PR #27528, opened 2022-03-12
Delete useless checks

kurnakovv:
I hope my changes make more readable code :)
* 16aa7a45c12f0c8167be9a277b2eccbf3f468bc0 This expression always be true, because we already have above this code null check, for example [here](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/signalr/Microsoft.Azure.Management.SignalR/src/Generated/SignalRManagementClient.cs#L206).
* 620e5affcdf10e6c6a257fca15176fd32359c13b This expression always be true, because [here](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/signalr/Microsoft.Azure.Management.SignalR/src/Generated/Models/SignalRFeature.cs#L123) we have already have check.


KurnakovMaksim:
Hmm, local tests in this solution have been passed.

KurnakovMaksim:
I'm active :)

KurnakovMaksim:
@m-nash do i need do something more? Can you say something, please? :)

Issue #27414, opened 2022-03-08
[BUG] Newtonsoft.Json default settings (JsonSerializerSettings) are ignored since version 1.7.0


ilya-git:
### Library name and version

Microsoft.Azure.WebJobs.Extensions.SignalRService 1.7.0

### Describe the bug

I have the following code to configure the json output for the signalr messages:

```csharp
JsonConvert.DefaultSettings = () =>
        {
            var settings = new JsonSerializerSettings
            {
                ContractResolver = new CamelCasePropertyNamesContractResolver()
            };
            settings.Converters.Add(new StringEnumConverter());
            return settings;
        };
```

In version 1.7.0-beta.1 this code worked properly and I got camel case and enum conversion. After upgrade to 1.7.0 it stopped working.

This looks like an undocumented breaking change that prevents us from upgrading the package.

### Expected behavior

There should be a way to configure JSON output. E.g.:

1. Allow for the default settings to work as before the 1.7.0
2. Allow for some configuration of newtonsoft.json or system.text.json via some callback
3. Allow to configure the `SignalRMessage` so that it does not serialize at all (and will allow only string arguments e.g.). This way users would be able to serialize however they like. Right now it is impossible since the package will encode all json.

I would personally prefer number 3. It would also probably mean least amount of work and configuration in the package. Everyone who needs custom serialization would be able to do as they please without the package providing complex mechanisms to do so.

### Actual behavior

No way to configure JSON output in 1.7.0.

Or if there is, I cannot find anything in either Changelog or the documentation.

### Reproduction Steps

Try to use the package and produce an output with e.g. enum converted as string. Since default is to use a number it seems impossible to change right now in version 1.7.0.

### Environment

net6.0, Azure function version 4, AzureSignalRServiceTransportType = transient

SatishBoddu-MSFT:
Hello @ilya-git 
What does your JSON message look like? any error message? or Debug log?
Does it break when you go back to the beta version (1.7.0-beta.1)?

ilya-git:
There is no error message, just my settings are ignored. So e.g. if I have the object with the only enum property called `Field`, it will be like this:

1.7.0-beta.1: {"field":"enumValue"}
17.0: {"Field":1}

because the json settings are not applied anymore.

Y-Sindo:
[...]

PR #27529, opened 2022-03-12
Add SignalR Functions extension contact

Y-Sindo:


.github/CODEOWNERS
```diff
@@ -575,6 +575,10 @@ sdk/trafficmanager/Microsoft.Azure.Management.TrafficManager/            @tmsupp
 # ServiceLabel: %SignalR %Service Attention
 /sdk/signalr/Microsoft.Azure.Management.SignalR/            @sffamily @chenkennt
 
+# ServiceLabel: %SignalR %Service Attention
```
jsquire, line 578:
```suggestion
```

jsquire: There can't be two annotations for the same "Service Attention" + label combination.  The bot is unable to consider paths; it just keys on the label combination.

Y-Sindo: I'd like to make issues like #27414 to be routed to me, but only path under "/sdk/signalr/Microsoft.Azure.WebJobs.Extensions.SignalRService/".  Can PR label do that?

jsquire: Unfortunately, it isn't possible to filter service attention by path.  The closest that you'd be able to achieve is to be part of the mention list for anything tagged `SignalR` and `Service Attention`.

jsquire (approved):

PR #27547, opened 2022-03-14
Add example of namespace imports

heaths:
Ideally, samples should also show importing namespaces. Updated the template to support this.


jsquire (approved):
Good idea!

PR #27551, opened 2022-03-15
Build snippets

### LINKED ISSUE #20432: Consider building samples with SNIPPET preprocessor directive defined
To verify that whatever gets embedded into READMEs and XML docs builds.

heaths:
Resolves #20432 for the build pipeline, and fixes resulting errors in the Key Vault SDKs.


m-nash (approved):
mgmt changes look good

eng/pipelines/templates/jobs/ci.yml
```diff
@@ -189,6 +189,17 @@ jobs:
            -SpellCheckPublicApiSurface
           pwsh: true
           failOnStderr: false
+      - script: >-
```
jsquire, line 192: Should we consider some kind of opt-in/out so that we can gradually transition everything that breaks, or are you thinking that we'd do it in one fell swoop?

eng/Directory.Build.Common.props
```diff
@@ -1,4 +1,4 @@
-﻿<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+﻿<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003" TreatAsLocalProperty="DefineConstants">
```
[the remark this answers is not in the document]
heaths: If you pass `/p:DefineConstants=SNIPPET` it becomes a global variable that cannot be overridden without this, so any project that tries to add to `DefineConstants` will no-op.

Speaking of which...I forgot to add the build step. 🤦‍♂️

eng/pipelines/templates/jobs/ci.yml
heaths: We're probably going to have to, if we even want to do this based on our internal Teams discussion.

heaths:
@weshaggard @benbp you ever seen an error like this?

```
Error : Unable to locate repository with working directory that contains directory '/mnt/vss/_work/1/s/sdk/core/Azure.Core/src'
```

Given it's under the working source directory, how could it not have an identifiable git repo? Running this same command locally works just fine.

benbp:
> Given it's under the working source directory, how could it not have an identifiable git repo? Running this same command locally works just fine.

Do you have a pipeline link example? My suspicion would be the tool is running in a context where we have only done sparse checkout for `/eng`.

heaths:
Thanks @benbp, you're right: https://dev.azure.com/azure-sdk/public/_build/results?buildId=1433496&view=logs&j=b70e5e73-bbb6-5567-0939-8415943fadb9&t=2c4000de-e41e-51f5-d969-886b675842ec

I'll have to see if I can either disable what wants a git repo or perhaps move it to the build job. Or perhaps I could just switch to a shallow checkout so we still get a git repo but should impact the timing very little. The "Analyze" job is already so quick, and this does have to do with ensuring documentation (specifically code snippets in documentation) is correct, so this made sense conceptually.

eng/Directory.Build.Common.props
weshaggard, line 1: Can you give more info on why you need this? I don't like adding these in the common props because it will block folks adding additional ones in their projects.

eng/pipelines/templates/jobs/ci.yml
```diff
@@ -189,6 +192,19 @@ jobs:
...
+          dotnet build eng/service.proj -warnaserror
+          /t:rebuild
+          /p:DebugType=none
+          /p:SDKType=${{ parameters.SDKType }}
+          /p:ServiceDirectory=${{ parameters.ServiceDirectory }}
+          /p:IncludePerf=false
+          /p:IncludeStress=false
+          /p:PublicSign=false
+          /p:Configuration=$(BuildConfiguration)
+          /p:DefineConstants=SNIPPET
```
weshaggard, line 205: I think I would pass another property that would add SNIPPET to the define list that already exists instead of overriding DefineConstants for everyone.

eng/pipelines/templates/jobs/ci.yml
```diff
@@ -17,6 +17,9 @@ parameters:
 - name: ServiceToTest
   type: string
   default: ''
+- name: TestDocSnippets
```
weshaggard, line 20: Who will pass this value as true?

eng/Directory.Build.Common.props
heaths: If you pass `/p:DefineConstants=SNIPPET` it becomes a global variable that cannot be overridden without this, so any project that tries to add to `DefineConstants` will no-op.

Alternatively, I could add some other condition and arbitrary property to instead use an in-props `DefineConstants` if you'd prefer, but all this does is allow global variables to be overridden. IMO, passing constants via command line can be useful to conditionalize builds so there's little to no harm in this.

heaths: ...which is what I see you suggested below, so I'll just do that instead. Still, I think DefineConstants is one of those that would be safe, but I don't feel strongly about it.

jsquire (approved):

PR #27556, opened 2022-03-15
Init Azure.ResourceManager.HybridConnectivity

ArthurMa1978:


sdk/hybridconnectivity/Azure.ResourceManager.HybridConnectivity/README.md
```diff
@@ -0,0 +1,77 @@
+# Azure hybridconnectivity Management client library for .NET
```
Yao725, line 1: This should be `HybridConnectivity`.

sdk/hybridconnectivity/Azure.ResourceManager.HybridConnectivity/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+This package follows the [new Azure SDK guidelines](https://azure.github.io/azure-sdk/general_introduction.html) which provide a number of core capabilities that are shared amongst all Azure SDKs, including the intuitive Azure Identity library, an HTTP Pipeline with custom policies, error-handling, distributed tracing, and much more.
+
+## Getting started 
+
+### Install the package
+
+Install the Azure hybridconnectivity management library for .NET with [NuGet](https://www.nuget.org/):
```
Yao725, line 9: Same as above.

sdk/hybridconnectivity/Azure.ResourceManager.HybridConnectivity/src/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Runtime.CompilerServices;
+
+[assembly: Azure.Core.AzureResourceProviderNamespace("hybridconnectivity")]
```
Yao725, line 6: I think the namespace should be `Microsoft.HybridConnectivity`

sdk/hybridconnectivity/Azure.ResourceManager.HybridConnectivity/src/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,8 @@
...
+
+[assembly: InternalsVisibleTo("Azure.ResourceManager.HybridConnectivity.Tests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100d15ddcb29688295338af4b7686603fe614abd555e09efba8fb88ee09e1f7b1ccaeed2e8f823fa9eef3fdd60217fc012ea67d2479751a0b8c087a4185541b851bd8b16f8d91b840e51b1cb0ba6fe647997e57429265e85ef62d565db50a69ae1647d54d7bd855e4db3d8a91510e5bcbd0edfbbecaa20a7bd9ae74593daa7b11b4")]
```
Yao725, line 8: Please also add this line here:
```
[assembly: InternalsVisibleTo("DynamicProxyGenAssembly2, PublicKey=0024000004800000940000000602000000240000525341310004000001000100c547cac37abd99c8db225ef2f6c8a3602f3b3606cc9891605d02baa56104f4cfc0734aa39b93bf7852f7d9266654753cc297e7d2edfe0bac1cdcf9f717241550e0a7b191195b7667bb4f64bcb8e2121380fd1d9d46ad2d92d2d15605093924cceaf74c4861eff62abf69b9291ed0a340e113be11e6a7d3113e92484cf7045cc7")]
```

ArthurMa1978: Should this be in template

sdk/hybridconnectivity/test-resources.json
```diff
@@ -0,0 +1,6 @@
+{
```
benbp, line 1: FYI this file is unnecessary unless you add a `tests.yml` file for live testing.

sdk/hybridconnectivity/Azure.ResourceManager.HybridConnectivity/README.md
```diff
@@ -0,0 +1,77 @@
+# Azure HybridConnectivity Management client library for .NET
```
ArcturusZhang, line 1: Should we separate the word HybridConnectivity?
```suggestion
# Azure Hybrid Connectivity Management client library for .NET
```

sdk/hybridconnectivity/Azure.ResourceManager.HybridConnectivity/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+This package follows the [new Azure SDK guidelines](https://azure.github.io/azure-sdk/general_introduction.html) which provide a number of core capabilities that are shared amongst all Azure SDKs, including the intuitive Azure Identity library, an HTTP Pipeline with custom policies, error-handling, distributed tracing, and much more.
+
+## Getting started 
+
+### Install the package
+
+Install the Azure HybridConnectivity management library for .NET with [NuGet](https://www.nuget.org/):
```
ArcturusZhang, line 9:
```suggestion
Install the Azure Hybrid Connectivity management library for .NET with [NuGet](https://www.nuget.org/):
```

sdk/hybridconnectivity/Azure.ResourceManager.HybridConnectivity/src/Generated/EndpointResource.cs
```diff
@@ -0,0 +1,284 @@
...
+{
+    /// <summary> A Class representing a EndpointResource along with the instance operations that can be performed on it. </summary>
+    public partial class EndpointResource : ArmResource
+    {
+        /// <summary> Generate the resource identifier of a <see cref="EndpointResource"/> instance. </summary>
+        public static ResourceIdentifier CreateResourceIdentifier(string resourceUri, string endpointName)
+        {
+            var resourceId = $"{resourceUri}/providers/Microsoft.HybridConnectivity/endpoints/{endpointName}";
+            return new ResourceIdentifier(resourceId);
+        }
```
ArcturusZhang, line 29: We should rename the parameter `resourceUri` to `resourceId` or `scope` since this is actually a resource Id of another resource instead of a URI

ArcturusZhang: Another question for this parameter is that do we have any knowledge about what kind of resource can be the scope of this resource? Is there a limited list of valid resource types? Or this resource could apply to any resource type, like subscription, resource group, virtual machine, virtual network, etc?

sdk/hybridconnectivity/Azure.ResourceManager.HybridConnectivity/src/Generated/Models/EndpointAccessResource.cs
```diff
@@ -0,0 +1,44 @@
...
+// Licensed under the MIT License.
+
+// <auto-generated/>
+
+#nullable disable
+
+namespace Azure.ResourceManager.HybridConnectivity.Models
+{
+    /// <summary> The endpoint access for the target resource. </summary>
+    public partial class EndpointAccessResource
```
ArcturusZhang, line 11: This model is not really a resource - therefore could we drop the suffix from this model?

sdk/hybridconnectivity/Azure.ResourceManager.HybridConnectivity/src/autorest.md
```diff
@@ -0,0 +1,40 @@
...
+  VMScaleSet: VmScaleSet
+  DNS: Dns
+  VPN: Vpn
+  NAT: Nat
+  WAN: Wan
+  Ipv4: IPv4
+  Ipv6: IPv6
+  Ipsec: IPsec
+  URI: Uri
+  TypePropertiesType: EndpointType
```
ArcturusZhang, line 34: We should not use this dictionary to transform names of property - could we change this to a directive? You can add `x-ms-client-name` to the corresponding property to do this

ArthurMa1978: This name `TypePropertiesType` is generated by codegen which is cased by flatten

sdk/hybridconnectivity/Azure.ResourceManager.HybridConnectivity/src/Generated/Models/EndpointAccessResource.cs
ArthurMa1978: good question, to this kind of non-resource model that swagger defined, what can we do? Should we have a generic rule to remove the suffix? Here the original name 'EndpointAccessResource' is not good, but remove the suffix also not a good name...

ArthurMa1978: Since the `resource` suffix change is ongoing, I'd like to level it as is now

sdk/hybridconnectivity/Azure.ResourceManager.HybridConnectivity/src/Generated/EndpointResource.cs
ArthurMa1978: @archerzz can you help us to understand how does codegen generate this method which is not same as other RPs

ArthurMa1978: Here `resourceUri` is a real `uri` so change it to `id` is not current

Yao725: I think maybe `scope` is better so we can keep align with `Deployment`.

Yao725 (approved):

sdk/hybridconnectivity/Azure.ResourceManager.HybridConnectivity/src/Generated/EndpointResource.cs
ArthurMa1978: `scope` is right, but can't fix by use `directive`

ArthurMa1978: will try later

PR #27363, opened 2022-03-04
[DO NOT MERGE] Experiment Metrics Advisor - Add Grow-up methods

ShivangiReja:
As an experiment to confirm that DPG can grow to a meaningful SDK, we want to experiment on Metrics Advisor:

* Generate DPG of Metrics Advisor
   * [Here](https://github.com/ShivangiReja/azure-sdk-for-net/tree/MetricsAdvisor-Experiment/sdk/metricsadvisor/Azure.AI.MetricsAdvisor/src/Generated) is the generated code for DPG
* Add grow up methods for all the operations mentioned [here](https://github.com/Azure/autorest.csharp/issues/1943#issuecomment-1051260711). 
* Run sample tests and live tests for all the operations mentioned [here](https://github.com/Azure/autorest.csharp/issues/1943#issuecomment-1051260711).

Adding the links for the grow up methods for all the operations mentioned [here](https://github.com/Azure/autorest.csharp/issues/1943#issuecomment-1051260711):
* [MetricsAdvisorClient](https://github.com/ShivangiReja/azure-sdk-for-net/blob/MetricsAdvisor-Experiment/sdk/metricsadvisor/Azure.AI.MetricsAdvisor/src/MetricsAdvisorClient.cs)
     * [GetAllFeedback](https://github.com/ShivangiReja/azure-sdk-for-net/blob/MetricsAdvisor-Experiment/sdk/metricsadvisor/Azure.AI.MetricsAdvisor/src/MetricsAdvisorClient.cs#L138-L208)
     * [AddFeedback](https://github.com/ShivangiReja/azure-sdk-for-net/blob/MetricsAdvisor-Experiment/sdk/metricsadvisor/Azure.AI.MetricsAdvisor/src/MetricsAdvisorClient.cs#L220-L302)
     * [GetFeedback](https://github.com/ShivangiReja/azure-sdk-for-net/blob/MetricsAdvisor-Experiment/sdk/metricsadvisor/Azure.AI.MetricsAdvisor/src/MetricsAdvisorClient.cs#L314-L370)
     * [GetMetricDimensionValues](https://github.com/ShivangiReja/azure-sdk-for-net/blob/MetricsAdvisor-Experiment/sdk/metricsadvisor/Azure.AI.MetricsAdvisor/src/MetricsAdvisorClient.cs#L71-L121)
[...]

PR #27559, opened 2022-03-15
Add developer driven evolution document

pshao25:
Addresses https://github.com/Azure/autorest.csharp/issues/1945

### DIFF
--- a/doc/Data Plane Code Generation/DeveloperDrivenEvolution.md
+++ b/doc/Data Plane Code Generation/DeveloperDrivenEvolution.md
@@ -185,10 +185,10 @@ namespace Azure.AI.MetricsAdvisor
         public virtual AsyncPageable<MetricFeedback> GetMetricFeedbacksValuesAsync(Guid feedbackId, CancellationToken cancellationToken = default)
         {
             // Call protocol method
-            AsyncPageable<BinaryData> pageableBinadryData = GetMetricFeedbacksAsync(feedbackId, new RequestContext() { CancellationToken = cancellationToken });
+            AsyncPageable<BinaryData> pageableBinaryData = GetMetricFeedbacksAsync(feedbackId, new RequestContext() { CancellationToken = cancellationToken });
 
             // Calling deserialization helper
-            return PageableHelpers.Select(pageableBinadryData, response => ConvertToDataFeeds(DataFeedList.FromResponse(response).Value));
+            return PageableHelpers.Select(pageableBinaryData, response => ConvertToDataFeeds(DataFeedList.FromResponse(response).Value));
         }
     }
 }
--- a/doc/Data Plane Code Generation/DeveloperDrivenEvolution.md
+++ b/doc/Data Plane Code Generation/DeveloperDrivenEvolution.md
@@ -1,5 +1,5 @@
 # Developer driven evolution
-Convenience API is an additive layer on top of protocol mathod so it could produce improved user experience.
+Convenience API is an additive layer on top of protocol method so it could produce improved user experience.
 
 To add a convenience API that takes and returns strongly-typed input and output models instead of raw JSON, steps are:
 1. [Create the models](#create-the-models)
--- a/doc/Data Plane Code Generation/DeveloperDrivenEvolution.md
+++ b/doc/Data Plane Code Generation/DeveloperDrivenEvolution.md
@@ -2,18 +2,46 @@
 Convenience API is an additive layer on top of protocol mathod so it could produce improved user experience.
 
 To add a convenience API that takes and returns strongly-typed input and output models instead of raw JSON, steps are:
-1. [Create the convenience method with new signature](#create-the-convenience-method-with-new-signature)
+1. [Create the models](#create-the-models)
+2. [Create the convenience method with new signature](#create-the-convenience-method-with-new-signature)
 - [Replace input parameters](#replace-input-parameters)
 - [Replace output value](#replace-output-value)
 - [Rename method if needed](#rename-method-if-needed)
-2. [Create the models](#create-the-models)
 3. [Implement the method](#implement-the-method)
 - [Implement GET method](#implement-get-method)
 - [Implement POST method](#implement-post-method)
 - [Implement GET paging method](#implement-get-paging-method)
 
+## Create the models
+Currently the models should be created manually. We will support generate it automatically. So, to avoid the models overwritten, we suggest putting the models under folder {sdk folder}/Models which is the same layer as folder {sdk folder}/Generated.
+
+After you create the models you want, you also need to add the helper methods mapping between raw response and model. Examples for model `MetricFeedback` (Models/MetricFeedback/MetricFeedback.cs) are:
+```C#
+namespace Azure.AI.MetricsAdvisor
+{
+    public partial class MetricFeedback
+    {
+        // Mapping raw response to model
+        internal static MetricFeedback FromResponse(Response response)
+        {
+            using var document = JsonDocument.Parse(response.Content);
+            // Add your deserialization logic in DeserializeMetricFeedback
+            return DeserializeMetricFeedback(document.RootElement);
+        }
+
+        // Mapping model to raw request
+        internal static RequestContent ToRequestContent(MetricFeedback metricFeedback)
+        {
+            var content = new Utf8JsonRequestContent();
+            content.JsonWriter.WriteObjectValue(metricFeedback);
+            return content;
+        }
+    }
+}
+```
+
 ## Create the convenience method with new signature
-To create a convenience method on top of the protocol method, you should not modify generated code. The convenience methods should be added in a saparate file as partial class of the class containing protocol methods. You could check out the full examples in [Implement the method](#implement-the-method).
+To create a convenience method on top of the protocol method, you should not modify generated code. The convenience methods should be added in a separate file as partial class of the class containing protocol methods. You could check out the full examples in [Implement the method](#implement-the-method).
 
 Below is how you could create the convenience signature from protocol method. Here is an example of your generated protocol method.
 ```C#
@@ -67,33 +95,7 @@ public virtual async Task<Response<MetricFeedback>> GetMetricFeedbackValuesAsync
 **If this name doesn't make sense, pick the best name**
 
 If the suggested rename does not apply to your scenario, you could pick the best name fitting your API.
-## Create the models
-Currently the models should be created manually. We will support generate it automatically. So, to avoid the models overwritten, we suggest putting the models under folder {sdk folder}/Models which is the same layer as folder {sdk folder}/Generated.
 
-After you create the models, you also need to add the helper methods mapping between raw response and model. Examples for model `MetricFeedback` (Models/MetricFeedback/MetricFeedback.cs) are:
-```C#
-namespace Azure.AI.MetricsAdvisor
-{
-    public partial class MetricFeedback
-    {
-        // Mapping raw response to model
-        internal static MetricFeedback FromResponse(Response response)
-        {
-            using var document = JsonDocument.Parse(response.Content);
-            // Add your deserialization logic in DeserializeMetricFeedback
-            return DeserializeMetricFeedback(document.RootElement);
-        }
-
-        // Mapping model to raw request
-        internal static RequestContent ToRequestContent(MetricFeedback metricFeedback)
-        {
-            var content = new Utf8JsonRequestContent();
-            content.JsonWriter.WriteObjectValue(metricFeedback);
-            return content;
-        }
-    }
-}
-```
 ## Implement the method
 You could call the protocol method inside your convenience method and map the returned raw response to model. See different scenarios and corresponding examples below. 
 
@@ -111,7 +113,7 @@ namespace Azure.AI.MetricsAdvisor
     }
 }
 ```
-**Improve the GET method (MetricsAdvisorClient.cs):**
+**Improve the GET method ({Other manual folder}/MetricsAdvisorClient.cs):**
 ``` C#
 namespace Azure.AI.MetricsAdvisor
 {
@@ -143,7 +145,7 @@ namespace Azure.AI.MetricsAdvisor
     }
 }
 ```
-**Improve the POST method (MetricsAdvisorClient.cs):**
+**Improve the POST method ({Other manual folder}/MetricsAdvisorClient.cs):**
 ``` C#
 namespace Azure.AI.MetricsAdvisor
 {
@@ -179,7 +181,7 @@ namespace Azure.AI.MetricsAdvisor
     }
 }
 ```
-**Improve the GET paging method (MetricsAdvisorClient.cs):**
+**Improve the GET paging method ({Other manual folder}/MetricsAdvisorClient.cs):**
 ``` C#
 namespace Azure.AI.MetricsAdvisor
 {
--- a/doc/Data Plane Code Generation/DeveloperDrivenEvolution.md
+++ b/doc/Data Plane Code Generation/DeveloperDrivenEvolution.md
@@ -1,22 +1,72 @@
 # Developer driven evolution
-To add a convenience API that returns raw JSON to a model, steps are:
+To add a convenience API that takes and returns strongly-typed input and output models instead of raw JSON, steps are:
+1. [Create the convenience method with new signature](#create-the-convenience-method-with-new-signature)
+- [Replace input parameters](#replace-input-parameters)
+- [Replace output value](#replace-output-value)
+- [Rename method if needed](#rename-method-if-needed)
+2. [Create the models](#create-the-models)
+3. [Implement the method](#implement-the-method)
+- [Implement GET method](#implement-get-method)
+- [Implement POST method](#implement-post-method)
+- [Implement GET paging method](#implement-get-paging-method)
 
-**Get convenience method signiture from protocol method signiture.** Replace the optional `RequestContext` parameter with an optional `CancellationToken`. If there is an input `RequestContent`, replace it with a model type `T`. Replace the return value `Response` with `Response<T>`.
+## Create the convenience method with new signature
+You should first decide how the convenience method grows up from protocol method. Below is an example of your generated protocol method.
 ```C#
-- public virtual async Task<Response> CreateMetricFeedbackAsync(RequestContent content, RequestContext context = null);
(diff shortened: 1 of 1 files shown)


doc/Data Plane Code Generation/DeveloperDrivenEvolution.md
```diff
@@ -0,0 +1,144 @@
...
+        {
+        }
+    }
+}
+```
+**Improve the GET method**
+``` C#
+namespace Azure.AI.MetricsAdvisor
+{
+    public partial class MetricsAdvisorClient
```
annelo-msft, line 56: We should be clear that this is a different file -- any code developers add to the generated client will be deleted when the client is generated.

doc/Data Plane Code Generation/DeveloperDrivenEvolution.md
```diff
@@ -0,0 +1,144 @@
...
+    {
+        // Suggest appending value to method name
+        public virtual async Task<Response<MetricFeedback>> GetMetricFeedbackValueAsync(Guid feedbackId, CancellationToken cancellationToken = default)
+        {
+            // Call protocol method
```
annelo-msft, line 61: We will want to add a DiagnosticScope outside the method, as a result of this discussion: https://github.com/Azure/azure-sdk-for-net/pull/27363#discussion_r824997122

We will use whatever mechanism is developed here to suppress the trace from the inner DPG client: https://github.com/azure/autorest.csharp/issues/2069

doc/Data Plane Code Generation/DeveloperDrivenEvolution.md
```diff
@@ -0,0 +1,144 @@
...
+- Adding a suffix `Values` to the initial method name as the grow-up method name, if the initial method is plural (e.g., `GetMetricFeedbacksAsync`).
+
+**Add implict casting or helper method in generated or handcrafted model.** Examples for model `MetricFeedback` (Models/MetricFeedback/MetricFeedback.cs) are:
+```C#
+namespace Azure.AI.MetricsAdvisor
+{
+    public partial class MetricFeedback
+    {
+        // Add implicit casting
+        public static implicit operator MetricFeedback(Response response)
```
annelo-msft, line 19: Let's not show model casting in this document, since we don't need it for the kind of convenience APIs we'll add in the GA timeframe.

doc/Data Plane Code Generation/DeveloperDrivenEvolution.md
```diff
@@ -0,0 +1,144 @@
...
+        {
+            // Add your deserialization logic in DeserializeMetricFeedback
+            return DeserializeMetricFeedback(JsonDocument.Parse(response.Content.ToMemory()).RootElement);
+        }
+
+        // Add helper method 
+        internal static MetricFeedback FromResponse(Response response)
```
annelo-msft, line 26: It would be good to have one for each of FromResponse and ToRequestContext here, such as: https://github.com/ShivangiReja/azure-sdk-for-net/blob/MetricsAdvisor-Experiment/sdk/metricsadvisor/Azure.AI.MetricsAdvisor/src/Models/DataFeedDetail.Serialization.cs#L487-L498

annelo-msft (commented):
I think we can restructure this to simplify it a bit.  Instead of enumerating GET/PUT/POST/etc., what if we gave the rules for writing a convenience API?

- Replace the RequestContext parameter with CancellationToken
- Replace the input parameter RequestContent with a model type
- Replace the return value Response with Response\<T\>
- If you end up with an ambiguous call-site, rename the method
  - Add suffix Value or Values
  - If this isn't a good name, pick a good name for the method
- Create the models for the convenience APIs you've added

doc/Data Plane Code Generation/DeveloperDrivenEvolution.md
```diff
@@ -0,0 +1,144 @@
+# Developer driven evolution
+To grow up a protocol method that returns raw JSON to a model, steps are:
```
annelo-msft, line 2: Instead of "growing up a protocol method," let's give this its formal name, "adding a convenience API."

pshao25:
> I think we can restructure this to simplify it a bit. Instead of enumerating GET/PUT/POST/etc., what if we gave the rules for writing a convenience API?
> 
> * Replace the RequestContext parameter with CancellationToken
> * Replace the input parameter RequestContent with a model type
> * Replace the return value Response with Response<T>
> * If you end up with an ambiguous call-site, rename the method
>   
>   * Add suffix Value or Values
>   * If this isn't a good name, pick a good name for the method
> * Create the models for the convenience APIs you've added

Yes, I do follow this by giving rules, and I added more rules as you listed above. Enumerating GET/PUT/POST/etc. is just examples for how to improve to grow up method.

doc/Data Plane Code Generation/DeveloperDrivenEvolution.md
```diff
@@ -0,0 +1,145 @@
+# Developer driven evolution
+To add a convenience API that returns raw JSON to a model, steps are:
+
+**Get convenience method signiture from protocol method signiture.** Replace the optional `RequestContext` parameter with an optional `CancellationToken`. If there is an input `RequestContent`, replace it with a model type `T`. Replace the return value `Response` with `Response<T>`.
+```C#
+- public virtual async Task<Response> CreateMetricFeedbackAsync(RequestContent content, RequestContext context = null);
++ public virtual async Task<Response<MetricFeedback>> CreateMetricFeedbackAsync(MetricFeedback feedback, CancellationToken cancellationToken = default);
+```
+**Pick the best convenience method name fitting your scenario.** You could use the same method name as the initial protocol method only when your convenience method has different **required** parameter list. E.g.,
```
annelo-msft, line 9: I think partner service teams following this guidance will think we are telling them to "pick a new name for the method."  

Let's include this as a sub-bullet of "Get convenience method signature," where the steps are:
1. Replace input parameters
2. Replace output value
3. Rename method only if needed
    - Only needed if ambiguous call site
    - Suggested rename for most cases
    - If this name doesn't make sense, pick the best name

It would be better in the explanation to show full method signatures rather than parameter lists - it's unclear what the parameter lists are trying to describe.

doc/Data Plane Code Generation/DeveloperDrivenEvolution.md
```diff
@@ -0,0 +1,145 @@
...
+- ```(string feedId, RequestContext requestContext = null)``` and ```(string feedId, CancellationToken cancellationToken = default)``` have the same required parameter list. You cannot add the convenience method as an overload method by leveraging the initial method name, because it will cause an ambiguous compile error when you call it without `RequestContext`/`CancellationToken`.
+- ```(RequestContent content, RequestContext requestContext = null)``` and ```(string feedId, CancellationToken cancellationToken = default)``` have different required parameter list. You are safe to use same method name with the initial protocol name.
+
+If your convenience method has the same parameter list as that of the protocol method, we suggest 
+- Adding a suffix `Value` to the initial method name as the convenience method name, if the initial method is singular (e.g., `GetMetricFeedbackAsync`).
+- Adding a suffix `Values` to the initial method name as the convenience method name, if the initial method is plural (e.g., `GetMetricFeedbacksAsync`).
+
+**Create the models for the convenience APIs you've added.** Currently it can be generated manually. We will support generate it automatically.
```
annelo-msft, line 17: This is basically step two, where the steps are:
1. Create the convenience method with new signature
2. Create the models
3. Implement the method

It would be good to clarify that this is the outline

doc/Data Plane Code Generation/DeveloperDrivenEvolution.md
```diff
@@ -0,0 +1,145 @@
...
+            content.JsonWriter.WriteObjectValue(metricFeedback);
+            return content;
+        }
+    }
+}
+```
+**Call the protocol method and map returned raw response to model.** You could call the protocol method inside your convenience method and map the returned raw response to model. See different scenarios and corresponding examples below. 
+
+
+## Improve a GET method that returns raw JSON to return a model
```
annelo-msft, line 46: It is fine to provide these as examples, but right now the outline of the article is structured so that it looks like these are the main point and not supporting examples.  Let's please fix this.

doc/Data Plane Code Generation/DeveloperDrivenEvolution.md
```diff
@@ -0,0 +1,145 @@
...
+``` C#
+namespace Azure.AI.MetricsAdvisor
+{
+    public partial class MetricsAdvisorClient
+    {
+        // Suggest appending values to a plural method name
+        public virtual AsyncPageable<MetricFeedback> GetMetricFeedbacksValuesAsync(Guid feedbackId, CancellationToken cancellationToken = default)
+        {
+            // Call protocol method
+            AsyncPageable<BinaryData> pageableBindaryData = GetMetricFeedbacksAsync(feedbackId, new RequestContext() { CancellationToken = cancellationToken });
```
annelo-msft, line 103: Nit:

```suggestion
            AsyncPageable<BinaryData> pageableBinaryData = GetMetricFeedbacksAsync(feedbackId, new RequestContext() { CancellationToken = cancellationToken });
```

doc/Data Plane Code Generation/DeveloperDrivenEvolution.md
```diff
@@ -0,0 +1,145 @@
...
+{
+    public partial class MetricsAdvisorClient
+    {
+        public virtual AsyncPageable<BinaryData> GetMetricFeedbacksAsync(Guid feedbackId, RequestContext context = null)
+        {
+        }
+    }
+}
+```
+**Improve the GET paging method (MetricsAdvisorClient.cs):**
```
annelo-msft, line 93: Let's move the paging example after the GET and POST examples, since paging is more complicated.  Is this all they'll need to do?

annelo-msft (commented):
Thanks!

We'll want to make sure we sort out the diagnostic scope questions and provide guidance for that before we GA -- @pshao25, would you create a separate issue for this, so we don't block this PR on that?

pshao25:
> diagnostic scope

pshao25:
> Thanks!
> 
> We'll want to make sure we sort out the diagnostic scope questions and provide guidance for that before we GA -- @pshao25, would you create a separate issue for this, so we don't block this PR on that?

Yes, created here: https://github.com/Azure/autorest.csharp/issues/2115

doc/Data Plane Code Generation/DeveloperDrivenEvolution.md
pshao25: Yes, the complicated logic is in PageableHelpers.Select.

doc/Data Plane Code Generation/DeveloperDrivenEvolution.md
```diff
@@ -0,0 +1,195 @@
...
+- [Implement GET method](#implement-get-method)
+- [Implement POST method](#implement-post-method)
+- [Implement GET paging method](#implement-get-paging-method)
+
+## Create the convenience method with new signature
+You should first decide how the convenience method grows up from protocol method. Below is an example of your generated protocol method.
+```C#
+public virtual async Task<Response> CreateMetricFeedbackAsync(RequestContent content, RequestContext context = null);
+```
+You should do the following steps to create the convenience method signature:
```
lirenhe, line 18: Could you explain more on this?  Are we expecting developer to modify the existing method or to create a new method with the same name but different input/output?

doc/Data Plane Code Generation/DeveloperDrivenEvolution.md
```diff
@@ -0,0 +1,195 @@
...
+### Replace input parameters
+- Replace the parameter `RequestContext` with an optional `CancellationToken` with default value.
+- If there is an input parameter `RequestContent`, replace it with a wanted input type (e.g., `MetricFeedback`).
```
lirenhe, line 21: How can we generate the model for 'MetricFeedback'?

lirenhe: And what is the folder structure for the manual code to avoid them to be overwritten when we call codegen again?

doc/Data Plane Code Generation/DeveloperDrivenEvolution.md
```diff
@@ -0,0 +1,195 @@
...
+        // Mapping raw response to model
+        internal static MetricFeedback FromResponse(Response response)
+        {
+            using var document = JsonDocument.Parse(response.Content);
+            // Add your deserialization logic in DeserializeMetricFeedback
+            return DeserializeMetricFeedback(document.RootElement);
+        }
+
+        // Mapping model to raw request
+        internal static RequestContent ToRequestContent(MetricFeedback metricFeedback)
```
lirenhe, line 84: Do we need to write 2 conversion methods for each model?   Can we leverage template to reduce the effort?

doc/Data Plane Code Generation/DeveloperDrivenEvolution.md
```diff
@@ -0,0 +1,195 @@
...
+            Response response = await GetMetricFeedbackAsync(feedbackId, new RequestContext() { CancellationToken = cancellationToken });
+
+            // Calling deserialization helper
+            MetricFeedback value = MetricFeedback.FromResponse(response);
+        }
+    }
+}
+```
+
+### Implement POST method
```
annelo-msft, line 129: This doesn't illustrate that the client API before is:

```csharp
namespace Azure.AI.MetricsAdvisor {
       public class MetricsAdvisorClient {
           // ...
           public virtual async Task<Response> CreateMetricFeedbackAsync(RequestContent content, RequestContext context = null);
           // ...
           }
    }
```

And after the addition of the grow-up method it is:

```csharp
namespace Azure.AI.MetricsAdvisor {
       public class MetricsAdvisorClient {
           // ...
           public virtual async Task<Response> CreateMetricFeedbackAsync(RequestContent content, RequestContext context = null);
public virtual async Task<Response<MetricFeedback>> CreateMetricFeedbackAsync(MetricFeedback feedback, CancellationToken cancellationToken = default);
           // ...
           }
    }
```

Please make this clear.  

The section https://github.com/Azure/autorest.csharp#change-model-namespace-or-accessibility-in-bulk is a nice example of this, with diff-color highlighting

image

doc/Data Plane Code Generation/DeveloperDrivenEvolution.md
annelo-msft: We do need one for serialization and one for deserialization.

doc/Data Plane Code Generation/DeveloperDrivenEvolution.md
annelo-msft: They should not modify generated code.  We use partial classes to add grow-up methods in a separate file that will not be overwritten by the generator.  This information should be added so partners know where to add the grow-up methods.

doc/Data Plane Code Generation/DeveloperDrivenEvolution.md
pshao25: > How can we generate the model for 'MetricFeedback'?

How about we put section [Create the models](#create-the-models) as the first step?

doc/Data Plane Code Generation/DeveloperDrivenEvolution.md
```diff
@@ -0,0 +1,199 @@
...
+- [Replace output value](#replace-output-value)
+- [Rename method if needed](#rename-method-if-needed)
+2. [Create the models](#create-the-models)
+3. [Implement the method](#implement-the-method)
+- [Implement GET method](#implement-get-method)
+- [Implement POST method](#implement-post-method)
+- [Implement GET paging method](#implement-get-paging-method)
+
+## Create the convenience method with new signature
+To create a convenience method on top of the protocol method, you should not modify generated code. The convenience methods should be added in a saparate file as partial class of the class containing protocol methods. You could check out the full examples in [Implement the method](#implement-the-method).
```
lirenhe, line 16: typo:  separate

doc/Data Plane Code Generation/DeveloperDrivenEvolution.md
lirenhe: It is better to tell what 'MetricFeedback' is before referencing this model.

doc/Data Plane Code Generation/DeveloperDrivenEvolution.md
```diff
@@ -0,0 +1,199 @@
...
+// protocol method
+public virtual async Task<Response> GetMetricFeedbackAsync(string feedId, RequestContext context = null);
+// Convenience method
+public virtual async Task<Response<MetricFeedback>> GetMetricFeedbackAsync(string feedId, CancellationToken cancellationToken = default);
+```
+Ambiguous call exists when you call
+```C#
+GetMetricFeedbackAsync(feedId);
+```
+In this situation, you have to change the convenience name.
```
lirenhe, line 54: Do we have a rule about what is the acceptable name for the 'convenience name'?
In this case, how can i rename 'GetMetricFeedbackAsync'

doc/Data Plane Code Generation/DeveloperDrivenEvolution.md
```diff
@@ -0,0 +1,199 @@
...
+{
+    public partial class MetricsAdvisorClient
+    {
+        public virtual async Task<Response> CreateMetricFeedbackAsync(RequestContent content, RequestContext context = null)
+        {
+        }
+    }
+}
+```
+**Improve the POST method (MetricsAdvisorClient.cs):**
```
lirenhe, line 146: Could you highlight this is the manual written file?   is it better to use  './MetricsAdvisorClient.cs'  otherwise i thought this is the same file compared with previous one.

lirenhe (approved):
LGTM

doc/Data Plane Code Generation/DeveloperDrivenEvolution.md
```diff
@@ -0,0 +1,201 @@
+# Developer driven evolution
+Convenience API is an additive layer on top of protocol mathod so it could produce improved user experience.
```
lirenhe, line 2:
```suggestion
Convenience API is an additive layer on top of protocol method so it could produce improved user experience.
```

doc/Data Plane Code Generation/DeveloperDrivenEvolution.md
annelo-msft: I disagree that this is clear, but we can test it with partner teams referencing the docs and improve if necessary.

annelo-msft (approved):

PR #27577, opened 2022-03-15
Azure Monitor Exporter - Enable persistent storage for logs and metrics

rajkumar-rangaraj:
- Moved the logic for `TransmitFromStorage` from trace exporter to common class `AzureMonitorPersistentStorage`
- Enabled persistent storage for logs and metrics exporter

### DIFF
--- a/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/AzureMonitorLogExporter.cs
+++ b/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/AzureMonitorLogExporter.cs
@@ -16,6 +16,7 @@ namespace Azure.Monitor.OpenTelemetry.Exporter
         private readonly ITransmitter _transmitter;
         private readonly string _instrumentationKey;
         private readonly ResourceParser _resourceParser;
+        private readonly AzureMonitorPersistentStorage _persistentStorage;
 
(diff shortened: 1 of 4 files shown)


sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/AzureMonitorLogExporter.cs
```diff
@@ -26,11 +27,17 @@ internal AzureMonitorLogExporter(ITransmitter transmitter)
             _transmitter = transmitter;
             _instrumentationKey = transmitter.InstrumentationKey;
             _resourceParser = new ResourceParser();
+
+            // TODO: Add check if offline storage is enabled by user via options
+            // If disabled do not initialize AzureMonitorPersistentStorage
+            _persistentStorage = new AzureMonitorPersistentStorage(transmitter);
```
vishweshbankwar, line 33: Should we just add check here? if (_transmitter.storage != null)

vishweshbankwar (approved):

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/AzureMonitorLogExporter.cs
rajkumar-rangaraj: This is a nice idea and we will have all implementations within `AzureMonitorTransmitter`. Will send a follow up PR for this one.

Issue #27549, opened 2022-03-15
[BUG] `SearchOptions.SessionId` is not passed to the service


Mohit-Chakraborty:
### Library name and version

Azure.Search.Documents

### Describe the bug

When a user sets the `SessionId` in `SearchOptions` for a query, the service does not receive the property.

### Expected behavior

We expect to see `SessionId` passed to the service.

### Actual behavior

Service-side logs show that `SessionId` was not passed.

### Reproduction Steps

```
      private static void RunQueries(SearchClient srchclient)
      {
          SearchOptions options;
          SearchResults<Hotel> response;

          Console.WriteLine("Query #1: Search on empty term '*' to return all documents, showing a subset of fields...\n");

          options = new SearchOptions()
          {
              IncludeTotalCount = true,  
              SessionId = "123",
          };

          options.Select.Add("HotelId");
          options.Select.Add("HotelName");
          options.Select.Add("Rating");

          while (true)
          {
              response = srchclient.Search<Hotel>("*", options);
              WriteDocuments(response);
          }
      }
```

pallavit:
This issue was fixed.

PR #27582, opened 2022-03-16
[Search] Add missing property when cloning `SearchOptions`

Mohit-Chakraborty:
Copy all properties using reflection to alleviate the issue of missing some properties in the copy code.

### DIFF
--- a/sdk/search/Azure.Search.Documents/src/Options/SearchOptions.cs
+++ b/sdk/search/Azure.Search.Documents/src/Options/SearchOptions.cs
@@ -310,10 +310,33 @@ namespace Azure.Search.Documents
             Debug.Assert(source != null);
             Debug.Assert(destination != null);
 
-            foreach (System.Reflection.PropertyInfo propertyInfo in typeof(SearchOptions).GetProperties())
-            {
-                propertyInfo.SetValue(destination, propertyInfo.GetValue(source));
-            }
+            destination.Facets = source.Facets;
+            destination.Filter = source.Filter;
+            destination.HighlightFields = source.HighlightFields;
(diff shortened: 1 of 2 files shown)


Mohit-Chakraborty:
Issue #27549

Mohit-Chakraborty:
> Issue #27549

This is the category of issues I am trying to kill with this approach. It has happened a few times now. The current issue is for an old property, so the bug has existed for a while.

Mohit-Chakraborty:
cc: @tg-msft
What do you think of this solution? Alternatively, we could add the one we know is missing currently and be more diligent in the future.

JoshLove-msft:
It looks like these are all value types, would a [MemberwiseClone](https://docs.microsoft.com/en-us/dotnet/api/system.object.memberwiseclone?redirectedfrom=MSDN&view=net-6.0#System_Object_MemberwiseClone) work?

JoshLove-msft:
Another idea would be to just do the reflection in the unit test, and keep the production code up to date by hand. This way you will catch future missing properties automatically in the test, but won't have to use reflection in production code.

Mohit-Chakraborty:
> Another idea would be to just do the reflection in the unit test, and keep the production code up to date by hand. This way you will catch future missing properties automatically in the test, but won't have to use reflection in production code.

What would the test check? The count of the properties?

JoshLove-msft:
> > Another idea would be to just do the reflection in the unit test, and keep the production code up to date by hand. This way you will catch future missing properties automatically in the test, but won't have to use reflection in production code.
> 
> What would the test check? The count of the properties?

The test could check that each property on the source is equal to the corresponding property on the destination.

JoshLove-msft (approved):

Issue #27030, opened 2022-02-15
[QUERY] [EventHub] PartitionKey HashCode Generation


danielmarbach:
### Library name and version

Azure.Messaging.EventHubs

### Query/Question

When converting [PartitionResolver.ComputeHash](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Azure.Messaging.EventHubs/src/Core/PartitionResolver.cs#L123) to span and BinaryPrimitives it would be possible to squeeze out some perf. See the below benchmark

``` ini

BenchmarkDotNet=v0.13.1, OS=Windows 10.0.22000
11th Gen Intel Core i7-1185G7 3.00GHz, 1 CPU, 8 logical and 4 physical cores
.NET SDK=6.0.102
  [Host]     : .NET 6.0.2 (6.0.222.6406), X64 RyuJIT
  DefaultJob : .NET 6.0.2 (6.0.222.6406), X64 RyuJIT


```
|                  Method |     Mean |     Error |   StdDev |    StdErr |   Median |      Min |       Q1 |        Q3 |      Max |          Op/s | Ratio | RatioSD |
|------------------------ |---------:|----------:|---------:|----------:|---------:|---------:|---------:|----------:|---------:|--------------:|------:|--------:|
| ComputeHashBitConverter | 9.599 ns | 0.3469 ns | 1.007 ns | 0.1022 ns | 9.421 ns | 7.989 ns | 8.835 ns | 10.279 ns | 12.61 ns | 104,174,105.6 |  1.00 |    0.00 |
|         ComputeHashSpan | 8.792 ns | 0.4335 ns | 1.216 ns | 0.1274 ns | 8.444 ns | 6.987 ns | 7.875 ns |  9.409 ns | 12.60 ns | 113,736,404.7 |  0.93 |    0.16 |

Is that something you'd be interested in?

Another thing. I noticed currently it is required to allocate a byte array to compute the hash and the size of the byte array is proportional to the size of the partition key. I see the following options:

- Pool the buffer
- Stackalloc the buffer for short string and for longer pool the buffer
- If the `EnqueueEventOptions` are most likely to be reused per partition, it would be possible to turn the partition key into bytes there and the just pass those bytes into `ComputeHash` method (then the options act as sort of root for the byte array)

Just as an example, pooling the byte array there is slightly slower but gets rid of all the allocations on that path.

``` ini

[...]

PR #27598, opened 2022-03-16
[Event Hubs] Optimize partition resolver compute hash

danielmarbach:
Relates to https://github.com/Azure/azure-sdk-for-net/issues/27030

[Benchmark](https://github.com/danielmarbach/MicroBenchmarks/commit/d1c45dd7fe90326b7c49513c938fdcf078990c1e) uses the same string of various sizes (see size column in the screenshot) as a partition key.

## With SkipLocalsInit

image

## Without SkipLocalsInit

image


sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/EncodingExtensions.cs
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Text;
+
+namespace Azure.Messaging.EventHubs.Primitives
+{
+    internal static class EncodingExtensions
```
danielmarbach, line 9: These methods are actually available on NETStandard 2.1 and NetCoreApp. Should we ifdef here? If so any advises what types of conditions we would normally use?

Example

https://docs.microsoft.com/en-us/dotnet/api/system.text.encoding.getbytes?view=net-6.0#system-text-encoding-getbytes(system-readonlyspan((system-char))-system-span((system-byte)))

sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/SkipLocalsInitAttribute.cs
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
danielmarbach, line 1: The original file has https://github.com/dotnet/corert/blob/master/src/System.Private.CoreLib/shared/System/Runtime/CompilerServices/SkipLocalsInitAttribute.cs#L1 is it a problem if we change it? I'd say not because in order to shim they are expected to be copied.

sdk/eventhub/Azure.Messaging.EventHubs/src/Core/PartitionResolver.cs
```diff
@@ -92,15 +95,26 @@ public virtual string AssignRoundRobin(string[] partitions)
...
             if (partitionKey == null)
             {
                 return 0;
             }
 
-            ComputeHash(Encoding.UTF8.GetBytes(partitionKey), seed1: 0, seed2: 0, out uint hash1, out uint hash2);
+            byte[] sharedBuffer = null;
+            var partitionKeySpan = partitionKey.AsSpan();
+            var partitionKeyByteLength = Encoding.UTF8.GetByteCount(partitionKeySpan);
+            Span<byte> hashBuffer = partitionKeyByteLength <= 256 ? stackalloc byte[partitionKeyByteLength] :
```
danielmarbach, line 109: I used 256 as a max size similar to what Json does here

https://github.com/dotnet/runtime/blob/ab02bf6745a80d991ccfbdac0f4e029b9d16f14b/src/libraries/System.Text.Json/src/System/Text/Json/Reader/JsonReaderHelper.Unescaping.cs#L74-L76

since it is hard to come up with a good threshold

sdk/eventhub/Azure.Messaging.EventHubs/src/Core/PartitionResolver.cs
```diff
@@ -92,15 +95,26 @@ public virtual string AssignRoundRobin(string[] partitions)
...
+                sharedBuffer = ArrayPool<byte>.Shared.Rent(partitionKeyByteLength);
+
+            var written = Encoding.UTF8.GetBytes(partitionKeySpan, hashBuffer);
+            ComputeHash(hashBuffer.Slice(0, written), seed1: 0, seed2: 0, out uint hash1, out uint hash2);
+            if (sharedBuffer != null)
+            {
+                ArrayPool<byte>.Shared.Return(sharedBuffer);
```
danielmarbach, line 116: I assumed we don't bother clearing the buffer here

sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/SkipLocalsInitAttribute.cs
danielmarbach: Given SkipLocalsInit becomes only relevant for slightly larger stack alloc byte arrays we might also decide to get rid of it depending on the discussion (or consider it slightly too esoteric)

danielmarbach:
@jsquire I think these are great results. I'm sure we could squeeze out more by actually also doing something with the hash algorithm but I felt this might be too complex for now for me to dive into and like you mentioned we would need to be super extra careful not not break anything. With these changes we can get nice gains without really touching the hashing function that much.

sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/EncodingExtensions.cs
danielmarbach: Well the benefit of ifdef would be that instead of using the local implementation the code path would be using the framework built-in versions and hence automatically benefit from any optimizations being done there

sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/SkipLocalsInitAttribute.cs
danielmarbach: I'm totally ok with that

danielmarbach:
SkipLocalsInit is removed. Last oustanding question is whether to ifdef or not. Otherwise I guess this is more or less ready to rumble.

sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/EncodingExtensions.cs
jsquire: The question that I'm considering is whether those potential optimizations are likely to be impactful enough that it warrants the lack of testability that we'd have and, to a lesser extend, warrants future maintainers (including future me) having to reason about what's being called.   If we continue to accept that the big win is that you got us to non-allocating, would saving a couple of ns make an impact?   On the other hand, wrapping the entire class in an `#ifdef` with a comment explaining things isn't a ton of cognitive burden or complexity.

I remain pretty neutral, but I'd slant towards leaving it without the `#ifdef`, potentially with a comment to reconsider if we start to explicitly support `netstandard2.1` as a target.

sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/EncodingExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+    {
+        public static unsafe int GetBytes(this Encoding encoding, ReadOnlySpan<char> src, Span<byte> dest)
```
jsquire, line 11:
```suggestion
        public static unsafe int GetBytes(this Encoding encoding, 
                                           ReadOnlySpan<char> src, 
                                           Span<byte> dest)
```

sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/EncodingExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+                    return encoding.GetBytes(
+                        chars: charPointer,
+                        charCount: src.Length,
+                        bytes: bytePointer,
+                        byteCount: dest.Length);
+                }
+            }
+        }
+
+        public static unsafe int GetByteCount(this Encoding encoding, ReadOnlySpan<char> src)
```
jsquire, line 36:
```suggestion
        public static unsafe int GetByteCount(this Encoding encoding, 
                                              ReadOnlySpan<char> src)
```

sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/EncodingExtensions.cs
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Text;
+
+namespace Azure.Messaging.EventHubs.Primitives
```
jsquire, line 7: If you'd be so kind, doc comments would be appreciated on the class and all methods.

jsquire (commented):
Since we think we're just about ready, I gave it the nitpick formatting pass for consistency with other Event Hubs assets.


By chance, did you run any tests comparing the output of the old and new implementation across a large set of random strings?    If not, I'll do that before we merge for confidence that we didn't accidentally change behavior.

danielmarbach:
@jsquire 

> By chance, did you run any tests comparing the output of the old and new implementation across a large set of random strings? If not, I'll do that before we merge for confidence that we didn't accidentally change behavior.

I did not yet. Normally, what I would do is write an explicit test that copies the old implementation local into the test and then exercise the new implementation together with the old and verify that for the same input the same output occurs on both sides. If you are OK with that approach, I'm happy to push that.

jsquire:
> I did not yet. Normally, what I would do is write an explicit test that copies the old implementation local into the test and then exercise the new implementation together with the old and verify that for the same input the same output occurs on both sides. If you are OK with that approach, I'm happy to push that.

I wasn't even thinking that level of rigor.   I was going to copy the old + new into LinqPad and just run in a giant loop with a few tens of thousands of random inputs.

jsquire:
We ended up with inconsistent results compared to the original implementation.   I traced this back to `BitConverter` reading with the correct endian-ness  for the host where the new approach is using `BinaryPrimitives` and assuming big endian reads.    

I was able to work around this by aliasing the `BinaryPrimitives` reads:

```csharp
private delegate uint ReadUInt32(ReadOnlySpan<byte> bytes);

private static ReadUInt32 ReadUInt32CorrectEndian = BitConverter.IsLittleEndian
    ? BinaryPrimitives.ReadUInt32LittleEndian
    : BinaryPrimitives.ReadUInt32BigEndian;
```

```csharp
// ... [SNIP]

while (size > 12)
{
    a += ReadUInt32CorrectEndian(data.Slice(index));
    b += ReadUInt32CorrectEndian(data.Slice(index + 4));
    c += ReadUInt32CorrectEndian(data.Slice(index + 8));

// ... [SNIP]
```

I did not (yet) have a chance to run another benchmark pass.

danielmarbach:
Could we use a method that can be inlined that uses something like

``` 
BitConverter.IsLittleEndian ? BinaryPrimitives.ReverseEndianness(tempVal) : tempVal
```

?

danielmarbach:
I can rerun the perf test later tonight with a proposal

jsquire:
I generated a test bed when I was doing the compat runs, if that helps.     It's in a simple JSON array format.   
[partition-key-hashes.txt](https://github.com/Azure/azure-sdk-for-net/files/8352724/partition-key-hashes.txt)

danielmarbach:
Hmmm... It does do an endian check.

https://github.com/dotnet/runtime/blob/57bfe474518ab5b7cfe6bf7424a79ce3af9d6657/src/libraries/System.Private.CoreLib/src/System/Buffers/Binary/ReaderBigEndian.cs#L127

I need to have a closer look what's going on there

danielmarbach:
> I'm open to it, sure. I don't think that CLS compliant comes into play unless we're doing something with the public API, no?

Yeah you are right. I haven't had to worry about CLS compliance for a while... Completely forgot that part

danielmarbach:
I think I will simply bring the span based API that netstandard 2.1 has into the lib. Then we are using the BitConverter approach

https://docs.microsoft.com/en-us/dotnet/api/system.bitconverter.toint32?view=net-6.0

danielmarbach:
@jsquire I'm starting to wonder if this is not a bug in the original implementation. `BitConverter.ToUInt32` does an unaligned read. This means the compute hash will depending on the endianness for the same input return different output. Was that intended? Wouldn't you want for the same partition key input always the same output?

jsquire:
> @jsquire I'm starting to wonder if this is not a bug in the original implementation. `BitConverter.ToUInt32` does an unaligned read. This means the compute hash will depending on the endianness for the same input return different output. Was that intended? 

Hrmm....  I hadn't stopped to question the reference version, but yeah, I can see where you're coming from.

> Wouldn't you want for the same partition key input always the same output?

yes, we do.    We also need this to be consistent across languages, which may be a challenge.     Let me verify whether our other SDKs can force that kind of low-level consistency.

jsquire:
The more that I thought about this over the weekend and played around, the more I'm convinced that your observation is spot on - we should assume a consistent endian-ness for the conversion.   I think this means that your original version was right.   I would ask that we assume little endian rather than big, though, since that will be natural to the majority of hosts.

Thoughts?

danielmarbach:
>  I would ask that we assume little endian rather than big, though, since that will be natural to the majority of hosts.

Done. Do we need to worry about backward compatibility?

sdk/eventhub/Azure.Messaging.EventHubs/src/Core/PartitionResolver.cs
danielmarbach: FYI There was recently an interesting conversation that has also been captured in 

https://github.com/dotnet/docs/issues/28823

I did run some tests and if I switch the pattern the numbers look like

image

but by adding `SkipLocalsInit` I'm now getting an impressive boost

image

jsquire:
> Done. Do we need to worry about backward compatibility?

Thankfully, no.  We're running with a slightly different version than the service currently uses, by request, and we've set expectations that the buffered producer will assign partitions differently than the producer.   Once we GA, we'll need to avoid client-side differences, but we're still in beta.

sdk/eventhub/Azure.Messaging.EventHubs/src/Core/PartitionResolver.cs
```diff
@@ -92,15 +94,30 @@ public virtual string AssignRoundRobin(string[] partitions)
         ///   the gateway hashing algorithm as closely as possible and should not be adjusted without careful
         ///   consideration.
         /// </remarks>
-        ///
         private static short GenerateHashCode(string partitionKey)
```
jsquire, line 97:
```suggestion
        ///
        private static short GenerateHashCode(string partitionKey)
```

sdk/eventhub/Azure.Messaging.EventHubs/src/Core/PartitionResolver.cs
```diff
@@ -134,9 +151,9 @@ private static short GenerateHashCode(string partitionKey)
             int index = 0, size = data.Length;
             while (size > 12)
             {
-                a += BitConverter.ToUInt32(data, index);
-                b += BitConverter.ToUInt32(data, index + 4);
-                c += BitConverter.ToUInt32(data, index + 8);
+                a += BinaryPrimitives.ReadUInt32LittleEndian(data.Slice(index) );
```
jsquire, line 154:
```suggestion
                a += BinaryPrimitives.ReadUInt32LittleEndian(data.Slice(index));
```

sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/EncodingExtensions.cs
```diff
@@ -0,0 +1,76 @@
...
+{
+    /// <summary>
+    /// Encoding extensions used to provide a compatibility shim for encoding methods that are missing in .NET Standard 2.0.
+    /// </summary>
```
jsquire, line 11:
```suggestion
    /// <summary>
    ///   Encoding extensions used to provide a compatibility shim for encoding methods that are missing in .NET Standard 2.0.
    /// </summary>
    ///
```

sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/EncodingExtensions.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// Encodes into a span of bytes a set of characters from the specified read-only span.
+        /// </summary>
+        /// <param name="encoding">The encoding to be used.</param>
+        /// <param name="src">The span containing the set of characters to encode.</param>
+        /// <param name="dest">The byte span to hold the encoded bytes.</param>
+        /// <returns>The number of encoded bytes.</returns>
+        /// <remarks>The method was introduced as a compatibility shim for .NET Standard and can be replaced once the
+        /// SDK moves to .NET Standard 2.1 or target frameworks that provides those methods out of the box. During
+        /// the reviews it was decided to not multi target due to the added complexity.
+        /// https://docs.microsoft.com/en-us/dotnet/api/system.text.encoding.getbytes?view=netstandard-2.1#system-text-encoding-getbytes(system-readonlyspan((system-char))-system-span((system-byte))).</remarks>
```
jsquire, line 24:
```suggestion
        /// <summary>
        ///   Encodes into a span of bytes a set of characters from the specified read-only span.
        /// </summary>
        ///
        /// <param name="encoding">The encoding to be used.</param>
        /// <param name="src">The span containing the set of characters to encode.</param>
        /// <param name="dest">The byte span to hold the encoded bytes.</param>
        ///
        /// <returns>The count of encoded bytes.</returns>
        ///
        /// <remarks>
        ///   The method was introduced as a compatibility shim for .NET Standard and can be replaced should the
        ///   SDK change targets to .NET Standard 2.1 or target frameworks that provides those methods out of the box.
        ///   During  reviews it was decided to not multi-target due to the added complexity.
        ///  </remarks>
        ///
        ///   <seealso="https://docs.microsoft.com/dotnet/api/system.text.encoding.getbytes?view=netstandard-2.1#system-text-encoding-getbytes(system-readonlyspan((system-char))-system-span((system-byte)))" />
        ///
```

sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/EncodingExtensions.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// <summary>
+        /// Calculates the number of bytes produced by encoding the characters in the specified character span.
+        /// </summary>
+        /// <param name="encoding">The encoding to be used.</param>
+        /// <param name="src">The span of characters to encode.</param>
+        /// <returns>The number of bytes produced by encoding the specified character span.</returns>
+        /// <remarks>The method was introduced as a compatibility shim for .NET Standard and can be replaced once the
+        /// SDK moves to .NET Standard 2.1 or target frameworks that provides those methods out of the box. During
+        /// the reviews it was decided to not multi target due to the added complexity.
+        /// https://docs.microsoft.com/en-us/dotnet/api/system.text.encoding.getbytecount?view=netstandard-2.1#system-text-encoding-getbytecount(system-readonlyspan((system-char)))</remarks>
```
jsquire, line 61:
```suggestion
        /// <summary>
        ///   Calculates the number of bytes produced by encoding the characters in the specified character span.
        /// </summary>
        ///
        /// <param name="encoding">The encoding to be used.</param>
        /// <param name="src">The span of characters to encode.</param>
        ///
        /// <returns>The count of bytes produced by encoding the specified character span.</returns>
        ///
       /// <remarks>
        ///   The method was introduced as a compatibility shim for .NET Standard and can be replaced should the
        ///   SDK change targets to .NET Standard 2.1 or target frameworks that provides those methods out of the box.
        ///   During  reviews it was decided to not multi-target due to the added complexity.
        ///  </remarks>
        ///
        ///   <seealso="https://docs.microsoft.com/dotnet/api/system.text.encoding.getbytes?view=netstandard-2.1#system-text-encoding-getbytes(system-readonlyspan((system-char))-system-span((system-byte)))" />
        ///
```

jsquire (approved):
Thank you, @danielmarbach, as always, for your taking this on and your flexibility talking through the ambiguity.  Pedantic formatting guy made one last pass, but I think the important parts are good to go.

danielmarbach:
@jsquire let me know what you think about the recent discussion point I brought up regarding following more the best practices of always stack allocating the max size and how in that scenario having `SkipLocalsInit` would actually improve things quite a bit.

On another note: Would you be able to run your fancy script against the latest changes?

sdk/eventhub/Azure.Messaging.EventHubs/src/Core/PartitionResolver.cs
danielmarbach: Or in other words I would like to bring back https://github.com/Azure/azure-sdk-for-net/pull/27598/commits/14912a3cfda8253debf92db9944c62a0d25ffcc2

danielmarbach: Brought the SkipLocalsInit back and aligned the stack alloc pattern to always go for the max size as suggested by the linked conversation

sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/SkipLocalsInitAttribute.cs
```diff
@@ -0,0 +1,23 @@
...
+// Licensed under the MIT License.
+
+namespace System.Runtime.CompilerServices
+{
+    /// <summary>
+    /// Used to indicate to the compiler that the <c>.locals init</c> flag should not be set in method headers.
+    /// </summary>
+    /// <remarks>Internal copy of the .NET 5 attribute.</remarks>
```
jsquire, line 9:
```suggestion
    /// <summary>
    ///   Used to indicate to the compiler that the <c>.locals init</c> flag should not be set in method headers.
    /// </summary>
    ///
    /// <remarks>Internal copy of the .NET 5 attribute.</remarks>
    ///
```

sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/SkipLocalsInitAttribute.cs
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace System.Runtime.CompilerServices
```
jsquire, line 4: Do we need to wrap this in an `#ifdef` or is it just a marker interface that would have the same effect as the actual system implementation?

jsquire:
> let me know what you think about the recent discussion point I brought up regarding following more the best practices of always stack allocating the max size and how in that scenario having `SkipLocalsInit` would actually improve things quite a bit.

I'm not opposed to bringing it back; I just question how much of an impact it will have with our `netstandard2.0` target.  If it's being evaluated by the jitter and would impact newer runtimes for applications using the SDK, I think it's a definite win.

jsquire:
> On another note: Would you be able to run your fancy script against the latest changes?

Will do.   I'll do another pass today.

sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/SkipLocalsInitAttribute.cs
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
```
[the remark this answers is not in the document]
danielmarbach: Done

danielmarbach:
> > let me know what you think about the recent discussion point I brought up regarding following more the best practices of always stack allocating the max size and how in that scenario having `SkipLocalsInit` would actually improve things quite a bit.
> 
> I'm not opposed to bringing it back; I just question how much of an impact it will have with our `netstandard2.0` target. If it's being evaluated by the jitter and would impact newer runtimes for applications using the SDK, I think it's a definite win.

It instructs the compiler to not add the locals init. So, the important part is to compile the code that uses it with a roslyn compiler version that supports that. I guess that is given here right?

jsquire:
> It instructs the compiler to not add the locals init. So, the important part is to compile the code that uses it with a roslyn compiler version that supports that. I guess that is given here right?

I'll be interesting to see if the net6 SDK compiler does the right thing for a netstandard2.0 target.  We'll always be current with the SDK/compiler versions, it's the targets that will lag behind for compatibility.

sdk/eventhub/Azure.Messaging.EventHubs/src/Core/PartitionResolver.cs
```diff
@@ -93,14 +95,31 @@ public virtual string AssignRoundRobin(string[] partitions)
...
+            var encoding = Encoding.UTF8;
+
+            var partitionKeyByteLength = encoding.GetMaxByteCount(partitionKey.Length);
+            var hashBuffer = partitionKeyByteLength <= MaxStackLimit ?
+                stackalloc byte[MaxStackLimit] :
+                sharedBuffer = ArrayPool<byte>.Shared.Rent(partitionKeyByteLength);
+
+            var written = encoding.GetBytes(partitionKeySpan, hashBuffer);
+            ComputeHash(hashBuffer.Slice(0, written), seed1: 0, seed2: 0, out uint hash1, out uint hash2);
+            if (sharedBuffer != null)
```
jsquire, line 119: Should we move this to a `finally` block?

jsquire:
> > On another note: Would you be able to run your fancy script against the latest changes?
> 
> Will do. I'll do another pass today.

Test pass is solid.   260,000+ random key variations with different patterns and results are consistent.

sdk/eventhub/Azure.Messaging.EventHubs/src/Core/PartitionResolver.cs
danielmarbach: Since this is perf sensitive I would lean towards leaving as is but can understand if you play the safe card 💳😊

jsquire: Talk me off the ledge.  😄    We've got user-provided input that we're assuming we can treat as UTF-8 bytes.   Failure causes a rented buffer to be orphaned.    However, I can't think of a reasonable way that you could create a normal .NET string that would cause an encoding failure.... you?

jsquire: You and me both.   Let's leave things as-is for now.   We can always guard if we have feedback related to errors.

jsquire:
@danielmarbach:  Looks like we're all wrapped up and ready.   Please let me know if you'd like me to merge or if you're still making tweaks.

danielmarbach:
Go ahead

PR #27599, opened 2022-03-16
Update to BinaryData deserializer to support Any and AnyObject

m-nash:

In this PR https://github.com/Azure/azure-sdk-for-net/pull/27382 we were focused on supporting `BinaryData` for `AnyObject`.  After looking at the resulting regeneration in this PR https://github.com/Azure/azure-sdk-for-net/pull/27572 we realized we need to address two other cases.

1. We want to support both `Any` and `AnyObject` the same way so we will change the public API from `ToDictionaryFromJson` to `ToObjectFromJson` with no T parameter to indicate we want a typed generic object returned.  Customers would original wanted the dictionary can call the new extension method and cast to `Dictionary<string, object>`. 
2. We need a way to handle cases where the replacement results in either a `List<BinaryData>` or a `Dictionary<string, BinaryData>`.  This portion will go into autorest since that is where the decision is made, but the affect will be in this repository after the generator is updated.


sdk/core/Azure.Core/src/Serialization/AzureCoreExtensions.cs
```diff
@@ -44,17 +44,17 @@ public static class AzureCoreExtensions
...
-        /// Each value in the key value pair will be strongly typed as an int, long, string, Guid, double or bool.
-        /// Each value can also be another Dictionary of string object representing an inner object or
-        /// a List of objects representing an array.
+        /// Converts the <see cref="BinaryData"/> to an object based on the json value it contains.
+        /// If it is a primitive type such as string, int, double, bool, or null literal you will get that back.
+        /// Otherwise you will get back either an array of object or Dictionary of string to object.
+        /// Each value in the key value pair or list will be also be converted into a primitive or another complex type recursively.
         /// </summary>
         /// <param name="data">The <see cref="BinaryData"/> instance to convert.</param>
         /// <returns>The data converted to the Dictionary of string to object.</returns>
```
christothes, line 53:
```suggestion
        /// Converts the json value represented by <see cref="BinaryData"/> to an object of a specific type.
        /// If it is a primitive type such as string, int, double, bool, or null literal you will get that back.
        /// Otherwise you will get back either an array of object or Dictionary of string to object.
        /// Each value in the key value pair or list will be also be converted into a primitive or another complex type recursively.
        /// </summary>
        /// <param name="data">The <see cref="BinaryData"/> instance to convert.</param>
       /// <returns> The object value of the json value.
        /// If the object contains a primitive type such as string, int, double, bool, or null literal, it returns that type.
        /// Otherwise, it returns either an <see cref="object[]"/> or <see cref="Dictionary{TKey, TValue}"/> where TKey is <see cref="string"/> and TValue is <see cref="object"/>.
        /// Each value in the key value pair or list will be also be converted into a primitive or another complex type.
        /// </returns>
```

sdk/core/Azure.Core/src/Serialization/AzureCoreExtensions.cs
```diff
@@ -44,17 +44,21 @@ public static class AzureCoreExtensions
...
 
         /// <summary>
-        /// Converts the <see cref="BinaryData"/> to a Dictionary of string to object.
-        /// Each value in the key value pair will be strongly typed as an int, long, string, Guid, double or bool.
-        /// Each value can also be another Dictionary of string object representing an inner object or
-        /// a List of objects representing an array.
+        /// Converts the json value represented by <see cref="BinaryData"/> to an object of a specific type.
+        /// If it is a primitive type such as string, int, double, bool, or null literal you will get that back.
+        /// Otherwise you will get back either an array of object or Dictionary of string to object.
+        /// Each value in the key value pair or list will be also be converted into a primitive or another complex type recursively.
```
christothes, line 50: sorry, meant to dedupe this from the <returns> block

sdk/core/Azure.Core/src/Serialization/AzureCoreExtensions.cs
```diff
@@ -44,17 +44,21 @@ public static class AzureCoreExtensions
...
         /// </summary>
         /// <param name="data">The <see cref="BinaryData"/> instance to convert.</param>
-        /// <returns>The data converted to the Dictionary of string to object.</returns>
-        public static IDictionary<string, object?> ToDictionaryFromJson(this BinaryData data)
+       /// <returns> The object value of the json value.
+        /// If the object contains a primitive type such as string, int, double, bool, or null literal, it returns that type.
+        /// Otherwise, it returns either an <see cref="object[]"/> or <see cref="Dictionary{TKey, TValue}"/> where TKey is <see cref="string"/> and TValue is <see cref="object"/>.
+        /// Each value in the key value pair or list will be also be converted into a primitive or another complex type.
```
christothes, line 56:
```suggestion
        /// Each value in the key value pair or list will also be converted into a primitive or another complex type recursively.
```

christothes (approved):

PR #19105, opened 2021-03-01
[Core] Improve OperationHelpers used by our LROs

kinelski:
Fixes https://github.com/azure/azure-sdk-for-net/issues/18958.

Creating new `OperationInternal` class to replace the current `OperationHelpers` class. It provides:
- Implementation for `Value`, `HasValue`, `HasCompleted`, `RawResponse`.
- Implementation for `UpdateStatus` and `UpdateStatusAsync`, taking diagnostic scope creation into consideration.
- Implementation for `WaitForCompletionAsync` (both overloads), taking the retry-after header into consideration.

Updated one Form Recognizer operation to illustrate how the new `OperationInternal` should be used.

PR #27639, opened 2022-03-18
Synchronize OperationInternalBase calls

AlexanderSher:
This is a draft of the change to resolve several issues:
- When two `WaitForCompletion` calls are made concurrently, `RawResponse` property may contain response that is not from the final state. For details, see:  https://github.com/Azure/azure-sdk-for-net/pull/19105#discussion_r621554812
- Because of the problem above, right now we have a workaround that every call for `WaitForCompletion` makes at least one call even if operation is completed, including cases when operation was created as completed
- `OperationInternal<T>.Value` may contain value that can't be recreated from `RawResponse`
- `OperationInternal<T>.HasValue` may be `true` while `RawResponse` contains "pending" response

### DIFF
--- a/sdk/confidentialledger/Azure.Security.ConfidentialLedger/src/PostLedgerEntryOperation.cs
+++ b/sdk/confidentialledger/Azure.Security.ConfidentialLedger/src/PostLedgerEntryOperation.cs
@@ -61,15 +61,13 @@ namespace Azure.Security.ConfidentialLedger
                     .ConfigureAwait(false)
                 : _client.GetTransactionStatus(Id, new RequestContext { CancellationToken = cancellationToken, ErrorOptions = ErrorOptions.NoThrow });
 
-            _operationInternal.RawResponse = statusResponse;
-
             if (statusResponse.Status != (int)HttpStatusCode.OK)
             {
                 var error = new ResponseError(null, exceptionMessage);
                 var ex = async
                     ? await _client.ClientDiagnostics.CreateRequestFailedExceptionAsync(statusResponse, error).ConfigureAwait(false)
                     : _client.ClientDiagnostics.CreateRequestFailedException(statusResponse, error);
-                return OperationState.Failure(GetRawResponse(), new RequestFailedException(exceptionMessage, ex));
+                return OperationState.Failure(statusResponse, new RequestFailedException(exceptionMessage, ex));
             }
 
             string status = JsonDocument.Parse(statusResponse.Content)
--- a/sdk/core/Azure.Core/src/Shared/AsyncLockWithValue.cs
+++ b/sdk/core/Azure.Core/src/Shared/AsyncLockWithValue.cs
@@ -1,0 +1,227 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#nullable enable
+
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Azure.Core.Pipeline;
+
+namespace Azure.Core
+{
+    /// <summary>
+    /// Primitive that combines async lock and value cache
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    internal sealed class AsyncLockWithValue<T>
+    {
+        private readonly object _syncObj = new object();
+        private Queue<TaskCompletionSource<Lock>>? _waiters;
+        private bool _isLocked;
+        private bool _hasValue;
+        private T? _value;
+
+        public bool HasValue
+        {
+            get
+            {
+                lock (_syncObj)
+                {
+                    return _hasValue;
+                }
+            }
+        }
+
+        public AsyncLockWithValue() { }
+
+        public AsyncLockWithValue(T value)
+        {
+            _hasValue = true;
+            _value = value;
+        }
+
+        public bool TryGetValue(
+            out T? value)
+        {
+            lock (_syncObj)
+            {
+                if (_hasValue)
+                {
+                    value = _value;
+                    return true;
+                }
+            }
+
+            value = default;
+            return false;
+        }
+
+        /// <summary>
+        /// Method that either returns cached value or acquire a lock.
+        /// If one caller has acquired a lock, other callers will be waiting for the lock to be released.
+        /// If value is set, lock is released and all waiters get that value.
+        /// If value isn't set, the next waiter in the queue will get the lock.
+        /// </summary>
+        /// <param name="async"></param>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        public async ValueTask<Lock> GetLockOrValueAsync(bool async, CancellationToken cancellationToken = default)
+        {
+            TaskCompletionSource<Lock> valueTcs;
+            lock (_syncObj)
+            {
+                // If there is a value, just return it
+                if (_hasValue)
+                {
+                    return new Lock(_value!);
+                }
+
+                // If lock isn't acquire yet, acquire it and return to the caller
+                if (!_isLocked)
+                {
+                    _isLocked = true;
+                    return new Lock(this);
+                }
+
+                // Check cancellationToken before instantiating waiter
+                cancellationToken.ThrowIfCancellationRequested();
+
+                // If lock is already taken, create a waiter and wait either until value is set or lock can be acquired by this waiter
+                _waiters ??= new Queue<TaskCompletionSource<Lock>>();
+                // if async == false, valueTcs will be waited only in this thread and only synchronously, so RunContinuationsAsynchronously isn't needed.
+                valueTcs = new TaskCompletionSource<Lock>(async ? TaskCreationOptions.RunContinuationsAsynchronously : TaskCreationOptions.None);
+                _waiters.Enqueue(valueTcs);
+            }
+
+            try
+            {
+                if (async)
+                {
+                    return await valueTcs.Task.AwaitWithCancellation(cancellationToken);
+                }
+
+#pragma warning disable AZC0104 // Use EnsureCompleted() directly on asynchronous method return value.
+#pragma warning disable AZC0111 // DO NOT use EnsureCompleted in possibly asynchronous scope.
+                valueTcs.Task.Wait(cancellationToken);
+                return valueTcs.Task.EnsureCompleted();
+#pragma warning restore AZC0111 // DO NOT use EnsureCompleted in possibly asynchronous scope.
+#pragma warning restore AZC0104 // Use EnsureCompleted() directly on asynchronous method return value.
+            }
+            catch (OperationCanceledException)
+            {
+                // Throw OperationCanceledException only if another thread hasn't set a value to this waiter
+                // by calling either Reset or SetValue
+                if (valueTcs.TrySetCanceled(cancellationToken))
+                {
+                    throw;
+                }
+
+                return valueTcs.Task.Result;
+            }
+        }
+
+        /// <summary>
+        /// Set value to the cache and to all the waiters
+        /// </summary>
+        /// <param name="value"></param>
+        private void SetValue(T value)
+        {
+            Queue<TaskCompletionSource<Lock>> waiters;
+            lock (_syncObj)
+            {
+                _value = value;
+                _hasValue = true;
+                _isLocked = false;
+                if (_waiters == default)
+                {
+                    return;
+                }
+
+                waiters = _waiters;
+                _waiters = default;
+            }
+
+            while (waiters.Count > 0)
+            {
+                waiters.Dequeue().TrySetResult(new Lock(value));
+            }
+        }
+
+        /// <summary>
+        /// Release the lock and allow next waiter acquire it
+        /// </summary>
+        private void Reset()
+        {
+            TaskCompletionSource<Lock>? nextWaiter = UnlockOrGetNextWaiter();
+            while (nextWaiter != default && !nextWaiter.TrySetResult(new Lock(this)))
+            {
+                nextWaiter = UnlockOrGetNextWaiter();
+            }
+        }
+
+        private TaskCompletionSource<Lock>? UnlockOrGetNextWaiter()
+        {
+            lock (_syncObj)
+            {
+                if (!_isLocked)
+                {
+                    return default;
+                }
+
+                if (_waiters == default)
+                {
+                    _isLocked = false;
+                    return default;
+                }
+
+                while (_waiters.Count > 0)
+                {
+                    var nextWaiter = _waiters.Dequeue();
+                    if (!nextWaiter.Task.IsCompleted)
+                    {
+                        // Return the waiter only if it wasn't canceled already
+                        return nextWaiter;
+                    }
+                }
+
+                _isLocked = false;
+                return default;
+            }
+        }
+
+        public readonly struct Lock : IDisposable
+        {
+            private readonly AsyncLockWithValue<T>? _owner;
+            public bool HasValue => _owner == default;
+            public T? Value { get; }
+
+            public Lock(T value)
+            {
+                _owner = default;
+                Value = value;
+            }
+
+            public Lock(AsyncLockWithValue<T> owner)
+            {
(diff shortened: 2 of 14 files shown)


sdk/core/Azure.Core/src/Shared/AsyncLockWithValue.cs
```diff
@@ -0,0 +1,227 @@
...
+using System.Threading.Tasks;
+using Azure.Core.Pipeline;
+
+namespace Azure.Core
+{
+    /// <summary>
+    /// Primitive that combines async lock and value cache
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    internal sealed class AsyncLockWithValue<T>
```
AlexanderSher, line 18: This type is a copy of type from [Azure.Identity](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/identity/Azure.Identity/src/AsyncLockWithValue.cs) with added `HasValue` property and `TryGetValue` method.

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
```diff
@@ -18,14 +18,30 @@ internal abstract class OperationInternalBase
...
+
+        protected OperationInternalBase(ClientDiagnostics clientDiagnostics, Response rawResponse, string operationTypeName, IEnumerable<KeyValuePair<string, string>>? scopeAttributes = null, DelayStrategy? fallbackStrategy = null, OperationState? finalState = null)
         {
             _diagnostics = clientDiagnostics;
             _updateStatusScopeName = $"{operationTypeName}.UpdateStatus";
-            _scopeAttributes = scopeAttributes?.ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
             RawResponse = rawResponse;
-            _fallbackStrategy = fallbackStrategy;
+
+            if (finalState != null)
```
AlexanderSher, line 35: To avoid overriding of one final state with another, final state is assigned in constructor instead of separate method.

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
```diff
@@ -176,14 +190,38 @@ public virtual Response WaitForCompletionResponse(CancellationToken cancellation
...
 
+        public bool TrySetPendingResponse(Response response, CancellationToken cancellationToken)
+            => TrySetPendingResponseAsync(false, response, cancellationToken).EnsureCompleted();
+
+        public async ValueTask<bool> TrySetPendingResponseAsync(Response response, CancellationToken cancellationToken)
+            => await TrySetPendingResponseAsync(true, response, cancellationToken).ConfigureAwait(false);
+
+        private async ValueTask<bool> TrySetPendingResponseAsync(bool async, Response response, CancellationToken cancellationToken)
+        {
+            using var asyncLock = await _operationStateLock.GetLockOrValueAsync(async, cancellationToken).ConfigureAwait(false);
```
AlexanderSher, line 207: This method allows update of the RawResponse up until operation is completed

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
```diff
@@ -176,14 +190,38 @@ public virtual Response WaitForCompletionResponse(CancellationToken cancellation
...
+                return false;
+            }
+
+            RawResponse = response;
+            return true;
+        }
+
         private async ValueTask<Response> UpdateStatusAsync(bool async, CancellationToken cancellationToken)
         {
+            using var asyncLock = await _operationStateLock.GetLockOrValueAsync(async, cancellationToken).ConfigureAwait(false);
```
AlexanderSher, line 219: Here is where synchronization of calls happens.

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
```diff
@@ -198,43 +236,37 @@ private async ValueTask<Response> UpdateStatusAsync(bool async, CancellationToke
...
+                }
 
-            if (!hasCompleted)
-            {
-                return response;
-            }
+                if (!state.HasSucceeded && state.OperationFailedException == null)
+                {
+                    var exception = async
+                        ? await _diagnostics.CreateRequestFailedExceptionAsync(state.RawResponse).ConfigureAwait(false)
```
AlexanderSher, line 249: @annelo-msft when custom error parsing introduced, we will have to update this place as well

sdk/core/Azure.Core/tests/AsyncLockWithValueTests.cs
```diff
@@ -0,0 +1,251 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using NUnit.Framework;
+
+namespace Azure.Core.Tests
+{
+    public class AsyncLockWithValueTests
```
AlexanderSher, line 12: Tests also copied from `Azure.Identity`

sdk/confidentialledger/Azure.Security.ConfidentialLedger/src/PostLedgerEntryOperation.cs
```diff
@@ -61,15 +61,13 @@ async ValueTask<OperationState> IOperation.UpdateStateAsync(bool async, Cancella
                     .ConfigureAwait(false)
                 : _client.GetTransactionStatus(Id, new RequestContext { CancellationToken = cancellationToken, ErrorOptions = ErrorOptions.NoThrow });
 
-            _operationInternal.RawResponse = statusResponse;
```
AlexanderSher, line 64: This is not needed, `_operationInternal` updates its `RawResponse`

AlexanderSher:
I'll add more tests for this if design is approved.

sdk/core/Azure.Core/src/Shared/AsyncLockWithValue.cs
```diff
@@ -0,0 +1,227 @@
...
+                {
+                    _owner.SetValue(value);
+                }
+                else
+                {
+                    throw new InvalidOperationException("Value for the lock is set already");
+                }
+            }
+
+            public void Dispose() => _owner?.Reset();
```
heaths, line 224: Might be worth a comment here clarifying what this is intended to do. I believe the goal here is just to release any queued lock - not necessarily this one being disposed - and cycle through until we get another lock that hasn't completed, effectively like a semaphore, right?

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
```diff
@@ -105,7 +119,7 @@ protected OperationInternalBase(ClientDiagnostics clientDiagnostics, Response ra
         /// <param name="cancellationToken">A <see cref="CancellationToken"/> controlling the request lifetime.</param>
         /// <returns>The last HTTP response received from the server, including the final result of the long-running operation.</returns>
         /// <exception cref="RequestFailedException">Thrown if there's been any issues during the connection, or if the operation has completed with failures.</exception>
-        public virtual async ValueTask<Response> WaitForCompletionResponseAsync(CancellationToken cancellationToken)
+        public async ValueTask<Response> WaitForCompletionResponseAsync(CancellationToken cancellationToken)
```
heaths, line 122: Should these remain virtual for mocking? This would also be a breaking binary change: can't `callvirt` a non-virtual method.

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
heaths: Could you add that as a comment? Would help with future maintenance.

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
heaths: Same here. Would help with future maintenance.

sdk/keyvault/Azure.Security.KeyVault.Certificates/src/DeleteCertificateOperation.cs
```diff
@@ -24,18 +24,15 @@ internal DeleteCertificateOperation(KeyVaultPipeline pipeline, Response<DeletedC
         {
             _pipeline = pipeline;
             _value = response.Value ?? throw new InvalidOperationException("The response does not contain a value.");
+            // The recoveryId is only returned if soft delete is enabled.
```
heaths, line 27: Nit: can you add a blank line between 26 and 27. IIRC, this may cause some code style errors.

heaths (approved):
General and KV changes LGTM, though I had a few suggestions and question.

sdk/core/Azure.Core/src/Shared/OperationInternalOfT.cs
```diff
@@ -78,17 +77,22 @@ internal class OperationInternal<T> : OperationInternalBase
...
         public OperationInternal(
             ClientDiagnostics clientDiagnostics,
             IOperation<T> operation,
             Response rawResponse,
             string? operationTypeName = null,
             IEnumerable<KeyValuePair<string, string>>? scopeAttributes = null,
-            DelayStrategy? fallbackStrategy = null)
-            : base(clientDiagnostics, rawResponse, operationTypeName ?? operation.GetType().Name, scopeAttributes, fallbackStrategy)
+            DelayStrategy? fallbackStrategy = null,
+            OperationState<T>? finalState = null)
```
m-nash, line 88: Given we are taking this in on the constructor now I think we should move this class into its own file along with the non T version

https://github.com/Azure/azure-sdk-for-net/pull/27639/files#diff-f4e6fe06a0be62f7078fe921b22173ed7ad094e129c90206626734d527eb31f8R286

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
AlexanderSher: This type is internal

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
```diff
@@ -198,43 +236,37 @@ private async ValueTask<Response> UpdateStatusAsync(bool async, CancellationToke
...
-                throw OperationFailedException;
+                scope.Failed(e);
+                throw;
             }
-
-            return response;
         }
 
-        protected abstract ValueTask<Response> UpdateStateAsync(bool async, CancellationToken cancellationToken);
+        protected abstract ValueTask<OperationState> UpdateStateAsync(bool async, CancellationToken cancellationToken);
```
m-nash, line 270: It feels weird that we have a `State` class representing the state of the current operation yet that state is never stored in the instance of the class it simply creates one, then assigns some member variables.  Should we maybe store this as a member variable and have the HasCompleted / HasValue / Value all lambda into this object to get their answers?

sdk/core/Azure.Core/src/Shared/OperationInternal.cs
```diff
@@ -68,32 +68,22 @@ internal class OperationInternal : OperationInternalBase
...
         public OperationInternal(
             ClientDiagnostics clientDiagnostics,
             IOperation operation,
             Response rawResponse,
             string? operationTypeName = null,
             IEnumerable<KeyValuePair<string, string>>? scopeAttributes = null,
-            DelayStrategy? fallbackStrategy = null)
-            :base(clientDiagnostics, rawResponse, operationTypeName ?? operation.GetType().Name, scopeAttributes, fallbackStrategy)
+            DelayStrategy? fallbackStrategy = null,
+            OperationState? finalState = null)
```
m-nash, line 79: Do we worry about the conflict between the response in this final state and the response as the parameter?  Should we take one or the other?

sdk/core/Azure.Core/src/Shared/OperationInternalOfT.cs
```diff
@@ -223,27 +223,14 @@ public Response<T> WaitForCompletion(TimeSpan pollingInterval, CancellationToken
...
-            {
-                Value = state.Value!;
-            }
-            ApplyStateAsync(false, state.RawResponse, state.HasCompleted, state.HasSucceeded, state.OperationFailedException, throwIfFailed: false).EnsureCompleted();
-        }
-
-        protected override async ValueTask<Response> UpdateStateAsync(bool async, CancellationToken cancellationToken)
+        protected override async ValueTask<OperationState> UpdateStateAsync(bool async, CancellationToken cancellationToken)
         {
             OperationState<T> state = await _operation.UpdateStateAsync(async, cancellationToken).ConfigureAwait(false);
```
m-nash, line 228: This method on IOperation.UpdateStateAsync feels like it should be named GetStateAsync since it shouldn't be updating anything only this instance method should?

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
```diff
@@ -198,43 +236,37 @@ private async ValueTask<Response> UpdateStatusAsync(bool async, CancellationToke
...
-            }
+                if (!state.HasSucceeded && state.OperationFailedException == null)
+                {
+                    var exception = async
+                        ? await _diagnostics.CreateRequestFailedExceptionAsync(state.RawResponse).ConfigureAwait(false)
+                        : _diagnostics.CreateRequestFailedException(state.RawResponse);
+                    state = OperationState.Failure(state.RawResponse, exception);
+                }
 
-            HasCompleted = true;
```
m-nash, line 219: I would like to understand more why this race condition has to be fixed this way.  I don't like that we are forcing an extra call even when the operation is completed (unless there is some nuance in your statement there).

m-nash (changes requested):

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
AlexanderSher: It is stored: https://github.com/Azure/azure-sdk-for-net/pull/27639/files#diff-f9f7a9d046f21930f94a8ca1d25df78506aabb692ce00a3504bdce32498b4ef3R254

sdk/core/Azure.Core/src/Shared/OperationInternal.cs
AlexanderSher: @heaths , is it possible that these responses are different in keyvault?

heaths: For the operations you changed, they are all pseudo-LROs anyway. The service backed endpoints aren't actually LROs, but we poll the final resource URI until it's not 404. Given the response and the final state should be agreed in that case, I don't think it should be possible.

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
m-nash: I read the OP comment of `To mitigate this problem` we are adding an extra call, but it was referring to the previous bullet and how it was previously fixed.  My understanding is now that the extra call is being removed.

I still would like to understand if there is a simpler way to resolve the race condition and allow the occasional thread to need an extra call

AlexanderSher: Depends on what do you mean by "simpler". If we make an extra call, then we need to throw away its response if operation is in a final state. Otherwise, `Response.FromValue(Value, rawResponse)` may use `Value` from one request and `rawResponse` from another.
There are two options how to achieve it: using locks/semaphores or using state object that can be read/replaced in atomic operation. Since we have `OperationInternal` and `OperationInternal<T>`, there will be at least two objects (one derived from another). I can implement this version in a separate branch so we can compare.

sdk/core/Azure.Core/src/Shared/AsyncLockWithValue.cs
christothes: Assuming we make this change, we should change Identity's reference to this shared source version also.

AlexanderSher: We can do that.

AlexanderSher:
Closing in favor of https://github.com/Azure/azure-sdk-for-net/pull/28375

Issue #27614, opened 2022-03-17
[BUG] HLC + DPG + Exceptions? 


tg-msft:
I’m having a problem trying to use DPG via HLC around exceptions:
- Storage HLCs use its own [StorageClientDiagnostics ExtractFailureContent](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/storage/Azure.Storage.Common/src/Shared/StorageClientDiagnostics.cs#L26) when [it throws](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/storage/Azure.Storage.Blobs/src/Generated/BlobRestClient.cs#L150). 
- Storage [using DPG](https://github.com/Azure/autorest.csharp/blob/feature/v3/src/assets/Generator.Shared/HttpPipelineExtensions.cs#L53) throws [via RequestFailedException’s .ctor](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/RequestFailedException.cs#L79) which just calls [the naïve default](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/Shared/ClientDiagnostics.cs#L64) for extracting that can’t parse Storage’s XML response body.
 
If I’m surfacing this to Storage customers, it won’t populate the RequestFailedException.ErrorCode, it won’t use the additionalInfo for a great exception message, etc.  Not having ErrorCode in particular is a deal breaker.

I think this also breaks the DPG "don't throw" story because I'll get different exceptions if the client throws vs. if I do myself.

tg-msft:
Closing as this no longer repros:

```C#
using Azure;
using Azure.Storage;
using Azure.Storage.Blobs;
using Azure.Storage.Blobs.Specialized;

BlobClient blob = new(new Uri("..."), new StorageSharedKeyCredential("...", "..."));
BlobLeaseClient lease = blob.GetBlobLeaseClient();

try
{
    await lease.AcquireAsync(TimeSpan.FromHours(3));
}
catch (RequestFailedException ex)
{
    Console.WriteLine($"Convenience: {ex}");
}

try
{
    await lease.AcquireAsync(TimeSpan.FromHours(3), conditions: null, new RequestContext());
}
catch (RequestFailedException ex)
{
    Console.WriteLine($"Protocol: {ex}");
}

try
{
    RequestContext ctx = new();
    ctx.AddClassifier(400, isError: false);
[...]

Issue #27490, opened 2022-03-10
Support custom error data extraction in RequestFailedException


AlexanderSher:
`ClientDiagnostics` type has a virtual method that can be overridden: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/Shared/ClientDiagnostics.cs#L52 
If `ExtractFailureContent` is overridden in derived type, there is no way to use it when we call `RequestFailedException` constructor directly. We need some way to inject custom error data extraction logic into `RequestFailedException`

annelo-msft:
Related to https://github.com/Azure/azure-sdk-for-net/issues/24031

Issue #24031, opened 2021-09-15
[Core] Move exception creation functionality out of ClientDiagnostics


annelo-msft:
This needs to be addressed as a part of https://github.com/Azure/azure-sdk-for-net/issues/23372

- [ ] Add the functionality to new type in Azure.Core
- [ ] Update generator shared source to use new Response-based APIs
- [ ] Update libraries using ClientDiagnostics to move to the new type for exception creation
- [ ] Remove functionality from ClientDiagnostics

annelo-msft:
See nice pre-work done by @pakrym: https://github.com/Azure/azure-sdk-for-net/compare/main...pakrym%3Apakrym%2FException-creation-in-core?expand=1

annelo-msft:
Investigation into services that have subclassed ResponseClassifier:
- https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/identity/Azure.Identity/src/CredentialPipeline.cs
- https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/identity/Azure.Identity/src/ManagedIdentitySource.cs
- https://github.com/Azure/azure-sdk-for- net/blob/main/sdk/storage/Azure.Storage.Common/src/Shared/StorageResponseClassifier.cs

annelo-msft:
Services that have [subclassed ClientDiagnostics](https://grep.app/search?q=%3A%20ClientDiagnostics&case=true&filter[repo][0]=Azure/azure-sdk-for-net):
- sdk/servicebus/Azure.Messaging.ServiceBus/src/Administration/ServiceBusClientDiagnostics.cs
- sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClientDiagnostics.cs
- sdk/communication/Azure.Communication.PhoneNumbers/src/CommunicationClientDiagnostics.cs
- sdk/tables/Azure.Data.Tables/src/TablesClientDiagnostics.cs
- sdk/storage/Azure.Storage.Common/src/Shared/StorageClientDiagnostics.cs

annelo-msft:
Related to: https://github.com/Azure/azure-sdk-for-net/issues/25583

annelo-msft:
Work remains to move all libraries to the new implementation and remove the old functionality from ClientDiagnostics - that would be great to do in MQ.

PR #27643, opened 2022-03-18
[Prototype] Move service-library customization into RequestFailedException constructor

annelo-msft:
Prototype addressing https://github.com/Azure/azure-sdk-for-net/issues/27614

Also:
- https://github.com/Azure/azure-sdk-for-net/issues/24031
- https://github.com/Azure/azure-sdk-for-net/issues/27490


sdk/core/Azure.Core/src/Shared/ClientDiagnostics.cs
```diff
@@ -18,7 +18,7 @@ namespace Azure.Core.Pipeline
 {
     internal class ClientDiagnostics : DiagnosticScopeFactory
     {
-        private const string DefaultMessage = "Service request failed.";
+        //private const string DefaultMessage = "Service request failed.";
 
         private readonly HttpMessageSanitizer _sanitizer;
```
annelo-msft, line 23: We'll want to remove this if it's never used.

Issue #25626, opened 2021-12-01
Expose `RequestOptions` for HLC methods causing noisy logs/tracing


tg-msft:
The upcoming `Azure.RequestOptions` gives us the chance to expose [`ErrorOptions`](https://docs.microsoft.com/en-us/dotnet/api/azure.erroroptions?view=azure-dotnet) from our convenience methods.  We're not planning to do this everywhere, but there have been a handful of cases where the noise of an expected exception degraded the observability/diagnosability experience for customers.  This was discussed in detail in the ArchBoard meetings at https://github.com/Azure/azure-sdk/issues/3391.  This work item tracks enabling the first set of experiences for HLC libraries.

#### Foundational
- [x] Add `RequestOptions` - https://github.com/Azure/autorest.csharp/issues/1666
- [x] Support `RequestOptions` via the AutoRest's HLC generator - https://github.com/Azure/autorest.csharp/issues/1734

#### Use internally for both sync/async overloads:
- [ ] (Azure.Storage.Blobs) Blob Download - https://github.com/Azure/azure-sdk-for-net/issues/9934
- [x] (Azure.Storage.Blobs) Blob Exists - https://github.com/Azure/azure-sdk-for-net/issues/18592

#### Expose publicly both sync/async overloads:
- [ ] (Microsoft.Azure.WebJobs.Extensions.Storage) Receipt Manager - https://github.com/Azure/azure-sdk-for-net/issues/21511

#### Could either expose publicly or add an `If(Not)Exists` method to use internally for both sync/async overloads:
- [ ] (Azure.Search.Documents) Get Document - https://github.com/Azure/azure-sdk-for-net/issues/17445
- [ ] (Azure.Security.KeyVault.*) Delete - https://github.com/Azure/azure-sdk-for-net/issues/25125
- [x] (Azure.Data.Tables) CreateIfNotExists, CreateTableIfNotExists, GetEntitiy - https://github.com/Azure/azure-sdk-for-net/issues/16251
- [ ] (Azure.Data.AppConfiguration) Create - https://github.com/Azure/azure-sdk-for-net/issues/14802#issuecomment-996976716

longtimedeveloper:
@tg-msft can you please add both 'Azure.Data.Tables.GetEntity' and 'Azure.Data.Tables.GetEntityAsync'? 

Thank you!!

tg-msft:
[...]

Issue #21511, opened 2021-06-02
[BUG] Lease already present after upgrading to Microsoft.Azure.WebJobs.Extensions.Storage 5.0.0-beta.3


mauve:
**Describe the bug**

```
2021-06-02T12:43:49.889 [Warning] Error response [a89392d0-8a3c-44ba-b55e-8c56563e74ec] 409 There is already a lease present. (00.0s)Server:Windows-Azure-Blob/1.0,Microsoft-HTTPAPI/2.0x-ms-request-id:5a92e3cb-801e-002d-1eac-57e3e5000000x-ms-client-request-id:a89392d0-8a3c-44ba-b55e-8c56563e74ecx-ms-version:2020-04-08x-ms-error-code:LeaseAlreadyPresentDate:Wed, 02 Jun 2021 12:43:49 GMTContent-Length:221Content-Type:application/xml
```

**Expected behavior**

No warning

**Actual behavior (include Exception or Stack Trace)**

**To Reproduce**

1. update to 5.0.0-beta.3

**Environment:**
 - azure functions v3
 - dotnet core 3.1
 - Microsoft.Azure.WebJobs.Extensions.Storage 5.0.0-beta.3

All project dependencies:

```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <TargetFramework>netcoreapp3.1</TargetFramework>
    <AzureFunctionsVersion>V3</AzureFunctionsVersion>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Azure.Identity" Version="1.4.0" />
    <PackageReference Include="Azure.Messaging.ServiceBus" Version="7.2.0-beta.3" />
    <PackageReference Include="Azure.Security.KeyVault.Secrets" Version="4.1.1" />
    <PackageReference Include="Azure.Storage.Blobs" Version="12.9.0-beta.4" />
    <PackageReference Include="Microsoft.Azure.Cosmos" Version="3.19.0" />
    <PackageReference Include="Microsoft.Azure.Devices" Version="1.34.0" />
    <PackageReference Include="Microsoft.Azure.Functions.Extensions" Version="1.1.0" />
    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.DurableTask" Version="2.4.3" />
    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.Storage.Blobs" Version="5.0.0-beta.3" />
[...]

Issue #18592, opened 2021-02-09
[BUG] BlobBaseClient.Exists always logs exceptions to App Insights when the blob does not exist.


mrpmorris:
**Describe the bug**
BlobBaseClient.Exists files lots of false exceptions in App Insights when the blob does not exist.

**Expected behavior**
Checking if a blob exists should simply return false, not log an exception in App Insights.

**Actual behavior (include Exception or Stack Trace)**
App Insights is littered with false-positive exceptions due to a 404 response code to a HEAD request.

**To Reproduce**
Steps to reproduce the behavior (include a code snippet, screenshot, or any additional information that might help us reproduce the issue)

1. Create an app that connects to a valid blob storage
2. Enable AppInsights
3. Access BlobBaseClient.Exists for a blob you know does not exist
4. Look in AppInsights

**Environment:**
 - Name and version of the Library package used: Azure.Storage.Blobs 12.4.4
 - Hosting platform or OS and .NET runtime version: Azure Function app (net core 3.1)

The problem seems to be here
https://github.com/Azure/azure-sdk-for-net/blob/f01cbdc6b939331123af0c682892f3f478350f27/sdk/storage/Azure.Storage.Queues/src/Generated/ServiceRestClient.cs#L162

The code throws an exception for anything other than a 200 response - but 404 is not an exceptional response, it is very much expected - in fact, I expect to call it for non-existent blobs more often than any other blob scenario.

Ultimately, the exception is swallowed here
https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/storage/Azure.Storage.Blobs/src/BlobBaseClient.cs#L2952

But, by this point in time, the exception has already been logged to App Insights.

jsquire:
//cc: @tg-msft, as he was involved in an earlier discussion on this topic.

jsquire:
Thank you for your feedback.  Tagging and routing to the team best able to assist.

tg-msft:
This is waiting on #9934.

janid1967:
I am experiencing the same problem and looking forward to a fix of this

seanmcc-msft:
@tg-msft, what is the status of https://github.com/Azure/azure-sdk-for-net/issues/9934 ?

tg-msft:
[...]

PR #27645, opened 2022-03-18
[Draft] Minimize noisy Storage exceptions via DPG methods

tg-msft:
This shows how we can reduce logging noise for expected failures by using DPG methods from HLCs.  https://github.com/Azure/azure-sdk-for-net/issues/25626 is tracking the meta issue, but this is a proof of concept for both internal only (like BlobClient.Exists for https://github.com/Azure/azure-sdk-for-net/issues/18592) and externally visible (like BlobLeaseClient.Acquire for https://github.com/Azure/azure-sdk-for-net/issues/21511) cases.

This is currently blocked on https://github.com/Azure/azure-sdk-for-net/issues/27614 before we can ship it.

There are also a few tests in DataLake like FileClientTest.ExistsAsync_CPK complaining about not matching the recordings, but it's difficult to debug at the moment because of a known issue in the test proxy.  I'm ignoring those for the moment.

### DIFF
--- a/sdk/storage/Azure.Storage.Blobs/api/Azure.Storage.Blobs.netstandard2.0.cs
+++ b/sdk/storage/Azure.Storage.Blobs/api/Azure.Storage.Blobs.netstandard2.0.cs
@@ -1578,7 +1578,9 @@ namespace Azure.Storage.Blobs.Specialized
         protected virtual Azure.Storage.Blobs.BlobContainerClient BlobContainerClient { get { throw null; } }
         public virtual string LeaseId { get { throw null; } }
         public System.Uri Uri { get { throw null; } }
+        public virtual Azure.Response<Azure.Storage.Blobs.Models.BlobLease> Acquire(System.TimeSpan duration, Azure.RequestConditions conditions, Azure.RequestContext context) { throw null; }
         public virtual Azure.Response<Azure.Storage.Blobs.Models.BlobLease> Acquire(System.TimeSpan duration, Azure.RequestConditions conditions = null, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
+        public virtual System.Threading.Tasks.Task<Azure.Response<Azure.Storage.Blobs.Models.BlobLease>> AcquireAsync(System.TimeSpan duration, Azure.RequestConditions conditions, Azure.RequestContext context) { throw null; }
         public virtual System.Threading.Tasks.Task<Azure.Response<Azure.Storage.Blobs.Models.BlobLease>> AcquireAsync(System.TimeSpan duration, Azure.RequestConditions conditions = null, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
         public virtual Azure.Response<Azure.Storage.Blobs.Models.BlobLease> Break(System.TimeSpan? breakPeriod = default(System.TimeSpan?), Azure.RequestConditions conditions = null, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
         public virtual System.Threading.Tasks.Task<Azure.Response<Azure.Storage.Blobs.Models.BlobLease>> BreakAsync(System.TimeSpan? breakPeriod = default(System.TimeSpan?), Azure.RequestConditions conditions = null, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
--- a/sdk/storage/Azure.Storage.Blobs/src/BlobBaseClient.cs
+++ b/sdk/storage/Azure.Storage.Blobs/src/BlobBaseClient.cs
@@ -4448,38 +4448,24 @@ namespace Azure.Storage.Blobs.Specialized
                     scope.Start();
 
                     // Get the blob properties, but don't throw any exceptions
-                    RequestContext context = new()
-                    {
-                        ErrorOptions = ErrorOptions.NoThrow,
-                        CancellationToken = cancellationToken
-                    };
+                    RequestContext context =
+                        new RequestContext()
+                        {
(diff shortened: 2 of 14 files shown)


sdk/storage/Azure.Storage.Blobs/src/BlobBaseClient.cs
```diff
@@ -4442,39 +4442,63 @@ private void SetNameFieldsIfNull()
...
-                    return Response.FromValue(false, default);
-                }
-                catch (RequestFailedException storageRequestFailedException)
-                when (storageRequestFailedException.ErrorCode == BlobErrorCode.BlobUsesCustomerSpecifiedEncryption)
-                {
-                    return Response.FromValue(true, default);
+                    // Get the blob properties, but don't throw any exceptions
+                    RequestContext context = new()
+                    {
+                        ErrorOptions = ErrorOptions.NoThrow,
```
annelo-msft, line 4453: This will suppress an exception from being thrown from within GetProperties, but error logging and tracing will still occur in the pipeline.  In order to suppress those as well, you'll need to add classifiers using context.AddClassifier(...)

tg-msft: Sigh...  That's what I get for rushing and only checking the Storage tests.  I updated and verified that distributed tracing is clean too.

Current:
Current

Fixed:
Fixed

```C#
#define Fixed

using Azure;
using Azure.Core;
using Azure.Monitor.OpenTelemetry.Exporter;
using Azure.Storage.Blobs;
using Azure.Storage.Blobs.Models;
using Azure.Storage.Blobs.Specialized;
using OpenTelemetry.Trace;
using System.Diagnostics;

// Settings
const string StorageConnStr = "...";
const string AppInsightsConnStr = "...";
const string Name =
#if !Fixed
    "Current";
#else
    "Fixed";
#endif

// Collect telemetry in Azure Monitor
AppContext.SetSwitch("Azure.Experimental.EnableActivitySource", true);
using TracerProvider telemetry =
    OpenTelemetry.Sdk
    .CreateTracerProviderBuilder()
    .AddSource(Name, "Azure.*")
    .AddAzureMonitorTraceExporter(options => options.ConnectionString = AppInsightsConnStr)
    .Build();
using ActivitySource source = new(Name);
using Activity? group = source.StartActivity(Name)!;


// Check if a fake blob exists
BlobClient fake = new BlobClient(StorageConnStr, "fakecontainer", "fakeblob");
Console.WriteLine($"Exists: {fake.Exists().Value}");


// Try acquiring a lease from a fake blob
#if !Fixed

try
{
    BlobLease lease = fake.GetBlobLeaseClient().Acquire(TimeSpan.FromSeconds(15));
    Console.WriteLine($"Lease: {lease.LeaseId}");
}
catch (RequestFailedException ex)
{
    Console.WriteLine($"Failed to obtain lease!\n{ex}");
}

#else

RequestContext context =
    new RequestContext() { ErrorOptions = ErrorOptions.NoThrow }
    .AddClassifier(404, BlobErrorCode.BlobNotFound.ToString(), false)
    .AddClassifier(404, BlobErrorCode.ContainerNotFound.ToString(), false)
    .AddClassifier(409, BlobErrorCode.LeaseAlreadyPresent.ToString(), false);
Response<BlobLease> response =
    fake.GetBlobLeaseClient().Acquire(
        TimeSpan.FromSeconds(15),
        conditions: null,
        context);
Response raw = response.GetRawResponse();
if (raw.Status == 201)
{
    Console.WriteLine($"Lease: {response.Value.LeaseId}");
}
else
{
    Console.WriteLine($"Failed to obtain lease!\n{raw.Content.ToString()}");
}

#endif
```

sdk/storage/Azure.Storage.Blobs/api/Azure.Storage.Blobs.netstandard2.0.cs
```diff
@@ -1578,7 +1578,9 @@ public partial class BlobLeaseClient
         protected virtual Azure.Storage.Blobs.BlobContainerClient BlobContainerClient { get { throw null; } }
         public virtual string LeaseId { get { throw null; } }
         public System.Uri Uri { get { throw null; } }
+        public virtual Azure.Response<Azure.Storage.Blobs.Models.BlobLease> Acquire(System.TimeSpan duration, Azure.RequestConditions conditions, Azure.RequestContext context) { throw null; }
```
kasobol-msft, line 1581: How do we determine that we want to add `RequestContext` to HLC methods vs using it internally.
I.e. why this instead of `AcquireIfNotYetAcuired` API that uses `RequestContext` internally?

sdk/storage/Azure.Storage.Blobs/src/BlobBaseClient.cs
```diff
@@ -4442,39 +4442,59 @@ private void SetNameFieldsIfNull()
                     nameof(BlobBaseClient),
                     message:
                     $"{nameof(Uri)}: {Uri}");
-
-                string operationName = $"{nameof(BlobBaseClient)}.{nameof(Exists)}";
-
+                DiagnosticScope scope = ClientConfiguration.ClientDiagnostics.CreateScope($"{nameof(BlobBaseClient)}.{nameof(Exists)}");
```
annelo-msft, line 4445: Curious: why add a scope now when there wasn't one there before?

sdk/storage/Azure.Storage.Blobs/api/Azure.Storage.Blobs.netstandard2.0.cs
tg-msft: Because it's not really doing `AcquireIfNotYetAcquired`.  It's doing `AcquireIfBlobNotFoundOrContainerNotFoundOrLeaseAlreadyAcquired` and we don't want to add an overload for that specific, non-general scenario.

tg-msft:
Be gone from here, bot.

PR #27652, opened 2022-03-19
Add additional transaction tests

JoshLove-msft:
Validates that a sender can be reused across non-tx and tx simultaneously.


sdk/servicebus/Azure.Messaging.ServiceBus/tests/Transactions/TransactionLiveTests.cs
```diff
@@ -355,6 +355,53 @@ await receiver.CompleteMessageAsync(receivedMessage), Throws.InstanceOf<ServiceB
...
+
+                ServiceBusMessage message = ServiceBusTestUtilities.GetMessage();
+                await sender1.SendMessageAsync(message);
+
+                ServiceBusReceivedMessage receivedMessage = await receiver.ReceiveMessageAsync();
+                Assert.NotNull(receivedMessage);
+
+                var tcs = new TaskCompletionSource<bool>(TaskCreationOptions.RunContinuationsAsynchronously);
+
+                _ = SendMessageAsync();
```
jsquire, line 386: Should we consider capturing the task and awaiting at the very end of the test, just to force any errors to surface?

jsquire (approved):

PR #27037, opened 2022-02-16
Initial set of diagnostics

JoshLove-msft:
Contributes to https://github.com/Azure/azure-sdk-for-net/issues/8561

PR #27663, opened 2022-03-21
Add ServiceBusTransportMetrics

JoshLove-msft:
Second iteration of https://github.com/Azure/azure-sdk-for-net/pull/27037

### DIFF
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpUsageMeter.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpUsageMeter.cs
@@ -1,0 +1,48 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using Microsoft.Azure.Amqp;
+
+namespace Azure.Messaging.ServiceBus.Amqp
+{
+    internal class AmqpUsageMeter : IAmqpUsageMeter
+    {
+        private readonly ServiceBusTransportMetrics _metrics;
+
+        public AmqpUsageMeter(ServiceBusTransportMetrics metrics)
+        {
+            _metrics = metrics;
+        }
+        public void OnTransportWrite(int bufferSize, int writeSize, long queueSize, long latencyTicks)
+        {
+            // not implemented
+        }
+
+        public void OnTransportRead(int bufferSize, int readSize, int cacheHits, long latencyTicks)
+        {
+            // not implemented
+        }
+
+        public void OnRead(AmqpConnection connection, ulong frameCode, int numberOfBytes)
+        {
+            switch (frameCode)
+            {
+                case 0x00:
+                    _metrics.LastHeartBeat = DateTimeOffset.UtcNow;
+                    break;
+                case 0x10:
+                    _metrics.LastConnectionOpen = DateTimeOffset.UtcNow;
+                    break;
+                case 0x18:
+                    _metrics.LastConnectionClose = DateTimeOffset.UtcNow;
+                    break;
+            }
+        }
+
+        public void OnWrite(AmqpConnection connection, ulong frameCode, int numberOfBytes)
+        {
+            // not implemented
+        }
+    }
+}
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpUsageMeter.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpUsageMeter.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Threading;
 using Microsoft.Azure.Amqp;
 
 namespace Azure.Messaging.ServiceBus.Amqp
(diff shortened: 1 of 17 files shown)


sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpUsageMeter.cs
```diff
@@ -0,0 +1,48 @@
...
+        {
+            // not implemented
+        }
+
+        public void OnRead(AmqpConnection connection, ulong frameCode, int numberOfBytes)
+        {
+            switch (frameCode)
+            {
+                case 0x00:
+                    _metrics.LastHeartBeat = DateTimeOffset.UtcNow;
```
jsquire, line 32: Are we sure that this is an atomic assignment?   `DateTimeOffset` is a struct and will, I believe, need to be boxed.   Should we do an `Interlocked.Exchange` here to be safe?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Client/ServiceBusClient.cs
```diff
@@ -87,17 +87,24 @@ public virtual async ValueTask DisposeAsync()
...
         /// <summary>
-        /// Can be used for mocking.
+        /// Gets the metrics associated with this <see cref="ServiceBusClient"/> instance. In order to use this property,
+        /// <see cref="ServiceBusClientOptions.EnableTransportMetrics"/> must be set to <value>true</value>.
         /// </summary>
-        protected ServiceBusClient()
-        {
-        }
+        public ServiceBusTransportMetrics TransportMetrics
+            => Connection.InnerClient.Metrics ?? throw new InvalidOperationException("Transport metrics are not enabled. To enable transport metrics, set the EnableTransportMetrics property on the ServiceBusClientOptions.");
```
jsquire, line 94: Should we consider cloning the instance to ensure that data doesn't mutate for callers?   As a caller, I would expect that anything you return to me will be stable when I'm reading it.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Client/ServiceBusTransportMetrics.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// Initializes a new instance of the <see cref="ServiceBusTransportMetrics"/> class for mocking.
+        /// </summary>
+        protected internal ServiceBusTransportMetrics()
+        {
+        }
+        /// <summary>
+        /// Gets the last time that a heartbeat was received from the Service Bus service. These heartbeats are sent from the
+        /// service approximately every 30 seconds.
+        /// </summary>
+        public virtual DateTimeOffset? LastHeartBeat { get; internal set; }
```
jsquire, line 23: Do these need to be virtual?   Is there a scenario where creating an instance via the model factory wouldn't be sufficient?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Core/TransportClient.cs
```diff
@@ -35,6 +35,11 @@ internal abstract class TransportClient : IAsyncDisposable
         ///
         public virtual Uri ServiceEndpoint { get; }
 
+        /// <summary>
+        /// The metrics related to the client.
+        /// </summary>
+        public virtual ServiceBusTransportMetrics Metrics { get; }
```
jsquire, line 41: Should we consider calling this `TransportMetrics`, since the associated type is strongly bound to that context?

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Infrastructure/ServiceBusLiveTestBase.cs
```diff
@@ -70,5 +73,18 @@ protected ServiceBusClient CreateClient(int tryTimeout = DefaultTryTimeout)
                 batch = default;
             }
         }
+
+        protected static void SimulateNetworkFailure(ServiceBusClient client)
```
jsquire, line 77: Clever!

sdk/servicebus/Azure.Messaging.ServiceBus/src/Client/ServiceBusClient.cs
JoshLove-msft: We could, but then I think we'd need to switch to a method rather than property, which might be okay.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Client/ServiceBusTransportMetrics.cs
JoshLove-msft: A model factory would work as well.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpUsageMeter.cs
JoshLove-msft: Nullable is a struct.

jsquire (approved):

Issue #23842, opened 2021-09-07
Avro Schema Registry serializer requires group name


JoshLove-msft:
It makes sense that group name is required for serialization, as it is required to look up the schema ID since there could be multiple, identical schemas in different schema groups.

When deserializing, you shouldn't need to specify a group name because the schema ID is unique across the endpoint.

JoshLove-msft:
Discussed using an optional parameter here. We would validate that the group name is not null in the serialize method.

PR #27687, opened 2022-03-22
Fix issue with null options and add constructor that omits groupName.

JoshLove-msft:
Addresses https://github.com/Azure/azure-sdk-for-net/issues/23842


sdk/schemaregistry/Microsoft.Azure.Data.SchemaRegistry.ApacheAvro/src/SchemaRegistryAvroSerializer.cs
```diff
@@ -118,12 +147,20 @@ private enum SupportedType
             bool async,
             CancellationToken cancellationToken)
         {
+            if (_groupName == null)
```
jsquire, line 150: What if it's empty or whitespace?   It doesn't appear that we're doing any validation for those in the constructor.

jsquire (approved):

christothes (approved):

sdk/schemaregistry/Microsoft.Azure.Data.SchemaRegistry.ApacheAvro/src/SchemaRegistryAvroSerializer.cs
JoshLove-msft: We'd just send it on to the service and it wouldn't work I presume.

sdk/keyvault/Azure.Security.KeyVault.Certificates/tests/CertificatesTestBase.cs
```diff
@@ -42,6 +42,8 @@ public CertificatesTestBase(bool isAsync, CertificateClientOptions.ServiceVersio
             : base(isAsync, mode)
         {
             _serviceVersion = serviceVersion;
+            // temporary until https://github.com/Azure/azure-sdk-for-net/issues/27688 is addressed
```
heaths, line 45: Based on the history of this and the KeysTestBase.cs, it's probably fine to still compare bodies despite our offline conversation. Maybe I fixed the problem I mentioned a while ago. But I do know there's a few other libraries I worked on / own that I set this to `false` because of floating point serialization issues between .NET Framework and .NET (Core).

heaths (approved):
General and KV changes LGTM.

sdk/schemaregistry/Microsoft.Azure.Data.SchemaRegistry.ApacheAvro/api/Microsoft.Azure.Data.SchemaRegistry.ApacheAvro.netstandard2.0.cs
```diff
@@ -12,7 +12,9 @@ public partial class AvroSerializationException : System.Exception
     }
     public partial class SchemaRegistryAvroSerializer
     {
-        public SchemaRegistryAvroSerializer(Azure.Data.SchemaRegistry.SchemaRegistryClient client, string groupName, Microsoft.Azure.Data.SchemaRegistry.ApacheAvro.SchemaRegistryAvroSerializerOptions options = null) { }
+        public SchemaRegistryAvroSerializer(Azure.Data.SchemaRegistry.SchemaRegistryClient client) { }
+        public SchemaRegistryAvroSerializer(Azure.Data.SchemaRegistry.SchemaRegistryClient client, string groupName) { }
+        public SchemaRegistryAvroSerializer(Azure.Data.SchemaRegistry.SchemaRegistryClient client, string groupName, Microsoft.Azure.Data.SchemaRegistry.ApacheAvro.SchemaRegistryAvroSerializerOptions options) { }
```
deyaaeldeen, line 17: groupName is part of the options in JS because it is optional. It is only needed when serializing but it is valid to use the serializer for deserializing only.

JoshLove-msft: I don't think it would make sense to put in the options for .NET, as it makes the mainline scenario more complicated.

kinelski (approved):
FR looks good to me. I'll investigate the issues later today.

sdk/schemaregistry/Microsoft.Azure.Data.SchemaRegistry.ApacheAvro/api/Microsoft.Azure.Data.SchemaRegistry.ApacheAvro.netstandard2.0.cs
JoshLove-msft: when serializing, I'd need to do:
```c#
var serializer = new SchemaRegistryAvroSerializer(client, new SchemaRegistryAvroSerializerOptions { GroupName = "groupName"});
```
rather than:
```c# 
var serializer = new SchemaRegistryAvroSerializer(client, "groupName");
```

PR #27708, opened 2022-03-23
[FormRecognizer] Enable StartBuildModelSucceedsWithValidPrefix

### LINKED ISSUE #24448: Enable StartTrainingSucceedsWithValidPrefix
@maririos commented on [Mon Aug 30 2021](https://github.com/Azure/azure-sdk-for-net-pr/issues/1442)

Currently service fails when trying to set a prefix

kinelski:
Fixes https://github.com/azure/azure-sdk-for-net/issues/24448.


sdk/formrecognizer/Azure.AI.FormRecognizer/tests/DocumentModelAdministrationClient/DocumentModelAdministrationClientLiveTests.cs
```diff
@@ -113,14 +113,13 @@ public async Task StartBuildModelWithNeuralBuildMode()
...
         [RecordedTest]
-        [Ignore("Issue https://github.com/Azure/azure-sdk-for-net-pr/issues/1442")]
         public async Task StartBuildModelSucceedsWithValidPrefix()
         {
             var client = CreateDocumentModelAdministrationClient();
             var trainingFilesUri = new Uri(TestEnvironment.BlobContainerSasUrl);
             var modelId = Recording.GenerateId();
 
-            BuildModelOperation operation = await client.StartBuildModelAsync(trainingFilesUri, DocumentBuildMode.Template, modelId, new BuildModelOptions() { Prefix = "subfolder" });
+            BuildModelOperation operation = await client.StartBuildModelAsync(trainingFilesUri, DocumentBuildMode.Template, modelId, new BuildModelOptions() { Prefix = "subfolder/" });
```
kinelski, line 122: Test was failing for 2 reasons:
1. Appending a slash `/` to the subfolder name is required. This behavior is intentional, according to the service team, but they intend to improve this scenario.
2. Only the subfolder specified by `Prefix` will be probed when searching for training files, so we need to include all files in the `subfolder` directory. We had only 1 file there and were expecting the service to find the other ones in the root. We have updated our training container to fix this, and we're also uploading the files to our test project so contributors can use them when running our tests.

jsquire (approved):

PR #27713, opened 2022-03-23
[DPG] Add service driven evolution documentation

fengzhou-msft:

### DIFF
--- a/doc/Data Plane Code Generation/ServiceDrivenEvolution.md
+++ b/doc/Data Plane Code Generation/ServiceDrivenEvolution.md
@@ -1,6 +1,8 @@
 # Service driven evolution
 
-As a service evolves with non-breaking changes in the REST API, it may or may not cause breaking changes in the generated SDK. For instance, adding a new method in an existing path or in a new path won't cause breaking changes in the new SDK. But some changes like adding an optional query parameter could cause breaking changes in the generated SDK. In order to make the SDK backward compatible, we can add customized code such as overload methods. The following sections will show some examples of such scenarios and how to handle them.
+As a service evolves with non-breaking changes in the REST API, it may or may not cause breaking changes in the generated SDK. For instance, adding a new method in an existing path or in a new path won't cause breaking changes in the new SDK. See details [here](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#api-versioning). 
+
+However, even no breaking changes exist from the REST API perspective, some changes like adding an optional query parameter could still cause breaking changes in the generated SDK. In order to make the SDK backward compatible, we can add customized code such as overload methods for now. We will make it automatically in the near future. The following sections will show some examples of such scenarios and how to handle them.
 
 ## A method gets a new optional parameter
 
@@ -9,20 +11,13 @@ Generated code in a V1 client with a required query parameter and an optional qu
 ``` C#
 public virtual async Task<Response> PutRequiredOptionalAsync(string requiredParam, string optionalParam = null, RequestContext context = null)
 {
-    Argument.AssertNotNull(requiredParam, nameof(requiredParam));
-
-    using var scope = ClientDiagnostics.CreateScope("ParamsClient.PutRequiredOptional");
-    scope.Start();
+    ...
     try
     {
         using HttpMessage message = CreatePutRequiredOptionalRequest(requiredParam, optionalParam, context);
         return await _pipeline.ProcessMessageAsync(message, context).ConfigureAwait(false);
     }
-    catch (Exception e)
-    {
-        scope.Failed(e);
-        throw;
-    }
+    ...
 }
 ```
 
@@ -31,20 +26,13 @@ Generated code in a V2 client with a new optional query parameter:
 ``` C#
 public virtual async Task<Response> PutRequiredOptionalAsync(string requiredParam, string optionalParam = null, string newParameter = null, RequestContext context = null)
 {
-    Argument.AssertNotNull(requiredParam, nameof(requiredParam));
-
-    using var scope = ClientDiagnostics.CreateScope("ParamsClient.PutRequiredOptional");
-    scope.Start();
+    ...
     try
     {
         using HttpMessage message = CreatePutRequiredOptionalRequest(requiredParam, optionalParam, newParameter, context);
         return await _pipeline.ProcessMessageAsync(message, context).ConfigureAwait(false);
     }
-    catch (Exception e)
-    {
-        scope.Failed(e);
-        throw;
-    }
+    ...
 }
 ```
 
@@ -53,22 +41,18 @@ Since `context` is always the last parameter, V2 inserts the new optional parame
 ``` C#
 public virtual async Task<Response> PutRequiredOptionalAsync(string requiredParam, string optionalParam, RequestContext context)
 {
-    Argument.AssertNotNull(requiredParam, nameof(requiredParam));
-
-    using var scope = ClientDiagnostics.CreateScope("ParamsClient.PutRequiredOptional");
-    scope.Start();
+    ...
     try
     {
         using HttpMessage message = CreatePutRequiredOptionalRequest(requiredParam, optionalParam, null, context);
         return await _pipeline.ProcessMessageAsync(message, context).ConfigureAwait(false);
     }
-    catch (Exception e)
-    {
-        scope.Failed(e);
-        throw;
-    }
+    ...
 }
 ```
+
+**Notice:**
+Now the helper method has one mpre parameter, so we should call it with `newParameter` `null` as `CreatePutRequiredOptionalRequest(requiredParam, optionalParam, null, context);`.
 
 Now calling `client.PutRequiredOptionalAsync("requiredParam", "optionalParam", ErrorOptions.NoThrow)` with V2 will invoke the manual method we added.
 
@@ -79,32 +63,18 @@ Generated code in a V1 client with only `application/json` as the content type:
 ``` C#
 public virtual async Task<Response> PostParametersAsync(RequestContent content, RequestContext context = null)
 {
-    Argument.AssertNotNull(content, nameof(content));
-
-    using var scope = ClientDiagnostics.CreateScope("ParamsClient.PostParameters");
-    scope.Start();
+    ...
     try
     {
         using HttpMessage message = CreatePostParametersRequest(content, context);
         return await _pipeline.ProcessMessageAsync(message, context).ConfigureAwait(false);
     }
-    catch (Exception e)
-    {
-        scope.Failed(e);
-        throw;
-    }
+    ...
 }
 
 internal HttpMessage CreatePostParametersRequest(RequestContent content, RequestContext context)
 {
-    var message = _pipeline.CreateMessage(context, ResponseClassifier200);
-    var request = message.Request;
-    request.Method = RequestMethod.Post;
-    var uri = new RawRequestUriBuilder();
-    uri.Reset(_endpoint);
-    uri.AppendPath("/serviceDriven/parameters", false);
-    request.Uri = uri;
-    request.Headers.Add("Accept", "application/json");
+    ...
     request.Headers.Add("Content-Type", "application/json");
     request.Content = content;
     return message;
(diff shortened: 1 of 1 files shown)


doc/Data Plane Code Generation/ServiceDrivenEvolution.md
```diff
@@ -0,0 +1,177 @@
...
+As a service evolves with non-breaking changes in the REST API, it may or may not cause breaking changes in the generated SDK. For instance, adding a new method in an existing path or in a new path won't cause breaking changes in the new SDK. But some changes like adding an optional query parameter could cause breaking changes in the generated SDK. In order to make the SDK backward compatible, we can add customized code such as overload methods. The following sections will show some examples of such scenarios and how to handle them.
+
+## A method gets a new optional parameter
+
+Generated code in a V1 client with a required query parameter and an optional query parameter:
+
+``` C#
+public virtual async Task<Response> PutRequiredOptionalAsync(string requiredParam, string optionalParam = null, RequestContext context = null)
+{
+    Argument.AssertNotNull(requiredParam, nameof(requiredParam));
```
annelo-msft, line 12: We should not be validating query parameters in our clients.
https://azure.github.io/azure-sdk/dotnet_introduction.html#parameter-validation

doc/Data Plane Code Generation/ServiceDrivenEvolution.md
```diff
@@ -0,0 +1,177 @@
...
+Now calling `client.PutRequiredOptionalAsync("requiredParam", "optionalParam", ErrorOptions.NoThrow)` with V2 will invoke the manual method we added.
+
+## A new body type is added
+
+Generated code in a V1 client with only `application/json` as the content type:
+
+``` C#
+public virtual async Task<Response> PostParametersAsync(RequestContent content, RequestContext context = null)
+{
+    Argument.AssertNotNull(content, nameof(content));
```
annelo-msft, line 82: We shouldn't be validating request bodies in our clients: https://azure.github.io/azure-sdk/dotnet_introduction.html#parameter-validation

doc/Data Plane Code Generation/ServiceDrivenEvolution.md
```diff
@@ -0,0 +1,177 @@
...
+    uri.AppendPath("/serviceDriven/parameters", false);
+    request.Uri = uri;
+    request.Headers.Add("Accept", "application/json");
+    request.Headers.Add("Content-Type", contentType.ToString());
+    request.Content = content;
+    return message;
+}
+```
+
+Since the V2 client added a required parameter, it breaks the code that is written for V1. We can manually add an overload method that has the same signature as V1 to make it backward compatible:
```
annelo-msft, line 154:
```suggestion
Since the V2 client added a required parameter to an existing client method, it breaks the code that is written for the V1 client. We can manually add an overload method that has the same signature as V1 to make it backward compatible:
```

doc/Data Plane Code Generation/ServiceDrivenEvolution.md
```diff
@@ -0,0 +1,177 @@
...
+    }
+    catch (Exception e)
+    {
+        scope.Failed(e);
+        throw;
+    }
+}
+```
+
+Notice that we leverage the `CreatePostParametersRequest` method in V2 but always pass in the `ContentType.ApplicationJson` as the `contentType` parameter in the manual method to make the behavior align with V1. We can add `[EditorBrowsable(EditorBrowsableState.Never)]` to the manual method to make it invisible in IDE if we don't want to confuse new customers that never used V1 with this overload method that uses a hard-coded content type in addition to the generated method that uses a `contentType` parameter.
```
annelo-msft, line 177: For this part:

> We can add `[EditorBrowsable(EditorBrowsableState.Never)]` to the manual method to make it invisible in IDE if we don't want to confuse new customers that never used V1 with this overload method that uses a hard-coded content type in addition to the generated method that uses a `contentType` parameter.

It would be good to call out that the business rules for the service are what would determine whether or not to add `[EditorBrowsable(EditorBrowsableState.Never)]`.  If the service added the new contentType, for example, to improve performance, and wants to discourage callers from using the old contentType, then they would definitely want to add `[EditorBrowsable(EditorBrowsableState.Never)]` to the old method so new users wouldn't discover it.  If the old contentType remains the default preference, they wouldn't want to add `[EditorBrowsable(EditorBrowsableState.Never)]` to make it easy for users to discover the default.

annelo-msft (commented):
This is looking really good, thanks, @fengzhou-msft!

I left a few comments.

annelo-msft:
@lirenhe, who is working on this PR now?

doc/Data Plane Code Generation/ServiceDrivenEvolution.md
```diff
@@ -0,0 +1,177 @@
...
+    }
+}
+```
+
+Since `context` is always the last parameter, V2 inserts the new optional parameter before `context` parameter and this could break code that works with V1 such as `client.PutRequiredOptionalAsync("requiredParam", "optionalParam", ErrorOptions.NoThrow)` (`ErrorOptions.NoThrow` was passed in as `context` parameter but now it's treated as `newParameter`). We can manually add an overload method in V2 which transforms the optional parameters in V1 to required parameters to make V2 backward compatible with V1:
+
+``` C#
+public virtual async Task<Response> PutRequiredOptionalAsync(string requiredParam, string optionalParam, RequestContext context)
+{
+    Argument.AssertNotNull(requiredParam, nameof(requiredParam));
```
lirenhe, line 56: Do we need to validate other parameters of this function?

doc/Data Plane Code Generation/ServiceDrivenEvolution.md
```diff
@@ -0,0 +1,177 @@
...
+
+    using var scope = ClientDiagnostics.CreateScope("ParamsClient.PutRequiredOptional");
```
lirenhe, line 58: It looks this part the same compared with 
PutRequiredOptionalAsync(string requiredParam, string optionalParam = null, string newParameter = null, RequestContext context = null)

So can we use the following in function body directly?  
  this.PutRequiredOptionalAsync(requiredParam, optionalParam, null, context);

doc/Data Plane Code Generation/ServiceDrivenEvolution.md
pshao25: Just curious: what is the difference between URL segments and path parameters. It says we should validate path parameter, while should not validate URL segments.

annelo-msft: @AlexanderSher, is this a bug in the guidelines?

doc/Data Plane Code Generation/ServiceDrivenEvolution.md
annelo-msft: https://azure.github.io/azure-sdk/general_implementation.html#parameter-validation

doc/Data Plane Code Generation/ServiceDrivenEvolution.md
pshao25: I created an issue here: https://github.com/Azure/autorest.csharp/issues/2155

lirenhe:
@annelo-msft,  @pshao25 will continue working on it. Thanks

doc/Data Plane Code Generation/ServiceDrivenEvolution.md
pshao25: It is OK technically. Not sure if there is any other aspects should take care. I would expect @annelo-msft to comment.

doc/Data Plane Code Generation/ServiceDrivenEvolution.md
```diff
@@ -0,0 +1,191 @@
...
+    {
+        using HttpMessage message = CreatePutRequiredOptionalRequest(requiredParam, optionalParam, null, context);
+        return await _pipeline.ProcessMessageAsync(message, context).ConfigureAwait(false);
+    }
+    ...
+}
+```
+
+**Notice:**
+Now the helper method has one mpre parameter, so we should call it with `newParameter` `null` as `CreatePutRequiredOptionalRequest(requiredParam, optionalParam, null, context);`.
```
lirenhe, line 55:
```suggestion
Now the helper method has one more parameter, so we should call it with `newParameter` `null` as `CreatePutRequiredOptionalRequest(requiredParam, optionalParam, null, context);`.
```
```suggestion
Now the helper method has one mpre parameter, so we should call it with `newParameter` `null` as `CreatePutRequiredOptionalRequest(requiredParam, optionalParam, null, context);`.
```

lirenhe, line 55:
```suggestion
Now the helper method has one more parameter, so we should call it with `newParameter` `null` as `CreatePutRequiredOptionalRequest(requiredParam, optionalParam, null, context);`.
```
```suggestion
Now the helper method has one mpre parameter, so we should call it with `newParameter` `null` as `CreatePutRequiredOptionalRequest(requiredParam, optionalParam, null, context);`.
```

lirenhe, line 55:
```suggestion
Now the helper method has one more parameter, so we should call it with `newParameter` `null` as `CreatePutRequiredOptionalRequest(requiredParam, optionalParam, null, context);`.
```
```suggestion
Now the helper method has one mpre parameter, so we should call it with `newParameter` `null` as `CreatePutRequiredOptionalRequest(requiredParam, optionalParam, null, context);`.
```

lirenhe, line 55:
```suggestion
Now the helper method has one more parameter, so we should call it with `newParameter` `null` as `CreatePutRequiredOptionalRequest(requiredParam, optionalParam, null, context);`.
```
```suggestion
Now the helper method has one mpre parameter, so we should call it with `newParameter` `null` as `CreatePutRequiredOptionalRequest(requiredParam, optionalParam, null, context);`.
```

doc/Data Plane Code Generation/ServiceDrivenEvolution.md
```diff
@@ -0,0 +1,191 @@
...
+    {
+        using HttpMessage message = CreatePutRequiredOptionalRequest(requiredParam2, requiredParam1, context);
+        return await _pipeline.ProcessMessageAsync(message, context).ConfigureAwait(false);
+    }
+    ...
+}
+```
+You have to handle this via a manually-written convenience overload. E.g.,
+``` C#
+public virtual async Task<Response> PutRequiredOptionalAsync(string requiredParam2, string requiredParam1 = null, RequestContext context = null)
```
lirenhe, line 179: For this part, i don't quite get it.  It seems the api signature is the same, but you change the parameter order inside  CreatePutRequiredOptionalRequest.   Will the latter one replace the generated code?

lirenhe, line 179: For this part, i don't quite get it.  It seems the api signature is the same, but you change the parameter order inside  CreatePutRequiredOptionalRequest.   Will the latter one replace the generated code?

pshao25: Yeah, this suggestion is incorrect. So I put another one.

lirenhe (approved):

doc/Data Plane Code Generation/ServiceDrivenEvolution.md
pshao25: This looks by design, see https://github.com/Azure/autorest.csharp/issues/2155#issuecomment-1101111229

doc/Data Plane Code Generation/ServiceDrivenEvolution.md
```diff
@@ -0,0 +1,179 @@
...
+    {
+        using HttpMessage message = CreatePutRequiredOptionalRequest(requiredParam, optionalParam, null, context);
+        return await _pipeline.ProcessMessageAsync(message, context).ConfigureAwait(false);
+    }
+    ...
+}
+```
+
+**Notice:**
+Now the helper method has one more parameter, so we should call it with `newParameter` `null` as `CreatePutRequiredOptionalRequest(requiredParam, optionalParam, null, context);`.
```
annelo-msft, line 55: I don't understand this.  Why should we call it with null?  In what case do we want to do that, and to achieve what goal?

annelo-msft, line 55: I don't understand this.  Why should we call it with null?  In what case do we want to do that, and to achieve what goal?

doc/Data Plane Code Generation/ServiceDrivenEvolution.md
```diff
@@ -0,0 +1,179 @@
...
+
+Now calling `client.PutRequiredOptionalAsync("requiredParam", "optionalParam", ErrorOptions.NoThrow)` with V2 will invoke the manual method we added.
```
annelo-msft, line 57: True, but it's not clear to me why we need to say this.  Executing this line of code would not have worked before, correct?

annelo-msft, line 57: True, but it's not clear to me why we need to say this.  Executing this line of code would not have worked before, correct?

doc/Data Plane Code Generation/ServiceDrivenEvolution.md
annelo-msft, line 55: We're calling these "helper method" and "manual method" -- there's some ambiguity there.  Is there a more clear name we could use?

"helper" implies a convenience method, but these aren't provided for convenience so much as backward compatibility.


My thinking is this.  I find it hard to communicate about these methods without stating clearly:
- What the client looked like in v_1
- What was added to the REST API in v_2
- What the client looks like in v_2
 
If you do this, you can start to see the pattern, which is that in new versions of the client, we make the longest overload take default parameters and make the parameters of shorter overloads from earlier versions all required.  This then maps to recommendations from the .NET Framework Design Guidelines, such as:
- CONSIDER using default parameters on the longest overload of a method.
- DO provide a simple overload, with no default parameters, for any method with two or more defaulted parameters.
- DO NOT have two overloads of a method with “compatible” required parameters that both use default parameters.
- AVOID having two overloads of the same method that both use default parameters.
- DO move all default parameters to the new, longer overload when adding optional parameters to an existing method.

See: https://learning.oreilly.com/library/view/framework-design-guidelines/9780135896457/ch05.xhtml#sec5_1

annelo-msft, line 55: We're calling these "helper method" and "manual method" -- there's some ambiguity there.  Is there a more clear name we could use?

doc/Data Plane Code Generation/ServiceDrivenEvolution.md
```diff
@@ -0,0 +1,179 @@
...
+    try
+    {
+        using HttpMessage message = CreatePutRequiredOptionalRequest(param1, param2, context);
+        return await _pipeline.ProcessMessageAsync(message, context).ConfigureAwait(false);
+    }
+    ...
+}
+```
+
+**Required input is changed to an optional input when it is not in the last position (e.g., `param1`).**
```
annelo-msft, line 162: It feels like this section would go better after the first section -- parameter ordering for these cases is closely related.

annelo-msft, line 162: It feels like this section would go better after the first section -- parameter ordering for these cases is closely related.

annelo-msft (commented):
This is looking good - added a few more comments.

doc/Data Plane Code Generation/ServiceDrivenEvolution.md
pshao25: This aimed to be compatible with V1 call. In V1, there is no `newParameter`, so we pass in `null`.

doc/Data Plane Code Generation/ServiceDrivenEvolution.md
pshao25: Yes, `client.PutRequiredOptionalAsync("requiredParam", "optionalParam", ErrorOptions.NoThrow)` worked in V1. But in V2, the method signature becomes 
```
public virtual async Task<Response> PutRequiredOptionalAsync(string requiredParam, string optionalParam = null, string newParameter = null, RequestContext context = null)
```
So, the V1 call `client.PutRequiredOptionalAsync("requiredParam", "optionalParam", ErrorOptions.NoThrow)` does not work any more. So we add this customization.

doc/Data Plane Code Generation/ServiceDrivenEvolution.md
pshao25: I think we do follow this. I made some rephrasing. Hope it could be more clear.

doc/Data Plane Code Generation/ServiceDrivenEvolution.md
```diff
@@ -50,130 +76,147 @@ public virtual async Task<Response> PutRequiredOptionalAsync(string requiredPara
...
 ```
+In summary of what this evolution looks like now:
+```C#
+// Generated in V2
+public virtual async Task<Response> PutRequiredOptionalAsync(string requiredParam, string optionalParam = null, string newParameter = null, RequestContext context = null)
+{
+}
+
+// Generated in V2
+internal HttpMessage CreatePutRequiredOptionalRequest(string requiredParam, string optionalParam, string newParameter, RequestContext context)
```
annelo-msft, line 87: Please remove the CreateRequest methods from the sample, as they are not part of the public API.

doc/Data Plane Code Generation/ServiceDrivenEvolution.md
```diff
@@ -50,130 +76,147 @@ public virtual async Task<Response> PutRequiredOptionalAsync(string requiredPara
...
+
+// Customization method in V2 to be compatible with V1 calls
+public virtual async Task<Response> PutRequiredOptionalAsync(string requiredParam, string optionalParam, RequestContext context)
+{
+}
+```
 
 **Notice:**
-Now the helper method has one more parameter, so we should call it with `newParameter` `null` as `CreatePutRequiredOptionalRequest(requiredParam, optionalParam, null, context);`.
+Now the customization method has one more parameter, so we should call it with `newParameter` `null` as `CreatePutRequiredOptionalRequest(requiredParam, optionalParam, null, context);`.
```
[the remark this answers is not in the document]
annelo-msft: I see - it's not a helper or customization method, it's the create request method.  I understand now, let's change `customization` to `CreateXxRequest` method.

doc/Data Plane Code Generation/ServiceDrivenEvolution.md
annelo-msft: Replied below.

annelo-msft (approved):

doc/Data Plane Code Generation/ServiceDrivenEvolution.md
pshao25: As synced, we will keep this.

PR #27718, opened 2022-03-23
Add App Service 2019 endpoint to Managed Identity Sources

### LINKED ISSUE #26578: Switch to using the newest API version for Managed Identity tokens on App Service platform
https://github.com/Azure/azure-sdk-for-net/blob/0909c641d6066094b524028cb14eadc50f8a6df2/sdk/identity/Azure.Identity/src/ManagedIdentityClient.cs#L45

Current ManagedIdentityClient uses the 2017 API, but the 2019 API is in this class: https://github.com/Azure/azure-sdk-for-net/blob/0909c641d6066094b524028cb14eadc50f8a6df2/sdk/identity/Azure.Identity/src/AppServiceV2019ManagedIdentitySource.cs#L12

Please switch over to using the new API as it enforces proper behavior on choosing the right Managed Identities

Changes with the new 2019 api versus old 2017 api:
- query string parameters changed from camel case (2017) to snake case (2019)
- If a user assigned identity is specified in the query parameter, a BadRequest is given if not found (2019). In 2017 API, the SystemAssigned was used if available
- Secret header changed from "SECRET" (2017) to "X-IDENTITY-HEADER" (2019). If invalid header, response changed from Unauthorized (2017) to Forbidden (2019).
- Expires_on in the response payload is now unix time seconds (2019) instead of local datetime (2017)

### LINKED ISSUE #27673: Make "2019-08-01" API version of the App Service Managed Identity endpoint available
### Library name

Azure.Identity

Make "2019-08-01" API version of the App Service Managed Identity endpoint appear before the older API version in the Managed Identity sources tried by the Managed Identity Credential.

christothes:
closes #27673
closes #26578

### DIFF
--- a/sdk/identity/Azure.Identity/src/ManagedIdentityClient.cs
+++ b/sdk/identity/Azure.Identity/src/ManagedIdentityClient.cs
@@ -55,11 +55,11 @@ namespace Azure.Identity
         private static ManagedIdentitySource SelectManagedIdentitySource(ManagedIdentityClientOptions options)
         {
             return
+                ServiceFabricManagedIdentitySource.TryCreate(options) ??
+                AppServiceV2019ManagedIdentitySource.TryCreate(options) ??
                 AppServiceV2017ManagedIdentitySource.TryCreate(options) ??
                 CloudShellManagedIdentitySource.TryCreate(options) ??
                 AzureArcManagedIdentitySource.TryCreate(options) ??
-                ServiceFabricManagedIdentitySource.TryCreate(options) ??
-                AppServiceV2019ManagedIdentitySource.TryCreate(options) ??
                 TokenExchangeManagedIdentitySource.TryCreate(options) ??
                 new ImdsManagedIdentitySource(options);
         }
--- a/sdk/identity/Azure.Identity/src/ManagedIdentityClient.cs
+++ b/sdk/identity/Azure.Identity/src/ManagedIdentityClient.cs
@@ -10,20 +10,24 @@ namespace Azure.Identity
 {
     internal class ManagedIdentityClient
     {
-        internal const string MsiUnavailableError = "ManagedIdentityCredential authentication unavailable. No Managed Identity endpoint found.";
+        internal const string MsiUnavailableError =
+            "ManagedIdentityCredential authentication unavailable. No Managed Identity endpoint found.";
 
         private Lazy<ManagedIdentitySource> _identitySource;
 
         protected ManagedIdentityClient()
-        { }
+        {
+        }
 
         public ManagedIdentityClient(CredentialPipeline pipeline, string clientId = null)
-            : this(new ManagedIdentityClientOptions { Pipeline = pipeline, ClientId = clientId })
-        { }
+            : this(new ManagedIdentityClientOptions {Pipeline = pipeline, ClientId = clientId})
+        {
+        }
 
         public ManagedIdentityClient(CredentialPipeline pipeline, ResourceIdentifier resourceId)
-            : this(new ManagedIdentityClientOptions { Pipeline = pipeline, ResourceIdentifier = resourceId })
-        { }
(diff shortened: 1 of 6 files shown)


sdk/identity/Azure.Identity/src/ManagedIdentityClient.cs
```diff
@@ -48,7 +48,8 @@ public virtual async ValueTask<AccessToken> AuthenticateAsync(bool async, TokenR
 
         private static ManagedIdentitySource SelectManagedIdentitySource(ManagedIdentityClientOptions options)
         {
-            return AppServiceV2017ManagedIdentitySource.TryCreate(options) ??
+            return AppServiceV2019ManagedIdentitySource.TryCreate(options) ??
+                AppServiceV2017ManagedIdentitySource.TryCreate(options) ??
```
schaabs, line 52: There is some nuance here. I think with adding support for the 2019 protocol we need to consider both `AppServiceV2019ManagedIdentitySource` and `AppServiceV2017ManagedIdentitySource` after `ServiceFabricManagedIdentitySource` otherwise we will break Service Fabric  managed identity support. This is because `ServiceFabricManagedIdentitySource` should be used if the environment variables `IDENTITY_ENDPOINT`, `IDENTITY_HEADER` _AND_ `IDENTITY_SERVER_THUMBPRINT` are set. If we consider `AppServiceV2019ManagedIdentitySource` before it, we will incorrectly identify the Service Fabric host as the App Service host. 

Also `AppServiceV2017ManagedIdentitySource` must be considered after `AppServiceV2019ManagedIdentitySource` (as you currently are doing) because App Service hosts which support the 2019 protocol will have both pairs of `IDENTITY_ENDPOINT`, `IDENTITY_HEADER` _AND_ `MSI_ENDPOINT`, `MSI_SECRET` set, so in this case we want to prefer the newer protocol.

We should add comment to the rational behind there ordering here, and make sure we have tests covering these cases.

schaabs (changes requested):

sdk/identity/Azure.Identity/src/ManagedIdentityClient.cs
christothes: In this case, we are handling it properly in the 2019 provider, but I think the ordering fix is better because it doesn't require the V2019 source from having to check the `IDENTITY_SERVER_THUMBPRINT` environment variable.

sdk/identity/Azure.Identity/src/ManagedIdentityClient.cs
```diff
@@ -41,20 +46,22 @@ public ManagedIdentityClient(ManagedIdentityClientOptions options)
...
+            CancellationToken cancellationToken)
         {
             return await _identitySource.Value.AuthenticateAsync(async, context, cancellationToken).ConfigureAwait(false);
         }
 
         private static ManagedIdentitySource SelectManagedIdentitySource(ManagedIdentityClientOptions options)
         {
-            return AppServiceV2019ManagedIdentitySource.TryCreate(options) ??
+            return
                 AppServiceV2017ManagedIdentitySource.TryCreate(options) ??
```
schaabs, line 58: I believe `AppServiceV2017ManagedIdentitySource` and `CloudShellManagedIdentitySource` should come after `AppServiceV2019ManagedIdentitySource` since App Service hosts which support `2019-08-01` will likely still set `MSI_ENDPOINT` and `MSI_SECRET` for backwards compatibility in this case we want to use the latest protocol. If we don't have a test which set's all four environment variables which app services uses `IDENTITY_ENDPOINT`, `IDENTITY_HEADER`, `MSI_ENDPOINT` and `MSI_SECRET`, and ensure we use the `AppServiceV2019ManagedIdentitySource` in this case.

schaabs (changes requested):

schaabs (approved):

PR #27719, opened 2022-03-23
Prepare Key Vault SDK release for 7.3

### LINKED ISSUE #24965: [KeyVault] - Prepare for GA release
In order to prepare for GA release, ensure the following have been considered for all 4 packages:

- [ ] Regenerate code from the stable 7.3 swagger when available for all 4 libraries
- [ ] API view is approved (all 4 likely will have changes - either due to new RBAC roles or just service version change)
- [ ] Version constants are updated to remove the "-beta.x" suffix
- [ ] Documentation pass: ensure it meets [this bar](https://dev.azure.com/azure-sdk/internal/_wiki/wikis/internal.wiki/122/Documentation-Release-Checklist) and that generated docs look as expected
- [ ] Ensure release artifacts can be installed cleanly
- [ ] Ensure upgrade path succeeds without breaking changes (use existing GA package, run a few well-known scenarios, then upgrade and rerun)
- [ ] Add a link to release policy grammar, see https://github.com/Azure/azure-sdk-for-js/pull/20941/files as an example
- [ ] Ensure any language-specific necessary action items have been added and addressed

heaths:
Resolves #24965


heaths:
Note to @maorleger @christothes @schaabs: if you ignore the session records, it's not that bad.

christothes (approved):

maorleger (approved):

heaths:
The build breaks were primarily from missing declarations to not compare bodies. I also forgot the one-off SecretClientLiveTests manual date update and unification. Maybe I should just get rid of that. There's a comment in there in case I get hit by the lottery, but still not the most intuitive to update.

PR #27731, opened 2022-03-24
[ResourceManager] Rename duplicates in resource manager

Yao725:


sdk/resourcemanager/Azure.ResourceManager/src/ManagementGroup/Generated/ManagementGroupCollection.cs
```diff
@@ -297,7 +297,7 @@ Page<ManagementGroupResource> NextPageFunc(string nextLink, int? pageSizeHint)
         /// <param name="checkNameAvailabilityRequest"> Management group name availability check parameters. </param>
         /// <param name="cancellationToken"> The cancellation token to use. </param>
         /// <exception cref="ArgumentNullException"> <paramref name="checkNameAvailabilityRequest"/> is null. </exception>
-        public virtual async Task<Response<CheckNameAvailabilityResult>> CheckNameAvailabilityAsync(CheckNameAvailabilityOptions checkNameAvailabilityRequest, CancellationToken cancellationToken = default)
+        public virtual async Task<Response<ManagementGroupNameAvailabilityResult>> CheckNameAvailabilityAsync(CheckNameAvailabilityOptions checkNameAvailabilityRequest, CancellationToken cancellationToken = default)
```
Yao725, line 300: Two things that needs to confirm here:

- We might need to change the method name to `CheckManagementGroupNameAvailability`.
- Should we consider modifying the name of `CheckNameAvailabilityOptions` to `ManagementGroupNameAvailabilityOptions`?

archerzz: If `CheckNameAvailabilityOptions` is public, then we should do it. Otherwise, it's optional, IMHO.

archerzz (approved):

Issue #27619, opened 2022-03-17
Make sure snippets can be compiled


heaths:
The following projects have snippets that do not compile correctly. Owners should make sure their snippets build correctly to make sure that customers copying snippets (and replacing any placeholders necessary) are successful with our sample snippets.

1. Build your service directory. I recommend enabling binlogs (`/bl`) and using the [MSBuild Log Viewer](https://msbuildlog.com/) to view the file (default is "msbuild.binlog" in the current working directory, which is ignored by git already):
   ```bash
   dotnet build eng/service.proj -warnaserror /t:rebuild /p:DebugType=none /p:SDKType=all /p:ServiceDirectory={ServiceDirectory} /p:IncludePerf=false /p:IncludeStress=false /p:PublicSign=false /p:Configuration=Release /p:EnableSourceLink=false /p:BuildSnippets=true /bl
   ```
2. After you can build cleanly, in your `sdk/{ServiceDirectory}/ci.yml` add the `BuildSnippets=true` parameter:
   ```yaml
   parameters:
     ServiceDirectory: "{ServiceDirectory}"
     BuildSnippets: true
   ```

For more information and any tips about how I solved some of these problems for Key Vault, see PR #27551.

* [ ] samples/CloudClipboard
* [ ] sdk/appconfiguration/Azure.Data.AppConfiguration
* [x] sdk/cognitivelanguage/Azure.AI.Language.QuestionAnswering
* [ ] sdk/compute/Azure.ResourceManager.Compute
* [ ] sdk/confidentialledger/Azure.Security.ConfidentialLedger
* [x] sdk/containerregistry/Azure.Containers.ContainerRegistry
* [ ] sdk/core/Azure.Core
* [x] sdk/eventhub/Azure.EventHubs
* [x] sdk/eventhub/Azure.Messaging.EventHubs.Processor
* [x] sdk/eventhub/Azure.ResourceManager.EventHubs
* [x] sdk/formrecognizer/Azure.AI.FormRecognizer
* [ ] sdk/identity/Azure.Identity
* [ ] sdk/keyvault/Azure.ResourceManager.KeyVault
* [x] sdk/metricsadvisor/Azure.AI.MetricsAdvisor
* [ ] sdk/network/Azure.ResourceManager.Network
* [ ] sdk/personalizer/Azure.AI.Personalizer
* [ ] sdk/resourcemanager/Azure.ResourceManager
* [x] sdk/search/Azure.Search.Documents
[...]

PR #27738, opened 2022-03-24
[Event Hubs] Enable snippet compilation

jsquire:
# Summary 

The focus of these changes is to opt-into building snippets during CI runs and fixing the snippet build errors across all track two projects in the `eventhub` service directory.

# References and Related

- [Make sure snippets can be compiled (#27619)](https://github.com/Azure/azure-sdk-for-net/issues/27619)

### DIFF
--- a/sdk/eventhub/Azure.Messaging.EventHubs/tests/Snippets/Sample04_PublishingEventsLiveTests.cs
+++ b/sdk/eventhub/Azure.Messaging.EventHubs/tests/Snippets/Sample04_PublishingEventsLiveTests.cs
@@ -589,8 +589,6 @@ namespace Azure.Messaging.EventHubs.Tests.Snippets
 #else
             var connectionString = EventHubsTestEnvironment.Instance.EventHubsConnectionString;
             var eventHubName = scope.EventHubName;
-            var sentEventCount = 0;
-            var batchEventCount = 0;
 #endif
 
             var producer = new EventHubProducerClient(connectionString, eventHubName);
@@ -606,15 +604,9 @@ namespace Azure.Messaging.EventHubs.Tests.Snippets
 
                 batches = await BuildBatchesAsync(eventsToSend, producer);
 
-#if !SNIPPET
-                batchEventCount = batches.Sum(batch => batch.Count);
-#endif
                 foreach (var batch in batches)
                 {
                     await producer.SendAsync(batch);
-#if !SNIPPET
-                    sentEventCount += batch.Count;
-#endif
                 }
             }
             finally
@@ -628,10 +620,6 @@ namespace Azure.Messaging.EventHubs.Tests.Snippets
             }
 
             #endregion
-
-#if !SNIPPET
-            Assert.That(batchEventCount, Is.EqualTo(sentEventCount));
-#endif
         }
 
         /// <summary>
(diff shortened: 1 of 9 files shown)


sdk/eventhub/Azure.ResourceManager.EventHubs/CHANGELOG.md
```diff
@@ -101,10 +101,10 @@ var createEventHubResponse = this.EventHubManagementClient.EventHubs.CreateOrUpd
 
 After upgrade:
 ```C# Snippet:ChangeLog_Sample
-using Azure.Identity;
-using Azure.ResourceManager.Resources;
-using Azure.ResourceManager.EventHubs.Models;
-using Azure.Core;
+// using Azure.Identity;
```
JoshLove-msft, line 104: Why are these commented out?

JoshLove-msft: I think I understand - it's because we can't prevent building this using the ifdef SNIPPET anymore. An alternative would be to create a separate region around where the usings are declared in the cs file.

jsquire: Because they're not compilable in the location they were included - neither for the tests nor for customers doing a copy/paste on the snippet.   The most common approach that I've seen for including "using" hints in samples across MS Docs and elsewhere is to illustrate via comment, as I've done here.  

The other possibility that I see would be to create a separate snippet that demonstrates only the using statements, but that looked strange when I prototyped it, and I don't think the lack of locality was helpful in keeping the association of "using" context with the snippet; the comment does that better.

sdk/eventhub/Azure.Messaging.EventHubs/tests/Snippets/Sample04_PublishingEventsLiveTests.cs
```diff
@@ -629,7 +629,9 @@ public async Task MultipleBatches()
 
             #endregion
 
+#if !SNIPPET
```
heaths, line 632: Was this necessary? Seems the region ends right above it so this line shouldn't even be in the snippet to begin with.

sdk/eventhub/Azure.ResourceManager.EventHubs/CHANGELOG.md
```diff
@@ -47,8 +47,8 @@ Example: Create an Event Hub:
 
 Before upgrade:
 ```csharp
-using Microsoft.Azure.Management.EventHub;
-using Microsoft.Azure.Management.EventHub.Models;
+// using Microsoft.Azure.Management.EventHub;
+// using Microsoft.Azure.Management.EventHub.Models;
```
heaths, line 51: I recommend having a separate snippet block for imports to avoid having to do this in the source where it's not possible outside of comments. I made a similar comment on the Service Hub PR.

heaths (approved):
Approving to unblock, but I do recommend a separate namespace import block as with the Service Hub PR.

sdk/eventhub/Azure.ResourceManager.EventHubs/CHANGELOG.md
jsquire: I'm not a fan of the separate block, personally.  I think it harms locality and context.   That said, if we've agreed on that as a common format, I'll conform.

sdk/eventhub/Azure.Messaging.EventHubs/tests/Snippets/Sample04_PublishingEventsLiveTests.cs
jsquire: It was necessary because the variables it was using were only declared for non-snippet scenarios.   The bigger question is why were we doing test assertions and tracking in the sample snippet... since I can't answer that and we have tests that cover this scenario, I removed the non-snippet logic.

PR #27769, opened 2022-03-25
Add DPG codeowners to repo

annelo-msft:
We'll need to set up ownership for various areas of DPG.


.github/CODEOWNERS
```diff
@@ -703,3 +703,11 @@ sdk/trafficmanager/Microsoft.Azure.Management.TrafficManager/            @tmsupp
 /eng/pipelines/aggregate-reports.yml            @AlexGhiondea @maririos
 /sdk/eventhub/tests.data.yml  @serkantkaraca @sjkwak
 /sdk/servicebus/tests.data.yml @shankarsama @DorothySun216 
+
+# ######## DPG ########
+
+# Onboarding Documentation and Quickstarts
+/doc/Data\%20Plane\%20Code\%20Generation           @chunyu3 @pshao25 @fengzhou-msft @lirenhe @annelo-msft
```
annelo-msft, line 710: @jsquire, do you know how to escape %20's in the paths?  Or will we need to take the spaces out of this path?

jsquire: Better to have no spaces, in my opinion, as it is difficult to work with them on non-Windows systems.   I can't think of any other place in the repo that we have paths with spaces.  My vote would be to go with either `data-place-code-generation` or `DataPlaneCodeGeneration`.

That said, since CODEOWNERS files .gitignore file patterns, you can encode them as if they were *nix paths and it should work.

jsquire, line 710:
```suggestion
/doc/Data\ Plane\ Code\ Generation           @chunyu3 @pshao25 @fengzhou-msft @lirenhe @annelo-msft
```

If you decide against removing the spaces, I believe this is the format that you'd need.

jsquire (approved):

PR #27775, opened 2022-03-25
Block any project reference to Azure.Core that is approved by core owners

weshaggard:
This is a proof of concept for how we could make sure that anyone taking a project reference to Azure.Core would need to get review from the owners and get added to a list of approved projects.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -19,6 +19,7 @@
 
 # PRLabel: %Azure.Core
 /sdk/core/                                      @JoshLove-msft @christothes @annelo-msft @KrzysztofCwalina
+/eng/AllowedAzureCorePackageReferences.targets  @JoshLove-msft @christothes @annelo-msft @KrzysztofCwalina
 
 # PRLabel: %Azure.Identity
 /sdk/identity/                                  @schaabs @christothes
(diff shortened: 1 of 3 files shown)


jsquire:
Can you help me understand the problem that we're trying to solve with this?

weshaggard:
The thing we are trying to do is ensure that anyone that depends on a unrelease ProjectReference to Azure.Core will have to get reviewed and approved by the owners of Azure.Core. This is just one potential option I put together for consideration. I think the main thing we are trying to do is make sure folks realize that taking a project dependency creates a release ordering dependency and some folks the folks that have to release now because of that should be informed somehow.

JoshLove-msft:
How would we ensure the list of allowed libraries doesn't included extra libraries that are no longer needing the project reference?

annelo-msft:
@AlexGhiondea FYI - we may not need this if we move to a new process @AlexGhiondea has proposed where we no longer take project references to Azure.Core in the repo.  Or @AlexGhiondea, was this scoped only to DPG?

christothes:
Thanks for exploring this. My initial thought is that making it easier to track dependencies on core is a goal, but I'm not sure that making it more difficult to take a Core project reference is desirable. A client would only add a project reference to Azure.Core after or in conjunction with a PR that already must be approved by the normal PR review process.

weshaggard:
Closing this PR given there doesn't seem to be much interest. It was just a proof of concept so if you guys want to do something similar you can.

PR #27789, opened 2022-03-28
Remove useless 'lenient-model-deduplication'

ArthurMa1978:
Part of https://github.com/azure/autorest.csharp/issues/1818

### DIFF
--- a/sdk/resourcemanager/Azure.ResourceManager/eng/autorest-resources.md
+++ b/sdk/resourcemanager/Azure.ResourceManager/eng/autorest-resources.md
@@ -23,8 +23,6 @@ namespace: Azure.ResourceManager.Resources
 model-namespace: false
 public-clients: false
 head-as-boolean: false
-modelerfour:
-    lenient-model-deduplication: true
 skip-csproj: true
 payload-flattening-threshold: 2
 operation-group-to-resource-type:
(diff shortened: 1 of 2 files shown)


sdk/resourcemanager/Azure.ResourceManager/eng/autorest-resources.md
```diff
@@ -23,8 +23,6 @@ namespace: Azure.ResourceManager.Resources
 model-namespace: false
 public-clients: false
 head-as-boolean: false
-modelerfour:
-    lenient-model-deduplication: true
```
ArcturusZhang, line 27: As far as I know, this file does not have any effect, because the resourcemanager is generated from this configuration file: https://github.com/Azure/azure-sdk-for-net/blob/25c9dd2fb0f48628f531e16612b8cc0b0d781179/sdk/resourcemanager/Azure.ResourceManager/src/autorest.md
I am not sure why @fengzhou-msft still kept these files, but I suggest we remove this entire directory `sdk/resourcemanager/Azure.ResourceManager/eng`

ArthurMa1978: Leave it as is now

Yao725 (approved):

PR #27829, opened 2022-03-29
Update doc comments on GetUserDelegationKey

pharring:
Fix the description of the return parameter for GetUserDelegationKey


pharring:
@seanmcc-msft would you mind taking a look at this?

PR #27836, opened 2022-03-30
Fix system model namespace

JoshLove-msft:
Hiding models that were released in an erroneous namespace. New versions (with new names) are created in the correct namespace and new properties (with new names) have been added.


sdk/eventgrid/Azure.Messaging.EventGrid/src/Customization/EventGridModelFactory.cs
```diff
@@ -437,7 +437,25 @@ public static MediaLiveEventChannelArchiveHeartbeatEventData MediaLiveEventChann
...
+            return AcsRecordingFileStatusUpdatedEventData(recordingStorageInfo: recordingStorageInfo, recordingStartTime: recordingStartTime, recordingDurationMs: recordingDurationMs, sessionEndReason: sessionEndReason, channelType: null);
+        }
+
+        /// <summary> Initializes a new instance of AcsRecordingFileStatusUpdatedEventData. </summary>
+        /// <param name="recordingStorageInfo"> The details of recording storage information. </param>
+        /// <param name="recordingStartTime"> The time at which the recording started. </param>
+        /// <param name="recordingDurationMs"> The recording duration in milliseconds. </param>
+        /// <param name="recordingContentType"> The recording content type- AudioVideo, or Audio. </param>
+        /// <param name="recordingChannelType"> The recording  channel type - Mixed, Unmixed. </param>
+        /// <param name="recordingFormatType"> The recording format type - Mp4, Mp3, Wav. </param>
```
jsquire, line 449:
```suggestion
        /// <param name="recordingFormatType"> The recording format type.  For example, Mp4, Mp3, Wav. </param>
```

We may want to be explicit about these being examples rather than an always-up-to-date exhaustive list.

jsquire (approved):

PR #27858, opened 2022-03-30
Update azure.core to have reference type on ResponseError

m-nash:
Implementing the issue discussed here https://github.com/Azure/azure-sdk/issues/4091#issuecomment-1081101442


sdk/core/Azure.Core/src/Internal/TypeReferenceTypeAttribute.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+
+namespace Azure.Core
+{
+    /// <summary>
+    /// An attribute class indicating a reference type which can replace a type in target RPs.
```
annelo-msft, line 9: Curious: RPs is a management plane-specific term -- is there a more general use case than this, or would it only ever be used for management plane?

sdk/core/Azure.Core/src/Internal/InitializationConstructorAttribute.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+
+namespace Azure.Core
+{
+    /// <summary>
+    /// An attribute class indicating the constructor to use for initialization.
```
christothes, line 9:
```suggestion
    /// An attribute class indicating to AutoRest which constructor to use for initialization.
```

sdk/core/Azure.Core/src/Internal/SerializationConstructorAttribute.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+
+namespace Azure.Core
+{
+    /// <summary>
+    /// An attribute class indicating the constructor to use for serialization.
```
christothes, line 9:
```suggestion
    /// An attribute class indicating to AutoRest which constructor to use for serialization.
```

sdk/core/Azure.Core/src/Internal/TypeReferenceTypeAttribute.cs
christothes: 👍 We should generalize this desscription.

sdk/core/Azure.Core/src/Internal/InitializationConstructorAttribute.cs
```diff
@@ -0,0 +1,15 @@
...
+
+using System;
+
+namespace Azure.Core
+{
+    /// <summary>
+    /// An attribute class indicating to AutoRest which constructor to use for initialization.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Constructor)]
+    internal class InitializationConstructorAttribute : Attribute
```
JoshLove-msft, line 12: Don't these types already exist in the repo for the mgmt plane usage? Should we just move those into core and pull them into mgmt projects as needed?

m-nash: We thought about making these internal shared classes, but wanted to move that out of scope for this change since we don't necessarily want to add to the existing circular internal shared reference issue.

m-nash: I could do the first version of breaking this dependency if we feel its needed for this, but IMHO this can be done after GA since its all internal code

JoshLove-msft: Can you explain what you mean about the existing circular reference issue?

m-nash: instead of an SDK pulling shared source from azure.core (its dependency), Azure.Core gives its files to Autorest, the SDK references Azure.Core but pulls all of the shared Azure.Core files from Autorest.

JoshLove-msft (approved):

PR #27966, opened 2022-04-04
Warn on unsupported managed identity configuration

### LINKED ISSUE #26714: Warn on unsupported managed identity configuration
### Library name

Azure.Identity

Cloud Shell's token endpoint ignores unexpected parameters and always authenticates the signed in user. This means that when ManagedIdentityCredential requests a token for a user assigned identity, Cloud Shell returns a token for the signed in user instead. The credential should therefore throw an exception when asked to get a token for a user assigned identity in Cloud Shell.

christothes:
closes #26714


sdk/identity/Azure.Identity/src/CloudShellManagedIdentitySource.cs
```diff
@@ -36,38 +35,31 @@ public static ManagedIdentitySource TryCreate(ManagedIdentityClientOptions optio
...
         }
 
-        private CloudShellManagedIdentitySource(CredentialPipeline pipeline, Uri endpoint, string clientId) : base(pipeline)
+        private CloudShellManagedIdentitySource(Uri endpoint, ManagedIdentityClientOptions options) : base(options.Pipeline)
         {
             _endpoint = endpoint;
-            _clientId = clientId;
+            if (!string.IsNullOrEmpty(options.ClientId) || null == options.ResourceIdentifier)
+            {
+                AzureIdentityEventSource.Singleton.UserAssignedManagedIdentityNotSupported("CloudShell");
```
chlowell, line 46: Should this be the product name?
```suggestion
                AzureIdentityEventSource.Singleton.UserAssignedManagedIdentityNotSupported("Cloud Shell");
```

sdk/identity/Azure.Identity/src/AzureArcManagedIdentitySource.cs
```diff
@@ -51,6 +51,7 @@ protected override Request CreateRequest(string[] scopes)
             // arc MI endpoint doesn't support user assigned identities so if client id was specified throw AuthenticationFailedException
             if (!string.IsNullOrEmpty(_clientId))
```
chlowell, line 52: Is it possible to pass a ResourceID to this source?

chlowell (approved):

sdk/identity/Azure.Identity/src/AzureArcManagedIdentitySource.cs
christothes: In theory any of them are subject to the option if it was specified. We happen not to explicitly do anything with it here, but we should warn if it was sent in the options either way. I'll add the additional check.

sdk/identity/Azure.Identity/src/CloudShellManagedIdentitySource.cs
```diff
@@ -36,38 +35,31 @@ public static ManagedIdentitySource TryCreate(ManagedIdentityClientOptions optio
...
-            return new CloudShellManagedIdentitySource(options.Pipeline, endpointUri, options.ClientId);
+            return new CloudShellManagedIdentitySource(endpointUri, options);
         }
 
-        private CloudShellManagedIdentitySource(CredentialPipeline pipeline, Uri endpoint, string clientId) : base(pipeline)
+        private CloudShellManagedIdentitySource(Uri endpoint, ManagedIdentityClientOptions options) : base(options.Pipeline)
         {
             _endpoint = endpoint;
-            _clientId = clientId;
+            if (!string.IsNullOrEmpty(options.ClientId) || null == options.ResourceIdentifier)
```
billwert, line 44: Should this be `null != options.ResourceIdentifier` like the others?

christothes: Yes, thanks for catching this!

PR #27982, opened 2022-04-05
Fix namespace of internal shared type

JoshLove-msft:


annelo-msft (commented):
For context - is this needed because the namespace change was missed when moving these in from Experimental?

christothes (commented):
I think we need to update all clients that reference this type also.

JoshLove-msft:
> For context - is this needed because the namespace change was missed when moving these in from Experimental?

No, I just noticed that the namespace for this shared type was Azure.Monitor.Query which seemed wrong given that the type was used in other libraries like Tables.

JoshLove-msft:
> I think we need to update all clients that reference this type also.

Yes, I think I updated them all. The latest CI failures were unrelated.

christothes (approved):

JoshLove-msft:
> > No, I just noticed that the namespace for this shared type was Azure.Monitor.Query which seemed wrong given that the type was used in other libraries like Tables.
> 
> Have the types GA'ed? Is it a breaking change?

This is an internal shared type so it doesn't impact the API surface anywhere.

annelo-msft:
> No, I just noticed that the namespace for this shared type was Azure.Monitor.Query which seemed wrong given that the type was used in other libraries like Tables.

Cool, thanks!

PR #27987, opened 2022-04-05
Prepare release for Azure.ResourceManager.Grafana 1.0.0-beta.1

ArthurMa1978:


Yao725 (commented):
Considering that the `Azure.ResourceManager 1.0.0-beta.9` depends on  `Azure.Core 1.23.0` and we have introduced the change that uses the `ResponseError` from `Azure.Core 1.24.0`,  I think we have two options here:
1. Release Dashboard from this [tag](https://github.com/Azure/azure-sdk-for-net/tree/Azure.ResourceManager_1.0.0-beta.9)
2. Hold on this release until we release the GA version of `Azure.ResourceManager`

m-nash (approved):

PR #28016, opened 2022-04-06
DPG best-practice and automation template

chunyu3:

As part of the DPG effort, we are providing a DPG template as a tool and guidance to our service team partners:
 1. To enable them to create a new SDK in the repo for their service via this template, that builds and passes CIs in 10 minutes
That has all the information they need to follow Azure SDK best practices to create tests, samples, and other customer-facing documents in a matter of days
2.	This template sdk/template/Azure.Template is DPG + grow-up story, and it can pass the CI mock test to validate the changes of eng scripts and pipeline.


annelo-msft (commented):
Please add a link to the issue in the PR description.

sdk/template/Azure.ServiceTemplate.Template/tests/TemplateClientTest.cs
```diff
@@ -18,24 +18,7 @@ public TemplateClientTest(bool isAsync) : base(isAsync)
...
-        //     return client;
-        // }
-        // Add Test here as following.
-        // [RecordedTest]
-        // public async Task TestOperation()
-        // {
-        // }
+        /* please refer to https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/template/Azure.Template/tests/TemplateClientLiveTests.cs to write tests. */
 
         [RecordedTest]
```
annelo-msft, line 23: It looks like we are referencing Synapse examples below -- this service has not GA'ed, so these are probably not great examples to reference.  Would it be possible to reference files in the template as examples here?

sdk/template/Azure.Template/CHANGELOG.md
```diff
@@ -1,106 +1,23 @@
...
-- Testing Changelog added section
-- Testing Links [Project Site](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/template/Azure.Template/README.md)
-- Test `Highlighted Code`
-
-## 1.0.2-preview.4 (2019-12-17)
-- Version Bump
-
-## 1.0.2-preview.3 (2019-12-17)
-- Started changelog to capture release notes.
+## 1.0.0-beta.1 (Unreleased)
```
annelo-msft, line 3: Would it make sense to preserve the history of the template in the CHANGELOG?  I'm actually not sure which is better, to show the history or provide a template for it.

sdk/template/Azure.Template/README.md
```diff
@@ -1,11 +1,11 @@
 # README.md template
 
-Use the guidelines in each section of this template to ensure consistency and readability of your README. The README resides in your package's GitHub repository at the root of its directory within the repo. It's also used as the package distribution page (NuGet, PyPi, npm, etc.) and as a Quickstart on docs.microsoft.com. See [Azure.Template/README.md](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/template/Azure.Template/README.md) for an example following this template.
+Use the guidelines in each section of this template to ensure consistency and readability of your README. The README resides in your package's GitHub repository at the root of its directory within the repo. It's also used as the package distribution page (NuGet) and as a Quickstart on docs.microsoft.com. See [Azure.Template/README.md](https://github.com/Azure/azure-sdk-for-net/blob/bb0fbccfc33dd27d1ec6f0870022824d47181e61/sdk/template-dpg/Azure.Template/README.md) for an example following this template.
```
annelo-msft, line 3: If we pin the example to a specific version, will we be able to keep it updated?

sdk/template/Azure.Template/samples/Sample1_CreateResource.md
```diff
@@ -0,0 +1,46 @@
...
+
+Use the guidelines in each section of this template to write samples, for an example following this template.
+
+This sample shows how to detect all the anomalies in the entire time series.
+
+To get started, make sure you have satisfied all the prerequisites and got all the resources required by [README][README].
+
+**write the sample code snippets as following example**
+
+## Detect anomalies of the entire series
```
annelo-msft, line 11: Is it confusing that we're using examples from different service libraries in the template?  Would it make sense to stick to one service to illustrate that all the parts of the template work together for a single service SDK?

sdk/template/Azure.Template/src/Azure.Template.csproj
```diff
@@ -1,22 +1,23 @@
...
   <PropertyGroup>
-    <Description>This is a template project to demonstrate how to create a package that uses code generation as well as use for testing our build and release pipelines</Description>
-    <AssemblyTitle>Azure SDK Template</AssemblyTitle>
-    <Version>1.0.3-beta.19</Version>
+    <Description>This is the Template client library for developing .NET applications with rich experience.</Description>
+    <AssemblyTitle>Azure SDK Code Generation Template for Azure Data Plane</AssemblyTitle>
+    <Version>1.0.0-beta.1</Version>
     <PackageTags>Azure Template</PackageTags>
     <TargetFrameworks>$(RequiredTargetFrameworks)</TargetFrameworks>
+    <NoWarn>$(NoWarn);419;AZC0012</NoWarn>
```
annelo-msft, line 8: We should not use `<NoWarn>` in a template -- we don't want library authors suppressing errors.

sdk/template/Azure.Template/src/Azure.Template.csproj
```diff
@@ -1,22 +1,23 @@
...
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
```
annelo-msft, line 9: This should be removed as well.

sdk/template/Azure.Template/src/Azure.Template.csproj
```diff
@@ -1,22 +1,23 @@
...
+    <DefineConstants>$(DefineConstants);EXPERIMENTAL</DefineConstants>
```
annelo-msft, line 10: Do we still need this?

sdk/template/Azure.Template/src/Azure.Template.csproj
```diff
@@ -1,22 +1,23 @@
...
     <IncludeOperationsSharedSource>true</IncludeOperationsSharedSource>
   </PropertyGroup>
 
-  <ItemGroup>
-    <PackageReference Include="Azure.Core" />
-    <PackageReference Include="System.Text.Json" />
```
annelo-msft, line 13: Why remove the reference to System.Text.Json?  I expect most libraries will need it, and it will help clarify that this is the Azure SDK supported library for Json.

sdk/template/Azure.Template/src/Generated/TemplateClient.cs
```diff
@@ -0,0 +1,144 @@
...
+using Azure;
+using Azure.Core;
+using Azure.Core.Pipeline;
+
+namespace Azure.Template
+{
+    /// <summary> The Template service client. </summary>
+    public partial class TemplateClient
+    {
+        private static readonly string[] AuthorizationScopes = new string[] { "https://api.adu.microsoft.com/.default" };
```
annelo-msft, line 19: Why this auth scope?  Should we provide one that is more clearly an example?

sdk/template/Azure.Template/src/autorest.md
```diff
@@ -1,8 +1,17 @@
...
 
+### AutoRest Configuration
+> see https://aka.ms/autorest
+
 ``` yaml
 input-file:
-    -  $(this-folder)/swagger/mini-secrets.json
+- $(this-folder)/swagger/mini-secrets.json
+namespace: Azure.Template
+public-clients: true
```
annelo-msft, line 12: We should not use `public-clients: true` in a template -- this is not a recommended approach for data plane.

sdk/template/Azure.Template/src/Generated/TemplateClient.cs
```diff
@@ -0,0 +1,144 @@
...
+        private readonly TokenCredential _tokenCredential;
+        private readonly HttpPipeline _pipeline;
+        private readonly string _vaultBaseUrl;
```
annelo-msft, line 22: This should be a `Uri` type.

sdk/template/Azure.Template/src/Generated/TemplateClient.cs
```diff
@@ -0,0 +1,144 @@
...
+        protected TemplateClient()
+        {
+        }
+
+        /// <summary> Initializes a new instance of TemplateClient. </summary>
+        /// <param name="vaultBaseUrl"> The vault name, for example https://myvault.vault.azure.net. </param>
+        /// <param name="credential"> A credential used to authenticate to an Azure Service. </param>
+        /// <param name="options"> The options for configuring the client. </param>
+        /// <exception cref="ArgumentNullException"> <paramref name="vaultBaseUrl"/> or <paramref name="credential"/> is null. </exception>
+        public TemplateClient(string vaultBaseUrl, TokenCredential credential, TemplateClientOptions options = null)
```
annelo-msft, line 41: This constructor doesn't follow the Azure SDK guidelines -- it should take a Uri for the endpoint, an have no default parameters in the constructor.  We should regenerate once https://github.com/azure/autorest.csharp/issues/2105 is resolved.

sdk/template/Azure.Template/src/GlobalSuppressions.cs
```diff
@@ -0,0 +1,7 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Diagnostics.CodeAnalysis;
+
+[assembly: SuppressMessage("Usage", "AZC0002:DO ensure all service methods, both asynchronous and synchronous, take an optional CancellationToken parameter called cancellationToken.", Justification = "CancellationToken can be passed through RequestOptions")]
```
annelo-msft, line 6: We should not suppress this warning.

sdk/template/Azure.Template/src/GlobalSuppressions.cs
```diff
@@ -0,0 +1,7 @@
...
+[assembly: SuppressMessage("Usage", "AZC0016:Invalid ServiceVersion member name.", Justification = "Generated code: https://github.com/Azure/autorest.csharp/issues/1524")]
```
annelo-msft, line 7: Do we need this? The version enum doesn't look like it has `preview` in the name.

sdk/template/Azure.Template/src/TemplateClient.cs
```diff
@@ -0,0 +1,71 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+// <auto-generated/>
```
annelo-msft, line 4: Was this file auto-generated?

sdk/template/Azure.Template/src/TemplateClient.cs
```diff
@@ -0,0 +1,71 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Azure;
+using Azure.Core;
+using Azure.Core.Pipeline;
+using Azure.Template.Models;
+
+namespace Azure.Template
+{
+    /// <summary> The Template service client. </summary>
```
annelo-msft, line 18: It might be worth calling out here or elsewhere in the file that this partial class implements the Convenience APIs for the DPG client.  You could even link back to where this is discussed in the getting-started documentation.  I think this would help service-team partners make this connection, and make it easier to understand the overall Azure SDK client development story.

sdk/template/ci.yml
```diff
@@ -24,21 +20,12 @@ pr:
...
 extends:
-  template: ../../eng/pipelines/templates/stages/archetype-sdk-client.yml
+  template: /eng/pipelines/templates/stages/archetype-sdk-client.yml
   parameters:
     ServiceDirectory: template
     ArtifactName: packages
     Artifacts:
     - name: Azure.Template
-      safeName: AzureTemplate
+      safeName: AzureIoTTemplate
```
annelo-msft, line 31: Why IoT?

weshaggard:
For the content\ci.yml files you might need to rename them so that our automation that sets up pipelines doesn't try to setup a pipeline for them. Perhaps we can name them ci.yml.template or something like that and as part of the template generation process rename the file to ci.yml.

weshaggard:
FYI @ArthurMa1978 given that this change impacts the mgmt template as well.

heaths:
1. Tip: for a PR, instead of merging from main which can make PRs difficult to review in some scenarios, you can more cleanly rebase on main to merge e.g.,
   ```bash
   git fetch upstream main
   git rebase upstream/main # run this in your topic branch
   git push --force
   ```
2. Given the email thread, perhaps we want to consider keeping eng/template the *only* templates and have that generate sdk/sample/Azure.Template.Sample or maybe sdk/template/Azure.Template.Sample and remove the .template.config directory from sdk/template now.

sdk/template/Azure.Template/samples/README.md
```diff
@@ -11,4 +11,13 @@ description: Samples for the Azure.Template client library.
 
 # Azure.Template Samples
 
-- [Creating Resource](https://github.com/Azure/azure-sdk-for-net/blob/3ac301ac6435c818ad7a9946ab1c4023cee236ff/eng/templates/Azure.Template/samples/Sample1_CreateResource.md)
\ No newline at end of file
+These code samples show common scenario operations with the Azure Template client library.
```
heaths, line 14:
```suggestion
These code samples show common scenarios with the Azure Template client library.
```

sdk/template/Azure.Template/samples/README.md
```diff
@@ -11,4 +11,13 @@ description: Samples for the Azure.Template client library.
...
+
+|**Sample Name**|**Description**|
```
heaths, line 16: Nit: we don't generally format table headers and just let the default markdown rendering styles apply.

sdk/template/Azure.Template/samples/README.md
```diff
@@ -11,4 +11,13 @@ description: Samples for the Azure.Template client library.
...
+|----------------|-------------|
+|[Sample1_HelloWorld][sample_helloworld] |Get secrets synchronously.|
```
heaths, line 18: Tip: if you name the links the same token value you can use an abbreviated form:

```markdown
[Sample1_HelloWorld]
[Sample1_HelloWorldAsync]

[Sample1_HelloWorld]: link1
[Sample1_HelloWorldAsync]: link2
```

sdk/template/Azure.Template/samples/Sample1.HelloWorldAsync.md
```diff
@@ -0,0 +1,25 @@
+# Get secret
```
heaths, line 1:
```suggestion
# Get secret asynchronously
```

heaths: Also, why does this file use a period as a separator in the file name while Sample1_HelloWorld.md does not? I'd stick with an underscore for both as we normally do throughout the repo (mostly).

sdk/template/Azure.Template/tests/Samples/Sample1_HelloWorld.cs
```diff
@@ -0,0 +1,37 @@
...
+
+namespace Azure.Template.Tests.Samples
+{
+    public partial class Sample1_HelloWorld: SamplesBase<TemplateClientTestEnvironment>
+    {
+        [Test]
+        [SyncOnly]
+        public void GettingASecret()
+        {
+            var endpoint = TestEnvironment.KeyVaultUri;
```
heaths, line 21: We try to avoid using `var` throughout our source, especially for samples. A few instances throughout this PR to also update.

sdk/template/Azure.Template/tests/Samples/Sample1_HelloWorld.cs
```diff
@@ -0,0 +1,37 @@
...
+
+            #region Snippet:Azure_Template_GetSecret
+#if SNIPPET
+            string endpoint = "https://myvault.vault.azure.net";
```
heaths, line 25: This was already defined and won't compile. Instead, replace the definition of `endpoint` (including line 21 above) with code like this:

```csharp
    string endpoint = "https://myvault.vault.azure.net";
#if !SNIPPET
    endpoint = TestEnvironment.KeyVaultUri;
#endif
    SecretClient client = new SecretClient(new Uri(endpoint), new DefaultAzureCredential());
```

sdk/template/Azure.Template/tests/Samples/Sample1_HelloWorld.cs
```diff
@@ -0,0 +1,37 @@
...
+#endif
+            var client = new TemplateClient(endpoint, new DefaultAzureCredential());
```
heaths, line 27: This won't compile:

1. `endpoint1 is not a `Uri`. See my sample below for how to fix that.
2. This was a grow-up example and this should probably be a `SecretClient`.

I highly recommend using a real PR as opposed to a draft so it goes through all the normal checks which would've caught this. Draft PRs are more for when you're prototyping some idea and don't want it merged.

sdk/template/Azure.Template/tests/Samples/Sample1_HelloWorldAsync.cs
```diff
@@ -12,28 +12,8 @@
 
 namespace Azure.Template.Tests.Samples
 {
-    public class TemplateClientSamples: SamplesBase<TemplateClientTestEnvironment>
+    public partial class Sample1_HelloWorld: SamplesBase<TemplateClientTestEnvironment>
```
heaths, line 15: Just call the class `Sample1` and avoid the underscores in names. Likely will trigger Roslyn errors. All the SampleN samples should be unique anyway - the words after are just to provide some meaning to the file names. The classes themselves aren't really paramount to the sample - just the method names.

heaths (changes requested):
I suggest converting this to a full PR instead of a draft. It'll run more checks that would've cause various syntax errors I noticed. We typically only use draft PRs when you basically want to try out some idea and make sure it's obvious it shouldn't be merged yet, even if reviewed and approved.

sdk/template/Azure.Template/src/Azure.Template.csproj
chunyu3: removed

sdk/template/Azure.Template/src/Generated/TemplateClient.cs
chunyu3: If swagger does not define its format uri and no customization configuration added, codegen will generate it as string. This is an example, we should not add customization to do so.

sdk/template/Azure.Template/src/Generated/TemplateClient.cs
chunyu3: This is generated by current autorest.csharp. When the autorest.csharp upgrade, the file should be re-generated.

sdk/template/Azure.Template/src/TemplateClient.cs
chunyu3: No. removed this line.

sdk/template/Azure.Template/src/TemplateClient.cs
chunyu3: updated

sdk/template/ci.yml
```diff
@@ -25,18 +26,9 @@ pr:
   paths:
     include:
     - sdk/template/
-#if (false)
```
weshaggard, line 28: Did you intend to remove these?

chunyu3: updated

sdk/template/Azure.Template/src/GlobalSuppressions.cs
chunyu3: Removed in the Azure.Template

sdk/template/Azure.Template/tests/Samples/Sample1_HelloWorld.cs
chunyu3: Current codegen only generate according to the swagger definition, that is the endpoint is string. It is compiled successfully.

sdk/template/Azure.Template/tests/Samples/Sample1_HelloWorldAsync.cs
chunyu3: updated

sdk/template/Azure.ServiceTemplate.Template/tests/TemplateClientTest.cs
chunyu3: removed

sdk/template/Azure.Template/src/Generated/TemplateClient.cs
chunyu3: use a faked scope https://sample/.default

sdk/template/Azure.Template/src/GlobalSuppressions.cs
chunyu3: remove from the Azure.Template

sdk/template/Azure.Template/tests/Samples/Sample1_HelloWorld.cs
chunyu3: updated

sdk/template/Azure.Template/tests/Samples/Sample1_HelloWorldAsync.cs
```diff
@@ -0,0 +1,35 @@
...
+// Licensed under the MIT License.
+
+using System;
+using System.Threading.Tasks;
+using Azure.Core.TestFramework;
+using Azure.Identity;
+using Azure.Template.Models;
+using NUnit.Framework;
+
+namespace Azure.Template.Tests.Samples
```
lirenhe, line 11: It looks the file content has nothing related to 'HelloWorld', can we rename the file name to 'TemplateSamples' directly? (same as the class name)

sdk/template/Azure.Template/src/TemplateClient.cs
```diff
@@ -0,0 +1,69 @@
...
+                throw;
+            }
+        }
+
+        /// <summary> The GET operation is applicable to any secret stored in Azure Key Vault. This operation requires the secrets/get permission. </summary>
+        /// <param name="secretName"> The name of the secret. </param>
+        /// <param name="cancellationToken"> The cancellation token to use. </param>
+        public virtual Response<SecretBundle> GetSecretValue(string secretName, CancellationToken cancellationToken = default)
+        {
+            using var scope = ClientDiagnostics.CreateScope("TemplateClient.GetSecretValue");
```
lirenhe, line 49: Seems in our code, we use a combination of 'var' and actual type when declare a varible: 
     var scope
     RequestContext context 
Do we have a guidance about what pattern to use and shall we make the code consistent for this?

annelo-msft: Ah, @heaths  confirms this here: https://github.com/Azure/azure-sdk-for-net/pull/28016#discussion_r850879368

sdk/template/Azure.Template/tests/Samples/Sample1_HelloWorldAsync.cs
chunyu3: No.  each operation will have two sample file (sync, and async), and the two files will define the same class. The file name is consistent with the guideline. Please check https://azure.github.io/azure-sdk/dotnet_introduction.html#dotnet-samples

sdk/template/Azure.Template/src/TemplateClient.cs
chunyu3: resolved

lirenhe (approved):
LGTM. 
Please work with Wes and Heath to get it reviewed.

chunyu3:
Hi @weshaggard  @heaths  would you please help to review this PR? Thanks

sdk/template/Azure.Template/src/TemplateClient.cs
heaths: FWIW, it's actually in our style settings: https://github.com/Azure/azure-sdk-for-net/blob/951e011c5423c2e427a4068a3281f314f4c8c4c9/.editorconfig#L42-L45

heaths (changes requested):
Close, but a few blockers in comments and below:

1. When I used this template, I passed `-g Sample -li Client` but found lots of instances of "sample" instead of "Sample", including the project name. This means package file names won't be properly cased. See my old .template.config for how I resolved this. Case can be important, so you may need to calculate variables with both case-honoring names and tolower() variants. Instances of "sample" were all over. I recommend using `dotnet new dpg -g Sample -li Client -s http://localhost/swagger.json` or similar and going through the output. To list a few...
   1. sample.sln (the name, and in the contents.
   2. README.md and other sample markdowns: "sample"
   3. All file names I noticed.
   4. autorest.md: `namespace sample` (will affect generated code namespace).

sdk/template/Azure.Template/.content/samples/README.md
```diff
@@ -0,0 +1,14 @@
...
+products:
+# Including relevant stubs from https://review.docs.microsoft.com/help/contribute/metadata-taxonomies#product
+- azure
+name: Azure.Template samples for .NET
+description: Samples for the Azure.Template client library.
+---
+
+# Azure.ServiceTemplate.Template Samples
+
+<!-- please refer to <https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/template/Azure.Template/samples/README.md> to write sample readme. -->
```
weshaggard, line 14: Isn't that this file?

sdk/template/Azure.Template/tests/Azure.Template.Tests.csproj
```diff
@@ -11,7 +20,15 @@
   </ItemGroup>
 
   <ItemGroup>
-    <ProjectReference Include="$(AzureCoreTestFramework)" />
-    <ProjectReference Include="..\src\Azure.Template.csproj" />
+    <Folder Include="SessionRecords\" />
+  </ItemGroup>
+
+  <ItemGroup Condition="$(TargetFramework) == 'net461'">
```
weshaggard, line 26: Do we need System.Web reference in the template?

sdk/template/ci.yml
weshaggard: Looks like this if condition is still removed I suspect we still want these paths here.

eng/scripts/automation/GenerateAndBuildLib.ps1
```diff
@@ -103,27 +127,29 @@ function New-DataPlanePackageFolder() {
   $inputfile = ""
   $fileArray = $inputfiles.Split(";")
   if (($inputfiles -ne "") -And ($fileArray.Length -gt 0)) {
-    $inputfile = "- " + $fileArray[0];
-    for ($i = 1; $i -lt $fileArray.Count ; $i++) {
+    # $inputfile = "- " + $fileArray[0];
```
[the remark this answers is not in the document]
chunyu3: removed

eng/scripts/automation/GenerateAndBuildLib.ps1
```diff
@@ -135,9 +161,9 @@ function New-DataPlanePackageFolder() {
 
     $libraryName = $namespaceArray[-1]
     $groupName = $namespaceArray[1]
-    $dotnetNewCmd = "dotnet new dataplane --libraryName $libraryName --groupName $groupName --includeCI true --force"
+    $dotnetNewCmd = "dotnet new dpg --libraryName $libraryName --groupName $groupName --force"
```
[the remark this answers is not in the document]
chunyu3: Use azsdkdpg as the template name. Thanks

eng/scripts/automation/Invoke-GenerateAndBuild.ps1
```diff
@@ -45,6 +45,12 @@ if ( $? -ne $True) {
   Write-Error "Failed to generate sdk. exit code: $?"
   exit 1
 }
+
+Invoke-Build -sdkfolder $projectFolder
+if ( $? -ne $True) {
```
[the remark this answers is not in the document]
chunyu3: updated. Thanks

sdk/template/Azure.Template/.template.config/template.json
```diff
@@ -1,15 +1,14 @@
 {
   "$schema": "http://json.schemastore.org/template",
-  "author": "chunyu@microsoft.com",
   "classifications": [ "Azure", "ClassLibrary" ],
-  "identity": "Azure.ServiceTemplate.TemplateProject",
+  "identity": "Azure.Template",
   "name": "Azure DataPlane SDK template: client project",
-  "shortName": "dataplane",
+  "shortName": "dpg",
```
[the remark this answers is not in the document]
chunyu3: updated to azsdkdpg

sdk/template/Azure.Template/.template.config/template.json
```diff
@@ -26,15 +25,41 @@
...
+            "src/swagger/**/*",
+            "tests/SessionRecords/**/*",
+            ".vs/**/*"
           ]
         }
       ]
     },
     {
-      "source": "./content",
+      "source": ".content/testResource",
```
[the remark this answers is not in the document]
chunyu3: added.

sdk/template/Azure.Template/.template.config/template.json
```diff
@@ -26,15 +25,41 @@
...
+    {
+      "source": ".content/tests",
+      "target": "./tests"
+    },
+    {
+      "source": ".content/src",
+      "target": "./src"
     }
   ],
   "symbols": {
```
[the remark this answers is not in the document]
chunyu3: Hi @heaths  do you mean AzureSDKCodeGeneration_DataPlane_Quickstart.md and Invoke-DataPlaneGenerateSDKPackage.ps1?  Yes, I updated the doc to use the new template

sdk/template/Azure.Template/CHANGELOG.md
```diff
@@ -1,106 +1,7 @@
 # Release History
 
-## 1.0.3-beta.19 (2020-09-24)
-- Test Submit-PR
+## 1.0.0-beta.1 (2022-04-13)
```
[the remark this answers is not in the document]
chunyu3: we will release DPG. so the change of 1.0.3-beta.20 will be 'Release DPG library'?

eng/scripts/automation/GenerateAndBuildLib.ps1
```diff
@@ -147,18 +173,26 @@ function New-DataPlanePackageFolder() {
...
     $file = (Join-Path $projectFolder "src" $AUTOREST_CONFIG_FILE)
-    Update-AutorestConfigFile -autorestFilePath $file -inputfile -readme $readme
+    Update-AutorestConfigFile -autorestFilePath $file -readme $readme
     # dotnet sln
     dotnet sln remove src\$namespace.csproj
     dotnet sln add src\$namespace.csproj
     dotnet sln remove tests\$namespace.Tests.csproj
     dotnet sln add tests\$namespace.Tests.csproj
     Pop-Location
+    dotnet new -u $sdkPath/sdk/template/Azure.Template
```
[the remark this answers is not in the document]
chunyu3: I thought it was better to release unused resource. But what your concerned the scenario that the someone installed the template manually should also be considered. In order not to uninstall the template which is installed manually, It is better not uninstall here. 
I will update it. Thanks

sdk/template/Azure.Template/CHANGELOG.md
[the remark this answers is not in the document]
chunyu3: updated

sdk/template/Azure.Template/.content/CHANGELOG.md
```diff
@@ -0,0 +1,11 @@
...
+
+## 1.0.0-beta.1 (Unreleased)
+
+### Features Added
+
+### Breaking Changes
+
+### Bugs Fixed
+
+### Other Changes
```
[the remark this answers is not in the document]
chunyu3: Add a skeleton file under "sdk/template/Azure.Template/.content/api/Azure.Template.netstandard2.0.cs"

sdk/template/Azure.Template/README.md
```diff
@@ -78,10 +78,10 @@ Each example in the *Examples* section starts with an H3 that describes the exam
 The `GetSecret` method retrieves a secret from the service.
 
 ```C# Snippet:Azure_Template_GetSecret
-string endpoint = "https://myvault.vault.azure.net";
-var client = new MiniSecretClient(new Uri(endpoint), new DefaultAzureCredential());
+endpoint = "https://myvault.vault.azure.net";
+TemplateClient client = new TemplateClient(endpoint, new DefaultAzureCredential());
```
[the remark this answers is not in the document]
chunyu3: updated

sdk/template/Azure.Template/.content/samples/README.md
chunyu3: No. All files under .content are skeleton files to replace the files in `sdk\template\Azure.Template` when generate sdk package via `dotnet new`.  They will show in the generated packages for the service except Azure.Template. I updated to use  placeholder for this.

sdk/template/ci.yml
chunyu3: Hi @heaths Shall we still add this if condition here in ci.yml? Thanks

weshaggard (commented):
This looks reasonable to me. Thanks for the consolidation. I will defer to @heaths for the final signoff.

eng/scripts/automation/Invoke-DataPlaneGenerateSDKPackage.ps1
```diff
@@ -22,17 +22,10 @@ $outputJson = Get-Content $outputJsonFile | Out-String | ConvertFrom-Json
 $projectFolder = $outputJson.projectFolder
 Write-Host "projectFolder:$projectFolder"
 Remove-Item $outputJsonFile
+# Generate Code
 Invoke-Generate -sdkfolder $projectFolder
-if ( $? -ne $True) {
-  Write-Error "Failed to create generate sdk."
-  exit 1
-}
-
```
heaths: So no error checking here? Alternatively, I often start my PowerShell scripts with `$ErrorActionPreference = 'Stop'` so any PowerShell error within the scripts or other scripts/commands it calls halts script execution.

sdk/template/Azure.Template/.template.config/template.json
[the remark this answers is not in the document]
heaths: I meant for any replacements in general, but probably the easiest way to find out is to use your template and just search for "template" (case-insensitive). I'll do that again before I sign off.

heaths (changes requested):
When I install the template from your PR and create a new project, I still see a number of issues:

1. Lowercase project name, which will impact the output file name and other issues:
   image
   image
2. The README and CHANGELOG.md files were in the current directory. You need to create a subdirectory for the calculated Package name. For example, if I run `dotnet new azsdkdpg -g Test -li Sample -s local.json` in sdk/test then I should see a number of files in sdk/test but then also an Azure.Test.Sample directory that has the src, tests, etc. directories.
3. If I search sdk/test (from my example above) for "template", I still find a lot of links with "template" in the URL:
   image
   If you checkout main and create a new project with the old sdk/template these all get cleaned up. Take a look at all the replacements I do: https://github.com/Azure/azure-sdk-for-net/blob/dfcc1226d2de34048723977f499f40fb4abe34eb/sdk/template/.template.config/template.json#L28-L91
   Everywhere you see "replaces" I'm replacing patterns. You do a great job of adding placeholders, but those need to be used in more places or just replace "sdk/template" with "sdk/{ServiceDirectory}" where "ServiceDirectory" is one of the required inputs I take. In fact, I was able to calculate just about everything with only 2 inputs but you could use the separate GroupName and LibraryName you do now - just add a calculated field that combines "Azure.{GroupName}.{LibraryName" into one symbol you replace "Azure.Template", create another replacement with "Azure{GroupName}{LibraryName}" for "AzureTemplate" (the "safe" name, as I called it), etc.

I recommend using `dotnet new azsdkdpg` in a directory. What comes out should match the same structure we use for any track 2 SDK e.g.,

* sdk/
  * {ServiceDirectory}/
    * Azure.{GroupName}.{LibraryName}/
      * src/
        * Azure.{GroupName}.{LibraryName}.csproj
        * (other files)
      * tests/
        * Azure.{GroupName}.{LibraryName}.Tests.csproj
        * (other files)
      * perf/ (and other directories, with proper project names)
      * Azure.{GroupName}.{LibraryName}.sln
      * README.md
      * CHANGELOG.md
    * Directory.Build.props
    * test-resources.json
    * ci.yml
    * tests.yml

weshaggard:
@chunyu3 to help us see what the final output is could you please run the template generation step and push the changes to another branch so we can see what it contains. Also list the exact generation command you used to generate it here.

eng/scripts/automation/Invoke-DataPlaneGenerateSDKPackage.ps1
heaths: 1. I explained in a previous comment that the lowercase package directory and project file names will impact the output name. We don't want to publish packages like Azure.test.sample. It looks bad, and is inconsistent with our package formatting. Your template config needs separate replacements for case-sensitive replacements, as well as case-sensitive replacements with lowercase replacement text just like I have in my original template config.
2. So then how do you copy the ci.yml, Directory.Build.props, test-resources.json, and tests.yml files in the service directory? I tried creating a template in different directories and never saw this important files.

chunyu3: #1 Yes, we need to avoid lowercase package directory. But current, DPG eng script will use the package folder name as the namespace, even we force to use the tempelate.json to re-generate the package name, in the Generated Code, the namespace is also the package name, it will cause build fail. So we write script and enforce this constrains in it and describe it in the start doc.
#2 you can set --includeCI in the `dotnet new` command ( `dotnet new azsdkdpg -g Test -li Sample -s local.json --includeCI`), then those ci.yml, Directory.Build.props, test-resources.json, and tests.yml files will be copied into the service directory.

heaths: Tip: don't use `#<num>` unless referring to an issue or pull request. It adds a link to these comments on those (very old) issues. Just use numbered lists e.g., `1.`, or quote and reply from the comment toolbar.

1. That sounds like a problem. Instead, I recommend the docs say to pass the service directory name and package name cased as they want it, and the template defines a separate variable with the lowercase name. That's what my previous template had and works just fine. The DPG script shouldn't be a factor. Instead of deriving the package name from the directory name, derive the directory name from the package name. Much easier to control casing. See my old template for an example. I recommend, like with your extension, you check out an older branch and run my extension to see how it works.
2. But those files should be there by default unless excluded or already there. Every track 2 library should have those files. Directory.Build.props is optional, but doesn't hurt and helps prevent teams from inventing their own ways when needed.

chunyu3:
Hi @weshaggard  @heaths  I updated the dpg template according to our discussion in the meeting. And I generated two sample SDKs via azsdkdpg template.  You can specify `dotnet new` `--name` option or not.

1. under `sdk/sample` directory, specify the `dotnet --name` option, the value of name is `Azure.<group>.<service>`, run cmd `dotnet new azsdkdpg --name Azure.IoT.MyService --groupName IoT --clientName MyService --serviceDirectory test --swagger https://github.com/dpokluda/azure-rest-api-specs/blob/be397aa65510bd4e8f87da539af2b0025f6f44ca/specification/deviceupdate/data-plane/Microsoft.DeviceUpdate/preview/2020-09-01/deviceupdate.json --securityScopes https://api.adu.microsoft.com/.default`

The generated SDK is https://github.com/chunyu3/azure-sdk-for-net/tree/TemplateSample/sdk/sample

2. under `sdk/test` directory, don't specify the `dotnet --name` option, it will use the test as the package directory. run cmd `dotnet new azsdkdpg --groupName IoT --clientName MyService --serviceDirectory test --swagger https://github.com/dpokluda/azure-rest-api-specs/blob/be397aa65510bd4e8f87da539af2b0025f6f44ca/specification/deviceupdate/data-plane/Microsoft.DeviceUpdate/preview/2020-09-01/deviceupdate.json --securityScopes https://api.adu.microsoft.com/.default`

The generated SDK is https://github.com/chunyu3/azure-sdk-for-net/tree/TemplateSample/sdk/test

chunyu3:
Also, we can run `Invoke-DataPlaneGenerateSDKPackage.ps1` to generate the whole SDK starter package.
```powershell
pwsh eng/scripts/automation/Invoke-DataPlaneGenerateSDKPackage.ps1 -service test -namespace Azure.Identity.MySample -sdkPath ./azure-sdk-for-net -inputfiles https://github.com/dpokluda/azure-rest-api-specs/blob/be397aa65510bd4e8f87da539af2b0025f6f44ca/specification/deviceupdate/data-plane/Microsoft.DeviceUpdate/preview/2020-09-01/deviceupdate.json -securityScope https://myservice/.default
```

sdk/template/ci.yml
weshaggard: I'm going to say we need to keep this removed section so that our template will still trigger to validate things under those paths. I'm not sure why this ended up getting removed but I think we should definitely add it back before merging this change.

heaths (approved):

sdk/template/ci.yml
benbp: @chunyu3 can you please add this back? We currently do not have pipeline runs for various engsys related PRs that rely on the template tests as our primary CI gate.

PR #28031, opened 2022-04-06
Update handling for host disposed case

JoshLove-msft:


sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusListener.cs
```diff
@@ -307,21 +301,16 @@ internal async Task ProcessSessionMessageAsync(ProcessSessionMessageEventArgs ar
             }
         }
 
-        private bool EnsureIsRunning()
+        private void EnsureIsRunning()
         {
-            if (!Started)
+            if (!Started || Disposed)
```
jsquire, line 306: Sorry... I feel like I should have thought of that earlier.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusListener.cs
```diff
@@ -307,21 +301,16 @@ internal async Task ProcessSessionMessageAsync(ProcessSessionMessageEventArgs ar
...
             {
-                _logger.LogWarning($"Message received for a listener that is not in started state ({_details.Value})");
-                return false;
+                var message =
+                    $"Message received for a listener that is not in a running state. The message will not be delivered to the function, and instead will be abandoned. " +
+                    $"(Listener started = {Started}, Listener disposed = {Disposed}, {_details.Value})";
+                _logger.LogWarning(message);
+                throw new InvalidOperationException(message);
```
jsquire, line 312: I thought we wanted to protect the customer code from exceptions.   Is this observable only on the function host?

jsquire (approved):

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusListener.cs
JoshLove-msft: This exception will be caught by the ServiceBusProcessor which will result in the message being abandoned rather than potentially autocompleted.

PR #28044, opened 2022-04-07
remove resource link from resourcemanager

ArcturusZhang:


m-nash (approved):
need to update api

Yao725 (approved):
LGTM

archerzz (approved):
If you want to disable the whole test case, you can use `#if false ......#endif`

sdk/resourcemanager/Azure.ResourceManager/tests/Scenario/ResourceLinkCollectionTests.cs
```diff
@@ -11,90 +11,90 @@
 
 namespace Azure.ResourceManager.Tests
 {
-    public class ResourceLinkCollectionTests : ResourceManagerTestBase
-    {
-        public ResourceLinkCollectionTests(bool isAsync)
-            : base(isAsync)//, RecordedTestMode.Record)
-        {
-        }
+    //public class ResourceLinkCollectionTests : ResourceManagerTestBase
```
ArthurMa1978, line 14: consider to just remove test cases

ArthurMa1978 (approved):

sdk/resourcemanager/Azure.ResourceManager/tests/Scenario/ResourceLinkCollectionTests.cs
ArcturusZhang: I would like to keep it in case we could easily find it when we want to put them back

Issue #9569, opened 2020-01-21
Revisit the ability to cache AMQP messages when adding events to the `AmqpEventBatch`


jsquire:
# Summary

Because an event batch is size-constrained and ensures that it only accepts events that will fit into the batch, when events are added to the batch using `TryAdd` they are first measured.   Because the size of an event depends on how it is represented for a network request, the transport batch is responsible for transforming the event to its wire format before measuring its size.

In the case of the `AmqpEventBatch`, this means that an `EventData` instance is transformed into an `AmqpMessage`, which is a disposable resource.  In order to make efficient use of resources, the `AmqpEventBatch` was intended to cache the `AmqpMessage` for each event that was accepted into the batch so that the transformation from `EventData` did not need to be repeated when a batch was sent.

Because an AMQP message cannot be sent more than once, whether the send was successful or not, each time sending of a batch is requested, the contents of the batch are copied into a new AMQP message that serves as a batch envelope, allowing the entire batch to be sent to the Event Hubs service as a single message.  This limitation means that the `AmqpEventBatch` being sent may not be disposed until after all attempts to send the batch have completed, or the event messages will not be available for copying into a new batch.

During stress testing, a scenario was discovered in which the `AmqpEventBatch` was being prematurely disposed when the associated AMQP connection/links to the service were under load, despite the dispose request being part of a continuation only expected to execute once the send operations were complete.  This proved to be elusive and a true root cause has not yet been discovered.

In order to ensure stability of the library, the decision was made to cache the non-volatile `EventData` instances within the `AmqpEventBatch` and perform the full translation to the corresponding AMQP format on send, in order to remove the possibility of premature disposal.  Going forward, it would be desirable to return to caching the AMQP message for each event, in order to make more efficient use of resources and avoid additional allocations and computation. 

# Scope of Work

- Determine a root cause and reliable reproduction for the premature disposal of the `AmqpEventBatch` within the [`AmqpProducer.SendAsync`](https://github.com/Azure/azure-sdk-for-net/blob/02c143cfff60080424bf13dd1acda0c8595b7083/sdk/eventhub/Azure.Messaging.EventHubs/src/Amqp/AmqpProducer.cs#L178) method as it existed in the linked version.

**--EITHER--** 

- Devise an approach that would protect against a race condition and ensure that the batch is not disposed before all send attempts have fully completed.

[...]

PR #28055, opened 2022-04-07
[Event Hubs] Event Batch Enhancements

jsquire:
# Summary

The focus of these changes is to avoid paying the cost of serialization from `EventData` instances to AMQP messages multiple times, where possible.

# Details

Previously, events were serialized to AMQP format when `TryAdd` was called for the `EventDataBatch` in order to measure the event size, and then the `EventData` instance was cloned and held in the batch.  When calling `SendAsync`, the events held by the batch were again serialized to AMQP format for each attempt at publishing; when retries were needed, the full serialization pass was made against all batch events.

This approach was used for two reasons.  First, the AMQP transport tracks delivery information on the serialized message directly and prevents messages from being sent twice; in error scenarios, attempts to retry sending were failing because of the transport behavior.  Additionally, when idempotent retries are enabled, the metadata for sequencing events must be applied as part of the send operation, preventing eager serialization and caching.

The changes included herein work around the transport limitation by allowing previously serialized messages to be packaged in a new AMQP batch envelope for each attempt to publish.  The transport tracks delivery on the envelope, which is discarded in the event of retries and is rebuilt without needing to perform a full serialization pass over the batch events.  This removes the need to clone the source event in `TryAdd` and allows the batch to hold the serialized AMQP message, eliminating the need to perform serialization for batch events when `SendAsync` is called.

When idempotent retries are enabled, the idempotent metadata written to an event cannot be calculated eagerly and must be determined at the time `SendAsync` is called.  Because of this, the transport batch now holds responsibility for applying sequencing and other publisher properties to AMQP messages in batch scenarios.  For non-batch scenarios, the producer will continue to apply sequencing to the event prior to serialization.

# References and Related

- [Revisit the ability to cache AMQP messages when adding events to the AmqpEventBatch (#9569)](https://github.com/Azure/azure-sdk-for-net/issues/9569)


sdk/eventhub/Azure.Messaging.EventHubs/src/Amqp/AmqpProducer.cs
```diff
@@ -454,23 +470,17 @@ public override async Task CloseAsync(CancellationToken cancellationToken)
...
                     try
                     {
-                        using AmqpMessage batchMessage = messageFactory();
-
-                        // Creation of the link happens without explicit knowledge of the cancellation token
-                        // used for this operation; validate the token state before attempting link creation and
-                        // again after the operation completes to provide best efforts in respecting it.
-
-                        EventHubsEventSource.Log.EventPublishStart(EventHubName, logPartition, operationId);
+                        using AmqpMessage batchMessage = MessageConverter.CreateBatchFromMessages(messages);
```
jsquire, line 477: This approach requires converting each AMQP message into an `ArraySegment<byte>` to be packaged in the envelope.  For retry scenarios, this work is performed for each attempt.  The alternative would be to create the batch envelope once and then clone it for each send attempt.   That requires the first send attempt to pay for creation of the batch envelope creation and a clone, while retries pay only for the clone.

I chose to optimize for the golden path here - since a single send attempt is the majority outcome, by creating a new batch envelope during each attempt, the first send avoids paying for an unneeded clone.

danielmarbach:
@jsquire happy to look at it some time on the weekend. When do you need feedback the latest?

jsquire:
> @jsquire happy to look at it some time on the weekend. When do you need feedback the latest?

Awesome!  I'd like to start running things through prolonged stress tests the week of the 18th.   Even if I end up closing this PR out get some burn in during nightly test runs before you get a chance to give it a look, I'd love to incorporate your feedback in another set of changes.

sdk/eventhub/Azure.Messaging.EventHubs/src/Amqp/AmqpMessageConverter.cs
```diff
@@ -295,16 +295,21 @@ public virtual PartitionProperties CreatePartitionPropertiesFromResponse(AmqpMes
...
+            // delivery state which prevents it from being sent during retries,
+            // clone it.
+
             if (source.Count == 1)
             {
                 switch (source)
                 {
                     case List<AmqpMessage> messageList:
-                        batchEnvelope = messageList[0];
+                        batchEnvelope = messageList[0].Clone();
```
danielmarbach, line 308: It seems it does only a shallow copy. Is that a problem?

https://github.com/Azure/azure-amqp/blob/master/src/AmqpMessage.cs#L466-L468

sdk/eventhub/Azure.Messaging.EventHubs/src/Amqp/AmqpProducer.cs
```diff
@@ -203,8 +203,27 @@ internal class AmqpProducer : TransportProducer
             Argument.AssertNotClosed(_closed, nameof(AmqpProducer));
             Argument.AssertNotClosed(ConnectionScope.IsDisposed, nameof(EventHubConnection));
 
-            AmqpMessage messageFactory() => MessageConverter.CreateBatchFromEvents(events, sendOptions?.PartitionKey);
```
danielmarbach, line 206: Yehaa. This factory was ugly... :partying_face:

sdk/eventhub/Azure.Messaging.EventHubs/src/Amqp/AmqpProducer.cs
```diff
@@ -311,7 +327,12 @@ internal class AmqpProducer : TransportProducer
             options.MaximumSizeInBytes ??= MaximumMessageSize;
             Argument.AssertInRange(options.MaximumSizeInBytes.Value, EventHubProducerClient.MinimumBatchSizeLimit, MaximumMessageSize.Value, nameof(options.MaximumSizeInBytes));
 
-            return new AmqpEventBatch(MessageConverter, options, ActiveFeatures);
+            // If idempotent publishing is enabled, a specialized transport batch is needed to
+            // support sequencing semantics during send operations.
+
+            return ((ActiveFeatures & TransportProducerFeatures.IdempotentPublishing) == 0)
```
danielmarbach, line 333: This is nitpicky. I was just wondering if this should be abstracted into `AmqpEventBatch.Create(MessageConverter, options, TransportProducerFeatures, ActiveFeatures)` and then return the base.

I was immediately wondering why we access ActiveFeatures here and then also passing it to the batch creation.

sdk/eventhub/Azure.Messaging.EventHubs/src/Amqp/IdempotentAmqpEventBatch.cs
```diff
@@ -0,0 +1,206 @@
...
+using Microsoft.Azure.Amqp;
+
+namespace Azure.Messaging.EventHubs.Amqp
+{
+    /// <summary>
+    ///   A set of events with known size constraints, based on messages to be sent
+    ///   using an AMQP-based transport appropriate for idempotent publishing.
+    /// </summary>
+    ///
+    internal class IdempotentAmqpEventBatch : AmqpTransportEventBatch
```
danielmarbach, line 19: would it make sense to seal?

sdk/eventhub/Azure.Messaging.EventHubs/src/Amqp/IdempotentAmqpEventBatch.cs
```diff
@@ -0,0 +1,206 @@
...
+                {
+                    return false;
+                }
+
+                // Clone the event when accepting into the batch to ensure
+                // that its size does change if the original reference is
+                // mutated.
+
+                _sizeBytes = size;
+                BatchEvents.Add(eventData.Clone());
```
danielmarbach, line 116: See my previous comment about shallow vs deep clone

danielmarbach (commented):
Very clean and good separation of concerns. I think the design is intention revealing. 

I was wondering if it would be possible to separate the payload from the metadata since the AmqpMessage allows getting the payload and then maybe it would be possible to reconstruct the AmqpMessage with the payload and fresh metadata. With that, you wouldn't have to pay the price of serializing the payload again and again on the idempotent path. But I'm not deep enough into all the mechanics here, so this might just be nonsensical gibberish what I'm writing here.

sdk/eventhub/Azure.Messaging.EventHubs/src/Amqp/AmqpProducer.cs
jsquire: yeah, it was.  The pivot to "keep events because the AMQP messages are throwing on retries" was very late in the ship cycle.

sdk/eventhub/Azure.Messaging.EventHubs/src/Amqp/AmqpProducer.cs
jsquire: Interesting thought.  I'll prototype that and see how it feels.  

The feature flags are intended to support a few things forthcoming that may need to be considered as part of measurement, conversion, or sending.   Currently, it's just the idempotent coarse-grained flag.

sdk/eventhub/Azure.Messaging.EventHubs/src/Amqp/IdempotentAmqpEventBatch.cs
jsquire: Can't do it in this case;  the model factory implementation is extending this.  😄

sdk/eventhub/Azure.Messaging.EventHubs/src/Amqp/IdempotentAmqpEventBatch.cs
jsquire: The intent, as you're likely aware, is to prevent mutating the batch size by manipulating the `EventData` source.   

Generally, we're trying to guard against setting the direct properties or adding additional members into the `ApplicationProperties` dictionary.   The properties are comprised of primitives and immutable types that we copy over to the clone.  The dictionary only supports primitive types and streams for serialization - the latter of which is a very unusual case.   _([ref](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Azure.Messaging.EventHubs/src/Amqp/AmqpMessageConverter.cs#L942))_

In the case where someone put a stream in the application properties dictionary and is writing to it, and has idempotent retries enabled ... our abstraction leaks and manipulating the stream does mutate the batch.   I'd call it more like a mid-level copy - not fully deep but also not totally shallow.   It's not ideal, but it's been a decent balance for the mainstream case.

sdk/eventhub/Azure.Messaging.EventHubs/src/Amqp/AmqpProducer.cs
jsquire: I like it; centralizes the concern.   I'll push this up shortly.

jsquire:
> I was wondering if it would be possible to separate the payload from the metadata since the AmqpMessage allows getting the payload and then maybe it would be possible to reconstruct the AmqpMessage with the payload and fresh metadata. With that, you wouldn't have to pay the price of serializing the payload again and again on the idempotent path. But I'm not deep enough into all the mechanics here, so this might just be nonsensical gibberish what I'm writing here.

I'm not sure that I'm following.   I think what you're suggesting is to shift some of the logic for sequencing to the transport batch and manipulate it at the AMQP message level.    Something like:

```csharp
// Producer.SendIdempotent

var publishingState = GetPublishingState(partitionId);

try
{
    var lastBatchSequenceNumber= batch.ApplyPublishingState(publishingState);

    await transportProducer.SendAsync(batch, ...);
    publishingState.LastSequenceNumber = lastBatchSequenceNumber;
}
catch
{
    batch.ResetPublishingState();
}
```

Where internally the batch would do something like:

```csharp
public class Batch
{
    public long ApplyPublishingState(PartitionPublishingState state)
    {
        var sequenceNumber = GetNextSequenceNumber(state.LastSequenceNumber);

        foreach (var message in _batchMessages)
        {
              message.MessageAnnotations.Map[AmqpProperty.ProducerSequenceNumber] = sequenceNumber;
              sequenceNumber = GetNextSequenceNumber(sequenceNumber);

              message.MessageAnnotations.Map[AmqpProperty.ProducerGroupId] = stateProducerGroupId;
              message.MessageAnnotations.Map[AmqpProperty.ProducerOwnerLevel] = state.ProducerOwnerLevel;
        }
    }

    public long ResetPublishingState()
    {
        foreach (var message in _batchMessages)
        {
              message.MessageAnnotations.Map.TryRemoveValue<int?>(AmqpProperty.ProducerSequenceNumber, out _);
              message.MessageAnnotations.Map.TryRemoveValue<long?>(AmqpProperty.ProducerGroupId, out _);
              message.MessageAnnotations.Map.TryRemoveValue<short?>(AmqpProperty.ProducerOwnerLevel,  out _);
        }
    }
}
```
_(I'd probably do that in the message converter, but close enough for illustration....)_


Am I close?

danielmarbach:
Yep exactly. It was just a wild thought and I haven't validated if that would even work or give us much.

jsquire:
> It was just a wild thought and I haven't validated if that would even work or give us much.

Not so wild; its a great suggestion that I didn't think of.   I refactored for this approach, which let me remove a lot of the idempotent-specific infrastructure.   I think it's a huge improvement to what was here.   Thanks!

sdk/eventhub/Azure.Messaging.EventHubs.sln
```diff
@@ -35,14 +35,6 @@ Global
 		Release|Any CPU = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(ProjectConfigurationPlatforms) = postSolution
-		{6221D9A4-8563-4B8B-86B2-73585D54F840}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
```
jsquire, line 38: I do not know what caused this, but things seem happy...

danielmarbach:
This looks nice! :partying_face: :balloon:

sdk/eventhub/Azure.Messaging.EventHubs.sln
danielmarbach: :magic_wand:

sdk/eventhub/Azure.Messaging.EventHubs/src/Amqp/AmqpEventBatch.cs
```diff
@@ -123,43 +126,26 @@ public override bool TryAdd(EventData eventData)
             Argument.AssertNotNull(eventData, nameof(eventData));
             Argument.AssertNotDisposed(_disposed, nameof(EventDataBatch));
 
-            // Reserve space for producer-owned fields that correspond to special
-            // features, if enabled.
-
-            if ((ActiveFeatures & TransportProducerFeatures.IdempotentPublishing) != 0)
```
jsquire, line 129: I forgot to add idempotent property reservation back in after collapsing back to a single batch.     Will fix.

sdk/eventhub/Azure.Messaging.EventHubs/src/Amqp/AmqpEventBatch.cs
```diff
@@ -183,12 +174,59 @@ public override void Clear()
...
+        public override int ApplyBatchSequencing(int lastSequenceNumber,
+                                                 long? producerGroupId,
+                                                 short? ownerLevel)
+        {
+            if (_batchMessages.Count == 0)
+            {
+                return lastSequenceNumber;
+            }
+
+            _startingSequenceNumber = NextSequence(lastSequenceNumber);
```
jsquire, line 206: I wanted to avoid embedding an `if` statement in the `foreach` loop, so I'm paying the cost of the unchecked increment/rollover twice rather than the conditional for each tick.

jsquire, line 206: If felt it was cleaner to do this outside the loop and pay for an extra unchecked increment/rollover rather than have a conditional for each tick of the loop just to set this on the first tick.

sdk/eventhub/Azure.Messaging.EventHubs/src/Producer/EventHubProducerClient.cs
```diff
@@ -1111,46 +1110,33 @@ public virtual async ValueTask DisposeAsync()
...
-                    var firstSequence = NextSequence(lastSequence);
-
-                    foreach (var eventData in eventSet)
-                    {
-                        lastSequence = NextSequence(lastSequence);
-                        eventData.PendingPublishSequenceNumber = lastSequence;
-                        eventData.PendingProducerGroupId = partitionState.ProducerGroupId;
-                        eventData.PendingProducerOwnerLevel = partitionState.OwnerLevel;
-                    }
+                    var lastSequence = eventBatch.ApplyBatchSequencing(partitionState.LastPublishedSequenceNumber.Value, partitionState.ProducerGroupId, partitionState.OwnerLevel);
```
christothes, line 1113: is `partitionState.LastPublishedSequenceNumber.HasValue` guaranteed to be true here?

christothes (approved):

sdk/eventhub/Azure.Messaging.EventHubs/src/Producer/EventHubProducerClient.cs
jsquire: It is.   The service sets a value for it when the link is opened if the idempotent publishing capability is set.  That happens via the `InitializePartitionStateAsync` call on L1108.   If the service can't initialize the link properties, an exception flow is triggered.

danielmarbach:
Looking forward to seeing the impact of this change with some nice profiler screenshots :D

PR #28066, opened 2022-04-07
Support suppressing nested client/internal scopes

lmolkova:
Related to https://github.com/Azure/autorest.csharp/issues/2069 and https://github.com/Azure/autorest.csharp/issues/2127

## What

Public client methods can call into each other:
- grow-up methods call into protocol methods
- LRO and convenience methods calling into other convenience methods
- different clients calling into each other (blob checkpoint store)

It can lead to multiple nested scopes created, which can be noisy and increases customer bills on telemetry. Writing/generating code in a way to prevent it is quite complicated.

We want to be able to detect nested scopes at runtime and suppress the creation of the inner scope if the outer scope is present. It should affect scopes created for public methods. HTTP spans should not be suppressed.

Example: 
- EventHubs.process (server/consumer scope)
  - [user code] BlobClient.Upload
    - HTTP HEAD /blob 200
    - HTTP PUT /blob 200
  - [eventhubs code] EventHubs.Checkpoint
    - [suppressed with this PR] ~~BlobClient.SetMetada~~
    - HTTP PUT /metadata 200

There is no intention to change the behavior of existing clients, this change initially targets new SDKs.

## How

Add user-configurable setting that enables suppression. (currently implemented with `bool DiagnosticsOptions.SuppressNestedClientSpans`)
   - it's useful to let users configure it: for existing clients, customers might want to enable suppression. For new clients, it might be useful to disable suppression for debugging purposes (e.g. logs also follow scopes for correlation)
   - If we do this,  DPG clients need to be generated in a way that sets `ClientOptions.DiagnosticsOptions.SuppressNestedClientSpans = true`

### DIFF
--- a/sdk/core/Azure.Core.TestFramework/src/Instrumentation/DiagnosticScopeValidatingInterceptor.cs
+++ b/sdk/core/Azure.Core.TestFramework/src/Instrumentation/DiagnosticScopeValidatingInterceptor.cs
@@ -204,7 +204,9 @@ namespace Azure.Core.TestFramework
                         }
 
                         Activity parent = e.Activity.Parent;
-                        if (parent != null && !parent.Tags.Any(tag => tag.Key == "kind" && (tag.Value == "client" || tag.Value == "internal")))
+                        if (parent != null &&
+                            parent.Tags.Any(tag => tag.Key == "az.namespace") &&
+                            parent.Tags.Any(tag => tag.Key == "kind" && (tag.Value == "client" || tag.Value == "internal")))
                         {
                             throw new InvalidOperationException($"Client diagnostic scopes should not be nested except for backward compatibility reasons. Use ClientDiagnostics constructor overload to suppress nested scopes.");
                         }
--- a/sdk/core/Azure.Core.TestFramework/src/Instrumentation/DiagnosticScopeValidatingInterceptor.cs
+++ b/sdk/core/Azure.Core.TestFramework/src/Instrumentation/DiagnosticScopeValidatingInterceptor.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.Linq;
 using System.Reflection;
 using System.Threading;
@@ -201,14 +200,6 @@ namespace Azure.Core.TestFramework
                         if (!e.Activity.Tags.Any(tag => tag.Key == "az.namespace"))
                         {
                             throw new InvalidOperationException($"All diagnostic scopes should have 'az.namespace' attribute, make sure the assembly containing **ClientOptions type is marked with the AzureResourceProviderNamespace attribute specifying the appropriate provider. This attribute should be included in AssemblyInfo, and can be included by pulling in AzureResourceProviderNamespaceAttribute.cs using the AzureCoreSharedSources alias.");
-                        }
-
-                        Activity parent = e.Activity.Parent;
-                        if (parent != null &&
-                            parent.Tags.Any(tag => tag.Key == "az.namespace") &&
-                            parent.Tags.Any(tag => tag.Key == "kind" && (tag.Value == "client" || tag.Value == "internal")))
-                        {
-                            throw new InvalidOperationException($"Client diagnostic scopes should not be nested except for backward compatibility reasons. Use ClientDiagnostics constructor overload to suppress nested scopes.");
                         }
 
                         if (lastException != null && !e.IsFailed)
--- a/sdk/core/Azure.Core.TestFramework/src/Instrumentation/DiagnosticScopeValidatingInterceptor.cs
+++ b/sdk/core/Azure.Core.TestFramework/src/Instrumentation/DiagnosticScopeValidatingInterceptor.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Reflection;
 using System.Threading;
@@ -200,6 +201,12 @@ namespace Azure.Core.TestFramework
                         if (!e.Activity.Tags.Any(tag => tag.Key == "az.namespace"))
                         {
                             throw new InvalidOperationException($"All diagnostic scopes should have 'az.namespace' attribute, make sure the assembly containing **ClientOptions type is marked with the AzureResourceProviderNamespace attribute specifying the appropriate provider. This attribute should be included in AssemblyInfo, and can be included by pulling in AzureResourceProviderNamespaceAttribute.cs using the AzureCoreSharedSources alias.");
+                        }
+
+                        Activity parent = e.Activity.Parent;
+                        if (parent != null && !parent.Tags.Any(tag => tag.Key == "kind" && (tag.Value == "client" || tag.Value == "internal")))
+                        {
+                            throw new InvalidOperationException($"Client diagnostic scopes should not be nested except for backward compatibility reasons. Use ClientDiagnostics constructor overload to suppress nested scopes.");
                         }
 
                         if (lastException != null && !e.IsFailed)
--- a/sdk/core/Azure.Core/api/Azure.Core.net461.cs
+++ b/sdk/core/Azure.Core/api/Azure.Core.net461.cs
@@ -395,11 +395,11 @@ namespace Azure.Core
         public bool IsDistributedTracingEnabled { get { throw null; } set { } }
         public bool IsLoggingContentEnabled { get { throw null; } set { } }
         public bool IsLoggingEnabled { get { throw null; } set { } }
-        public bool IsNestedSpanSuppressionEnabled { get { throw null; } set { } }
         public bool IsTelemetryEnabled { get { throw null; } set { } }
         public int LoggedContentSizeLimit { get { throw null; } set { } }
         public System.Collections.Generic.IList<string> LoggedHeaderNames { get { throw null; } }
         public System.Collections.Generic.IList<string> LoggedQueryParameters { get { throw null; } }
+        public bool SuppressNestedClientSpans { get { throw null; } set { } }
     }
     [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
     public readonly partial struct HttpHeader : System.IEquatable<Azure.Core.HttpHeader>
--- a/sdk/core/Azure.Core/api/Azure.Core.net461.cs
+++ b/sdk/core/Azure.Core/api/Azure.Core.net461.cs
@@ -395,7 +395,7 @@ namespace Azure.Core
         public bool IsDistributedTracingEnabled { get { throw null; } set { } }
         public bool IsLoggingContentEnabled { get { throw null; } set { } }
         public bool IsLoggingEnabled { get { throw null; } set { } }
-        public bool IsNestingClientSpansEnabled { get { throw null; } set { } }
+        public bool IsNestedSpanSuppressionEnabled { get { throw null; } set { } }
         public bool IsTelemetryEnabled { get { throw null; } set { } }
         public int LoggedContentSizeLimit { get { throw null; } set { } }
         public System.Collections.Generic.IList<string> LoggedHeaderNames { get { throw null; } }
--- a/sdk/core/Azure.Core/api/Azure.Core.net461.cs
+++ b/sdk/core/Azure.Core/api/Azure.Core.net461.cs
@@ -399,7 +399,7 @@ namespace Azure.Core
         public int LoggedContentSizeLimit { get { throw null; } set { } }
         public System.Collections.Generic.IList<string> LoggedHeaderNames { get { throw null; } }
         public System.Collections.Generic.IList<string> LoggedQueryParameters { get { throw null; } }
-        public bool AreNestedClientSpansEnabled { get { throw null; } set { } }
+        public bool IsNestingClientSpansEnabled { get { throw null; } set { } }
     }
     [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
     public readonly partial struct HttpHeader : System.IEquatable<Azure.Core.HttpHeader>
--- a/sdk/core/Azure.Core/api/Azure.Core.net461.cs
+++ b/sdk/core/Azure.Core/api/Azure.Core.net461.cs
@@ -399,7 +399,7 @@ namespace Azure.Core
         public int LoggedContentSizeLimit { get { throw null; } set { } }
         public System.Collections.Generic.IList<string> LoggedHeaderNames { get { throw null; } }
         public System.Collections.Generic.IList<string> LoggedQueryParameters { get { throw null; } }
-        public bool SuppressNestedClientSpans { get { throw null; } set { } }
+        public bool AreNestedClientSpansEnabled { get { throw null; } set { } }
     }
     [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
     public readonly partial struct HttpHeader : System.IEquatable<Azure.Core.HttpHeader>
--- a/sdk/core/Azure.Core/api/Azure.Core.net461.cs
+++ b/sdk/core/Azure.Core/api/Azure.Core.net461.cs
@@ -399,6 +399,7 @@ namespace Azure.Core
         public int LoggedContentSizeLimit { get { throw null; } set { } }
         public System.Collections.Generic.IList<string> LoggedHeaderNames { get { throw null; } }
         public System.Collections.Generic.IList<string> LoggedQueryParameters { get { throw null; } }
+        public bool SuppressNestedClientSpans { get { throw null; } set { } }
     }
     [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
     public readonly partial struct HttpHeader : System.IEquatable<Azure.Core.HttpHeader>
--- a/sdk/core/Azure.Core/api/Azure.Core.net461.cs
+++ b/sdk/core/Azure.Core/api/Azure.Core.net461.cs
@@ -395,11 +395,11 @@ namespace Azure.Core
         public bool IsDistributedTracingEnabled { get { throw null; } set { } }
         public bool IsLoggingContentEnabled { get { throw null; } set { } }
         public bool IsLoggingEnabled { get { throw null; } set { } }
+        public bool IsNestingClientSpansEnabled { get { throw null; } set { } }
         public bool IsTelemetryEnabled { get { throw null; } set { } }
         public int LoggedContentSizeLimit { get { throw null; } set { } }
         public System.Collections.Generic.IList<string> LoggedHeaderNames { get { throw null; } }
         public System.Collections.Generic.IList<string> LoggedQueryParameters { get { throw null; } }
-        public bool IsNestingClientSpansEnabled { get { throw null; } set { } }
     }
     [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
     public readonly partial struct HttpHeader : System.IEquatable<Azure.Core.HttpHeader>
--- a/sdk/core/Azure.Core/src/DefaultClientOptions.cs
+++ b/sdk/core/Azure.Core/src/DefaultClientOptions.cs
@@ -13,7 +13,6 @@ namespace Azure.Core
             Transport = HttpPipelineTransport.Create();
             Diagnostics.IsTelemetryEnabled = !EnvironmentVariableToBool(Environment.GetEnvironmentVariable("AZURE_TELEMETRY_DISABLED")) ?? true;
             Diagnostics.IsDistributedTracingEnabled = !EnvironmentVariableToBool(Environment.GetEnvironmentVariable("AZURE_TRACING_DISABLED")) ?? true;
-            Diagnostics.IsNestingClientSpansEnabled  = true;
         }
 
         private static bool? EnvironmentVariableToBool(string? value)
--- a/sdk/core/Azure.Core/src/DefaultClientOptions.cs
+++ b/sdk/core/Azure.Core/src/DefaultClientOptions.cs
@@ -13,7 +13,7 @@ namespace Azure.Core
             Transport = HttpPipelineTransport.Create();
             Diagnostics.IsTelemetryEnabled = !EnvironmentVariableToBool(Environment.GetEnvironmentVariable("AZURE_TELEMETRY_DISABLED")) ?? true;
             Diagnostics.IsDistributedTracingEnabled = !EnvironmentVariableToBool(Environment.GetEnvironmentVariable("AZURE_TRACING_DISABLED")) ?? true;
-            Diagnostics.SuppressNestedClientSpans = false;
+            Diagnostics.AreNestedClientSpansEnabled = true;
         }
 
         private static bool? EnvironmentVariableToBool(string? value)
--- a/sdk/core/Azure.Core/src/DefaultClientOptions.cs
+++ b/sdk/core/Azure.Core/src/DefaultClientOptions.cs
@@ -13,7 +13,7 @@ namespace Azure.Core
             Transport = HttpPipelineTransport.Create();
             Diagnostics.IsTelemetryEnabled = !EnvironmentVariableToBool(Environment.GetEnvironmentVariable("AZURE_TELEMETRY_DISABLED")) ?? true;
             Diagnostics.IsDistributedTracingEnabled = !EnvironmentVariableToBool(Environment.GetEnvironmentVariable("AZURE_TRACING_DISABLED")) ?? true;
-            Diagnostics.AreNestedClientSpansEnabled = true;
+            Diagnostics.IsNestingClientSpansEnabled  = true;
         }
 
         private static bool? EnvironmentVariableToBool(string? value)
--- a/sdk/core/Azure.Core/src/DefaultClientOptions.cs
+++ b/sdk/core/Azure.Core/src/DefaultClientOptions.cs
@@ -13,6 +13,7 @@ namespace Azure.Core
             Transport = HttpPipelineTransport.Create();
             Diagnostics.IsTelemetryEnabled = !EnvironmentVariableToBool(Environment.GetEnvironmentVariable("AZURE_TELEMETRY_DISABLED")) ?? true;
             Diagnostics.IsDistributedTracingEnabled = !EnvironmentVariableToBool(Environment.GetEnvironmentVariable("AZURE_TRACING_DISABLED")) ?? true;
+            Diagnostics.SuppressNestedClientSpans = false;
         }
 
         private static bool? EnvironmentVariableToBool(string? value)
--- a/sdk/core/Azure.Core/src/DiagnosticsOptions.cs
+++ b/sdk/core/Azure.Core/src/DiagnosticsOptions.cs
@@ -38,7 +38,7 @@ namespace Azure.Core
                 LoggedContentSizeLimit = diagnosticsOptions.LoggedContentSizeLimit;
                 IsDistributedTracingEnabled = diagnosticsOptions.IsDistributedTracingEnabled;
                 IsLoggingContentEnabled = diagnosticsOptions.IsLoggingContentEnabled;
-                IsNestedSpanSuppressionEnabled  = diagnosticsOptions.IsNestedSpanSuppressionEnabled;
+                SuppressNestedClientSpans  = diagnosticsOptions.SuppressNestedClientSpans;
             }
             else
             {
@@ -87,9 +87,9 @@ namespace Azure.Core
         public bool IsDistributedTracingEnabled { get; set; } = true;
 
         /// <summary>
-        /// Gets or sets value indicating whether nested distributed tracing spans are going to be created for client methods when called by other instrumented client methods.
+        /// Gets or sets value indicating whether distributed tracing spans are going to be created when this client's methods are called by other instrumented client methods.
         /// </summary>
-        public bool IsNestedSpanSuppressionEnabled  { get; set; }
+        public bool SuppressNestedClientSpans  { get; set; }
 
         /// <summary>
         /// Gets or sets value indicating whether the "User-Agent" header containing <see cref="ApplicationId"/>, client library package name and version, <see cref="RuntimeInformation.FrameworkDescription"/>
--- a/sdk/core/Azure.Core/src/DiagnosticsOptions.cs
+++ b/sdk/core/Azure.Core/src/DiagnosticsOptions.cs
@@ -38,7 +38,7 @@ namespace Azure.Core
                 LoggedContentSizeLimit = diagnosticsOptions.LoggedContentSizeLimit;
                 IsDistributedTracingEnabled = diagnosticsOptions.IsDistributedTracingEnabled;
                 IsLoggingContentEnabled = diagnosticsOptions.IsLoggingContentEnabled;
-                IsNestingClientSpansEnabled  = diagnosticsOptions.IsNestingClientSpansEnabled ;
+                IsNestingClientSpansEnabled  = diagnosticsOptions.IsNestingClientSpansEnabled;
             }
             else
             {
--- a/sdk/core/Azure.Core/src/DiagnosticsOptions.cs
+++ b/sdk/core/Azure.Core/src/DiagnosticsOptions.cs
@@ -38,7 +38,7 @@ namespace Azure.Core
                 LoggedContentSizeLimit = diagnosticsOptions.LoggedContentSizeLimit;
                 IsDistributedTracingEnabled = diagnosticsOptions.IsDistributedTracingEnabled;
                 IsLoggingContentEnabled = diagnosticsOptions.IsLoggingContentEnabled;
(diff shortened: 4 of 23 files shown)


sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -399,6 +399,7 @@ public partial class DiagnosticsOptions
         public int LoggedContentSizeLimit { get { throw null; } set { } }
         public System.Collections.Generic.IList<string> LoggedHeaderNames { get { throw null; } }
         public System.Collections.Generic.IList<string> LoggedQueryParameters { get { throw null; } }
+        public bool SuppressNestedClientSpans { get { throw null; } set { } }
```
annelo-msft, line 402: Do we want to allow customers to suppress inner scopes?

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
```diff
@@ -94,7 +103,11 @@ public void AddLink(string traceparent, string tracestate, IDictionary<string, s
 
         public void Start()
         {
-            _activityAdapter?.Start();
+            bool? started = _activityAdapter?.Start();
+            if (started.GetValueOrDefault() && _suppressNestedClientScopes)
```
annelo-msft, line 107: Are there any concurrency concerns here if two client methods are called from different threads?

jsquire (approved):
The Event Hubs and Service Bus parts look good to me.

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
christothes, line 107: nit - for the default case, it might be a bit more efficient to check `_suppressNestedClientScopes` first

sdk/core/Azure.Core/tests/ClientDiagnosticsTests.Net50.cs
```diff
@@ -99,6 +99,75 @@ public void StartsActivitySourceActivity()
...
+            DiagnosticScopeFactory clientDiagnostics = new DiagnosticScopeFactory("Azure.Clients", "Microsoft.Azure.Core.Cool.Tests", true, true);
+
+            using DiagnosticScope scope = clientDiagnostics.CreateScope("ClientName.ActivityName", (DiagnosticScope.ActivityKind)kind);
+            Assert.IsTrue(scope.IsEnabled);
+
+            scope.Start();
+            Assert.AreEqual("ClientName.ActivityName", Activity.Current.OperationName);
+            DiagnosticScope nestedScope = clientDiagnostics.CreateScope("ClientName.NestedActivityName", (DiagnosticScope.ActivityKind)kind);
+            Assert.IsFalse(nestedScope.IsEnabled);
+            nestedScope.Start();
```
annelo-msft, line 120: Curious: why start it after checking IsEnabled?

sdk/core/Azure.Core/tests/ClientDiagnosticsTests.Net50.cs
```diff
@@ -99,6 +99,75 @@ public void StartsActivitySourceActivity()
...
+
+        [Test]
+        [NonParallelizable]
+        public void NestedActivitiesOuterSampledOut()
+        {
+            using var _ = SetAppConfigSwitch();
+            using var activityListener = new TestActivitySourceListener("Azure.Clients.ClientName");
+            DiagnosticScopeFactory clientDiagnostics = new DiagnosticScopeFactory("Azure.Clients", "Microsoft.Azure.Core.Cool.Tests", true, true);
+            DiagnosticScope scope = clientDiagnostics.CreateScope("ClientName.ActivityName");
+            scope.AddAttribute("sampled-out", true);
```
annelo-msft, line 156: Curious: is this used outside the test framework?

sdk/core/Azure.Core/tests/ClientDiagnosticsTests.Net50.cs
```diff
@@ -99,6 +99,75 @@ public void StartsActivitySourceActivity()
             }
         }
 
+        [Theory]
+        [TestCase(0)]
+        [TestCase(2)]
```
christothes, line 104: I think you can change these to the actual  `ActivityKind` values since they are constant

sdk/core/Azure.Core/tests/ClientDiagnosticsTests.cs
```diff
@@ -237,11 +237,176 @@ public void CreatesASingleListenerPerNamespace()
...
+            nestedScope.Start();
+            Assert.AreEqual("ClientName.NestedActivityName", Activity.Current.OperationName);
+            nestedScope.Dispose();
+            Assert.AreEqual("ClientName.ActivityName", Activity.Current.OperationName);
+        }
+
+        [Theory]
+        [TestCase(null, false)]
+        [TestCase(false, false)]
+        [TestCase(true, true, Ignore = "ClientDiagnostics need new Azure.Core to ship before using DiagnosticsOptions.SuppressNestedClientSpans flag")]
```
annelo-msft, line 294: Curious: why?

sdk/core/Azure.Core/tests/ClientDiagnosticsTests.cs
```diff
@@ -237,11 +237,176 @@ public void CreatesASingleListenerPerNamespace()
...
+            scope.Dispose();
+        }
+
+        [Test]
+        [NonParallelizable]
+        public void NestedActivitiesNoSuppressionOuterDisabled()
+        {
+            DiagnosticScopeFactory clientDiagnostics = new DiagnosticScopeFactory("Azure.Clients", "Microsoft.Azure.Core.Cool.Tests", true, true);
+            DiagnosticScope scope = clientDiagnostics.CreateScope("ClientName.ActivityName");
+            scope.Start();
```
annelo-msft, line 380: Can you help me understand how the outer scope is disabled here?  Or is suppression prevented because they have different namespaces?

annelo-msft (commented):
This is super cool.  I left some questions, and I'll let other Azure.Core folks comment as well.  Thanks @lmolkova!

sdk/core/Azure.Core/src/Shared/DiagnosticScopeFactory.cs
```diff
@@ -16,11 +16,13 @@ internal class DiagnosticScopeFactory
         private static Dictionary<string, DiagnosticListener>? _listeners;
         private readonly string? _resourceProviderNamespace;
         private readonly DiagnosticListener? _source;
+        private readonly bool _suppressNestedClientScopes;
 
-        public DiagnosticScopeFactory(string clientNamespace, string? resourceProviderNamespace, bool isActivityEnabled)
+        public DiagnosticScopeFactory(string clientNamespace, string? resourceProviderNamespace, bool isActivityEnabled, bool suppressNestedClientScopes)
```
christothes, line 21: Should we assign a default value here so that new clients fall into the pit of success?

sdk/core/Azure.Core/tests/ClientDiagnosticsTests.cs
```diff
@@ -237,11 +237,176 @@ public void CreatesASingleListenerPerNamespace()
...
-            _ = new DiagnosticScopeFactory("Azure.Clients.1",  "Microsoft.Azure.Core.Cool.Tests", true);
-            _ = new DiagnosticScopeFactory("Azure.Clients.2",  "Microsoft.Azure.Core.Cool.Tests", true);
+            _ = new DiagnosticScopeFactory("Azure.Clients.1",  "Microsoft.Azure.Core.Cool.Tests", true, false);
+            _ = new DiagnosticScopeFactory("Azure.Clients.1",  "Microsoft.Azure.Core.Cool.Tests", true, false);
+            _ = new DiagnosticScopeFactory("Azure.Clients.2",  "Microsoft.Azure.Core.Cool.Tests", true, false);
 
             Assert.AreEqual(2, testListener.Sources.Count);
         }
+
+        [Theory]
```
christothes, line 247: Did we intentionally not use `[Test]` here?

sdk/core/Azure.Core/tests/ClientDiagnosticsTests.cs
```diff
@@ -237,11 +237,176 @@ public void CreatesASingleListenerPerNamespace()
...
+        [TestCase(0)]
+        [TestCase(2)]
```
christothes, line 249: Should we add a case here for an ActivityKind we ignore and validate that also?

sdk/core/Azure.Core/src/DiagnosticsOptions.cs
```diff
@@ -81,10 +83,15 @@ internal DiagnosticsOptions(DiagnosticsOptions? diagnosticsOptions)
...
         /// <summary>
-        /// Gets or sets value indicating whether distributed tracing spans are going to be created for this clients methods calls and HTTP calls.
+        /// Gets or sets value indicating whether distributed tracing spans are going to be created for the clients methods calls and HTTP calls.
         /// </summary>
         public bool IsDistributedTracingEnabled { get; set; } = true;
 
+        /// <summary>
+        /// Gets or sets value indicating whether distributed tracing spans created for inner public method calls should be suppressed.
+        /// </summary>
+        public bool SuppressNestedClientSpans { get; set; }
```
JoshLove-msft, line 93: I'm curious whether we actually need this - it looks like only three libraries are being updated to set this to False - does that mean those libraries have nested client spans? If they don't it may be worth just changing the default behavior rather than introducing this property.

sdk/core/Azure.Core.TestFramework/src/Instrumentation/DiagnosticScopeValidatingInterceptor.cs
```diff
@@ -202,6 +203,12 @@ internal static async ValueTask<T> ValidateDiagnosticScope<T>(Func<ValueTask<(T
                             throw new InvalidOperationException($"All diagnostic scopes should have 'az.namespace' attribute, make sure the assembly containing **ClientOptions type is marked with the AzureResourceProviderNamespace attribute specifying the appropriate provider. This attribute should be included in AssemblyInfo, and can be included by pulling in AzureResourceProviderNamespaceAttribute.cs using the AzureCoreSharedSources alias.");
                         }
 
+                        Activity parent = e.Activity.Parent;
+                        if (parent != null && !parent.Tags.Any(tag => tag.Key == "kind" && (tag.Value == "client" || tag.Value == "internal")))
+                        {
+                            throw new InvalidOperationException($"Client diagnostic scopes should not be nested except for backward compatibility reasons. Use ClientDiagnostics constructor overload to suppress nested scopes.");
```
JoshLove-msft, line 209: So if you do need to set the flag to False, how do you avoid this error? Wouldn't we need to check the value of the property to determine whether this check should be done?

sdk/core/Azure.Core/src/DiagnosticsOptions.cs
JoshLove-msft: Thinking about it more - I can see that we do need this. Even if it doesn't impact any of our released libraries, other consumers of Core could be broken by making this a default change to the behavior.
Still, I think we may be able to get away with passing "true" for the updated libraries.

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
```diff
@@ -12,27 +12,36 @@
...
 
-        internal DiagnosticScope(string scopeName, DiagnosticListener source, object? diagnosticSourceArgs, object? activitySource, ActivityKind kind)
+        internal DiagnosticScope(string scopeName, DiagnosticListener source, object? diagnosticSourceArgs, object? activitySource, ActivityKind kind, bool suppressNestedClientScopes)
         {
             IsEnabled = source.IsEnabled() || ActivityExtensions.ActivitySourceHasListeners(activitySource);
+            _suppressNestedClientScopes = suppressNestedClientScopes && (kind == ActivityKind.Internal || kind == ActivityKind.Client);
+            // ActivityKind.Internal and Client both can represent public API calls.
+            if (_suppressNestedClientScopes)
+            {
+                IsEnabled &= !ClientCustomerScopeStarted.Value;
```
JoshLove-msft, line 43: So with this, would we be preventing Scope2 from being created? 

- Create Scope1
- Start Scope 1
- Create Scope2
- Dispose Scope 1
- Start Scope 2
- Dispose Scope 2

I'm wondering if this check should be in Start rather than in the constructor.

sdk/core/Azure.Core/src/Shared/ClientDiagnostics.cs
```diff
@@ -30,7 +30,7 @@ public ClientDiagnostics(ClientOptions options)
         }
 
         public ClientDiagnostics(string optionsNamespace, string? providerNamespace, DiagnosticsOptions diagnosticsOptions)
-            : base(optionsNamespace, providerNamespace, diagnosticsOptions.IsDistributedTracingEnabled)
+            : base(optionsNamespace, providerNamespace, diagnosticsOptions.IsDistributedTracingEnabled, false)
```
JoshLove-msft, line 33: Should this respect the value in diagnosticOptions?

sdk/core/Azure.Core.TestFramework/src/Instrumentation/DiagnosticScopeValidatingInterceptor.cs
JoshLove-msft: Does this make the check here no longer necessary - https://github.com/Azure/azure-sdk-for-net/blob/d82b2ce6271cb4017f4037496fb973a9a3c78490/sdk/core/Azure.Core.TestFramework/src/ClientDiagnosticListener.cs#L147-L153

JoshLove-msft, line 209: The guidance in the exception message confused me a bit. I think we want people to set SuppressScopes to true if they are using DPG, but this may not actually be the cause of the issue. This exception could occur in an existing HLC or even in a grow-up DPG, where the root cause isn't that they need to SuppressScopes - but that there may be a usage issue where they have nested scopes in the convenience layer.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
lmolkova: discussed offline, agreed to allow users to configure it.

Stordlord:
We get a review for it and get back to you

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
lmolkova: If two client methods are called in two different threads, they will end up having two different executions contexts (and two different `AsyncLocal` instances), so we should be fine here.

I've also had to change the underlying implementation to rely on [`Activity.SetCustomProperty`](https://github.com/dotnet/runtime/blob/ed50b63b404cec5ab03cff0b0fde1d2f7327751e/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Activity.cs#L980) for other reasons, which is thread-safe

sdk/core/Azure.Core/tests/ClientDiagnosticsTests.Net50.cs
lmolkova: no, it's just a condition I simulated here and it's never used out of the test framework.

Stordlord (approved):
Approve

sdk/core/Azure.Core/tests/ClientDiagnosticsTests.cs
lmolkova: There is no listener created for outer scope , i.e. the scope will be disabled

sdk/core/Azure.Core/tests/ClientDiagnosticsTests.Net50.cs
lmolkova: unfortunately no, since ActivityKind is an internal type and can't be used in the public test

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
lmolkova: I want to avoid creating ActivityAdapter to save a bit of perf. Also the scope should have `IsEnabled` false so that we don't waste any perf populating attributes on it.

sdk/core/Azure.Core/src/Shared/ClientDiagnostics.cs
lmolkova: we can't until we ship new Azure.Core:

> E.g. Azure.Identity uses shared ClientDiagnostics source file. Identity has package reference to Azure.Core, i.e. it uses released version of Core.
> So when I define new property in DiagnositcOptions, Identity can't use it yet, i.e. ClientDiagnostics (being part of the identity package) does not know about it too.

JoshLove-msft: Maybe I'm misunderstanding, but with shared source the file will be compiled into each library that it is shared with, so dependencies don't come into play.

JoshLove-msft: Ah nevermind - I see your point. The public property is in DiagnosticOptions which is not yet shipped.

JoshLove-msft: Usually we would update to a project reference in this case or file an issue so that we don't forget to update this once Core ships.

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
```diff
@@ -370,9 +378,56 @@ static ActivityExtensions()
...
         private static Func<ICollection<KeyValuePair<string,object>>?>? CreateTagsCollectionMethod;
-
+        private static Func<Activity, string, object?>? GetCustomPropertyMethod;
+        private static Action<Activity, string, object>? SetCustomPropertyMethod;
+        private static string AzureSdkScopeLabel = "az.sdk.scope";
         private static readonly ParameterExpression ActivityParameter = Expression.Parameter(typeof(Activity));
 
+        public static bool IsNestedScopeEnabled(this Activity activity)
+        {
+            if (GetCustomPropertyMethod == null)
```
lmolkova, line 388: unfortunately, we can't use our own `AsyncLocal`. the problem is that we use shared sources in clietn libraries and have multiple copies of `ClientDiagnsotics`, `DiagnsoticScope`, etc classes. So we can't rely on shared static properties.

So we have to rely on something public, which is `Activity`. 
We should consider making DiagnosticScope public and then we can make these methods more efficient.

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
JoshLove-msft: Yeah, that's fair - it is definitely an edge case.

sdk/core/Azure.Core/tests/ClientDiagnosticsTests.Net50.cs
christothes: Is it not this type? https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.activitykind?view=net-6.0

sdk/core/Azure.Core.TestFramework/src/Instrumentation/DiagnosticScopeValidatingInterceptor.cs
lmolkova: I reverted the change to this file for now.  The new check has to be conditional and depend on `DiagnsoticsOptions.AreNestedClientSpansEnabled` flag value.

But  I can't do it until the new Azure.Core is shipped (TestFramework uses package reference). There are some storage tests that fail with this check on.

And you're right. With this check, `A scope has already started for event '{producedDiagnosticScope.Name}` becomes unnecessary.

JoshLove-msft: Up to you, but it may be easier to just flip to a project reference where needed in this PR, and then we can flip everything back after the Core release.

sdk/core/Azure.Core/src/DefaultClientOptions.cs
```diff
@@ -13,6 +13,7 @@ public DefaultClientOptions(): base(null, null)
             Transport = HttpPipelineTransport.Create();
             Diagnostics.IsTelemetryEnabled = !EnvironmentVariableToBool(Environment.GetEnvironmentVariable("AZURE_TELEMETRY_DISABLED")) ?? true;
             Diagnostics.IsDistributedTracingEnabled = !EnvironmentVariableToBool(Environment.GetEnvironmentVariable("AZURE_TRACING_DISABLED")) ?? true;
+            Diagnostics.AreNestedClientSpansEnabled = true;
```
JoshLove-msft, line 16: Suggest IsNestingClientSpansEnabled to better align with the other bool properties.

sdk/core/Azure.Core/tests/ClientDiagnosticsTests.Net50.cs
Stordlord: Make use of new

sdk/core/Azure.Core/tests/ClientDiagnosticsTests.Net50.cs
lmolkova: no intention, just copy/paste  :) refactored it

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
```diff
@@ -370,9 +378,56 @@ static ActivityExtensions()
...
+            {
+                var method = typeof(Activity).GetMethod("GetCustomProperty");
```
JoshLove-msft, line 390: Why do we need to use reflection? Is this method not available on all target platforms? If not, then does this mean the feature only works on .NET 5+?

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
christothes: Would it be possible to use the [Baggage ](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.activity.setbaggage?view=net-6.0)collection on the Activity itself to stash the context we need?

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
lmolkova: We use reflection here a lot for the following reason:
- many APIs are available since DiagnosticSource package version 5.0.0 (or sometimes 6.0). CustomProperties are one of these relatively new APIs
- DiagnsoticSource is OOB nuget package - you can install 5.0.0 on .NET 4.6. So we can't use .NET version as a heuristic
- Consumers of our telemetry require DiagnsoticSource 5.0+
  - ApplicationInsights .NET SDK started depending on it in Nov 2020 and versions which don't depend on it are deprecated.
  - OpenTelemetry has similar timeline and there was never a GA verion of it which depended on older DiagnosticSoruce

So de-facto absolute majority of consumers has DiagnosticSource 5.0.0+, regardless of their .NET version.

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
lmolkova: Good question, Baggage is something that propagates between services, there is no way to remove the baggage (ever), and is intended for applications to use.

christothes: According to [SetBaggage](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.activity.setbaggage?view=net-6.0) docs, items can be removed:

> Add or update the Activity baggage with the input key and value. If the input value is null - if the collection has any baggage with the same key, then this baggage will get removed from the collection.

It seemed like it could be appropriate because this is an application specific use case.

lmolkova: Ah, I didn't know you could remove it. Thanks! Still, it's not the right place to put out flags into since what will happen is:
- we start Activity with Baggage
- all its child activities will have copies of this package - in case user callback is called anywhere - there can be many of them which we have no control over
- we'll need to remove the baggage before headers are set - Baggage goes into headers
- we'll need to add baggage again after request is done to suppress anything under.

So, Baggage is not designed for it, CustomProperties are designed exactly for it.

JoshLove-msft (approved):

sdk/core/Azure.Core/tests/ClientDiagnosticsTests.Net50.cs
```diff
@@ -99,6 +99,168 @@ public void StartsActivitySourceActivity()
...
+            {
+                Assert.IsFalse(nestedScope.IsEnabled);
+                Assert.AreEqual("ClientName.ActivityName", Activity.Current.OperationName);
+            }
+
+            nestedScope.Dispose();
+            Assert.AreEqual("ClientName.ActivityName", Activity.Current.OperationName);
+        }
+
+        [TestCase(0, true)]  // Internal
```
AlexanderSher, line 139: You can use enum directly:

```cs
[TestCase(DiagnosticScope.ActivityKind.Internal, true)]
[TestCase(DiagnosticScope.ActivityKind.Server, false)]
[TestCase(DiagnosticScope.ActivityKind.Client, true)]
[TestCase(DiagnosticScope.ActivityKind.Producer, false)]
[TestCase(DiagnosticScope.ActivityKind.Consumer, false)]
```

sdk/core/Azure.Core/src/Shared/ClientDiagnostics.cs
```diff
@@ -30,8 +30,10 @@ public ClientDiagnostics(ClientOptions options)
         }
 
         public ClientDiagnostics(string optionsNamespace, string? providerNamespace, DiagnosticsOptions diagnosticsOptions)
-            : base(optionsNamespace, providerNamespace, diagnosticsOptions.IsDistributedTracingEnabled)
+            : base(optionsNamespace, providerNamespace, diagnosticsOptions.IsDistributedTracingEnabled, true)
         {
+            // TODO: use DiagnosticsOptions.IsNestingClientSpansEnabled  here after new Azure.Core ships:
```
annelo-msft, line 35: I think it would be best to update dependent projects to use a project reference, so we can see and test the changes end to end before merging them.  This is a standard Azure.Core procedure -- let us know if you need help with this.

sdk/core/Azure.Core/tests/ClientDiagnosticsTests.Net50.cs
lmolkova: `DiagnosticScope.ActivityKind` is internal, and the test method is public, so no, I can't

AlexanderSher: image

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
```diff
@@ -370,9 +378,56 @@ static ActivityExtensions()
         private static Func<object, bool>? ActivitySourceHasListenersMethod;
         private static Func<string, string?, ICollection<KeyValuePair<string, object>>?, object?>? CreateActivityLinkMethod;
         private static Func<ICollection<KeyValuePair<string,object>>?>? CreateTagsCollectionMethod;
-
+        private static Func<Activity, string, object?>? GetCustomPropertyMethod;
+        private static Action<Activity, string, object>? SetCustomPropertyMethod;
+        private static string AzureSdkScopeLabel = "az.sdk.scope";
```
AlexanderSher, line 383: From what I understand, `ActivityExtensions` type is a hack that uses reflection to access methods that we can't bind in compile time. `"az.sdk.scope"` is specific for nested scope only.

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
```diff
@@ -370,9 +378,56 @@ static ActivityExtensions()
...
         private static readonly ParameterExpression ActivityParameter = Expression.Parameter(typeof(Activity));
 
+        public static bool IsNestedScopeEnabled(this Activity activity)
```
AlexanderSher, line 386: We probably should define `GetCustomProperty` / `SetCustomProperty` pair of methods and have the bit that accesses `"az.sdk.scope"` property in `DiagnosticScope`.

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
```diff
@@ -299,6 +305,8 @@ public void Start()
                 }
 
                 _diagnosticSource.Write(_activityName + ".Start", _diagnosticSourceArgs ?? _currentActivity);
+
+                return _currentActivity;
```
AlexanderSher, line 309: Is it safe to return `Activity` instance here?

lmolkova: yes, it's safe. Anyone can also get it with `Activity.Current`, but we're making a small optimization. Also, `ActivityAdapter` (where this method is defined) is a private class and only used in `DiagnosticScope`, so we fully control who/how will call it

sdk/core/Azure.Core/src/Pipeline/Internal/RequestActivityPolicy.cs
```diff
@@ -54,7 +54,7 @@ public override void Process(HttpMessage message, ReadOnlyMemory<HttpPipelinePol
 
         private async ValueTask ProcessAsync(HttpMessage message, ReadOnlyMemory<HttpPipelinePolicy> pipeline, bool async)
         {
-            using var scope = new DiagnosticScope("Azure.Core.Http.Request", s_diagnosticSource, message, s_activitySource, DiagnosticScope.ActivityKind.Client);
+            using var scope = new DiagnosticScope("Azure.Core.Http.Request", s_diagnosticSource, message, s_activitySource, DiagnosticScope.ActivityKind.Client, true);
```
annelo-msft, line 57: We wouldn't need to change this file if the constructor took a default value here.

sdk/core/Azure.Core/tests/ClientDiagnosticsTests.Net50.cs
```diff
@@ -99,6 +99,168 @@ public void StartsActivitySourceActivity()
             }
         }
 
+        [TestCase(null)]
+        [TestCase(true)]
+        [NonParallelizable]
+        public void NestedClientActivitiesConfigurationClientOptions(bool? enableNestedScopes)
+        {
+            // TODO: ClientDiagnostics need new Azure.Core to ship before using
```
JoshLove-msft, line 107: Is this accurate?

sdk/core/Azure.Core/src/Shared/ClientDiagnostics.cs
lmolkova: discussed offline: it'd require switching everything to ProjectReference and decided not to do it.

sdk/core/Azure.Core/src/DiagnosticsOptions.cs
```diff
@@ -81,10 +83,15 @@ internal DiagnosticsOptions(DiagnosticsOptions? diagnosticsOptions)
...
         /// <summary>
-        /// Gets or sets value indicating whether distributed tracing spans are going to be created for this clients methods calls and HTTP calls.
+        /// Gets or sets value indicating whether distributed tracing spans are going to be created for the clients methods calls and HTTP calls.
         /// </summary>
         public bool IsDistributedTracingEnabled { get; set; } = true;
 
+        /// <summary>
+        /// Gets or sets value indicating whether nested distributed tracing spans are going to be created for client methods when called by other instrumented client methods.
+        /// </summary>
+        public bool IsNestingClientSpansEnabled  { get; set; }
```
lmolkova, line 93: discussed naming with @annelo-msft and @ShivangiReja. The suggestion is to have a flag that's `false` by default and we'll override it in new DPG clients to `true` to make it less confusing.

then the name should be something along the lines of `IsNestedSpanSuppressionEnabled`

sdk/core/Azure.Core/src/Pipeline/Internal/RequestActivityPolicy.cs
lmolkova: `RequestActivityPolicy` is special since it produces HTTP spans which should not be suppressed.

I added default value though (to suppress), but then it means that existing clients still need to set it (to not suppress).

sdk/core/Azure.Core/src/DiagnosticsOptions.cs
JoshLove-msft: In that case, why not go back to the initial name of SupressNestedSpans or the like? I find IsNestedSpanSuppressionEnabled even more confusing than it is currently.

lmolkova: @JoshLove-msft thank you for the suggestion, I was trying to follow the existing pattern of `Is*Enabled`, but `SupressNestedSpans` is much better.

sdk/core/Azure.Core/tests/ClientDiagnosticsTests.Net50.cs
lmolkova: discussed offline and yes.

annelo-msft (approved):

lmolkova:
Per offline discussion: the plan is that all DPG will set `SuppressNestedSpans = true` in their `ClientOptions` impl contructor.
The problem is that  WebPubSub was already GA-ed and it'd be breaking to change it. Went through the WebPubSub code and didn't find nested calls there, so it's ok to start suppressing calls there.

PR #28108, opened 2022-04-08
IoT Central 2021-11-01-preview .Net SDK Release

chgennar:


sdk/iotcentral/Microsoft.Azure.Management.IotCentral/src/Generated/Models/App.cs
```diff
@@ -21,7 +22,7 @@ namespace Microsoft.Azure.Management.IotCentral.Models
     /// The IoT Central application.
     /// </summary>
     [Rest.Serialization.JsonTransformation]
-    public partial class App : Resource
+    public partial class App : IResource
```
chgennar, line 25: @Yao725, this App class is missing the location and tags fields defined in the specification. In the specification, the App definition is defined to inherit all properties of the TrackedResource definition, which in turn inherits from Resource. Do you have any ideas why these are missing?

sdk/iotcentral/Microsoft.Azure.Management.IotCentral/src/Generated/AppsOperations.cs
```diff
@@ -292,10 +287,10 @@ public async Task<AzureOperationResponse<App>> CreateOrUpdateWithHttpMessagesAsy
...
         /// The cancellation token.
         /// </param>
-        public async Task<AzureOperationResponse<App>> UpdateWithHttpMessagesAsync(string resourceGroupName, string resourceName, AppPatch appPatch, Dictionary<string, List<string>> customHeaders = null, CancellationToken cancellationToken = default(CancellationToken))
+        public async Task<AzureOperationHeaderResponse<AppsUpdateHeaders>> UpdateWithHttpMessagesAsync(string resourceGroupName, string resourceName, AppPatch appPatch, Dictionary<string, List<string>> customHeaders = null, CancellationToken cancellationToken = default(CancellationToken))
         {
             // Send Request
-            AzureOperationResponse<App> _response = await BeginUpdateWithHttpMessagesAsync(resourceGroupName, resourceName, appPatch, customHeaders, cancellationToken).ConfigureAwait(false);
+            AzureOperationHeaderResponse<AppsUpdateHeaders> _response = await BeginUpdateWithHttpMessagesAsync(resourceGroupName, resourceName, appPatch, customHeaders, cancellationToken).ConfigureAwait(false);
             return await Client.GetPutOrPatchOperationResultAsync(_response, customHeaders, cancellationToken).ConfigureAwait(false);
```
chgennar, line 294: @Yao725, autorest appears to be generating this incorrectly. `GetPutOrPatchOperationResultAsync` returns an `AzureOperationResponse` but the method in this `AppsOperations` class expects an `AzyreOperationHeaderResponse`. Do you know who to contact about this?

sdk/iotcentral/Microsoft.Azure.Management.IotCentral/src/Generated/Models/App.cs
Yao725: Will take a look

chgennar: @Yao725, for this one it looks like the problem is caused by the fact that the App definition in the swagger spec is decorated with the "x-ms-azure-resource" property. When I remove this, the App class properly inherits from TrackedResource.

sdk/iotcentral/Microsoft.Azure.Management.IotCentral/src/Generated/AppsOperations.cs
chgennar: @Yao725, for this one I think the problem is that the AzureClientExtensions class is missing a method definition for:

  ```
  public static async Task<AzureOperationHeaderResponse<THeader>> GetPutOrPatchOperationResultAsync<THeader>(this IAzureClient client, AzureOperationHeaderResponse<THeader> response, Dictionary<string, List<string>> customHeaders, CancellationToken cancellationToken)
  ```

sdk/iotcentral/Microsoft.Azure.Management.IotCentral/src/Generated/Models/App.cs
Yao725: I agree with your comment, on the one hand, you marked `App` as `azure resource`, but on the other hand, the common type `Resource` which is inherited by `TrackedResource`, is also marked as `azure resource`. I think it is this duplication of definitions that causes this error. After I removed this one in `App`, the generated code is right:
```csharp
public partial class App : TrackedResource
```

sdk/iotcentral/Microsoft.Azure.Management.IotCentral/src/Generated/AppsOperations.cs
Yao725: I think this problem is also from the swagger file. According to the ARM RPC, a patch method should have a response of [200](https://github.com/Azure/azure-resource-manager-rpc/blob/master/v1.0/resource-api-reference.md#response-1), so, is there any reason that this patch method doesn't have this [response](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/iotcentral/resource-manager/Microsoft.IoTCentral/preview/2021-11-01-preview/iotcentral.json#L192-L217)? I tried to manually add a 200 response and the generated code is what we are expecting:
```csharp
AzureOperationResponse<App,AppsUpdateHeaders> _response = await BeginUpdateWithHttpMessagesAsync(resourceGroupName, resourceName, appPatch, customHeaders, cancellationToken).ConfigureAwait(false);
```

sdk/iotcentral/Microsoft.Azure.Management.IotCentral/src/Generated/Models/App.cs
chgennar: I'll remove that property from the spec.

sdk/iotcentral/Microsoft.Azure.Management.IotCentral/src/Generated/AppsOperations.cs
chgennar: This patch operation is asynchronous so 202 is returned.

sdk/iotcentral/Microsoft.Azure.Management.IotCentral/src/Generated/Models/App.cs
chgennar: PR for this update to the spec: https://github.com/Azure/azure-rest-api-specs/pull/18736

sdk/iotcentral/Microsoft.Azure.Management.IotCentral/src/Generated/AppsOperations.cs
Yao725: Yeah, I understand that it needs 202 because this operation is asynchronous. But I think it should both follow the docs that you and I mentioned. As 200 is missed, what will happen if you call the patch method but nothing is actually updated? I think return 200 is more reasonable in this scenario.

PR #28113, opened 2022-04-08
[Event Hubs] Cancellation Token Exception Guards

jsquire:
# Summary

The focus of these changes is to guard against callbacks defined using `CancellationToken.Register` in handlers which throw.  These exceptions bubble up from developer code through the linked tokens and can cause processor infrastructure to fail when the processor is stopping.

These guards will intercept the exceptions and log a warning before ignoring them so that the shutdown process is able to complete normally.

### DIFF
--- a/sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/EventProcessor.cs
+++ b/sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/EventProcessor.cs
@@ -1264,7 +1264,14 @@ namespace Azure.Messaging.EventHubs.Primitives
                     // Canceling the main source here won't cause a problem and will help expedite stopping
                     // the processor later.
 
-                    _runningProcessorCancellationSource?.Cancel();
+                    try
+                    {
+                        _runningProcessorCancellationSource?.Cancel();
+                    }
+                    catch (Exception cancelEx)
+                    {
+                        Logger.ProcessorStoppingCancellationWarning(Identifier, EventHubName, ConsumerGroup, cancelEx.Message);
+                    }
                 }
             }
             catch (OperationCanceledException ex)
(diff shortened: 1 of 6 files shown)


sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/EventProcessor.cs
```diff
@@ -1832,7 +1861,6 @@ private async Task RunProcessingAsync(CancellationToken cancellationToken)
                 }
 
                 partitionProcessor.Dispose();
-                cancellationToken.ThrowIfCancellationRequested<TaskCanceledException>();
```
jsquire, line 1835: This is somewhat orthogonal, but it should never have been here.   At this point, the partition processor has been stopped.  Honoring the token here just caused the handlers not to be invoked, violating their contract.

sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/EventProcessor.cs
```diff
@@ -1264,7 +1264,14 @@ protected virtual LastEnqueuedEventProperties ReadLastEnqueuedEventProperties(st
                     // Canceling the main source here won't cause a problem and will help expedite stopping
                     // the processor later.
 
-                    _runningProcessorCancellationSource?.Cancel();
+                    try
+                    {
+                        _runningProcessorCancellationSource?.Cancel();
```
JoshLove-msft, line 1269: Ah I missed the start processing cancellation in SB, though thinking about it I'm not sure how this would ever not be null.

JoshLove-msft (approved):

christothes (approved):

sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/EventProcessor.cs
jsquire: It never should if things are working correctly.  The only time that would happen is if Stop were broken and left an inconsistent state.

JoshLove-msft: Do we not set to null in a finally block in stop?

jsquire: We do not.   _([ref](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/EventProcessor.cs#L1365))_

PR #28118, opened 2022-04-08
Fix Cognitive Languages live tests

heaths:


heaths:
@ZulaMostafa were we using a complete different project for .NET then? Assertions are way off, expecting intents like `SushiMaking` instead of `EmailIntent`.

I fixed the variables (don't set them in the test-resources.json as empty, which will override those in tests.yml - I was afraid of that, but wasn't sure when I reviewed the PR), but if we want to use unified assets with Python we'll probably need to update our tests and samples.

/cc @annatisch @mshaban-msft

PR #28139, opened 2022-04-11
Set Operation Name for Consumer Kind

RDalziel:
When integrating with Mass Transit's OTEL implementation most items come through perfectly aside from consumer kind traces.

This is just a small update to at least allow the operation name to avoid the below output in app insights:

image


jsquire:
//fyi:  @lmolkova

RDalziel:
Any thoughts on this one folks ?

vishweshbankwar:
@RDalziel - Thank you for your PR. Schema mapping for consumer kind is not completed yet on our side. There is no ETA right now - Will update here when there are more details.

RDalziel:
Thanks @vishweshbankwar  - is there any scope for iterations here which would at least allow for consumers kinds to not show as blank on Application Insights?

PR #28146, opened 2022-04-11
Update Azure Deploy Button

jsquire:
# Summary

The focus of these changes is to update to the new Azure Deploy button image.


christothes (commented):
Looks like there are a [few others as well ](https://github.com/Azure/azure-sdk-for-net/search?q=%22http%3A%2F%2Fazuredeploy.net%2Fdeploybutton.png%22)(assuming the search index is not stale). Should we update those as well?

jsquire:
> Looks like there are a [few others as well ](https://github.com/Azure/azure-sdk-for-net/search?q=%22http%3A%2F%2Fazuredeploy.net%2Fdeploybutton.png%22)(assuming the search index is not stale). Should we update those as well?

Indeed we should.   I didn't think to run a search.  😄     I'll put out PRs for those shortly  since I've been doing it via the web UI.

christothes (approved):

PR #28186, opened 2022-04-13
Fix SDK automation script

ShivangiReja:
PR - https://github.com/Azure/azure-rest-api-specs/pull/18628
Issue - https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1493678&view=logs&j=03afb3bb-7296-55ad-aa07-ceee610c73b2&t=0f320c29-1c90-5304-d2d3-c58a6bb76437&l=128

Currently SDK automation script updates sha in autorest.md for only the [changed](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1493678&view=logs&j=03afb3bb-7296-55ad-aa07-ceee610c73b2&t=0f320c29-1c90-5304-d2d3-c58a6bb76437&l=69) files in the PR.

We found an issue in [ConfidentialLedger](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/confidentialledger/Azure.Security.ConfidentialLedger/src/autorest.md), It has three input files in autorest.md and [this](https://github.com/Azure/azure-rest-api-specs/pull/18628) PR updates only one  input file. So current script replaces the sha only for the changed input file, and it has a [common.json](https://github.com/Azure/azure-rest-api-specs/blob/8130d43a9d72b8281ecc84157cea2933208b1f04/specification/confidentialledger/data-plane/Microsoft.ConfidentialLedger/preview/0.1-preview/confidentialledger.json#L438) reference but other input files also have a common.json reference, so we load 2 different common.json because of there is 2 different hash to the input files, therefore the generation fails with  duplicate schema names error.

Fix: If there is any match found in autorest.md for the changed input files, update sha for all the input files.

### DIFF
--- a/eng/scripts/Automation-Sdk-Generate.ps1
+++ b/eng/scripts/Automation-Sdk-Generate.ps1
@@ -25,29 +25,24 @@ Write-Host "Updating autorest.md files for all the changed swaggers."
 $sdksInfo = @{}
 $headSha = $inputFileContent.headSha
 $repoHttpsUrl = $inputFileContent.repoHttpsUrl
+$regexToFindSha = "https:\/\/[^`"]*[\/][0-9a-f]{4,40}[\/]"
 foreach ($path in $autorestFilesPath) {
   $fileContent = Get-Content $path
   foreach ($inputFilePath in $inputFilePaths) {
-    $isUpdatedLines = $false
     $escapedInputFilePath = [System.Text.RegularExpressions.Regex]::Escape($inputFilePath)
-    $regexForMatchingShaAndPath = "https:\/\/[^`"]*[\/][0-9a-f]{4,40}[\/]$escapedInputFilePath"
-    $updatedLines = foreach ($line in $fileContent) {
-      if ($line -match $regexForMatchingShaAndPath) {
-        $line -replace $regexForMatchingShaAndPath, "$repoHttpsUrl/blob/$headSha/$inputFilePath"
+    $regexForMatchingShaAndPath = $regexToFindSha + $escapedInputFilePath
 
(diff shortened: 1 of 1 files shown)


eng/scripts/Automation-Sdk-Generate.ps1
```diff
@@ -25,29 +25,24 @@ Write-Host "Updating autorest.md files for all the changed swaggers."
...
-        $line -replace $regexForMatchingShaAndPath, "$repoHttpsUrl/blob/$headSha/$inputFilePath"
+        $fileContent -replace $regexToFindSha, "$repoHttpsUrl/blob/$headSha/" | Set-Content -Path $path
 
-        $isUpdatedLines = $true
         $sdkpath = (get-item $path).Directory.Parent.FullName | Resolve-Path -Relative
         if (!$sdksInfo.ContainsKey($sdkpath)) {
           $specReadmePath = $inputFileContent.relatedReadmeMdFiles -match $inputFilePath
           $sdksInfo.Add($sdkpath, $specReadmePath)
         }
+        break
```
christothes, line 44: Why do we want to break here?

ShivangiReja: If we found any match with the changed file in PR in an autorest.md, then [here](https://github.com/Azure/azure-sdk-for-net/blob/ee20c1adfdcc63342626f09153866a6c48d24bf0/eng/scripts/Automation-Sdk-Generate.ps1#L37) I'm updating all the sha in an autorest.md and added the `break` to get out of the loop. And then it will check the next autorest.md file in the outer loop to do the same.

christothes (approved):

Issue #28052, opened 2022-04-07
Event Hubs: Buffered Producer Performance Tests 


jsquire:
# Summary

In order to ensure consistent performance and measure across releases of the SDK, performance tests are needed for the `EventHubBufferedProducerClient`.  

# Scope of Work

- Create new scenarios in the existing [Azure.Messaging.EventHubs performance tests](https://github.com/Azure/azure-sdk-for-net/tree/master/sdk/eventhub/Azure.Messaging.EventHubs/perf) for the buffered producer:

  - Buffered send with automatic routing
  - Buffered send with partition keys
  - Buffered send to partitions

# Success Criteria

- The new scenarios for the `EventHubBufferedProducerClient` have been created in the Event Hubs performance project, following established patterns for processor-like types.

# References and Related

- [Azure.Messaging.EventHubs Performance Tests](https://github.com/Azure/azure-sdk-for-net/tree/master/sdk/eventhub/Azure.Messaging.EventHubs/perf)
- [Performance Testing Framework and Sample Project](https://github.com/Azure/azure-sdk-for-net/tree/master/common/Perf)

PR #28190, opened 2022-04-13
[Event Hubs] Performance Test Updates

jsquire:
# Summary

The focus of these changes is to add a set of performance tests for the `EventHubBufferedProducerClient`, and to flesh out and improve the existing tests for the `EventProcessorClient`.

# References and Related

- [Event Hubs: Buffered Producer Performance Tests (#28052)](https://github.com/Azure/azure-sdk-for-net/issues/28052)


sdk/eventhub/Azure.Messaging.EventHubs.Processor/perf/Azure.Messaging.EventHubs.Processor.Perf.csproj
```diff
@@ -6,6 +6,10 @@
     <LangVersion>latest</LangVersion>
     <IsTestSupportProject>true</IsTestSupportProject>
   </PropertyGroup>
+  <ItemGroup>
```
jsquire, line 9:
```suggestion

  <ItemGroup>
```

christothes: should this be removed?

sdk/eventhub/Azure.Messaging.EventHubs/perf/Azure.Messaging.EventHubs.Perf.csproj
```diff
@@ -7,6 +7,11 @@
     <IsTestSupportProject>true</IsTestSupportProject>
   </PropertyGroup>
 
+  <ItemGroup>
```
christothes, line 10: Is this needed?

christothes (approved):

sdk/eventhub/Azure.Messaging.EventHubs.Processor/perf/Azure.Messaging.EventHubs.Processor.Perf.csproj
jsquire: Nope; the omission is intentional.   I just forgot to make the spacing pretty.  😄

jsquire:
@mikeharder:  I'm going to commit this for now.   I'm happy to discuss the approach and incorporate any feedback that you may have when you return.

sdk/eventhub/Azure.Messaging.EventHubs.Processor/perf/Infrastructure/ProcessorTest.cs
```diff
@@ -3,61 +3,169 @@
...
         {
-            await base.SetupAsync();
-            await _checkpointStore.CreateIfNotExistsAsync();
+            try
+            {
+                await _checkpointStore.DeleteIfExistsAsync().ConfigureAwait(false);
+            }
+            finally
+            {
+                await Scope.DisposeAsync().ConfigureAwait(false);
```
mikeharder, line 124: NIT: Should `Scope.DisposeAsync()` be called from `CleanupAsync()` or `DisposeAsyncCore()`?

https://github.com/Azure/azure-sdk-for-net/blob/c56d65f9357b7c5f64aa7ed2fa31f0098480ee74/common/Perf/Azure.Test.Perf/PerfTestBase.cs#L87

sdk/eventhub/Azure.Messaging.EventHubs.Processor/perf/Scenarios/ProcessEventTest.cs
```diff
@@ -82,21 +102,48 @@ private async Task ProcessEventAsync(ProcessEventArgs args)
...
+        ///   Performs the tasks needed to initialize and set up the environment for the test scenario.
+        ///   This setup will take place once for each instance, running after the global setup has
+        ///   completed.
+        /// </summary>
+        ///
         public override async Task SetupAsync()
         {
             await base.SetupAsync();
+
+            EventProcessorClient.ProcessEventAsync += ProcessEventAsync;
```
mikeharder, line 128: I don't feel too strongly either way, but the typical pattern in perf tests is to do things in ctor/dispose if possible, and only use setup/cleanup if required (e.g. calling async methods).   So I think attach/detach event handlers could live in ctor/dispose.  But if you think the code is cleaner having everything in setup/cleanup, that's fine too.  It's just a convention we've been using.

sdk/eventhub/Azure.Messaging.EventHubs/perf/Infrastructure/BufferedPublishPerfTest.cs
```diff
@@ -0,0 +1,247 @@
...
+                }
+
+                return Task.CompletedTask;
+            };
+
+            // Read the available partitions and buffer a single events to establish the
+            // connection and link and start reading the buffers.
+
+            Partitions = await _producer.GetPartitionIdsAsync().ConfigureAwait(false);
+            await _producer.EnqueueEventsAsync(EventGenerator.CreateEvents(1)).ConfigureAwait(false);
```
mikeharder, line 116: Buffering a single event to establish the connection seems unnecessary, since the existing `warmup` period should handle this.

sdk/eventhub/Azure.Messaging.EventHubs/perf/Scenarios/BufferedPublishToPartitionsTest.cs
```diff
@@ -0,0 +1,49 @@
...
+        ///
+        /// <remarks>
+        ///     This method is expected to be called each time that a set of events
+        ///     is to be enqueued; it should be kept light and fast.
+        /// </remarks>
+        ///
+        protected override EnqueueEventOptions CreateEnqueueOptions() =>
+            new EnqueueEventOptions
+            {
+                PartitionId = Partitions[_random.Next(0, Partitions.Length - 1)]
```
mikeharder, line 46: The existing perf test `PublishEventsToPartition` uses `ParallelIndex` to determine the partition, so each parallel instance will always publish to the same partition, and you need to use N parallel instances to publish to N partitions:

https://github.com/Azure/azure-sdk-for-net/blob/c56d65f9357b7c5f64aa7ed2fa31f0098480ee74/sdk/eventhub/Azure.Messaging.EventHubs/perf/Scenarios/PublishEventsToPartition.cs#L37-L43

I don't have a preference which is better, but I think both tests should use the same algorithm, so we can get a fair comparison between the two publishers.

sdk/eventhub/Azure.Messaging.EventHubs.Processor/perf/Infrastructure/ProcessorTest.cs
jsquire: The existing tests in the core library are doing the scope manipulation in GlobalSetup/Setup and the teardown in the GlobalCleanup/Cleanup.  I went with the same pattern here for consistency.

I don't have strong opinions, but I'd prefer to keep it in Cleanup since we don't have the option of a GlobalDispose.

sdk/eventhub/Azure.Messaging.EventHubs.Processor/perf/Scenarios/ProcessEventTest.cs
jsquire: We need it in Setup because the client doesn't exist in the constructor.   We have to grab consumer group from the transient scope in the base class.   The scope is an async operation, so the client isn't created until base.Setup.

jsquire: We could assume the default consumer group, but generally, we want to allow the scope to dictate since it owns the set of resources and may have been asked to use specific consumer groups.

sdk/eventhub/Azure.Messaging.EventHubs/perf/Infrastructure/BufferedPublishPerfTest.cs
jsquire: Fair; we used this pattern in the other tests but I suppose that's mostly because of a lack of understanding at the time.   I'll make a pass to tweak that across the tests.

sdk/eventhub/Azure.Messaging.EventHubs/perf/Scenarios/BufferedPublishToPartitionsTest.cs
jsquire: Yeah, there's some differences between the types that we may or may not need to account for.   I'm going to drop 30 on your calendar next week;  I'd like to run the entire context by you for your thoughts.

PR #28193, opened 2022-04-13
Add validation for installation snippets

JoshLove-msft:


eng/scripts/CodeChecks.ps1
```diff
@@ -94,6 +96,45 @@ try {
         & $PSScriptRoot\Export-API.ps1 -ServiceDirectory $ServiceDirectory -SDKType $SDKType -SpellCheckPublicApiSurface:$SpellCheckPublicApiSurface
     }
 
+    Write-Host "Validating installation instructions"
+    Join-Path "$PSScriptRoot/../../sdk" $ServiceDirectory  `
```
heaths, line 100: `Join-Path` is pretty dumb. I'd use `[System.IO.Path]::Combine()`. If `$ServiceDirectory` is an absolute folder (can't remember if we supported that in these scripts, but we do in some places in the repo) then it'll be used as-is instead. `Join-Path` would just concat them for an invalid path.

eng/scripts/CodeChecks.ps1
```diff
@@ -94,6 +96,45 @@ try {
...
+        | Resolve-Path `
+        | % { Get-ChildItem $_ -Filter "README.md" -Recurse } `
+        | % {
+            $readmePath = $_
+            $readmeContent = Get-Content $readmePath
+            if ($readmeContent -Match "dotnet add .*--version.*$")
+            {
+                LogError "Specific versions should not be specified in the installation instructions. For beta versions, include the --prelease flag."
```
heaths, line 108:
```suggestion
                LogError "Specific versions should not be specified in the installation instructions. For beta versions, include the --prerelease flag."
```

eng/scripts/CodeChecks.ps1
```diff
@@ -94,6 +96,45 @@ try {
...
     }
 
+    Write-Host "Validating installation instructions"
+    Join-Path "$PSScriptRoot/../../sdk" $ServiceDirectory  `
+        | Resolve-Path `
+        | % { Get-ChildItem $_ -Filter "README.md" -Recurse } `
+        | % {
+            $readmePath = $_
+            $readmeContent = Get-Content $readmePath
+            if ($readmeContent -Match "dotnet add .*--version.*$")
```
heaths, line 106: The `.*$` doesn't really add any value and would impact performance (not that it matters here).
```suggestion
            if ($readmeContent -Match "dotnet add .*--version")
```

eng/scripts/CodeChecks.ps1
```diff
@@ -94,6 +96,45 @@ try {
...
+                    $entry = $changeLogEntries[$key]
+                    if ($entry.ReleaseStatus -ne "(Unreleased)" -and $entry.ReleaseVersion -notmatch "beta")
+                    {
+                        $hasGa = $true;
+                        break
+                    }
+                }
+                if ($hasGa)
+                {
+                    if (-Not ($readmeContent -Match "dotnet add (?!.*--prerelease).*$")){
```
heaths, line 128: Same here:
```suggestion
                    if (-Not ($readmeContent -Match "dotnet add (?!.*--prerelease)")){
```

eng/scripts/CodeChecks.ps1
```diff
@@ -94,6 +96,45 @@ try {
...
+                        LogError `
+"No GA installation instructions found in [$readmePath] but there was a GA entry in the Changelog [$changelogPath]. `
```
heaths, line 130: Nit: we typically surround paths in single quotes:
```suggestion
"No GA installation instructions found in '$readmePath' but there was a GA entry in the Changelog '$changelogPath'. `
```

heaths (approved):
Just a few nits/suggestions, but excellent idea!

eng/scripts/CodeChecks.ps1
JoshLove-msft: serviceDirectory is just the name of the service folder, e.g. "eventgrid"

eng/scripts/CodeChecks.ps1
jsquire: aww.   I like "prelease" better.

eng/scripts/CodeChecks.ps1
```diff
@@ -94,6 +96,45 @@ try {
...
+            
+            if ($readmeContent -Match "dotnet add")
+            {
+                $changelogPath = Join-Path $(Split-Path -Parent $readmePath) "CHANGELOG.md"
+                $changeLogEntries = Get-ChangeLogEntries -ChangeLogLocation $changelogPath
+                $hasGa = $false;
+                foreach ($key in $changeLogEntries.Keys)
+                {
+                    $entry = $changeLogEntries[$key]
+                    if ($entry.ReleaseStatus -ne "(Unreleased)" -and $entry.ReleaseVersion -notmatch "beta")
```
jsquire, line 120: Can't we sniff this from the package feed common function, and would that be more reliable?

eng/scripts/CodeChecks.ps1
jsquire, line 128: What about the scenario where the README demonstrates the GA and beta addition?   We've been asked by service teams in the Cognitive areas to consider doing this.   We haven't yet done so or formally discussed with the architects, but it may make sense to consider it being a possibility.   For example, something like:



### Install the package

Install the Azure Text Analytics client library for .NET with [NuGet](https://www.nuget.org/):

```dotnetcli
dotnet add package Azure.AI.TextAnalytics
```

### Install the latest beta release

Features currently in preview can be explored by installing the latest beta release:

```dotnetcli
dotnet add package Azure.Messaging.EventHubs
```

jsquire (approved):

eng/scripts/CodeChecks.ps1
JoshLove-msft: The check just ensures that there is at least a GA installation instruction if the package is GA. As the error message indicates, it is still okay to _also_ have prerelease instructions.

eng/scripts/CodeChecks.ps1
JoshLove-msft: The problem I see there is that we'd have to do two separate PRs - one to update changelog, and then a separate to update README after the release.

eng/scripts/CodeChecks.ps1
jsquire: my bad, then.   I thought it was enforcing that ALL instances of `dotnet add` were GA

JoshLove-msft: Nope, it is saying that if there is not at least one instance without --prerelease, fail the validation.

eng/scripts/CodeChecks.ps1
jsquire: Yeah, I see your point.   I was thinking the package feed would more reliably detect if there was already a GA, but it would fail to detect when this is the first ever GA until after the package was published - which is too late for the README riding along with it.

eng/scripts/CodeChecks.ps1
```diff
@@ -94,6 +96,44 @@ try {
...
+                    {
+                        $hasGa = $true;
+                        break
+                    }
+                }
+                if ($hasGa)
+                {
+                    if (-Not ($readmeContent -Match "dotnet add (?!.*--prerelease)")){
```
christothes, line 127: Might be worth while to add a some text describing examples of valid and invalid installation instructions just in case it's not clear.

christothes (approved):

eng/scripts/CodeChecks.ps1
JoshLove-msft: Perhaps, but this check is predicated on there already being a dotnet add entry, so the only thing we are really checking is that there is no version flag, and that there is one without a prerelease flag if already GAed. If we expand this check to look for the presence of these instructions, then I would agree having an example would be helpful.

eng/scripts/CodeChecks.ps1
heaths: Some scripts - like the TestResources script - do also accept a full directory spec. These may not, but in general I stay away from `Join-Path`. I opened a bug years ago and the PowerShell team agreed that for the FileSystem provider they probably should've used `Path.Combine` but that it'd be a breaking change.

heaths: To note, though, I'm not saying this has to be fixed. More of "tip".

JoshLove-msft: Yeah, I actually just copied this from another place in the same script where we are using Join-Path. But good to know about the potential pitfalls.

Mohit-Chakraborty (approved):

m-nash:
Is there any reason this is only done for a subset? https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/resourcemanager/Azure.ResourceManager/README.md#install-the-package.  How was this subset chosen and does there need to be a follow up to convert all?

m-nash (commented):
Is there any reason this is only done for a subset? https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/resourcemanager/Azure.ResourceManager/README.md#install-the-package. How was this subset chosen and does there need to be a follow up to convert all?

JoshLove-msft:
> Is there any reason this is only done for a subset? https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/resourcemanager/Azure.ResourceManager/README.md#install-the-package. How was this subset chosen and does there need to be a follow up to convert all?

Good question - I'm not sure why that one was not caught.

JoshLove-msft:
> > Is there any reason this is only done for a subset? https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/resourcemanager/Azure.ResourceManager/README.md#install-the-package. How was this subset chosen and does there need to be a follow up to convert all?
> 
> Good question - I'm not sure why that one was not caught.

Actually, looks like the Core analyze is still running so that should catch it.

m-nash (approved):
I would expect another 25 mgmt plane readme's to get caught by this, but the changes look good once they are all included

reyang (approved):

PR #28200, opened 2022-04-14
Added logging function name on ServiceBus exception.

alrod:
Port from T1:
https://github.com/Azure/azure-functions-servicebus-extension/commit/cd7528ae4aab8ba90531a90e33e41b4e46cc4801

### DIFF
--- a/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusExtensionConfigProvider.cs
+++ b/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusExtensionConfigProvider.cs
@@ -74,14 +74,6 @@ namespace Microsoft.Azure.WebJobs.ServiceBus.Config
                 throw new ArgumentNullException(nameof(context));
             }
 
-            // Set the default exception handler for background exceptions
-            // if not already set.
-            Options.ProcessErrorAsync ??= (e) =>
-            {
-                LogExceptionReceivedEvent(e, _loggerFactory);
-                return Task.CompletedTask;
-            };
-
             context
                 .AddConverter(new MessageToStringConverter())
                 .AddConverter(new MessageToByteArrayConverter())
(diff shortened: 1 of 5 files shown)


sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusExtensionConfigProvider.cs
```diff
@@ -74,14 +74,6 @@ public void Initialize(ExtensionConfigContext context)
                 throw new ArgumentNullException(nameof(context));
             }
 
-            // Set the default exception handler for background exceptions
-            // if not already set.
-            Options.ProcessErrorAsync ??= (e) =>
```
JoshLove-msft, line 79: Why do we move this from Initialize?

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Triggers/ServiceBusTriggerAttributeBindingProvider.cs
```diff
@@ -112,5 +120,40 @@ private bool GetAutoCompleteMessagesOptionToUse(ServiceBusTriggerAttribute attri
 
             return _options.AutoCompleteMessages;
         }
+
+        internal static void LogExceptionReceivedEvent(ProcessErrorEventArgs e, string functionName, ILoggerFactory loggerFactory)
+        {
+            try
+            {
+                var errorSource = e.ErrorSource;
+                var logger = loggerFactory?.CreateLogger(LogCategories.CreateFunctionCategory(functionName));
```
JoshLove-msft, line 129: Won't this end up creating a new logger for every error?

JoshLove-msft: I guess it was already like this, but would we be able to pass in a specific logger to this method?

JoshLove-msft (commented):
Thanks @alrod for contributing this! I just had a couple questions.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusExtensionConfigProvider.cs
alrod: This was moved to `ServiceBusTriggerAttributeBindingProvider` as we need to access the function name. But I just realized that in T1 we clone `ServiceBusOptions` and `ServiceBusOptions `is different for each function. For T2 ServiceBusOptions is the same and it will report only the function name which was on create.

I do not see how it can be done without cloning..

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Triggers/ServiceBusTriggerAttributeBindingProvider.cs
alrod: yes, this can be done.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusExtensionConfigProvider.cs
JoshLove-msft: I see.. I'm a little unsure about this change in general because we are changing the category for existing events.

If we think this change is important, we'd probably need to implement it by not mutating ProcessErrorAsync value on the options and instead pass in the handler as a new argument to ServiceBusListener constructor. Given that the event category already has ServiceBusListener in the name, and it matches what we do for all other events (using the type name from where it was logged), I'd suggest we hold off on this change until we get feedback that it is necessary. If I'm not mistaken in T1 we went from having a category of just Executor to function name, whereas in T2 we already have the extension-specific type.

alrod:
Closing the PR as we can correlate function name by using new logs for ServiceBus listener details.

PR #28202, opened 2022-04-14
upgrade autorest.csharp version and remove data-plane flag

chunyu3:

We will set 'data-plane' as autorest.csharp codegen default target. and will add new tag 'generation1-convenience-client' tag for the HLC Client Library.

In order to upgrade to the new autorest.csharp which will use 'data-plane' as default target, we need to migrate current SDKs in azure-sdk-for-net step by step as following:

1. Add new flag 'generation1-convenience-client' to all HLC SDKs
2. Merge change in autorest.csharp https://github.com/Azure/autorest.csharp/pull/2146
3. Upgrade autorest.csharp version and Remove old flag (data-plane) from the DPG SDKs

This pull request is #3

### DIFF
--- a/sdk/cognitivelanguage/Azure.AI.Language.QuestionAnswering/src/autorest.md
+++ b/sdk/cognitivelanguage/Azure.AI.Language.QuestionAnswering/src/autorest.md
@@ -16,7 +16,6 @@ batch:
 - input-file: https://raw.githubusercontent.com/Azure/azure-rest-api-specs/b791f57426508cb2793a8911650a416dcb11c6a6/specification/cognitiveservices/data-plane/Language/stable/2021-10-01/questionanswering-authoring.json
 # namespace: Azure.AI.Language.QuestionAnswering.Projects
   add-credentials: true
-  data-plane: true
 
 modelerfour:
   lenient-model-deduplication: true
--- a/eng/Packages.Data.props
+++ b/eng/Packages.Data.props
@@ -152,7 +152,7 @@
     All should have PrivateAssets="All" set so they don't become package dependencies
   -->
   <ItemGroup>
-    <PackageReference Update="Microsoft.Azure.AutoRest.CSharp" Version="3.0.0-beta.20220407.2" PrivateAssets="All" />
+    <PackageReference Update="Microsoft.Azure.AutoRest.CSharp" Version="3.0.0-beta.20220412.1" PrivateAssets="All" />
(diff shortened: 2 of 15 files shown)


christothes (approved):

sdk/cognitivelanguage/Azure.AI.Language.QuestionAnswering/src/autorest.md
```diff
@@ -16,7 +16,6 @@ batch:
 - input-file: https://raw.githubusercontent.com/Azure/azure-rest-api-specs/b791f57426508cb2793a8911650a416dcb11c6a6/specification/cognitiveservices/data-plane/Language/stable/2021-10-01/questionanswering-authoring.json
 # namespace: Azure.AI.Language.QuestionAnswering.Projects
   add-credentials: true
-  data-plane: true
```
heaths, line 19: Can you describe in your PR description what the effect of these changes should be instead of the standard external contributor template? For example, what was the intended impact of his directive given that there was seemingly no other effect on sources? And shy isn't the `generation1-convenience-client` directive added given that this swagger is meant to generate an LLC (the old style)?

I don't feel I have enough information to sign off given we're about to GA this SDK, and are about to release a beta of another one I don't see modified here that is also an older LLC (we'll use DPG for beta 4).

heaths (changes requested):

sdk/cognitivelanguage/Azure.AI.Language.QuestionAnswering/src/autorest.md
heaths: I'll take this offline to unblock the PR since there's no source changes, though I don't understand why there aren't any changes besides this.

heaths (approved):

sdk/cognitivelanguage/Azure.AI.Language.QuestionAnswering/src/autorest.md
chunyu3: We will change to set 'data-plane' as a default target in autorest.csharp codegen. And previous HLC library will use a new tag 'generation1-convenience-client'. So when 'azure-arm' is set, it will generate management plane, if 'generation1-convenience-client' is set, it will generate HLC, otherwise it will generate DPG(LLC). 

The issue for this is https://github.com/Azure/azure-sdk-for-net/issues/assigned/chunyu3

PR #28221, opened 2022-04-15
Update ServiceBus documentation to fix message batching

Ricky-G:
Receiving messages in a batch is not correctly documented so fixing this so there should be no confusion for developers.


sdk/servicebus/Azure.Messaging.ServiceBus/README.md
```diff
@@ -182,7 +182,7 @@ string body = receivedMessage.Body.ToString();
 Console.WriteLine(body);
 ```
 
-### Send and receive a batch of messages
+### Send a batch of messages
```
jsquire, line 185: We'll need to reflect the new sections in the list on L145, replacing the existing item.

jsquire, line 185: We'll also want to update the snippet name for sending on L189 so that it reflects that it is not receiving anything.

jsquire (commented):
Hi @Ricky-G.   Thank you for your contribution and helping to improve the Azure SDK experience.   I left a couple of minor suggestions, but the changes look great to me.  

I'm going to defer to @JoshLove-msft for additional feedback and merge decisions.

sdk/servicebus/Azure.Messaging.ServiceBus/README.md
```diff
@@ -238,6 +238,23 @@ messages.Add(new ServiceBusMessage("Second"));
 // send the messages
 await sender.SendMessagesAsync(messages);
 ```
+### Receive a batch of messages
+
+In order to receive a batch of messages from queue, topic or subscription:
+
+```C# Snippet:ReceiveBatchOfMessages
+// create a receiver that we can use to receive the messages
```
jsquire, line 246: I believe that we'll want to include context here for creating the client and where `queueName` comes from.    Our goal is to allow each snippet to be as self-contained as possible so that it can be copy/pasted without the need to search through the sample for the related parts.

sdk/servicebus/Azure.Messaging.ServiceBus/README.md
Ricky-G: Hi @jsquire thank you for reviewing the PR.  The background for this change; a couple of engineers on the customer side mentioned the heading said **"Send and receive a batch of messages"** but there was no actual code for batch receiving of messages in the markdown file.

What I have done is go back to the original markdown file and just include the missing code which is supposed to receive a batch of messages.

Ricky-G: Reverted to the original markdown file, just added the missing code which is supposed to receive a batch of messages.

sdk/servicebus/Azure.Messaging.ServiceBus/README.md
Ricky-G: Makes sense.  Reverted back to how it was with the old heading, just included the 3 missing lines that were supposed to show the batch receiving of messages.

sdk/servicebus/Azure.Messaging.ServiceBus/MigrationGuide.md
```diff
@@ -212,6 +212,20 @@ while (messages.Count > 0)
 
     // if there are any remaining messages in the .NET queue, the while loop repeats
 }
+
+// create a receiver that we can use to receive the messages
```
JoshLove-msft, line 216: I don't think we want this change in the migration guide. What we can do instead is define a new region for the receiving part and include both the sending and receiving snippets in the README, but leave only the sending snippet in the migration guide.

Ricky-G: hi @JoshLove-msft thank you for the comment. I have done as suggested, the migration guide remains untouched.  Introduced a new snippet and included that in the Readme.

JoshLove-msft: Excellent - I think all that is remaining is to update the table of contents of the README to reflect the new section and the updated name of the existing section - https://github.com/Azure/azure-sdk-for-net/pull/28221/files#diff-6b20936315df43a04f92031bb5622ac62ef10233db814c3e9aae4f89611bd5c2R145

sdk/servicebus/Azure.Messaging.ServiceBus/README.md
```diff
@@ -142,7 +142,8 @@ We guarantee that all client instance methods are thread-safe and independent of
 ## Examples
 
 * [Send and receive a message](#send-and-receive-a-message)
-* [Send and receive a batch of messages](#send-and-receive-a-batch-of-messages)
+* [Sending a batch of messages](#Sending-a-batch-of-messages)
```
JoshLove-msft, line 145: It looks like you need to lowercase the anchor tags based on the CI error.

sdk/servicebus/Azure.Messaging.ServiceBus/README.md
```diff
@@ -142,7 +142,8 @@ We guarantee that all client instance methods are thread-safe and independent of
...
+* [Receiving a batch of messages](#Receiving-a-batch-of-messages)
```
JoshLove-msft, line 146:
```suggestion
* [Receiving a batch of messages](#receiving-a-batch-of-messages)
```

sdk/servicebus/Azure.Messaging.ServiceBus/README.md
JoshLove-msft, line 145:
```suggestion
* [Sending a batch of messages](#sending-a-batch-of-messages)
```

JoshLove-msft (approved):
Thanks again for the contribution @Ricky-G

PR #28225, opened 2022-04-15
Review work on Feature/text analytics language api

maririos:
The purpose of this PR is to review the progress done in the feature/text-analytics-language-api branch with no intentions of merging the branch to master.


sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.cs
```diff
@@ -32,6 +32,22 @@ internal static TextAnalyticsError ConvertToError(TextAnalyticsErrorInternal err
             return new TextAnalyticsError(errorCode, message, target);
         }
 
+        internal static TextAnalyticsError ConvertToError(DocumentError error)
```
maririos, line 35: From this code, it looks like you are only using the `error.Error` property, so instead of passing the whole `DocumentError` object, consider sending to this function just `DocumentError.Error`.

Context: `DocumentError` has 2 main properties: the `id` of the document and the `error` itself. This function is already called per each document, which is why you never use `id` in here.

sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.cs
```diff
@@ -73,6 +89,17 @@ internal static DetectedLanguage ConvertToDetectedLanguage(DocumentLanguage docu
             return new DetectedLanguage(documentLanguage.DetectedLanguage, ConvertToWarnings(documentLanguage.Warnings));
         }
 
+        // CHANGED: added overload for New Swagger
+        internal static DetectedLanguage ConvertToDetectedLanguage(LanguageDetectionDocumentResult documentLanguage)
+        {
+            List<TextAnalyticsWarning> warnings = new List<TextAnalyticsWarning>();
+            foreach (var warning in documentLanguage.Warnings)
```
maririos, line 96: If the service doesn't return `Warnings`, `document.Language.Warnings` is going to be null and the code here is going to throw.

Because this check and loop happen in every endpoint, we extracted that logic into the [ConvertToWarnings](https://github.com/Azure/azure-sdk-for-net/blob/2c9bf3756510134d789bf86e85c3ce7ab013b6d5/sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.cs#L67) method.
We should do the same or similar for new code too.

sdk/textanalytics/Azure.AI.TextAnalytics/src/Generated/Models/DetectedLanguageInternal.cs
```diff
@@ -14,7 +14,7 @@ namespace Azure.AI.TextAnalytics.Models
         /// <param name="name"> Long name of a detected language (e.g. English, French). </param>
         /// <param name="iso6391Name"> A two letter representation of the detected language according to the ISO 639-1 standard (e.g. en, fr). </param>
         /// <param name="confidenceScore"> A confidence score between 0 and 1. Scores close to 1 indicate 100% certainty that the identified language is true. </param>
-        internal DetectedLanguageInternal(string name, string iso6391Name, double confidenceScore)
+        public DetectedLanguageInternal(string name, string iso6391Name, double confidenceScore)
```
maririos, line 17: curious, why the need to make this class public? also, did you make the class `public` directly here? or in another file that I missed?

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
```diff
@@ -19,7 +20,8 @@ namespace Azure.AI.TextAnalytics
     public class TextAnalyticsClient
     {
         private readonly Uri _baseUri;
-        internal readonly TextAnalyticsRestClient _serviceRestClient;
+        internal readonly AnalyzeTextRestClient _serviceRestClient;
+        internal readonly MicrosoftCognitiveLanguageServiceRestClient _cognitiveRestClient;
```
maririos, line 24: what is the difference between these two? I don't see the `MicrosoftCognitiveLanguageServiceRestClient` referenced here

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
```diff
@@ -65,7 +67,7 @@ public TextAnalyticsClient(Uri endpoint, TokenCredential credential, TextAnalyti
             _options = options;
 
             var pipeline = HttpPipelineBuilder.Build(options, new BearerTokenAuthenticationPolicy(credential, defaultScope));
-            _serviceRestClient = new TextAnalyticsRestClient(_clientDiagnostics, pipeline, endpoint.AbsoluteUri, TextAnalyticsClientOptions.GetVersionString(options.Version));
+            _serviceRestClient = new AnalyzeTextRestClient(_clientDiagnostics, pipeline, endpoint.AbsoluteUri, TextAnalyticsClientOptions.GetVersionString(options.Version));
```
maririos, line 70: we need to instantiate `_cognitiveRestClient` too otherwise the calls are not going to get executed

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
```diff
@@ -195,17 +203,26 @@ public virtual Response<DetectedLanguage> DetectLanguage(string document, string
...
+                {
+                    input.Documents.Add(doc);
+                }
+                var analyzeLanguageDetection = new AnalyzeTextLanguageDetectionInput { AnalysisInput = input };
+                Response<AnalyzeTextTaskResult> result = _cognitiveRestClient.AnalyzeText(analyzeLanguageDetection, cancellationToken: cancellationToken);
 
-                if (result.Value.Errors.Count > 0)
+                var languageDetection = result.Value as LanguageDetectionTaskResult;
+                Debug.Assert(languageDetection != null);
+                Response response = result.GetRawResponse();
```
maririos, line 216: need to cleanup these 2 lines

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
```diff
@@ -195,17 +203,26 @@ public virtual Response<DetectedLanguage> DetectLanguage(string document, string
...
+                var input = new LanguageDetectionAnalysisInput();
+                foreach (var doc in documents)
+                {
+                    input.Documents.Add(doc);
+                }
+                var analyzeLanguageDetection = new AnalyzeTextLanguageDetectionInput { AnalysisInput = input };
+                Response<AnalyzeTextTaskResult> result = _cognitiveRestClient.AnalyzeText(analyzeLanguageDetection, cancellationToken: cancellationToken);
 
-                if (result.Value.Errors.Count > 0)
+                var languageDetection = result.Value as LanguageDetectionTaskResult;
```
maririos, line 214: before you were doing a [cast expression](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/type-testing-and-cast#cast-expression) `var languageDetection = (LanguageDetectionTaskResult)result.Value;` and here you are using the `as` [operator](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/type-testing-and-cast#as-operator).

which do you think should be the patter we should use everywhere?

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
```diff
@@ -364,6 +377,33 @@ private async Task<Response<DetectLanguageResultCollection>> DetectLanguageBatch
...
+        private async Task<Response<DetectLanguageResultCollection>> DetectLanguageBatchAsync(LanguageDetectionAnalysisInput batchInput, TextAnalyticsRequestOptions options, CancellationToken cancellationToken)
+        {
+            using DiagnosticScope scope = _clientDiagnostics.CreateScope($"{nameof(TextAnalyticsClient)}.{nameof(DetectLanguageBatch)}");
+            scope.Start();
+
+            try
+            {
+                var analyzeLanguageDetection = new AnalyzeTextLanguageDetectionInput { AnalysisInput = batchInput };
+                analyzeLanguageDetection.Parameters.ModelVersion = options.ModelVersion;
+                analyzeLanguageDetection.Parameters.LoggingOptOut = options.DisableServiceLogs;
```
maririos, line 390: curious why the `Parameters` are treated differently than `AnalysisInput`

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
```diff
@@ -195,17 +203,26 @@ public virtual Response<DetectedLanguage> DetectLanguage(string document, string
...
+                foreach (var doc in documents)
+                {
+                    input.Documents.Add(doc);
+                }
+                var analyzeLanguageDetection = new AnalyzeTextLanguageDetectionInput { AnalysisInput = input };
+                Response<AnalyzeTextTaskResult> result = _cognitiveRestClient.AnalyzeText(analyzeLanguageDetection, cancellationToken: cancellationToken);
 
-                if (result.Value.Errors.Count > 0)
+                var languageDetection = result.Value as LanguageDetectionTaskResult;
+                Debug.Assert(languageDetection != null);
```
maririos, line 215: need to clean up this line

sdk/textanalytics/Azure.AI.TextAnalytics/src/Azure.AI.TextAnalytics.csproj
```diff
@@ -24,4 +24,8 @@
     <Compile Include="$(AzureCoreSharedSources)PageResponseEnumerator.cs" LinkBase="Shared" />
     <Compile Include="$(AzureCoreSharedSources)AzureKeyCredentialPolicy.cs" LinkBase="Shared" />
   </ItemGroup>
+
+  <ItemGroup>
+    <Folder Include="Models\" />
```
maririos, line 29: curious what this is for?

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
maririos: actually, I am unclear what the difference is between the two. Could you explain when to use which?

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
```diff
@@ -241,11 +258,12 @@ public virtual Response<DetectedLanguage> DetectLanguage(string document, string
         /// status code.</exception>
         public virtual async Task<Response<DetectLanguageResultCollection>> DetectLanguageBatchAsync(IEnumerable<string> documents, string countryHint = default, TextAnalyticsRequestOptions options = default, CancellationToken cancellationToken = default)
         {
-            Argument.AssertNotNullOrEmpty(documents, nameof(documents));
-            options ??= new TextAnalyticsRequestOptions();
-            LanguageBatchInput detectLanguageInputs = ConvertToLanguageInputs(documents, countryHint);
-
-            return await DetectLanguageBatchAsync(detectLanguageInputs, options, cancellationToken).ConfigureAwait(false);
+            //Argument.AssertNotNullOrEmpty(documents, nameof(documents));
```
maririos, line 261: curious, why did you have to comment these lines out? I thought this path was already working

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
```diff
@@ -637,31 +674,37 @@ public virtual async Task<Response<RecognizeEntitiesResultCollection>> Recognize
...
-                Response<EntitiesResult> result = await _serviceRestClient.EntitiesRecognitionGeneralAsync(
-                    batchInput,
-                    options.ModelVersion,
-                    options.IncludeStatistics,
-                    options.DisableServiceLogs,
-                    Constants.DefaultStringIndexType,
-                    cancellationToken).ConfigureAwait(false);
-                var response = result.GetRawResponse();
+                var input = entityRecognitionInput.AnalysisInput;
+                var analyzeRecognizeEntities = new AnalyzeTextEntityRecognitionInput { AnalysisInput = input };
```
maririos, line 695: could we eliminate `input` and pass the value directly?

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
```diff
@@ -429,21 +449,26 @@ public virtual async Task<Response<CategorizedEntityCollection>> RecognizeEntiti
             try
             {
                 var documents = new List<MultiLanguageInput>() { ConvertToMultiLanguageInput(document, language) };
-
-                Response<EntitiesResult> result = await _serviceRestClient.EntitiesRecognitionGeneralAsync(
-                    new MultiLanguageBatchInput(documents),
-                    stringIndexType: Constants.DefaultStringIndexType,
```
maririos, line 435: this parameter is no longer passed. Could you please add it again?

maririos: I see this one is missing in other places too. 
`StringIndexType` is what tells the service what type of encoding they should use when managing strings. (think for example when someone sends an emoji. According to the encoding, the properties of a string will change).
By default, .NET manages `Utf16codeUnit` so we always need to send this property to the service. See how the library has a default value [here](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/textanalytics/Azure.AI.TextAnalytics/src/Constants.cs#L10).

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
```diff
@@ -744,26 +791,26 @@ public virtual async Task<Response<PiiEntityCollection>> RecognizePiiEntitiesAsy
...
+                var input = new MultiLanguageAnalysisInput();
+                foreach (var doc in documents)
+                {
+                    input.Documents.Add(doc);
+                }
+                var analyzePiiEntites = new AnalyzeTextPiiEntitiesRecognitionInput { AnalysisInput = input };
+                Response<AnalyzeTextTaskResult> result = await _cognitiveRestClient.AnalyzeTextAsync(analyzePiiEntites, cancellationToken: cancellationToken).ConfigureAwait(false);
 
-                Response<PiiEntitiesResult> result = await _serviceRestClient.EntitiesRecognitionPiiAsync(
-                    new MultiLanguageBatchInput(documents),
```
maririos, line 749: We need to make sure we send these properties too

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsOperationStatus.cs
```diff
@@ -11,5 +11,8 @@ namespace Azure.AI.TextAnalytics
     [CodeGenModel("State")]
     public partial struct TextAnalyticsOperationStatus
     {
+        /// <summary> Added for compilation. </summary>
+        /// TODO: delete later
+        public static TextAnalyticsOperationStatus Rejected { get; } = new TextAnalyticsOperationStatus(CancellingValue);
```
maririos, line 16: could u add an issue for this so we don't forget to clean it up? it needs to happen during this release cycle

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClientDiagnostics.cs
```diff
@@ -32,28 +33,29 @@ public TextAnalyticsClientDiagnostics(ClientOptions options) : base(options)
             ResponseHeaders responseHeaders,
             ref IDictionary<string, string>? additionalInfo)
         {
-            if (!string.IsNullOrEmpty(content))
```
maririos, line 35: curious about why was this causing a problem? 
can you create an issue to undo this?

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsOperationStatus.cs
nisha-bhatia: Issue: https://github.com/Azure/azure-sdk-for-net/issues/28326

sdk/textanalytics/Azure.AI.TextAnalytics/src/Generated/Models/DetectedLanguageInternal.cs
nisha-bhatia: I think this should be reverted

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsOperationStatus.cs
nisha-bhatia: https://github.com/Azure/azure-sdk-for-net/issues/28326

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
nisha-bhatia: I updated the Parameter construction to match the AnalysisInput

sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.cs
nisha-bhatia: added a ConvertToWarnings method

sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.cs
nisha-bhatia: updated

sdk/textanalytics/Azure.AI.TextAnalytics/src/Azure.AI.TextAnalytics.csproj
nisha-bhatia: I'm not sure how this got updated

maririos: Jesse fixed this. resolving comment

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClientDiagnostics.cs
maririos: Fixed in PR https://github.com/Azure/azure-sdk-for-net/pull/28449

PR #28229, opened 2022-04-16
[Text Analytics] Legacy Generation and Abstractions

jsquire:
# Summary

The focus of these changes is to add the generated surface for the legacy Text Analytics REST API in a stand-alone project that is compatible with the AutoRest generator and its conventions.  The types for the legacy API are generated under a unique namespace and will be included in the Text Analytics  client library via source linking.  This is intended to allow the legacy API to be managed and regenerated in a standard way while avoiding conflicts with the Cognitive Language API that will be used for service operations on new version going forward.

Also included are the stubs for the `ServiceClient` abstractions, forming the basis of the contract by which the `TextAnalyticsClient` will be able to target different REST services through a common interface.

## Next Steps

- Link the legacy API surface into the project
- Implement the legacy service client
- Update TextAnalyticsClient to delegate to the correct service client
- Shift Nisha's work over to the language service client


sdk/textanalytics/Azure.AI.TextAnalytics.Legacy.Shared/src/GeneratorStubs/InternalTypes.cs
```diff
@@ -0,0 +1,38 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Azure.Core;
+
+namespace Azure.AI.TextAnalytics.Legacy.Models
```
jsquire, line 6: There is precedent for this pattern in Search and Storage.

sdk/textanalytics/Azure.AI.TextAnalytics.Legacy.Shared/src/autorest.md
```diff
@@ -0,0 +1,244 @@
...
+  from: swagger-document
+  where: $.operations.*
+  transform: >
+    $["x-namespace"] = "Azure.AI.TextAnalytics.Legacy"
+
+```
+### Make types internal
+
+``` yaml
+directive:
```
jsquire, line 42: I went with the expanded form after discussing with Heath; functionally, its the same as having 3 separate sections.

sdk/textanalytics/Azure.AI.TextAnalytics.Legacy.Shared/src/autorest.md
```diff
@@ -0,0 +1,244 @@
...
+- from: swagger-document
+  where: $.definitions.*
+  transform: >
+    $["x-accessibility"] = "internal"
+- from: swagger-document
+  where: $.operations.*
+  transform: >
+    $["x-accessibility"] = "internal"
+- from: swagger-document
+  where: $.definitions..properties.*
```
jsquire, line 52: @maririos:  I tried your enum trick in this form and in the expanded form;  unfortunately, I was still getting `public` types emitted.

sdk/textanalytics/Azure.AI.TextAnalytics.Legacy.Shared/src/Custom/DocumentSentimentValue.cs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Azure.Core;
+
```
jsquire, line 5: I was unable to get the generator transforms to prevent emitting the parameter-related enumerations as public.   @maririos had some suggestions that I was unable to get to work;  I'll be opening an issue for us to revisit this and see if we can get the generator to play nice.

For now, I believe this will suffice as we know there will not be any large changes to the REST API going forward.

sdk/textanalytics/Azure.AI.TextAnalytics.Legacy.Shared/src/Custom/JobManifestTasks.cs
```diff
@@ -0,0 +1,60 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Collections.Generic;
+using Azure.Core;
+
+namespace Azure.AI.TextAnalytics.Legacy
+{
```
jsquire, line 8: I absorbed the content for things in the `Custom` area from @nisha-bhatia's work.  I'm not entirely sure that we need the extra code; I was able to get the legacy bits to generate and compile without it, but I left it for now under the assumption that we'll need it as we start building out the service clients.   If that turns out to be incorrect, I'll look at trimming things back then.

sdk/textanalytics/Azure.AI.TextAnalytics.Legacy.Shared/src/GeneratorStubs/InternalTypes.cs
jsquire: These types will not need to be linked into the core Text Analytics client library - they exist only to influence code generation.

sdk/textanalytics/Azure.AI.TextAnalytics.Legacy.Shared/src/Azure.AI.TextAnalytics.Legacy.csproj
```diff
@@ -0,0 +1,34 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
jsquire, line 1: To make sure that things play nice with the generator, I used a standard project here, rather than using a "shared code" type project.   This allows us to also test compilation of the generated surface independently.   

The only downstream impact is that we'll need to do source linking via glob pattern rather than having a single ready-made package of code to import.

jsquire:
CI failures are a hold-over from the conversion-to-language-service efforts being in progress; I confirmed that the legacy generated code compiles and that we can regenerate.   Likewise, the service client stubs have been confirmed as compilable.

maririos (approved):
Forgot to submit my review :(

sdk/textanalytics/Azure.AI.TextAnalytics.Legacy.Shared/src/Azure.AI.TextAnalytics.Legacy.csproj
jsquire: There won't be a shared class between them - at least not in the sense of partials or true code sharing.   There may be custom types that are used by both, such as the `Transformer`, but there should not be anything that is generated that crosses between the two layers.   Editing things shouldn't require any special steps.

sdk/textanalytics/Azure.AI.TextAnalytics.Legacy.Shared/src/Custom/JobManifestTasks.cs
jsquire: Yup; that's the scenario.   Nothing public, just overrides and extensions for generated types that were in the `Internal` folder.

Issue #27138, opened 2022-02-22
[QUERY]How to replace IMessageSession.ReceiveAsync() in new Azure.Messaging.ServiceBus?


caiyiJu:
### Library name and version

Microsoft.Azure.WebJobs.Extensions.ServiceBus 5.2.0

### Query/Question

Hi,
When trying to update code that use Microsoft.Azure.WebJobs.Extensions.ServiceBus" Version="4.1.1" to new 5.2.0. There are some differences.

code when using version 4.1.1
```csharp
public async static Task RunAsync([ServiceBusTrigger("%queuename%", Connection = "ServiceBusStorageConnectionString", IsSessionsEnabled = true)]
        Message message, IMessageSession messageSession, 
            ILogger log)
        {
            var msgs=await messageSession.ReceiveAsync(10);
```

code when using version 5.2.0

```csharp
public async static Task RunAsync([ServiceBusTrigger("%queuename%", Connection = "ServiceBusStorageConnectionString", IsSessionsEnabled = true)]
        ServiceBusReceivedMessage message, ServiceBusSessionMessageActions messageSession,
            ILogger log)
```

How to call: messageSession.ReceiveAsync() now?

### Environment

Net core 3.1

jsquire:
Thank you for your feedback.  Tagging and routing to the team member best able to assist.

caiyiJu:
Hello team, could we have an update for this?

JoshLove-msft:
@caiyiJu, can you explain why you want to receive another message from your function? The function already executes whenever a message is received.

caiyiJu:
@JoshLove-msft As batch and concurrent session cannot be used together per [the documentation](https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-service-bus). Will it be more efficient if we receive another message at the same time? By the way, could you explain why the method is removed if there is no such method in the new version? This will also be helpful to understand the usage.
<html>
<body>


maxConcurrentSessions | 8 | The maximum number of sessions that can be handled concurrently per scaled instance. This setting only applies for functions that receive a single message at a time.
-- | -- | --



</body>
</html>

JoshLove-msft:
I think using the function in that way to receive additional messages is a workaround for what you would really want - session concurrency when using batch receive, right? If that's the case, I'd rather attempt to address this by extending maxConcurrentSessions to work for functions that take an array of messages.

caiyiJu:
[...]

Issue #22083, opened 2021-06-23
[FEATURE REQ] Service bus ReceiveDeferredMessageAsync from a processor


shlomiassaf:
**Library or service name.**
Azure.Messaging.ServiceBus
 
**Is your feature request related to a problem? Please describe.**
Currently, receiving of a deferred message can only be done through a receiver.  

Since `ServiceBusProcessor` encapsulate the receiver, the only way to receive a deferred message is by creating
a duplicate receiver instance.

I understand that the processor is not a receiver and thus might want to encapsulate the receiver but it might be
that defer operation logic does exist within the processor and perhaps it is possible only to expose a method
that will allow receiving deferred message through the internal receiver.
The processor args already allow deferring, so it seems logical (to me) that it will also allow receiving these deferred message later.

Thanks!

jsquire:
Thank you for your feedback.  Tagging and routing to the team member best able to assist.

JoshLove-msft:
So if we were to expose receiving deferred messages in the args, you would only be able to do so within the message handler when you receive a different message, which doesn't seem like an intuitive flow. Can you tell us more about when you would need to receive deferred messages when using the processor?

JoshLove-msft:
Related https://github.com/Azure/azure-sdk-for-net/issues/22154

shlomiassaf:
Well, we've implemented a small library on top of `Azure.Messaging.ServiceBus`.

Basically we've added plugins and pipelines, wrapped with container support for DI.
We've made each bus message closer to an incoming HTTP request in a server.

We have a backoff retry mechanism which, based on certain exceptions, emits the message back to service bus with a schedule time and a label to follow on the count.

We've used to do it via message cloning, which OK but not optimal (as we see it)

- Its not transactional as 2 calls are required (complete current, submit new).
- It's a bit verbose (cloning mostly)
- It requires a special, short-circuiting filter rule, on subscriptions so we only retry on the relevant subscription within the topic. And ofcourse submitting the clone with a header-flag for that filter.

The filtering issue has the biggest impact since rule filters are "OR" but we must short-circuit on retry submit so we need to join filters into one expression.  
Of course, filters comes with a performance hit... 

All of the above is solvable.

- We can use a transaction scope  
- We can schedule and submit to a designated queue with a "response" address which will have a handler that will forward it to...

Lot's of solution... but all has tradeoff.

The best one, as we see it, is using the defer mechanizem.
[...]

PR #28230, opened 2022-04-17
Support receiving messages from processor

JoshLove-msft:
Addresses https://github.com/Azure/azure-sdk-for-net/issues/27138 and https://github.com/Azure/azure-sdk-for-net/issues/22083


sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpReceiver.cs
```diff
@@ -1034,6 +1034,10 @@ private void ThrowLockLostException()
             {
                 DateTime[] lockedUntilUtcTimes = amqpResponseMessage.GetValue<DateTime[]>(ManagementConstants.Properties.Expirations);
                 lockedUntil = lockedUntilUtcTimes[0];
+                if (_requestResponseLockedMessages.Contains(lockToken))
```
JoshLove-msft, line 1037: Ran into this while adding new tests - we were not updating the receive deferred lockedUntil time so completing these messages would not work properly if the initial lock time was passed (they would get removed from requestResponseLockedMessages ConcurrentExpiringSet).

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpReceiver.cs
```diff
@@ -1075,9 +1079,8 @@ internal async Task<DateTimeOffset> RenewSessionLockInternal(TimeSpan timeout)
 
             if (!_receiveLink.TryGetOpenedObject(out ReceivingAmqpLink receiveLink))
             {
-                receiveLink = await _receiveLink.GetOrCreateAsync(UseMinimum(_connectionScope.SessionTimeout, timeout)).ConfigureAwait(false);
```
JoshLove-msft, line 1078: We should not attempt to reopen a closed session link as this can cause an inconsistent error experience, i.e. you would get an AcceptSession error rather than SessionLockLost.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusSessionProcessor.cs
```diff
@@ -145,6 +145,7 @@ protected ServiceBusSessionProcessor()
         protected internal virtual async Task OnProcessSessionMessageAsync(ProcessSessionMessageEventArgs args)
         {
             await InnerProcessor.OnProcessSessionMessageAsync(args).ConfigureAwait(false);
+            args.EndExecutionScope();
```
jsquire, line 148: Should we consider putting this in a finally block so that we don't potentially miss it in the case of an exception?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Receiver/ServiceBusReceiver.cs
```diff
@@ -498,14 +498,16 @@ public virtual async Task CloseAsync(CancellationToken cancellationToken = defau
...
-        /// </exception>
-        /// <exception cref="ServiceBusException">
-        ///   The lock for the session has expired or the message has already been completed. This only applies for session-enabled
-        ///   entities.
-        ///   The <see cref="ServiceBusException.Reason" /> will be set to <see cref="ServiceBusFailureReason.SessionLockLost"/> in this case.
+        ///   <list type="bullet">
+        ///     <item>
+        ///       The lock for the message has expired or the message has already been completed. This does not apply for session-enabled entities.
+        ///       The <see cref="ServiceBusException.Reason" /> will be set to <see cref="ServiceBusFailureReason.MessageLockLost"/> in this case.
+        ///     </item>
```
jsquire, line 505:
```suggestion
        ///     <item>
        ///        <description>
        ///          The lock for the message has expired or the message has already been completed. This does not apply for session-enabled entities.
        ///          The <see cref="ServiceBusException.Reason" /> will be set to <see cref="ServiceBusFailureReason.MessageLockLost"/> in this case.
        ///        </description> 
        ///     </item>
```

There needs to be a `<description>` element wrapping the text for a list tiem to be valid.   MS Docs won't render these if the `<description>` is missing.    _([ref](https://docs.microsoft.com/dotnet/csharp/language-reference/language-specification/documentation-comments#d37-list))_

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Processor/ProcessorLiveTests.cs
```diff
@@ -236,7 +239,7 @@ async Task ProcessMessage(ProcessMessageEventArgs args)
                 }
                 await Task.WhenAll(completionSources.Select(source => source.Task));
                 Assert.IsTrue(processor.IsProcessing);
-                await processor.StopProcessingAsync();
+                await processor.CloseAsync();
```
jsquire, line 242: This seems odd;  don't we normally expect the previous pattern in customer code?    Just calling stop and then close/dispose when the application is exiting?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ProcessMessageEventArgs.cs
```diff
@@ -36,11 +45,38 @@ public class ProcessMessageEventArgs : EventArgs
...
+        /// in the event that <see cref="ServiceBusProcessor.StopProcessingAsync"/> is called.
+        /// </param>
+        internal ProcessMessageEventArgs(
+            ServiceBusReceivedMessage message,
+            ReceiverManager manager,
+            CancellationToken cancellationToken)
+        {
+            Message = message;
+            Messages[message] = default;
+            _manager = manager;
```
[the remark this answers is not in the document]
JoshLove-msft: The manager's unmanaged objects would still get cleaned up. GC doesn't trigger the clean up in any case.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessor.cs
```diff
@@ -520,6 +520,7 @@ protected ServiceBusProcessor()
         protected internal virtual async Task OnProcessMessageAsync(ProcessMessageEventArgs args)
         {
             await _processMessageAsync(args).ConfigureAwait(false);
+            args.EndExecutionScope();
```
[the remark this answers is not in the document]
JoshLove-msft: good catch

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Processor/ProcessorLiveTests.cs
JoshLove-msft: Yeah, I may have went overboard with the replacements here. I updated a few tests where we were checking that all of the messages were gone from the queue because otherwise in the case of sessions, the session would remain locked when just calling StopProcessingAsync.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ProcessMessageEventArgs.cs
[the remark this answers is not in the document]
jsquire: > GC doesn't trigger the clean up in any case.

Except in the case of finalizers, which I believe are used in several places in the network areas of the BCL.   That said, I was more thinking about the wasted memory in this case.   I don't have a strong concern.

jsquire (approved):

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ProcessMessageEventArgs.cs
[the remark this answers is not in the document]
JoshLove-msft: Ah, I see your point - the explicit cleanup that we are doing in the processor will still happen regardless of whether the user is holding on to a reference, so it shouldn't matter whether or not the finalizer runs, at least I don't think it would.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ProcessMessageEventArgs.cs
```diff
@@ -117,5 +153,107 @@ await _receiver.AbandonMessageAsync(message, propertiesToModify, cancellationTok
...
+                foreach (ServiceBusReceivedMessage message in messages)
+                {
+                    Messages[message] = default;
+                }
+            }
+
+            return messages;
+        }
+
+        internal void EndExecutionScope()
```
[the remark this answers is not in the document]
JoshLove-msft: To be honest, I wish we had implemented this for all methods in the args to make the contract more explicit. It wasn't strictly necessary to add here, but I think it can prevent users from shooting themselves in the foot.

PR #28232, opened 2022-04-17
Fixing and Re-recording CLU live tests

ZulaMostafa:


sdk/cognitivelanguage/test-resources.json
```diff
@@ -23,42 +23,42 @@
...
+        // },
+        // "AZURE_CONVERSATIONS_WORKFLOW_PROJECT_NAME": {
+        //     "type": "string",
+        //     "value": "antischTwo"
+        // },
+        // "AZURE_CONVERSATIONS_WORKFLOW_DEPLOYMENT_NAME": {
+        //     "type": "string",
+        //     "value": "production"
+        // },
+        // "QUESTIONANSWERING_ENDPOINT": {
```
heaths, line 51: Even though you're commenting this out, use the correct "AZURE_" prefix. Make sure the rest line up. That way, if/when we do restore these, they are already correct and there's no confusion.

heaths (approved):
Before I merge, were these recorded with the local env vars that @mshaban-msft passed along internally? I updated our shared Key Vault this afternoon using those values except for `qna-key` and `qna-uri` but, as we're discussing in internally, maybe we should - expecting that QA live tests would then fail (but not important right now).

sdk/cognitivelanguage/test-resources.json
ZulaMostafa: Yes, they were recorded with the vars that Mohammed passed.

heaths (approved):

heaths:
Be sure eventually to update _QuestionAnsweringTestEnvironment.cs_ with the new env vars. Technically, these should still work since the base `TestEnvironment` class falls back to prepending "AZURE_", but while we're busy fixing up QA anyway, let's get everything in sync for easier troubleshooting, should the need ever arise.

PR #28236, opened 2022-04-18
Don't save recordings for failed tests

JoshLove-msft:
Also add back SaveDebugRecordingsOnFailure that @heaths added prior to Test Proxy migration, now that there is a way to control whether or not to save in Test Proxy.

### DIFF
--- a/sdk/core/Azure.Core.TestFramework/src/Generated/TestProxyRestClient.cs
+++ b/sdk/core/Azure.Core.TestFramework/src/Generated/TestProxyRestClient.cs
@@ -239,7 +239,7 @@ namespace Azure.Core.TestFramework
             }
         }
 
-        internal HttpMessage CreateStopRecordRequest(string xRecordingId, IDictionary<string, string> variables)
+        internal HttpMessage CreateStopRecordRequest(string xRecordingId, IDictionary<string, string> variables, string xRecordingSkip)
         {
             var message = _pipeline.CreateMessage();
             var request = message.Request;
@@ -249,6 +249,10 @@ namespace Azure.Core.TestFramework
             uri.AppendPath("/record/stop", false);
             request.Uri = uri;
             request.Headers.Add("x-recording-id", xRecordingId);
+            if (xRecordingSkip != null)
+            {
+                request.Headers.Add("x-recording-skip", xRecordingSkip);
+            }
             request.Headers.Add("Content-Type", "application/json");
             var content = new Utf8JsonRequestContent();
             content.JsonWriter.WriteStartObject();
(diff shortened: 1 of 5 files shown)


sdk/core/Azure.Core.TestFramework/src/RecordedTestBase.cs
```diff
@@ -97,6 +97,28 @@ public abstract class RecordedTestBase : ClientTestBase
         /// </summary>
         public List<(string Header, string QueryParameter)> SanitizedQueryParametersInHeaders { get; } = new();
 
+        /// <summary>
+        /// Flag you can (temporarily) enable to save failed test recordings
+        /// and debug/re-run at the point of failure without re-running
+        /// potentially lengthy live tests.  This should never be checked in
+        /// and will throw an exception from CI builds to help make that easier
+        /// to spot.
+        /// The list of JSON path sanitizers to use when sanitizing a JSON request or response body.
```
christothes, line 106: Is this last comment relevant to this setting? If so, could it be clarified?

sdk/core/Azure.Core.TestFramework/src/Generated/TestProxyRestClient.cs
```diff
@@ -249,6 +249,10 @@ internal HttpMessage CreateStopRecordRequest(string xRecordingId, IDictionary<st
             uri.AppendPath("/record/stop", false);
             request.Uri = uri;
             request.Headers.Add("x-recording-id", xRecordingId);
+            if (xRecordingSkip != null)
+            {
+                request.Headers.Add("x-recording-skip", xRecordingSkip);
```
christothes, line 254: What values would be valid for `xRecordingSkip`? If it's just a `bool`, should we make it a `bool` and `ToString` it here?

sdk/core/Azure.Core.TestFramework/src/Generated/TestProxyRestClient.cs
```diff
@@ -265,9 +269,10 @@ internal HttpMessage CreateStopRecordRequest(string xRecordingId, IDictionary<st
         /// <summary> Stop recording for a test. </summary>
         /// <param name="xRecordingId"> The recording ID. </param>
         /// <param name="variables"> Variables for the recording. </param>
+        /// <param name="xRecordingSkip"> Set to request-response to skip recording this session. </param>
         /// <param name="cancellationToken"> The cancellation token to use. </param>
         /// <exception cref="ArgumentNullException"> <paramref name="xRecordingId"/> or <paramref name="variables"/> is null. </exception>
-        public async Task<Response> StopRecordAsync(string xRecordingId, IDictionary<string, string> variables, CancellationToken cancellationToken = default)
+        public async Task<Response> StopRecordAsync(string xRecordingId, IDictionary<string, string> variables, string xRecordingSkip = null, CancellationToken cancellationToken = default)
```
christothes, line 275: Should `xRecordingSkip` be a bool?

christothes, line 275: Is this new argument plumbed through to `TestRecording` somewhere?

sdk/core/Azure.Core.TestFramework/src/RecordedTestBase.cs
JoshLove-msft: Woops will remove.

sdk/core/Azure.Core.TestFramework/src/Generated/TestProxyRestClient.cs
JoshLove-msft: It could be a bool, but the Test Proxy is expecting "request-response" as the value, and this is generated code.

sdk/core/Azure.Core.TestFramework/src/Generated/TestProxyRestClient.cs
JoshLove-msft: Test Proxy expects "request-response" as the value here.

JoshLove-msft: It is passed to this method from DisposeAsync in TestRecording.

christothes (approved):

PR #28238, opened 2022-04-18
add eula/privacy statement to gallery SDK release

kangsun-ctrl:


sdk/compute/Microsoft.Azure.Management.Compute/src/Generated/Models/Gallery.cs
```diff
@@ -74,15 +75,11 @@ public Gallery(string location, string id = default(string), string name = defau
...
+        /// Gets or sets possible values include: 'Creating', 'Updating',
+        /// 'Failed', 'Succeeded', 'Deleting', 'Migrating'
         /// </summary>
-        /// <remarks>
-        /// The provisioning state, which only appears in the response.
-        /// Possible values include: 'Creating', 'Updating', 'Failed',
-        /// 'Succeeded', 'Deleting', 'Migrating'
-        /// </remarks>
         [JsonProperty(PropertyName = "properties.provisioningState")]
-        public string ProvisioningState { get; private set; }
```
Sandido, line 85: This removal of Private seems incorrect. The Swagger [PR ](https://github.com/Azure/azure-rest-api-specs/pull/18680/files#diff-2861a3a31f910fbcb3cbc0af6f313527001ff65e8360cfcc566ae280a27b8bf9R2887)is not removing the ReadOnly: True tag, so this should still be ReadOnly.

Sandido: This issue occurs in multiple files in this PR.

Sandido (changes requested):

sdk/compute/Microsoft.Azure.Management.Compute/src/Generated/Models/Gallery.cs
Sandido: Might need to keep the ReadOnly:True tag directly in the ProvisioningState property instead of the model. Try that.

kangsun-ctrl: this is resolved

Yao725 (approved):
I'm ok with these changes, do we need to merge this PR now?

Sandido (approved):

sdk/compute/Microsoft.Azure.Management.Compute/src/Generated/Models/CommunityGalleryImage.cs
```diff
@@ -71,6 +75,8 @@ public CommunityGalleryImage(OperatingSystemTypes osType, OperatingSystemStateTy
             Features = features;
             PurchasePlan = purchasePlan;
             Architecture = architecture;
+            PrivacyStatementUri = privacyStatementUri;
```
Sandido, line 78: Where are the .Net SDK test(s) for this feature?

Sandido (changes requested):

sdk/compute/Microsoft.Azure.Management.Compute/src/Generated/Models/CommunityGalleryImage.cs
kangsun-ctrl: test added, also add new recorded test

sdk/compute/Microsoft.Azure.Management.Compute/tests/ScenarioTests/CommunityGalleryTests.cs
```diff
@@ -105,6 +105,8 @@ private void ValidateCommunityGalleryImage(CommunityGalleryImage communityGaller
         {
             string expectedId = "/CommunityGalleries/" + PublicGalleryName + "/Images/" + GalleryImageName;
             Assert.Equal(expectedId, communityGalleryImage.UniqueId);
+            Assert.NotNull(communityGalleryImage.Eula);
```
Sandido, line 108: Can you check for the exact expected value?

Sandido (changes requested):

sdk/compute/Microsoft.Azure.Management.Compute/tests/ScenarioTests/CommunityGalleryTests.cs
Sandido: Just in case something unexpectedly changes the value in an incorrect or breaking way.

kangsun-ctrl: the exact value is a random string i put. It does not add much value here to check exact value because it might gets updated by mistake and this test will fail. But the only thing we want here is the string flow-through and NotNull assert do can 100% of that job right.

PR #28248, opened 2022-04-19
Add Avro perf tests

JoshLove-msft:


sdk/schemaregistry/Microsoft.Azure.Data.SchemaRegistry.ApacheAvro/perf/Scenarios/Deserialize.cs
```diff
@@ -0,0 +1,41 @@
...
+using TestSchema;
+
+namespace Microsoft.Azure.Data.SchemaRegistry.ApacheAvro.Perf.Scenarios
+{
+    public class Deserialize : SchemaRegistryAvroSerializerPerfTestBase
+    {
+        private readonly List<MessageContent> _serializedData = new();
+        public Deserialize(SizeCountOptions options) : base(options)
+        {
+            var employee = new Employee() { Name = "Caketown", Age = 42 };
```
jsquire, line 18: This is probably best done as part of `GlobalSetup` and used as a static set across parallel instances.

sdk/schemaregistry/Microsoft.Azure.Data.SchemaRegistry.ApacheAvro/perf/Scenarios/Serialize.cs
```diff
@@ -0,0 +1,40 @@
...
+
+namespace Microsoft.Azure.Data.SchemaRegistry.ApacheAvro.Perf.Scenarios
+{
+    public class Serialize : SchemaRegistryAvroSerializerPerfTestBase
+    {
+        private readonly List<Employee> _employees = new();
+
+        public Serialize(SizeCountOptions options) : base(options)
+        {
+            for (int i = 0; i < options.Count; i++)
```
jsquire, line 18: This is probably best done as part of `GlobalSetup` and used as a static set across parallel instances.

jsquire (approved):

sdk/schemaregistry/Microsoft.Azure.Data.SchemaRegistry.ApacheAvro/perf/Scenarios/Deserialize.cs
JoshLove-msft: GlobalSetup is an instance method, so I'm not sure how that would work. This can be done in a static constructor to share across instances, but I don't think it really matters.

jsquire: Global setup is called once across all instances of the tests before anything is run.   It's intended to be used similarly to a static constructor.

PR #28264, opened 2022-04-19
Adding package exclusion feature for smoke test

### LINKED ISSUE #28265: Smoke test should have package filter to exclude specified packages for nightly
Right now the Azure.AI.Personalizer package is having type loading issue that won't be resolved anytime soon. To maintain usability of smoke test nightly run, we should have an exclude package feature for scenarios like this.

ckairen:
Closes #28265 
Pipeline Ref:
https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1515783&view=results

### DIFF
--- a/eng/scripts/smoketest/Get-Package-Version.ps1
+++ b/eng/scripts/smoketest/Get-Package-Version.ps1
@@ -23,7 +23,8 @@ function Get-SmokeTestPkgProperties
 
     $NugetSource="https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-for-net/nuget/v2"
     $allPackages = Get-AllPkgProperties
-    $newPackages = $allPackages.Where({ $_.IsNewSdk })
+    $excludePackages = (Get-Content 'SmokeTestExcludePackage.json' | ConvertFrom-Json).ExcludePackages
+    $newPackages = $allPackages.Where({ $_.IsNewSdk -and !$excludePackages.Contains($_.Name) })
 
     $azureCorePkgInfo = $newPackages.Where({ $_.Name -eq "Azure.Core"})
     $azureCoreVer = [AzureEngSemanticVersion]::ParseVersionString($azureCorePkgInfo.Version)
--- a/eng/scripts/smoketest/Get-Package-Version.ps1
+++ b/eng/scripts/smoketest/Get-Package-Version.ps1
@@ -23,7 +23,7 @@ function Get-SmokeTestPkgProperties
 
     $NugetSource="https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-for-net/nuget/v2"
     $allPackages = Get-AllPkgProperties
-    $excludePackages = (Get-Content 'SmokeTestExcludePackage.json' | ConvertFrom-Json).ExcludePackages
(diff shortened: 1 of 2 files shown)


ckairen:
TODO:
After merging this in, set the schedule for net - smoketest

eng/scripts/smoketest/Get-Package-Version.ps1
```diff
@@ -23,7 +23,8 @@ function Get-SmokeTestPkgProperties
 
     $NugetSource="https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-for-net/nuget/v2"
     $allPackages = Get-AllPkgProperties
-    $newPackages = $allPackages.Where({ $_.IsNewSdk })
+    $excludePackages = (Get-Content "$PSScriptRoot/SmokeTestExcludePackage.json" | ConvertFrom-Json).ExcludePackages
```
benbp, line 26: I think you need `-Raw` here, otherwise `Get-Content` will return a newline delimited array instead of a json string (see [docs](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.management/get-content?view=powershell-7.2#example-6-get-raw-content))

weshaggard: Given this list is only for this one smoke test project I'd suggest just doing it inline in the PS script similar to https://github.com/Azure/azure-sdk-for-js/blob/main/eng/scripts/Language-Settings.ps1#L235-L238 and include a link to the issue as well.

eng/scripts/smoketest/Get-Package-Version.ps1
```diff
@@ -23,7 +23,10 @@ function Get-SmokeTestPkgProperties
 
     $NugetSource="https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-for-net/nuget/v2"
     $allPackages = Get-AllPkgProperties
-    $newPackages = $allPackages.Where({ $_.IsNewSdk })
+    $excludePackages = @{
+        "Azure.AI.Personalizer" = 'Unable to load DLL rl.net.native.dll https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1492856&view=ms.vss-test-web.build-test-results-tab&runId=30630890&resultId=100012&paneView=debug';
```
weshaggard, line 27: Is there an issue filed for this? If so please link the issue here and if not please file an issue at put this context in it and link it here.

weshaggard (approved):
One comment about linking to a github issue otherwise looks good.

eng/scripts/smoketest/Get-Package-Version.ps1
benbp: I would just paste the full error message in the comment instead of adding a link to an internal pipeline:

```
System.TypeInitializationException : The type initializer for 'Rl.Net.Native.NativeMethods' threw an exception.
----> System.DllNotFoundException : Unable to load DLL 'rl.net.native.dll': The specified module could not be found. (Exception from HRESULT: 0x8007007E)
```

benbp: We can link to the internal pipeline for debugging in the issue description after it's created.

weshaggard (approved):

PR #28267, opened 2022-04-19
Adding support for retry policy (SupportsRetryAttribute)

alrod:
EventHub trigger binding needs `SupportsRetryAttribute` in order to support retry policy.
Corresponding change in webjobs-sdk:
https://github.com/Azure/azure-webjobs-sdk/commit/830b86a1f1f03a7405315b0cc0b94a310d1df5ee

### DIFF
--- a/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Triggers/EventHubTriggerAttributeBindingProvider.cs
+++ b/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Triggers/EventHubTriggerAttributeBindingProvider.cs
@@ -70,11 +70,11 @@ namespace Microsoft.Azure.WebJobs.EventHubs
                                                 _loggerFactory);
                  return Task.FromResult(listener);
              };
-
 #pragma warning disable 618
             ITriggerBinding binding = BindingFactory.GetTriggerBinding(new EventHubTriggerBindingStrategy(), parameter, _converterManager, createListener);
 #pragma warning restore 618
-            return Task.FromResult(binding);
+            EventHubTriggerBindingWrapper eventHubTriggerBindingWrapper = new EventHubTriggerBindingWrapper(binding);
+            return Task.FromResult(eventHubTriggerBindingWrapper as ITriggerBinding);
         }
     } // end class
 }
--- a/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Triggers/EventHubTriggerAttributeBindingProvider.cs
+++ b/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Triggers/EventHubTriggerAttributeBindingProvider.cs
@@ -73,8 +73,8 @@ namespace Microsoft.Azure.WebJobs.EventHubs
 #pragma warning disable 618
             ITriggerBinding binding = BindingFactory.GetTriggerBinding(new EventHubTriggerBindingStrategy(), parameter, _converterManager, createListener);
 #pragma warning restore 618
-            EventHubTriggerBindingWrapper eventHubTriggerBindingWrapper = new EventHubTriggerBindingWrapper(binding);
-            return Task.FromResult(eventHubTriggerBindingWrapper as ITriggerBinding);
+            ITriggerBinding eventHubTriggerBindingWrapper = new EventHubTriggerBindingWrapper(binding);
+            return Task.FromResult(eventHubTriggerBindingWrapper);
         }
     } // end class
 }
--- a/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Triggers/EventHubTriggerBindingWrapper.cs
+++ b/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Triggers/EventHubTriggerBindingWrapper.cs
@@ -1,0 +1,44 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Azure.WebJobs.Host;
+using Microsoft.Azure.WebJobs.Host.Bindings;
+using Microsoft.Azure.WebJobs.Host.Listeners;
+using Microsoft.Azure.WebJobs.Host.Protocols;
+using Microsoft.Azure.WebJobs.Host.Triggers;
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs.EventHubs
+{
+    [SupportsRetry]
+    internal class EventHubTriggerBindingWrapper : ITriggerBinding
+    {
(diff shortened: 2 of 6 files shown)


sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Triggers/EventHubTriggerAttributeBindingProvider.cs
```diff
@@ -70,11 +70,11 @@ public Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext contex
                                                 _loggerFactory);
                  return Task.FromResult(listener);
              };
-
 #pragma warning disable 618
             ITriggerBinding binding = BindingFactory.GetTriggerBinding(new EventHubTriggerBindingStrategy(), parameter, _converterManager, createListener);
 #pragma warning restore 618
-            return Task.FromResult(binding);
+            EventHubTriggerBindingWrapper eventHubTriggerBindingWrapper = new EventHubTriggerBindingWrapper(binding);
```
JoshLove-msft, line 76:
```suggestion
            ITriggerBinding eventHubTriggerBindingWrapper = new EventHubTriggerBindingWrapper(binding);
```

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Triggers/EventHubTriggerAttributeBindingProvider.cs
```diff
@@ -70,11 +70,11 @@ public Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext contex
...
+            return Task.FromResult(eventHubTriggerBindingWrapper as ITriggerBinding);
```
JoshLove-msft, line 77:
```suggestion
            return Task.FromResult(eventHubTriggerBindingWrapper);
```

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Triggers/EventHubTriggerBindingWrapper.cs
```diff
@@ -0,0 +1,44 @@
...
+        public IReadOnlyDictionary<string, Type> BindingDataContract => _innerTriggerBinding.BindingDataContract;
+
+        public Task<ITriggerData> BindAsync(object value, ValueBindingContext context)
+        {
+            return _innerTriggerBinding.BindAsync(value, context);
+        }
+
+        public Task<IListener> CreateListenerAsync(ListenerFactoryContext context)
+        {
+            return _innerTriggerBinding.CreateListenerAsync(context);
```
JoshLove-msft, line 36: nit: suggest awaiting to have more understandable call stack in case of errors.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Triggers/EventHubTriggerBindingWrapper.cs
```diff
@@ -0,0 +1,44 @@
...
+            _innerTriggerBinding = triggerBinding;
+        }
+
+        public Type TriggerValueType => _innerTriggerBinding.TriggerValueType;
+
+        public IReadOnlyDictionary<string, Type> BindingDataContract => _innerTriggerBinding.BindingDataContract;
+
+        public Task<ITriggerData> BindAsync(object value, ValueBindingContext context)
+        {
+            return _innerTriggerBinding.BindAsync(value, context);
```
JoshLove-msft, line 31: nit: suggest awaiting to have more understandable call stack in case of errors.

JoshLove-msft (commented):
Can we add some tests?

JoshLove-msft:
@jsquire it looks like the extensions currently have a project reference to the EH library. Is this something that can be feasibly removed prior to the EH GA? The functions team is hoping to ship an update to the EH extension to support the retry feature which is about to GA.

alrod:
> Can we add some tests?
Right now we allow all triggers for the retry policy, it will be changed once we release extensions with retry only support(Eventhub, Timmer). webjobs sdk will throw the [warning](https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs#L433) for all non-supported triggers.

The test can be added after [the change](https://github.com/Azure/azure-webjobs-sdk/issues/2788) in webjob sdk.

JoshLove-msft:
> > Can we add some tests?
> > Right now we allow all triggers for the retry policy, it will be changed once we release extensions with retry only support(Eventhub, Timmer). webjobs sdk will throw the [warning](https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs#L433) for all non-supported triggers.
> 
> The test can be added after [the change](https://github.com/Azure/azure-webjobs-sdk/issues/2788) in webjob sdk.

Well we can still test that the feature actually works, right? I assume if a function throws an exception it will get retried if it has the attribute?

JoshLove-msft (approved):

jsquire:
> @jsquire it looks like the extensions currently have a project reference to the EH library. Is this something that can be feasibly removed prior to the EH GA? The functions team is hoping to ship an update to the EH extension to support the retry feature which is about to GA.

You could do it if you backout the changes made in [a5f2ba22c34a6d931b3717aa36f1c286c4647d13](https://github.com/Azure/azure-sdk-for-net/commit/a5f2ba22c34a6d931b3717aa36f1c286c4647d13) - it'll be messy, though.  In addition to the changes made in the extension itself, there were changes in the Shared library that are imported.  Most notably, the internal `BlobsCheckpointStore` became `BlobsCheckpointStoreInternal` and the signature for creating checkpoints was updated.    

To avoid breaking the Event Hubs packages, you'd have to copy the prior version of the shared checkpoint store into the extensions package, remove the linkages in the extensions + extensions tests and then play a bit of whack-a-mole for other compilation issues.   I'd count on me having forgotten a thing or two.

alrod:
@jsquire, do you think we can do a release from the commit [Increment version for eventhub releases](https://github.com/Azure/azure-sdk-for-net/commit/4b61e879b3b1f620a0a5efd9f14b3f4579e7e1c4) - the commit before [[Event Hubs] Checkpoint Store Abstraction](https://github.com/Azure/azure-sdk-for-net/commit/a5f2ba22c34a6d931b3717aa36f1c286c4647d13) + the cherry pick the change in the PR.

image

jsquire:
> @jsquire, do you think we can do a release from the commit [Increment version for eventhub releases](https://github.com/Azure/azure-sdk-for-net/commit/4b61e879b3b1f620a0a5efd9f14b3f4579e7e1c4) - the commit before [[Event Hubs] Checkpoint Store Abstraction](https://github.com/Azure/azure-sdk-for-net/commit/a5f2ba22c34a6d931b3717aa36f1c286c4647d13) + cherry pick the change in the PR.

That would be my recommendation.   I believe there would be less risk of accidentally breaking something that way.

PR #28274, opened 2022-04-19
Remove SchemaRegistryAvroException

JoshLove-msft:


sdk/schemaregistry/Microsoft.Azure.Data.SchemaRegistry.ApacheAvro/src/SchemaRegistryAvroSerializer.cs
```diff
@@ -392,10 +397,9 @@ private static SupportedType GetSupportedTypeOrThrow(Type type)
             }
             catch (SchemaParseException ex)
             {
-                throw new SchemaRegistryAvroException(
+                throw new FormatException(
                     $"An error occurred while attempting to parse the schema (schema ID: {schemaId}) that was used to serialize the Avro. " +
                     $"Make sure that the schema represents valid Avro.",
-                    schemaId,
```
swathipil, line 398: wondering if not including schemaID in the exception changes error scenarios/how we expect customers to handle errors?

or, since this can be added in the future when customers ask, is this not really a big deal?

JoshLove-msft: The assumption is that this is not an error that users would handle programmatically. You raise a good point though because by throwing a FormatException rather than base Exception we are probably closing the door on introducing a custom type in the future that would have the SchemaId property. I will wait to hear back from the .NET architects.

JoshLove-msft: That said, when deserializing the schema ID will be in the content type, and when serializing the user could look up the schema ID based on the schema if needed.

deyaaeldeen (approved):
(not) so long my friend!

jsquire (approved):
Not sure that `InvalidOperationException` feels right for some of those scenarios, but I don’t have a better suggestion that would cover all cases.

sdk/schemaregistry/Microsoft.Azure.Data.SchemaRegistry.ApacheAvro/src/SchemaRegistryAvroSerializer.cs
JoshLove-msft: Updated to wrapping with Exception.

JoshLove-msft:
> Not sure that `InvalidOperationException` feels right for some of those scenarios, but I don’t have a better suggestion that would cover all cases.

Changed to wrapping with Exception.

PR #28298, opened 2022-04-20
Update Chaos SDK to latest version of the OpenAPI spec

jaysvoboda:
- Incremented version from `0.9.15-preview` to `0.9.15-preview.1`.
- Includes change for `startTime` and `endTime` in `ActionStatus` model.
- Includes change for `experimentName` Regex pattern in the OpenAPI spec, but this change is not applied to SDK because client-side validation is not enabled.


Yao725 (approved):
LGTM

PR #28300, opened 2022-04-20
update Code Owner for Form Recognizer

anrothMSFT:
I've left the team, so I've updated it with the current PM owner.


maririos (approved):
Thank you so much for helping us update this file :)

PR #28316, opened 2022-04-21
[Fabric Bot] Update Form Recognizer Contacts

jsquire:
# Summary

The focus of these changes is to update the bot rules to mirror the CODEOWNERS changes made in #28300.

### DIFF
--- a/.github/fabricbot.json
+++ b/.github/fabricbot.json
@@ -1463,7 +1463,7 @@
           ],
           "mentionees": [
             "ctstone",
-            "anrothMSFT"
+            "vkurpad"
           ]
         },
         {


christothes (approved):

.github/fabricbot.json
```diff
@@ -1463,7 +1463,7 @@
           ],
           "mentionees": [
             "ctstone",
-            "anrothMSFT"
+            "vkurpad"
```
maririos, line 1466: out of curiosity, si there a way to keep both files synced so we don't forget to update this in the future?

jsquire: Not currently, this is one of the big weaknesses of Fabric Bot.   We're looking at some alternatives so that we can move these out of CODEOWNERS and into a single data file, but that's a bit of a longer-term project.

PR #28320, opened 2022-04-21
Eventhub Stress Test Onboarding

m-redding:


sdk/eventhub/Azure.Messaging.EventHubs.Shared/src/Azure.Messaging.EventHubs.Shared.Stress.projitems
```diff
@@ -0,0 +1,15 @@
...
+    <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
+    <HasSharedItems>true</HasSharedItems>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <Import_RootNamespace>Azure.Messaging.EventHubs.Tests</Import_RootNamespace>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="$(MSBuildThisfileDirectory)Testing\EventGenerator.cs" Link="SharedSource\Stress\%(Filename)%(Extension)" />
```
benbp, line 13:
```suggestion
    <Compile Include="$(MSBuildThisFileDirectory)Testing\EventGenerator.cs" Link="SharedSource\Stress\%(Filename)%(Extension)" />
```

sdk/eventhub/Azure.Messaging.EventHubs/stress/Chart.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v2
+name: eventhub-net-stress-test
+version: 0.1.1
+description: Stress tests for Event Hubs for .NET
+
+dependencies:
+- name: stress-test-addons
+  version: 0.1.13
+  repository: https://stresstestcharts.blob.core.windows.net/helm/
```
benbp, line 9: FYI @ckairen for the list to update.

sdk/eventhub/Azure.Messaging.EventHubs/stress/Dockerfile
```diff
@@ -0,0 +1,28 @@
+# right version of dotnet?
+FROM mcr.microsoft.com/dotnet/sdk:6.0 AS build-env 
+
+# Copy in engineering system needed to build - TODO: cut this down to just what is needed
+COPY /sdk/eventhub/Azure.Messaging.EventHubs/ /app/sdk/eventhub/Azure.Messaging.EventHubs/
+COPY /eng /app/eng
+COPY /tools /app/tools
```
benbp, line 7: Do these also need to be `./` instead of `/`?

sdk/eventhub/Azure.Messaging.EventHubs/stress/buildstress.ps1
```diff
@@ -0,0 +1,6 @@
+# TODO: how to update registry path
+
+Set-Location -Path ..\..\..\..\
+Set-Location -Path .\artifacts\bin\Azure.Messaging.EventHubs.Stress\Debug\netcoreapp3.1
```
benbp, line 4: This script will fail the first time you run it if you have no eventhubs build.

sdk/eventhub/Azure.Messaging.EventHubs/stress/Chart.yaml
```diff
@@ -0,0 +1,15 @@
...
+
+annotations:
+  stressTest: 'true'
+  namespace: 'net'
+  dockerbuilddir: '../../../..'
+  dockerfile: './Dockerfile'
```
benbp, line 15: @ckairen I'm also realizing here that we have no easy way to docker build without deploying when adding the custom build dirs.

sdk/eventhub/Azure.Messaging.EventHubs/stress/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Local Stress Test Runs
+### Setting up resources and packages
+```cmd
+(env) ~/stress/src> dotnet clean
+(env) ~/stress/src> dotnet build
+```
+
+### Running Tests
+When tests are run locally, Azure resources need to be created prior to running the test. This can be done through the Azure CLI, an ARM template or bicep file, or the Azure Portal. The user is required to input the connection strings upon request on the command line when the test is being run. For more information about what resources are needed for each test, see the "Scenario Information" section below.
```
benbp, line 18: @ckairen I think we should also support a "deploy resources only" type mode to make local testing easier with resources.

sdk/eventhub/Azure.Messaging.EventHubs/stress/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+RUN dotnet restore './sdk/eventhub/Azure.Messaging.EventHubs/stress/src/'
+RUN dotnet build './sdk/eventhub/Azure.Messaging.EventHubs/stress/src/'
+COPY /artifacts/bin/Azure.Messaging.EventHubs.Stress/Debug /app/bin/Debug
+
+# Copy in the dll files to be ready to run
+FROM build-env as publish
+WORKDIR /app
+COPY --from=build-env /app/artifacts/bin/Azure.Messaging.EventHubs.Stress/Debug /app/artifacts/bin/Azure.Messaging.EventHubs.Stress/Debug
+COPY --from=build-env /app/sdk/eventhub/Azure.Messaging.EventHubs.Shared /app/sdk/evethub/Azure.Messaging.EventHubs.Shared
+COPY --from=build-env /app/sdk/eventhub/Azure.Messaging.EventHubs/stress /app/sdk/eventhub/Azure.Messaging.EventHubs/stress
```
benbp, line 26: I'm pretty sure this is a "me" problem because I don't have any of the other NET repo build configs set up.

```
--> 9ab93d936cd
[1/2] STEP 14/14: COPY /artifacts/bin/Azure.Messaging.EventHubs.Stress/Debug /app/bin/Debug
Error: error building at STEP "COPY /artifacts/bin/Azure.Messaging.EventHubs.Stress/Debug /app/bin/Debug": checking on sources under "/home/ben/sdk/azure-sdk-for-net": copier: stat: "/artifacts/bin/Azure.Messaging.EventHubs.Stress/Debug": no such file or directory
```

benbp: Also it may be helpful to add an `ENTRYPOINT` or `RUN` directive to support local `docker run` commands easily (unless I'm reading the dockerfile wrong).

sdk/eventhub/Azure.Messaging.EventHubs/stress/stress-test-resources.json
```diff
@@ -0,0 +1,165 @@
+{
```
benbp, line 1: I would avoid checking this file in and also add a `.gitignore` for it ([example](https://github.com/Azure/azure-sdk-tools/blob/main/tools/stress-cluster/chaos/examples/.gitignore#L1)) since it's an intermediate artifact.

sdk/eventhub/Azure.Messaging.EventHubs/stress/stress-test-resources.bicep
```diff
@@ -0,0 +1,95 @@
...
+@description('The maximum duration, in minutes, that a single test is permitted to run before it is considered at-risk for being hung.')
+param perTestExecutionLimitMinutes string = '15'
+
+@description('The client OID to grant access to test resources.')
+param testApplicationOid string
+
+@description('The location of the resource group.')
+param location string = resourceGroup().location
+
+var eventHubsNamespace_var = 'eh-${resourceGroup().name}'
```
benbp, line 13: If you only use the resource group name here, it will be a little simpler to use the auto-generated namespace URL in the network blocking config, but it should still work fine with `eh-` we'll just have to remember to add it in other places.

benbp:
A couple nits but the stress configuration pieces LGTM.

sdk/eventhub/Azure.Messaging.EventHubs/stress/Chart.yaml
ckairen: Well taking the publish image code out as a separate script is not too hard, also helps decrease the size of deployment script which I'm sure will still grow. Do you think it's better to do it that way?

benbp: I wonder if having a higher level type CLI/script for working with stress tests could be useful. E.g. `stress build`, `stress start`, `stress deploy`, `stress debug` etc. etc.

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Infrastructure/Metrics.cs
```diff
@@ -0,0 +1,58 @@
...
+        public string LostEventDataEvent = "LostEventDataEvent";
+
+        // Environment statistics
+        public MetricIdentifier MemorySamples = new MetricIdentifier("MemorySamples");
+        public MetricIdentifier GenerationZeroCollections = new MetricIdentifier("GenerationZeroCollections");
+        public MetricIdentifier GenerationOneCollections = new MetricIdentifier("GenerationOneCollections");
+        public MetricIdentifier GenerationTwoCollections = new MetricIdentifier("GenerationTwoCollections");
+
+        // Publisher statistics
+        public MetricIdentifier BatchesCount = new MetricIdentifier("BatchesCount");
```
jsquire, line 29: Wouldn't we want to have distinct names for each scenario?   Otherwise, they'd bleed over, no?

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Infrastructure/Publisher.cs
```diff
@@ -0,0 +1,174 @@
...
+                }
+                catch (Exception ex)
+                {
+                    // If this catch is hit, there's a problem with the producer itself; cancel the
+                    // background sending and wait before allowing the producer to restart.
+
+                    backgroundCancellationSource.Cancel();
+                    await Task.WhenAll(sendTasks).ConfigureAwait(false);
+
+                    //Interlocked.Increment(ref Metrics.ProducerRestarted);
```
jsquire, line 121: We'll definitely want an equivalent for this.  It is used to tell us if the producer is crashing, and we had to restart it.

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Infrastructure/BufferedPublisher.cs
```diff
@@ -0,0 +1,155 @@
...
+                        metrics.Client.GetMetric(metrics.SuccessfullySentFromQueue, args.PartitionId).TrackValue(numEvents);
+
+                        return Task.CompletedTask;
+                    };
+
+                    producer.SendEventBatchFailedAsync += args =>
+                    {
+                        var numEvents = args.EventBatch.ToList().Count;
+
+                        var eventProperties = new Dictionary<String, String>();
```
jsquire, line 59: Doesn't look like we're using this...

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Infrastructure/BufferedPublisher.cs
```diff
@@ -0,0 +1,155 @@
...
+        {
+            var enqueueTasks = new List<Task>();
+
+            while (!cancellationToken.IsCancellationRequested)
+            {
+                using var backgroundCancellationSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+
+                try
+                {
+                    var producer = new EventHubBufferedProducerClient(connectionString, eventHubName);
```
jsquire, line 44: We'll definitely want to refactor a bit to make this class (and similar role-type things) async disposable;  we need to be able to dispose the producer without flushing as part of cleanup.

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Infrastructure/BufferedPublisher.cs
```diff
@@ -0,0 +1,155 @@
...
+                    producer.SendEventBatchSucceededAsync += args =>
+                    {
+                        var numEvents = args.EventBatch.ToList().Count;
+                        var eventProperties = new Dictionary<String, String>();
```
jsquire, line 48: Doesn't look like we're using this...

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Infrastructure/BufferedPublisher.cs
```diff
@@ -0,0 +1,155 @@
...
+
+                        var eventProperties = new Dictionary<String, String>();
+
+                        metrics.Client.GetMetric(metrics.EventsNotSentAfterEnqueue, args.PartitionId).TrackValue(numEvents);
+                        metrics.Client.TrackException(args.Exception);
+
+                        return Task.CompletedTask;
+                    };
+
+                    await using (producer.ConfigureAwait(false))
```
jsquire, line 67: ooohhh... I see what you're up to.    Clever idea.   We should probably consider being more explicit about the lifecycle, though.   It would be easy for someone to miss this if they weren't familiar.

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Infrastructure/BufferedPublisher.cs
```diff
@@ -0,0 +1,155 @@
...
+                    throw;
+                }
+                catch (Exception ex)
+                {
+                    metrics.Client.TrackException(ex);
+                }
+            }
+        }
+
+        private async Task PerformSend(EventHubBufferedProducerClient producer,
```
jsquire, line 131: nit:  We probably should rename this to denote that it's enqueuing rather than sending.

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Program.cs
```diff
@@ -0,0 +1,132 @@
...
+namespace Azure.Messaging.EventHubs.Stress
+{
+    public class Program
+    {
+        public static async Task Main(string[] args)
+        {
+            var testsToRun = ParseArguments(args);
+
+            // Determine which tests to run
+            testsToRun.TryGetValue("EventProd", out var runEventProducerTest);
```
jsquire, line 25: How does this work with the number of pods?   I don't think we want to allow concurrent tests in the same pod.   We should be able to run concurrent tests in different pods... but that would likely be outside of the purview of this executable.

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Program.cs
```diff
@@ -0,0 +1,132 @@
...
+
+            while (string.IsNullOrEmpty(testConfiguration.EventHub))
+            {
+                Console.Write($"Please provide the name of the Event Hub that you'd like to use for the {testName} and then press Enter: ");
+                testConfiguration.EventHub = Console.ReadLine().Trim();
+                Console.WriteLine();
+            }
+        }
+
+        private static Dictionary<string,bool> ParseArguments(string[] args)
```
jsquire, line 120: nifty.   We can probably shift to using something like `System.CommandLine.Dragonfruit` to make arguments typed and avoid having to parse.

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Program.cs
```diff
@@ -0,0 +1,132 @@
...
+            var testsToRun = ParseArguments(args);
+
+            // Determine which tests to run
+            testsToRun.TryGetValue("EventProd", out var runEventProducerTest);
+            testsToRun.TryGetValue("EventBuffProd", out var runEventBufferedProducerTest);
+            testsToRun.TryGetValue("ConcurBuffProd", out var runConcurrentBufferedProducerTest);
+            testsToRun.TryGetValue("BurstBuffProd", out var runBurstBufferedProducerTest);
+
+            var ENV_FILE = Environment.GetEnvironmentVariable("ENV_FILE");
+            var environment = EnvironmentReader.LoadFromFile(ENV_FILE);
```
jsquire, line 31:
```suggestion
            var environment = EnvironmentReader.LoadFromFile(Environment.GetEnvironmentVariable("ENV_FILE"));
```

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Program.cs
```diff
@@ -0,0 +1,132 @@
...
+
+            if (runEventProducerTest)
+            {
+                var testConfiguration = new ProducerConfiguration();
+                environment.TryGetValue("EVENTHUB_NAMESPACE_CONNECTION_STRING", out testConfiguration.EventHubsConnectionString);
```
jsquire, line 36: We should consider a set of constant strings for these:

```csharp
public static class EnvironmentVariables
{
    public const string ApplicationInsightsKey = "APPINSIGHTS_INSTRUMENTATIONKEY";
    // ...  more stuffs
}
```

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Program.cs
```diff
@@ -0,0 +1,132 @@
...
+
+                var testRun = new BufferedProducerTest(testConfiguration);
+                await testRun.Run();
+            }
+        }
+
+        private static void PromptForResources(ProducerConfiguration testConfiguration, string testName)
+        {
+            // Prompt for the App Insights instrumentation key
+            while (string.IsNullOrEmpty(testConfiguration.InstrumentationKey))
```
jsquire, line 96: Not sure this works well in the stress environment;  should we consider locking this down to some kind of -`--interactive` flag and throwing if we don't have enough information otherwise?

jsquire (commented):
Great start!    I left a bunch of thoughts, most are minor.    We probably want to remove the ARM template output and add `.gitignore` before committing.     Otherwise, let me know how you'd prefer to move forward.   We can commit this and then do some follow-up PRs or we can iterate here and commit after.

I blatantly ignored any formatting that we'll want to update and mostly ignored naming stuff.   We've got time to revisit that once we feel good about being able to run the buffered producer tests "for real"

sdk/eventhub/Azure.Messaging.EventHubs/stress/Dockerfile
benbp: @jsquire I'm definitely in favor of common stress components, either global or language level. For this dockerfile, my initial thought would be to create an intermediate builder image, so you then you can do something like `FROM azsdkengsys.azurecr.io/stress/dotnet/base AS build-env`?

benbp: Development could be somewhat annoying if you're iterating on both, so we may want to provide a way to build multiple dockerfiles in sequence if it starts getting iterated on heavily.

sdk/eventhub/Azure.Messaging.EventHubs/stress/Chart.yaml
benbp: @jsquire right now the script automation basically assumes you're doing everything: logging in, building and pushing images, and deploying the stress tests to the cluster. You can disable some of this functionality via script parameters but it doesn't apply to everything and isn't optimized for all dev workflows. I think that it would be better to re-think the UX of the stress command line tools.

benbp: https://github.com/Azure/azure-sdk-tools/issues/3193

sdk/eventhub/Azure.Messaging.EventHubs/stress/Dockerfile
benbp: @m-redding thinking more about my first comment, is it expected that these dockerfile commands will work in a fresh/from scratch environment? Eventually we will want the stress test to be able to be deployable in CI to the "stress-prod" environment, meaning the docker build command alone can get everything built the way it needs to. Currently any stress tests checked in to main will get auto-deployed weekly to that env.

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Azure.Messaging.EventHubs.Stress.csproj
```diff
@@ -0,0 +1,28 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <Version>1.0.0</Version>
+    <OutputType>Exe</OutputType>
+    <LangVersion>latest</LangVersion>
+    <IsTestSupportProject>true</IsTestSupportProject>
+    <IsStressProject>false</IsStressProject>
```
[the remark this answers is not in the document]
benbp: I was not familiar with these, but a cursory glance at the repo makes me think we can. @mikeharder?

sdk/eventhub/Azure.Messaging.EventHubs/stress/templates/deploy-job.yaml
```diff
@@ -0,0 +1,17 @@
+{{- include "stress-test-addons.deploy-job-template.from-pod" (list . "stress.event-hub") -}}
```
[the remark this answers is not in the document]
benbp: It's the curly braces that stand out to you here and not the ridiculous constructing of a list to pass to a template with quasi-function arguments while in a config block? :D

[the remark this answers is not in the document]
benbp: (By ridiculous I'm referring to the necessities of how I've implemented the helm boilerplate and not the PR changes in question).

sdk/eventhub/Azure.Messaging.EventHubs/stress/stress-test-resources.bicep
```diff
@@ -0,0 +1,95 @@
...
+var defaultSASKeyName = 'RootManageSharedAccessKey'
+var eventHubsAuthRuleResourceId = resourceId('Microsoft.EventHub/namespaces/authorizationRules', eventHubsNamespace_var, defaultSASKeyName)
+
+var ehBufferedProducer = 'bufferedproducertest'
+var bufferproducerPartitions = 10
```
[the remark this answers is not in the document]
benbp: Do you have a preference about how you'd like to configure this? We used to support env-specific `parameters.json` files for ARM template deployment, but it wasn't used to I removed it.

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Program.cs
benbp: The changes here do concurrent tests in different pods. The [values.yaml](https://github.com/Azure/azure-sdk-for-net/blob/52bfa862fa9cdae6172aa0462a465bbe862a3c34/sdk/eventhub/Azure.Messaging.EventHubs/stress/values.yaml) config has a `scenarios` key. When the test is deployed, it will be replicated for each scenario in that list, with the scenario value passed down through the configs. The scenario name is then passed to the app here: https://github.com/Azure/azure-sdk-for-net/blob/52bfa862fa9cdae6172aa0462a465bbe862a3c34/sdk/eventhub/Azure.Messaging.EventHubs/stress/templates/deploy-job.yaml#L14-L15

sdk/eventhub/Azure.Messaging.EventHubs/stress/Dockerfile
jsquire: I really like the idea of the intermediate image.

sdk/eventhub/Azure.Messaging.EventHubs/stress/Chart.yaml
jsquire: ahhhh....  Got it!

sdk/eventhub/Azure.Messaging.EventHubs/stress/stress-test-resources.bicep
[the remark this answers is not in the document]
jsquire: No preference beyond the standard ARM template experience.   If we can make them end up being required parameters that are only injected via params file when we're running on the platform, that would allow interactive prompting when deploying locally or via the portal.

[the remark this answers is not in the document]
benbp: Do you think it's likely that we would also need different config values per "scenario" ?

sdk/eventhub/Azure.Messaging.EventHubs/stress/stress-test-resources.bicep
```diff
@@ -0,0 +1,95 @@
+@description('The url suffix to use when creating storage connection strings.')
```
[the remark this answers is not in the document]
benbp: `az bicep decompile -f <arm template>` is a great way to kickstart this.

sdk/eventhub/Azure.Messaging.EventHubs/stress/stress-test-resources.bicep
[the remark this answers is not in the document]
jsquire: It'd be good to have that flexibility; depending on the scenario, having individual control over things like partition count or number of consumer groups would potentially be needed.

sdk/eventhub/Azure.Messaging.EventHubs/stress/Dockerfile
m-redding: I had hoped that they would, so I'll try to figure out what is causing that error, because yes, we definitely will want them to be deployable in CI eventually.

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Infrastructure/Metrics.cs
m-redding: @jsquire I thought about this too, but the log entries have a column that says which test it comes from, so I went with simplifying the metrics class and filtering it on the other end, but I'm not attached to doing it one way or the other. Is it confusing to have the tests share metrics names?

I think it is easier to combine metrics graphs across tests if they have their own name, if that seems like a useful visualization? I'm not doing that right now. Another option is to include a metrics dimension that includes which test it is, but I'm not sure how much more helpful that is than having it in a separate column.

sdk/eventhub/Azure.Messaging.EventHubs/stress/Dockerfile
m-redding: > Also it may be helpful to add an `ENTRYPOINT` or `RUN` directive to support local `docker run` commands easily (unless I'm reading the dockerfile wrong).

@benbp Just to clarify since I'm still a little new to docker, what would be the usual use case for running `docker run` ? Running all the tests in succession?

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Infrastructure/Metrics.cs
jsquire: Nice.   I didn't realize that we already had structured data for that.   Good by me - that covers my concern!

sdk/eventhub/Azure.Messaging.EventHubs/stress/stress-test-resources.bicep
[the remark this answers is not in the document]
benbp: Makes sense. I've filed https://github.com/Azure/azure-sdk-tools/issues/3208 for this.

sdk/eventhub/Azure.Messaging.EventHubs/stress/README.md
```diff
@@ -49,3 +49,15 @@ This test requires an event hub namespace, an event hub, and an application insi
...
+
+When you submit a pull request, a CLA-bot will automatically determine whether you need to provide a CLA and decorate the PR appropriately (e.g., label, comment). Simply follow the instructions provided by the bot. You will only need to do this once across all repos using our CLA.
+
+This project has adopted the [Microsoft Open Source Code of Conduct](https://opensource.microsoft.com/codeofconduct/). For more information see the [Code of Conduct FAQ](https://opensource.microsoft.com/codeofconduct/faq/) or contact [opencode@microsoft.com](mailto:opencode@microsoft.com) with any additional questions or comments.
+
+Please see our [contributing guide](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Azure.Messaging.EventHubs/CONTRIBUTING.md) for more information.
+  
+![Impressions](https://azure-sdk-impressions.azurewebsites.net/api/impressions/azure-sdk-for-net%2Fsdk%2Feventhub%2FAzure.Messaging.EventHubs%2stress%2FREADME.png)
```
benbp, line 63: This png fails to load on the readme.

sdk/eventhub/Azure.Messaging.EventHubs/stress/README.md
```diff
@@ -0,0 +1,63 @@
+# Event Hub Stress Tests
```
benbp, line 1: I just thought of this, but can you put a link to the stress readme (https://github.com/Azure/azure-sdk-tools/blob/main/tools/stress-cluster/chaos/README.md) somewhere in here?

sdk/eventhub/Azure.Messaging.EventHubs/stress/README.md
```diff
@@ -0,0 +1,63 @@
...
+The recommended approach is to run tests one at a time when running locally.
+To run any one test, run the following:
+```cmd
+(env) ~/stress/src> dotnet run -f netcoreapp3.1 BasicPublishReadTest local
+```
+
+## Deploy a stress test
+In order to deploy stress tests to be run in kubernetes clusters, run:
+```cmd 
+(env) ~/azure-sdk-for-net/eng/common/scripts/stress-testing/deploy-stress-tests.ps1 `
```
benbp, line 29: There are a couple places where the `~` home directory is referenced in this readme, though not everyone will have the same path structure for their repo. I usually type examples like `<git root>/eng/common/...` though I imagine there's a better solution out there.

sdk/eventhub/Azure.Messaging.EventHubs/stress/Dockerfile
benbp: @m-redding One example use case could be: let's say your program was working locally, but wasn't working in the cluster, possibly because of a missing file or dependency. It would be much easier to iterate on a bug like this if you were able to do a dev/test loop like `docker build -t testing && docker run -it testing` vs. having to `push the image/re-deploy the stress test/wait for image download/wait for resource deployment/wait for pod start/run kubectl logs`.

So if you set `ENTRYPOINT` and `CMD` like:

```
ENTRYPOINT ["Azure.Messaging.EventHubs.Stress.exe", "--tests"]
CMD ["EventProd"]
```

Then `docker run -it <image tag>` would execute `Azure.Messaging.Eventhubs.Stress.exe --tests EventProd` in the container on startup. You could then override the scenario like `docker run -it <image tag> BuffProd` which would replace the `CMD` directive and execute `Azure.Messaging.Eventhubs.Stress.exe --tests BuffProd`.

When setting `command` and `args` in the pod yaml spec it tells kubernetes to override anything set in the dockerfile, hence why it's only useful for local dev.

sdk/eventhub/Azure.Messaging.EventHubs/stress/stress-test-resources.json
```diff
@@ -5,7 +5,7 @@
     "_generator": {
       "name": "bicep",
       "version": "0.4.1318.3566",
-      "templateHash": "9849786101028970025"
```
benbp: I think because you previously checked this in, the .gitignore won't automatically remove it. So you have to manually remove it with `git rm` and commit the file deletion.

sdk/eventhub/Azure.Messaging.EventHubs/stress/README.md
jsquire: Loads for me; it's one of our tracking pixels, so it will look like nothing.

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Scenarios/BufferedProducerTest.cs
```diff
@@ -0,0 +1,77 @@
...
+
+            var runDuration = TimeSpan.FromHours(durationInHours);
+            enqueueingCancellationSource.CancelAfter(runDuration);
+
+            var enqueuingTasks = default(IEnumerable<Task>);
+
+            try
+            {
+                enqueuingTasks = Enumerable
+                    .Range(0, 2)
```
jsquire, line 48: I wonder... do we need more than 2?   Generally, we advise customers to use a single instance in their application unless they have unusual needs for a second connection.    Having a second instance in this test is helpful to ensure that we don't have any shared state that would have issues when multiple clients are used.   But I don't know that we have value above 2.

jsquire:
We should add the `ConfigureAwait(false)` to all the things at some point in the near future.   It'll be helpful for ensuring that things keep flowing efficiently since we're in a resource-limited host environment.

sdk/eventhub/Azure.Messaging.EventHubs/stress/README.md
benbp: @jsquire it looks like the problem is a missing `F` character: `EventHubs %2 stress`

image

sdk/eventhub/Azure.Messaging.EventHubs/stress/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+When you submit a pull request, a CLA-bot will automatically determine whether you need to provide a CLA and decorate the PR appropriately (e.g., label, comment). Simply follow the instructions provided by the bot. You will only need to do this once across all repos using our CLA.
+
+This project has adopted the [Microsoft Open Source Code of Conduct](https://opensource.microsoft.com/codeofconduct/). For more information see the [Code of Conduct FAQ](https://opensource.microsoft.com/codeofconduct/faq/) or contact [opencode@microsoft.com](mailto:opencode@microsoft.com) with any additional questions or comments.
+
+Please see our [contributing guide](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Azure.Messaging.EventHubs/CONTRIBUTING.md) for more information.
+  
+![Impressions](https://azure-sdk-impressions.azurewebsites.net/api/impressions/azure-sdk-for-net%2Fsdk%2Feventhub%2FAzure.Messaging.EventHubs%2stress%2FREADME.png)
```
benbp, line 85:
```suggestion
![Impressions](https://azure-sdk-impressions.azurewebsites.net/api/impressions/azure-sdk-for-net%2Fsdk%2Feventhub%2FAzure.Messaging.EventHubs%2Fstress%2FREADME.png)
```

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Program.cs
benbp: I have some examples of this [here](https://github.com/Azure/azure-sdk-tools/blob/main/tools/stress-cluster/cluster/kubernetes/generator/Stress.Generator/Program.cs#L17) if you're interested.

sdk/eventhub/Azure.Messaging.EventHubs/stress/stress-test-resources.bicep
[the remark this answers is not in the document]
benbp: Huh, that's unfortunate, because all the normal deploy sub-commands in the CLI tools DO support native bicep templates (or at least do the conversion under the hood, though I doubt the direct API takes bicep).

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Infrastructure/BufferedProducerTestConfig.cs
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+
+namespace Azure.Messaging.EventHubs.Stress
+{
+    internal class BufferedProducerTestConfig : TestConfig
+    {
+        public int EventEnqueueListSize = 50;
```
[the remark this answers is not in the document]
m-redding: @jsquire Which member of the options would I use for this? There isn't any significance to the value I chose, I'm just using this value for the size of the event list I'm generating from the `EventGenerator` and passing into `EnqueueEventsAsync` in the `BufferedPublisher`.  Thinking about this more though it would probably be helpful to have this be a dynamic value.

sdk/eventhub/Azure.Messaging.EventHubs/stress/README.md
jsquire: ahhh.... I see it now.   Yeah, that's on me.   Thanks for the catch!

jsquire: Chrome/Edge do a nice job of fixing the malformed encoding.   Checking in the browser worked.

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Infrastructure/BufferedProducerTestConfig.cs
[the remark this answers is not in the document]
jsquire: oooohhhhh.....  that's my bad.   I interpreted that as the buffer length, not the number of events that we were generating per enqueue call.

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Program.cs
```diff
@@ -0,0 +1,177 @@
...
+        {
+            var environment = new Dictionary<string, string>();
+            var appInsightsKey = String.Empty;
+            var eventHubsConnectionString = String.Empty;
+            var eventProducerTestConfig = new EventProducerTestConfig();
+            var bufferedProducerTestConfig = new BufferedProducerTestConfig();
+            var burstBufferedProducerTestConfig = new BufferedProducerTestConfig();
+            var concurrentBufferedProducerTestConfig = new BufferedProducerTestConfig();
+
+            foreach (var testRun in opts.Tests)
```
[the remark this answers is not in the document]
benbp: Yeah or rather if no test names are passed in (empty flag) then it defaults to all?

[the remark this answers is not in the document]
jsquire: I'd prefer to be explicit.   If we assume `all` when someone forgets to pass a test, then we could end up spending a bunch of resources on tests that someone didn't mean to launch.

sdk/eventhub/Azure.Messaging.EventHubs/stress/README.md
```diff
@@ -0,0 +1,85 @@
+# Event Hub Stress Tests
```
jsquire, line 1: This is really, really nice.    Well done!

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Configurations/BufferedProducerTestConfig.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Azure.Messaging.EventHubs.Stress
+{
+    internal class BufferedProducerTestConfig : TestConfig
+    {
+        // Test Configurations
+
+        // The number of events to generate and enqueue during each iteration of PerformEnqueue
+        public int EventEnqueueListSize = 50;
+        public long MaximumEventListSize = 1000000;
```
jsquire, line 14:
```suggestion
        public long MaximumEventListSize = 1_000_000;
```

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Configurations/BufferedProducerTestConfig.cs
```diff
@@ -0,0 +1,20 @@
...
+
+        // Buffered Producer Configuration
+
+        public TimeSpan MaxWaitTime = TimeSpan.FromMilliseconds(5000);
```
jsquire, line 18:
```suggestion
        public TimeSpan MaxWaitTime = TimeSpan.FromSeconds(5);
```

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Infrastructure/BufferedPublisher.cs
```diff
@@ -0,0 +1,175 @@
...
+        {
+            var enqueueTasks = new List<Task>();
+
+            while (!cancellationToken.IsCancellationRequested)
+            {
+                using var backgroundCancellationSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+
+                // Create the buffered producer client and register the success and failure handlers
+
+                var options = new EventHubBufferedProducerClientOptions
```
jsquire, line 45: The Buffered Producer has some adjustments to the standard defaults for retry policy if the options aren't explicitly passed.    We should mirror them for the test, since that'll be our majority case.  

Source:  https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Azure.Messaging.EventHubs/src/Producer/EventHubBufferedProducerClient.cs#L68

jsquire: If you have ideas on how we can better detect and adjust retry configuration when the individual values aren't set, I'd be very interested in hearing them.   The best that I can think of is creating a mirrored type inside the buffered options.      Maybe we should explore that as a quick interlude between phase 1 and phase 2.

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Infrastructure/BufferedPublisher.cs
```diff
@@ -0,0 +1,175 @@
...
+using Azure.Messaging.EventHubs.Tests;
+
+namespace Azure.Messaging.EventHubs.Stress
+{
+    internal class BufferedPublisher
+    {
+        private string _connectionString;
+        private string _eventHubName;
+        private Metrics _metrics;
+        private BufferedProducerTestConfig _testConfiguration;
```
jsquire, line 24: these can probably be `readonly`, I think

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Infrastructure/BufferedPublisher.cs
```diff
@@ -0,0 +1,175 @@
...
+                    || ex is StackOverflowException
+                    || ex is ThreadAbortException)
+                {
+                    throw;
+                }
+                catch (Exception ex)
+                {
+                    // If this catch is hit, it means the producer has restarted, collect metrics.
+
+                    _metrics.Client.GetMetric(_metrics.ProducerRestarted).TrackValue(1);
```
jsquire, line 134: Does this increment the count or are we collecting these as individual events and then aggregating via Kusto?

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Infrastructure/EnvironmentReader.cs
```diff
@@ -0,0 +1,126 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using Azure.Core;
+
+namespace Azure.Messaging.EventHubs.Stress;
```
jsquire, line 9: At some indeterminate point in the future, we should probably look to shift the rest of the files to this form.   Very low priority, but I love that we can reduce the nesting.

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Infrastructure/Publisher.cs
```diff
@@ -0,0 +1,176 @@
...
+{
+    internal class Publisher
+    {
+        private static readonly ThreadLocal<Random> RandomNumberGenerator = new ThreadLocal<Random>(() => new Random(Interlocked.Increment(ref _randomSeed)), false);
+
+        private Metrics _metrics;
+
+        private static int _randomSeed = Environment.TickCount;
+
+        private EventProducerTestConfig _testConfiguration;
```
jsquire, line 25:
```suggestion
        private static readonly ThreadLocal<Random> RandomNumberGenerator = new ThreadLocal<Random>(() => new Random(Interlocked.Increment(ref _randomSeed)), false);
        private static int _randomSeed = Environment.TickCount;
        
        private readonly Metrics _metrics;      
        private readonly EventProducerTestConfig _testConfiguration;
```

jsquire: just me being pedantic and grouping the statics together.

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Infrastructure/Publisher.cs
```diff
@@ -0,0 +1,176 @@
...
+using Azure.Messaging.EventHubs;
+using Azure.Messaging.EventHubs.Producer;
+using Azure.Messaging.EventHubs.Tests;
+using Microsoft.ApplicationInsights;
+
+namespace Azure.Messaging.EventHubs.Stress
+{
+    internal class Publisher
+    {
+        private static readonly ThreadLocal<Random> RandomNumberGenerator = new ThreadLocal<Random>(() => new Random(Interlocked.Increment(ref _randomSeed)), false);
```
jsquire, line 19: Are we using this anywhere?   I think this was originally put in before I pulled stuff into the event generator.   We may be able to remove it.

jsquire (approved):

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Infrastructure/BufferedPublisher.cs
m-redding: The GetMetric method is cool it aggregates (in this case sums) up all the values sent via "TrackValue" and sends them all at once:
"In summary GetMetric() is the recommended approach since it does pre-aggregation, it accumulates values from all the Track() calls and sends a summary/aggregate once every minute. This can significantly reduce the cost and performance overhead by sending fewer data points, while still collecting all relevant information." [ref](https://docs.microsoft.com/en-us/azure/azure-monitor/app/get-metric)
The long-term aggregation like total sum is done in Kusto.

PR #28340, opened 2022-04-22
[TA] Merge rest clients into one by renaming OperationIds

maririos:
I saw this idea in the JS implementation and loved it. Better to have all operations in one REST class to avoid confusion.
I also added `batch` to the analyze LRO to distinguish them from non LRO


sdk/textanalytics/Azure.AI.TextAnalytics/src/autorest.md
```diff
@@ -11,6 +11,23 @@ input-file:
 generation1-convenience-client: true
 ```
 
+### Modify operationId names
+
+``` yaml
+directive:
```
jsquire, line 17: Nice!   I didn't know that you could do this.

jsquire (approved):

sdk/textanalytics/Azure.AI.TextAnalytics/src/autorest.md
maririos: Autorest has a lot of magic. If only we knew it all :P

PR #28345, opened 2022-04-22
ContentType small improvements

Ratomir:
# ContentType struct improvements

Improvements for the struct [ContentType](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/ContentType.cs).

Comments should reflect the actual state of properties and be meaningful.
The method ToString is overridden and returns the value of the ContentType in the string format. Which is not the actual behavior that ToString() method does. 

Kind regards,
Ratomir


christothes:
Thanks for your contribution, @Ratomir !

sdk/core/Azure.Core/src/ContentType.cs
```diff
@@ -81,7 +81,10 @@ public override int GetHashCode()
         public static bool operator !=(ContentType left, ContentType right)
             => !left.Equals(right);
 
-        /// <inheritdoc />
-        public override string ToString() => _contentType == null?"":_contentType;
+        /// <summary>
+        /// Returns value of <see cref="ContentType"/>.
+        /// </summary>
+        /// <returns>Value of content type <see cref="ContentType"/>.</returns>
```
christothes, line 87: I think we should revert this - the inheritdoc comment is `Returns a string that represents the current object.` which seems accurate and is consistent with other overrides in the repo.

Ratomir: Thanks @christothes, I will revert to the old comment.

christothes (approved):

PR #28351, opened 2022-04-22
Don't close session when closing link if cross entity transactions are enabled

### LINKED ISSUE #28025: [BUG] RenewMessageLockAsync succeeds but CompleteMessageAsync fails on  +12min + processing durations
### Library name and version

Azure.Messaging.ServiceBus 7.7.0

### Describe the bug

Invoking `RenewMessageLockAsync` works, it updates the lease but eventually when processing is completed the call to `CompleteMessageAsync` fails.

> Azure.Messaging.ServiceBus.ServiceBusException: The lock supplied is invalid. Either the lock expired, or the message has already been removed from the queue, or was received by a different receiver instance. (MessageLockLost)

This while `LockedUntil` on message object is minutes into the future.

What is important to note is that the `ServiceBusReceiver` instances are the same for invoking `ReceiveMessageAsync` and `CompleteMessageAsync` but not for `RenewMessageLockAsync`. 

Interestingly on processing duration between 5 and 10-12 minutes things just work. Sometimes it also magically works for longer processing duration.

At the moment it seems when all methods share the same `ServiceBusReceiver` instance no `MessageLockLost` occurs when invoking `CompleteMessageAsync`. 

Is it supported to invoke `RenewMessageLockAsync` on a different  `ServiceBusReceiver` instance that invoked  `ReceiveMessageAsync` ?

### Expected behavior

I would expect that when `RenewMessageLockAsync` is invoked and it returns a message object with a valid `LockedUntil` that invoking `CompleteMessageAsync`  will not result in a `MessageLockLost`, even if done so on a different `ServiceBusReceiver` instance.

### Actual behavior

Invoking  `CompleteMessageAsync` results in the following exception if the processing duration takes roughly 12 minutes or longer.

> Azure.Messaging.ServiceBus.ServiceBusException: The lock supplied is invalid. Either the lock expired, or the message has already been removed from the queue, or was received by a different receiver instance. (MessageLockLost)

### Reproduction Steps

Working on this, currently  got a super hacky reproduction.

### Environment

Tested with .NET 6 and .NET Framework 4.8

JoshLove-msft:
Fixes https://github.com/Azure/azure-sdk-for-net/issues/28025

### DIFF
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpReceiver.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpReceiver.cs
@@ -167,11 +167,11 @@ namespace Azure.Messaging.ServiceBus.Amqp
                         // it is okay to register the user provided cancellationToken from the AcceptNextSessionAsync call in
                         // the fault tolerant object because session receivers are never reconnected.
                         cancellationToken: cancellationToken),
-                link => CloseLink(link));
+                link => _connectionScope.CloseLink(link));
 
             _managementLink = new FaultTolerantAmqpObject<RequestResponseAmqpLink>(
                 timeout => OpenManagementLinkAsync(timeout),
-                link => CloseLink(link));
+                link => _connectionScope.CloseLink(link));
         }
 
         private async Task<RequestResponseAmqpLink> OpenManagementLinkAsync(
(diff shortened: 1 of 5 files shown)


sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpReceiver.cs
```diff
@@ -1325,10 +1313,10 @@ public override async Task CloseAsync(CancellationToken cancellationToken)
         private void OnReceiverLinkClosed(object receiver, EventArgs e)
         {
             var receivingAmqpLink = (ReceivingAmqpLink)receiver;
-            if (_isSessionReceiver && receivingAmqpLink != null)
+            if (receivingAmqpLink != null)
```
JoshLove-msft, line 1316: Include LinkException in logs even for non-session receivers.

jsquire (approved):

danielmarbach:
Thanks for taking care of this!

PR #28371, opened 2022-04-25
[Identity] Update static token sample to use DelegatedTokenCredential

schaabs:
Fixing prefetched token sample which incorrectly calls `TokenCredential.Create` rather than `DelegatedTokenCredential.Create`.


christothes (approved):

scottaddie (approved):
:shipit:

PR #28375, opened 2022-04-25
Synchronize OperationInternal calls

AlexanderSher:
Changes `OperationInternal` / `OperationInternal<T>` so that only one call of `UpdateStatusAsync` and `WaitForCompletionAsync` can be executed concurrently. After `OperationInternal` / `OperationInternal<T>` is set in its final state, no more call to the service is made.

Implementation is updated based on the discussion in https://github.com/Azure/azure-sdk-for-net/pull/27639


sdk/core/Azure.Core/src/Shared/AsyncLockWithValue.cs
```diff
@@ -0,0 +1,230 @@
...
+using System.Threading.Tasks;
+using Azure.Core.Pipeline;
+
+namespace Azure.Core
+{
+    /// <summary>
+    /// Primitive that combines async lock and value cache
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    internal sealed class AsyncLockWithValue<T>
```
christothes, line 19: can we delete the duplicate instance of this from identity?

AlexanderSher: Done

sdk/core/Azure.Core/src/Shared/AsyncLockWithValue.cs
```diff
@@ -0,0 +1,230 @@
...
+
+        public AsyncLockWithValue(T value)
+        {
+            _hasValue = true;
+            _value = value;
+        }
+
+        public bool TryGetValue(out T? value)
+        {
+            lock (_syncObj)
```
heaths, line 48: Any perf advantage to using a `ReaderWriterLockSlim` instead?

sdk/core/Azure.Core/src/Shared/AsyncLockWithValue.cs
```diff
@@ -0,0 +1,230 @@
...
+                        return nextWaiter;
+                    }
+                }
+
+                _isLocked = false;
+                return default;
+            }
+        }
+
+        public readonly struct Lock : IDisposable
```
heaths, line 194: Can't this just be private since you're returning an `IDisposable`? If not, I'd document its public members. Makes it harder to use without having to grok source.

sdk/core/Azure.Core/src/Shared/OperationInternalOfT.cs
```diff
@@ -223,27 +247,53 @@ public Response<T> WaitForCompletion(TimeSpan pollingInterval, CancellationToken
...
             }
-            return await ApplyStateAsync(async, state.RawResponse, state.HasCompleted, state.HasSucceeded, state.OperationFailedException).ConfigureAwait(false);
+
+            throw state.OperationFailedException!;
+        }
+
+        private class FinalOperation : IOperation<T>
+        {
+            public ValueTask<OperationState<T>> UpdateStateAsync(bool async, CancellationToken cancellationToken)
+                => throw new NotSupportedException("Type is in final state");
```
heaths, line 296: If it's technically possible this could throw out to user code, it should be more actionable and/or diagnosable. Perhaps:
```suggestion
                => throw new NotSupportedException("The operation has already completed");
```

sdk/keyvault/Azure.Security.KeyVault.Certificates/src/DeleteCertificateOperation.cs
```diff
@@ -24,17 +24,20 @@ internal DeleteCertificateOperation(KeyVaultPipeline pipeline, Response<DeletedC
...
-                new KeyValuePair<string, string>("secret", _value.Name), // Retained for backward compatibility.
-                new KeyValuePair<string, string>("certificate", _value.Name),
-            });
 
             // The recoveryId is only returned if soft delete is enabled.
             if (_value.RecoveryId is null)
             {
                 // If soft delete is not enabled, deleting is immediate so set success accordingly.
-                _operationInternal.SetState(OperationState.Success(response.GetRawResponse()));
+                _operationInternal = OperationInternal.Succeeded(response.GetRawResponse());
```
heaths, line 32: The variables have been dropped and should be retained.

heaths: Same in the two other instances below.

heaths (dismissed):

sdk/core/Azure.Core/tests/OperationInternalTests.cs
```diff
@@ -14,256 +15,113 @@
 
 namespace Azure.Core.Tests
 {
-    [TestFixture(true)]
-    [TestFixture(false)]
     public class OperationInternalTests
```
christothes, line 18: Would it be possible to refactor the tests in a separate PR just to ensure we didn't regress anything? It's harder to tell when both the implementation and the tests all change at the same time.

AlexanderSher: Unfortunately, some asserts in these were simply incorrect. For example, initial response should **not** be returned after `UpdateStatusAsync` when state isn't final. So I'm not sure what those tests have been validating exactly,

sdk/core/Azure.Core/src/Shared/AsyncLockWithValue.cs
AlexanderSher: Only for scenario when lots of subscribers decided to cancel their wait, and have done so concurrently, but that is an edge case. In more common scenarios overhead of the `ReaderWriterLockSlim` is higher than code bits in the critical section itself.

sdk/keyvault/Azure.Security.KeyVault.Certificates/src/DeleteCertificateOperation.cs
AlexanderSher: Sorry, can you please clarify this? Which variables have been dropped?

sdk/core/Azure.Core/src/Shared/OperationPoller.cs
```diff
@@ -6,96 +6,99 @@
...
-                {
-                    return Response.FromValue(value(), getRawResponse());
-                }
-
-                Thread.Sleep(_delayStrategy.GetNextDelay(response, suggestedInterval));
+                await Task.Delay(delay, cancellationToken).ConfigureAwait(false);
+            }
+            else if (cancellationToken.CanBeCanceled)
+            {
+                cancellationToken.WaitHandle.WaitOne(delay);
```
JoshLove-msft, line 97: Should we throw if the token is canceled here?

AlexanderSher: Awesome catch!

sdk/core/Azure.Core/src/Shared/OperationInternalOfT.cs
AlexanderSher: Fixed

sdk/core/Azure.Core/src/Shared/AsyncLockWithValue.cs
AlexanderSher: Added comments

sdk/core/Azure.Core/src/Shared/AsyncLockWithValue.cs
```diff
@@ -1,25 +1,63 @@
...
     /// <summary>
     /// Primitive that combines async lock and value cache
     /// </summary>
     /// <typeparam name="T"></typeparam>
     internal sealed class AsyncLockWithValue<T>
     {
-        private readonly object _syncObj = new object();
-        private Queue<TaskCompletionSource<Lock>> _waiters;
+        private readonly object _syncObj = new();
+        private Queue<TaskCompletionSource<LockOrValue>>? _waiters;
```
annelo-msft, line 22: How are we regression testing any libraries that use this type?

sdk/core/Azure.Core/src/Shared/AsyncLockWithValue.cs
```diff
@@ -108,73 +154,107 @@ private void SetValue(T value)
...
                 }
 
+                // If no next waiter has been found, unlock and return
                 _isLocked = false;
-                return default;
             }
         }
 
-        public readonly struct Lock : IDisposable
+        public readonly struct LockOrValue : IDisposable
```
annelo-msft, line 209: It seems we're changing the functionality of this type.  Is there still a use case for just the lock, and would it make sense to instead introduce a second LockOrValue type for just the use cases where value is needed?

AlexanderSher: No, functionality remains the same. It has always been "lock or value" depending on either value is cached or not. I've just thought that `LockOrValue` name is better.

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
```diff
@@ -38,7 +48,7 @@ protected OperationInternalBase(ClientDiagnostics clientDiagnostics, Response ra
         /// </code>
         /// </example>
         /// </summary>
-        public Response RawResponse { get; set; }
+        public abstract Response RawResponse { get; }
```
annelo-msft, line 51: Will this break anything that derives from this type?

sdk/core/Azure.Core/src/Shared/AsyncLockWithValue.cs
AlexanderSher: There is no change in functionality. It is just that `Azure.Identity` doesn't have nullable reference types analysis enabled, so it didn't require `_waiters` field to be marked as "can be null".

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
AlexanderSher: No. It is internal shared type, and we have only `OperationInternal` and `OperationInternalOfT` derived from it.

sdk/keyvault/Azure.Security.KeyVault.Certificates/src/DeleteCertificateOperation.cs
heaths: Before, the "secret" and "certificate" variables were added invariably:
image
But now they are only added if `RecoveryId` is null. One could argue that they are necessary if there's nothing to do, but I do worry about a regression in behavior. @KrzysztofCwalina or @tg-msft what do you think from a breaking change standpoint?

AlexanderSher: When `OperationInternal` is in the final state, no scope will be open, so these scope attributes will never be used. This is different from the previous behavior when calling `UpdateStatusAsync` would make at least one call to the server. However, that call would also override the response assigned by `_operationInternal.SetState`, so I don't think that previous behavior was correct.

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
```diff
@@ -176,15 +175,31 @@ public virtual Response WaitForCompletionResponse(CancellationToken cancellation
         /// <param name="cancellationToken">A <see cref="CancellationToken"/> controlling the request lifetime.</param>
         /// <returns>The last HTTP response received from the server, including the final result of the long-running operation.</returns>
         /// <exception cref="RequestFailedException">Thrown if there's been any issues during the connection, or if the operation has completed with failures.</exception>
-        public virtual Response WaitForCompletionResponse(TimeSpan pollingInterval, CancellationToken cancellationToken)
+        public Response WaitForCompletionResponse(TimeSpan pollingInterval, CancellationToken cancellationToken)
+            => WaitForCompletionResponseAsync(async: false, pollingInterval, cancellationToken).EnsureCompleted();
+
+        private async ValueTask<Response> WaitForCompletionResponseAsync(bool async, TimeSpan? pollingInterval, CancellationToken cancellationToken)
```
annelo-msft, line 181: This is where the synchronization happens.

sdk/core/Azure.Core/src/Shared/OperationInternal.cs
```diff
@@ -43,7 +42,20 @@ namespace Azure.Core
     /// </summary>
     internal class OperationInternal : OperationInternalBase
     {
-        private readonly IOperation _operation;
+        private readonly OperationInternal<VoidValue> _internalOperation;
```
annelo-msft, line 45: Please add a comment here regarding VoidValue

annelo-msft: include that this avoids introducing a fourth abstraction

sdk/core/Azure.Core/src/Shared/OperationInternal.cs
```diff
@@ -75,24 +87,46 @@ internal class OperationInternal : OperationInternalBase
...
+
+            public OperationWrapper(IOperation operation)
+            {
+                _operation = operation;
+            }
+
+            public async ValueTask<OperationState<VoidValue>> UpdateStateAsync(bool async, CancellationToken cancellationToken)
+            {
+                var state = await _operation.UpdateStateAsync(async, cancellationToken).ConfigureAwait(false);
+                return state.HasCompleted
```
annelo-msft, line 124: Could we convert this to an if/else statement for readability?

sdk/core/Azure.Core/src/Shared/OperationInternal.cs
```diff
@@ -75,24 +87,46 @@ internal class OperationInternal : OperationInternalBase
...
+        public override Response RawResponse => _internalOperation.RawResponse;
+
+        public override bool HasCompleted => _internalOperation.HasCompleted;
+
+        protected override async ValueTask<Response> UpdateStatusAsync(bool async, CancellationToken cancellationToken) =>
+            async ? await _internalOperation.UpdateStatusAsync(cancellationToken).ConfigureAwait(false) : _internalOperation.UpdateStatus(cancellationToken);
+
+        private readonly struct VoidValue { }
+
+        private readonly struct OperationWrapper : IOperation<VoidValue>
```
annelo-msft, line 112: Shoudl this be a class?

annelo-msft: Let's give this a name that implies it converts it to a generic, and/or have a good comment here -- wrapper could mean lots of things.

sdk/core/Azure.Core/src/Shared/OperationInternalOfT.cs
```diff
@@ -85,12 +99,25 @@ internal class OperationInternal<T> : OperationInternalBase
...
             _operation = operation;
-            RawResponse = rawResponse;
+            _rawResponse = rawResponse;
+            _stateLock = new AsyncLockWithValue<OperationState<T>>();
+        }
+
+        private OperationInternal(OperationState<T> finalState)
+            : base(finalState.RawResponse)
+        {
+            _operation = new FinalOperation();
```
annelo-msft, line 112: Please add a comment explaining this.

sdk/core/Azure.Core/src/Shared/OperationInternalOfT.cs
```diff
@@ -223,27 +247,53 @@ public Response<T> WaitForCompletion(TimeSpan pollingInterval, CancellationToken
...
 
-        /// <summary>
-        /// Sets the <see cref="OperationInternal{T}"/> state immediately.
-        /// </summary>
-        /// <param name="state">The <see cref="OperationState{T}"/> used to set <see cref="OperationInternalBase.HasCompleted"/> and other members.</param>
-        public void SetState(OperationState<T> state)
+        protected override async ValueTask<Response> UpdateStatusAsync(bool async, CancellationToken cancellationToken)
         {
-            if (state.HasCompleted && state.HasSucceeded)
+            using var asyncLock = await _stateLock.GetLockOrValueAsync(async, cancellationToken).ConfigureAwait(false);
```
annelo-msft, line 252: Let's add a comment here to explain the difference between this lock and the lock we keep for the value.

sdk/core/Azure.Core/src/Shared/OperationPoller.cs
```diff
@@ -6,96 +6,102 @@
...
         {
-            while (true)
+            if (async)
             {
-                Response response = updateStatus(cancellationToken);
-
-                if (hasCompleted())
+                await Task.Delay(delay, cancellationToken).ConfigureAwait(false);
+            }
+            else if (cancellationToken.CanBeCanceled)
```
annelo-msft, line 95: Let's add tests for this.

sdk/core/Azure.Core/src/Shared/AsyncLockWithValue.cs
```diff
@@ -1,25 +1,63 @@
...
+        {
+            _hasValue = true;
+            _value = value;
+        }
+
+        public bool TryGetValue(out T? value)
+        {
+            lock (_syncObj)
+            {
+                if (_hasValue)
```
KrzysztofCwalina, line 51: Can the value ever be removed? If not, you might want to first check if the value is there (without locking) and then only when it's not there, lock. In such case, I think you also need to make the value field volatile such that it gets synchronized.

KrzysztofCwalina: Of course, I am not even sure it's worth the trouble. Depends on how hot this path it.

sdk/core/Azure.Core/src/Shared/OperationPoller.cs
AlexanderSher: Done

sdk/textanalytics/Azure.AI.TextAnalytics/src/AnalyzeHealthcareEntitiesOperation.cs
```diff
@@ -243,8 +243,7 @@ public virtual void Cancel(CancellationToken cancellationToken = default)
             scope.Start();
             try
             {
-                ResponseWithHeaders<TextAnalyticsCancelHealthJobHeaders> response = _serviceClient.CancelHealthJob(new Guid(_jobId), cancellationToken);
-                _operationInternal.RawResponse = response.GetRawResponse();
+                _serviceClient.CancelHealthJob(new Guid(_jobId), cancellationToken);
```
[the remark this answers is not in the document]
AlexanderSher: Because `UpdateStateAsync` works with response from [`_serviceClient.HealthStatusAsync`](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/textanalytics/Azure.AI.TextAnalytics/src/AnalyzeHealthcareEntitiesOperation.cs#L344)
and it is `UpdateStateAsync` that produces [correct state](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/textanalytics/Azure.AI.TextAnalytics/src/AnalyzeHealthcareEntitiesOperation.cs#L372) for operation when it is canceled.

AlexanderSher:
> From my understanding, WaitForCompletion relies on UpdateStatus. If UpdateStatus is synchronized, do we need to synchronize WaitForCompletion too? Can we get a race condition if we don't do so?

We can't, but there is no reason to have multiple loops requesting server status - one in enough.

sdk/core/Azure.Core/src/Shared/OperationInternal.cs
```diff
@@ -109,23 +112,30 @@ private OperationInternal(OperationState finalState)
...
 
             public async ValueTask<OperationState<VoidValue>> UpdateStateAsync(bool async, CancellationToken cancellationToken)
             {
                 var state = await _operation.UpdateStateAsync(async, cancellationToken).ConfigureAwait(false);
-                return state.HasCompleted
-                    ? state.HasSucceeded
-                        ? OperationState<VoidValue>.Success(state.RawResponse, new VoidValue())
-                        : OperationState<VoidValue>.Failure(state.RawResponse, state.OperationFailedException)
-                    : OperationState<VoidValue>.Pending(state.RawResponse);
+                if (!state.HasCompleted)
```
annelo-msft, line 128: ❤

sdk/core/Azure.Core/src/Shared/AsyncLockWithValue.cs
AlexanderSher: Yeah, it's not a hot path where the type is used right now. I'd rather keep it simpler until we get any perf issues.

annelo-msft (approved):

sdk/core/Azure.Core/src/Shared/OperationInternalOfT.cs
AlexanderSher: We can't, but there is no reason to have multiple loops requesting server status - one in enough.

sdk/textanalytics/Azure.AI.TextAnalytics/src/AnalyzeHealthcareEntitiesOperation.cs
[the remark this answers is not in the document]
kinelski: Could you add this to the bug fixes in the Text Analytics changelog? This should be enough:
> In `AnalyzeHealthcareEntitiesOperation`, `Cancel` calls won't overwrite the response from `GetRawResponse` anymore.

kinelski (approved):
LGTM

sdk/textanalytics/Azure.AI.TextAnalytics/src/AnalyzeHealthcareEntitiesOperation.cs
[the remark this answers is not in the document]
AlexanderSher: Done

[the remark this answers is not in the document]
maririos: @AlexanderSher do you have the link to the issue you opened?

Issue #8561, opened 2019-11-01
[Service Bus] Detect connection status


giggio:
Right now it does not seem possible to detect the connection state with Azure Service Bus. There is no property that indicates if we are connected or if there is some connectivity issue (or other issue, like subscription deleted etc).

**Describe the solution you'd like**
I want a boolean property on the `SubscriptionClient` that tells me if it is connected. And I want an event on connection changes, i.e. `Connected` and `Disconnected`.

**Describe alternatives you've considered**
The alternative I'm using now uses Reflection, which is fragile. It does a little bit like this:

````csharp
var subscriptionClient = new SubscriptionClient(connStr, topic, subscription, ReceiveMode.PeekLock);
var conn = subscriptionClient.ServiceBusConnection;
FaultTolerantAmqpObject<AmqpConnection> connectionManager = (FaultTolerantAmqpObject<AmqpConnection>)typeof(ServiceBusConnection).GetProperty("ConnectionManager", BindingFlags.Instance | BindingFlags.NonPublic).GetMethod.Invoke(conn, null);
var amqpConnection = await connectionManager.GetOrCreateAsync(TimeSpan.FromSeconds(2));
var amqpSessionSettings = new AmqpSessionSettings { Properties = new Fields() };
while (true)
{
    var session = amqpConnection.CreateSession(amqpSessionSettings);
    var connected = false;
    try
    {
        await session.OpenAsync(TimeSpan.FromSeconds(2));
        await session.CloseAsync(TimeSpan.FromSeconds(2));
        connected = true;
    }
    catch (TimeoutException)
    {
    }
    Console.WriteLine(connected);
    Thread.Sleep(2000);
}
````

**Additional context**
I'll use this to inform the healthcheck of a container on a Kubernetes environment.

We shouldn't need to have to use reflection. And I get that the library is resilient and will reconnect. But I need to setup scaling, alerts and other infrastructure based on the fact that there is a worker doing what it is supposed to be doing. If for some reason a container gets disconnected from Service Bus, I need to know, I might setup infrastructure that will recreate it, setup alarms, etc. My queues/topics might be filling up, and I don't get to react to it.

AlexGhiondea:
/cc @binzywu.

Also, @jsquire to see how this feedback applies to our Track 2 libraries.

jfggdl:
@giggio, thanks for opening this issue. 
[...]

PR #28379, opened 2022-04-26
Add Connected/Disconnected events

JoshLove-msft:
Alternative approach to address https://github.com/Azure/azure-sdk-for-net/issues/8561

### DIFF
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpConnectionScope.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpConnectionScope.cs
@@ -153,12 +153,14 @@ namespace Azure.Messaging.ServiceBus.Amqp
         /// <summary>
         ///   The AMQP connection that is active for the current scope.
         /// </summary>
-        private FaultTolerantAmqpObject<AmqpConnection> ActiveConnection { get; }
+        internal FaultTolerantAmqpObject<AmqpConnection> ActiveConnection { get; }
 
         /// <summary>
         ///  The controller responsible for managing transactions.
         /// </summary>
         internal FaultTolerantAmqpObject<Controller> TransactionController { get; }
+
+        private readonly string _fullyQualifiedHostName;
 
         private readonly bool _useSingleSession;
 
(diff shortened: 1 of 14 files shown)


sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpConnectionScope.cs
```diff
@@ -168,37 +170,42 @@ public override bool IsDisposed
 
         private readonly object _syncLock = new();
         private readonly TimeSpan _operationTimeout;
+        private readonly ServiceBusClient _client;
```
jsquire, line 173: I'm not a fan of inverting the dependency flow here.   My goal was to have awareness flow in one direction - the client layer was aware of the transport layer but not vice-versa.   This makes it circular.    

Have we considered putting the connected/disconnected events on the connection scope and surfacing it through the transport client?    That would add a small bit more plumbing, but keep the dependency flow in a single direction.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Client/ServiceBusClient.cs
```diff
@@ -44,6 +43,59 @@ public virtual bool IsClosed
             private set => _closed = value;
         }
 
+        /// <summary>
+        /// An event that can be subscribed to for notification when the client connects to the service.
+        /// </summary>
+#pragma warning disable AZC0002
```
jsquire, line 49: nit:  Comments for what these warnings are would be helpful.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Client/ServiceBusClient.cs
```diff
@@ -44,6 +43,59 @@ public virtual bool IsClosed
...
+        /// <summary>
+        /// Gets whether or not the client is currently connected to the service. No action is required when the client temporarily
+        /// disconnects due to a transient network or service issue, as the client will attempt to re-establish the connection
+        /// automatically on the next operation. As such, it is not recommended that this property is checked before performing operations with the <see cref="ServiceBusClient"/>
+        /// or any of the other Service Bus types. This property should be used only for diagnostic information.
+        /// </summary>
+        public virtual bool IsConnected => Connection.IsConnected;
+
+        /// <summary>
+        ///
```
jsquire, line 76: nit:  Needs content.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Client/ServiceBusClient.cs
```diff
@@ -44,6 +43,59 @@ public virtual bool IsClosed
...
+        /// </summary>
+        /// <returns></returns>
+        protected internal virtual async Task OnConnectedAsync(ServiceBusConnectionEventArgs args)
+        {
+            if (ConnectedAsync != null)
+            {
+                await ConnectedAsync(args).ConfigureAwait(false);
+            }
```
jsquire, line 84:
```suggestion
            var handler = ConnectedAsync;
           
            if (handler != null)
            {
                await handler(args).ConfigureAwait(false);
            }
```
This is a race condition, since we're not blocking handlers being added/removed at any time.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Client/ServiceBusClient.cs
```diff
@@ -44,6 +43,59 @@ public virtual bool IsClosed
...
+        /// <summary>
+        ///
+        /// </summary>
+        /// <returns></returns>
+        protected internal virtual async Task OnDisconnectedAsync(ServiceBusConnectionEventArgs args)
+        {
+            if (ConnectedAsync != null)
+            {
+                await DisconnectedAsync(args).ConfigureAwait(false);
+            }
```
jsquire, line 96:
```suggestion
            var handler = DisconnectedAsync;
           
            if (handler != null)
            {
                await handler(args).ConfigureAwait(false);
            }
```
L93 was checking the wrong handler.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Client/ServiceBusConnectionEventArgs.cs
```diff
@@ -0,0 +1,39 @@
...
+    /// </summary>
+    public class ServiceBusConnectionEventArgs : EventArgs
+    {
+        /// <inheritdoc cref="ServiceBusClient.FullyQualifiedNamespace"/>
+        public string FullyQualifiedNamespace { get; }
+
+        /// <summary>
+        /// The proxy being used when communicating with the service.
+        /// </summary>
+        public IWebProxy Proxy { get; }
```
jsquire, line 20: Proxy and TransportType are interesting additions.    What's the scenario where they'd be useful?   I'd see the client identifier being more helpful to differentiate between multiple instances.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Client/ServiceBusClient.cs
```diff
@@ -44,6 +43,59 @@ public virtual bool IsClosed
...
+#pragma warning restore AZC0002
+#pragma warning restore AZC0003
+
+        /// <summary>
+        /// Gets whether or not the client is currently connected to the service. No action is required when the client temporarily
+        /// disconnects due to a transient network or service issue, as the client will attempt to re-establish the connection
+        /// automatically on the next operation. As such, it is not recommended that this property is checked before performing operations with the <see cref="ServiceBusClient"/>
+        /// or any of the other Service Bus types. This property should be used only for diagnostic information.
+        /// </summary>
+        public virtual bool IsConnected => Connection.IsConnected;
```
jsquire, line 73: We should also consider noting that this property does not indicate that read/publish operations are actively taking place.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/ServiceBusConnection.cs
```diff
@@ -69,18 +71,19 @@ internal class ServiceBusConnection : IAsyncDisposable
...
+        /// <param name="client"></param>
         /// <remarks>
         ///   If the connection string is copied from the Service Bus entity itself, it will contain the name of the desired Service Bus entity,
         ///   and can be used directly without passing the  name="entityName" />.  The name of the Service Bus entity should be
         ///   passed only once, either as part of the connection string or separately.
         /// </remarks>
         internal ServiceBusConnection(
             string connectionString,
-            ServiceBusClientOptions options)
+            ServiceBusClient client)
```
jsquire, line 82: I have the same hesitation here that I mentioned in the transport client; this changes the flow of dependencies from a single direction into circular.   Have we considered surfacing the connected/disconnected events from the transport client through the connection?

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Client/ServiceBusClientTests.cs
```diff
@@ -473,11 +473,6 @@ public void MetricsPropertyDoesNotThrowWhenEnabled()
         ///
         public class ReadableOptionsMock : ServiceBusClient
         {
```
jsquire, line 475: We should consider some unit tests for the handlers here, since we can fully control when they fire and what args they emit.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpConnectionScope.cs
JoshLove-msft: Yes, that is a fair point - was trying to just get things working quickly to validate the overall approach.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Client/ServiceBusConnectionEventArgs.cs
JoshLove-msft: Yeah, I considered that but was thinking it might be better to wait until we actually expose Identifier on the client. Mostly, I just wanted to add the event args type now so that we can evolve it in the future.

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Primitives/ServiceBusConnectionTests.cs
```diff
@@ -518,10 +519,11 @@ private class ObservableTransportClient : TransportClient
         {
             public bool WasCloseCalled;
 
-            public override Task CloseAsync(CancellationToken cancellationToken)
+            public override async Task CloseAsync(CancellationToken cancellationToken)
             {
                 WasCloseCalled = true;
-                return Task.CompletedTask;
+                await DisconnectedAsync(new ServiceBusConnectionEventArgs(ServiceEndpoint.Host, ServiceBusTransportType.AmqpTcp, null));
```
JoshLove-msft, line 525: Need to update to add explicit tests for these.

PR #28382, opened 2022-04-26
generated sdk from #18819 swagger PR

IliasKhan:
https://github.com/Azure/azure-rest-api-specs/pull/18819/files


sdk/sqlmanagement/Microsoft.Azure.Management.Sql/src/Generated/Models/UnlinkParameters.cs
```diff
@@ -1,53 +0,0 @@
-// <auto-generated>
```
archerzz, line 1: This is breaking change.  Here are 2 options:
- Add manual codes to provide backward compatibility, see compute for reference: https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/compute/Microsoft.Azure.Management.Compute/src/Customizations
- Or you need to bump the major version

archerzz (changes requested):
@IliasKhan Please check the release process, especially [this step](https://dev.azure.com/azure-sdk/internal/_wiki/wikis/internal.wiki/78/Mgmt-.NET-SDK-release-process?anchor=sdk-generation-from-updated-spec). There are a few files need to be updated:
- `.csproj` file
- `AzSdk.RP.props`
- `AssemblyInfo.cs`

Thanks.

PR #28388, opened 2022-04-26
Regen for remover

### LINKED ISSUE #2157: Remove the models that are not even internally referenced
We have some logic to make the models that are not publicly referenced `internal`.
We need to remove the models that are not even internally referenced.

ArcturusZhang:
Fixes https://github.com/Azure/autorest.csharp/issues/2157


ArcturusZhang:
The CI failures are all complaining about LongRunningOperation method - I believe this should be introduced by recent LRO changes such as this: https://github.com/Azure/autorest.csharp/pull/2213

ArcturusZhang:
Updated the autorest.csharp version to a newer one, this should contain the missing files and resolve the compile issues.

Yao725 (approved):
I roughly checked `Cdn`, `DeviceUpdate` and `Dynatrace`, code change LGTM, thanks!

weshaggard:
@ArcturusZhang how often do you think these types of changes will need to be made? I ask because it causes a large amount of CI jobs to be triggered which can cause what we call is a build storm which can consume a lot of our resources and cause delays in processing jobs.

PR #28400, opened 2022-04-26
example of no dependency impl

### LINKED ISSUE #28185: Cross RP dependency - Orphaned models and operations between Network and Compute need dependency resolved
Problem
•	There are APIs defined under different provider, like [this](https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2FAzure%2Fazure-rest-api-specs%2Fblob%2Fmain%2Fspecification%2Fnetwork%2Fresource-manager%2FMicrosoft.Network%2Fstable%2F2021-05-01%2FvmssNetworkInterface.json&data=05%7C01%7Cmicnash%40microsoft.com%7C6c1b15ddb24346b1ea3408da1953a72e%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C637850144357805836%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=dXoMZq9I48HNTd%2BaCbEixtU6UzvavZYTxWbyUcUTYhc%3D&reserved=0).
•	Those API resources are orphans, so that the hierarchy API cannot be generated. But the extension API are generated.
o	For example, this API is orphan because VirtualMachineScaleSet and VirtualMachine are not defined in “network”.
o	I haven’t tried if APIs under different provider are not orphans. I guess they can be generated. If not, that is another issue.

Solution
Packaging
•	We cannot generate the API in the RP which should own the provider name (e.g. “compute”), because API Version could be different.
•	We can generate the API in where it’s defined but choose different packaging strategy.
o	We can just generate the API in “network”, and then mark “network” depending on “compute”.
	Prons: simple because the API is defined in “network”
	Crons: “network” has to depend on “compute”
•	There is a potential risk that two RPs have cross dependency (e.g. both have API defined under other’s provider name)
o	Or we can generate the API in a separate package, and then mark that package pending on both “network” and “compute”.
	Prons:
•	“network” do not depend on “compute”
•	Eliminate the cross dependency
	Crons:
•	Make the dependency more complex
•	People might have difficult to find those API (especially for track1 users who get used to)

How to generate the codes
Take "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/microsoft.Compute/virtualMachineScaleSets/{virtualMachineScaleSetName}/virtualMachines/{virtualmachineIndex}/networkInterfaces" as example.
•	Create a new directive “virtual-resources”
o	Each entry is a pair of path and resource name. For example, virtual machine scale set should have:
	Path: “/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/microsoft.Compute/virtualMachineScaleSets”
	Name: “VirtualMachineSacleSet”
•	For generator:
o	It will assume a virtual resource exists, so that the leaf resources can be connected to the hierarchy (e.g., “networkInterfaces”)
o	It will only generate codes of a virtual resource if its child resource is non-virtual (e.g., “VirtualMachineScaleSet” will not be generated, but “VirtualMachine” will be generated)
	It will only generate codes to get child resources (e.g., “VirtualMachine.GetNetworkInterfaces”)

m-nash:
Fixes https://github.com/Azure/azure-sdk-for-net/issues/28185


sdk/network/Azure.ResourceManager.Network/src/Custom/VirtualMachineScaleSetExtensionResource.cs
```diff
@@ -0,0 +1,125 @@
...
+using System.Threading.Tasks;
+using Azure.Core;
+using Azure.Core.Pipeline;
+
+namespace Azure.ResourceManager.Network
+{
+    /// <summary>
+    /// Extensions for VirtualMachineScaleSet.
+    /// </summary>
+    public class VirtualMachineScaleSetExtensionResource : ArmResource
```
ArcturusZhang, line 19: according to [this comment](https://github.com/Azure/azure-sdk-for-net/issues/28185#issuecomment-1116659864) this should rename to `VirtualMachineScaleSetNetworkResource`

ArcturusZhang, line 19: according to [this comment](https://github.com/Azure/azure-sdk-for-net/issues/28185#issuecomment-1116659864) this should rename to `VirtualMachineScaleSetNetworkResource`

PR #28403, opened 2022-04-26
Add Event Hubs integration to Avro sample

JoshLove-msft:


sdk/schemaregistry/Microsoft.Azure.Data.SchemaRegistry.ApacheAvro/tests/SchemaRegistryAvroObjectSerializerLiveTests.cs
```diff
@@ -212,6 +214,24 @@ public async Task CanUseEncoderWithEventData()
...
+            var eventHubName = TestEnvironment.EventHubName;
+            var credential = TestEnvironment.Credential;
+#endif
+            await using var producer = new EventHubProducerClient(fullyQualifiedNamespace, eventHubName, credential);
+            using EventDataBatch eventBatch = await producer.CreateBatchAsync();
+            if (!eventBatch.TryAdd(eventData))
+            {
+                throw new InvalidOperationException("Payload is too large to be sent to Event Hubs service.");
+            }
+            await producer.SendAsync(eventBatch);
```
jsquire, line 234:
```suggestion
            // because we're sending a single event, we'll avoid the overhead of creating a batch. 
            await producer.SendAsync(new EventData[] { eventData });
```

I'm thinking it would be simpler to show the enumerable overload.   Since we've got a single event, we'll know exactly what failed if it was too large, so the batch doesn't buy us anything.  We probably want to consider tweaking the comment a bit, though.

sdk/schemaregistry/test-resources.json
```diff
@@ -85,6 +108,10 @@
     "SCHEMAREGISTRY_ENDPOINT": {
       "type": "string",
       "value": "[variables('schemaRegistryEndpoint')]"
+    },
+    "EVENTHUB_NAME": {
```
jsquire, line 112: Should this be `SCHEMAREGISTRY_EVENTHUB_NAME` ?

jsquire (approved):

PR #28405, opened 2022-04-26
Fix nullref exception when ScaledAssetDimensionsWrapper is null

RamonArguelles:

### DIFF
--- a/sdk/objectanchors/Azure.MixedReality.ObjectAnchors.Conversion/src/Customizations/Models/AssetConversionConfiguration.cs
+++ b/sdk/objectanchors/Azure.MixedReality.ObjectAnchors.Conversion/src/Customizations/Models/AssetConversionConfiguration.cs
@@ -61,17 +61,17 @@ namespace Azure.MixedReality.ObjectAnchors.Conversion
         /// <summary>
         /// Dimensions of the asset.
         /// </summary>
-        public System.Numerics.Vector3? AssetDimensions { get => AssetDimensionsWrapper == null ? null : AssetDimensionsWrapper; }
+        public System.Numerics.Vector3? AssetDimensions { get => AssetDimensionsWrapper == null ? null : (System.Numerics.Vector3)AssetDimensionsWrapper; }
 
         /// <summary>
         /// BoundingBoxCenter of the asset.
         /// </summary>
-        public System.Numerics.Vector3? BoundingBoxCenter { get => BoundingBoxCenterWrapper == null ? null : BoundingBoxCenterWrapper; }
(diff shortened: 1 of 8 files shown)


RamonArguelles:
When a job has been submitted, but hasn't finished processing yet, ScaledAssetDimensionsWrapper is null. Same will be for a job that has finished, but failed before scaled asset dimensions could be calculated (for example, when the file format is invalid). 

ScaledAssetDimensionsWrapper is an internal property. When the customer tries to access the public property, ScaledAssetDimensions, the logic that runs in the code is:

        public System.Numerics.Vector3? ScaledAssetDimensions { get => ScaledAssetDimensionsWrapper == null ? null : ScaledAssetDimensionsWrapper; }

This code should be simply returning null. However, I suspect there is an optimization happening here, and the code is likely being optimized, even on Debug, to do this:

        public System.Numerics.Vector3? ScaledAssetDimensions { get => ScaledAssetDimensionsWrapper; }

I know this, because, in the current implementation, the implicit conversion operator from Azure.MixedReality.ObjectAnchors.Conversion.Models.Vector3 to System.Numerics.Vector3 is getting invoked:

        public static implicit operator System.Numerics.Vector3(Vector3 v) => v.data;

And this is causing a null-ref, since Vector3 v is null. The implicit conversion operator should NOT be called in this case. But is, and the only explanation I can give is my suggestion above: code is being optimized.

With my proposed fix, I'm forcing an explicit cast. This way, the compiler does not apply the optimization, and leaves the code as it was actually intended. The end result: the implicit operator is no longer invoked when the value is null, and it is invoked explicitly when the value isn't null.

I've tested this with our sample code, and we behave as expected in both scenarios now.

RamonArguelles:
@craigktreasure

craigktreasure:
If `ScaledAssetDimensionsWrapper` can be `null`, then we should mark it as nullable, right? That would better represent the value and allow you to instead call `HasValue` on it to reliably determine if it is `null` or not. The cast seems a bit dodgy to me.

craigktreasure (changes requested):
If `ScaledAssetDimensionsWrapper` can be `null`, then we should mark it as nullable, right? That would better represent the value and allow you to instead call `HasValue` on it to reliably determine if it is `null` or not. The cast seems a bit dodgy to me.

craigktreasure:
> If `ScaledAssetDimensionsWrapper` can be `null`, then we should mark it as nullable, right? That would better represent the value and allow you to instead call `HasValue` on it to reliably determine if it is `null` or not. The cast seems a bit dodgy to me.

Talked offline. I've confused the model `Vector3` with `System.Numerics.Vector3`. The model `Vector3` type has an implicit converter.

Should we instead make the converter explicit rather than implicit?

craigktreasure (approved):

PR #28408, opened 2022-04-27
Playbackonly for MHSM mgmt test

### LINKED ISSUE #26841: Some Key Vault tests running too slowly
Some management plane tests are running too slowly, currently `ManagedHsmListKeys` which creates an HSM (very long, expensive operation), checks keys, and deletes them.

While the creation is a long-enough operation we may need to work out something different (there's a 10s global max or the tests fail), the deletes are likely unnecessary and removing or refactoring those out could help bring the time down.

The data-plane schedules deletes to happen as soon as a resource is created. This also has the benefit that any test failure won't be able to clean up the resources.

Take a look at https://github.com/Azure/azure-sdk-for-net/blob/23771fc956029d35b225adb4b8b50b5152a94fe1/sdk/keyvault/Azure.Security.KeyVault.Keys/tests/KeysTestBase.cs#L111-L133 and the functions within that class we call, and a usage example at https://github.com/Azure/azure-sdk-for-net/blob/23771fc956029d35b225adb4b8b50b5152a94fe1/sdk/keyvault/Azure.Security.KeyVault.Keys/tests/KeyClientLiveTests.cs#L37-L47 where we call `RegisterForCleanup()`. Not only does this work for test failures, but doesn't count against individual test time.

Alternatively - and we're considering this for data plane - you could just not delete. It's slow and often unnecessary. Where you tend to list resources, you could instead assert existence against a list of known IDs and ignore all others.

For now I need to mark the test as `[Ignore]` because it's blocking release and failing pretty consistently on MacOS for some reason, but I'll link back to this bug for tracking.

ArthurMa1978:


Yao725 (approved):
Do we need to mark all the mgmt plane KeyVault tests as `PlaybackOnly`, or is this one enough?

ArthurMa1978:
> Do we need to mark all the mgmt plane KeyVault tests as `PlaybackOnly`, or is this one enough?

Good question, compare with vault 2 reasons support to disable it only for daily run
- The time of HSM testing is extremly longer than valut, so this might make daily run timout.
- The HSM is expensive and have resource limitation (5 per region), so should not be tested too frequently.

Issue #28330, opened 2022-04-21
Hook batch endpoints from TextAnalyticsClient to use the new Language API endpoints


maririos:
The endpoints are:

- [x] Healthcare
- [x] AnalyzeActions

Build should be green, there shouldn't be any modifications to Public API, and at least some tests need to pass per endpoint

maririos:
Need to solve https://github.com/Azure/azure-sdk-for-net/pull/28423#discussion_r860172827

PR #28423, opened 2022-04-27
[Text Analytics] Legacy Service Client

jsquire:
# Summary

The focus of these changes is to introduce the service client responsible for interacting with the legacy Text Analytics REST API.  This client accepts and returns types from the SDK's public API, translating them as needed for service operations.

## Next steps for legacy support

- Update TextAnalyticsClient to delegate to the correct service client
- Shift Nisha's work over to the language service client


sdk/textanalytics/Azure.AI.TextAnalytics/src/AnalyzeActionsOperation.cs
```diff
@@ -18,7 +17,7 @@ namespace Azure.AI.TextAnalytics
     public class AnalyzeActionsOperation : PageableOperation<AnalyzeActionsResult>, IOperation<AsyncPageable<AnalyzeActionsResult>>
     {
         /// <summary>Provides communication with the Text Analytics Azure Cognitive Service through its REST API.</summary>
-        //private readonly TextAnalyticsRestClient _serviceClient;
+        private readonly ServiceClient _serviceClient;
```
jsquire, line 20: Not sure if this pattern will ultimately work out. I shifted it for now, as it gives us a neutral client surface. The previous implementation was commented out, so this has no functionality impact just yet.

sdk/textanalytics/Azure.AI.TextAnalytics/src/Azure.AI.TextAnalytics.csproj
```diff
@@ -17,15 +17,18 @@
...
-    <Compile Include="$(AzureCoreSharedSources)PageResponseEnumerator.cs" LinkBase="Shared" />
-    <Compile Include="$(AzureCoreSharedSources)AzureKeyCredentialPolicy.cs" LinkBase="Shared" />
+    <Compile Include="../../Azure.AI.TextAnalytics.Legacy.Shared/src/Generated/**/*" LinkBase="Shared/Legacy/Generated" />
+    <Compile Include="../../Azure.AI.TextAnalytics.Legacy.Shared/src/Custom/**/*" LinkBase="Shared/Legacy/Custom" />
   </ItemGroup>
 
+  <!-- Import Azure.Core shared source -->
   <ItemGroup>
-    <Folder Include="Models\" />
+    <Compile Include="$(AzureCoreSharedSources)AzureResourceProviderNamespaceAttribute.cs" LinkBase="Shared/Core" />
```
jsquire, line 28: The LRO shared source was already using Shared/Core as a base; this update unifies it so that all the Azure.Core shared source is hanging out in the same place.

sdk/textanalytics/Azure.AI.TextAnalytics/src/CategorizedEntity.cs
```diff
@@ -14,15 +14,20 @@ namespace Azure.AI.TextAnalytics
     public readonly struct CategorizedEntity
     {
         internal CategorizedEntity(Entity entity)
+            : this(entity.Text, entity.Category, entity.Subcategory, entity.ConfidenceScore, entity.Offset, entity.Length)
+        {
+        }
+
+        internal CategorizedEntity(string text, string category, string subcategory, double confidenceScore, int offset, int length)
```
jsquire, line 21: I intentionally avoided using the legacy type directly in constructors. My goal was to not allow the legacy surface to get deeply embedded so that it can be removed at a later point, as there are potentially plans to sunset the associated REST service.

sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/LegacyServiceClient.cs
```diff
@@ -16,87 +19,1378 @@ namespace Azure.AI.TextAnalytics.ServiceClients
     ///
     internal class LegacyServiceClient : ServiceClient
```
jsquire, line 20: Essentially, this is the `TextAnalyticsClient` from the previous version.   Some types have been remapped as the legacy generated surface is entirely internalized.

sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/LegacyServiceClient.cs
```diff
@@ -16,87 +19,1378 @@ namespace Azure.AI.TextAnalytics.ServiceClients
...
+        private static readonly AnalyzeSentimentOptions DefaultAnalyzeSentimentOptions = new();
+
+        internal readonly TextAnalyticsRestClient _serviceRestClient;
+        internal readonly ClientDiagnostics _clientDiagnostics;
+
+        private readonly Uri _baseUri;
+        private readonly string _defaultLanguage;
+        private readonly string _defaultCountryHint;
+
+        public LegacyServiceClient(Uri endpoint, TokenCredential credential, string defaultLanguage, string defaultCountryHint, string authorizationScope, string serviceVersion, ClientOptions options)
```
jsquire, line 34: I waffled on whether to avoid passing `TextAnalyticsClientOptions` directly here, as this is not the `TextAnalyticsClient` type.   For now, I chose to have the actual client prepare scopes and specify defaults rather than having both service clients (legacy/language) duplicate doing so.

sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.cs
```diff
@@ -10,58 +10,29 @@
 
 namespace Azure.AI.TextAnalytics
 {
-    internal static class Transforms
+    internal static partial class Transforms
     {
-        #region Common
-
-        public static readonly Regex _targetRegex = new Regex("#/tasks/(keyPhraseExtractionTasks|entityRecognitionPiiTasks|entityRecognitionTasks|entityLinkingTasks|sentimentAnalysisTasks|extractiveSummarizationTasks|customSingleClassificationTasks|customMultiClassificationTasks|customEntityRecognitionTasks)/(\\d+)", RegexOptions.Compiled, TimeSpan.FromSeconds(2));
+        #region Needs Review
```
jsquire, line 15: We should consider auditing anything with the `Internal` suffix to see if it is still needed with the language API - they're not used by the legacy integration.

sdk/textanalytics/Azure.AI.TextAnalytics.Legacy.Shared/src/GeneratorStubs/InternalTypes.cs
```diff
@@ -32,7 +32,7 @@ internal partial class KeyPhrasesTaskParameters { }
     internal partial class SentimentAnalysisTaskParameters { }
 
     [CodeGenModel("PiiCategory")]
-    internal readonly partial struct PiiEntityCategory { }
+    internal readonly partial struct PiiEntityLegacyCategory { }
```
jsquire, line 35: This was needed because there's a `PiiEntityCategory` in the `Azure.AI.TextAnalytics` namespace.  Because the legacy generated code does not fully qualify the type name, the compiler was matching the wrong class in the main project.

sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.Legacy.cs
```diff
@@ -0,0 +1,644 @@
...
+        internal static TextSentiment ConvertToTextSentiment(Legacy.Models.DocumentSentimentValue sentiment) =>
+            sentiment switch
+            {
+                Legacy.Models.DocumentSentimentValue.Neutral => TextSentiment.Neutral,
+                Legacy.Models.DocumentSentimentValue.Positive => TextSentiment.Positive,
+                Legacy.Models.DocumentSentimentValue.Negative => TextSentiment.Negative,
+                _ => throw new NotSupportedException($"The document sentiment, { sentiment }, is not supported for conversion.")
+            };
+
+        //TODO:  Check this conversion.  Seems to match: https://github.com/Azure/azure-sdk-for-net/blob/Azure.AI.TextAnalytics_5.2.0-beta.2/sdk/textanalytics/Azure.AI.TextAnalytics/src/DocumentSentiment.cs#L56
```
jsquire, line 136: @maririos:   I could use your expertise here.  From what I can tell, this is the same conceptual transform that was being performed.   But, the existing transformation is odd and I'm not sure that I'm interpreting it correctly.

sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.cs
```diff
@@ -10,58 +10,29 @@
...
-                return new TextAnalyticsError(innerError.Code.ToString(), innerError.Message, innerError.Target);
-            }
+        #region Common
 
-            return new TextAnalyticsError(errorCode, message, target);
-        }
+        public static readonly Regex _targetRegex = new Regex("#/tasks/(keyPhraseExtractionTasks|entityRecognitionPiiTasks|entityRecognitionTasks|entityLinkingTasks|sentimentAnalysisTasks|extractiveSummarizationTasks|customSingleClassificationTasks|customMultiClassificationTasks|customEntityRecognitionTasks)/(\\d+)", RegexOptions.Compiled, TimeSpan.FromSeconds(2));
 
-        internal static List<TextAnalyticsError> ConvertToErrors(IReadOnlyList<TextAnalyticsErrorInternal> internalErrors)
+        internal static TextAnalyticsError ConvertToError(DocumentError error)
```
jsquire, line 29: This removed some potentially unused intermediate variables, for efficiency.

sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.cs
```diff
@@ -191,13 +157,17 @@ internal static ExtractKeyPhrasesResultCollection ConvertToExtractKeyPhrasesResu
 
         #region Recognize Entities
 
-        internal static List<CategorizedEntity> ConvertToCategorizedEntityList(List<Entity> entities)
-            => entities.Select((entity) => new CategorizedEntity(entity)).ToList();
+        internal static List<CategorizedEntity> ConvertToCategorizedEntityList(IList<Entity> entities)
```
jsquire, line 160: This and what follows was just unrolling LINQ for efficiency.

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
```diff
@@ -157,8 +158,8 @@ public virtual async Task<Response<DetectedLanguage>> DetectLanguageAsync(string
                 if (languageDetection.Results.Errors.Count > 0)
                 {
                     // only one document, so we can ignore the id and grab the first error message.
-                    var error = Transforms.ConvertToError(languageDetection.Results.Errors.FirstOrDefault().Error);
-                    throw await _clientDiagnostics.CreateRequestFailedExceptionAsync(response, new ResponseError(error.ErrorCode.ToString(), error.Message), CreateAdditionalInformation(error)).ConfigureAwait(false);
+                    var error = languageDetection.Results.Errors.First().Error;
```
jsquire: I tripped over these after a rebase due to a name collision.   In looking at it, we didn't need to convert - we have the available information.  Since we already know there's at least one error, I also removed the `FirstOrDefault` in favor of just grabbing the first element.

sdk/textanalytics/Azure.AI.TextAnalytics/src/AnalyzeActionsOperation.cs
maririos: hmm interesting. I forgot that LROs are like little mini clients. Def need to think about this one more

sdk/textanalytics/Azure.AI.TextAnalytics/src/AnalyzeActionsOperation.cs
```diff
@@ -157,28 +156,28 @@ public AnalyzeActionsOperation(string operationId, TextAnalyticsClient client)
...
-        //    _showStats = showStats;
-        //    _operationInternal = new OperationInternal<AsyncPageable<AnalyzeActionsResult>>(_diagnostics, this, rawResponse: null);
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AnalyzeActionsOperation"/> class.
+        /// </summary>
+        /// <param name="serviceClient">The client for communicating with the Form Recognizer Azure Cognitive Service through its REST API.</param>
+        /// <param name="diagnostics">The client diagnostics for exception creation in case of failure.</param>
+        /// <param name="operationLocation">The address of the long-running operation. It can be obtained from the response headers upon starting the operation.</param>
+        /// <param name="idToIndexMap">A map of identifiers to their associated index.</param>
+        /// <param name="showStats"><c>true</c> to show statistics; otherwise, <c>false</c>.</param>
```
maririos, line 166: thanks for adding this haha I guess we got lazy added descriptions :(

sdk/textanalytics/Azure.AI.TextAnalytics/src/CategorizedEntity.cs
maririos: what do you mean with sunset? :)

jsquire: Sunset = retirement on a incremental plan.

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
```diff
@@ -157,8 +158,8 @@ public virtual async Task<Response<DetectedLanguage>> DetectLanguageAsync(string
                 if (languageDetection.Results.Errors.Count > 0)
                 {
                     // only one document, so we can ignore the id and grab the first error message.
-                    var error = Transforms.ConvertToError(languageDetection.Results.Errors.FirstOrDefault().Error);
-                    throw await _clientDiagnostics.CreateRequestFailedExceptionAsync(response, new ResponseError(error.ErrorCode.ToString(), error.Message), CreateAdditionalInformation(error)).ConfigureAwait(false);
+                    var error = languageDetection.Results.Errors[0].Error;
```
maririos, line 161: Although technically correct, in TA we always surface the `InnerError` information instead of the first level `Error` .
See for example:
```
"error": {
        "code": "InvalidRequest",
        "message": "Invalid document in request.",
        "innererror": {
            "code": "InvalidDocument",
            "message": "Request contains duplicated Ids. Make sure each document has a unique Id."
        }
    }
```

The real information is in the `InnerError` property, which is why the transform was looking for it and overwriting the `Error` type.

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
```diff
@@ -2425,12 +2426,10 @@ private MultiLanguageAnalysisInput ConvertToMultiLanguageInputs(IEnumerable<Text
         private LanguageInput ConvertToLanguageInput(string document, string countryHint, int id = 0)
             => new LanguageInput($"{id}", document) { CountryHint = countryHint ?? _options.DefaultCountryHint };
 
-        private static IDictionary<string, string> CreateAdditionalInformation(TextAnalyticsError error)
-        {
-            if (string.IsNullOrEmpty(error.Target))
-                return null;
-            return new Dictionary<string, string> { { "Target", error.Target } };
-        }
+        private static IDictionary<string, string> CreateAdditionalInformation(Error error) =>
```
maririos, line 2429: I can't think, help! what is the difference between this and what it had before

jsquire: The incoming type changed from TextAnalytics to Error.    😄

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
jsquire: hrmm....   yeah, I overlooked that.   I'll fix that up.

sdk/textanalytics/Azure.AI.TextAnalytics/src/AnalyzeActionsOperation.cs
jsquire: Yeah.... this was a surprise that'll need a bit more design.

sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.Legacy.cs
maririos: Could you remind me what the need to add this code here instead of leaving it as is?

sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.Legacy.cs
```diff
@@ -0,0 +1,641 @@
...
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Azure.AI.TextAnalytics
+{
+    /// <summary>
+    ///   This portion of the class defines the transformations needed for interacting
+    ///   with the legacy Text Analytics REST API generated types.
+    /// </summary>
+    internal static partial class Transforms
```
maririos, line 14: ideally, this class should be the same as https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.cs right?
I mean, I can tell it is not, so can you help me understand the disconnect?

sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.Legacy.cs
jsquire: The types changed and the structure didn't line up in the way that I expected.    Did I overlook something that would make that unnecessary?

sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.Legacy.cs
jsquire: The `main` branch mixes type renames and such from the core namespace, sharing the public API with the generated code.   In this form, the legacy API is generated as purely internal and isolated.   Mostly, types moved namespaces and occasionally reverted to the pure generated name.   (renames were mostly things previously hit with the `Internal` suffix)

sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.cs
```diff
@@ -10,58 +10,38 @@
...
-                return new TextAnalyticsError(innerError.Code, message, target);
-            }
+        #region Common
 
-            return new TextAnalyticsError(errorCode, message, target);
-        }
+        public static readonly Regex _targetRegex = new Regex("#/tasks/(keyPhraseExtractionTasks|entityRecognitionPiiTasks|entityRecognitionTasks|entityLinkingTasks|sentimentAnalysisTasks|extractiveSummarizationTasks|customSingleClassificationTasks|customMultiClassificationTasks|customEntityRecognitionTasks)/(\\d+)", RegexOptions.Compiled, TimeSpan.FromSeconds(2));
 
-        internal static TextAnalyticsError ConvertToError(Error error)
+        internal static TextAnalyticsError ConvertToError(DocumentError error)
```
maririos, line 29: from looking at the PR is hard to see what the difference between `ConvertToError(DocumentError error)` and ConvertToError(Error error)`` is. Do you happen to know?

jsquire: Not sure that I understand.  They're different types with no common ancestor, so they needed to be overloaded unless we're going to unify the types.

sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.cs
```diff
@@ -159,11 +139,6 @@ internal static AnalyzeSentimentResultCollection ConvertToAnalyzeSentimentResult
 
         #region KeyPhrases
 
-        //internal static KeyPhraseCollection ConvertToKeyPhraseCollection(DocumentKeyPhrases documentKeyPhrases)
```
maririos, line 162: sorry Jesse. :( 
The reason why we delete this is because we are using the same method that lives in `Transforms.Legacy` , right?

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
jsquire: The type change here is being reverted due to restoring the inner error extraction that I accidentally removed.

sdk/textanalytics/Azure.AI.TextAnalytics/src/AnalyzeActionsOperation.cs
maririos: Added comment to issue https://github.com/Azure/azure-sdk-for-net/issues/28330

sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.cs
jsquire: No.... this was a mistake.    😄

jsquire: Restored.

sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.cs
maririos: I believe I understand what is happening here. `DocumentError` contains `Error`. It also has an `Id` property. in our client logic, we always look at the `Id` of the document first and then assign to that `Document` the appropriate `Error`.
so if this is true, no one should call `ConvertToError(DocumentError error)` .

Is hard to see if this is true from the PR. Can you see if someone is actually calling that method?

sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.Legacy.cs
```diff
@@ -0,0 +1,641 @@
...
+                sentences.Add(new SentenceSentiment(
+                    ConvertToTextSentiment(sentiment.Sentiment),
+                    sentiment.Text,
+                    sentiment.ConfidenceScores.Positive,
+                    sentiment.ConfidenceScores.Neutral,
+                    sentiment.ConfidenceScores.Negative,
+                    sentiment.Offset,
+                    sentiment.Length,
+                    ConvertToSentenceOpinions(sentiment.Assessments)));
+            }
```
maririos, line 185: nit: would be easier to read 
```suggestion
            foreach (var sentence in legacySentiments)
            {
                sentences.Add(new SentenceSentiment(
                    ConvertToTextSentiment(sentence.Sentiment),
                    sentence.Text,
                    sentence.ConfidenceScores.Positive,
                    sentence.ConfidenceScores.Neutral,
                    sentence.ConfidenceScores.Negative,
                    sentence.Offset,
                    sentence.Length,
                    ConvertToSentenceOpinions(sentence.Assessments)));
            }
```

sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.cs
jsquire: This looks to be a hold-over from before Nisha's latest merge that I reformatted as I was adding the overloads.   _([src](https://github.com/Azure/azure-sdk-for-net/blob/08e2e0401ece971c0d560991f38c4373b6ad7629/sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.cs#L36))_    It does not appear that there are any references to it.   I'll remove it for now; it's simple enough to add back in if we end up needing it somewhere in the future.

sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.Legacy.cs
jsquire: Changed all `sentiment` references in the method to `sentence`.

sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.Legacy.cs
maririos: what I see missing is how are we treating the [TargetRelation](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cognitiveservices/data-plane/Language/preview/2022-03-01-preview/textanalytics.json#L1386) .
This type has a json pointer that correlates the assessment to a specific sentence which is what the code in [here](https://github.com/Azure/azure-sdk-for-net/blob/Azure.AI.TextAnalytics_5.2.0-beta.2/sdk/textanalytics/Azure.AI.TextAnalytics/src/SentenceSentiment.cs#L86) is doing.

jsquire: @maririos:  Thank you very much for all of your help in sorting out the issues here.  Based on our conversation, I refactored and I think we've now got the correct transform.

sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.Legacy.cs
```diff
@@ -130,58 +134,51 @@ sentiment switch
...
+                foreach (var relation in target.Relations)
+                {
+                    if (relation.RelationType == Legacy.Models.TargetRelationType.Assessment)
+                    {
+                        assessments.Add(ResolveAssessmentReference(allSentences, relation.Ref));
+                    }
+                }
 
-            var opinions = new List<SentenceOpinion>(targetList.Count);
+                var sentiment = new TargetSentiment(
```
maririos, line 153: nit: might be better to use `target`

eng/.docsettings.yml
```diff
@@ -72,15 +72,13 @@ known_presence_issues:
...
   - ['sdk/search/Microsoft.Azure.Search.Data','Track 1']
-  - ['sdk/search/Microsoft.Azure.Search.Service','Track 1']
+  - ['sdk/search/Microsoft.Azure.Search.Service','Track 1']  
   - ['sdk/operationalinsights/Microsoft.Azure.OperationalInsights','Track 1']
   - ['sdk/search/Microsoft.Azure.Search','Track 1']
   - ['sdk/search/Microsoft.Azure.Search.Common/CHANGELOG.md','Track 1']
   - ['sdk/search/Microsoft.Azure.Search.Data/CHANGELOG.md','Track 1']
   - ['sdk/search/Microsoft.Azure.Search.Service/CHANGELOG.md','Track 1']
-
-  - ['sdk/textanalytics/Azure.AI.TextAnalytics.Legacy.Shared/CHANGELOG.md','Only used for code sharing; not a shipping library']
```
maririos: ohh interesting. I thought this was needed in order to make the build work :|

sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.cs
```diff
@@ -164,7 +135,7 @@ internal static AnalyzeSentimentResultCollection ConvertToAnalyzeSentimentResult
         //    return new KeyPhraseCollection(documentKeyPhrases.KeyPhrases.ToList(), ConvertToWarnings(documentKeyPhrases.Warnings));
         //}
 
-        internal static ExtractKeyPhrasesResultCollection ConvertToExtractKeyPhrasesResultCollection(KeyPhraseResult results, IDictionary<string, int> idToIndexMap)
+    internal static ExtractKeyPhrasesResultCollection ConvertToExtractKeyPhrasesResultCollection(KeyPhraseResult results, IDictionary<string, int> idToIndexMap)
```
maririos, line 138: tiny nit: spaces

maririos (approved):
🚢

eng/.docsettings.yml
jsquire: So did I.... but apparently, if we set `<IsTestSupportLibrary>` AND `<IsShippingLibrary>` in the project file to the right things, we don't need this.

sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.Legacy.cs
jsquire: Target is used in the outer loop.   I changed to `targetSentiment`.

PR #28440, opened 2022-04-28
IoT Central Track2 Tests

### LINKED ISSUE #28876: Initial Azure.ResourceManager.IotCentral

chgennar:
This PR adds some initial tests for the Track2 Iot Central management plane SDK.

Fixes: #28876


sdk/iotcentral/Azure.ResourceManager.IotCentral/Directory.Build.props
```diff
@@ -1,4 +1,7 @@
 ﻿<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <IsClientLibrary>true</IsClientLibrary>
```
Yao725, line 3: I don't think we need this, actually we use `IsMgmtLibrary` for all track 2 mgmt. plane sdk.

sdk/iotcentral/Azure.ResourceManager.IotCentral/tests/Azure.ResourceManager.IotCentral.Tests.csproj
```diff
@@ -1,5 +1,9 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <ItemGroup>
     <ProjectReference Include="..\src\Azure.ResourceManager.IotCentral.csproj" />
+    <ProjectReference Include="$(AzureCoreTestFramework)" />
```
Yao725, line 4: You don't need to include test framework here, it is already be done in [Directory.Build.Common.targets](https://github.com/Yao725/azure-sdk-for-net/blob/main/eng/Directory.Build.Common.targets#L226)

sdk/iotcentral/Azure.ResourceManager.IotCentral/tests/IotCentralLifeCycleTests.cs
```diff
@@ -0,0 +1,200 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
Yao725, line 1: Could you please follow our track 2 sdk test pattern? 

1. First, we will define a [base test class](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/cdn/Azure.ResourceManager.Cdn/tests/CdnManagementTestBase.cs) which inherits from `ManagementRecordedTestBase` and its name will be in the form `{RP}ManagementTestBase`. And in this base class we defined the `Client` field and used the attribute `SetUp` to construct it. 
2. Then we will have two cs files to test different methods of the one resource separately. In other words, the file named `{Resource}CollectionTests` test the methods in `{Resource}Collection` and `{Resource}OperationsTests` test the methods in `{Resource}Resource`. Both of them are the derived classed of the base test class. Here is the [example](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/cdn/Azure.ResourceManager.Cdn/tests/Scenario/AfdCustomDomainCollectionTests.cs) of collection test in Cdn.
3. One more thing to say is, we use these [two attributes](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/cdn/Azure.ResourceManager.Cdn/tests/Scenario/AfdCustomDomainCollectionTests.cs#L37-L38) to mark test instead of `AsyncOnly`. You can just test the asynchronous methods and our test framework will automatically help you to test the corresponding synchronous methods.

Yao725 (changes requested):
As for the error in CI, we run the tests in pipeline with playback mode, which means when you finish writing the tests, you must set up these [environment variables](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/resourcemanager/Azure.ResourceManager/docs/AuthUsingEnvironmentVariables.md) in your local machine and run the test in record mode. This will record the tests and generate corresponding [test record](https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/cdn/Azure.ResourceManager.Cdn/tests/SessionRecords) in the `SessionRecords` folder, so we can use them to playback.

sdk/iotcentral/Azure.ResourceManager.IotCentral/tests/Tests/IotCentralExtensionsTests.cs
```diff
@@ -0,0 +1,152 @@
...
+                var iotCentralAppData = new IotCentralAppData(AzureLocation.WestUS, new AppSkuInfo(AppSku.ST0))
+                {
+                    DisplayName = appName,
+                    Subdomain = appName,
+                };
+                await appsCollection.CreateOrUpdateAsync(WaitUntil.Completed, appName, iotCentralAppData, CancellationToken.None);
+            }
+
+            // Get and delete IoT Central apps.
+            await foreach (var rgApp in rg.GetIotCentralApps())
```
Yao725, line 107: Nit: can we directly call the `GetAll` method?
```suggestion
            await foreach (var rgApp in appsCollection.GetAllAsync())
```

Yao725 (approved):
Thanks @chgennar for the contribution, I left a minor comment and the others LGTM. That's all the tests, right? If so, I think we can merge it.

sdk/iotcentral/Azure.ResourceManager.IotCentral/tests/Tests/IotCentralExtensionsTests.cs
chgennar: this test is meant to check whether that rg.GetIotCentralApps() calls returns all the created applications.

chgennar:
> Thanks @chgennar for the contribution, I left a minor comment and the others LGTM. That's all the tests, right? If so, I think we can merge it.

Thanks for the review! Yes, that is all the tests.

sdk/iotcentral/Azure.ResourceManager.IotCentral/tests/Tests/IotCentralExtensionsTests.cs
Yao725: Yeah, I understand your intention and I'm ok with that. I have this suggestion because the `IotCentralAppCollection` class is derived from [IAsyncEnumerable](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/iotcentral/Azure.ResourceManager.IotCentral/src/Generated/IotCentralAppCollection.cs#L28), so this call will implicitly call the [GetAsyncEnumerator](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/iotcentral/Azure.ResourceManager.IotCentral/src/Generated/IotCentralAppCollection.cs#L329) method, in which the `GetAllAsync()` method is called to return all the created apps in this resource group. So I'd like to directly use the `GetAllAsync()` method to make the code more intuitive.

chgennar: i see, updated!

PR #28449, opened 2022-04-28
[TA] Test enabled for Detect Language

maririos:
- Enabled for Detect Language:
  - Live tests
  - Mock tests
  - Samples
  - Sample snippets
- Recordings added
- Disable the rest of the tests for easier viewing
- Commented mock tests for Internal types


sdk/textanalytics/Azure.AI.TextAnalytics/tests/MockTests/TextAnalyticsClientMockTests.cs
```diff
@@ -0,0 +1,679 @@
...
+                Transport = transport
+            };
+
+            var client = InstrumentClient(new TextAnalyticsClient(new Uri(s_endpoint), new AzureKeyCredential(s_apiKey), options));
+
+            return client;
+        }
+
+        [Test]
+        [Ignore ("Not implemented yet")]
```
maririos, line 38: I could also delete all of the not implemented tests and add them as things start to work... but I prefer this way so there is less chance we forget to do them

sdk/textanalytics/Azure.AI.TextAnalytics/tests/MockTests/TextAnalyticsClientMockTests.cs
```diff
@@ -0,0 +1,679 @@
...
+            var resultCollection = response.Value;
+
+            Assert.AreEqual("4", resultCollection[0].Id);
+            Assert.AreEqual("5", resultCollection[1].Id);
+            Assert.AreEqual("2", resultCollection[2].Id);
+            Assert.AreEqual("3", resultCollection[3].Id);
+        }
+
+        [Test]
+        public async Task DetectedLanguageNullName()
```
maririos, line 211: we do have 2 DT tests here that work :)

sdk/textanalytics/Azure.AI.TextAnalytics/tests/MockTests/Legacy/TextAnalyticsClientMockTests.cs
```diff
@@ -11,8 +11,9 @@
 using System.Text;
 using System.Threading.Tasks;
 
-namespace Azure.AI.TextAnalytics.Tests
+namespace Azure.AI.TextAnalytics.Legacy.Tests
```
jsquire, line 14: Interesting that we've got tests dedicated to legacy only.   I'd like to better understand the strategy there.   In theory, we should be able to test all service versions with the same test, no?

jsquire (approved):
Approving to keep things unblocked.   Nothing crazy jumped out at me in a quick and shallow pass.   I'll give this another, deeper look, tomorrow morning.

sdk/textanalytics/Azure.AI.TextAnalytics/tests/MockTests/Legacy/TextAnalyticsClientMockTests.cs
maririos: This happens only for the Mock tests. The reason why I wanted to separate these tests was that the input and output from the service are slightly different so the "output mock" is affected.

jsquire (approved):

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClientDiagnostics.cs
```diff
@@ -33,29 +33,28 @@ public TextAnalyticsClientDiagnostics(ClientOptions options) : base(options)
...
+            if (!string.IsNullOrEmpty(content))
+            {
+                try
+                {
+                    // Try to parse the failure content and use that as the
+                    // default value for the message, error code, etc.
+                    using JsonDocument doc = JsonDocument.Parse(content);
+                    if (doc.RootElement.TryGetProperty("error", out JsonElement errorElement))
+                    {
+                        TextAnalyticsError error = Transforms.ConvertToError(Error.DeserializeError(errorElement));
```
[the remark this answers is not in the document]
maririos: Good point. We could always overwrite the `DeserializeError` method. I avoided that in the past in case the Service kept modifying the structure of it. But this is not going to change too much, especially now that `Error` is a common type across services and follows the REST guidelines.

Issue https://github.com/Azure/azure-sdk-for-net/issues/28461

sdk/textanalytics/Azure.AI.TextAnalytics/tests/ExtractSummaryTests.cs
```diff
@@ -10,6 +10,7 @@
 namespace Azure.AI.TextAnalytics.Tests
 {
     [ServiceVersion(Min = TextAnalyticsClientOptions.ServiceVersion.V3_2_Preview_2)]
```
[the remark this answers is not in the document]
maririos: My understanding is it using the enum value. In this case, it is saying run this set of tests for whatever version that goes after and including `v3_2_preview_2`.

See https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core.TestFramework/src/ClientTestFixtureAttribute.cs#L61

PR #28452, opened 2022-04-28
Update codespaces config to use default universal image

hallipr:
Codespaces are currently broken due to disk space and docker init issues. The latest universal image already contains docker. This change removes our custom docker work and uses the default universal image.


chidozieononiwu (approved):
This definitely works for me.

PR #28454, opened 2022-04-28
Move receive actions onto separate type

JoshLove-msft:


jsquire (approved):
Out of curiosity, what was the motivation to split?

JoshLove-msft:
> Out of curiosity, what was the motivation to split?

This was based on a suggestion from @tg-msft

PR #28463, opened 2022-04-29
[Text Analytics] Delegation to Serivce Clients

jsquire:
# Summary

The focus of these changes is to shift functionality for the Language REST API integration into the `LanguageServiceClient` and update the `TextAnalyticsClient` to delegate all operations to the service client.

## Next steps (BEFORE MERGE)

- Integrate the recent work for the language service integration via rebase.
- Coordinate with language service integration efforts to move work into the LanguageServiceClient

### DIFF
--- a/sdk/textanalytics/Azure.AI.TextAnalytics/src/AnalyzeActionsOperation.cs
+++ b/sdk/textanalytics/Azure.AI.TextAnalytics/src/AnalyzeActionsOperation.cs
@@ -152,7 +152,7 @@ namespace Azure.AI.TextAnalytics
 
             Id = operationId;
             //_serviceClient = client._serviceRestClient;
-            _diagnostics = client._clientDiagnostics;
+            _diagnostics = client._serviceClient.Diagnostics;
             _operationInternal = new OperationInternal<AsyncPageable<AnalyzeActionsResult>>(_diagnostics, this, rawResponse: null);
         }
 
--- a/sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/LanguageServiceClient.cs
+++ b/sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/LanguageServiceClient.cs
@@ -54,10 +54,8 @@ namespace Azure.AI.TextAnalytics.ServiceClients
             Argument.AssertNotNull(credential, nameof(credential));
             Argument.AssertNotNull(options, nameof(options));
 
-            Argument.AssertNotNullOrEmpty(defaultLanguage, nameof(defaultLanguage));
-            Argument.AssertNotNullOrEmpty(defaultCountryHint, nameof(defaultCountryHint));
-            Argument.AssertNotNull(authorizationScope, nameof(authorizationScope));
-            Argument.AssertNotNull(serviceVersion, nameof(serviceVersion));
+            Argument.AssertNotNullOrEmpty(authorizationScope, nameof(authorizationScope));
+            Argument.AssertNotNullOrEmpty(serviceVersion, nameof(serviceVersion));
 
             _baseUri = endpoint;
             _clientDiagnostics = new TextAnalyticsClientDiagnostics(options);
--- a/sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/LanguageServiceClient.cs
+++ b/sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/LanguageServiceClient.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
@@ -64,28 +65,6 @@ namespace Azure.AI.TextAnalytics.ServiceClients
 
         #region Detect Language
 
-        /// <summary>
-        /// Runs a predictive model to determine the language the passed-in
-        /// document is written in, and returns the detected language as well
-        /// as a score indicating the model's confidence that the inferred
-        /// language is correct. Scores close to 1 indicate high certainty in
-        /// the result.
-        /// <para>For document length limits, maximum batch size, and supported text encoding, see
-        /// <see href="https://docs.microsoft.com/azure/cognitive-services/text-analytics/concepts/data-limits?tabs=version-3"/>.</para>
-        /// </summary>
-        /// <param name="document">The document to analyze.</param>
-        /// <param name="countryHint">Indicates the country of origin of the
-        /// document to assist the Text Analytics model in predicting the language
-        /// it is written in. If unspecified, this value will be set to the
-        /// default country hint in <see cref="TextAnalyticsClientOptions.DefaultCountryHint"/>
-        /// in the request sent to the service.
-        /// To remove this behavior, set to <see cref="DetectLanguageInput.None"/>.</param>
-        /// <param name="cancellationToken">A <see cref="CancellationToken"/>
-        /// controlling the request lifetime.</param>
-        /// <returns>A result containing the detected language or an error if
-        /// the model could not analyze the document.</returns>
-        /// <exception cref="RequestFailedException">Service returned a non-success
-        /// status code.</exception>
         public override async Task<Response<DetectedLanguage>> DetectLanguageAsync(string document, string countryHint = default, CancellationToken cancellationToken = default)
         {
             Argument.AssertNotNullOrEmpty(document, nameof(document));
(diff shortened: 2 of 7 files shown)


sdk/textanalytics/Azure.AI.TextAnalytics/src/AnalyzeActionsOperation.cs
```diff
@@ -152,7 +152,7 @@ public AnalyzeActionsOperation(string operationId, TextAnalyticsClient client)
 
             Id = operationId;
             //_serviceClient = client._serviceRestClient;
-            _diagnostics = client._clientDiagnostics;
+            _diagnostics = client._serviceClient.Diagnostics;
```
jsquire, line 155: When we look at the LROs, I'd like to look deeper here and see if we can't remove external access to underscore members in favor of an abstraction.

sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/LanguageServiceClient.cs
```diff
@@ -16,87 +20,2193 @@ namespace Azure.AI.TextAnalytics.ServiceClients
     ///
     internal class LanguageServiceClient : ServiceClient
```
jsquire, line 21: This is where we'll need to shift effort for integrating the Language REST API.

//cc: @nisha-bhatia

sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/LanguageServiceClient.cs
```diff
@@ -3,8 +3,12 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
```
jsquire, line 6: This includes all work that was merged into the feature branch at the time the PR was opened.   I'll do another rebase before we take next steps to pull in any changes after that point.

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
```diff
@@ -19,12 +20,14 @@ namespace Azure.AI.TextAnalytics
     /// </summary>
     public class TextAnalyticsClient
```
jsquire, line 21: Currently, this class is a pure pass-through; it has responsibility for creating the service client that corresponds to the requested version, and nothing else.   This was chosen primarily because it was the most expedient path forward.   In the future, we could consider pulling some common functionality here rather than delegating to the service clients and duplicating code.  For example:

- Parameter validation
- Doing the transforms/packaging that used the private helpers previously in the `Common` section

The trade-off for centralizing that logic would be that we lose the freedom to have logic specific to the service versions - which may or may not be desirable.   In the longer term, since the Legacy API is frozen, it may be a better pattern to localize logic being changed to the Language Service client.   Something to consider.

maririos: I have to say, I love how this class looks :)

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
```diff
@@ -62,13 +65,29 @@ public TextAnalyticsClient(Uri endpoint, TokenCredential credential, TextAnalyti
...
-            var pipeline = HttpPipelineBuilder.Build(options, new BearerTokenAuthenticationPolicy(credential, defaultScope));
-            _languageRestClient = new MicrosoftCognitiveLanguageServiceRestClient(_clientDiagnostics, pipeline, endpoint.AbsoluteUri, TextAnalyticsClientOptions.GetVersionString(options.Version));
+            _serviceClient = options.Version switch
+            {
+                var version when s_legacyServiceVersions.Contains(version) =>
+                    new LegacyServiceClient(
+                        endpoint,
+                        credential,
+                        options.DefaultLanguage,
+                        options.DefaultCountryHint,
```
maririos, line 77: curious on why the need to pass these 2 properties directly, instead of getting their values from `options` in the respective clients.
My concern is if in the future we keep adding defaults to the library, we are going to have to add more parameters to the method

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
```diff
@@ -102,12 +121,29 @@ public TextAnalyticsClient(Uri endpoint, AzureKeyCredential credential, TextAnal
             Argument.AssertNotNull(credential, nameof(credential));
             Argument.AssertNotNull(options, nameof(options));
 
-            _baseUri = endpoint;
-            _clientDiagnostics = new TextAnalyticsClientDiagnostics(options);
-            _options = options;
+            var authorizationScope = $"{(string.IsNullOrEmpty(options.Audience?.ToString()) ? TextAnalyticsAudience.AzurePublicCloud : options.Audience)}/.default";
```
maririos, line 124: for API Key, we don't need the `authorizationScope` so you could remove this variable and not pass it in the client ctrs

sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/LanguageServiceClient.cs
```diff
@@ -16,87 +20,2189 @@ namespace Azure.AI.TextAnalytics.ServiceClients
...
-        public override Task<Response<DetectLanguageResultCollection>> DetectLanguageBatchAsync(IEnumerable<string> documents, string countryHint = default, TextAnalyticsRequestOptions options = default, CancellationToken cancellationToken = default) => throw new NotImplementedException();
-        public override Response<DetectLanguageResultCollection> DetectLanguageBatch(IEnumerable<string> documents, string countryHint = default, TextAnalyticsRequestOptions options = default, CancellationToken cancellationToken = default) => throw new NotImplementedException();
-        public override Task<Response<DetectLanguageResultCollection>> DetectLanguageBatchAsync(IEnumerable<DetectLanguageInput> documents, TextAnalyticsRequestOptions options = default, CancellationToken cancellationToken = default) => throw new NotImplementedException();
-        public override Response<DetectLanguageResultCollection> DetectLanguageBatch(IEnumerable<DetectLanguageInput> documents, TextAnalyticsRequestOptions options = default, CancellationToken cancellationToken = default) => throw new NotImplementedException();
+        /// <summary>
```
maririos, line 71: Currently `TextAnalyticsClient` and `LanguageServiceClient` have them. I think Intellisense will use the override, so will it be better to have the docstrings moved from `TextAnalyticsClient` to the `LegacyServiceClient`?

sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/ServiceClient.cs
```diff
@@ -16,6 +17,8 @@ namespace Azure.AI.TextAnalytics.ServiceClients
     ///
     internal abstract class ServiceClient
     {
+        public abstract ClientDiagnostics Diagnostics { get; }
```
maririos, line 20: Could you explain to me why the need for this?

jsquire: Because the LROs are using it, currently.   Generally, we don't want to allow access to underscore members; those are intended to be private to the class.   Once we need access outside, it's better to have a property or a field named like a property to send a clear signal that it's safe to use this thing.

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
jsquire: There's context in the previous PR, but the short version is that I was on the fence about how proper it was to expose the `TextAnalyticsClientOptions` to a class that wasn't `TextAnalyticsClient` - it just feels like a leaky abstraction.    

I'm open to changing it, it's not something that I felt strongly about.

sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/ServiceClient.cs
maririos: :| ugh right

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
maririos: Apologies I missed that in the last review.
I don't feel strongly about it either. We can leave it as is and if it ever is a problem, then change it

maririos (approved):

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
jsquire: It was a good call-out; I ended up needing `ClientOptions` anyway, which I wasn't accounting for initially.   I updated to just pass the options.

PR #28466, opened 2022-04-29
Extract custom error creation from ClientDiagnostics (shared code)

christothes:
## Moving custom error creation out of `ClientDiagnostics` and into `RequestFailedException` along with `Response`.

related to #27490 #24031

### Changes required for the generator to utilize this new approach:

- Construction of `RequestFailedException` would change from 

```C#
RequestFailedException rfex = await ClientDiagnostics.CreateRequestFailedExceptionAsync(failedSubResponse).ConfigureAwait(false);
```
to 

```C#
RequestFailedException rfex = new RequestFailedException(failedSubResponse);
```

### Changes required for clients that want to customize their Error content changes

Previously, clients created a sub-class of `ClientDiagnostics` and created an override for `ExtractFailureContent`

Now clients would use a new overload to `HttpPipelineBuilder.Build` which accepts a new options type called `HttpPipelineBuildOptions`. Clients constructing the pipline would use the new overload as follows:

```C#
var pipelineOptions = new HttpPipelineBuildOptions(
	options,
	perCallPolicies,
	new[] { new TableBearerTokenChallengeAuthorizationPolicy(tokenCredential, TableConstants.StorageScope, options.EnableTenantDiscovery) },
	new ResponseClassifier(),
	new TablesRequestFailedDetailsParser());

_pipeline = HttpPipelineBuilder.Build(pipelineOptions);
```

```c#
internal class TablesRequestFailedDetailsParser : RequestFailedDetailsParser
{
    public override RequestFailedDetails GetRequestFailedDetails(Response response)
    { ... }
}
```


annelo-msft:
Would it make sense to link this to the issue it's addressing?

annelo-msft:
> FooRestClient would take DiagnosticsOptions as an additional parameter.

Does this only apply to HLCs?  I expect we'll want to do it in DPG clients and probably Mgmt plane libraries as well.

christothes:
> > FooRestClient would take DiagnosticsOptions as an additional parameter.
> 
> Does this only apply to HLCs? I expect we'll want to do it in DPG clients and probably Mgmt plane libraries as well.

Do DPGs not utilize `DiagnosticsOptions` ?

sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -200,6 +200,20 @@ public partial class RequestContext
         public void AddPolicy(Azure.Core.Pipeline.HttpPipelinePolicy policy, Azure.Core.HttpPipelinePosition position) { }
         public static implicit operator Azure.RequestContext (Azure.ErrorOptions options) { throw null; }
     }
+    [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
+    public partial struct RequestFailedDetails
+    {
+        private object _dummy;
+        private int _dummyPrimitive;
+        public RequestFailedDetails(Azure.ResponseError? error, System.Collections.Generic.IDictionary<string, string>? additionalInfo) { throw null; }
```
annelo-msft, line 208: Curious: why are both inputs nullable? Is there a case where we'd have error details without an error?

sdk/core/Azure.Core/src/Pipeline/RequestFailedDetailsParser.cs
```diff
@@ -0,0 +1,18 @@
...
+    ///
+    /// </summary>
+    public abstract class RequestFailedDetailsParser
+    {
+        /// <summary>
+        ///
+        /// </summary>
+        /// <param name="response"></param>
+        /// <returns></returns>
+        public abstract RequestFailedDetails GetRequestFailedDetails(Response response);
```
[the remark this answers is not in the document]
christothes: There is no reason to implement this class if you don't need to override this method. Is there a use case you had in mind where we'd want to call the base implementation?

sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -211,6 +225,8 @@ public partial class RequestFailedException : System.Exception, System.Runtime.S
         public RequestFailedException(string message, System.Exception? innerException) { }
         public string? ErrorCode { get { throw null; } }
         public int Status { get { throw null; } }
+        public static Azure.RequestFailedException Create(Azure.Response response, Azure.ResponseError? error = null, System.Exception? innerException = null) { throw null; }
+        public static System.Threading.Tasks.ValueTask<Azure.RequestFailedException> CreateAsync(Azure.Response response, Azure.ResponseError? error = null, System.Exception? innerException = null, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
```
annelo-msft, line 229: Curious: Did we add this factory method to handle the case where the response is not buffered, because there is no way to add an async constructor?

For the sync one, is that to be symmetric with the async factory method?  Because otherwise it might make sense to add a third constructor that takes Azure.ResponseError? to maintain the constructor pattern we've already established.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -932,6 +953,17 @@ public static partial class HttpPipelineBuilder
...
+        public static Azure.Core.Pipeline.HttpPipeline Build(Azure.Core.Pipeline.HttpPipelineBuildOptions options) { throw null; }
+        public static Azure.Core.Pipeline.DisposableHttpPipeline Build(Azure.Core.Pipeline.HttpPipelineBuildOptions options, Azure.Core.Pipeline.HttpPipelineTransportOptions transportOptions) { throw null; }
+    }
+    public partial class HttpPipelineBuildOptions
+    {
+        public HttpPipelineBuildOptions(Azure.Core.ClientOptions options, Azure.Core.Pipeline.HttpPipelinePolicy[] perCallPolicies, Azure.Core.Pipeline.HttpPipelinePolicy[] perRetryPolicies, Azure.Core.ResponseClassifier? responseClassifier, Azure.RequestFailedDetailsParser? requestFailedDetailsParser = null) { }
+        public Azure.Core.ClientOptions ClientOptions { get { throw null; } set { } }
+        public Azure.Core.Pipeline.HttpPipelinePolicy[] PerCallPolicies { get { throw null; } set { } }
+        public Azure.Core.Pipeline.HttpPipelinePolicy[] PerRetryPolicies { get { throw null; } set { } }
+        public virtual Azure.RequestFailedDetailsParser RequestFailedDetailsParser { get { throw null; } set { } }
```
annelo-msft, line 965: Curious: why is this property `virtual` when others aren't?

sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -200,6 +200,20 @@ public partial class RequestContext
         public void AddPolicy(Azure.Core.Pipeline.HttpPipelinePolicy policy, Azure.Core.HttpPipelinePosition position) { }
         public static implicit operator Azure.RequestContext (Azure.ErrorOptions options) { throw null; }
     }
+    [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
+    public partial struct RequestFailedDetails
```
annelo-msft, line 204: I'm wondering if we need this as a public type.  It seems like RequestFailedDetailsParser could have two methods: GetResponseError(Response response) and GetAdditionalInfo(), and we could reduce the number of types we're adding to the API.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -200,6 +200,20 @@ public partial class RequestContext
...
+    {
+        private object _dummy;
+        private int _dummyPrimitive;
+        public RequestFailedDetails(Azure.ResponseError? error, System.Collections.Generic.IDictionary<string, string>? additionalInfo) { throw null; }
+        public System.Collections.Generic.IDictionary<string, string>? AdditionalInfo { get { throw null; } }
+        public Azure.ResponseError? Error { get { throw null; } set { } }
+    }
+    public partial class RequestFailedDetailsParser
```
annelo-msft, line 212: Would `ErrorResponseParser` be a better name for this?  It would match the `Response.IsError` terminology elsewhere in the API.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -464,6 +480,11 @@ public enum HttpPipelinePosition
         PerRetry = 1,
         BeforeTransport = 2,
     }
+    public static partial class MultipartResponse
```
annelo-msft, line 483: Curious: why add this type to the public API as part of this change?

sdk/core/Azure.Core/src/Pipeline/Internal/Multipart/BufferedReadStream.cs
```diff
@@ -243,7 +244,9 @@ public async override Task<int> ReadAsync(byte[] buffer, int offset, int count,
                 return toCopy;
             }
 
+#pragma warning disable CA1835 // WriteAsync(Memory<>) overload is not available in all targets
```
annelo-msft, line 247: Curious: why suppress this warning?

sdk/core/Azure.Core/src/Pipeline/Internal/Multipart/BufferedReadStream.cs
```diff
@@ -288,7 +293,7 @@ public bool EnsureBuffered(int minCount)
         {
             if (minCount > _buffer.Length)
             {
-                throw new ArgumentOutOfRangeException(nameof(minCount), minCount, "The value must be smaller than the buffer size: " + _buffer.Length.ToString());
+                throw new ArgumentOutOfRangeException(nameof(minCount), minCount, "The value must be smaller than the buffer size: " + _buffer.Length.ToString(CultureInfo.InvariantCulture));
```
annelo-msft, line 296: Curious: why add this?  I thought we didn't like to use InvariantCulture, but I don't remember why I think that.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
annelo-msft, line 229: Curious: did we consider an approach where we take the custom parser directly into the constructor/factory method instead of adding it to pipeline build options?

christothes: No - it's because the exception message is readonly, so we need to process the response before constructing the exception.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
christothes: Good catch - this was not intentional.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
christothes: Unless I am misunderstanding the suggestion, this would mean that we'd have to process the response content twice - once to extract the error details and once to extract the additionalInfo collection. I not sure it's worth doing that just to avoid the type.

sdk/core/Azure.Core/src/Pipeline/Internal/Multipart/BufferedReadStream.cs
christothes: It's done in a [few other places](https://grep.app/search?q=%23pragma%20warning%20disable%20CA1835&regexp=true&filter[repo][0]=Azure/azure-sdk-for-net&filter[path][0]=sdk/) - this is because the analyzer complains about not using the correct overload for the net6.0 target, but it's not available in the others. The analyzer didn't complain when it was in shared source.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
christothes: > What is the advantage of having it baked into the pipeline?

So we can get the parser set on the Response (currently it's done by the `HttpPipelineTransportPolicy` just like the Sanitizer)

If we don't set it on the `Response` we have to solve the problem of how to get the parser to the factory method.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
christothes: I don't feel strongly, but one reason I like the current name is that everything centers around `RequestFailedException`. Are there scenarios you are aware of where these concepts stand alone?

That said, if others like these names better, I'm good with it.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
annelo-msft: Why not process the response in the constructor, like we did for the other constructors?

annelo-msft: > If we don't set it on the Response we have to solve the problem of how to get the parser to the factory method.

Would passing the parser as a parameter into the factory method not work to get the parser to the factory method?

sdk/core/Azure.Core/api/Azure.Core.net461.cs
annelo-msft: I would hesitate to put this type in the public API unless there is a clear use case for it from a customer perspective -- either a library author or an end user customer (typically the former for shared source).  There is usually a way to work around the Core/internal divide in other ways.

If we do decide it makes sense to put this in the public API, I think we should take a step back and get clear on what the requirements/needs for this type are and make sure we design it appropriately.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
christothes: One of the reasons we stashed it on the `Response` in the first place was to avoid having to pass the parser around. 

Could you say more about the concern with having it be part of the pipeline build?

sdk/core/Azure.Core/src/Pipeline/Internal/Multipart/BufferedReadStream.cs
annelo-msft: Is the analyzer telling us something important that we should pay attention to?  Why didn't it complain when it was in shared source?

In general, I think it's preferred for us to be paying down technical debt rather than increasing it as we move forward.

sdk/core/Azure.Core/src/Pipeline/Internal/Multipart/BufferedReadStream.cs
annelo-msft: Ok, cool, I don't remember why I think that, so I could be completely wrong.

Do we know why the analyzer complains if we don't do it?  It would be good to understand why we're making this change.

sdk/core/Azure.Core/src/Pipeline/RequestFailedDetailsParser.cs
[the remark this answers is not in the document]
annelo-msft: I think it's worth considering @JoshLove-msft's suggestion.  What would it look like if instead of checking the parser for null, we just always called the parser, and if no one had subclassed it, we got the default implementation?  ResponseClassifier uses this design, so there's a precedent for it in the API.  Does it make things simpler, or more complex?

sdk/core/Azure.Core/src/Pipeline/Internal/Multipart/BufferedReadStream.cs
christothes: This is net zero for current debt (as the code was simply relocated). I'd rather file an issue for the broader concern to investigate if this is worth changing everywhere it exists, rather than just here as part of this PR.

sdk/core/Azure.Core/src/Pipeline/Internal/Multipart/BufferedReadStream.cs
christothes: https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1305

sdk/core/Azure.Core/api/Azure.Core.net461.cs
christothes: The constructor needs to ultimately call the base constructor on `Exception`, and we need to manipulate all the inputs based on the response. We could make it work, but I believe it would involve duplicating work parsing the response. Secondly, to your original question, we'd need to do async work as well.

sdk/core/Azure.Core/src/Pipeline/RequestFailedDetailsParser.cs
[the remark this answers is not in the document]
christothes: Ideally, I agree with the idea that we should always call the parser rather than checking for null. My main concern with doing that is it relies on the `HttpPipelineTransportPolicy` which will almost always be there but is not guaranteed to be there. My thought was that we should be erring on the side of avoiding a NRE or failing to call even the default parser. 

But the main concern is just that making it virtual doesn't force sub-classes to provide an implementation at all, which is really the only reason one would sub-class the type.

sdk/core/Azure.Core/src/Pipeline/Internal/Multipart/BufferedReadStream.cs
christothes: created #29197

sdk/core/Azure.Core/api/Azure.Core.net461.cs
annelo-msft: Does it make sense to make an allocation for a type that contains only null properties?  Could you return null instead?

sdk/core/Azure.Core/src/Pipeline/RequestFailedDetailsParser.cs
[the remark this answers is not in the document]
annelo-msft: > My concern with doing that is it relies on the HttpPipelineTransportPolicy

Can you help me understand this depencency?

> But the main concern is just that making it virtual doesn't force sub-classes to provide an implementation at all, which is really the only reason one would sub-class the type.

Do we need to design for the case where someone wants to subclass a type with one virtual method and not override the virtual method?  I don't understand why someone would do that.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
annelo-msft: What I don't understand is that we're already parsing the response in two constructors on this type.  Why would we add a new creation pattern instead of following the existing pattern?

sdk/core/Azure.Core/api/Azure.Core.net461.cs
christothes: It's in the exception path, so I don't think that matters all too much. I tried a different shape in the latest iterations.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
JoshLove-msft: I also like the idea of passing this into the factory method. This is where the parser will be used so it seems reasonable to pass it if you want to do something custom.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
annelo-msft: I think it'll be good to decide in context once the rest of the API additions are closer to completion.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
annelo-msft: I was just wondering if we'd considered that approach.  We're adding a lot of types and constructors to support putting it on Response (which we'll probably need in the future), and if the RequestFailedException constructor is always called from a client method implementation, it might reduce complexity to just pass it in at the call site when the exception is created.  If there are other places we create exceptions, then I agree this may not be ideal.

JoshLove-msft: It looks like in the [existing code](https://github.com/Azure/azure-sdk-for-net/pull/28466/files#diff-d201011507f12f72bd9d970a544a1f9f1395e535dd55549a6419cb46ecf33b7bR67), we are using the sync path for reading the response content. I suppose it might be worth making this tradeoff if it keep the API cleaner.

christothes: > I was just wondering if we'd considered that approach. We're adding a lot of types and constructors to support putting it on Response (which we'll probably need in the future), and if the RequestFailedException constructor is always called from a client method implementation, it might reduce complexity to just pass it in at the call site when the exception is created. If there are other places we create exceptions, then I agree this may not be ideal.

Yes, in fact that was the original design. But in the arch board discussion on this topic, the idea was raised to stash it on the `Response` instead. After implementing it this way, I like the idea even more. It avoids the caller from having to make decisions when raising an exception, just once when configuring the pipeline. In addition, to your point, the context for how to raise an exception is always with `Response`. The advantage there is this avoids ever having to refactor just to make the error customization available in additional code paths. 

That said, I'm happy to revisit the decision if we agree that the advantages are more compelling.

sdk/core/Azure.Core/src/Pipeline/RequestFailedDetailsParser.cs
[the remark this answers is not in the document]
christothes: discussed offline - some takeaways:
- There is no reason to implement a sub-class of this type unless one intends to do custom error parsing
- There are no current examples of custom `ClientDiagnostics.ExtractFailureContent` implementations that call the default implementation as part of their custom implementation
- Regardless of whether this is purely abstract or virtual, the client author code that is written is identical. 
- question for the architects: Is the existence of a pure abstract type for client authors to implement inconsistent with our current APIs? If so, how much should this inconsistency be weighed against the use case?

sdk/core/Azure.Core/api/Azure.Core.net461.cs
christothes: I wrote up some pros/cons of the various approaches we have discussed [here](https://gist.github.com/christothes/35f276c3b316cb2d4c7b53ee8da85425)

sdk/core/Azure.Core/api/Azure.Core.net461.cs
christothes: In the latest iteration, I eliminated this type.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
JoshLove-msft: We could ensure that error responses are always buffered by adding this into RetryPolicy right before throwing.

sdk/core/Azure.Core/src/Azure.Core.csproj
```diff
@@ -7,7 +7,7 @@
     <ApiCompatVersion>1.24.0</ApiCompatVersion>
     <PackageTags>Microsoft Azure Client Pipeline</PackageTags>
     <Nullable>enable</Nullable>
-    <DefineConstants>$(DefineConstants);AZURE_NULLABLE</DefineConstants>
+    <DefineConstants>$(DefineConstants);AZURE_NULLABLE;HAS_INTERNALS_VISIBLE_CORE</DefineConstants>
```
JoshLove-msft, line 10: Should this be here?

sdk/core/Azure.Core/src/Pipeline/RequestFailedDetailsParser.cs
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Collections.Generic;
+
+namespace Azure.Core
+{
+    /// <summary>
+    ///
```
JoshLove-msft, line 9: nit: docs

sdk/core/Azure.Core/src/RequestFailedException.cs
```diff
@@ -112,5 +125,134 @@ public override void GetObjectData(SerializationInfo info, StreamingContext cont
...
+                string? content = null;
+                if (response.ContentStream != null && response.ContentStream.CanSeek)
+                {
+                    content = response.Content.ToString();
+                }
+                else
+                {
+                    // this path should only happen in exceptional cases such as when
+                    // the RFE ctor was called directly by client or customer code with an un-buffered response.
+                    // Generated code would never do this.
```
JoshLove-msft, line 233: What about storage generated code?

christothes: Once the full set of generator changes come in, this will be true for all error paths

JoshLove-msft: Is there an issue filed for updating ResponseBodyPolicy to use IsErrorResponse, and updating generater to use this as well?

sdk/core/Azure.Core/src/RequestFailedException.cs
```diff
@@ -112,5 +125,134 @@ public override void GetObjectData(SerializationInfo info, StreamingContext cont
...
+                    content = ReadContent(response);
+                }
+                // Optimistic check for JSON object we expect
+                if (content == null || !content.StartsWith("{", StringComparison.OrdinalIgnoreCase))
+                {
+                    return false;
+                }
+                error = System.Text.Json.JsonSerializer.Deserialize<ErrorResponse>(content)?.Error;
```
JoshLove-msft, line 241: nit: avoid fully qualifying

sdk/core/Azure.Core/src/RequestFailedException.cs
```diff
@@ -112,5 +125,134 @@ public override void GetObjectData(SerializationInfo info, StreamingContext cont
...
+                // Ignore any failures - unexpected content will be
+                // included verbatim in the detailed error message
+            }
+
+            return error != null;
+        }
+
+        private class ErrorResponse
+        {
+            [System.Text.Json.Serialization.JsonPropertyName("error")]
```
JoshLove-msft, line 254: nit: fully qualified

sdk/search/Microsoft.Azure.Search/tests/Tests/CustomAnalyzerTests.cs
```diff
@@ -540,7 +540,7 @@ public void CanUseAllRegexFlags()
             });
         }
 
-        [Fact(Skip = "https://github.com/Azure/azure-sdk-for-net/issues/29190")]
+        [Fact(Skip="https://github.com/Azure/azure-sdk-for-net/issues/29190")]
```
JoshLove-msft, line 543: Why?

sdk/core/Azure.Core/src/RequestFailedException.cs
christothes: This was to avoid conflicts with the `System.Runtime.Serialization`

sdk/core/Azure.Core/src/RequestFailedException.cs
christothes: [yes - Al](https://github.com/Azure/autorest.csharp/issues/2385)

sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -932,6 +943,17 @@ public static partial class HttpPipelineBuilder
...
         public static Azure.Core.Pipeline.HttpPipeline Build(Azure.Core.ClientOptions options, Azure.Core.Pipeline.HttpPipelinePolicy[] perCallPolicies, Azure.Core.Pipeline.HttpPipelinePolicy[] perRetryPolicies, Azure.Core.ResponseClassifier? responseClassifier) { throw null; }
+        public static Azure.Core.Pipeline.HttpPipeline Build(Azure.Core.Pipeline.HttpPipelineOptions options) { throw null; }
+        public static Azure.Core.Pipeline.DisposableHttpPipeline Build(Azure.Core.Pipeline.HttpPipelineOptions options, Azure.Core.Pipeline.HttpPipelineTransportOptions transportOptions) { throw null; }
+    }
+    public partial class HttpPipelineOptions
+    {
+        public HttpPipelineOptions(Azure.Core.ClientOptions options) { }
+        public Azure.Core.ClientOptions ClientOptions { get { throw null; } }
+        public Azure.Core.Pipeline.HttpPipelinePolicy[] PerCallPolicies { get { throw null; } set { } }
```
JoshLove-msft, line 953: Should the this still be a settable array? Or an IList with a getter only?

sdk/core/Azure.Core/src/Pipeline/Internal/HttpPipelineTransportPolicy.cs
```diff
@@ -11,11 +11,13 @@ internal class HttpPipelineTransportPolicy : HttpPipelinePolicy
     {
         private readonly HttpPipelineTransport _transport;
         private readonly HttpMessageSanitizer _sanitizer;
+        private readonly RequestFailedDetailsParser? _failureContentExtractor;
```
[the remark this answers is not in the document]
JoshLove-msft: or something similar to reflect the new name.

sdk/core/Azure.Core/src/Pipeline/MultipartResponse.cs
```diff
@@ -38,8 +39,39 @@ internal static class Multipart
...
+        /// <param name="response">The response containing multi-part content.</param>
+        /// <param name="expectCRLF">Controls whether the parser will expect all multi-part boundaries to use CRLF line breaks. This should be true unless more permissive line break parsing is required.</param>
+        /// <param name="cancellationToken">
+        /// Optional <see cref="CancellationToken"/> to propagate notifications
+        /// that the operation should be cancelled.
+        /// </param>
+        /// <returns>The parsed <see cref="Response"/>s.</returns>
+        public static Response[] Parse(
+            Response response,
+            bool expectCRLF,
```
JoshLove-msft, line 51: expectWindowsNewLine?

sdk/core/Azure.Core/api/Azure.Core.net461.cs
christothes: Now that this is not taken in the ctor, it should have a setter. Since we ultimately need the array, I think it makes sense to stay explicitly as an array.

sdk/core/Azure.Core/src/Pipeline/MultipartResponse.cs
christothes: My opinion here is that `WindowsNewLine` could be a bit obscure (Is it clear that this means CRLF?). Curious to hear @KrzysztofCwalina 's thoughts.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
JoshLove-msft: I thought there was a guideline against this. I found this analyzer -https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1819

sdk/core/Azure.Core/src/Pipeline/MultipartResponse.cs
JoshLove-msft: Is this parameter essentially saying whether the boundary will start with a CrLf?

christothes: yes - https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/Shared/Multipart/MultipartBoundary.cs#L31-L38

JoshLove-msft: The RFC seems to suggest that this is required - https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html. Is there a use case for omitting this?

christothes: Somebody forgot to tell the service about this requirement :)

sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -932,6 +943,17 @@ public static partial class HttpPipelineBuilder
...
+        public static Azure.Core.Pipeline.HttpPipeline Build(Azure.Core.Pipeline.HttpPipelineOptions options) { throw null; }
+        public static Azure.Core.Pipeline.DisposableHttpPipeline Build(Azure.Core.Pipeline.HttpPipelineOptions options, Azure.Core.Pipeline.HttpPipelineTransportOptions transportOptions) { throw null; }
+    }
+    public partial class HttpPipelineOptions
+    {
+        public HttpPipelineOptions(Azure.Core.ClientOptions options) { }
+        public Azure.Core.ClientOptions ClientOptions { get { throw null; } }
+        public System.Collections.Generic.IList<Azure.Core.Pipeline.HttpPipelinePolicy> PerCallPolicies { get { throw null; } }
+        public System.Collections.Generic.IList<Azure.Core.Pipeline.HttpPipelinePolicy> PerRetryPolicies { get { throw null; } }
+        public Azure.Core.RequestFailedDetailsParser RequestFailedDetailsParser { get { throw null; } set { } }
```
JoshLove-msft, line 955: Why is this not nullable but ResponseClassifier is nullable?

JoshLove-msft (approved):

sdk/core/Azure.Core/api/Azure.Core.net461.cs
JoshLove-msft: I guess we don't set it to the default response classifier?

christothes: The implementation sets a default in the ctor.

Issue #27828, opened 2022-03-29
Document distributed tracing policy


lmolkova:
### Library name

Azure.Core

### Please describe the feature.

https://azure.github.io/azure-sdk/dotnet_implementation.html#dotnet-distributedtracing - we should document the policy here
- public methods
- pipeline policy
- grow-up methods

lmolkova:
Fixed in https://github.com/Azure/azure-sdk/pull/4367

PR #28469, opened 2022-04-29
Validate that library scope is present when using ForwardClientCalls

### LINKED ISSUE #11966: Test for diagnostic scope doesn't seem to fail if there is no diagnostic scope
When PR #11860 was created, it effectively removed `DiagnosticScope`s from Search, since we were calling into the *RestClient classes directly, bypassing the *Client classes that were generated. Since creating `DiagnosticScope`s was moved up from *RestClient to *Client classes, there should've been a failure in the tests for `DiagnosticScope`. I know, at least, then they fail when the scope name doesn't match the method name (sans "Async") because, when refactoring Key Vault toward the end before release, I forgot to change it in a couple places. But, seemingly, if there's no scope, the tests don't fail and probably should so we can catch cases like this where there is no more distributed tracing for this or any other reason.

JoshLove-msft:
Fixes https://github.com/Azure/azure-sdk-for-net/issues/11966

### DIFF
--- a/sdk/core/Azure.Core.TestFramework/src/Instrumentation/DiagnosticScopeValidatingInterceptor.cs
+++ b/sdk/core/Azure.Core.TestFramework/src/Instrumentation/DiagnosticScopeValidatingInterceptor.cs
@@ -215,7 +215,7 @@ namespace Azure.Core.TestFramework
                         {
                             throw new InvalidOperationException(
                                 "Expected some diagnostic scopes to be created other than the Azure.Core scopes, but no such scopes were present. " +
-                                $"Ensure that the inner method that client calls are being forwarded to from the '{{source}}' method has diagnostic scopes " +
+                                $"Ensure that the inner method that client calls are being forwarded to from the '{source}' method has diagnostic scopes " +
                                 "defined by using clientDiagnostics.CreateScope(...).");
                         }
                     }
--- a/sdk/core/Azure.Core.TestFramework/src/Instrumentation/DiagnosticScopeValidatingInterceptor.cs
+++ b/sdk/core/Azure.Core.TestFramework/src/Instrumentation/DiagnosticScopeValidatingInterceptor.cs
@@ -209,8 +209,8 @@ namespace Azure.Core.TestFramework
                     }
                     else
                     {
-                        // If ForwardsClientCallsAttribute is being used on the method, we don't know what the name of the scope should be (because there could be many
-                        // differently named methods sharing the same scope name, but we know that there should be some scope created other than the Azure.Core scope.
+                        // If ForwardsClientCallsAttribute is being used on the method, we don't know what the name of the scope should be because there could be many
+                        // differently named methods sharing the same scope name, but we do know that there should be some scope created other than the Azure.Core scope.
                         if (!diagnosticListener.Scopes.Any(e => !e.Name.StartsWith("Azure.Core")))
                         {
                             throw new InvalidOperationException(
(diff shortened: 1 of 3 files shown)


sdk/core/Azure.Core.TestFramework/src/Instrumentation/DiagnosticScopeValidatingInterceptor.cs
```diff
@@ -209,9 +209,13 @@ internal static async ValueTask<T> ValidateDiagnosticScope<T>(Func<ValueTask<(T
...
                     else
                     {
-                        if (!diagnosticListener.Scopes.Any())
+                        // If ForwardsClientCallsAttribute is being used on the method, we don't know what the name of the scope should be because there could be many
+                        // differently named methods sharing the same scope name, but we do know that there should be some scope created other than the Azure.Core scope.
+                        if (!diagnosticListener.Scopes.Any(e => !e.Name.StartsWith("Azure.Core")))
                         {
-                            throw new InvalidOperationException($"Expected some diagnostic scopes to be created, found none");
+                            throw new InvalidOperationException(
+                                $"Expected some diagnostic scopes to be created other than the Azure.Core scopes, but no such scopes were present. {Environment.NewLine} Ensure that the inner method that " +
```
heaths, line 217: Nit: remove the space after the new line (alignment will be wonky otherwise), or just remove the new line entirely. It's rare to see or need formatted exception message.

heaths (approved):

sdk/core/Azure.Core.TestFramework/src/Instrumentation/DiagnosticScopeValidatingInterceptor.cs
```diff
@@ -209,9 +209,14 @@ internal static async ValueTask<T> ValidateDiagnosticScope<T>(Func<ValueTask<(T
                     }
                     else
                     {
-                        if (!diagnosticListener.Scopes.Any())
+                        // If ForwardsClientCallsAttribute is being used on the method, we don't know what the name of the scope should be because there could be many
```
annelo-msft, line 212: Why are Azure.Core scopes created?  Scopes are supposed to correspond to customer calls of public methods.  Or are these the inner HTTP scopes?

annelo-msft: Maybe we could defined "The Azure.Core scope?"

annelo-msft: Per https://github.com/Azure/azure-sdk-for-net/issues/27828#issuecomment-1115069467, perhaps we could call the "nested HTTP scopes"?

JoshLove-msft: > Why are Azure.Core scopes created? Scopes are supposed to correspond to customer calls of public methods. Or are these the inner HTTP scopes?

Azure.Core creates scopes for outgoing requests - https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/Pipeline/Internal/RequestActivityPolicy.cs#L57

JoshLove-msft: > , perhaps we could call the "nested HTTP scopes"?

I'm not sure what this comment is referring to.

annelo-msft: > there should be some scope created other than the Azure.Core scope.

It's a nit, but in the distributed tracing guidelines, I've heard these referred to as the "nested HTTP/REST spans", and never as Azure.Core scopes, see https://azure.github.io/azure-sdk/general_implementation.html#distributed-tracing

JoshLove-msft: Ah I see. That is a fair point.

PR #28472, opened 2022-04-29
[AVA] Add deprecation notice to Video Analyzer data plane sdk

hivyas:
Added a deprecation notice to Video Analyzer data plane sdk


heaths (approved):
Once the new package is published, we may want to remove it from the repo. The source will always be there, but does improve performance of some activities like initial checking out of files from the index. Or maybe after December.

Just a thought.

heaths:
Do you want me to merge this as soon as the build completes successfully? Or should we wait a while? Package builds are kicked off manually, so merely merging this won't provoke any publishing or anything.

giakas (approved):

PR #28473, opened 2022-04-29
[TA] Reuse ConvertToMultiLanguageInputs functions

maririos:
We already have functions that do the conversion for us and that applies to all non-batch endpoints. Instead of creating specific ones, we should just reuse these ones.

You can see test passing with the fix in the draft PR => https://github.com/Azure/azure-sdk-for-net/pull/28474

### DIFF
--- a/sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/LanguageServiceClient.cs
+++ b/sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/LanguageServiceClient.cs
@@ -1174,9 +1174,10 @@ namespace Azure.AI.TextAnalytics.ServiceClients
         private MultiLanguageAnalysisInput ConvertToMultiLanguageInputs(IEnumerable<string> documents, string language)
         {
             MultiLanguageAnalysisInput input = new MultiLanguageAnalysisInput();
+            int i = 0;
             foreach (var document in documents)
             {
-                input.Documents.Add(ConvertToMultiLanguageInput(document, language));
+                input.Documents.Add(ConvertToMultiLanguageInput(document, language, i++));
             }
             return input;
         }
--- a/sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/LanguageServiceClient.cs
+++ b/sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/LanguageServiceClient.cs
@@ -343,7 +343,7 @@ namespace Azure.AI.TextAnalytics.ServiceClients
         {
             Argument.AssertNotNullOrEmpty(documents, nameof(documents));
             options ??= s_defaultRequestOptions;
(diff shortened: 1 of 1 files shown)


nisha-bhatia (approved):
LGTM

jsquire (approved):

sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/LanguageServiceClient.cs
```diff
@@ -429,34 +445,6 @@ private Response<RecognizeEntitiesResultCollection> RecognizeEntitiesBatch(Analy
...
-        private static AnalyzeTextEntityRecognitionInput DocumentsToEntityRecognition(IEnumerable<string> documents, string language)
-        {
-            AnalyzeTextEntityRecognitionInput textEntityInputs = new AnalyzeTextEntityRecognitionInput();
-            int id = 0;
-            foreach (var document in documents)
-            {
-                var input = new MultiLanguageInput(id: id.ToString(CultureInfo.InvariantCulture), text: document);
-                id++;
-                input.Language = language;
-                textEntityInputs.AnalysisInput.Documents.Add(input);
```
maririos, line 441: @nisha-bhatia the problem was in here because `AnalysisInput` was never instantiated.
See how you created an object `AnalyzeTextEntityRecognitionInput` with the name `textEntityInputs`. Its [constructor](https://github.com/Azure/azure-sdk-for-net/blob/feature/text-analytics-language-api/sdk/textanalytics/Azure.AI.TextAnalytics/src/Generated/Models/AnalyzeTextEntityRecognitionInput.cs#L14) doesn't initialize either the `AnalysisInput` or the `Parameters` properties.
So when you try to do `textEntityInputs.AnalysisInput.Documents.Add()` it throws a `NullReferenceException`.

A fix could have been to do:
```
AnalyzeTextEntityRecognitionInput textEntityInputs = new();
AnalysisInput analysisInput= new();
int id = 0;
foreach (var document in documents)
{
    var input = new MultiLanguageInput(id: id.ToString(CultureInfo.InvariantCulture), text: document);
    id++;
    input.Language = language;
    input.Documents.Add(input);
}
textEntityInputs.AnalysisInput = analysisInput;
return textEntityInputs;
```

To keep improving the function notice how we don't need the `AnalyzeTextEntityRecognitionInput` object yet. Here, we just want to transform user input into `AnalysisInput`, so the function will then look like this:

```
private static MultiLanguageAnalysisInput DocumentsToEntityRecognition(IEnumerable<string> documents, string language)
{
    MultiLanguageAnalysisInput analysisInput= new MultiLanguageAnalysisInput();
    int i = 0;
    foreach (var document in documents)
    {
         analysisInput.Documents.Add(ConvertToMultiLanguageInput(document, language, i++));
     }
     return analysisInput;
}
```

`MultiLanguageAnalysisInput` is used for all inputs so at this point we just need a general function to do this transformation. And it turns out we already have it, it is called `ConvertToMultiLanguageInputs`.

PR #28476, opened 2022-04-30
Smoke Tests:  Fix dangling parenthesis

jsquire:
The focus of these changes is to fix a dangling parenthesis emitted as part of information log messages when running in CI.    This is specific to the CI formatting and does not occur for local runs.


ebwinters:
do you have a picture or link to a log of the dangling parenthesis occuring?

jsquire:
> do you have a picture or link to a log of the dangling parenthesis occurring? Moreso just curious

It's pretty much exactly what it sounds like.  😄    

Note the unmatched trailing close parenthesis from the typo in each log message:
image

heaths (approved):

PR #28480, opened 2022-04-30
Retry timeouts for recorded tests

### LINKED ISSUE #26311: [FEATURE REQ] Retry playback test timeouts once
### Library name

Azure.Core.TestFramework

Sometimes a test will get unlucky and timeout. There is nothing we can do to "fix" it. So to reduce the flakiness we should retry the timeout once.

/cc @JoshLove-msft @christothes

JoshLove-msft:
Fixes https://github.com/Azure/azure-sdk-for-net/issues/26311

### DIFF
--- a/sdk/core/Azure.Core.TestFramework/src/ClientTestBase.cs
+++ b/sdk/core/Azure.Core.TestFramework/src/ClientTestBase.cs
@@ -54,9 +54,13 @@ namespace Azure.Core.TestFramework
             var timeout = TestEnvironment.GlobalIsRunningInCI ? GLOBAL_TEST_TIMEOUT_IN_SECONDS : GLOBAL_LOCAL_TEST_TIMEOUT_IN_SECONDS;
             if (duration > TimeSpan.FromSeconds(timeout))
             {
-                executionContext.CurrentResult.SetResult(
-                    ResultState.Failure,
-                    $"Test exceeded global time limit of {timeout} seconds. Duration: {duration}");
+                string message = $"Test exceeded global time limit of {timeout} seconds. Duration: {duration} ";
+                if (this is RecordedTestBase &&
+                    !executionContext.CurrentTest.GetCustomAttributes<RecordedTestAttribute>(true).Any())
+                {
+                    message += Environment.NewLine + "Add the [RecordedTest] attribute to your test to allow an automatic retry for timeouts.";
+                }
+                throw new TestTimeoutException(message);
             }
         }
 
--- a/sdk/core/Azure.Core.TestFramework/src/ClientTestBase.cs
+++ b/sdk/core/Azure.Core.TestFramework/src/ClientTestBase.cs
@@ -58,7 +58,7 @@ namespace Azure.Core.TestFramework
                 if (this is RecordedTestBase &&
                     !executionContext.CurrentTest.GetCustomAttributes<RecordedTestAttribute>(true).Any())
                 {
-                    message += Environment.NewLine + "Add the [RecordedTest] attribute to your test to allow an automatic retry for timeouts.";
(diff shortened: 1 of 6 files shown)


sdk/core/Azure.Core.TestFramework/src/ClientTestBase.cs
```diff
@@ -54,9 +54,13 @@ public virtual void GlobalTimeoutTearDown()
...
             if (duration > TimeSpan.FromSeconds(timeout))
             {
-                executionContext.CurrentResult.SetResult(
-                    ResultState.Failure,
-                    $"Test exceeded global time limit of {timeout} seconds. Duration: {duration}");
+                string message = $"Test exceeded global time limit of {timeout} seconds. Duration: {duration} ";
+                if (this is RecordedTestBase &&
+                    !executionContext.CurrentTest.GetCustomAttributes<RecordedTestAttribute>(true).Any())
+                {
+                    message += Environment.NewLine + "Add the [RecordedTest] attribute to your test to allow an automatic retry for timeouts.";
```
christothes, line 61:
```suggestion
                    message += Environment.NewLine + "Replace the [Test] attribute with the [RecordedTest] attribute to your test to allow an automatic retry for timeouts.";
```

sdk/core/Azure.Core.TestFramework/src/RecordedTestAttribute.cs
```diff
@@ -43,44 +43,73 @@ public override TestResult Execute(TestExecutionContext context)
...
-                            "The [RecordedTest] attribute attempted to re-record, but failed: " + Environment.NewLine +
-                            context.CurrentResult.Message + Environment.NewLine);
-                    }
+                        // retry once
+                        context.CurrentResult = context.CurrentTest.MakeTestResult();
+                        context.CurrentResult = innerCommand.Execute(context);
 
-                    // revert RecordTestMode to Playback
-                    SetRecordMode(context.TestObject as RecordedTestBase, RecordedTestMode.Playback);
+                        if (context.CurrentResult.ResultState.Status != TestStatus.Passed)
```
christothes, line 85:
```suggestion
                        if (IsTestFailed(context))
```

sdk/core/Azure.Core.TestFramework/src/RecordedTestAttribute.cs
```diff
@@ -43,44 +43,73 @@ public override TestResult Execute(TestExecutionContext context)
...
+                        {
+                            context.CurrentResult.SetResult(
+                                ResultState.Error,
+                                "The test timed out twice:" + Environment.NewLine +
+                                $"First attempt: {originalResult.Message}" + Environment.NewLine +
+                                $"Second attempt: {context.CurrentResult.Message}");
+                        }
+                        else
+                        {
+                            context.CurrentResult.SetResult(
```
christothes, line 95: Should we include the original state here as well just to indicate how far out of the timeout we were?

sdk/core/Azure.Core.TestFramework/src/ClientTestBase.cs
```diff
@@ -54,9 +54,13 @@ public virtual void GlobalTimeoutTearDown()
...
+                }
+                throw new TestTimeoutException(message);
```
christothes, line 63: Should we still be calling `executionContext.CurrentResult.SetResult(` ? I'm not sure if a test with exception handling would potentially still pass here.

christothes (approved):

sdk/core/Azure.Core.TestFramework/src/ClientTestBase.cs
JoshLove-msft: I don't think it's necessary - this exception is raised during TearDown and I don't believe there is a way for a test to prevent this from throwing.

PR #28484, opened 2022-05-02
[TA] PII Entities fix, enable live tests and samples

maririos:
Although code was building, the `PiiEntitiesParameters` needed some work.
Live tests, parameters tests, and samples running

### DIFF
--- a/sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/LanguageServiceClient.cs
+++ b/sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/LanguageServiceClient.cs
@@ -466,13 +466,12 @@ namespace Azure.AI.TextAnalytics.ServiceClients
                 {
                     input.Documents.Add(doc);
                 }
-                var analyzePiiEntities = new AnalyzeTextPiiEntitiesRecognitionInput { AnalysisInput = input };
-                analyzePiiEntities.Parameters = new PiiTaskParameters(
-                    options.DisableServiceLogs,
-                    options.ModelVersion,
-                    new PiiDomain(options.DomainFilter.GetString()),
-                    options.CategoriesFilter,
-                    Constants.DefaultStringIndexType);
+
+                AnalyzeTextPiiEntitiesRecognitionInput analyzePiiEntities = new()
+                {
+                    AnalysisInput = input,
+                    Parameters = PiiEntitiesParameters(options)
+                };
 
                 Response<AnalyzeTextTaskResult> result = await _languageRestClient.AnalyzeAsync(analyzePiiEntities, cancellationToken: cancellationToken).ConfigureAwait(false);
 
@@ -513,13 +512,12 @@ namespace Azure.AI.TextAnalytics.ServiceClients
                 {
                     input.Documents.Add(doc);
                 }
-                var analyzePiiEntities = new AnalyzeTextPiiEntitiesRecognitionInput { AnalysisInput = input };
-                analyzePiiEntities.Parameters = new PiiTaskParameters(
-                    options.DisableServiceLogs,
-                    options.ModelVersion,
-                    new PiiDomain(options.DomainFilter.GetString()),
-                    options.CategoriesFilter,
-                    Constants.DefaultStringIndexType);
+
+                AnalyzeTextPiiEntitiesRecognitionInput analyzePiiEntities = new()
+                {
+                    AnalysisInput = input,
+                    Parameters = PiiEntitiesParameters(options)
+                };
 
                 Response<AnalyzeTextTaskResult> result = _languageRestClient.Analyze(analyzePiiEntities, cancellationToken: cancellationToken);
                 var piiEntities = (PiiTaskResult)result.Value;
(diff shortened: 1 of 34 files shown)


sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/LanguageServiceClient.cs
```diff
@@ -652,6 +640,23 @@ private Response<RecognizePiiEntitiesResultCollection> RecognizePiiEntitiesBatch
...
+            {
+                LoggingOptOut = options.DisableServiceLogs,
+                ModelVersion = options.ModelVersion,
+                Domain = options.DomainFilter.GetString() ?? (PiiDomain?)null,
+                StringIndexType = Constants.DefaultStringIndexType
+            };
+
+            if (options.CategoriesFilter.Count > 0)
+            {
+                parameters.PiiCategories = options.CategoriesFilter;
```
maririos, line 655: just by touching `PiiCategories`, even if it is to assign an empty list, codegen will add it to the content of the request (empty value though), but we don't want that. 
We had this problem before and did this (see [here](https://github.com/Azure/azure-sdk-for-net/blob/feature/text-analytics-language-api/sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.Legacy.cs#L486) in the Legacy code for PiiActions). There might be a better solution, but putting this one out for now.

sdk/textanalytics/Azure.AI.TextAnalytics/tests/RecognizePiiEntitiesTests.cs
```diff
@@ -10,8 +10,8 @@
 
 namespace Azure.AI.TextAnalytics.Tests
 {
-    [ServiceVersion(Min = TextAnalyticsClientOptions.ServiceVersion.V3_1)]
-    [Ignore("Not yet implemented")]
+    //[ServiceVersion(Min = TextAnalyticsClientOptions.ServiceVersion.V3_1)]
+    [ClientTestFixture(TextAnalyticsClientOptions.ServiceVersion.V2022_03_01_Preview)]
```
maririos, line 14: Before adding more tests to other functionalities, I'm going to start working on enabling tests for all versions. This means I will delete enable `[ServiceVersion(Min = TextAnalyticsClientOptions.ServiceVersion.V3_1)]` and remove the `ClientTestFixture` from here

sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/LanguageServiceClient.cs
```diff
@@ -466,13 +466,12 @@ public override async Task<Response<PiiEntityCollection>> RecognizePiiEntitiesAs
...
                 }
-                var analyzePiiEntities = new AnalyzeTextPiiEntitiesRecognitionInput { AnalysisInput = input };
-                analyzePiiEntities.Parameters = new PiiTaskParameters(
-                    options.DisableServiceLogs,
-                    options.ModelVersion,
-                    new PiiDomain(options.DomainFilter.GetString()),
-                    options.CategoriesFilter,
-                    Constants.DefaultStringIndexType);
+
+                AnalyzeTextPiiEntitiesRecognitionInput analyzePiiEntities = new()
```
jsquire, line 470: oooohhh... so much cleaner

sdk/textanalytics/Azure.AI.TextAnalytics/tests/samples/Sample5_RecognizePiiEntitiesAsync.cs
```diff
@@ -15,7 +15,7 @@ public async Task RecognizePiiEntitiesAsync()
             string endpoint = TestEnvironment.Endpoint;
             string apiKey = TestEnvironment.ApiKey;
 
-            var client = new TextAnalyticsClient(new Uri(endpoint), new AzureKeyCredential(apiKey), CreateSampleOptions());
+            var client = new TextAnalyticsClient(new Uri(endpoint), new AzureKeyCredential(apiKey), CreateSampleOptions(TextAnalyticsClientOptions.ServiceVersion.V2022_03_01_Preview));
```
jsquire, line 18: Strange not to see any kind of snippet anywhere in this one.   Should we consider removing it?

sdk/textanalytics/Azure.AI.TextAnalytics/tests/samples/Sample5_RecognizePiiEntitiesBatchAsync.cs
```diff
@@ -17,7 +17,7 @@ public async Task RecognizePiiEntitiesBatchAsync()
             string apiKey = TestEnvironment.ApiKey;
 
             // Instantiate a client that will be used to call the service.
-            var client = new TextAnalyticsClient(new Uri(endpoint), new AzureKeyCredential(apiKey), CreateSampleOptions());
+            var client = new TextAnalyticsClient(new Uri(endpoint), new AzureKeyCredential(apiKey), CreateSampleOptions(TextAnalyticsClientOptions.ServiceVersion.V2022_03_01_Preview));
```
jsquire, line 20: There's no snippet to consume this method; should we consider removing it?

jsquire (approved):

sdk/textanalytics/Azure.AI.TextAnalytics/tests/samples/Sample5_RecognizePiiEntitiesAsync.cs
maririos: I'm going to leave it as-is for now. Let me know what you think and if we would like to make this change,I will create a wide issue as it will impact all of our samples.

PR #28494, opened 2022-05-02
Add LinkedEntities to Azure.AI.TextAnalytics

nisha-bhatia:

### DIFF
--- a/sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/LanguageServiceClient.cs
+++ b/sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/LanguageServiceClient.cs
@@ -892,115 +892,186 @@ namespace Azure.AI.TextAnalytics.ServiceClients
 
         public override async Task<Response<LinkedEntityCollection>> RecognizeLinkedEntitiesAsync(string document, string language = default, CancellationToken cancellationToken = default)
         {
-            //Argument.AssertNotNullOrEmpty(document, nameof(document));
+            Argument.AssertNotNullOrEmpty(document, nameof(document));
 
-            //using DiagnosticScope scope = _clientDiagnostics.CreateScope($"{nameof(TextAnalyticsClient)}.{nameof(RecognizeLinkedEntities)}");
-            //scope.AddAttribute("document", document);
-            //scope.Start();
+            using DiagnosticScope scope = _clientDiagnostics.CreateScope($"{nameof(TextAnalyticsClient)}.{nameof(RecognizeLinkedEntities)}");
+            scope.AddAttribute("document", document);
+            scope.Start();
 
-            //try
-            //{
-            //    var documents = new List<MultiLanguageInput>() { ConvertToMultiLanguageInput(document, language) };
+            try
+            {
+                var documents = new List<MultiLanguageInput>() { ConvertToMultiLanguageInput(document, language) };
+                var input = new MultiLanguageAnalysisInput();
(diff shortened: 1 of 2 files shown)


sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.cs
```diff
@@ -80,7 +80,7 @@ internal static DetectedLanguage ConvertToDetectedLanguage(LanguageDetectionDocu
 
         internal static DetectLanguageResultCollection ConvertToDetectLanguageResultCollection(LanguageDetectionResult results, IDictionary<string, int> idToIndexMap)
         {
-            var detectedLanguages = new List<DetectLanguageResult>(results.Errors.Count);
```
nisha-bhatia, line 83: I updated this line because I don't think we need to give a length to the list before. When we add the DetectLanguageResults the list will grow to the size needed

nisha-bhatia: I also updated this for Recognize Entities and Linked Entities

maririos: This is true. There is a scenario when, if the list is big, setting the size beforehand is more performant.
See Remarks section [here](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.-ctor?redirectedfrom=MSDN&view=net-6.0#System_Collections_Generic_List_1__ctor_System_Int32_):
> If the size of the collection can be estimated, using the [List<T>(Int32)](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.-ctor?view=net-6.0#system-collections-generic-list-1-ctor(system-int32)) constructor and specifying the initial capacity eliminates the need to perform a number of resizing operations while adding elements to the [List<T>](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1?view=net-6.0).)).

I would prefer to leave as is just in case but don't feel strongly about it as the chances for the `Errors` array to be big are small. What do you think?

nisha-bhatia: I think if we want to specify the initial length of the list, maybe we can give the results.Documents length instead of the error? I think there will be more documents than errors so this way we can reduce the chance of resizing as well

maririos: ohhh interesting!!! yes please do that!!!

sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/LanguageServiceClient.cs
```diff
@@ -906,7 +906,11 @@ public override async Task<Response<LinkedEntityCollection>> RecognizeLinkedEnti
                 {
                     input.Documents.Add(doc);
                 }
-                var analyzeRecognizeEntities = new AnalyzeTextEntityLinkingInput { AnalysisInput = input };
+                AnalyzeTextEntityLinkingInput analyzeRecognizeEntities = new()
```
maririos, line 909: technically this should be `analyzeRecognizeLinkedEntities`

maririos (commented):
Before you merge, please make sure people look at your code and approve it.

nisha-bhatia:
> Before you merge, please make sure people look at your code and approve it.

sorry, I thought you said it looked good :( I'll make a pr to fix the comment

PR #28496, opened 2022-05-03
Keep the SuppressClientNestedActivities flag private for now

lmolkova:
Addressing [Review feedback](https://apiview.dev/Assemblies/Review/77bdbcef49a3465fa96fac3fb745bf57/cb4a44f8dda24ee49d5410d986524c6e?diffRevisionId=532c07579b3146faad925857b4d5353d&doc=False&diffOnly=True)

Related: https://github.com/Azure/azure-sdk/issues/4222

1. Do not  expose `SuppressNestedClientSpans` until we have some clear request, make it an explicit parameter on `ClientDiagnsotics`
2. Use `Activity` terminology instead of `span` similar to OpenTelemetry in .NET and BCL

### DIFF
--- a/sdk/core/Azure.Core/src/Shared/ClientDiagnostics.cs
+++ b/sdk/core/Azure.Core/src/Shared/ClientDiagnostics.cs
@@ -31,7 +31,7 @@ namespace Azure.Core.Pipeline
         }
 
         public ClientDiagnostics(string optionsNamespace, string? providerNamespace, DiagnosticsOptions diagnosticsOptions, bool? suppressNestedClientActivities = null)
-            : base(optionsNamespace, providerNamespace, diagnosticsOptions.IsDistributedTracingEnabled, suppressNestedClientActivities)
+            : base(optionsNamespace, providerNamespace, diagnosticsOptions.IsDistributedTracingEnabled, suppressNestedClientActivities.GetValueOrDefault(false))
         {
             // TODO: use DiagnosticsOptions.SuppressNestedClientSpans  here after new Azure.Core ships:
             // ClientDiagnostics class is shared as a source code and used in client libraries.
--- a/sdk/core/Azure.Core/src/Shared/ClientDiagnostics.cs
+++ b/sdk/core/Azure.Core/src/Shared/ClientDiagnostics.cs
@@ -22,15 +22,16 @@ namespace Azure.Core.Pipeline
 
         private readonly HttpMessageSanitizer _sanitizer;
 
-        public ClientDiagnostics(ClientOptions options)
+        public ClientDiagnostics(ClientOptions options, bool? suppressNestedClientActivities = null)
                     : this(options.GetType().Namespace!,
                     GetResourceProviderNamespace(options.GetType().Assembly),
-                    options.Diagnostics)
+                    options.Diagnostics,
+                    suppressNestedClientActivities)
         {
         }
 
-        public ClientDiagnostics(string optionsNamespace, string? providerNamespace, DiagnosticsOptions diagnosticsOptions)
-            : base(optionsNamespace, providerNamespace, diagnosticsOptions.IsDistributedTracingEnabled, false)
+        public ClientDiagnostics(string optionsNamespace, string? providerNamespace, DiagnosticsOptions diagnosticsOptions, bool? suppressNestedClientActivities = null)
+            : base(optionsNamespace, providerNamespace, diagnosticsOptions.IsDistributedTracingEnabled, suppressNestedClientActivities)
         {
             // TODO: use DiagnosticsOptions.SuppressNestedClientSpans  here after new Azure.Core ships:
             // ClientDiagnostics class is shared as a source code and used in client libraries.
--- a/sdk/core/Azure.Core/src/Shared/ClientDiagnostics.cs
+++ b/sdk/core/Azure.Core/src/Shared/ClientDiagnostics.cs
@@ -28,7 +28,7 @@ namespace Azure.Core.Pipeline
         /// <param name="options">The customer provided client options object.</param>
         /// <param name="suppressNestedClientActivities">Flag controlling if <see cref="System.Diagnostics.Activity"/>
         ///  created by this <see cref="ClientDiagnostics"/> for client method calls should be suppressed when called
-        ///  by other Azure SDK client methods.  It's recommended to set it to true for new clients; use default (null) for
+        ///  by other Azure SDK client methods.  It's recommended to set it to true for new clients; use default (null)
         ///  for backward compatibility reasons, or set it to false to explicitly disable suppression for specific cases.
         ///  The default value could change in the future, the flag should be only set to false if suppression for the client
         ///  should never be enabled.</param>
--- a/sdk/core/Azure.Core/src/Shared/ClientDiagnostics.cs
+++ b/sdk/core/Azure.Core/src/Shared/ClientDiagnostics.cs
@@ -33,8 +33,6 @@ namespace Azure.Core.Pipeline
         public ClientDiagnostics(string optionsNamespace, string? providerNamespace, DiagnosticsOptions diagnosticsOptions, bool? suppressNestedClientActivities = null)
             : base(optionsNamespace, providerNamespace, diagnosticsOptions.IsDistributedTracingEnabled, suppressNestedClientActivities.GetValueOrDefault(false))
         {
-            // TODO: use DiagnosticsOptions.SuppressNestedClientSpans  here after new Azure.Core ships:
-            // ClientDiagnostics class is shared as a source code and used in client libraries.
             _sanitizer = CreateMessageSanitizer(diagnosticsOptions);
         }
 
--- a/sdk/core/Azure.Core/src/Shared/ClientDiagnostics.cs
+++ b/sdk/core/Azure.Core/src/Shared/ClientDiagnostics.cs
@@ -22,6 +22,16 @@ namespace Azure.Core.Pipeline
 
         private readonly HttpMessageSanitizer _sanitizer;
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ClientDiagnostics"/> class.
+        /// </summary>
+        /// <param name="options">The customer provided client options object.</param>
+        /// <param name="suppressNestedClientActivities">Flag controlling if <see cref="System.Diagnostics.Activity"/>
+        ///  created by this <see cref="ClientDiagnostics"/> for client method calls should be suppressed when called
+        ///  by other Azure SDK client methods.  It's recommended to set it to true for new clients; use default (null) for
+        ///  for backward compatibility reasons, or set it to false to explicitly disable suppression for specific cases.
+        ///  The default value could change in the future, the flag should be only set to false if suppression for the client
+        ///  should never be enabled.</param>
         public ClientDiagnostics(ClientOptions options, bool? suppressNestedClientActivities = null)
                     : this(options.GetType().Namespace!,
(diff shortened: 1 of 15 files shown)


sdk/core/Azure.Core/src/Shared/DiagnosticScopeFactory.cs
```diff
@@ -16,13 +16,14 @@ internal class DiagnosticScopeFactory
         private static Dictionary<string, DiagnosticListener>? _listeners;
         private readonly string? _resourceProviderNamespace;
         private readonly DiagnosticListener? _source;
-        private readonly bool _suppressNestedClientScopes;
+        private readonly bool _suppressNestedClientActivities;
 
-        public DiagnosticScopeFactory(string clientNamespace, string? resourceProviderNamespace, bool isActivityEnabled, bool suppressNestedClientScopes = true)
+        public DiagnosticScopeFactory(string clientNamespace, string? resourceProviderNamespace, bool isActivityEnabled, bool suppressNestedClientActivities)
```
JoshLove-msft, line 21: Just curious why we are removing the default.

sdk/core/Azure.Core/src/Shared/ClientDiagnostics.cs
```diff
@@ -22,15 +22,16 @@ internal class ClientDiagnostics : DiagnosticScopeFactory
...
                     GetResourceProviderNamespace(options.GetType().Assembly),
-                    options.Diagnostics)
+                    options.Diagnostics,
+                    suppressNestedClientActivities)
         {
         }
 
-        public ClientDiagnostics(string optionsNamespace, string? providerNamespace, DiagnosticsOptions diagnosticsOptions)
-            : base(optionsNamespace, providerNamespace, diagnosticsOptions.IsDistributedTracingEnabled, false)
+        public ClientDiagnostics(string optionsNamespace, string? providerNamespace, DiagnosticsOptions diagnosticsOptions, bool? suppressNestedClientActivities = null)
```
JoshLove-msft, line 33: Why not `bool suppressNestedClientActivities = false` so callers know the default behavior more easily?

sdk/core/Azure.Core/src/Shared/ClientDiagnostics.cs
```diff
@@ -22,15 +22,16 @@ internal class ClientDiagnostics : DiagnosticScopeFactory
...
+            : base(optionsNamespace, providerNamespace, diagnosticsOptions.IsDistributedTracingEnabled, suppressNestedClientActivities.GetValueOrDefault(false))
         {
             // TODO: use DiagnosticsOptions.SuppressNestedClientSpans  here after new Azure.Core ships:
```
JoshLove-msft, line 36: nit: remove

sdk/core/Azure.Core/tests/ClientDiagnosticsTests.Net50.cs
```diff
@@ -165,24 +162,6 @@ public void NestedClientActivitiesSuppressed(int kind, bool expectSuppression)
             Assert.AreEqual("ClientName.ActivityName", Activity.Current.OperationName);
         }
 
-        [Test]
-        [NonParallelizable]
-        public void NestedClientActivitiesSuppressedDefaultDiagnosticScopeFactory()
```
JoshLove-msft, line 170: Why remove these tests?

sdk/core/Azure.Core/src/Shared/DiagnosticScopeFactory.cs
```diff
@@ -16,13 +16,14 @@ internal class DiagnosticScopeFactory
...
         {
             _resourceProviderNamespace = resourceProviderNamespace;
             IsActivityEnabled = isActivityEnabled;
-            _suppressNestedClientScopes = suppressNestedClientScopes;
+            _suppressNestedClientActivities = suppressNestedClientActivities;
+            ;
```
christothes, line 26:
```suggestion
```

sdk/core/Azure.Core/src/Shared/ClientDiagnostics.cs
annelo-msft: +1, curious about the reasoning behind having it be nullable?

sdk/core/Azure.Core/src/Shared/ClientDiagnostics.cs
annelo-msft: To make sure I understand: this will now be enabled once DPG passes `suppressNestedClientActivities = true` in ClientDiagnostics constructor? (per https://github.com/ShivangiReja/autorest.csharp/pull/2/files#diff-812ba0c844167e48b314a283eceeae710d5d35de9d3f758f61d3e6a6e2574f2f).  We no longer have a dependency on Core shipping, correct?

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
```diff
@@ -22,22 +22,22 @@ namespace Azure.Core.Pipeline
         private static readonly ConcurrentDictionary<string, object?> ActivitySources = new();
 
         private readonly ActivityAdapter? _activityAdapter;
-        private readonly bool _suppressNestedClientScopes;
+        private readonly bool _suppressNestedClientActivities;
 
-        internal DiagnosticScope(string ns, string scopeName, DiagnosticListener source, ActivityKind kind, bool suppressNestedClientScopes = true) :
-            this(scopeName, source, null, GetActivitySource(ns, scopeName), kind, suppressNestedClientScopes)
+        internal DiagnosticScope(string ns, string scopeName, DiagnosticListener source, ActivityKind kind, bool suppressNestedClientActivities) :
```
annelo-msft, line 27: Curious why we're removing the default here?

sdk/core/Azure.Core/tests/ClientDiagnosticsTests.Net50.cs
```diff
@@ -33,7 +33,7 @@ public void StartActivityNoOpsWithoutSwitch()
         {
             using var activityListener = new TestActivitySourceListener("Azure.Clients.ClientName");
 
-            DiagnosticScopeFactory clientDiagnostics = new DiagnosticScopeFactory("Azure.Clients", "Microsoft.Azure.Core.Cool.Tests", true);
+            DiagnosticScopeFactory clientDiagnostics = new DiagnosticScopeFactory("Azure.Clients", "Microsoft.Azure.Core.Cool.Tests", true, false);
```
annelo-msft, line 36: Am I understanding this correctly?  Before the default was `true`, but now we are passing `false`.  Does/should this change the behavior of the test?

sdk/core/Azure.Core/tests/ClientDiagnosticsTests.Net50.cs
```diff
@@ -100,6 +100,7 @@ public void StartsActivitySourceActivity()
 
         [TestCase(null)]
         [TestCase(false)]
+        [TestCase(true)]
```
annelo-msft, line 103: What does `[TestCase(null)]` signify?

sdk/core/Azure.Core/src/Shared/ClientDiagnostics.cs
lmolkova: the reasoning is to be able to change the default once we're ready. So we need three choices:
1. `true`: clients explicitly opting-into suppression (new ones) 
2. `false`: clients that explicitly opt-out. E.g. HTTP RequestActivityPolicy that must never be suppressed
3. `null`: clients that follow default (current SDKs) - that only want back-compat

If we merge `null` with `false` we'll have a very hard time separating explicit choice from following a default.

sdk/core/Azure.Core/src/Shared/ClientDiagnostics.cs
lmolkova: Correct!

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
lmolkova: If we keep default here, it should be consistent with `ClientDiagnostics` and would be suppression *off*.

But the right value for all future manually written clients would be suppression *on*.

If we're ok with having inconsistent default values, I'm happy to return default `= true` back, otherwise, I prefer to force callers to make a choice

sdk/core/Azure.Core/src/Shared/DiagnosticScopeFactory.cs
lmolkova: similar to above: 

If we keep default here, it should be consistent with `ClientDiagnostics` and would be suppression *off*.

`ClientDiagnostics` needs suppression *off* because all previous autogenerated code needs to be backward compatible. 

But the right value for all future manually written clients would be suppression *on*.

If we're ok with having inconsistent default values, I'm happy to return default `= true` back, otherwise, I prefer to force callers to make a choice

sdk/core/Azure.Core/tests/ClientDiagnosticsTests.Net50.cs
lmolkova: It's an extra parameter I added, this test does not validate suppression

sdk/core/Azure.Core/tests/ClientDiagnosticsTests.Net50.cs
lmolkova: The default value, that is suppression *off*, but can change after we decide to break old clients.

sdk/core/Azure.Core/tests/ClientDiagnosticsTests.Net50.cs
lmolkova: It was testing default value of suppression flag in  `DiagnosticScopeFactory` which I removed. There is no default value anymore and we have tests for explicit values.

sdk/core/Azure.Core/src/Shared/ClientDiagnostics.cs
JoshLove-msft: Would you mind adding a comment explaining the reasoning behind why this is nullable?

JoshLove-msft (approved):

christothes (approved):

sdk/core/Azure.Core/src/Shared/ClientDiagnostics.cs
annelo-msft: Oh, that is cool. 😎

sdk/core/Azure.Core/src/Shared/ClientDiagnostics.cs
```diff
@@ -22,18 +22,39 @@ internal class ClientDiagnostics : DiagnosticScopeFactory
...
         private readonly HttpMessageSanitizer _sanitizer;
 
-        public ClientDiagnostics(ClientOptions options)
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ClientDiagnostics"/> class.
+        /// </summary>
+        /// <param name="options">The customer provided client options object.</param>
+        /// <param name="suppressNestedClientActivities">Flag controlling if <see cref="System.Diagnostics.Activity"/>
+        ///  created by this <see cref="ClientDiagnostics"/> for client method calls should be suppressed when called
+        ///  by other Azure SDK client methods.  It's recommended to set it to true for new clients; use default (null) for
```
annelo-msft, line 31: super nit.

```suggestion
        ///  by other Azure SDK client methods.  It's recommended to set it to true for new clients; use default (null)
```

annelo-msft (approved):
Thank you, @lmolkova!

PR #28497, opened 2022-05-03
AOA Conversion SDK casing and null handling

craigktreasure:
This change addresses some big issues in the Object Anchors Conversion SDK related to case sensitive file extensions and handling of null references along with several other more minor enhancements.

  - Make solution file target VS 2022 since we now include a .NET 6 target.
  - Fix ObjectAnchorsConversionClientTests to properly test what we intend to test.
    - Note: We can't currently mock parts of the client, so we can't perform positive tests with it.
  - Make asset file type comparisons case insensitive
    - This change enables the `AssetFileType` to handle file extensions with case insensitivity and enabling it to be testable by centralizing that logic in the client options.
    - I've also hardened the `AssetFileType` type to better handle null values in the case of the default value and other cases as well as to perform parameter validate to ensure we are creating values that will be useful.
  - Enabled null reference types
    - This change enables the nullable context handling to help me catch and handle more nullable references properly. This also allowed me to remove some unused and unnecessary internal setters and things.
    - This led me to an issue in the `ObjectAnchorsConversionClient` where we weren't setting the AccountDomain property at all.

### DIFF
--- a/sdk/objectanchors/Azure.MixedReality.ObjectAnchors.Conversion/Azure.MixedReality.ObjectAnchors.Conversion.sln
+++ b/sdk/objectanchors/Azure.MixedReality.ObjectAnchors.Conversion/Azure.MixedReality.ObjectAnchors.Conversion.sln
@@ -1,7 +1,7 @@
-﻿
+
 Microsoft Visual Studio Solution File, Format Version 12.00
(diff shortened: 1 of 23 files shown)


craigktreasure:
Abandoning for a more targeted fix for now.

PR #28501, opened 2022-05-03
Asset file type case and null handling

craigktreasure:
- Make asset file type comparisons case insensitive
    - This change enables the `AssetFileType` to handle file extensions with case insensitivity.
    - I've also hardened the `AssetFileType` type to better handle `null` values in the case of the default value and other cases as well as to perform parameter validate to ensure we are creating values that will be useful.

### DIFF
--- a/sdk/objectanchors/Azure.MixedReality.ObjectAnchors.Conversion/src/AssetFileType.cs
+++ b/sdk/objectanchors/Azure.MixedReality.ObjectAnchors.Conversion/src/AssetFileType.cs
@@ -4,6 +4,7 @@
 using System;
 using System.ComponentModel;
 using System.IO;
+using Azure.Core;
 
 namespace Azure.MixedReality.ObjectAnchors.Conversion
 {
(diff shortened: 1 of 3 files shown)


craigktreasure:
@RamonArguelles

sdk/objectanchors/Azure.MixedReality.ObjectAnchors.Conversion/src/AssetFileType.cs
```diff
@@ -19,13 +20,23 @@ namespace Azure.MixedReality.ObjectAnchors.Conversion
...
-            _value = value ?? throw new ArgumentNullException(nameof(value));
+            Argument.AssertNotNullOrWhiteSpace(value, nameof(value));
+
+            if (!value.StartsWith(".", StringComparison.Ordinal))
+            {
+                throw new ArgumentException("The value must start with a '.' character.", nameof(value));
+            }
+
+            _value = value.ToLowerInvariant()
+                ?? throw new ArgumentNullException(nameof(value));
```
RamonArguelles, line 39: How can toLowerInvariant return null? Otherwise, why the extra null check here?

RamonArguelles (approved):

craigktreasure:
#sign-off

sdk/objectanchors/Azure.MixedReality.ObjectAnchors.Conversion/src/AssetFileType.cs
craigktreasure: It can't. Sorry. I moved the validation up above and forgot to "complete" the move.

jsquire (approved):

jsquire:
Automerge seems to be stuck; merging manually.

PR #28508, opened 2022-05-03
Event Hubs: Buffered Producer Improvements

jsquire:
# Summary 

The focus of these changes is to enhance the buffered producer with some minor improvements based on stress and performance test findings.

- When the service is throttling, logic has been added to allow for retrying with an additional back-off until the operation completes or has been cancelled. Because callers do not directly influence when publishing takes place, they should not be exposed to failures for throttling; doing so would require that they take extreme steps, such as manually publishing the throttled events or recreating their producer to reset buffers if the ordering of events needed to be restored.

- When events are not available in the buffer, an expanding back-off is now applied to the publishing loop.  This is intended to avoid wasting resources in a spin-wait scenario during periods of low activity.  In the current implementation, consecutive empty buffer checks will result in a delay pattern that looks like:

    - 25 ms
    - 35 ms
    - 48 ms
    - 68 ms
    - 96 ms
    - 134 ms
    - 188 ms
    - 263 ms
    - 368 ms
    - 516 ms
    - 723 ms
    - 1 second

  Once the back-off reaches one second, it will continue to use that value until an event is available in the buffer.

- The approach to setting buffered producer-specific retry defaults is now more robust and will be applied when options are passed rather than only in the case when they were not.  Customization of retry values will continue to be appropriately respected.

### DIFF
--- a/sdk/eventhub/Azure.Messaging.EventHubs/src/Producer/EventHubBufferedProducerClient.cs
+++ b/sdk/eventhub/Azure.Messaging.EventHubs/src/Producer/EventHubBufferedProducerClient.cs
@@ -58,6 +58,15 @@ namespace Azure.Messaging.EventHubs.Producer
         /// <summary>The maximum amount of time, in milliseconds, to allow for acquiring the semaphore guarding a partition's publishing eligibility.</summary>
         private const int PartitionPublishingGuardAcquireLimitMilliseconds = 100;
 
+        /// <summary>The maximum number of seconds to delay between checks when no events are available in the buffers for any partition.</summary>
+        private const double MaximumEmptyBufferDelaySeconds = 1;
+
+        /// <summary>The number of seconds to use as the starting delay between checks when no events are available in the buffers for any partition.</summary>
+        private static readonly double StartingEmptyBufferDelaySeconds = TimeSpan.FromMilliseconds(25).TotalSeconds;
+
+        /// <summary>The base interval to delay when publishing is throttled and an operation needs to back-off before retrying.  Four seconds is recommended by the service.</summary>
+        private static readonly TimeSpan ThrottleBackoffInterval = TimeSpan.FromSeconds(4);
+
         /// <summary>The minimum interval to allow for waiting when building a batch to publish.</summary>
         private static readonly TimeSpan MinimumPublishingWaitInterval = TimeSpan.FromMilliseconds(5);
 
@@ -65,11 +74,13 @@ namespace Azure.Messaging.EventHubs.Producer
         private static readonly TimeSpan DefaultPublishingDelayInterval = TimeSpan.FromMilliseconds(25);
 
         /// <summary>The set of client options to use when options were not passed when the producer was instantiated.</summary>
-        private static readonly EventHubBufferedProducerClientOptions DefaultOptions =
-            new EventHubBufferedProducerClientOptions
-            {
-                RetryOptions = new EventHubsRetryOptions { MaximumRetries = 15, TryTimeout = TimeSpan.FromMinutes(3) }
-            };
+        private static readonly EventHubBufferedProducerClientOptions DefaultOptions = new();
+
+        /// <summary>The random number generator to use for a specific thread.</summary>
+        private static readonly ThreadLocal<Random> RandomNumberGenerator = new ThreadLocal<Random>(() => new Random(Interlocked.Increment(ref s_randomSeed)), false);
+
+        /// <summary>The seed to use for initializing random number generated for a given thread-specific instance.</summary>
+        private static int s_randomSeed = Environment.TickCount;
 
         /// <summary>The set of currently active partition publishing tasks.  Partition identifiers are used as keys.</summary>
         private readonly ConcurrentDictionary<string, PartitionPublishingState> _activePartitionStateMap = new();
@@ -1377,7 +1388,7 @@ namespace Azure.Messaging.EventHubs.Producer
 
                 var totalWaitTime = _options.MaximumWaitTime ?? Timeout.InfiniteTimeSpan;
                 var remainingWaitTime = totalWaitTime;
-                var delayInterval = CalculateDelay(totalWaitTime, DefaultPublishingDelayInterval);
+                var delayInterval = CalculateBatchingDelay(totalWaitTime, DefaultPublishingDelayInterval);
 
                 // The wait time constraint should not consider creating the batch; start tracking after the batch is available to build.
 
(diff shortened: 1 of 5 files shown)


sdk/eventhub/Azure.Messaging.EventHubs/src/Producer/EventHubBufferedProducerClient.cs
```diff
@@ -2314,16 +2380,36 @@ private async Task UpdatePartitionInformation(CancellationToken cancellationToke
...
+        /// <param name="maximumBufferDelaySeconds">The maximum number of seconds to allow as a delay.</param>
+        ///
+        /// <returns>The amount of delay to apply before checking the buffers for available events.</returns>
+        ///
+        private static TimeSpan CalculateEmptyBufferDelay(int consecutiveEmptyBufferCount,
+                                                          double startingEmptyBufferDelaySeconds,
+                                                          double maximumBufferDelaySeconds)
+        {
+            var delay = (Math.Pow(1.4, consecutiveEmptyBufferCount) * startingEmptyBufferDelaySeconds);
+            return delay > maximumBufferDelaySeconds ? TimeSpan.FromSeconds(1) : TimeSpan.FromSeconds(delay);
```
jsquire, line 2411:
```suggestion
            return delay > maximumBufferDelaySeconds ? maximumBufferDelaySeconds  : TimeSpan.FromSeconds(delay);
```

jsquire: Will fix this after the test run completes.

sdk/eventhub/Azure.Messaging.EventHubs/src/Producer/EventHubBufferedProducerClient.cs
```diff
@@ -2314,16 +2380,36 @@ private async Task UpdatePartitionInformation(CancellationToken cancellationToke
...
+        /// <param name="startingEmptyBufferDelaySeconds">The number of seconds to use as the starting point for the delay.</param>
+        /// <param name="maximumBufferDelaySeconds">The maximum number of seconds to allow as a delay.</param>
+        ///
+        /// <returns>The amount of delay to apply before checking the buffers for available events.</returns>
+        ///
+        private static TimeSpan CalculateEmptyBufferDelay(int consecutiveEmptyBufferCount,
+                                                          double startingEmptyBufferDelaySeconds,
+                                                          double maximumBufferDelaySeconds)
+        {
+            var delay = (Math.Pow(1.4, consecutiveEmptyBufferCount) * startingEmptyBufferDelaySeconds);
```
christothes, line 2410: Just curious why we picked `Math.Pow(1.4, consecutiveEmptyBufferCount)` as opposed to something more straightforward like `consecutiveEmptyBufferCount * 2`.

sdk/eventhub/Azure.Messaging.EventHubs/src/Producer/EventHubBufferedProducerClient.cs
```diff
@@ -2145,14 +2204,21 @@ private async Task RunProducerManagementAsync(CancellationToken cancellationToke
                                 // Responsibility for releasing the guard semaphore is passed to the task.
 
                                 activeTasks.Add(PublishBatchToPartition(partitionState, releaseGuard: true, activeOperationCancellationSource.Token));
+                                consecutiveEmptyBufferCount = 0;
```
christothes, line 2207: perhaps I'm missing something, but where do we loop and allow this value to increment past 1?

sdk/eventhub/Azure.Messaging.EventHubs/src/Producer/EventHubBufferedProducerClientOptions.cs
```diff
@@ -32,7 +32,7 @@ public class EventHubBufferedProducerClientOptions
         private EventHubConnectionOptions _connectionOptions = new EventHubConnectionOptions();
 
         /// <summary>The set of options to govern retry behavior and try timeouts.</summary>
-        private EventHubsRetryOptions _retryOptions = new EventHubsRetryOptions();
+        private EventHubsRetryOptions _retryOptions = new EventHubBufferedProducerClientRetryOptions();
```
christothes, line 35: Just curious - why not new up an `EventHubsRetyOptions` with the Properties set rather than add a new type here?

jsquire: Um.... yeah.   For some reason, I did not think of this at all.   Wow, now I feel silly.   I can just set these in the constructor.

sdk/eventhub/Azure.Messaging.EventHubs/src/Producer/EventHubBufferedProducerClient.cs
jsquire: It's on L2217, each time the delay calculates it increments.

christothes: I see that increment from 0 to 1, but what is the path to get to 2? From what I can see the `while` loop is above this on L2157, so it seems like we need to execute this line each iteration.

jsquire: Don't we?   Each tick of that loop, if we have no active publishing tasks, we'll hit this block and evaluate if we have nothing active because there are no events or because we haven't gotten to their partition yet.    When the answer is "no events", we'll continue to hit the delay calculate and increment.    

That gets reset when we start a publishing task, and we'll short-circuit if there are events waiting but we haven't hit that partition yet.

Am I overlooking something?

m-redding: I might be a little confused- but maybe he's saying that it looks like the value gets reset to 0 every iteration? Like it should only be set to 0 when there are active tasks again?

christothes: @m-redding that is exactly what I was saying :)

christothes: chatted with @jsquire offline - I was missing that the if block on 2199 prevents the reset when the buffer is still empty

jsquire: Chris and I discussed offline; the logic is correct due to the conditional check for resetting only when events are present.   Chris also suggested potentially using a sync primitive after the initial back-off to avoid spinning, using `EnqueueEventsAsync` as the trigger to start processing again.   I'm going to open an issue to track that for an improvement in the next milestone.

christothes (approved):

sdk/eventhub/Azure.Messaging.EventHubs/src/Producer/EventHubBufferedProducerClient.cs
m-redding: Ok great, that makes sense! @christothes @jsquire

andreminelli:
Hello.
Sorry if this is a dumb question, but I couldn't find a way to confirm: is this improvement already released alongside version 5.7.0?

I have a scenario where, after had already sent hundreds of messages using `EventHubBufferedProducerClient`, CPU usage continues as high as during the sending process :(

If version 5.7.0 hasn't this code yet, can you please provide an estimate about when a new version with this enhancement is expected to be released?

jsquire:
Hi @andreminelli.   Yes, this was part of the v5.7.0 release.   If you're seeing unexpectedly high CPU with that package, we'd appreciate if you'd open an issue so that we can investigate.

PR #28510, opened 2022-05-03
Fixing CodeGen Capitalization

jsquire:
# Summary

The focus of these changes is to fix a typo in the `CustomEntitiesResultDocumentsItem` stub which was causing "item" to be generated as lower-case.  This differs from the capitalization of the file committed to the repository and was causing issues with case-sensitive file systems in CI when the "Analyze" step was run during API diff checks.


maririos (commented):
looks like you need to regenerate the code so you can get the other classes updated, like for example: https://github.com/Azure/azure-sdk-for-net/blob/feature/text-analytics-language-api/sdk/textanalytics/Azure.AI.TextAnalytics/src/Generated/Models/CustomEntitiesResultDocumentsItem.cs#L15

PR #28514, opened 2022-05-03
Azure Extensions: GUID Parameter Recognition

jsquire:
# Summary

The focus of these changes is to add recognition for GUID parameter types in client constructors when creating from configuration.  This is intended to unblock scenarios needed by the AnomalyDetector client library.

### DIFF
--- a/sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientFactory.cs
+++ b/sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientFactory.cs
@@ -319,6 +319,11 @@ namespace Microsoft.Extensions.Azure
                 return TryConvertFromString(configuration, parameterName, s => new Uri(s), out value);
             }
 
+            if (parameterType == typeof(Guid))
+            {
+                return TryConvertFromString(configuration, parameterName, s => Guid.Parse(s), out value);
+            }
+
             return TryCreateObject(parameterType, configuration.GetSection(parameterName), out value);
         }
 
--- a/sdk/extensions/Microsoft.Extensions.Azure/tests/ClientFactoryTests.cs
+++ b/sdk/extensions/Microsoft.Extensions.Azure/tests/ClientFactoryTests.cs
@@ -52,6 +52,24 @@ namespace Azure.Core.Extensions.Tests
         }
 
         [Test]
+        public void FailsToConvertInvalidUriConfiguration()
+        {
+            IConfiguration configuration = GetConfiguration(new KeyValuePair<string, string>("uri", "no it its not"));
+
+            var clientOptions = new TestClientOptions();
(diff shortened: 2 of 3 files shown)


sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientFactory.cs
```diff
@@ -319,6 +319,11 @@ private static bool TryConvertArgument(IConfiguration configuration, string para
                 return TryConvertFromString(configuration, parameterName, s => new Uri(s), out value);
             }
 
+            if (parameterType == typeof(Guid))
```
JoshLove-msft, line 322: I guess this is already an existing issue with the Uri parsing above, but should we consider using Guid.TryParse or wrapping the parse in a try/catch so we can continue evaluating other ctors if the configuration does not contain a GUID?

JoshLove-msft, line 322: I guess this is already an existing issue with the Uri parsing above, but should we consider using Guid.TryParse or wrapping the parse in a try/catch so we can continue evaluating other ctors if the configuration does not contain a GUID?

jsquire: Honestly, I don't know.   I considered it, but since I don't have great context around why things are structured this way, my approach was to mimic the existing surface.   For some reason, `Uri.TryCreate` was not used.   Because of that, I did not `TryParse`

JoshLove-msft: I think it is probably something that should have been done for URI as well, but it may be even more likely to cause issues with GUID.
Imagine if a service exposes two constructors one containing a GUID, and another constructor using a string for the same parameter name. This would not be a realistic scenario with URI because we have guidelines that require using the URI type.

JoshLove-msft (approved):

sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientFactory.cs
JoshLove-msft: Thinking about this more, I think if we were to hit exceptions for URI/GUID parsing, that means that user has entered bad data, and the exception that is propagated to them seems reasonable. There isn't really a realistic scenario where a client would have multiple constructors with identically named parameters having different types. And if there is such a client, then we may still get things wrong as it would just depend on the order that the constructors were evaluated.

jsquire: Discussed with Josh offline; we decided that the conversion exceptions are likely to be more helpful than the message from a `false` return from `TryConvert` and likely more discoverable than if we used a try/catch with custom error, pushing the actual failure to the inner exception.   We agreed to keep the implementation as-is for now and revisit if we see customer feedback.

PR #28518, opened 2022-05-03
Log more information when MHSM provisioning fails

heaths:


sdk/keyvault/test-resources-post.ps1
```diff
@@ -95,7 +95,13 @@ if (Test-Path $sdpath) {
     Remove-Item $sdPath -Force
 }
 
-Export-AzKeyVaultSecurityDomain -Name $hsmName -Quorum 2 -Certificates $wrappingFiles -OutputPath $sdPath
+Export-AzKeyVaultSecurityDomain -Name $hsmName -Quorum 2 -Certificates $wrappingFiles -OutputPath $sdPath -ErrorAction SilentlyContinue -Verbose
+if ( !$? ) {
+    Write-Host $Error[0].Exception
+    Write-Error $Error[0]
+
+    exit
```
benbp, line 103: I assume this will still fail because of stderr for devops, right?

benbp (approved):

sdk/keyvault/test-resources-post.ps1
heaths: Not sure. Didn't see a reason to take the risk. It's temporary anyway.

heaths:
Appears to be a transient service issue, but the deployment passed for Managed HSM which is what we were trying to reproduce so I'm going to force this in. We'll leave the logging code in for now in case the issue with security domain download the last couple nights happens again.

PR #28539, opened 2022-05-04
Additional informational logs for transmission

vishweshbankwar:


sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/AzureMonitorTransmitter.cs
```diff
@@ -156,10 +161,13 @@ private ExportResult HandleFailures(HttpMessage httpMessage)
                 // HttpRequestException
                 content = HttpPipelineHelper.GetRequestContent(httpMessage.Request.Content);
                 result = _storage.SaveTelemetry(content, HttpPipelineHelper.MinimumRetryInterval);
+                AzureMonitorExporterEventSource.Log.Write($"FailedToTransmit{EventLevelSuffix.Informational}", "Network Error");
```
rajkumar-rangaraj, line 164: Is this informational or Warning?

vishweshbankwar: We are saving telemetry so thought of it as informational message. But warning also makes sense. Should we change it in case of error response code as well?

rajkumar-rangaraj: Yes. Let us change it to warning in case of error response too.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/AzureMonitorTransmitter.cs
```diff
@@ -203,25 +202,34 @@ private ExportResult HandleFailures(HttpMessage httpMessage)
                 }
             }
 
+            if (result == ExportResult.Success)
+            {
+                AzureMonitorExporterEventSource.Log.Write($"FailedToTransmit{EventLevelSuffix.Warning}", $"Error code is {statusCode}: Telemetry is stored offline for retry");
```
rajkumar-rangaraj, line 207: We might see a message `Error code is 0`, when `_storage.SaveTelemetry(content, HttpPipelineHelper.MinimumRetryInterval);` returns success.

rajkumar-rangaraj (approved):

PR #28544, opened 2022-05-04
Remove references to TextAnalyticsErrorInternal from Azure.AI.TextAnalytics

nisha-bhatia:

### DIFF
--- a/sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.cs
+++ b/sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.cs
@@ -12,16 +12,6 @@ namespace Azure.AI.TextAnalytics
 {
     internal static partial class Transforms
     {
-        #region Needs Review
-
-        //
-        // REGION: Common
-        //
-        internal static TextAnalyticsError ConvertToError(TextAnalyticsErrorInternal error) => throw new NotImplementedException("I think the TextAnalyticsInternalType is no longer used.");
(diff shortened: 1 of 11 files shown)


sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.cs
```diff
@@ -35,6 +25,17 @@ internal static TextAnalyticsError ConvertToError(Error error)
                 : new TextAnalyticsError(error.Code.ToString(), error.Message, error.Target);
         }
 
+        internal static List<TextAnalyticsError> ConvertToErrors(IReadOnlyList<Error> internalErrors)
```
maririos, line 28: who is using this function?

maririos:
For now, the changes look fine. Even if `TextAnalyticsErrorInternal` is still in the LRO code, this is an improvement and we could consider just merging this and later on keep adjusting the code

sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.cs
nisha-bhatia: It is currently not being used - I wrote it because we used to have one for TextAnalyticsErrorInternal on line 21. Maybe it is needed in the LROs?

maririos: Hmm maybe. ok later on we can delete it if we realize no one used it

maririos (approved):

PR #28547, opened 2022-05-05
Resolve breaking changes in #28525

ArcturusZhang:


ArthurMa1978 (approved):

m-nash (changes requested):
The issue I see with this approach is we found these issues due to a diff.  In new generations we won't have this luxury and its unclear how we will know where to apply these config changes to work around the unexpected behavior in autorest core and modelerfour

PR #28574, opened 2022-05-06
[Azure.Communication.Common] Added support for rawId ⟷ CommunicationIdentifier conversion

petrsvihlik:
### Packages impacted by this PR
- Azure.Communication.Common

### Issues associated with this PR
- N/A

### Describe the problem that is addressed by this PR
The `CommunicationIdentifier` design provides a good abstraction of Azure Communication Services internal id format with better type safety, auto-complete and hides internal knowledge.

However, it doesn't lend well to storing identifiers in a database or using them as keys because there is no clear canonical way how to encode them. This PR introduces translation functions that lets developers use the underlying `rawId` for these purposes.

Details [Internal wiki](https://skype.visualstudio.com/SPOOL/_wiki/wikis/SPOOL.wiki/31075/Design-change-request-RawId-support-for-(de)serialization)

### What are the possible designs available to address the problem? If there are more than one possible design, why was the one in this PR chosen?

See the [Internal wiki](https://skype.visualstudio.com/SPOOL/_wiki/wikis/SPOOL.wiki/31075/Design-change-request-RawId-support-for-(de)serialization)

### Are there test cases added in this PR?
- [x] Yes.

### Provide a list of related PRs _(if any)_
- https://github.com/Azure/azure-sdk-for-js/pull/21699
- https://github.com/Azure/azure-sdk-for-ios/pull/1255

### Checklists
- [x] Added impacted package name to the issue description
- [x] Does this PR needs any fixes in the SDK Generator? **NO**
- [x] Added a changelog


sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
```diff
@@ -3,20 +3,61 @@
...
     /// <summary>Represents an identifier in Azure Communication Services.</summary>
-#pragma warning disable CS0659 // Type overrides Object.Equals(object o) but does not override Object.GetHashCode()
     public abstract class CommunicationIdentifier : IEquatable<CommunicationIdentifier>
-#pragma warning restore CS0659 // Type overrides Object.Equals(object o) but does not override Object.GetHashCode()
     {
+        /// <summary>
+        /// Returns the RawId representation of the CommunicationIdentifier.
+        /// You can use the RawId for encoding the identifier and then use it as a key in a database.
+        /// </summary>
+        public virtual string RawId { get; }
```
petrsvihlik, line 17: @DominikMe `abstract` would be a breaking change. you ok with this?

DominikMe: Technically it would be breaking because we hadn't marked CommunicationIdentfier as `sealed`. But I don't expect that any customer has sub-classed CommunicationIdentifier as we wouldn't know how to treat this sub-class. Maybe for testing, but customers would have probably used a mocking technique.
`abstract` is more correct, as we need each identifier to define how they encode themselves, so that is my preference.

@christothes or @tg-msft can you advise whether we can make an exception here and use `abstract`?

sdk/communication/Azure.Communication.Common/src/PhoneNumberIdentifier.cs
```diff
@@ -27,17 +35,17 @@ public PhoneNumberIdentifier(string phoneNumber, string rawId = null)
...
         public override string ToString() => PhoneNumber;
 
         /// <inheritdoc />
-        public override int GetHashCode() => PhoneNumber.GetHashCode();
+        public override int GetHashCode() => RawId.GetHashCode();
 
         /// <inheritdoc />
         public override bool Equals(CommunicationIdentifier other)
-            => other is PhoneNumberIdentifier otherId && otherId.PhoneNumber == PhoneNumber && (RawId is null || otherId.RawId is null || RawId == otherId.RawId);
+            => other is PhoneNumberIdentifier otherId && otherId.RawId == RawId;
```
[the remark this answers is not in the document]
petrsvihlik: @JoshuaLai raised [the same question](https://github.com/Azure/azure-sdk-for-ios/pull/1255/files/d3105c1e1efef9313aa05da135ba23df13f9b575#r866346831) in the iOS library. 

My implementation was based on @DominikMe's change request, specifically, this comment:
> Change Equals and GetHashCode implementations for each identifier to rely on RawId only and not on other properties.

@DominikMe shall we revert to the previous logic, considering @tjprescott's [comment](https://github.com/Azure/azure-sdk-for-ios/pull/1255/files/d3105c1e1efef9313aa05da135ba23df13f9b575#r868190665)?

sdk/communication/Azure.Communication.Common/src/PhoneNumberIdentifier.cs
```diff
@@ -8,8 +8,16 @@ namespace Azure.Communication
     /// <summary>Represents a Phone Number.</summary>
     public class PhoneNumberIdentifier : CommunicationIdentifier
     {
+        private readonly string _rawId;
+
         /// <summary>The optional raw id of the phone number.</summary>
```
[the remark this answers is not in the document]
petrsvihlik: fixed

sdk/communication/Azure.Communication.Common/src/PhoneNumberIdentifier.cs
```diff
@@ -27,17 +35,17 @@ public PhoneNumberIdentifier(string phoneNumber, string rawId = null)
         {
             Argument.AssertNotNullOrEmpty(phoneNumber, nameof(phoneNumber));
             PhoneNumber = phoneNumber;
-            RawId = rawId;
+            _rawId = rawId;
```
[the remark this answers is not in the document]
petrsvihlik: I prefer to have as little logic in the constructor for several reasons. E.g., not to waste processor time on things that might not be needed at the time of construction. In this case, it can be negligible but for the `MicrosoftTeamsUserIdentifier` it's already four if-elses. I changed the implementation though to support lazy initialization of the properties so that they don't need to be recalculated every time (https://github.com/Azure/azure-sdk-for-net/pull/28574/commits/9bd080b83af940120e1989eb872ec8f7c4f9afe7).

Please let me know, if it makes sense.

If this is not idiomatic for Java, feel free to do the stuff in the constructor.

sdk/communication/Azure.Communication.Common/src/PhoneNumberIdentifier.cs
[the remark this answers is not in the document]
DominikMe: @petrsvihlik I've commented on the other thread. In the case of a rawId override, the real id is the rawId and the phoneNumber doesn't need to match

sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
tg-msft: No, either changing the existing modifier or adding a new abstract member we forwarded this to would be an unacceptable breaking change per the Framework Design Guidelines (see D.11.21 and D.7.2).  I think we'll have to live with the design and just document the expectations.

DominikMe: :-( That's unfortunate, I cannot imagine that we would break anybody in reality. This class was never designed to be sub-classed by others.

If we have to live with it for strict reasons, we should maybe throw in `CommunicationIdentifier`'s implementation to discover bugs early where a sub-class doesn't override `RawId` as it's supposed to.

petrsvihlik: >we should maybe throw in CommunicationIdentifier's implementation to discover bugs early where a sub-class doesn't override RawId as it's supposed to.

yeah, but if we base `GetHashCode()` on it, then it's effectively the same as creating a new abstract member (the derived class is forced to implement it). Is the fact that the class was never designed to be sub-classed by others a good enough justification to do so? @tg-msft @DominikMe

sdk/communication/Azure.Communication.Common/src/PhoneNumberIdentifier.cs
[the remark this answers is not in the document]
petrsvihlik: ok, thanks. Then I'm resolving this thread.

sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
petrsvihlik: yes, we can definitely bring up the topic on Thursday! thanks for your comments so far 🙂 👍

petrsvihlik: @DominikMe 
We discussed it on Thursday (recording [here](https://microsoft-my.sharepoint.com/:v:/p/teglaza/EUmWLrxWxQVEqUsJs4_eceYB7YQ1HqCMPeQsqf9xOA2pxA) - it was the first topic of the session) and there were two suggestions for how to tackle this issue, both of which assume adding default implementation:

1. return a constant `UnknownIdentifier` which would equal to an identifier that would not collide with any other identifier (like `0:unknown`)
2. return `null`


The first solution is better in the case we want to always have some value available that we can insert into the database. If not, we should go with the second approach - returning null. Since we're doing this also to support inserting the IDs to a DB, I'm slightly in favor of the first way. How about you @DominikMe ?

As for making sure that our subclasses implement `RawId`, I suggest adding an automatic test that would check that via reflection.

Also, we need to update the docs, to instruct Contoso to implement this member if they have a derived class.

sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
```diff
@@ -3,20 +3,61 @@
...
+            return prefix switch
+            {
+                "8:teamsvisitor:" => new MicrosoftTeamsUserIdentifier(suffix, true),
+                "8:orgid:" => new MicrosoftTeamsUserIdentifier(suffix, false, CommunicationCloudEnvironment.Public),
+                "8:dod:" => new MicrosoftTeamsUserIdentifier(suffix, false, CommunicationCloudEnvironment.Dod),
+                "8:gcch:" => new MicrosoftTeamsUserIdentifier(suffix, false, CommunicationCloudEnvironment.Gcch),
+                "8:acs:" or "8:spool:" or "8:dod-acs:" or "8:gcch-acs:" => new CommunicationUserIdentifier(rawId),
+                _ => new UnknownIdentifier(rawId),
+            };
+        }
```
iaulakh, line 61: I think we can create helper class, the future changes to this will help not to update this CommunicationIdentifier.

petrsvihlik: Why don't you like updating the `CommunicationIdentifier` class? 
I'd prefer to keeping it this way instead of creating a helper class. Currently, it's nicely aligned with having the serialization logic in the specific implementations of the `CommunicationIdentifier.RawId`.

iaulakh: Sounds good @petrsvihlik

sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
DominikMe: To me, nobody should ever use the return value of the base class. If a sub-class doesn't implement this "virtual" method it's a bug. Having a string like `0:unknown` imo just hides the error. If this string gets sent to our services, it will just fail there. If the string gets persisted in the customer's db it won't be better than `null` because it's meaningless.

> Also, we need to update the docs, to instruct Contoso to implement this member if they have a derived class.
`CommunicationIdentifier` really isn't meant to be subclassed by Contoso and we shouldn't encourage them. The identifier is a 1:1 mapping to values that are entirely defined in the IC3 services.

petrsvihlik: keeping null + adding a test for subclasses implementing the member

sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
```diff
@@ -3,20 +3,61 @@
...
+            var segments = rawId.Split(':');
+            if (segments.Length < 3)
+                return new UnknownIdentifier(rawId);
+
+            var prefix = $"{segments[0]}:{segments[1]}:";
+            var suffix = rawId.Substring(prefix.Length);
+
+            return prefix switch
+            {
+                "8:teamsvisitor:" => new MicrosoftTeamsUserIdentifier(suffix, true),
```
AikoBB, line 54: maybe we should create const strings for these prefixes in this class and reuse them in [MicrosoftTeamsUserIdentifier.RawId](https://github.com/Azure/azure-sdk-for-net/blob/c68aa4ab661798d829ea183f4eb94fba2336d36d/sdk/communication/Azure.Communication.Common/src/MicrosoftTeamsUserIdentifier.cs#L17)?

petrsvihlik: yes good point. let's keep it consistent with other langs for now and reopen this when we'll get to adding the cloud property. we'll track it here: https://skype.visualstudio.com/SPOOL/_workitems/edit/2835378

AikoBB (approved):

PR #28595, opened 2022-05-07
support relative sdk path

chunyu3:

- support relative path
- update doc to indicate to use absolute path for local swagger file
- add AzureKeyCredentialPolicy shared class in project file


doc/DataPlaneCodeGeneration/AzureSDKCodeGeneration_DataPlane_Quickstart.md
```diff
@@ -80,7 +80,7 @@ pwsh /home/azure-sdk-for-net/eng/scripts/automation/Invoke-DataPlaneGenerateSDKP
...
+- `-inputfiles` takes the address of the Open API spec files,  separated by semicolon if there is more than one file.  The Open API spec file can be local file, or the web address of the file in the `azure-rest-api-specs` repo.  When pointing to a local file, make sure to use **absolute path**, i.e. /home/swagger/compute.json. When pointing to a file in the `azure-rest-api-specs` repo, make sure to include the commit id in the URI, i.e. `https://github.com/Azure/azure-rest-api-specs/blob/73a0fa453a93bdbe8885f87b9e4e9fef4f0452d0/specification/webpubsub/data-plane/WebPubSub/stable/2021-10-01/webpubsub.json`. This ensures that you can choose the time to upgrade to new swagger file versions.
```
archerzz, line 83: If we support [Resolve-Path](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.management/resolve-path?msclkid=fb39864acf3811ec9d01f492e29b7525&view=powershell-7.2), then we should support relative path, right? Then here we don't need to emphasize `absolute path`, I guess.

heaths (approved):

doc/DataPlaneCodeGeneration/AzureSDKCodeGeneration_DataPlane_Quickstart.md
chunyu3: If you want to refer to an local swagger files as input-file, you need to add absolute path, because it will be inserted into autorest.md and will be used late, so it will be absolute path, otherwise, when it use `dotnet build /t:GenerateCode` it will fail. So I emphasize `absolute path` to ask them to provide absolute path. 
Yes, we can use Resolve-Path to translate relative path to absolute path, but since the -inputfiles are list of files which may be local file or web address, we need to write script to support this. Current we have not support yet, so ask them to provide absolute path.

chunyu3: We may support in future

PR #28619, opened 2022-05-10
Add provenance for PemReader

### LINKED ISSUE #28613: [FEATURE REQ] Add some doc comments to PemReader to explain some background as to why it exists
### Library name

Azure.Core

n/a

heaths:
Fixes #28613

### DIFF
--- a/sdk/core/Azure.Core/src/Shared/PemReader.cs
+++ b/sdk/core/Azure.Core/src/Shared/PemReader.cs
@@ -15,11 +15,11 @@ namespace Azure.Core
     /// <remarks>
     /// This class provides a downlevel PEM decoder since <c>PemEncoding</c> wasn't added until net5.0.
     /// The <c>PemEncoding</c> class takes advantage of other implementation changes in net5.0 and,
-    /// based on conversations with the .NET team, runtime changes. This was based on <c>PemEncodiner</c>
-    /// and reviewed by @bartonjs on the .NET / cryptography team.
+    /// based on conversations with the .NET team, runtime changes.
     /// </remarks>
     internal static partial class PemReader
     {
+        // The following implementation was based on PemEncoding and reviewed by @bartonjs on the .NET / cryptography team.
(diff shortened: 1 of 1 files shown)


sdk/core/Azure.Core/src/Shared/PemReader.cs
```diff
@@ -12,6 +12,12 @@ namespace Azure.Core
     /// <summary>
     /// Reads PEM streams to parse PEM fields or load certificates.
     /// </summary>
+    /// <remarks>
+    /// This class provides a downlevel PEM decoder since <c>PemEncoding</c> wasn't added until net5.0.
+    /// The <c>PemEncoding</c> class takes advantage of other implementation changes in net5.0 and,
+    /// based on conversations with the .NET team, runtime changes. This was based on <c>PemEncodiner</c>
```
JoshLove-msft, line 18:
```suggestion
    /// based on conversations with the .NET team, runtime changes. This was based on <c>PemEncoding</c>
```

sdk/core/Azure.Core/src/Shared/PemReader.cs
```diff
@@ -12,6 +12,12 @@ namespace Azure.Core
...
+    /// and reviewed by @bartonjs on the .NET / cryptography team.
```
JoshLove-msft, line 19: Should we be putting GitHub aliases in ref docs?

JoshLove-msft (approved):

sdk/core/Azure.Core/src/Shared/PemReader.cs
heaths: Perhaps not in rendered doc comments. As just a code comment sure. I think it's important to bring up here as bartonjs is the .NET security expert worth mentioning in this regard. Maybe I could refactor that a bit.

sdk/core/Azure.Core/src/Shared/PemReader.cs
heaths: Ugh. I kept wanting to type PemEncoder even though I knew it was PemEncoding. Guess I got my wires crossed here.

Issue #28640, opened 2022-05-10
[Core] Allow configuration of ClientCertificates for the HttpPipelineTransport


christothes:
### API Change Needed

Yes

### Background and motivation

For clients that require simple client certificate authentication, there is currently no supported option to configure the pipeline this way. Clients such as Confidential Ledger, support this mode of authentication. 

**Without feature support for this, customers would be required to provide a custom Transport, which would look something like the following**:

```C#
var cert = new X509Certificate2("MyCert.pfx", "password");
var options = new ConfidentialLedgerClientOptions();
var handler = new HttpClientHandler();
handler.ClientCertificates.Add(cert);
// The ServerCertificateCustomValidationCallback  is simplified from what is required here
handler.ServerCertificateCustomValidationCallback += (_, _, _, _) => true;
var http = new HttpClient(handler);
var transport = new HttpClientTransport(http);
options.Transport = transport;
var client = new ConfidentialLedgerClient(new Uri("https://chrisscltest.confidential-ledger.azure.com"), options);
```

Note this also prevents the client from taking advantage of existing transport defaults that are normally configured.

**With this feature, client construction looks a follows**:

```c#
var cert = new X509Certificate2("MyCert.pfx", "password");
var client = new ConfidentialLedgerClient(new Uri("https://chrisscltest.confidential-ledger.azure.com"), cert);
```

The client internally does the following with the new CertificateCredential

```C#
transportOptions.ClientCertificates.Add(certificateCredential);
ClientDiagnostics = new ClientDiagnostics(actualOptions);
_tokenCredential = credential;
_pipeline = HttpPipelineBuilder.Build(
    actualOptions,
    Array.Empty<HttpPipelinePolicy>(),
    Array.Empty<HttpPipelinePolicy>(),
    transportOptions,
    new ResponseClassifier());
```

### API / Feature Proposal

```diff
namespace Azure.Core
{
    public partial class HttpPipelineTransportOptions
    {
        public HttpPipelineTransportOptions() { }
+       public System.Collections.Generic.IList<X509Certificate2> ClientCertificates { get { throw null; } }
        public System.Func<Azure.Core.Pipeline.ServerCertificateCustomValidationArgs, bool>? ServerCertificateCustomValidationCallback { get { throw null; } set { } }
    }
}
```

### Time Constraints and Dependencies

This is needed for Confidential Ledger GA

JoshLove-msft:
[...]

PR #28645, opened 2022-05-10
Client Certificate auth for Confidential Ledger

christothes:
#28640

docs changes in https://github.com/Azure/azure-sdk-tools/pull/3320


eng/common/TestResources/README.md
```diff
@@ -77,6 +77,48 @@ setx KEYVAULT_SKU ${env:KEYVAULT_SKU}
...
+
+**Snippet from the corresponding `test-resources.json`.**
+
+Note that the values present in `$templateFileParameters` will map to parameters of the same name.
+```json
+{
+  "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
+  "contentVersion": "1.0.0.0",
+  "parameters": {
+    <...>
```
heaths, line 106: Nit: I'd use a comment like `// ...` that won't throw the diff off and would at least be copy+pastable JSON (well, JSONC, which ARM technically is).

sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -43,6 +43,11 @@ public partial class AzureSasCredential
         public string Signature { get { throw null; } }
         public void Update(string signature) { }
     }
+    public partial class CertificateCredential
+    {
+        public CertificateCredential(System.Security.Cryptography.X509Certificates.X509Certificate2 certificate) { }
+        public System.Security.Cryptography.X509Certificates.X509Certificate2 ClientCertificate { get { throw null; } }
```
heaths, line 49: IMO, if you're goal is to avoid making breaking changes later if you ever support more types, calling this property simply `ClientCertificate` may be too general. Why not just `X509Certificate2`?

sdk/core/Azure.Core/src/Pipeline/HttpPipelineTransportOptions.cs
```diff
@@ -10,9 +11,23 @@ namespace Azure.Core.Pipeline
...
         /// <summary>
         /// A delegate that validates the certificate presented by the server.
         /// </summary>
         public Func<ServerCertificateCustomValidationArgs, bool>? ServerCertificateCustomValidationCallback { get; set; }
+
+        /// <summary>
+        ///
+        /// </summary>
+        /// <value></value>
+        public IList<CertificateCredential> ClientCertificates {get;}
```
heaths, line 31: Nit: just because the code style would otherwise do it:
```suggestion
        public IList<CertificateCredential> ClientCertificates { get; }
```

heaths (changes requested):
1. You can't make local changes to eng/common. They need to be in Azure/azure-sdk-tools in the same directory.
2. Suggestion: I wouldn't rename (move) the .psm1. It's not needed with `import-module -path`, which we could also document if necessary.

heaths:
> I had some issues with `-path` not working consistently. I'm fine either way, but this seems to be more reliable.

I'm okay with that change then. The only reason for the block was local changes to eng/common. FWIW, I do like to test in a local repo like -net, but shouldn't check in like that.

eng/common/TestResources/README.md
christothes: Fixed and moved these changes to the tools repo PR.

heaths (approved):

sdk/core/Azure.Core.TestFramework/src/RecordedTestBase.cs
```diff
@@ -195,8 +195,11 @@ protected RecordedTestBase(bool isAsync, RecordedTestMode? mode = null) : base(i
                 clientOptions.Retry.Delay = TimeSpan.FromMilliseconds(10);
                 clientOptions.Retry.Mode = RetryMode.Fixed;
             }
-
-            clientOptions.Transport = recording.CreateTransport(clientOptions.Transport);
+            // No need to set the transport if we are in Live mode
+            if (Mode != RecordedTestMode.Live)
```
JoshLove-msft, line 199: This is already handled inside of CreateTransport.

sdk/core/Azure.Core/src/ClientCertificate.cs
```diff
@@ -0,0 +1,30 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Security.Cryptography.X509Certificates;
+using Azure.Core;
+
+namespace Azure
+{
+    /// <summary>
+    ///
```
JoshLove-msft, line 10: nit: add docs

sdk/core/Azure.Core/src/Pipeline/HttpClientTransport.cs
```diff
@@ -558,18 +558,25 @@ private static SocketsHttpHandler ApplyOptionsToHandler(SocketsHttpHandler httpH
...
                         new ServerCertificateCustomValidationArgs(
                             certificate is { } ? new X509Certificate2(certificate) : null,
                             x509Chain,
                             sslPolicyErrors));
-#pragma warning restore CA1416 // 'X509Certificate2' is unsupported on 'browser'
             }
+            // Set ClientCertificates
+             foreach (var cred in options.ClientCertificates)
+            {
+               httpHandler.SslOptions ??= new System.Net.Security.SslClientAuthenticationOptions();
```
JoshLove-msft, line 575: nit: add a using

sdk/core/Azure.Core/src/ClientCertificate.cs
christothes: thanks - I need to go do that in a few places.

sdk/core/Azure.Core/src/Pipeline/HttpPipelineTransportOptions.cs
```diff
@@ -10,9 +11,23 @@ namespace Azure.Core.Pipeline
     /// </summary>
     public class HttpPipelineTransportOptions
     {
+        /// <summary>
+        ///
```
JoshLove-msft, line 15: nit: docs

sdk/core/Azure.Core.TestFramework/src/RecordedTestBase.cs
christothes: Not exactly, it creates a ProxyTransport using the current transport, which breaks the settings needed by the client.

sdk/core/Azure.Core/src/DefaultClientOptions.cs
```diff
@@ -10,7 +10,6 @@ internal class DefaultClientOptions: ClientOptions
     {
         public DefaultClientOptions(): base(null, null)
         {
-            Transport = HttpPipelineTransport.Create();
```
JoshLove-msft, line 13: Was this redundant?

sdk/core/Azure.Core.TestFramework/src/RecordedTestBase.cs
christothes: It's the explicit setting of the Transport that breaks the TransportOptions, since that is interpreted by the pipeline as the customer setting the transport via clientOptions.

sdk/core/Azure.Core/src/DefaultClientOptions.cs
christothes: setting of the public property (rather than the private backing field indicates to the pipeline that the Transport has been replaced by the customer and therefor TransportOptions cannot be safely applied.

christothes: It's done here also https://github.com/Azure/azure-sdk-for-net/pull/28645/files/21d226fbef097e370ae4a2d3629991407a124096#diff-e39b05dd3e9930120929e0513ea6029c1e4e44372c7bb0d896d6260b4afbedb0R64

sdk/core/Azure.Core/src/Pipeline/HttpClientTransport.cs
christothes: I think I did that on purpose here to avoid some type name collisions.

sdk/core/Azure.Core/src/DefaultClientOptions.cs
JoshLove-msft: I see.. so is this actually a bug fix?

JoshLove-msft (approved):

sdk/core/Azure.Core/src/DefaultClientOptions.cs
christothes: Yes - although it currently only affects clients that set TransportOptions.

sdk/confidentialledger/Azure.Security.ConfidentialLedger/api/Azure.Security.ConfidentialLedger.netstandard2.0.cs
```diff
@@ -5,6 +5,7 @@ public partial class ConfidentialLedgerClient
         protected ConfidentialLedgerClient() { }
         public ConfidentialLedgerClient(System.Uri ledgerUri, Azure.Core.TokenCredential credential) { }
         public ConfidentialLedgerClient(System.Uri ledgerUri, Azure.Core.TokenCredential credential, Azure.Security.ConfidentialLedger.ConfidentialLedgerClientOptions options) { }
+        public ConfidentialLedgerClient(System.Uri ledgerUri, System.Security.Cryptography.X509Certificates.X509Certificate2 clientCertificate, Azure.Security.ConfidentialLedger.ConfidentialLedgerClientOptions options = null) { }
```
heaths, line 8: Could they not just use a `ClientCertificateCredential`? We had talked about this in some form a while ago, but I must apologize that I forgot what we decided. IIRC, you were going to modify it in some way, or further abstract it to handle some case. Sorry, details have escaped me.

sdk/confidentialledger/Azure.Security.ConfidentialLedger/src/ConfidentialLedgerClient.cs
```diff
@@ -19,55 +18,44 @@ public partial class ConfidentialLedgerClient
...
-            _pipeline = HttpPipelineBuilder.Build(
-                actualOptions,
-                Array.Empty<HttpPipelinePolicy>(),
-                new HttpPipelinePolicy[] { authPolicy },
-                transportOptions,
-                new ResponseClassifier());
-            _ledgerUri = ledgerUri;
-            _apiVersion = actualOptions.Version;
-        }
+            : this(ledgerUri, credential: credential, options: options, identityServiceClient: default)
```
heaths, line 21: Nit: You use `default` here but `null` below, and neither is necessary since it defaults to `null` anyway. Perhaps just elide?

sdk/confidentialledger/Azure.Security.ConfidentialLedger/src/ConfidentialLedgerClient.cs
```diff
@@ -19,55 +18,44 @@ public partial class ConfidentialLedgerClient
...
+                if (clientCertificate == null)
+                    throw new ArgumentNullException(nameof(clientCertificate));
+                if (credential == null)
+                    throw new ArgumentNullException(nameof(credential));
             }
-
             var actualOptions = options ?? new ConfidentialLedgerClientOptions();
-            var transportOptions = GetIdentityServerTlsCertAndTrust(ledgerUri, actualOptions);
-            transportOptions.ClientCertificates.Add(certificateCredential);
+            var transportOptions = GetIdentityServerTlsCertAndTrust(ledgerUri, actualOptions, identityServiceClient).GetAwaiter().GetResult();
```
heaths, line 46: IIRC, isn't this bad in a constructor?

sdk/confidentialledger/Azure.Security.ConfidentialLedger/src/ConfidentialLedgerClient.cs
```diff
@@ -184,13 +172,14 @@ public ConfidentialLedgerClient(Uri ledgerUri, CertificateCredential certificate
             return operation;
         }
 
-        internal static HttpPipelineTransportOptions GetIdentityServerTlsCertAndTrust(Uri ledgerUri, ConfidentialLedgerClientOptions options)
+        internal static async Task<HttpPipelineTransportOptions> GetIdentityServerTlsCertAndTrust(Uri ledgerUri, ConfidentialLedgerClientOptions options, ConfidentialLedgerIdentityServiceClient identityServiceClient = null)
```
heaths, line 175: Recommend "Async" method suffix for consistency.

sdk/confidentialledger/Azure.Security.ConfidentialLedger/src/ConfidentialLedgerClient.cs
```diff
@@ -19,55 +18,44 @@ public partial class ConfidentialLedgerClient
...
         /// <param name="ledgerUri"> The Confidential Ledger URL, for example https://contoso.confidentialledger.azure.com. </param>
-        /// <param name="certificateCredential"> A credential used to authenticate to an Azure Service. </param>
+        /// <param name="clientCertificate"> A <see cref="X509Certificate2"/> used to authenticate to an Azure Service. </param>
         /// <param name="options"> The options for configuring the client. </param>
-        public ConfidentialLedgerClient(Uri ledgerUri, CertificateCredential certificateCredential, ConfidentialLedgerClientOptions options = null)
+        public ConfidentialLedgerClient(Uri ledgerUri, X509Certificate2 clientCertificate, ConfidentialLedgerClientOptions options = null)
+            : this(ledgerUri, clientCertificate: clientCertificate, options: options, identityServiceClient: null)
+        { }
+
+        internal ConfidentialLedgerClient(Uri ledgerUri, TokenCredential credential = null, X509Certificate2 clientCertificate = null, ConfidentialLedgerClientOptions options = null, ConfidentialLedgerIdentityServiceClient identityServiceClient = null)
```
heaths, line 32: At least from this PR - even expanding unmodified code lines in this area - when is `identityServiceClient` non-null? I see further below you even use `??` on it and create one, but haven't yet seen where this can be anything but `null`. Seems some of this code could be simplified.

heaths (approved):
I don't see anything blocking, but had a couple of nits and an open question (again, not blocking).

heaths:
Given the merge commit - though GitHub somehow showed a "condensed version" (wonder when they added that nifty little feature) - I'd generally recommend rebasing + force push. Most often GitHub still shows reviewers only subsequent commits.

If one were to look at the whole merge commit:

image

I see that often in the specs repo and have them do a rebase instead.

heaths (approved):

Issue #28608, opened 2022-05-09
Update code base with stable swagger


maririos:
Currently we are using a dev swagger [here](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/textanalytics/Azure.AI.TextAnalytics/src/autorest.md#autorest-configuration).
We need to change it to the [official one](https://github.com/Azure/azure-rest-api-specs/blob/cognitiveservices-Language-2022-04-01-preview/specification/cognitiveservices/data-plane/Language/preview/2022-04-01-preview/textanalytics.json) from the service and re-generate our code.

PR #28673, opened 2022-05-10
Update Azure.AI.TextAnalytics to latest Swagger

nisha-bhatia:


nisha-bhatia:
https://github.com/Azure/azure-sdk-for-net/issues/28608

maririos:
ohh interesting that when running codegen, the legacy project generated a file too.

@jsquire or @JoshLove-msft do you know if there is a way to tell codegen to ignore the legacy project?

jsquire:
@maririos : That looks like an API export.  If that was triggered by codegen, it's probably scoped to running against the service directory.   Because there isn't an API for that project, we did not generate one originally.   If we had, there probably would have been no change and no additional asset in this PR.

The interesting part for me is that ExportApi seems to be running against projects marked as "not a shipping library" - which I don't believe it should be.

maririos (approved):
I don't want to block your PR because of the addition of the legacy file. We will keep looking into it separately

Issue #28677, opened 2022-05-11
[BUG] Vault Creation failed even passed all the required parameters


Yao725:
### Library name and version

Azure.ResourceManager.KeyVault 1.0.0-beta.8

### Describe the bug

When creating a vault, following code will fail:
```csharp
ArmClient armClient = new ArmClient(new DefaultAzureCredential());
SubscriptionResource subscription = await armClient.GetDefaultSubscriptionAsync();
var rgLro = await subscription.GetResourceGroups().CreateOrUpdateAsync(WaitUntil.Completed, "testVaultCreate", new ResourceGroupData(AzureLocation.WestUS));
ResourceGroupResource resourceGroup = rgLro.Value;
VaultProperties vaultProperties = new VaultProperties(new Guid("myTenantId"), new KeyVaultSku(KeyVaultSkuFamily.A, KeyVaultSkuName.Standard));
VaultCreateOrUpdateContent content = new VaultCreateOrUpdateContent(AzureLocation.WestUS, vaultProperties);
var vaultLro = await resourceGroup.GetVaults().CreateOrUpdateAsync(WaitUntil.Completed, "testVault", content);
VaultResource vault = vaultLro.Value;
```

### Expected behavior

If `accessPolicies` is a required parameter, we should not have a ctor without this parameter. In contrast, if it is not required, the creationg using the sample code should succeed instead of throwing exception.

### Actual behavior

Creation failed with following message:
```
Azure.RequestFailedException : The parameter accessPolicies is not specified.
Status: 400 (Bad Request)
ErrorCode: BadRequest

Content:
{"error":{"code":"BadRequest","message":"The parameter accessPolicies is not specified."}}

Source=Azure.ResourceManager.KeyVault
  StackTrace:
   at Azure.ResourceManager.KeyVault.VaultsRestOperations.<CreateOrUpdateAsync>d__6.MoveNext()
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.ConfiguredTaskAwaitable`1.ConfiguredTaskAwaiter.GetResult()
   at Azure.ResourceManager.KeyVault.VaultCollection.<CreateOrUpdateAsync>d__5.MoveNext()
[...]

PR #28678, opened 2022-05-11
[KeyVault] Support vault creation when not specify accessPolicies

Yao725:
This is a work around to #28677


archerzz (approved):
I suggest to create an issue on `azure-rest-api-specs` for `KeyVault` team.

Yao725 (commented):
After investigation, I decided to still use customization instead of configuration to resolve this issue. Because if we want to generate following code, we need to add both configuration in `autorest.md` and modify autorest.csharp, this doesn't feel worth it.
```csharp
public VaultProperties(Guid tenantId, KeyVaultSku sku, IEnumerable<AccessPolicyEntry> accessPolicies = null)
{
    if (sku == null)
    {
        throw new ArgumentNullException(nameof(sku));
    }

    TenantId = tenantId;
    Sku = sku;
    AccessPolicies = accessPolicies?.ToList() ?? new List<AccessPolicyEntry>();
    PrivateEndpointConnections = new ChangeTrackingList<PrivateEndpointConnectionItemData>();
}
```

This is the specific actions we need to take:

- Add these configurations in autorest.md

```yaml
$.VaultProperties.required.push('accessPolicies');
$.VaultProperties.properties.accessPolicies['x-ms-client-default'] = [];
$.VaultProperties.properties.accessPolicies['x-nullable'] = true;
```

- Support ILise<> in [ParseConstant](https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/Common/Output/Builders/BuilderHelpers.cs#L24) because IList doesn't implement IConvertible.

And I also improve the customization by customizing the ctor instead of serialization method, this helps us to limit customization to the smallest possible extent.

By the way, the CI failure is expected because this customization will take affect after the `remover` change.

sdk/keyvault/Azure.ResourceManager.KeyVault/src/autorest.md
```diff
@@ -7,7 +7,7 @@ azure-arm: true
 library-name: KeyVault
 namespace: Azure.ResourceManager.KeyVault
 tag: package-2021-10
-output-folder: Generated/
+output-folder: $(this-folder)/Generated
```
ArthurMa1978, line 10: If I run the codegen command under `sdk/keyvault/Azure.ResourceManager.KeyVault`, what the real output folder?

Yao725: It will change nothing. I add this one because `Generated/` and `$(this-folder)/Generated` will have different behavior when we encounter the [code](https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/Common/AutoRest/Plugins/GeneratedCodeWorkspace.cs#L112) here, which have influence in the customization. But yes, for this RP and this change, it has nothing to do.

ArthurMa1978: Then consider to have this added into the template

ArthurMa1978 (approved):

sdk/keyvault/Azure.ResourceManager.KeyVault/src/autorest.md
Yao725: I already did this, thank you!

PR #28708, opened 2022-05-12
Use credential in samples

heaths:


heaths:
I decided not to override the constructor to use the Key Vault ChallengeBasedAuthenticationPolicy because it means either copying it, or referencing it, which seems unnecessary given the BearerTokenAuthenticationPolicy works

jsquire (approved):

christothes (approved):

annelo-msft (approved):

PR #28716, opened 2022-05-12
Sdk automation for management and data-plane track2

chunyu3:

The changes of the PR:
- .net track2 automation script for managment and data-plane in swagger CI pipeline
- .net track2 generation for SDK-Generation pipeline
- support autorest config from comments in swagger PR to support onboard new service


eng/scripts/automation/GenerateAndBuildLib.ps1
```diff
@@ -1,5 +1,9 @@
 #Requires -Version 7.0
 $CI_YAML_FILE = "ci.yml"
+
+#mgmt: resourceProvider to sdk package name map
+$packageNameHash = [ordered]@{"vmware" = "avs"; "azure-kusto"="kusto"; "cosmos-db"="cosmosdb";"customer-insights"="customerinsights";"monitor"="insights";"msi"="managedserviceidentity";"web"="appservice"}
```
fengzhou-msft, line 5: insights package is removed and the monitor package is created, so you can remove `"monitor"="insights"`.

This could be a many to many mapping. For instance the swaggers in Azure.ResourceManager comes from common-type, resources and managementgroups. The resources swaggers will affect Azure.ResourceManager and Azure.ResourceManager.Resources packages.

fengzhou-msft, line 5: It seems like resourceProvider to sdk service directory mapping? If so, the service directory for appservice is `websites`, not `appservice`.

chunyu3: @fengzhou-msft would you please provide the full mappings? I just got it from previous script.

eng/scripts/automation/GenerateAndBuildLib.ps1
```diff
@@ -43,43 +50,37 @@ function Update-AutorestConfigFile() {
...
-                (Get-Content $autorestFilePath) -notmatch "- .*.json" |Out-File $autorestFilePath
-                (Get-Content $autorestFilePath) -replace $inputfileRex, ("input-file:" + [Environment]::NewLine + "$inputfile") | Set-Content $autorestFilePath
-            } else {
-                $startNum = (Get-Content $autorestFilePath | Select-String -Pattern '```').LineNumber[0]
-                $fileContent = Get-Content -Path $autorestFilePath
-                $fileContent[$startNum - 1] += ([Environment]::NewLine + "input-file:" + [Environment]::NewLine + "$inputfile")
-                $fileContent | Set-Content $autorestFilePath
+            $inputfileRex = "input-file*:*"
+            # clear
+            (Get-Content $autorestFilePath) -notmatch $requirRex |Out-File $autorestFilePath
```
weshaggard, line 57: Why are we reading and writing this file multiple times as opposed to reading once and then working with the string in memory?

eng/swagger_to_sdk_config.json
```diff
@@ -2,7 +2,7 @@
   "$schema": "https://raw.githubusercontent.com/Azure/azure-rest-api-specs/master/documentation/sdkautomation/SwaggerToSdkConfigSchema.json",
   "generateOptions": {
     "generateScript": {
-      "path": "pwsh ./eng/scripts/Automation-Sdk-Generate.ps1"
+      "path": "pwsh ./eng/scripts/Invoke-GeneratAndBuildV2.ps1"
```
weshaggard, line 5: What is the plan for Automation-Sdk-Generate.ps1? If we are planning to use it any longer we should delete it.

eng/scripts/automation/GenerateAndBuildLib.ps1
weshaggard: Why are we hardcoding these here? Should this info be part of the configuration instead?

eng/scripts/automation/GenerateAndBuildLib.ps1
```diff
@@ -324,25 +322,112 @@ function Invoke-Pack() {
...
+        if ( !$?) {
+            Write-Error "Failed to create sdk project folder. exit code: $?"
+            exit 1
+        }
+        $newpackageoutputJson = Get-Content $newpackageoutput | Out-String | ConvertFrom-Json
+        $packagesToGen = $packagesToGen + @($newpackageoutputJson)
+        Remove-Item $newpackageoutput
+    } else {
+        Write-Host "Generate data-plane SDK client library."
+        npx autorest --version=3.7.3 --csharp $readmeFile --csharp-sdks-folder=$sdkRootPath --skip-csproj --clear-output-folder=true
```
weshaggard, line 374: Is there a reason we don't use the generate command? Such as https://github.com/Azure/azure-sdk-for-net/blob/main/eng/scripts/Automation-Sdk-Generate.ps1#L62? 

If we don't use that then we risk the generation process to get out of sync as we use that msbuild target in our builds that verify the generated code. We should share and use the same exact commands in both places.

eng/scripts/automation/GenerateAndBuildLib.ps1
chunyu3: If we add this info in configuration such as readme.md, we need to parse this information out at first.

eng/scripts/automation/GenerateAndBuildLib.ps1
chunyu3: Thanks @weshaggard  I will investigate if we can combine those all together and use common instance in memory to apply all those actions.

eng/swagger_to_sdk_config.json
chunyu3: We will not use Automation-sdk-Generate.ps1 anymore. I will delete it.

eng/scripts/automation/GenerateAndBuildLib.ps1
chunyu3: We need to know the package namespace(which will be the package folder name) at first, then to update/create the SDK package. Use this generation autorest generation command is an workaround to get the package folder name( or namespace), otherwise we need to parse readme.md ourself.

eng/scripts/automation/GenerateAndBuildLib.ps1
weshaggard: Aren't you already reading some of that configuration via Get-ResourceProviderFromReadme? If possible we should try to avoid hardcoding these extra configurations because they become difficult to maintain and leak details into every place that needs this information.

eng/scripts/automation/GenerateAndBuildLib.ps1
weshaggard: Is this code path all about creating a new package? Do we need only the package name or the namespace (although I think they might be the same)? How did you avoid this for the mgmt case?

eng/scripts/automation/GenerateAndBuildLib.ps1
```diff
@@ -324,25 +322,112 @@ function Invoke-Pack() {
...
+    if (!(Test-Path -Path $readmeFile)) {
+        Write-Error "readme file does not exist."
+        exit 1
+    }
+    
+    $packagesToGen = @()
+    $newpackageoutput = "newPackageOutput.json"
+    if ( $serviceType -eq "resource-manager" ) {
+        Write-Host "Generate resource-manager SDK client library."
+        $package = $service
```
weshaggard, line 360: Why would the package and service be the same here? Is this some special case where we expect mgmt to only have one package per service? If so how to we end up finding the namespace name from the service?

eng/scripts/automation/GenerateAndBuildLib.ps1
weshaggard: @dw511214992 @chunyu3 I suspect this exact code path is what we are talking about in the email thread about where we should put the namespace and package name for new SDKs. Is there any way we could provide this information from the input file data?

eng/scripts/Invoke-GeneratAndBuildV2.ps1
```diff
@@ -0,0 +1,43 @@
...
+foreach ( $relateReadmeFile in $readmeFiles ) {
+    $readmeFile = $relateReadmeFile -replace "\\", "/"
+    
+    $sdkPath =  (Join-Path $PSScriptRoot .. ..)
+    $sdkPath = Resolve-Path $sdkPath
+    $sdkPath = $sdkPath -replace "\\", "/"
+
+    
+    $readme = ""
+    if ($commitid -ne "") {
```
heaths, line 27: Tip: not necessary to compare. PowerShell is very truthy, so just `if ($commitid) {` checks that `$commitid` is not null or an empty string / collection. Same can be said for other comparisons below.

eng/scripts/automation/GenerateAndBuildLib.ps1
heaths, line 5: I recommend putting each element on separate lines (semicolon is optional then). It makes it much easier to read, and future diffs to be smaller. Of course, I agree with @weshaggard as well: this should come from configuration.

eng/scripts/automation/GenerateAndBuildLib.ps1
```diff
@@ -15,14 +19,17 @@ function Get-SwaggerInfo()
         $content = Get-Content .\$AUTOREST_CONFIG_FILE -Raw
         if ($content -match $swaggerInfoRegex)
         {
+            Pop-Location
```
heaths, line 22: This is a very problematic pattern, as you've already seen. Just use `Get-Content "$dir/$AUTOREST_CONFIG_FILE" -raw` above and don't worry about having to push and pop. If you forget to pop in some exit state, bad things can happen. So just avoid it. The only time this really makes sense is when you have to run some utility that needs the current working directory set. That's not the case here. Just use full paths. You're already constructing paths above so do it here as well, and you can do it more simply as I've shown above. `Join-Path` is often overkill.

eng/scripts/automation/GenerateAndBuildLib.ps1
```diff
@@ -223,58 +224,55 @@ function New-MgmtPackageFolder() {
...
       $folderinfo = Get-ChildItem -Path $projectFolder
-      $foldername = $folderinfo.Name
-      $projectFolder = "$sdkPath/sdk/$packageName/$foldername"
+      $mgmtPackageName = $folderinfo.Name
+      $projectFolder = "$sdkPath/sdk/$packageName/$mgmtPackageName"
     } else {
       Write-Host "Path doesn't exist. create template."
       dotnet new -i $sdkPath/eng/templates/Azure.ResourceManager.Template
-      $projectFolder="$sdkPath/sdk/$packageName/Azure.ResourceManager.$packageName"
+      $CaptizedPackageName = (Get-Culture ).TextInfo.ToTitleCase($packageName)
```
heaths, line 247: This shouldn't be culture-specific. Use the invariant culture instead:

```powershell
[System.Globalization.CultureInfo]::InvariantCulture.TextInfo.ToTitleCase(...)
```

eng/scripts/automation/GenerateAndBuildLib.ps1
```diff
@@ -223,58 +224,55 @@ function New-MgmtPackageFolder() {
...
-  
-      $readmefilestr = Get-Content $file
-      Write-Output "autorest.md:$readmefilestr"
     }
   
-    $path=$projectFolder
-    $path=$path.Replace($sdkPath + "/", "")
+    Push-Location $sdkPath
+    $relativeFolderPath = Resolve-Path $projectFolder -Relative
+    Pop-Location
```
heaths, line 268: As noted above, don't abuse push and pop this way. It can cause catastrophic issues.
```suggestion
    $relativeFolderPath = Resolve-Path "$sdkPath/$projectFolder" -Relative
```

eng/scripts/automation/GenerateAndBuildLib.ps1
```diff
@@ -324,25 +322,112 @@ function Invoke-Pack() {
...
+        }
+
+        # pack
+        Invoke-Pack -sdkfolder $projectFolder
+        if ( !$? ) {
+            Write-Error "Failed to packe sdk. exit code: $?"
+            exit 1
+        }
+        # Generate APIs
+        Push-Location $sdkRootPath
```
heaths, line 419: You don't need to do this here as well. Just call `"&$sdkRootPath/eng/scripts/Export-API.ps1" $service` below. Even if  you did have to set the PowerShell-based current working directory (note: this doesn't impact Windows' CWD, which is what any other process you start will use), your `Write-Error` below - if someone has their `$ErrorActionPreference = 'Stop'` will be terminal and not restore the previous location. You could immediately after pushing type `trap { pop-location }` but just avoid this pattern. It's fragile at best.

heaths (dismissed):

eng/scripts/automation/GenerateAndBuildLib.ps1
chunyu3: Hi @weshaggard  Usually the service-directory has the same name of the folder under azure-rest-api-specs/specification, so we can this information via Get-ResourceProviderFromReadme. But in .NET management-plane, there are some exceptions that it will not use different name as sdk serivce-directory, such as `vmware`. In azure-rest-api-specs it called `wmware` , but the service directory for managment-plane sdk is 'avs'. This mapping is to handle such exceptions.

weshaggard: If this an exceptional list that we don't expect to be added to and we aren't repeating this same lookup logic somewhere else than I think we could live with this. However, please add a note in the comment about why this exists.

eng/scripts/automation/GenerateAndBuildLib.ps1
chunyu3: By default, the package of mgmt is the same as service name, and same as the serviceDirectory in azure-sdk-for-net, the  exception will be recorded in $packageNameHash

eng/scripts/Invoke-GeneratAndBuildV2.ps1
```diff
@@ -8,33 +8,78 @@ param (
...
+    }
+}
 
-foreach ( $relateReadmeFile in $readmeFiles ) {
-    $readmeFile = $relateReadmeFile -replace "\\", "/"
-    
-    $sdkPath =  (Join-Path $PSScriptRoot .. ..)
-    $sdkPath = Resolve-Path $sdkPath
-    $sdkPath = $sdkPath -replace "\\", "/"
+$generatedSDKPackages = New-Object 'Collections.Generic.List[System.Object]'
```
heaths, line 39: Tip: this is unnecessarily complicated. Just `$generatedSDKPackages = @()` accomplishes the same thing.

eng/scripts/Invoke-GeneratAndBuildV2.ps1
```diff
@@ -8,33 +8,78 @@ param (
...
+    } elseif ( $commitid) {
+        if ($repoHttpsUrl) {
+            $readme = "$repoHttpsUrl/blob/$commitid/$readmeFile"
+        } else {
+            $readme = "https://github.com/$org/azure-rest-api-specs/blob/$commitid/$readmeFile"
+        }
     } else {
-        $readme = "https://github.com/$org/azure-rest-api-specs/blob/$commitid/$readmeFile"
+        Write-Error "No readme File path provided."
+        exit 1
```
heaths, line 57: Tip: you can do both with just `throw "No readme file path provided."`. It will err and exit with err all at the same time.

eng/scripts/Invoke-GeneratAndBuildV2.ps1
```diff
@@ -8,33 +8,78 @@ param (
...
+
+    if ($autorestConfigYaml) {
+        $readmeFiles[$i] = $readme
+        $autorestConfigYaml = ConvertTo-YAML $yml
     }
+    Invoke-GenerateAndBuildSDK -readmeAbsolutePath $readme -sdkRootPath $sdkPath -autorestConfigYaml "$autorestConfigYaml" -downloadUrlPrefix "$downloadUrlPrefix" -generatedSDKPackages $generatedSDKPackages
+}
+
+#update services without readme.md
+[System.Collections.ArrayList] $serviceWithReadme = @()
```
heaths, line 68: Declaring the type is unnecessary. By default, `@()` is always a list of objects.

eng/scripts/Invoke-GeneratAndBuildV2.ps1
```diff
@@ -8,33 +8,78 @@ param (
...
+foreach( $readme in $readmeFiles) {
+    $service, $serviceType = Get-ResourceProviderFromReadme $readme
+    $serviceWithReadme.Add($service)
+}
+[System.Collections.ArrayList] $inputFileToGen = @()
```
heaths, line 73: Same here. `@()` is already a list of objects.

eng/scripts/Invoke-GeneratAndBuildV2.ps1
```diff
@@ -8,33 +8,78 @@ param (
...
+foreach( $file in $inputFilePaths) {
+    $file -match "specification/(?<specName>.*)/data-plane|resource-manager"
+    if (!$serviceWithReadme.Contains($matches["specName"])) {
+        $inputFileToGen.Add($file)
+    }
+}
 
-    Invoke-GenerateAndBuildSDK -readmeAbsolutePath $readme -sdkRootPath $sdkPath -generatedSDKPackages $generatedSDKPackages
+if ($inputFileToGen.Count -gt 0 ) {
```
heaths, line 81: Tip: really just need `if ($inputFileToGen)`. If it's null or empty, the condition evaluates as `false` already.

eng/scripts/automation/GenerateAndBuildLib.ps1
```diff
@@ -73,19 +79,56 @@ function Update-AutorestConfigFile() {
...
             $fileContent | Set-Content $autorestFilePath
             if ( !$? ) {
                 Write-Error "Failed to update autorest.md. exit code: $?"
                 exit 1
             }
-        }  
+        }
+        
+        # update autorest.md with configuration
+        if ( $autorestConfigYaml -ne "") {
```
heaths, line 92: Tip: just `if ($autorestConfigYaml)` is fine. If null or empty, it will evaluate as false anyway.

eng/scripts/automation/GenerateAndBuildLib.ps1
```diff
@@ -73,19 +79,56 @@ function Update-AutorestConfigFile() {
...
+                }
+            }
+
+            $fileContent | Out-File $autorestFilePath
+        }
     } else {
-        Write-Error "autorest.md doesn't exist."
-        exit 1
+        Write-Host "autorest.md does not exist. start to create one."
+        if ( $autorestConfigYaml -ne "") {
```
heaths, line 123: Tip: just `if ($autorestConfigYaml)` is fine. If null or empty, it will evaluate as false anyway.

eng/scripts/automation/GenerateAndBuildLib.ps1
```diff
@@ -392,42 +464,166 @@ function Invoke-GenerateAndBuildSDK () {
...
+        } else {
+            # pack
+            Write-Host "Start to pack sdk"
+            Invoke-Pack -sdkfolder $projectFolder
+            if ( !$? ) {
+                Write-Error "Failed to packe sdk. exit code: $?"
+            }
+            # Generate APIs
+            Write-Host "Start to export api for $service"
+            pwsh $sdkRootPath/eng/scripts/Export-API.ps1 $service
```
heaths, line 517: Why are you starting a new `pwsh` when you could just run the script? It's much more expensive. And if you set the `$ErrorActionPreference = 'Stop'` it will terminate the whole session automatically without having to check `$?`.

eng/scripts/Invoke-GeneratAndBuildV2.ps1
chunyu3: The $generatedSDKPackages will be passed to the inner function as reference value, and in the inner function, it will add new item into it. if define it as `$generatedSDKPackages = @()`, it is an object, but it failed to add the item into it in inner function. So I need to defined it as an Collections.Generic.List

eng/scripts/Invoke-GeneratAndBuildV2.ps1
chunyu3: `@()` is a fixed collection, it cannot add item into it. so I need to clearly define it as ArrayList.

eng/scripts/Invoke-GeneratAndBuildV2.ps1
chunyu3: updated

eng/scripts/automation/GenerateAndBuildLib.ps1
chunyu3: update to use `&` to call powershell script

eng/scripts/Invoke-GeneratAndBuildV2.ps1
heaths: You just use `$generatedSDKPackages += "new item"`, for example. PowerShell supports that for all collections by default. No need to call `Add` explicitly.

eng/scripts/Invoke-GeneratAndBuildV2.ps1
```diff
@@ -0,0 +1,87 @@
...
+}
+
+if ($inputFileToGen) {
+    UpdateExistingSDKByInputFiles -inputFilePaths $inputFileToGen -sdkRootPath $sdkPath -headSha $commitid -repoHttpsUrl $repoHttpsUrl -downloadUrlPrefix "$downloadUrlPrefix" -generatedSDKPackages $generatedSDKPackages
+}
+
+$outputJson = [PSCustomObject]@{
+    packages = $generatedSDKPackages
+}
+$outputJson | ConvertTo-Json -depth 100 | Out-File $outputJsonFile
```
weshaggard, line 87: Do you intend to just write this as output or have it be a return from the script?

eng/scripts/automation/GenerateAndBuildLib.ps1
```diff
@@ -32,59 +42,93 @@ function Get-SwaggerInfo()
...
-                (Get-Content $autorestFilePath) -notmatch "- .*.json" |Out-File $autorestFilePath
-                (Get-Content $autorestFilePath) -replace $requirRex, ("require:" + [Environment]::NewLine + "- " + "$requirefile") | Set-Content $autorestFilePath
-            } elseif ((Get-Content $autorestFilePath | Select-String -Pattern $inputfileRex).Matches.Success) {
-                (Get-Content $autorestFilePath) -notmatch "- .*.json" |Out-File $autorestFilePath
-                $requirefile = $requirefile + [Environment]::NewLine + "csharp: true";
-                (Get-Content $autorestFilePath) -replace $inputfileRex, ("require:" + [Environment]::NewLine + "- " + "$requirefile") | Set-Content $autorestFilePath
+            $inputfileRex = "input-file*:*"
+            $fileContent = Get-Content -Path $autorestFilePath
+            # clear
+            $fileContent = $fileContent -notmatch $requirRex
```
weshaggard, line 62: I'm not sure I understand what these are doing? Are you trying to strip out parts of the file? Should we instead switch to a model that just pulls out what we want with a regex instead?

eng/scripts/automation/GenerateAndBuildLib.ps1
```diff
@@ -32,59 +42,93 @@ function Get-SwaggerInfo()
...
+
+            $fileContent | Out-File $autorestFilePath
+        }
     } else {
-        Write-Error "autorest.md doesn't exist."
-        exit 1
+        Write-Host "autorest.md does not exist. start to create one."
+        if ( $autorestConfigYaml ) {
+            Write-Host "Create autorest.md with configuration."
+            $autorestConfigYaml = "# Azure.IoT.DeviceUpdate" + [Environment]::NewLine + '``` yaml' +  [Environment]::NewLine + $autorestConfigYaml + '```' + [Environment]::NewLine;
```
weshaggard, line 125: Did we intend to hardcode Azure.IoT.DeviceUpdate here?

eng/scripts/automation/GenerateAndBuildLib.ps1
```diff
@@ -223,58 +272,56 @@ function New-MgmtPackageFolder() {
...
         [string]$AUTOREST_CONFIG_FILE = "autorest.md",
         [string]$outputJsonFile = "newPacakgeOutput.json"
     )
   
-    $projectFolder="$sdkPath/sdk/$packageName/Azure.ResourceManager.*"
+    if ($packageName -eq "") {
+        $packageName = $service
+    }
+
+    $projectFolder = (Join-Path $sdkPath "sdk" $packageName "Azure.ResourceManager.*")
```
weshaggard, line 284: Do we need to resolve the `*`?

eng/scripts/automation/GenerateAndBuildLib.ps1
```diff
@@ -284,65 +331,299 @@ function Invoke-Generate() {
...
+            New-DataPlanePackageFolder -service $service -namespace $namespace -sdkPath $sdkRootPath -readme $readmeFile -autorestConfigYaml "$autorestConfigYaml" -outputJsonFile $newpackageoutput
+            if ( !$? ) {
+                Write-Error "Failed to create sdk project folder. exit code: $?"
+                exit 1
+            }
+            $newpackageoutputJson = Get-Content $newpackageoutput | Out-String | ConvertFrom-Json
+            $packagesToGen = $packagesToGen + @($newpackageoutputJson)
+            Remove-Item $newpackageoutput
+        } else {
+            npx autorest --version=3.8.4 --csharp $readmeFile --csharp-sdks-folder=$sdkRootPath --skip-csproj --clear-output-folder=true
```
weshaggard, line 445: Given our new plans I don't think we will ever be in this else block. If we don't have an autorest config we will either create one if we have the data from the input or we will skip for this library.

eng/scripts/Invoke-GeneratAndBuildV2.ps1
chunyu3: Thanks @heaths I will try.

eng/scripts/automation/GenerateAndBuildLib.ps1
chunyu3: I need to update the autorest.md from required readme to input-file json file, and vice versa. In order to support different scenarios, the input-file / require may define in one line, or use `-` to several lines (see following example), I need to clear all the previous lines and add the new lines. 
```
input-file: https://*****/**.json
```
or 
```
input-file:
- https://****/**.json
```

eng/scripts/automation/GenerateAndBuildLib.ps1
chunyu3: updated

eng/scripts/automation/GenerateAndBuildLib.ps1
chunyu3: Yes. for mgmt, most case the package name is `Azure.ResourceManager.<serviceDirectory>`, but it may have exception in current already released RPs. So we use this `*` to resolve out the exact existing package name.

eng/scripts/automation/GenerateAndBuildLib.ps1
chunyu3: This part is try to support existing rare scenarios which may have multiple sdk packages in one RP and it define batch task in the readme.csharp.md.

eng/scripts/Invoke-GeneratAndBuildV2.ps1
```diff
@@ -0,0 +1,87 @@
...
+$swaggerDir = $swaggerDir -replace "\\", "/"
+[string[]] $inputFilePaths = $inputJson.changedFiles;
+$readmeFiles = $inputJson.relatedReadmeMdFiles
+$commitid = $inputJson.headSha
+$repoHttpsUrl = $inputJson.repoHttpsUrl
+$downloadUrlPrefix = $inputJson.installInstructionInput.downloadUrlPrefix
+[string] $autorestConfig = $inputJson.autorestConfig
+
+$autorestConfigYaml = ""
+if ( $null -ne $autorestConfig -and $autorestConfig -ne "") {
```
archerzz, line 21: You can directly compare `if ($autorestConfigYaml)`, I guess. Or maybe `[string]::IsNullOrEmpty($autorestConfigYaml)`

eng/scripts/automation/GenerateAndBuildLib.ps1
weshaggard: I would like to avoid having this case if possible. Can you point me at an example in the specs repo that would hit this case?

eng/scripts/automation/GenerateAndBuildLib.ps1
```diff
@@ -32,59 +42,94 @@ function Get-SwaggerInfo()
...
+        [string]$readme = "",
+        [string]$autorestConfigYaml = ""
     )
-    if (Test-Path -Path $autorestFilePath) {
-        if ($readme -ne "") {
-            Write-Host "Updating autorest.md file to config required readme file."
+
+    if ((Test-Path -Path $autorestFilePath) -and (![String]::IsNullOrWhiteSpace((Get-Content -Path $autorestFilePath)))) {
+        if (($readme -ne "") -or ($inputfile -ne "")) {
             $requirRex = "require*:*";
```
weshaggard, line 59: This is not a valid regex. I suspect you want something like `require.*:.*`.

eng/scripts/automation/GenerateAndBuildLib.ps1
```diff
@@ -32,59 +42,94 @@ function Get-SwaggerInfo()
...
-            if ((Get-Content $autorestFilePath | Select-String -Pattern $requirRex).Matches.Success) {
-                (Get-Content $autorestFilePath) -notmatch "- .*.md" |Out-File $autorestFilePath
-                (Get-Content $autorestFilePath) -notmatch $inputfileRex |Out-File $autorestFilePath
-                (Get-Content $autorestFilePath) -notmatch "- .*.json" |Out-File $autorestFilePath
-                (Get-Content $autorestFilePath) -replace $requirRex, ("require:" + [Environment]::NewLine + "- " + "$requirefile") | Set-Content $autorestFilePath
-            } elseif ((Get-Content $autorestFilePath | Select-String -Pattern $inputfileRex).Matches.Success) {
-                (Get-Content $autorestFilePath) -notmatch "- .*.json" |Out-File $autorestFilePath
-                $requirefile = $requirefile + [Environment]::NewLine + "csharp: true";
-                (Get-Content $autorestFilePath) -replace $inputfileRex, ("require:" + [Environment]::NewLine + "- " + "$requirefile") | Set-Content $autorestFilePath
+            $inputfileRex = "input-file*:*"
```
weshaggard, line 60:
```suggestion
            $inputfileRegex = "input-file.*:.*"
```

eng/scripts/automation/GenerateAndBuildLib.ps1
```diff
@@ -32,59 +42,94 @@ function Get-SwaggerInfo()
...
+            $fileContent = $fileContent -notmatch "csharp: true"
+            $fileContent = $fileContent -notmatch $inputfileRex
+            $fileContent = $fileContent -notmatch "- .*.json"
+
+            $startNum = ($fileContent | Select-String -Pattern '```').LineNumber[0]
+            $configline = ""
+            if ($readme) {
+                Write-Host "Updating autorest.md file to config required readme file."
+                $requirefile = $readme
+                $configline = "require:" + [Environment]::NewLine + "- " + "$requirefile" + [Environment]::NewLine + "csharp: true"
```
weshaggard, line 74:
```suggestion
                $configline = "require:\n- ${requirefile}\ncsharp: true"
```

eng/scripts/automation/GenerateAndBuildLib.ps1
```diff
@@ -32,59 +42,94 @@ function Get-SwaggerInfo()
...
+            } elseif ($inputfile) {
+                Write-Host "Updating autorest.md file to update input-file."
+                if ($inputfile.StartsWith('-')) {
```
weshaggard, line 77: While I understand we support a list do we really need to use it when we generate our own file?

eng/scripts/automation/GenerateAndBuildLib.ps1
weshaggard: A similar regex can be used if you just want to replace similar to:
```
$fileContent = $fileContent -replace "(?s)(input-file:[\s-]*)(?<url>.+)","`$1$inputFile"
```

You can do something similar for the other settings.

eng/scripts/automation/GenerateAndBuildLib.ps1
```diff
@@ -32,59 +42,94 @@ function Get-SwaggerInfo()
...
+                $startNum = $range[0];
+                $lines = $range[1] - $range[0] - 1
+                $autorestConfigYaml = ($autorestConfigYaml | Select -Skip $startNum | Select -First $lines) |Out-String
             }
-            
-            if ( $? -ne $True) {
-                Write-Error "Failed to update autorest.md. exit code: $?"
-                exit 1
+
+            Install-Module -Name powershell-yaml -Force -Verbose -Scope CurrentUser
```
weshaggard, line 102: We should probably figure out if we want to use regex or yaml parsing for updating the file. Based on the settings I've seen I suspect we can just use regex to find/replace the options we need. The examples I showed above should get you most of the way.

eng/swagger_to_sdk_config.json
benbp, line 5:
```suggestion
      "path": "pwsh ./eng/scripts/Invoke-GenerateAndBuildV2.ps1"
```

eng/scripts/Invoke-GeneratAndBuildV2.ps1
```diff
@@ -0,0 +1,87 @@
...
+    $autorestConfig | Set-Content "config.md"
+    $autorestConfigYaml = Get-Content -Path .\config.md
```
benbp, line 23: Use forward slashes in paths for cross platform support.
```suggestion
    $autorestConfigYaml = Get-Content -Path ./config.md
```

eng/scripts/Invoke-GeneratAndBuildV2.ps1
```diff
@@ -0,0 +1,87 @@
...
+    $range = ($autorestConfigYaml | Select-String -Pattern '```').LineNumber
+    if ( $range.count -gt 1) {
+        $startNum = $range[0];
+        $lines = $range[1] - $range[0] - 1
+        $autorestConfigYaml = ($autorestConfigYaml | Select -Skip $startNum | Select -First $lines) |Out-String
```
benbp, line 28: When you call `Get-Content` without `-Raw` it returns an array type (the file split on newline). So this variable may be a string or an array depending on conditions which could cause unintended behavior.

eng/scripts/Invoke-GeneratAndBuildV2.ps1
```diff
@@ -0,0 +1,87 @@
...
+    }
+    Install-Module -Name powershell-yaml -Force -Verbose -Scope CurrentUser
+    Import-Module powershell-yaml
```
benbp, line 31: Please use some of our module install helpers. See [import example](https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/scripts/stress-testing/stress-test-deployment-lib.ps1#L6) and [usage example](https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/scripts/stress-testing/stress-test-deployment-lib.ps1#L260). Running `Install-Module` even if it's already installed adds a lot of unnecessary time to the script, e.g. when testing locally.

eng/scripts/Invoke-GeneratAndBuildV2.ps1
```diff
@@ -0,0 +1,87 @@
...
+
+#update services without readme.md
+[System.Collections.ArrayList] $serviceWithReadme = @()
+foreach( $readme in $readmeFiles) {
+    $service, $serviceType = Get-ResourceProviderFromReadme $readme
+    $serviceWithReadme.Add($service)
+}
+[System.Collections.ArrayList] $inputFileToGen = @()
+foreach( $file in $inputFilePaths) {
+    $file -match "specification/(?<specName>.*)/data-plane|resource-manager"
```
benbp, line 74: Do you need to check and handle the case where the match is false for some reason?

eng/scripts/automation/GenerateAndBuildLib.ps1
```diff
@@ -1,18 +1,28 @@
...
 $CI_YAML_FILE = "ci.yml"
+
+#mgmt: resourceProvider to sdk package name map
+$packageNameHash = [ordered]@{"vmware" = "avs"; 
+                            "azure-kusto"="kusto";
+                            "cosmos-db"="cosmosdb";
+                            "customer-insights"="customerinsights";
+                            "monitor"="insights";
+                            "msi"="managedserviceidentity";
+                            "web"="appservice"}
```
benbp, line 11: It's a little easier to read and update when the whitespace is consistent.
```suggestion
$packageNameHash = [ordered]@{
    "vmware" = "avs"; 
    "azure-kusto" = "kusto";
    "cosmos-db" = "cosmosdb";
    "customer-insights" = "customerinsights";
    "monitor" = "insights";
    "msi" = "managedserviceidentity";
    "web" = "appservice"
}
```

eng/scripts/automation/GenerateAndBuildLib.ps1
```diff
@@ -32,59 +42,94 @@ function Get-SwaggerInfo()
...
-            if ( $? -ne $True) {
-                Write-Error "Failed to update autorest.md. exit code: $?"
-                exit 1
+
+            Install-Module -Name powershell-yaml -Force -Verbose -Scope CurrentUser
+            Import-Module powershell-yaml
+            $yml = ConvertFrom-YAML $autorestConfigYaml
+
+            $fileContent = Get-Content -Path $autorestFilePath
+            Foreach ( $key in $yml.keys) {
```
benbp, line 107:
```suggestion
            foreach ( $key in $yml.keys) {
```

eng/scripts/automation/GenerateAndBuildLib.ps1
```diff
@@ -284,65 +332,299 @@ function Invoke-Generate() {
...
-    try
-    {
-        if ($readmeFile -match $readmeFileRegexWithSpec)
-        {
-            return $matches["specName"], $matches["serviceType"]
+    $readmeFile = $readmeFile -replace "\\", "/"
+    $pathArray = $readmeFile.Split("/");
+
+    if ( $pathArray.Count -lt 3) {
+        Write-Error "Error: invalid readme file path."
```
benbp, line 370: Why is it invalid? Would the user reading this error message know what they did wrong?

eng/scripts/automation/GenerateAndBuildLib.ps1
```diff
@@ -284,65 +332,299 @@ function Invoke-Generate() {
...
+        exit 1
+    }
+
+    $specName = $pathArray[-3]
+    $serviceType = $pathArray[-2]
+    Write-Host "specName:$specName, serviceType: $serviceType"
```
benbp, line 376:
```suggestion
    Write-Host "specName: $specName, serviceType: $serviceType"
```

eng/scripts/automation/GenerateAndBuildLib.ps1
```diff
@@ -284,65 +332,299 @@ function Invoke-Generate() {
...
+        [string]$downloadUrlPrefix = "",
+        [object]$generatedSDKPackages
+    )
+    $readmeFile = $readmeAbsolutePath -replace "\\", "/"
+    Write-Host "readmeFile:$readmeFile"
+    $service, $serviceType = Get-ResourceProviderFromReadme $readmeFile
+    Write-Host "service:$service, serviceType:$serviceType"
+    
+    if (!$readmeFile.StartsWith("http") -And !(Test-Path -Path $readmeFile)) {
+        Write-Error "readme file does not exist."
```
benbp, line 395:
```suggestion
        Write-Error "readme file '$readmeFile' does not exist."
```

eng/scripts/automation/GenerateAndBuildLib.ps1
```diff
@@ -284,65 +332,299 @@ function Invoke-Generate() {
...
+        $package = $service
+        if ($packageNameHash[$service] -ne "") {
+            $package = $packageNameHash[$service]
         }
-        if ($readmeFile -match $readmeFileRegex)
-        {
-            return $matches["specName"], $matches["serviceType"]
+        New-MgmtPackageFolder -service $service -packageName $package -sdkPath $sdkRootPath -commitid $commitid -readme $readmeFile -outputJsonFile $newpackageoutput
+        if ( !$?) {
+            Write-Error "Failed to create sdk project folder. exit code: $?"
```
benbp, line 409: If the inner function call invokes `exit` you will not hit this block of code.

eng/scripts/automation/GenerateAndBuildLib.ps1
```diff
@@ -284,65 +332,299 @@ function Invoke-Generate() {
...
+            exit 1
+        }
+        $newpackageoutputJson = Get-Content $newpackageoutput | Out-String | ConvertFrom-Json
```
benbp, line 412: Please use camel case for variable names for readability.

eng/scripts/automation/GenerateAndBuildLib.ps1
```diff
@@ -284,65 +332,299 @@ function Invoke-Generate() {
...
+            $outputFolder = $yml["output-folder"]
+            if ($outputFolder) {
+                $directories = $outputFolder.Split("/");
+                $service = $directories[-2];
+                $namespace = $directories[-1];
+            }
+
+            New-DataPlanePackageFolder -service $service -namespace $namespace -sdkPath $sdkRootPath -readme $readmeFile -autorestConfigYaml "$autorestConfigYaml" -outputJsonFile $newpackageoutput
+            if ( !$? ) {
+                Write-Error "Failed to create sdk project folder. exit code: $?"
```
benbp, line 439: Same comment as above re exit stopping execution in the inner function.

eng/scripts/automation/GenerateAndBuildLib.ps1
benbp: Unless I'm reading how things are invoked wrong. Also `$?` is a boolean, not an exit code. For powershell cmdlets/functions, prefer to throw. And similarly if you are handling an external binary call and want to propagate failures up to the caller, wrap them in a throw.

```
$ cat ./test.ps1
function fooExit()
{
    [CmdletBinding()]
    param()
    exit 1
}

function fooEntry()
{
    [CmdletBinding()]
    param()
    fooExit
    exit 2
}

fooEntry
$ ./test.ps1
$ $LASTEXITCODE
1
```

benbp:
This is quite a large change with a lot of functionality. It would be really good to add a set of unit tests to ensure all the different generation scenarios are covered, and to help illustrate for others what the expected inputs and outputs should look like. I have some (work in progress) documentation on good powershell testing practices [here](https://github.com/Azure/azure-sdk-tools/blob/75f0ce338829ea4b6af65f4850022494811fdf80/doc/development/powershell.md#testing).

eng/scripts/Invoke-GeneratAndBuildV2.ps1
chunyu3: updated.

eng/scripts/automation/GenerateAndBuildLib.ps1
chunyu3: yes. The script both support in swagger CI pipeline which will use readme as configure, and also it will support SDK-package-starter which may have input-file lists as input.

eng/scripts/automation/GenerateAndBuildLib.ps1
chunyu3: I may not now exactly which key/pair will be set in the autorestConfig xml. So it may not be able to define the regex to update the autorest.md, so I use yaml parsing to do it.

eng/swagger_to_sdk_config.json
chunyu3: update the script file name. Thanks @benbp

eng/scripts/Invoke-GeneratAndBuildV2.ps1
chunyu3: updated

eng/scripts/Invoke-GeneratAndBuildV2.ps1
chunyu3: Thanks @benbp . I updated accordingly.

eng/scripts/Invoke-GeneratAndBuildV2.ps1
chunyu3: Thanks @benbp. I need to read this content as string array.

eng/scripts/Invoke-GeneratAndBuildV2.ps1
chunyu3: added the check.

eng/scripts/automation/GenerateAndBuildLib.ps1
chunyu3: updated

eng/scripts/automation/GenerateAndBuildLib.ps1
chunyu3: Add the more description about the error: A valid readme file path should contain specName and serviceType and be of the form <specName>/<serviceType>/readme.md, e.g. specification/deviceupdate/data-plane/readme.md

eng/scripts/automation/GenerateAndBuildLib.ps1
chunyu3: updated

eng/scripts/Invoke-GeneratAndBuildV2.ps1
chunyu3: No. the script will support both scenarios : provide autorestConfig or not.

eng/scripts/Invoke-GeneratAndBuildV2.ps1
chunyu3: Hello @heaths I have tried again. the normal list cannot be passed as reference, the item added to it in the function, and after back to caller, the list does not contain item.

eng/scripts/automation/GenerateAndBuildLib.ps1
chunyu3: We need to support the scenario which will contain more than one line (see senario 2), and it also switch between input-file and require. The path of the file can be url or file path. So it is not be able to use regex and replace to support all those scenarios.

eng/scripts/automation/GenerateAndBuildLib.ps1
chunyu3: Hi @weshaggard Purview is such kind of RP, in sdk\purview it will contain more than one sdk packages (administrator, catalog, scanning), and there will be only one readme file. In order to support this, we need to use `batch` to defined in readme.md to support generate all those packages at once.

eng/scripts/Invoke-GeneratAndBuildV2.ps1
chunyu3: updated

eng/scripts/automation/GenerateAndBuildLib.ps1
chunyu3: Thanks @weshaggard. I updated the regex.

eng/scripts/automation/GenerateAndBuildLib.ps1
chunyu3: updated.

eng/scripts/automation/GenerateAndBuildLib.ps1
chunyu3: I will remove exit in inner function

eng/scripts/automation/GenerateAndBuildLib.ps1
weshaggard: @dw511214992 @josefree what is our strategy for handling batches? I know we had ah email about how that data would be passed for this multiple sdk case.

eng/scripts/automation/GenerateAndBuildLib.ps1
weshaggard: My replace statement also supports the multiple line case that is what the `(?s)` option is for. If you test that regex locally it should handle both cases. Even if it isn't a url the regex should work for those cases.

eng/scripts/automation/GenerateAndBuildLib.ps1
```diff
@@ -32,59 +42,94 @@ function Get-SwaggerInfo()
...
         [string]$inputfile = "",
-        [string]$readme = ""
+        [string]$readme = "",
+        [string]$autorestConfigYaml = ""
     )
-    if (Test-Path -Path $autorestFilePath) {
-        if ($readme -ne "") {
-            Write-Host "Updating autorest.md file to config required readme file."
+
+    if ((Test-Path -Path $autorestFilePath) -and (![String]::IsNullOrWhiteSpace((Get-Content -Path $autorestFilePath)))) {
```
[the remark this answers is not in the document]
weshaggard: You should still be able to capture that content into variable instead of needing to read it multiple times.

weshaggard:
@chunyu3 I still think we can do some better work on the parsing but as we talked it is really hard to tell without some tests so adding some basic testing with some inputs and outputs that will give so some thing to focus on to ensure the logic is working as expected.

eng/scripts/automation/GenerateAndBuildLib.ps1
chunyu3: If we read the file content without `-Raw`  `$filecontent = (Get-Content -path autorest.md)`, it will return an string array, one line on string, then the regex cannot handle multiple lines scenario.
If we read the file content with `-Raw` `$filecontent = (Get-Content -path autorest.md -Raw)`, it will return an string. The regex will return all the content after 'input-file' line (the input-file and other content behind)  which is not what we want, and cannot replace it.

eng/scripts/automation/GenerateAndBuildLib.ps1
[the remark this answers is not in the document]
chunyu3: Yes, I agree. I will define an variable with empty string, and Get-Content for the variable if I the file exist.

eng/scripts/automation/GenerateAndBuildLib.ps1
weshaggard: You can fix the regex such that it doesn't return the rest of the file by adding `(?s)(input-file:[\s-]*)(?<url>\S+)` which says match all the non-whitespace characters.

chunyu3:
> @chunyu3 I still think we can do some better work on the parsing but as we talked it is really hard to tell without some tests so adding some basic testing with some inputs and outputs that will give so some thing to focus on to ensure the logic is working as expected.

Hi @weshaggard  Yes. I will write document to provide some sample input ant output for testing. Thanks

chunyu3:
Hi @weshaggard  @benbp I add some comments in the script and some sample input ant output for testing different scenarios. Thanks.

benbp:
@chunyu3 can you turn the output docs into runnable unit tests? I have some examples of json validation in [here](https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/scripts/job-matrix/tests/job-matrix-functions.modification.tests.ps1) and in the unit testing doc I linked in my previous comment.

eng/scripts/automation/unified-pipeline-test.md
```diff
@@ -0,0 +1,186 @@
+# scenarios
```
weshaggard, line 1: It would be potentially helpful if these could be separate files or perhaps embedded in another PS1 script that we can run for testing purposes. It can be a simple script that just creates the inputs calls your script and then compares the output with what you expect it to be.

chunyu3: Hello @weshaggard So we will need to write script test as @benbp suggested. Right?

benbp: Thank you, just to re-iterate please follow the Pester [unit test structure](https://github.com/Azure/azure-sdk-tools/blob/75f0ce338829ea4b6af65f4850022494811fdf80/doc/development/powershell.md#unit-testing).

chunyu3: Hi @weshaggard  @benbp  I added tests under tests directories. Would you please take a look at it? Thanks

eng/scripts/automation/GenerateAndBuildLib.ps1
weshaggard: @chunyu3 I'd still like for you to try out my regex for replacement as I think it can make this code a lot simpler.

weshaggard:
@chunyu3 It looks like a lot of this script is about parsing and updating the autorest configuration files and I suspect the automation scripts in the other languages will need similar code. Please file an issue to add some common scripts (or tools) for handling this in a common way so that we can share it and we can be consistent about how we handle it.

eng/scripts/automation/GenerateAndBuildLib.ps1
chunyu3: As we discussed in the meeting, the new SDKs will provide namespace and package name in the swagger PR comment, and it will be passed in the input file when call the script. This code path here is for existing SDKs, and the one which has more than one SDK packages one readme file, and it define 'batch' to generate multiple SDK packages. When all the RPs have convert to one readme one SDK package structure, we will not call 'autorest' cmd here.

eng/scripts/automation/GenerateAndBuildLib.ps1
chunyu3: resolved.

eng/scripts/automation/GenerateAndBuildLib.ps1
chunyu3: Thanks @weshaggard I will try your Regex. The problem I met before is that in order to match multiple lines, I need to Get-content with `-Raw`, in that case the all content will be read as a string. And the problem is that it will return all the content starting from the first match line which is not what we want. I will retry and see if any Regex can resolve this issue and only return the matched lines only. Thanks

chunyu3: Hello @weshaggard I defined an Regex which can match out all scenarios and use `-replace` to update the autorest.md. Please take a look. Thanks.

chunyu3:
Hello @weshaggard  @benbp  I used `-replace` and  `Regex` to update autorest.md in script now. And I filed an issue https://github.com/Azure/azure-sdk-for-net/issues/30148 to track the autorest.md parsing and modification requirements.

eng/scripts/tests/Invoke-GenerateAndBuildV2.tests.ps1
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    It "scenaro 1: first onboard" {
+        & $PSScriptRoot/../Invoke-GenerateAndBuildV2.ps1 -inputJsonFile $PSScriptRoot/firstOnboardInput.json -outputJsonFile $PSScriptRoot/firstOnboardOutput.json
+        $outputJson = Get-Content $PSScriptRoot/firstOnboardOutput.json | Out-String | ConvertFrom-Json
+
+        $outputJson.packages.Count | Should -Be 1
+        $outputJson.packages[0].packageName | Should -Be "Azure.IoT.DeviceUpdate"
+        $outputJson.packages[0].result | Should -Be "succeeded"
+        Remove-Item $PSScriptRoot/firstOnboardOutput.json
```
benbp, line 28: You remove the output file here but not in the below tests. Was that intended?

eng/scripts/tests/Invoke-GenerateAndBuildV2.tests.ps1
```diff
@@ -0,0 +1,52 @@
...
+        $outputJson = Get-Content $PSScriptRoot/updateExistingOutput.json | Out-String | ConvertFrom-Json
+
+        $outputJson.packages.Count | Should -Be 1
+        $outputJson.packages[0].packageName | Should -Be "Azure.IoT.DeviceUpdate"
+        $outputJson.packages[0].result | Should -Be "succeeded"
+    }
+
+    It "scenaro 2: update existing SDK with autorest config" {
+        & $PSScriptRoot/../Invoke-GenerateAndBuildV2.ps1 -inputJsonFile $PSScriptRoot/updateExistingWithConfigInput.json -outputJsonFile $PSScriptRoot/updateExistingWithConfigOutput.json
+        $outputJson = Get-Content $PSScriptRoot/updateExistingWithConfigOutput.json | Out-String | ConvertFrom-Json
```
benbp, line 42: Since you have the output json already included in these changes, it may also be worth doing an object or text comparison of the expected output file and the one generated by this command.

eng/scripts/tests/Invoke-GenerateAndBuildV2.tests.ps1
```diff
@@ -0,0 +1,52 @@
...
+
+Install-ModuleIfNotInstalled "Pester" "5.3.3" | Import-Module
+
+Describe "Generate And Build SDK" -Tag "pipeline" {
+    BeforeAll {
+        # . $PSScriptRoot/../Invoke-GenerateAndBuildV2.ps1
+        # clean
+        $sdkRootPath =  (Join-Path $PSScriptRoot .. .. ..)
+        $sdkFolder = (Join-Path $sdkRootPath "sdk" "deviceupdate" "Azure.IoT.DeviceUpdate")
+        if (Test-Path $sdkFolder) {
```
benbp, line 12: Is this removing files that are already checked in? If so, maybe it would be better to copy them into a test directory and test against that target instead?

eng/scripts/tests/Invoke-GenerateAndBuildV2.tests.ps1
```diff
@@ -0,0 +1,52 @@
+#Requires -Version 7.0
```
benbp, line 1: It would be helpful to add a comment here or in a readme with a quick note about how to run these, e.g. https://github.com/Azure/azure-sdk-tools/blob/b6f98d351ceac565eeb2cbe23f762e123f120ef6/tools/asset-sync/contributing.md#using-pester

eng/scripts/automation/test/GenerateAndBuildLib-functions.tests.ps1
```diff
@@ -0,0 +1,109 @@
...
+        CreateOrUpdateAutorestConfigFile -autorestFilePath $PSScriptRoot/autorest.md -namespace "Azure.Purview.Account" -inputfile "$inputfile"
+
+        $autorestConfigYaml = Get-Content -Path $PSScriptRoot/autorest.md
+        $range = ($autorestConfigYaml | Select-String -Pattern '```').LineNumber
+        if ( $range.count -gt 1) {
+            $startNum = $range[0];
+            $lines = $range[1] - $range[0] - 1
+            $autorestConfigYaml = ($autorestConfigYaml | Select -Skip $startNum | Select -First $lines) |Out-String
+        }
+        Install-Module -Name powershell-yaml -Force -Verbose -Scope CurrentUser
```
weshaggard, line 35: Should we do this Install-Module/Import once at the top of the script?

eng/scripts/automation/test/GenerateAndBuildLib-functions.tests.ps1
```diff
@@ -0,0 +1,109 @@
...
+BeforeAll {
+    . $PSScriptRoot/../GenerateAndBuildLib.ps1
+}
+
+Describe "update autorest.md" -Tag "Unit" {
+    BeforeAll {
+        copy-item -Path $PSScriptRoot/autorest.test.md -Destination $PSScriptRoot/autorest.md
+    }
+
+    it("set mutilple input-file") {
```
weshaggard, line 16:
```suggestion
    it("set multiple input-file") {
```

eng/scripts/automation/unified-pipeline-test.md
weshaggard: Thanks for adding the tests and adding in the scenarios here. Can you also add some text in this MD that explains how to run the tests you created?

eng/scripts/automation/unified-pipeline-test.md
```diff
@@ -0,0 +1,186 @@
...
+
+## First Onboard
+For DPG, autorestConfig will be provided in input.json as following:
+
+input.json
+
+```json
```
weshaggard, line 8: Do these blocks exactly match the json test files if so perhaps you should just include a link to those files instead of duplicating the content that would need to be updated twice.

eng/scripts/automation/test/GenerateAndBuildLib-functions.tests.ps1
benbp: Also the underlying script imports it as well, so it may not be necessary?

eng/scripts/tests/updateExistingOutput-expected.json
```diff
@@ -0,0 +1,25 @@
...
+      },
+      "result": "succeeded",
+      "artifacts": [
+        ".\\artifacts\\packages\\Debug\\Azure.IoT.DeviceUpdate\\Azure.IoT.DeviceUpdate.1.0.0-alpha.20220726.1.nupkg"
+      ],
+      "changelog": {
+        "content": null,
+        "hasBreakingChange": false
+      },
+      "packageFolder": "C:/project/azure-sdk-for-net/sdk/deviceupdate/Azure.IoT.DeviceUpdate"
```
weshaggard, line 22: This path looks specific to your computer and will likely fail if ran on another.

eng/scripts/tests/Invoke-GenerateAndBuildV2.tests.ps1
weshaggard, line 1: Should these files be under eng/automation/test instead? It would be nice to keep them all in the same directory.

weshaggard (approved):
I've left a few more comments but nothing blocking. Once you finish your testing lets get these checked in so that we can start seeing them work.

benbp (approved):

benbp:
Some comments and I agree with @weshaggard's comments as well. +1 thanks for adding tests!

eng/scripts/tests/Invoke-GenerateAndBuildV2.tests.ps1
chunyu3: Yes, the output file is not used in below tests, so it is not needed.

eng/scripts/tests/Invoke-GenerateAndBuildV2.tests.ps1
chunyu3: In the output json there are some values varying every time, such as the 'artifacts', the name artifacts will contain the generate-date and it will vary every time, So I cannot just compare the whole object or text.

eng/scripts/automation/test/GenerateAndBuildLib-functions.tests.ps1
chunyu3: updated

eng/scripts/tests/Invoke-GenerateAndBuildV2.tests.ps1
chunyu3: Thanks, but I think it is better to put the test the same place with the script itself, then we can easy find out the test script.

eng/scripts/tests/Invoke-GenerateAndBuildV2.tests.ps1
chunyu3: Thanks @benbp . It is better not to remove the checkedin files. But we cannot copy them to a separate test directory, because all the sdk directory is under sdk\{serviceDirectory}, and the {serviceDiretory} is retried from the readme.md path. And we will use the {serviceDirectory} to get the swagger files in azure-rest-api-specs repo, so we cannot mock the readme.md path.

eng/scripts/tests/Invoke-GenerateAndBuildV2.tests.ps1
chunyu3: Add comment ahead of the test scripts.

eng/scripts/automation/unified-pipeline-test.md
chunyu3: The script only accepts json file as input, and in the input it will only can contain one json. so we cannot merge all those test json input as one and include link to it. 
Yes, it is not good to duplicate them, so I will remove unified-pipeline-test.md. thanks

eng/scripts/tests/updateExistingOutput-expected.json
chunyu3: the packageFolder is according to the running environment and the script will compute the path out. So all the outputJson files **-exptected.json may not exactly the same as the output of script. Those files are just for reference. I will update the files name to **-sample.json

PR #28737, opened 2022-05-13
Add IgnoreServiceErrorAttribute to Azure.Core.TestFramework

heaths:
Can be used for intermittent service errors. Ported loosely from sdk/keyvault.


sdk/core/Azure.Core.TestFramework/src/RecordedTestBaseOfTEnvironment.cs
```diff
@@ -7,6 +7,7 @@
 
 namespace Azure.Core.TestFramework
 {
+    [NonParallelizable]
```
JoshLove-msft, line 10: This can probably go on RecordedTestBase.

heaths: If the problem you mentioned is actually because of event trace testing, shouldn't we put this on ClientTestBase?

JoshLove-msft: No, I don't think so since there are unit tests that derive from ClientTestBase.

sdk/core/Azure.Core.TestFramework/src/IgnoreServiceErrorAttribute.cs
```diff
@@ -0,0 +1,55 @@
...
+        private readonly string _errorCodeMessage;
+
+        /// <summary>
+        /// Creates an instance of the <see cref="IgnoreServiceErrorAttribute"/> class.
+        /// </summary>
+        /// <param name="status">The HTTP status code to ignore e.g., 400.</param>
+        /// <param name="errorCode">The Azure error code to ignore e.g., BadParameter.</param>
+        public IgnoreServiceErrorAttribute(int status, string errorCode)
+        {
+            _statusMessage = $"Status: {status.ToString(CultureInfo.InvariantCulture)}";
```
JoshLove-msft, line 30: Should we throw if status or errorCode are null/empty?

sdk/core/Azure.Core.TestFramework/src/RecordedTestBaseOfTEnvironment.cs
heaths: Good point. I even do that. 🤦‍♂️

sdk/core/Azure.Core.TestFramework/src/IgnoreServiceErrorAttribute.cs
heaths: Could. I figured for an internal-ish attribute not necessary, but it wouldn't hurt.

JoshLove-msft:
Would you mind documenting this in the TestFramework readme?

JoshLove-msft (approved):
This is very cool!

sdk/core/Azure.Core.TestFramework/src/IgnoreServiceErrorAttribute.cs
```diff
@@ -0,0 +1,55 @@
...
+    /// <summary>
+    /// Marks tests failing with specific service error codes as "inconclusive".
+    /// </summary>
+    /// <remarks>
+    /// This is intended for use with intermittent service issues that happen only during internal testing,
+    /// or that are temporary issues causing test failures but will be fixed. Please engage with service teams
+    /// before using this attribute to see if live test issues can be resolved first.
+    /// </remarks>
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, AllowMultiple = true, Inherited = true)]
+    public class IgnoreServiceErrorAttribute : Attribute
```
JoshLove-msft, line 18: Should probably document that this only applies to RecordedTest tests. If possible, we should throw if used on a test without RecordedTest.

heaths: I'll document it, but it's (mostly: I did add a simple match method) just a data-only attribute. You can't really control to what it applies anyway, though there's thing I could do to make sure if it gets used it's base is RecordedTestBase, but not sure that's needed right now.

heaths: How this?

```     /// <summary>
    /// Marks tests failing with specific service error codes as "inconclusive". Must be used with test fixtures derived
    /// from <see cref="RecordedTestBase"/> and is used only by tests attributed with <see cref="RecordedTestAttribute"/>.
    /// </summary>
```

JoshLove-msft: I would think you could check it in RecordedTestBase Setup method and if RecordedTest is not there, then throw.
We do something similar in ClientTestBase - https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core.TestFramework/src/ClientTestBase.cs#L58-L59

JoshLove-msft: > How this?
> 
> ```
>     /// Marks tests failing with specific service error codes as "inconclusive". Must be used with test fixtures derived
>     /// from <see cref="RecordedTestBase"/> and is used only by tests attributed with <see cref="RecordedTestAttribute"/>.
>     /// </summary>
> ```

That works.

heaths:
I'm not repro'ing the exception locally. 🤔 Almost seems as if the attribute isn't encoded (or decoded at runtime) on the agents, but there's nothing special about the attribute.

sdk/core/Azure.Core.TestFramework/src/IgnoreServiceErrorAttribute.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// <summary>
+        /// Determines whether the <paramref name="message"/> matches the current attribute.
+        /// </summary>
+        /// <param name="message">The message to match.</param>
+        /// <returns>True if the <paramref name="message"/> matches the current attribute; otherwise, false.</returns>
+        internal protected bool Matches(string message) =>
+            !string.IsNullOrEmpty(message) &&
+            message.Contains(_statusMessage) &&
+            message.Contains(_errorCodeMessage) &&
+            string.IsNullOrWhiteSpace(Message) || message.Contains(Message);
```
JoshLove-msft, line 60: I think it's this or condition that is causing the error since it is not wrapped in parens.

JoshLove-msft, line 60:
```suggestion
            (string.IsNullOrWhiteSpace(Message) || message.Contains(Message));
```

JoshLove-msft (approved):
LGTM!

heaths:
Code within sdk/containerapps/Azure.ResourceManager.Applications.Containers is not up to date with code generation changes. I propose checking this in anyway, and do a separate clean-up PR. @jsquire?

PR #28738, opened 2022-05-13
Fix some naming issues on compute

### LINKED ISSUE #26391: Release stable version Azure.ResourceManager.Compute SDK

ArcturusZhang:


archerzz:
Why do we use `x-ms-client-name` for some renaming, but use `rename-model` for others?

ArcturusZhang:
> Why do we use `x-ms-client-name` for some renaming, but use `rename-model` for others?

1 . `rename-model` does not work for all cases.
2. In some cases, `rename-model` could work. But after renaming that model, I also need to change the name of some properties inside the same model, apparently using `x-ms-client-name` is much more simpler

sdk/compute/Azure.ResourceManager.Compute/src/autorest.md
```diff
@@ -82,15 +80,27 @@ directive:
...
-  - from: disk.json
-    where: $.definitions.PurchasePlan
-    transform: $["x-ms-client-name"] = "DiskPurchasePlan"
-  - from: swagger-document
-    where: $.definitions.VirtualMachineReimageParameters
-    transform: $["x-ms-client-name"] = "VirtualMachineReimageOptions"
-  - from: swagger-document
-    where: $.definitions.VirtualMachineScaleSetVMReimageParameters
-    transform: $["x-ms-client-name"] = "VirtualMachineScaleSetVmReimageOptions"
+  - rename-model:
```
ArthurMa1978, line 83: Consider to put `rename-model` at last because it changes the model name, then if one directiy after the rename which want to change something of the properties in this model, need to use the new name instead of the name in swagger spec. This will bring extra trouble if not aware the model name is renamed.

sdk/compute/Azure.ResourceManager.Compute/src/autorest.md
```diff
@@ -101,16 +111,59 @@ directive:
...
+      $.VirtualMachineImageProperties.properties.dataDiskImages.description = "The list of data disk images information.";
+      $.VirtualMachineInstallPatchesParameters.properties.maximumDuration["format"] = "duration";
+      $.VirtualMachineExtensionUpdateProperties.properties.type["x-ms-client-name"] = "ExtensionType";
+      $.VirtualMachineExtensionProperties.properties.type["x-ms-client-name"] = "ExtensionType";
+      $.VirtualMachineScaleSetExtensionProperties.properties.type["x-ms-client-name"] = "ExtensionType";
+      $.VirtualMachineScaleSetExtension.properties.type["x-ms-format"] = "resource-type";
+      $.VirtualMachineNetworkInterfaceIPConfigurationProperties.properties.privateIPAddressVersion["x-ms-enum"].name = "IPVersion";
+      $.VirtualMachinePublicIPAddressConfigurationProperties.properties.publicIPAddressVersion["x-ms-enum"].name = "IPVersion";
+      $.SubResource["x-ms-client-name"] = "ComputeWriteableSubResourceData";
+      $.SubResource.properties.id["x-ms-format"] = "arm-id";
```
ArthurMa1978, line 131: rename the `id`?

ArthurMa1978 (approved):

sdk/compute/Azure.ResourceManager.Compute/src/autorest.md
ArcturusZhang: Everything works here. I believe modeler will execute the rename-model directive at last

sdk/compute/Azure.ResourceManager.Compute/src/autorest.md
ArcturusZhang: No, we cannot rename the id property because this is designed to use for various resource references, including public ip, network interface, subnet, restorePointGroup, etc...
The reason why this model still is here, is that some model is inheriting from it. Therefore maybe "Sub"Resource is not a great name here.
But we could see what arch board will say about this...

PR #28739, opened 2022-05-13
Prepare release for storage

ArcturusZhang:

### DIFF
--- a/sdk/storage/Azure.ResourceManager.Storage/CHANGELOG.md
+++ b/sdk/storage/Azure.ResourceManager.Storage/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 - Flattened property from a read-only model no longer has setters.
 - The type of flattened primitive property changed to its corresponding nullable type.
+- Renamed class `PrivateLinkResource` to `StoragePrivateLinkResource`.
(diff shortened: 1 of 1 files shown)


ArthurMa1978 (approved):

sdk/storage/Azure.ResourceManager.Storage/CHANGELOG.md
```diff
@@ -1,14 +1,11 @@
 # Release History
 
-## 1.0.0-beta.9 (Unreleased)
-
-### Features Added
+## 1.0.0-beta.9 (2022-05-13)
```
Yao725, line 3: I think we might miss two things here:

- Add the `RP` prefix for the private link related resources and models
- Add the update method using put if this resource doesn't have a patch method

ArcturusZhang: Thank you! Let me add them

ArcturusZhang: OK now in storage it is called StoragePrivateLinkResource, but it is not inheriting from ArmResource which violates our rules.
But considering this is a beta release, I will let it slide.

Yao725 (approved):

PR #28768, opened 2022-05-13
Update CODEOWNERS

jsquire:
# Summary

The focus of these changes is to capture recent organizational updates for ownership

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -6,14 +6,14 @@
 ################
 
 # Git Hub integration and bot rules
-/.github/                                       @AlexGhiondea @jsquire
+/.github/                                       @jsquire
 
 #########
 # SDK
 #########
 
 # Catch all
-/sdk/                                           @AlexGhiondea
+/sdk/                                           
 
 # ######## Core Libraries ########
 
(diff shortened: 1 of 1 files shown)


christothes (approved):

.github/CODEOWNERS
```diff
@@ -6,22 +6,22 @@
...
-/.github/                                       @AlexGhiondea @jsquire
+/.github/                                       @jsquire
 
 #########
 # SDK
 #########
 
 # Catch all
-/sdk/                                           @AlexGhiondea
+/sdk/
```
weshaggard, line 16: @jsquire in some places we depend on this fallback having an owner (example docs or pipeline ownership) for any paths that have a mapping. Can we please add you and/or Pallavi to this set of owners?

pallavit: https://github.com/Azure/azure-sdk-for-net/pull/29166

jsquire: @weshaggard - Can we confirm behavior to be sure that my mental model is correct?   My understanding is that this will essentially end up sending alerts for anything in the repository that doesn't have a more specific path.   Assuming that I'm correct, I'm not comfortable having myself here _(I'll let Pallavi speak for herself)_

I'm still actively working in parts of the repository and rely on GH notifications via email for these.   The last I spoke with Alex, he received enough spam from the catch-all notifications that all GH emails were noise to him that were dealt with this by filtering into an "ignore" area.   That's not a situation that's going to work for me.

Can we define what areas, specifically, need the leads to watch and add explicit ownership for those areas?   Alternately, lets consider creating a GH handle and DG that Pallavi and I can subscribe to which I can use for filtering notifications for catch-all from those that I explicitly opt into.

PR #28771, opened 2022-05-13
Bind to PartitionContext in batch test

JoshLove-msft:


sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/tests/EventHubEndToEndTests.cs
```diff
@@ -572,6 +572,9 @@ public static void SendEvents_TestHub(int numEvents, string input, [EventHub(Tes
                     Assert.AreEqual(s_processedEventCount++, propertiesArray[i]["TestIndex"]);
                 }
 
+                Assert.NotNull(partitionContext.PartitionId);
+                Assert.NotNull(partitionContext.ReadLastEnqueuedEventProperties());
```
jsquire, line 576: `LastEnqueuedEventProperties` is a `struct`, so never null.    Maybe compare to `default`?

jsquire (approved):

Issue #28326, opened 2022-04-21
TextAnalytics - Clean up TextAnalyticsOperationStatus before release


nisha-bhatia:
https://github.com/Azure/azure-sdk-for-net/pull/28225#discussion_r854530317

Need to delete Rejected field on TextAnalyticsOperationStatus before next release

PR #28772, opened 2022-05-13
[TA] Implement Analyze Actions

### LINKED ISSUE #28330: Hook batch endpoints from TextAnalyticsClient to use the new Language API endpoints

maririos:
Includes:
- Legacy and Language implemented. 1/2 of https://github.com/Azure/azure-sdk-for-net/issues/28330
- Live tests with recordings 1/2 of https://github.com/Azure/azure-sdk-for-net/issues/28332
- Legacy mock tests
- Samples
- Fixes https://github.com/Azure/azure-sdk-for-net/issues/28330
Missing:
- Custom mock tests
- Pagination work on Language implementation

### DIFF
--- a/sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsModelFactory.cs
+++ b/sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsModelFactory.cs
@@ -271,8 +271,7 @@ namespace Azure.AI.TextAnalytics
         [EditorBrowsable(EditorBrowsableState.Never)]
         public static CategorizedEntity CategorizedEntity(string text, string category, string subCategory, double score)
         {
-            //return new CategorizedEntity(new Entity(text, category, subCategory, default, default, score));
-            throw new NotImplementedException();
+            return new CategorizedEntity(new Entity(text, category, subCategory, default, default, score));
         }
 
         /// <summary>
@@ -287,8 +286,7 @@ namespace Azure.AI.TextAnalytics
         /// <returns>A new instance of <see cref="TextAnalytics.CategorizedEntity"/> for mocking purposes.</returns>
         public static CategorizedEntity CategorizedEntity(string text, string category, string subCategory, double score, int offset, int length)
         {
-            //return new CategorizedEntity(new Entity(text, category, subCategory, offset, length, score));
-            throw new NotImplementedException();
+            return new CategorizedEntity(new Entity(text, category, subCategory, offset, length, score));
         }
 
         /// <summary>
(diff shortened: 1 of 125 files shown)


sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsOperationStatus.cs
```diff
@@ -11,8 +11,7 @@ namespace Azure.AI.TextAnalytics
     [CodeGenModel("State")]
     public partial struct TextAnalyticsOperationStatus
     {
-        /// <summary> Added for compilation. </summary>
-        /// TODO: delete later
-        public static TextAnalyticsOperationStatus Rejected { get; } = new TextAnalyticsOperationStatus(CancellingValue);
+        /// <summary> Rejected </summary>
+        public static TextAnalyticsOperationStatus Rejected { get; } = new TextAnalyticsOperationStatus("Rejected");
```
maririos, line 15: Issue https://github.com/Azure/azure-sdk-for-net/issues/28326
After consideration, decided to add "Rejected" state as this is an extensible enum. This is not sent from the service anymore and after talking to them, they didn't send them before either. 
Even with that, we can't delete it because a customer might have it in their code which will cause a build break.
So, let's leave it as is with the corresponding value

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsModelFactory.cs
```diff
@@ -271,8 +271,7 @@ public static DetectLanguageResultCollection DetectLanguageResultCollection(IEnu
         [EditorBrowsable(EditorBrowsableState.Never)]
         public static CategorizedEntity CategorizedEntity(string text, string category, string subCategory, double score)
         {
-            //return new CategorizedEntity(new Entity(text, category, subCategory, default, default, score));
-            throw new NotImplementedException();
```
maririos, line 275: most changes in this class are to remove comments and enable the method again

sdk/textanalytics/Azure.AI.TextAnalytics/tests/AnalyzeOperationTests.cs
```diff
@@ -423,6 +420,7 @@ public async Task AnalyzeOperationWithMultipleActionsOfSameType()
         }
 
         [RecordedTest]
+        [ServiceVersion(Max = TextAnalyticsClientOptions.ServiceVersion.V3_2_Preview_2)]
```
maririos, line 423: this is telling the tests, to run it in the Legacy side and not Language (which version is above 3.2-preview.2

sdk/textanalytics/Azure.AI.TextAnalytics/tests/MockTests/Legacy/TextAnalyticsClientMockTests.cs
```diff
@@ -289,53 +288,53 @@ public async Task DetectedLanguageNullIso6391Name()
         // We shipped TA 5.0.0 Text == string.Empty if the service returned a null value for Text.
         // We want to verify behavior is the same after code auto generated.
         [Test]
+        [Ignore("Figure out why it fails")]
```
maririos, line 291: By enabling this class, I noticed some tests were failing.
We can take a look at this in a future PR.

Issue => https://github.com/Azure/azure-sdk-for-net/issues/28774

jsquire (approved):

PR #28786, opened 2022-05-14
Abandon messages when an exception occurs in batch function regardless of autocomplete setting

JoshLove-msft:
Previously, we were only abandoning messages in batch functions if an error occurred _and_ AutoCompleteMessages was true. This was inconsistent with the behavior for single-dispatch functions, and with the behavior of the SDK when using the processor directly. The correct behavior is for abandoning to be independent of AutoCompleteMessages.

While working on this fix, an additional two issues were found and fixed:
- Messages received from ReceiveActions were not being abandoned when there was an exception
- Message lock renewal was not being stopped when user handler throws

### DIFF
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ReceiverManager.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ReceiverManager.cs
@@ -140,39 +140,32 @@ namespace Azure.Messaging.ServiceBus
             CancellationToken cancellationToken)
         {
             ServiceBusErrorSource errorSource = ServiceBusErrorSource.Receive;
-            EventArgs args = null;
+            EventArgs args = ConstructEventArgs(triggerMessage, cancellationToken);
 
             try
             {
                 errorSource = ServiceBusErrorSource.ProcessMessageCallback;
                 try
                 {
-                    ServiceBusEventSource.Log.ProcessorMessageHandlerStart(Processor.Identifier, triggerMessage.SequenceNumber,
-                        triggerMessage.LockTokenGuid);
-                    args = await OnMessageHandler(triggerMessage, cancellationToken).ConfigureAwait(false);
-                    ServiceBusEventSource.Log.ProcessorMessageHandlerComplete(Processor.Identifier, triggerMessage.SequenceNumber,
-                        triggerMessage.LockTokenGuid);
+                    ServiceBusEventSource.Log.ProcessorMessageHandlerStart(Processor.Identifier, triggerMessage.SequenceNumber, triggerMessage.LockTokenGuid);
+                    await OnMessageHandler(args).ConfigureAwait(false);
+                    ServiceBusEventSource.Log.ProcessorMessageHandlerComplete(Processor.Identifier, triggerMessage.SequenceNumber, triggerMessage.LockTokenGuid);
                 }
                 catch (Exception ex)
                 {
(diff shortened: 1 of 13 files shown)


sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/README.md
```diff
@@ -274,7 +274,9 @@ public static async Task Run(
 
 ## Troubleshooting
 
-Please refer to [Monitor Azure Functions](https://docs.microsoft.com/azure/azure-functions/functions-monitoring) for troubleshooting guidance.
+If your function triggers an unhandled exception, and you haven't already settled the message, the extension will attempt to abandon the message so that it becomes available for receiving again immediately.
```
jsquire, line 277:
```suggestion
If your function triggers an unhandled exception and you haven't already settled the message, the extension will attempt to abandon the message so that it becomes available for receiving again immediately.
```

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusListener.cs
```diff
@@ -403,42 +403,43 @@ private async Task RunBatchReceiveLoopAsync(CancellationToken cancellationToken)
...
-                                    abandonTasks.Add(receiver.AbandonMessageAsync(message, cancellationToken: CancellationToken.None));
+                                    continue;
                                 }
 
-                                await Task.WhenAll(abandonTasks).ConfigureAwait(false);
+                                // Pass CancellationToken.None to allow abandon to finish even when shutting down
+                                abandonTasks.Add(receiver.AbandonMessageAsync(message, cancellationToken: CancellationToken.None));
                             }
+
+                            await Task.WhenAll(abandonTasks).ConfigureAwait(false);
```
jsquire, line 442: In the off chance that one of the abandon calls throws and sessions are enabled, would this cause an issue because the block to release the session on L445 will get skipped?   Should we consider wrapping this in a `try/catch` ?

jsquire (approved):

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusListener.cs
JoshLove-msft: It shouldn't matter because we end up closing the receiver anyway in the fall through catch block.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ReceiverManager.cs
```diff
@@ -239,24 +235,25 @@ protected async Task ProcessOneMessageWithinScopeAsync(ServiceBusReceivedMessage
...
             return !Receiver.IsSessionReceiver &&
                    Receiver.ReceiveMode == ServiceBusReceiveMode.PeekLock &&
                    AutoRenewLock;
         }
 
-        protected virtual async Task<EventArgs> OnMessageHandler(
-            ServiceBusReceivedMessage message,
-            CancellationToken processorCancellationToken)
-        {
-            var args = new ProcessMessageEventArgs(
```
JoshLove-msft, line 253: Previously the args would not be returned to the caller in case of an exception, which broke the finally block that cancels lock renewal.

PR #28787, opened 2022-05-14
[TA] Samples fixed to latest version

maririos:
Now that all endpoints are enabled, we can make samples run on the default service version (which is latest).


jsquire (approved):
dang, that's a lot of files!

PR #28788, opened 2022-05-14
[TA] Fixes in legacy code

### LINKED ISSUE #28774: Legacy mock tests are failing
Why? they shouldn't fail. 
Tests are:
- AnalyzeSentimentNullText
- AnalyzeSentimentNotSupportedSentenceSentiment
- AnalyzeSentimentMixedSentenceSentiment
- RecognizeLinkedEntitiesNullText

maririos:
Fixes https://github.com/Azure/azure-sdk-for-net/issues/28774

### DIFF
--- a/sdk/textanalytics/Azure.AI.TextAnalytics.Legacy.Shared/src/Custom/Match.cs
+++ b/sdk/textanalytics/Azure.AI.TextAnalytics.Legacy.Shared/src/Custom/Match.cs
@@ -1,0 +1,21 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Azure.AI.TextAnalytics.Legacy
+{
+    /// <summary> The Match. </summary>
+    internal partial class Match
+    {
+        internal Match(double confidenceScore, string text, int offset, int length)
+        {
+            // We shipped TA 5.0.0 Text == string.Empty if the service returned a null value for Text.
+            // Because we don't want to introduce a breaking change, we are transforming that null to string.Empty
+            Text = text ?? string.Empty;
+
+            ConfidenceScore = confidenceScore;
+            Text = text;
+            Offset = offset;
+            Length = length;
+        }
+    }
+}
--- a/sdk/textanalytics/Azure.AI.TextAnalytics/tests/MockTests/Legacy/TextAnalyticsClientMockTests.cs
+++ b/sdk/textanalytics/Azure.AI.TextAnalytics/tests/MockTests/Legacy/TextAnalyticsClientMockTests.cs
@@ -288,7 +288,6 @@ namespace Azure.AI.TextAnalytics.Legacy.Tests
         // We shipped TA 5.0.0 Text == string.Empty if the service returned a null value for Text.
         // We want to verify behavior is the same after code auto generated.
         [Test]
-        [Ignore("Figure out why it fails")]
         public async Task AnalyzeSentimentNullText()
         {
             using var stream = new MemoryStream(Encoding.UTF8.GetBytes(@"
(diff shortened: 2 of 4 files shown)


sdk/textanalytics/Azure.AI.TextAnalytics.Legacy.Shared/src/Custom/Match.cs
```diff
@@ -0,0 +1,21 @@
...
+// Licensed under the MIT License.
+
+namespace Azure.AI.TextAnalytics.Legacy
+{
+    /// <summary> The Match. </summary>
+    internal partial class Match
+    {
+        internal Match(double confidenceScore, string text, int offset, int length)
+        {
+            // We shipped TA 5.0.0 Text == string.Empty if the service returned a null value for Text.
```
maririos, line 11: I had to bring this class back up again because of this

sdk/textanalytics/Azure.AI.TextAnalytics/tests/MockTests/Legacy/TextAnalyticsClientMockTests.cs
```diff
@@ -333,96 +332,6 @@ public async Task AnalyzeSentimentNullText()
...
-                            ""id"": ""1"",
-                            ""sentiment"": ""neutral"",
-                            ""confidenceScores"": {
-                                ""positive"": 0.1,
-                                ""neutral"": 0.88,
-                                ""negative"": 0.02
-                            },
-                            ""sentences"": [
-                                {
-                                ""sentiment"": ""confusion"",
```
maririos, line 353: This is an interesting case. In the legacy code, `SentenceSentiment.Sentiment` (previously called `SentenceSentimentInternal`) is of type `string`. We manually modified it [here](https://github.com/Azure/azure-sdk-for-net/blob/Azure.AI.TextAnalytics_5.2.0-beta.2/sdk/textanalytics/Azure.AI.TextAnalytics/src/Internal/SentenceSentimentInternal.cs#L15). 

I don't remember why, but it seems associated to [this line of code](https://github.com/Azure/azure-sdk-for-net/blob/Azure.AI.TextAnalytics_5.2.0-beta.2/sdk/textanalytics/Azure.AI.TextAnalytics/src/SentenceSentiment.cs#L37):
```
Sentiment = (TextSentiment)Enum.Parse(typeof(TextSentiment), sentenceSentiment.Sentiment, ignoreCase: true);
```

I think we wanted to just have one enum for the sentiments and we thought it was easier from string than doing a conversion.

This test is verifying that if the service sends a value that is not valid, our parsing `enum` throws an `ArgumentException`.

With the [new implementation](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/textanalytics/Azure.AI.TextAnalytics.Legacy.Shared/src/Generated/Models/SentenceSentiment.cs#L67), where we leave the `SentenceSentiment.Sentiment` as `SentenceSentimentValue` our deserialization [code fails](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/textanalytics/Azure.AI.TextAnalytics.Legacy.Shared/src/Generated/Models/SentenceSentimentValue.Serialization.cs#L19) as `confusion` is not a valid value.

Given that this is a change in behavior, the question is, what do we do?

What I am proposing is to delete the tests altogether and leave behavior as is. This is a validation we are doing in case the service fails (which we haven't seen), but we also introduced it by making the type a string. If we follow strictly what the REST API says and what we've seen from the service behavior, this is a very very unlikely scenario.

If we want to go back to the previous behavior, I will go ahead and make the modifications in the Legacy.Shared project

@jsquire what do you think?

sdk/textanalytics/Azure.AI.TextAnalytics/tests/MockTests/Legacy/TextAnalyticsClientMockTests.cs
```diff
@@ -333,96 +332,6 @@ public async Task AnalyzeSentimentNullText()
...
-
-            var mockTransport = new MockTransport(new[] { mockResponse });
-            var client = CreateTestClient(mockTransport);
-
-            Assert.ThrowsAsync<ArgumentException>(async () => await client.AnalyzeSentimentAsync("today is a hot day"));
-        }
-
-        [Test]
-        [Ignore("Figure out why it fails")]
-        public async Task AnalyzeSentimentMixedSentenceSentiment()
```
maririos, line 382: Same case about string vs enum

sdk/textanalytics/Azure.AI.TextAnalytics/tests/MockTests/Legacy/TextAnalyticsClientMockTests.cs
jsquire: Given that the legacy service is stable and not under active development, if an unknown value gets returned then it would be a service problem.   Since this wasn't an extensible enum to start with, my assumption is that the service is expected to return only one of the known set for that service version.

Based on that assumption, I agree with your assessment.   Having the enum directly in the generated model and allowing it to validate the service return makes sense.   Killing the test makes sense to me.

jsquire (approved):

Issue #28734, opened 2022-05-12
[FEATURE REQ]


coco1080:
### Library name

AzurePowershellCredentials

### Please describe the feature.

**AzurePowershellCredentials** library should be enhanced in a way the client code can configure timeout parameters depending on their needs and scenarios when generating token requests. The client can experience several transient timeout issues when requesting tokens by not having that option, which is the scenario we are experiencing. (Please see stack trace below for more details)

These are the timeout fixed values that are encapsulated in the following libraries:
 (AzurePowershellCredential 10s, AzureCLICredential 13s, VisualStudioCredential 30s) 

Stack Trace:
   at Azure.Identity.AzurePowerShellCredential.<RequestAzurePowerShellAccessTokenAsync>d__26.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Azure.Identity.AzurePowerShellCredential.<GetTokenImplAsync>d__25.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at Azure.Identity.CredentialDiagnosticScope.FailWrapAndThrow(Exception ex, String additionalMessage)
   at Azure.Identity.AzurePowerShellCredential.<GetTokenImplAsync>d__25.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Azure.Identity.AzurePowerShellCredential.GetToken(TokenRequestContext requestContext, CancellationToken cancellationToken)
   at Azure.Identity.DefaultAzureCredential.<GetTokenFromSourcesAsync>d__15.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Azure.Identity.DefaultAzureCredential.<GetTokenImplAsync>d__13.MoveNext()
[...]

Issue #23135, opened 2021-08-05
Add configurable process timeout for dev-time credentials


schaabs:
We still have reports of credentials which spawn subprocesses, specifically `AzurePowerShellCredential` and `AzureCliCredential`, still timing out in some environments despite having increased the timeouts. From a user account of debugging such an issue.

> Also, I copied the AzurePowerShellCredential code to my project and played with the timeout value. When the timeout is reduced from 10 seconds to 6 seconds, then the timeout always happens. When the timeout is increased from 10 seconds to 20 seconds, then the timeout never happens. This leads me to believe that there is no sinister race condition bug or a deadlock bug or something like that. It's just that PowerShell (or Azure CLI) starts up slow.

In order to make these credentials more resilient in various environments we should make this timeout configurable and possibly consider extending the default timeout a bit more.

dougdenn:
Excited to see #28792 merged and looking forward to trying the new option out. I'm facing timeouts when trying to connect to KeyVault and I think this will really help. Thanks.

scottaddie:
Retitled issue to align with other languages. Using this as a placeholder for the required archboard discussion.

joshfree:
@scottaddie would you mind filing the issues for the other languages and making sure they're added to the github project?

scottaddie:
This feature shipped in v1.9.0. See the `ProcessTimeout` property in the credential options bag for the following dev tool credential types:

- `AzureCliCredential`
- `AzureDeveloperCliCredential`
- `AzurePowerShellCredential`
- `VisualStudioCredential`

For example:

https://github.com/Azure/azure-sdk-for-net/blob/3294e7ee85814145bb30aadd42cb682d86936020/sdk/identity/Azure.Identity/src/Credentials/AzureCliCredentialOptions.cs#L29

Additionally, there's a `CredentialProcessTimeout` property in the `DefaultAzureCredential` options bag, which applies to all of the aforementioned dev tool credential types.

https://github.com/Azure/azure-sdk-for-net/blob/3294e7ee85814145bb30aadd42cb682d86936020/sdk/identity/Azure.Identity/src/Credentials/DefaultAzureCredentialOptions.cs#L206

PR #28792, opened 2022-05-15
Adding configurable process timeout for DeveloperCredentials

reynaldoburgos:
Adding configurable process timeout for DeveloperCredentials:
1. Azure CLI
2. Azure Powershell
3. VisualStudio Credential

Proposed solution for [FEATURE REQ]
#23135
#28734

### DIFF
--- a/sdk/identity/Azure.Identity/api/Azure.Identity.netstandard2.0.cs
+++ b/sdk/identity/Azure.Identity/api/Azure.Identity.netstandard2.0.cs
@@ -140,7 +140,7 @@ namespace Azure.Identity
     public partial class DefaultAzureCredentialOptions : Azure.Identity.TokenCredentialOptions
     {
         public DefaultAzureCredentialOptions() { }
-        public System.TimeSpan DeveloperCredentialTimeout { get { throw null; } set { } }
+        public System.TimeSpan? DeveloperCredentialTimeout { get { throw null; } set { } }
         public bool ExcludeAzureCliCredential { get { throw null; } set { } }
         public bool ExcludeAzurePowerShellCredential { get { throw null; } set { } }
         public bool ExcludeEnvironmentCredential { get { throw null; } set { } }
--- a/sdk/identity/Azure.Identity/api/Azure.Identity.netstandard2.0.cs
+++ b/sdk/identity/Azure.Identity/api/Azure.Identity.netstandard2.0.cs
@@ -58,6 +58,7 @@ namespace Azure.Identity
     public partial class AzureCliCredentialOptions : Azure.Identity.TokenCredentialOptions
     {
         public AzureCliCredentialOptions() { }
+        public System.TimeSpan? CliProcessTimeout { get { throw null; } set { } }
         public string TenantId { get { throw null; } set { } }
     }
     public partial class AzurePowerShellCredential : Azure.Core.TokenCredential
@@ -70,7 +71,7 @@ namespace Azure.Identity
     public partial class AzurePowerShellCredentialOptions : Azure.Identity.TokenCredentialOptions
     {
         public AzurePowerShellCredentialOptions() { }
-        public int PowerShellProcessTimeoutMs { get { throw null; } set { } }
+        public System.TimeSpan? PowerShellProcessTimeout { get { throw null; } set { } }
         public string TenantId { get { throw null; } set { } }
     }
     public partial class ChainedTokenCredential : Azure.Core.TokenCredential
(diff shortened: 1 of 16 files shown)


sdk/identity/Azure.Identity/src/Credentials/DefaultAzureCredentialOptions.cs
```diff
@@ -67,6 +67,11 @@ public class DefaultAzureCredentialOptions : TokenCredentialOptions
         /// </summary>
         public ResourceIdentifier ManagedIdentityResourceId { get; set; }
 
+        /// <summary>
+        /// Specifies Powershell process timeout in milliseconds.
+        /// </summary>
+        public int PowerShellProcessTimeoutMs { get; set; } = (int)TimeSpan.FromSeconds(10).TotalMilliseconds;
```
christothes, line 73: I'm wondering if we should generalize this to work for all process based credentials (Visual Studio, and AzureCli). 

@schaabs any thoughts ?

reynaldoburgos: This sounds reasonable, you mean calling this something like "SubProcessTimeOutMs" and then sharing this timeout between all providers using subproceses ? this sounds simple enough, but also wondering if clients would ever have the need to control each provider separately with its own timeout. 

I'm inclined to do the simplest approach of sharing single SubProcessTimeout since will add value soon and not complicate a lot the API so that users can configure this. Then evaluate later if we need to have greater control over each provider timeout.

thoughts ?

reynaldoburgos: @christothes @schaabs ping

christothes: Sorry for the delay on this issue. I am thinking we make it `TimeSpan DeveloperCredentialTimeout` which refers to the "Developer Credentials in [this flow](https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/identity/Azure.Identity#defaultazurecredential). I'd also adjust the comments to name the specific credentials that are affected (all the orange ones, currently). 

cc @scottaddie

sdk/identity/Azure.Identity/api/Azure.Identity.netstandard2.0.cs
```diff
@@ -138,6 +140,7 @@ public partial class DefaultAzureCredential : Azure.Core.TokenCredential
     public partial class DefaultAzureCredentialOptions : Azure.Identity.TokenCredentialOptions
     {
         public DefaultAzureCredentialOptions() { }
+        public System.TimeSpan DeveloperCredentialTimeout { get { throw null; } set { } }
```
snakefoot, line 143: Why not nullable to match the others?

sdk/identity/Azure.Identity/src/Credentials/VisualStudioCredential.cs
```diff
@@ -120,7 +123,7 @@ private async Task<AccessToken> RunProcessesAsync(List<ProcessStartInfo> process
                 string output = string.Empty;
                 try
                 {
-                    using var processRunner = new ProcessRunner(_processService.Create(processStartInfo), TimeSpan.FromSeconds(30), _logPII, cancellationToken);
+                    using var processRunner = new ProcessRunner(_processService.Create(processStartInfo), VisualStudioProcessTimeout, _logPII, cancellationToken);
```
christothes, line 126: There is a hard coded error message below that says "30 seconds" - can we change that to use the timeout value? Let's validate this on the other credentials also.

sdk/identity/Azure.Identity/tests/AzureCliCredentialTests.cs
```diff
@@ -113,5 +113,13 @@ public void AuthenticateWithCliCredential_CanceledByUser()
             AzureCliCredential credential = InstrumentClient(new AzureCliCredential(CredentialPipeline.GetInstance(null), new TestProcessService(testProcess)));
             Assert.CatchAsync<OperationCanceledException>(async () => await credential.GetTokenAsync(new TokenRequestContext(MockScopes.Default), cts.Token));
         }
+
+        [Test]
+        public void ConfigureCliProcessTimeout()
```
christothes, line 118: This and the other tests should also validate that GetToken throws after being cancelled by the (short) timeout just to validate that the option actually applies to the `ProcessRunner`.

christothes (approved):
Looks good!

PR #28794, opened 2022-05-16
remove if statement in ci.yml

chunyu3:

Remove if-statement since we donot have pipeline runs for various engsys related PRs that rely on the template tests as our primary CI gate.

### DIFF
--- a/sdk/template/ci.yml
+++ b/sdk/template/ci.yml
@@ -9,12 +9,7 @@ trigger:
   paths:
     include:
     - sdk/template/
-#if (false)
-    # The following paths should only be included in template/ci.yml, and removed from any other SDKs which copy this file.
-    # The surrounding conditions should accomplish that when installed with `dotnet new azsdk`.
-    # eng/common code changes trigger template pipeline for basic checking.
     - eng/common/
-#endif
 
 pr:
   branches:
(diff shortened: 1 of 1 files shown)


sdk/template/ci.yml
```diff
@@ -9,12 +9,7 @@ trigger:
   paths:
     include:
     - sdk/template/
-#if (false)
```
sima-zhu, line 12: Don't remove the if block. It is useful comment section.

sima-zhu (dismissed):
Could you please add the trigger of "eng/common" under pr triggers as well. It is at #L23

sdk/template/ci.yml
weshaggard: In fact we need to add back the if block that was removed as part of your other PR. @chunyu3 is there some reason you are trying to remove these?

weshaggard: See my comment in your other PR https://github.com/Azure/azure-sdk-for-net/pull/28016#discussion_r852480070.

benbp (approved):

sdk/template/ci.yml
chunyu3: I updated according to the comment and add eng/common in pr-trigger the same as trigger does

Issue #28807, opened 2022-05-16
Generate pagination code with a Transformation


maririos:
For TextAnalytics REST API, we had a [transformation](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/textanalytics/Azure.AI.TextAnalytics.Legacy.Shared/src/autorest.md#add-x-ms-paths-section-if-not-exists) that was in charge of generating the Pagination code for both Healthcare and AnalyzeActions functionalities.

Applying a similar transformation against the Language REST API code fails. Because of time constraints, for the May release, we are adding the code manually under an Internal `MicrosoftCognitiveLanguageServiceRestClient`. Note that the code is the same that codegeneration is supposed to generate.

The goal of this issue is to be able to generate the pagination code with a transformation, instead of us maintaining it to take full advantage of codegen.

heaths:
@m-nash does this sound like anything we still need for TypeSpec codegen? Otherwise, we can probably just close this. /cc @pallavit (this is a code gen issue, so I've removed Service Attention).

PR #28808, opened 2022-05-16
[TA] Add Pagination support for healthcare and AnalyzeActions

### LINKED ISSUE #28777: Write pagination support for analyzeActions and Healthcare

maririos:
Fixes: https://github.com/Azure/azure-sdk-for-net/issues/28777 by manually adding the pagination code (previously generated by codegen).

Issue to look into generating the Pagination code with transformations is https://github.com/Azure/azure-sdk-for-net/issues/28807

### DIFF
--- a/sdk/textanalytics/Azure.AI.TextAnalytics/src/Internal/MicrosoftCognitiveLanguageServiceRestClient.cs
+++ b/sdk/textanalytics/Azure.AI.TextAnalytics/src/Internal/MicrosoftCognitiveLanguageServiceRestClient.cs
@@ -1,0 +1,84 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Text.Json;
+using System.Threading;
+using System.Threading.Tasks;
+using Azure.AI.TextAnalytics.Models;
+using Azure.Core;
+
+namespace Azure.AI.TextAnalytics
+{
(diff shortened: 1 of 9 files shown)


sdk/textanalytics/Azure.AI.TextAnalytics/src/Internal/MicrosoftCognitiveLanguageServiceRestClient.cs
```diff
@@ -0,0 +1,84 @@
...
+using System;
+using System.Text.Json;
+using System.Threading;
+using System.Threading.Tasks;
+using Azure.AI.TextAnalytics.Models;
+using Azure.Core;
+
+namespace Azure.AI.TextAnalytics
+{
+    internal partial class MicrosoftCognitiveLanguageServiceRestClient
```
jsquire, line 13: That doesn't look nearly as scary as I would have thought.

jsquire (approved):

sdk/textanalytics/Azure.AI.TextAnalytics/src/Internal/MicrosoftCognitiveLanguageServiceRestClient.cs
maririos: I just don't like to do manual things when our tooling can do it for us. The less we have to think and change, the better :)

Issue #28060, opened 2022-04-07
[Text Analytics] Design Error Experience for Unsupported Fields and Operations in Legacy Versions


jsquire:
# Summary

With the Text Analytics client supporting the legacy Text Analytics REST API and new Language REST API, there will be operations, analysis types, and model fields not supported/recognized by the legacy version.   A design is needed for a consistent developer experience and an understanding of when to allow the service to perform validation and when client-side validation is needed.

# Scope of Work

- Understand the API gaps between the REST APIs and test service behavior for common scenarios

- Analyze service behavior and identify how they can be used in a consistent and clear developer experience

- Propose a design for addressing gaps and unsupported members across languages

# Success Criteria

- The design has been discussed with the feature team and feedback has been incorporated

- The design has been documented appropriate for discussion with the architects

- Architect discussions have been scheduled

maririos:
PR #28809 introduces new properties only supported in the Language client which are:
- `FhirVersion` input property
- `FhirBundle` output property.

I followed what we have done in this library when a new property is added that is only supported in specific versions and does not affect behavior of older version output properties:
- We add that information to docstrings for both input and output property.
- Input property:
  - If the property is a query parameter, we send it to the service and let them manage it.
  - If the property is a body parameter, we drop it.
- Output property: Value is always empty Dict even when targeting older versions

If as a result of the work in this issue, the guidance is different, please modify `Fhir` behavior

heaths:
Just to clarify, the output property will be the default value unless a collection/dictionary, in which case it's an empty collection/dictionary.

heaths:
Python errs when trying to use newer properties against older versions using decorators. Similar idea with attributes for .NET could work here.

heaths:
Could also just throw exceptions in either the _Transform.cs_ or _Transform.Legacy.cs_ files for newly-added model properties not updated to the legacy swagger. It really only matters in these cases anyway.

PR #28809, opened 2022-05-16
[TA] Implementation to support FHIR

### LINKED ISSUE #28328: Enable FHIR parameter into existing Healthcare endpoint
Swagger information => https://github.com/Azure/azure-rest-api-specs/pull/17710

maririos:
Fixes: https://github.com/Azure/azure-sdk-for-net/issues/28328


deyaaeldeen (approved):
Looks good. Perhaps one nitpicky thing to consider is mention a thing or two about FHIR protocol in the changelog to educate customers about the feature.

sdk/textanalytics/Azure.AI.TextAnalytics/tests/RecognizeHealthcareEntitiesTests.cs
```diff
@@ -443,6 +444,34 @@ public async Task AnalyzeHealthcareEntitiesPagination()
...
         }
 
+        [RecordedTest]
+        public async Task RecognizeHealthcareEntitiesWithFhir()
+        {
+            TextAnalyticsClient client = GetClient(useTokenCredential: true);
+
+            AnalyzeHealthcareEntitiesOptions options = new AnalyzeHealthcareEntitiesOptions()
+            {
+                FhirVersion = FhirVersion.Four01
```
kristapratico, line 454: I'm guessing this is just how it generates?

kristapratico (approved):

sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/LanguageServiceClient.cs
```diff
@@ -1286,7 +1286,8 @@ private static HealthcareLROTask CreateHealthcareTask(AnalyzeHealthcareEntitiesO
                 {
                     ModelVersion = options.ModelVersion,
                     StringIndexType = Constants.DefaultStringIndexType,
-                    LoggingOptOut = options.DisableServiceLogs
+                    LoggingOptOut = options.DisableServiceLogs,
```
jsquire, line 1289: Does this need to be reflected to the Legacy API?

sdk/textanalytics/Azure.AI.TextAnalytics/tests/RecognizeHealthcareEntitiesTests.cs
jsquire: woah.

jsquire (approved):

sdk/textanalytics/Azure.AI.TextAnalytics/tests/RecognizeHealthcareEntitiesTests.cs
maririos: Yes :( Issue https://github.com/Azure/azure-sdk-for-net/issues/28811 to make it better next release :)

sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/LanguageServiceClient.cs
maririos: No. it is an addition that was introduced only in the Language service.
There is no equivalent on where to assign it if the user passes a previous service version. If they still assign the value and target a previous version, the SDK drops the property

jsquire: Is this the first case of us hitting the "how do we handle mismatch errors?"    Does the behavior change or trigger anything else strange for the legacy surface?    Should we drop note in #28060 to mention this scenario for follow-up?

sdk/textanalytics/Azure.AI.TextAnalytics/tests/RecognizeHealthcareEntitiesTests.cs
```diff
@@ -443,6 +444,35 @@ public async Task AnalyzeHealthcareEntitiesPagination()
             Assert.AreEqual(2, pages[0].Count);
         }
 
+        [RecordedTest]
+        [ServiceVersion(Min = TextAnalyticsClientOptions.ServiceVersion.V2022_04_01_Preview)]
```
maririos, line 448: Fixed to only run for new versions

sdk/textanalytics/Azure.AI.TextAnalytics/tests/RecognizeHealthcareEntitiesTests.cs
```diff
@@ -497,6 +527,8 @@ private void ValidateBatchDocumentsResult(AnalyzeHealthcareEntitiesResultCollect
                 }
 
                 Assert.IsNotNull(entitiesInDocument.Warnings);
+                Assert.IsNotNull(entitiesInDocument.EntityRelations);
+                Assert.IsNotNull(entitiesInDocument.FhirBundle);
```
maririos, line 531: This validation happens in most of the tests (version independent)

sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/LanguageServiceClient.cs
maririos: > Is this the first case of us hitting the "how do we handle mismatch errors?" 

It indeed is :)

>  Does the behavior change or trigger anything else strange for the legacy surface?

It does not. The `FhriVersion` input property only affects the `FhirBundle` output property which was introduced in this new version too.

>Should we drop note in https://github.com/Azure/azure-sdk-for-net/issues/28060 to mention this scenario for follow-up?

Added. I'm following what we have done in this library when a new property is added that is only supported in specific versions and does not affect behavior of older version output properties:
- We add that information to docstrings for both input and output property.
- Input property:
  - If the property is a query parameter, we send it to the service and let them manage it.
  - If the property is a body parameter, we drop it.
- Output property: Value is always empty Dict even when targeting older versions (I added comments in the tests to pint this out)

sdk/textanalytics/Azure.AI.TextAnalytics/tests/RecognizeHealthcareEntitiesTests.cs
maririos: Running official build to run tests from all versions

mssfang (approved):

sdk/textanalytics/Azure.AI.TextAnalytics/tests/RecognizeHealthcareEntitiesTests.cs
maririos: yay for tests!. looking into it

maririos: Fixed

maririos:
@jsquire My last commit includes the fix for this to work in both legacy and language clients. To keep it consistent with the work you did in legacy, I decided on not touching `Transforms.Legacy` and instead, having another constructor that takes the new parameters. Apologies for the iterations in the PR while figuring out best approach.
Let me know what you think.

jsquire (approved):

Issue #28531, opened 2022-05-04
[FEATURE REQ] DocumentTranslationInput does not allow to specify categoryId in DocumentTranslationInput constructor


sabbadino:
### Library name

Azure.AI.Translation.Document

### Please describe the feature.

DocumentTranslationInput does not allow to specify categoryId in DocumentTranslationInput constructor

maririos:
Thanks @sabbadino for reporting this. We will include it in our next release

maririos:
Hi @sabbadino, a way to currently add `CategoryId` is to do:
```
           var source = new TranslationSource(new Uri("http://localhost"));
            var target = new TranslationTarget(new Uri("http://localhost"), "en")
            {
                CategoryId = "myCategoryId"
            };
            var input = new DocumentTranslationInput(source, new List<TranslationTarget>() { target });
```

I'm also working on adding it to the `AddTarget()` method under the `DocumentTranslationInput` class so it is easily discoverable. 
Let me know what you think

sabbadino:
Thanks, i had missed that possibility.

maririos:
We've also added it to the `DocumentTranslationInput.AddTarget` method. You will see it in our next release.
Thank you for your feedback!

PR #28813, opened 2022-05-16
[Translation] Make categoryId more visible

maririos:
Contributes to https://github.com/Azure/azure-sdk-for-net/issues/28531

### DIFF
--- a/sdk/translation/Azure.AI.Translation.Document/api/Azure.AI.Translation.Document.netstandard2.0.cs
+++ b/sdk/translation/Azure.AI.Translation.Document/api/Azure.AI.Translation.Document.netstandard2.0.cs
@@ -71,7 +71,7 @@ namespace Azure.AI.Translation.Document
         public Azure.AI.Translation.Document.TranslationSource Source { get { throw null; } }
         public Azure.AI.Translation.Document.StorageInputType? StorageType { get { throw null; } set { } }
         public System.Collections.Generic.IList<Azure.AI.Translation.Document.TranslationTarget> Targets { get { throw null; } }
-        public void AddTarget(System.Uri targetUri, string languageCode, Azure.AI.Translation.Document.TranslationGlossary glossary = null) { }
+        public void AddTarget(System.Uri targetUri, string languageCode, Azure.AI.Translation.Document.TranslationGlossary glossary = null, string categoryId = null) { }
     }
     public static partial class DocumentTranslationModelFactory
     {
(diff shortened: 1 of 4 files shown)


sdk/translation/Azure.AI.Translation.Document/api/Azure.AI.Translation.Document.netstandard2.0.cs
```diff
@@ -71,7 +71,7 @@ public partial class DocumentTranslationInput
         public Azure.AI.Translation.Document.TranslationSource Source { get { throw null; } }
```
kristapratico, line 71: I think just needs a changelog and maybe consider adding a test to make sure categoryId is set?

sdk/translation/Azure.AI.Translation.Document/src/DocumentTranslationInput.cs
```diff
@@ -50,13 +50,16 @@ public DocumentTranslationInput(Uri sourceUri, Uri targetUri, string targetLangu
         /// <a href="https://docs.microsoft.com/azure/cognitive-services/translator/language-support#translate"/>.</param>
         /// <param name="glossary">Custom <see cref="TranslationGlossary"/> to be used in the translation operation. For supported file types see
         /// <see cref="DocumentTranslationClient.GetSupportedGlossaryFormatsAsync(System.Threading.CancellationToken)"/>.</param>
-        public void AddTarget(Uri targetUri, string languageCode, TranslationGlossary glossary = default)
+        /// <param name="categoryId">Category/custom model ID for using custom translation.</param>
+        public void AddTarget(Uri targetUri, string languageCode, TranslationGlossary glossary = default, string categoryId = default)
```
jsquire, line 54: From a technical standpoint, this looks fine.   I'm not sure how I feel about the double optional.    Should we consider possibly making an overload with `categoryId` so that it's more discoverable, or maybe removing the optionals and adding an overload that takes a fully-formed `TranslationTarget`?

jsquire (approved):

sdk/translation/Azure.AI.Translation.Document/src/DocumentTranslationInput.cs
maririos: The idea behind this method is to use the progressive disclosure principle where a user already used the service to translate a document from one language to another and now wants to add another language to the translation.
Originally, we had the 2 required parameters plus a less common one which are the glossaries in case the user wanted to explore that option too. This made it look like there was no other way to also instantiate the `categoryId` using this flow.

> making an overload with categoryId so that it's more discoverable

Is this something we are applying in our libraries thanks to the studies we've done? I've been in some studies where optional parameters helped more than overloads but this was before my leave so I might be missing context here.

jsquire: It's a pattern that we've been applying to T2 when there are core scenarios where less experienced developers are included in the audience.   It's not explicitly called out in the guidelines, but tends to come up during archboard reviews.   The best example is the constructor patterns that we use.    If we feel that this is an advanced scenario, then I don't think we'll see concern with multiple optional.

maririos: Interesting. Thank you so much for sharing that.

This is def for advanced users. CategoryId means the user previously created a custom translation and now wants to use that model.

sdk/translation/Azure.AI.Translation.Document/api/Azure.AI.Translation.Document.netstandard2.0.cs
maririos: Added

kristapratico (approved):

PR #28814, opened 2022-05-16
[Reservations] Adding unit tests for track 2

gaoyp830:


sdk/reservations/Azure.ResourceManager.Reservations/src/Generated/RestOperations/AzureReservationAPIRestOperations.cs
```diff
@@ -42,13 +42,14 @@ internal HttpMessage CreateGetCatalogRequest(string subscriptionId, string reser
         {
             var message = _pipeline.CreateMessage();
             var request = message.Request;
+            var apiVersion = reservedResourceType.Equals("SqlDatabases", StringComparison.OrdinalIgnoreCase) ? "2020-10-01-preview" : _apiVersion;
```
gaoyp830, line 45: SqlDatabases is currently not supported in 2022-03-01 version, using 2020-10-01-preview for it instead

sdk/reservations/Azure.ResourceManager.Reservations/src/Generated/RestOperations/AzureReservationAPIRestOperations.cs
```diff
@@ -42,13 +42,15 @@ internal HttpMessage CreateGetCatalogRequest(string subscriptionId, string reser
         {
             var message = _pipeline.CreateMessage();
             var request = message.Request;
+            // SqlDatabases is currently not supported in 2022-03-01 version, using 2020-10-01-preview for it instead
```
Yao725, line 45: Unfortunately, we don't allow you to manually change the generated code in the `Generated` folder. Do we have a timeline to support SqlDatabases in `2022-03-01`.

sdk/reservations/Azure.ResourceManager.Reservations/tests/Tests/GetCatalogTests.cs
```diff
@@ -0,0 +1,495 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Azure.Core.TestFramework;
+using Azure.ResourceManager.Reservations.Models;
+using NUnit.Framework;
+
+namespace Azure.ResourceManager.Reservations.Tests
```
Yao725, line 10: Why are there only test about `GetCatalog` method here? Do we plan to continue to add other exposed public API's tests in this PR later?

sdk/reservations/Azure.ResourceManager.Reservations/tests/Tests/GetCatalogTests.cs
```diff
@@ -0,0 +1,495 @@
...
+            Assert.IsTrue(catalogResult.Count > 0);
+
+            catalogResult.ForEach(item =>
+            {
+                Assert.AreEqual("virtualMachines", item.ReservedResourceType);
+                Assert.NotNull(item.Locations);
+                Assert.IsTrue(item.Locations.Count == 1);
+                Assert.AreEqual("eastus", item.Locations[0].Name);
+                Assert.AreEqual("East US", item.Locations[0].DisplayName);
+            });
```
Yao725, line 44: Can we extract this code to a private method? I found they are duplicated in many tests.

sdk/reservations/Azure.ResourceManager.Reservations/tests/Tests/GetCatalogTests.cs
```diff
@@ -0,0 +1,495 @@
...
+            AsyncPageable<ReservationCatalog> catalogResponse = Subscription.GetCatalogAsync("SuseLinux");
+            List<ReservationCatalog> catalogResult = await catalogResponse.ToEnumerableAsync();
+
+            Assert.NotNull(catalogResult);
+            Assert.IsTrue(catalogResult.Count > 0);
+
+            catalogResult.ForEach(item =>
+            {
+                Assert.AreEqual("SuseLinux", item.ReservedResourceType);
+            });
```
Yao725, line 80: Same as above.

Yao725 (commented):
Left a few comments, thanks.

sdk/reservations/Azure.ResourceManager.Reservations/tests/Tests/GetCatalogTests.cs
gaoyp830: Initially I was thinking to keep this PR small, which was why I only added catalog tests. But I can add tests for other operations here as well so that we can have all the tests checked in one single PR.

sdk/reservations/Azure.ResourceManager.Reservations/tests/Tests/GetCatalogTests.cs
gaoyp830: will do

sdk/reservations/Azure.ResourceManager.Reservations/src/Generated/RestOperations/AzureReservationAPIRestOperations.cs
gaoyp830: Reverted back. We will be supporting in 2022-03-01 version soon

gaoyp830:
> Left a few comments, thanks.

@Yao725 Addressed comments with new changes

sdk/reservations/Azure.ResourceManager.Reservations/src/Generated/Models/PurchaseRequestContent.cs
```diff
@@ -79,7 +79,7 @@ public string SkuName
         /// <summary> Type of the Applied Scope. </summary>
         public AppliedScopeType? AppliedScopeType { get; set; }
         /// <summary> List of the subscriptions that the benefit will be applied. Do not specify if AppliedScopeType is Shared. </summary>
-        public IList<string> AppliedScopes { get; }
+        public IList<string> AppliedScopes { get; set; }
```
Yao725, line 82: For non-required property, we will initialize it using `ChangeTrackingList<T>` in ctor, then you can use `list = {}` or `list.Add()` to set its value, so the setter of this property is removed. This is by design, please don't add the setter.

sdk/reservations/Azure.ResourceManager.Reservations/src/Generated/Models/ReservationOrderResponseData.Serialization.cs
```diff
@@ -44,7 +44,7 @@ internal static ReservationOrderResponseData DeserializeReservationOrderResponse
                         property.ThrowNonNullablePropertyIsNull();
                         continue;
                     }
-                    etag = new ETag(property.Value.GetString());
+                    etag = new ETag(property.Value.ToString());
```
Yao725, line 47: This should not be changed, any reason we need this customization?

sdk/reservations/Azure.ResourceManager.Reservations/src/Generated/Models/PurchaseRequestContent.Serialization.cs
```diff
@@ -63,7 +63,7 @@ void IUtf8JsonSerializable.Write(Utf8JsonWriter writer)
                 writer.WritePropertyName("appliedScopeType");
                 writer.WriteStringValue(AppliedScopeType.Value.ToString());
             }
-            if (Optional.IsCollectionDefined(AppliedScopes))
+            if (Optional.IsCollectionDefined(AppliedScopes) && AppliedScopes != null)
```
Yao725, line 66: Please don't change the condition here.

Yao725 (changes requested):
Hi @gaoyp830, thanks for your update, I left some comments for the latest code. One thing I want to emphasize is, please don't manually change anything in the `Generated` directory. From maintenance perspective, if the generated code didn't work as expected or have some bugs, we should follow this order to fix it:

- See if we could update the swagger spec
- Using autorest.md to configure it
- Using customized code in separate customization directory (This should be our last choice)

sdk/reservations/Azure.ResourceManager.Reservations/src/autorest.md
```diff
@@ -57,8 +57,8 @@ directive:
       $.ReservationOrderResponse.properties.etag['x-ms-format'] = 'etag';
       $.PurchaseRequest['x-ms-client-name'] = 'PurchaseRequestContent';
       $.Price['x-ms-client-name'] = 'PurchasePrice';
-      $.Catalog.properties.resourceType['x-ms-format'] = 'resource-type';
+       $.Catalog.properties.resourceType['x-ms-client-name'] = 'reservedResourceType';
```
archerzz, line 60: remove the extra leading space

archerzz: also, any reason we change the type back to `string`?

Yao725: The reason that we revert this change is because the `resourceType` here in not a truly `ResourceType` that we defined in Azure.Core. It ‘s more like a simplified resource type. For instance, it will be `virtualMachines` instead of `Microsoft.Compute/virtualMachines`

sdk/reservations/Azure.ResourceManager.Reservations/src/Generated/ReservationOrderResponseData.cs
```diff
@@ -57,7 +56,7 @@ internal ReservationOrderResponseData(ResourceIdentifier id, string name, Resour
         }
 
         /// <summary> Gets the etag. </summary>
-        public ETag? Etag { get; }
+        public int? Etag { get; }
```
Yao725, line 59: Revert this change so it doesn't block testing. The backend service returns an integer.

sdk/reservations/Azure.ResourceManager.Reservations/src/Generated/Models/PurchaseRequestContent.cs
gaoyp830: Reverted back

sdk/reservations/Azure.ResourceManager.Reservations/tests/Tests/CalculatePriceTests.cs
```diff
@@ -0,0 +1,125 @@
...
+        [TestCase]
+        [RecordedTest]
+        public async Task TestCalculatePriceForSingleScopeUpfront()
+        {
+            var billingPlan = "Upfront";
+            var response = await Tenant.CalculateReservationOrderAsync(CreatePurchaseRequestContent("Single", billingPlan));
+            testCalculatePriceResponse(response, billingPlan);
+        }
+
+        private void testCalculatePriceResponse(Response<CalculatePriceResponse> response, string billingPlan)
```
Yao725, line 71: Based on the .NET convention, I think this should use `PascalCase`.
```suggestion
        private void TestCalculatePriceResponse(Response<CalculatePriceResponse> response, string billingPlan)
```

sdk/reservations/Azure.ResourceManager.Reservations/tests/Tests/GetCatalogTests.cs
```diff
@@ -0,0 +1,281 @@
...
+        [TestCase]
+        [RecordedTest]
+        public async Task TestGetCatalogForVirtualMachineSoftware()
+        {
+            AsyncPageable<ReservationCatalog> catalogResponse = Subscription.GetCatalogAsync("VirtualMachineSoftware", publisherId: "test_test_pmc2pc1", offerId: "mnk_vmri_test_001", planId: "testplan001");
+            List<ReservationCatalog> catalogResult = await catalogResponse.ToEnumerableAsync();
+            testForCatalogResult(catalogResult, "VirtualMachineSoftware");
+        }
+
+        private void testForCatalogResult(List<ReservationCatalog> catalogResult, string resourceTypeName, string alternateResourceTypeName = null, bool hasLocation = false, string location = null, string locationDisplayName = null)
```
Yao725, line 255: Considering naming convention and consistency, I'd like to rename this method as follows:
```suggestion
        private void TestGetCatalogResponse(List<ReservationCatalog> catalogResult, string resourceTypeName, string alternateResourceTypeName = null, bool hasLocation = false, string location = null, string locationDisplayName = null)
```

sdk/reservations/Azure.ResourceManager.Reservations/tests/Tests/PurchaseReservationOrderTests.cs
```diff
@@ -0,0 +1,124 @@
...
+
+        [TestCase]
+        [RecordedTest]
+        public async Task TestPurchaseReservationOrderSharedScopeMonthly()
+        {
+            var purchaseRequestContent = CreatePurchaseRequestContent("Shared", "Monthly");
+            var response = await Tenant.CalculateReservationOrderAsync(purchaseRequestContent);
+            var purchaseResponse = await collection.CreateOrUpdateAsync(WaitUntil.Completed, response.Value.Properties.ReservationOrderId, purchaseRequestContent);
+
+            await purchaseResponse.WaitForCompletionAsync();
```
Yao725, line 45: You don't need to wait by yourself, we will [call this method](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/reservations/Azure.ResourceManager.Reservations/src/Generated/ReservationOrderResponseCollection.cs#L80-L81) for you as you specified the `WaitUntil.Completed` in the `CreateOrUpdate` method, so please remove this line in this file.

sdk/reservations/Azure.ResourceManager.Reservations/tests/Tests/PurchaseReservationOrderTests.cs
```diff
@@ -0,0 +1,124 @@
...
+        public async Task TestPurchaseReservationOrderSingleScopeUpfront()
+        {
+            var purchaseRequestContent = CreatePurchaseRequestContent("Single", "Upfront");
+            var response = await Tenant.CalculateReservationOrderAsync(purchaseRequestContent);
+            var purchaseResponse = await collection.CreateOrUpdateAsync(WaitUntil.Completed, response.Value.Properties.ReservationOrderId, purchaseRequestContent);
+            await purchaseResponse.WaitForCompletionAsync();
+            validateTestResult(purchaseResponse, purchaseRequestContent, response.Value.Properties.ReservationOrderId);
+        }
+
+        private void validateTestResult(ArmOperation<ReservationOrderResponseResource> purchaseResponse, PurchaseRequestContent purchaseRequest, string reservationOrderId)
```
Yao725, line 83: Maybe a renaming is better:
```suggestion
        private void TestCreatePurchaseResponse(ArmOperation<ReservationOrderResponseResource> purchaseResponse, PurchaseRequestContent purchaseRequest, string reservationOrderId)
```

Yao725 (approved):
Generally, it LGTM, left a few comments to improve, thanks!
And FYI, after offline discussion, we think an int type `etag` is really too strange, so I renamed them to `version` to avoid breaking the consistency between different RPs.

gaoyp830:
> Generally, it LGTM, left a few comments to improve, thanks! And FYI, after offline discussion, we think an int type `etag` is really too strange, so I renamed them to `version` to avoid breaking the consistency between different RPs.

Thanks! I sent a new commit based on your comments. Please help merging this PR if you think everything looks good!

PR #28816, opened 2022-05-16
[MediaComposition] Introducing SDK Private Preview

minnieliu:

# Media Composition SDK Private Preview
- This PR is meant to introduce the initial version of the Media Composition SDK, which is a new service that we are working on.
- Our service is still a WIP and so we've skipped the E2E tests for now. A sample file has been included to get a feel for using the current SDK.
- Please keep in mind that this is just the initial version, we plan to have further improvements including: 
    - Polymorphic types of reduce nesting in models
    - Dealing with merge patch better by splitting up the update into smaller more focused functions like `AddInput`/`RemoveInput`/`UpdateLayout`/etc
    - More consideration for protocol methods, is this right for the Media Composition SDK?


sdk/communication/Azure.Communication.MediaComposition/src/Models/ImageInput.cs
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under the MIT License.
+
+using Azure.Core;
+
+namespace Azure.Communication.MediaComposition.Models
+{
+    /// <summary>Image input.</summary>
+    [CodeGenModel("Image")]
+    public partial class ImageInput
+    {
```
JianpingChen, line 11: Is there a flag we can set to not generate these "empty" files? Are they really useful?

minnieliu: This is the way to rename Image -> ImageInput because the .Net SDK does not allow for single word classes as it causes ambiguity

sdk/communication/Azure.Communication.MediaComposition.Tests/samples/Sample1_MediaCompositionClient.cs
```diff
@@ -0,0 +1,156 @@
...
+            {
+                Resolution = new LayoutResolution()
+                {
+                    Width = 1920,
+                    Height = 1080
+                },
+                Grid = new GridLayoutOptions(
+                    rows: 2,
+                    columns: 2,
+                    new string[2, 2] { { "jill", "jack" }, { "jane", "jerry" }
```
minnieliu, line 31: Is this the right format for the 2d array or is it a nested list?

sdk/communication/Azure.Communication.MediaComposition.Tests/samples/Sample1_MediaCompositionClient.cs
```diff
@@ -0,0 +1,156 @@
...
+    {
+        // TODO: Update the samples to tests when the E2E flow is working
+        private const string mediaCompositionId = "2x2Grid";
+        public async Task CreateMediaCompositionAsync()
+        {
+            var mediaCompositionClient = CreateMediaCompositionClient();
+
+            var layout = new MediaCompositionLayout()
+            {
+                Resolution = new LayoutResolution()
```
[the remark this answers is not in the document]
thdinizm: It can be empty if we have a default resolution. Is having a default resolution a good experience? Hard to answer..

sdk/communication/Azure.Communication.MediaComposition.Tests/samples/Sample1_MediaCompositionClient.cs
```diff
@@ -0,0 +1,156 @@
...
+                }),
+            };
+
+            var inputs = new Dictionary<string, MediaInput>()
+            {
+                {
+                    "jill",
+                    new MediaInput()
+                    { Participant = new ParticipantInput()
+                        { Id = new CommunicationIdentifierModel()
```
thdinizm, line 41: I think Dom mentioned before to make sure this property is using the standard all other ACS services use. Have we double checked that?

sdk/communication/Azure.Communication.MediaComposition.Tests/samples/Sample1_MediaCompositionClient.cs
```diff
@@ -0,0 +1,156 @@
...
+                            { MicrosoftTeamsUser = new MicrosoftTeamsUserIdentifierModel("f3ba9014-6dca-4456-8ec0-fa03cfa2b7b7") },
+                            Call = "teamsMeeting",
+                            PlaceholderImageUri = "https://imageendpoint"
```
thdinizm, line 44: I think placeholderImageUri should be option (if it already isn't)

minnieliu: Yup, currently, it's an optional property and users can decide if they want to set it.

sdk/communication/Azure.Communication.MediaComposition.Tests/samples/Sample1_MediaCompositionClient.cs
minnieliu: Yes, it is the standard one that all other ACS services are using. In the swagger, we link to this Model type which belong in the common 
image

sdk/communication/Azure.Communication.MediaComposition.Tests/samples/Sample1_MediaCompositionClient.cs
0rland0Wats0n: in my unit tests I used nested lists 
image

but I'm no c# expert

sdk/communication/Azure.Communication.MediaComposition/src/MediaCompositionClient.cs
```diff
@@ -0,0 +1,320 @@
...
+        /// <param name="inputs"> The media inputs. </param>
+        /// <param name="outputs"> The media outputs. </param>
+        /// <param name="cancellationToken"> The cancellation token to use. </param>
+        public virtual async Task<Response<MediaCompositionBody>> CreateAsync(string mediaCompositionId, MediaCompositionLayout layout, IDictionary<string, MediaInput> inputs, IDictionary<string, MediaOutput> outputs, CancellationToken cancellationToken = default)
+        {
+            using var scope = _clientDiagnostics.CreateScope("MediaCompositionClient.Create");
+            scope.Start();
+            try
+            {
+                // TODO: what is this id and is it needed?
```
minnieliu, line 139: Commenting for visibility: What is this second Id and can we merge it with the Media composition Id?

sdk/communication/Azure.Communication.MediaComposition/src/MediaCompositionClient.cs
```diff
@@ -0,0 +1,320 @@
...
+        /// <param name="outputs"> The updated media outputs. </param>
+        /// <param name="cancellationToken"> The cancellation token to use. </param>
+        public virtual async Task<Response<MediaCompositionBody>> UpdateAsync(string mediaCompositionId, MediaCompositionLayout layout = null, IDictionary<string, MediaInput> inputs = null, IDictionary<string, MediaOutput> outputs = null, CancellationToken cancellationToken = default)
+        {
+            using var scope = _clientDiagnostics.CreateScope("MediaCompositionClient.Update");
+            scope.Start();
+            try
+            {
+                // TODO: what is this id and how to retrieve for updates?
+                // Should the CompositionStreamState be exposed for update?
```
minnieliu, line 184: Commenting for visibility:

- What is this id and how to retrieve for updates?
- Should the CompositionStreamState be exposed for update?

sdk/communication/Azure.Communication.MediaComposition.Tests/samples/Sample1_MediaCompositionClient.cs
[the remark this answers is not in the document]
minnieliu: That's also a good point, I think adding a default resolution might be useful especially for those who aren't sure what the common dimensions are

sdk/communication/Azure.Communication.MediaComposition/src/MediaCompositionClient.cs
minnieliu: Had a discussion with team, we will be trying to remove the second `id` property because the two point to the same thing.

sdk/communication/Azure.Communication.MediaComposition/src/MediaCompositionClient.cs
minnieliu: Had a discussion with team, the CompositionStreamState will be updated to readOnly and should not be exposed to set

sdk/communication/Azure.Communication.MediaComposition.Tests/samples/Sample1_MediaCompositionClient.cs
minnieliu: Okay, I did the same in the Layout Component as well, so just updated accordingly,

sdk/communication/Azure.Communication.MediaComposition/src/MediaCompositionClient.cs
minnieliu: Doesn't look like removing the id parameter worked. @DominikMe what are the changes we can make on the controller side to remove the id? Since we are just passing back mediaCompositionId for both

sdk/communication/Azure.Communication.MediaComposition/tests/samples/Sample1_MediaCompositionClient.cs
```diff
@@ -0,0 +1,157 @@
...
+            var layout = new MediaCompositionLayout()
+            {
+                Resolution = new LayoutResolution(720, 480),
+                Presenter = new PresenterLayoutOptions("jill", "jack")
+                {
+                    SupportPosition = SupportPosition.BottomRight,
+                    SupportAspectRatio = 3/2
+                }
+            };
+            await mediaCompositionClient.UpdateAsync(mediaCompositionId, layout);
```
DominikMe, line 126: We'll need more discussions whether we want to support a full Update or instead have smaller methods like `AddSource`. The reason is that merge patch semantics are hard to convey, and it's not clear that passing a dictionary will only add new items but not remove pre-existing ones, or that one has to set a dictionary item to null to remove it.

Azure SDK is actually changing their engineering strategy to generate the data-plane client entirely and only start with what they call "protocol"-based methods instead of "model"-based methods. The idea is that a team can start with the protocol layer where developers send raw requests, and can then add model-based methods later. So far we haven't adopted that new approach in ACS yet, but maybe we can see if it would make sense for media composition?

https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/samples/ProtocolMethods.md

sdk/communication/Azure.Communication.MediaComposition/tests/samples/Sample1_MediaCompositionClient.cs
```diff
@@ -0,0 +1,157 @@
...
+                        {
+                            PlaceholderImageUri = "https://imageendpoint"
+                        }
+                    }
+                },
+                 {
+                    "jane",
+                    new MediaInput() {
+                        Participant = new ParticipantInput(
+                        id: new CommunicationIdentifierModel() { MicrosoftTeamsUser = new MicrosoftTeamsUserIdentifierModel("2dd69470-dc25-49cf-b5c3-f562f08bf3b2") },
```
DominikMe, line 68: In our final model-based API we should use the SDK identifier types and not the identifier model types.

sdk/communication/Azure.Communication.MediaComposition/src/autorest.md
```diff
@@ -0,0 +1,17 @@
...
+
+Run `dotnet msbuild /t:GenerateCode` to generate code.
+
+### AutoRest Configuration
+> see https://aka.ms/autorest
+
+``` yaml
+title: Media Composition
+tag: package-2022-06-26-preview1
+generation1-convenience-client: true
```
DominikMe, line 11: ah, interesting. Is that the new flag to opt-into the behavior before Dataplane-Generation?

sdk/communication/Azure.Communication.MediaComposition/src/MediaCompositionClient.cs
```diff
@@ -0,0 +1,314 @@
...
+                throw;
+            }
+        }
+
+        /// <param name="mediaCompositionId"> The media composition id. </param>
+        /// <param name="layout"> The layout to compose the media. </param>
+        /// <param name="inputs"> The media inputs. </param>
+        /// <param name="outputs"> The media outputs. </param>
+        /// <param name="cancellationToken"> The cancellation token to use. </param>
+        public virtual async Task<Response<MediaCompositionBody>> CreateAsync(string mediaCompositionId, MediaCompositionLayout layout, IDictionary<string, MediaInput> inputs, IDictionary<string, MediaOutput> outputs, CancellationToken cancellationToken = default)
```
DominikMe, line 133: Do we feel that layout, inputs, outputs should be required on creation?
How would the alternative of passing the whole MediaCompositionBody as an argument feel, pro/cons?

minnieliu: Yea, great point Dom and definitely something worth discussion.

I've created a sample file to get a feel of how the SDK works. Personally, I felt there was lots of nesting when building the inputs and the layouts. The objects can also get quite large. I think passing the entire MediaCompositionBody might make that even more apparent.

sdk/communication/Azure.Communication.MediaComposition/src/autorest.md
minnieliu: Actually, I found this flag in another ACS SDK and I had to include it for the models to be generated. I'm not sure exactly what it's meant to do but was required for autorest to work properly

sdk/communication/Azure.Communication.MediaComposition/tests/samples/Sample1_MediaCompositionClient.cs
minnieliu: Could you clarify what you mean here by the SDK identifier types?

DominikMe (approved):
This is a great start! lgtm

sdk/communication/Azure.Communication.MediaComposition/src/MediaCompositionClient.cs
minnieliu: Discussed with Dom offline, we will look into removing the additional id parameter, but okay to keep as is for now

sdk/communication/Azure.Communication.MediaComposition/tests/samples/Sample1_MediaCompositionClient.cs
minnieliu: So looks like right now, since it's nested inside the ParticipantInput, we won't be able to use the SDK identifier types. However, once we make polymorphic types for the ParticipantInput class, we should be able to switch to using the SDK Identifier Types

sdk/communication/ci.yml
```diff
@@ -38,6 +38,8 @@ extends:
       safeName: AzureCommunicationSms
     - name: Azure.Communication.Identity
       safeName: AzureCommunicationIdentity
+    - name: Azure.Communication.MediaComposition
+      safeName: AzureCommunicationMediaComposition
```
chidozieononiwu, line 42: Since you are not planning to do a public release, consider adding `skipReleaseStage: true` parameter so that the release is never triggered.

minnieliu:
@KrzysztofCwalina @tg-msft 
This PR is meant to introduce the initial version of the Media Composition SDK, which is a new service that we are working on.
Our service is still a WIP and we currently don't have the E2E running yet so it will be at least a few weeks before we release this SDK. We wanted to merge this initial version in first to the main branch, please let take a look and let us know if that's good to go. Thank you!

sdk/communication/ci.yml
minnieliu: Added in my latest commit, thank you for the suggestion!

Issue #28844, opened 2022-05-18
Support using ResourceId for ManagedIdentity configuration


JoshLove-msft:
This feature was added to to Identity. See https://github.com/Azure/azure-sdk-for-net/issues/24249

We would like to expose this in Microsoft.Extensions.Azure so that it can be used for functions extensions.

PR #28845, opened 2022-05-18
Support ResourceId configuration for ManagedIdentityCredential

JoshLove-msft:
Addresses https://github.com/Azure/azure-sdk-for-net/issues/28844

### DIFF
--- a/sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientFactory.cs
+++ b/sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientFactory.cs
@@ -88,7 +88,7 @@ namespace Microsoft.Extensions.Azure
             var credentialType = configuration["credential"];
             var clientId = configuration["clientId"];
             var tenantId = configuration["tenantId"];
-            var resourceId = configuration["resourceId"];
+            var resourceId = configuration["UserAssignedManagedIdentityResourceId"];
             var clientSecret = configuration["clientSecret"];
             var certificate = configuration["clientCertificate"];
             var certificateStoreName = configuration["clientCertificateStoreName"];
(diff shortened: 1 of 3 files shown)


sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientFactory.cs
```diff
@@ -88,13 +88,19 @@ internal static TokenCredential CreateCredential(IConfiguration configuration, T
...
             var tenantId = configuration["tenantId"];
+            var resourceId = configuration["resourceId"];
             var clientSecret = configuration["clientSecret"];
             var certificate = configuration["clientCertificate"];
             var certificateStoreName = configuration["clientCertificateStoreName"];
             var certificateStoreLocation = configuration["clientCertificateStoreLocation"];
 
             if (string.Equals(credentialType, "managedidentity", StringComparison.OrdinalIgnoreCase))
             {
+                if (!string.IsNullOrWhiteSpace(resourceId))
```
JoshLove-msft, line 99: I put this as the first check before ClientId, because apparently ClientId can be empty as it is an optional parameter to the ManagedIdentityCredential.

jsquire: Should we update the docs to indicate that if both are provided, `resourceId` takes precedence?

jsquire (approved):

sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientFactory.cs
christothes: In Identity, we throw if both are specified.

JoshLove-msft: > In Identity, we throw if both are specified.

I considered doing this here, but was hesitant because there is nothing that stopped users from putting resourceId in the config before, so wanted to attempt to handle this gracefully.

JoshLove-msft: Chatted offline - will follow same behavior as Identity and update the config key to reduce chance of customers being broken by the change.

christothes (approved):

PR #28870, opened 2022-05-18
Encryption v2

jaschrep-msft:
Adding authenticated encryption support.

## Behavioral Changes
- Ranged Download (`DownloadStreaming`, `DownloadContent`)
  - If client is configured with client-side encryption options, first fetches properties to determine the appropriate range adjustment, then does the ranged read.
  - This does not affect other downloads that use this under the hood
  - Pro: Not excessively extending ranges when unnecessary
  - Con: Extra rest call
- Partitioned Download (`DownloadTo`)
  - Decryption shifted from per-partition to the reassembled content stream
  - Pro: avoids overhead added to ranged download
  - Pro: avoids redownloading overlapping bytes from range extension
  - Pro: effectively separates transfer options from decryption logic
  - Con: v1 decryption loses parallelization (was tied to transfer parallelization)


sdk/storage/Azure.Storage.Blobs/src/Azure.Storage.Blobs.csproj
```diff
@@ -1,12 +1,12 @@
...
   <PropertyGroup>
-    <TargetFrameworks>$(RequiredTargetFrameworks)</TargetFrameworks>
+    <TargetFrameworks>$(RequiredTargetFrameworks);netstandard2.1</TargetFrameworks>
   </PropertyGroup>
   <PropertyGroup>
     <AssemblyTitle>Microsoft Azure.Storage.Blobs client library</AssemblyTitle>
     <Version>12.13.0-beta.1</Version>
     <!--The ApiCompatVersion is managed automatically and should not generally be modified manually.-->
-    <ApiCompatVersion>12.12.0</ApiCompatVersion>
+    <ApiCompatVersion Condition="'$(TargetFramework)' != 'netstandard2.1'">12.12.0</ApiCompatVersion>
```
jaschrep-msft, line 9: Temporary workaround from AzureSDK until an initial version with a netstandard2.1 target is on NuGet.

sdk/storage/Azure.Storage.Common/src/Shared/ClientsideEncryption/Models/TransformMode.cs
```diff
@@ -0,0 +1,27 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Azure.Storage.Cryptography.Models
+{
+    /// <summary>
+    /// Transformation mode for a give <see cref="IAuthenticatedCryptographicTransform"/>.
```
jaschrep-msft, line 13: "given"

jaschrep-msft:
@KrzysztofCwalina @tg-msft PR is in a state where your interests can be reviewed. You'll be interested in:

- csproj files
- Azure.Storage.Common.netstandard2.0.cs
- GcmAuthenticatedCryptographicTransform.cs
- BlobBaseClient.cs

Behavioral changes listed in PR description. Contents of Shared/ClientSideEncryption is where all the heavy operations take place if you need context. Happy to answer any questions.

QueueClient work is upcoming. Since queue message size is trivial that they are already buffered in the first place, there are no surprises coming in how data is managed.

sdk/storage/Azure.Storage.Blobs/src/BlobBaseClient.cs
```diff
@@ -1398,12 +1402,43 @@ private void SetNameFieldsIfNull()
...
+        /// Cancellation token.
+        /// </param>
+        /// <returns></returns>
+        internal virtual async Task<Response<BlobDownloadStreamingResult>> DownloadStreamingInternal(
             HttpRange range,
             BlobRequestConditions conditions,
             bool rangeGetContentHash,
             IProgress<long> progressHandler,
             string operationName,
+            bool bypassClientSideEncryption,
```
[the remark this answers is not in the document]
seanmcc-msft: Ya I like that better

sdk/storage/Azure.Storage.Common/tests/Azure.Storage.Common.Tests.csproj
```diff
@@ -8,6 +8,14 @@
     <IsPackable>false</IsPackable>
     <RootNamespace>Azure.Storage.Tests</RootNamespace>
   </PropertyGroup>
+
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net461'">
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
```
amnguye, line 13: Should we eventually move these tests to the Aes Gcm Tests package to avoid having to do this in the Common Tests package?

amnguye (approved):

sdk/storage/Azure.Storage.Blobs/tests/BlobClientOpenWriteTests.cs
```diff
@@ -18,7 +18,7 @@ namespace Azure.Storage.Blobs.Tests
     public class BlobClientOpenWriteTests : BlobBaseClientOpenWriteTests<BlobClient>
     {
         public BlobClientOpenWriteTests(bool async, BlobClientOptions.ServiceVersion serviceVersion)
-            : base(async, serviceVersion, null /* RecordedTestMode.Record /* to re-record */)
+            : base(async, serviceVersion, RecordedTestMode.Record /* RecordedTestMode.Record /* to re-record */)
```
amnguye, line 21: Remember to change back to null

sdk/storage/Azure.Storage.Common/tests/AuthenticatedRegionCryptoStreamTest.cs
```diff
@@ -0,0 +1,432 @@
...
+
+                CollectionAssert.AreEqual(
+                    plaintextAuthBlock.ToArray(),
+                    ciphertextAuthBlock.Slice(
+                        _nonceLength,
+                        Math.Min(_authRegionDataLength, ciphertextAuthBlock.Length - _nonceLength - _tagLength)).ToArray());
+            }
+        }
+
+#if NETCOREAPP3_1
```
jaschrep-msft, line 352: Remove conditional compilation.

sdk/storage/Azure.Storage.Common/tests/AuthenticatedRegionCryptoStreamTest.cs
```diff
@@ -0,0 +1,432 @@
...
+        [Test]
+        [Combinatorial]
+        public void RoundTrip_Read(
+            [Values(true, false)] bool useRealEncryption,
+            [Values(true, false)] bool alligned,
+            [Values(1, 3)] int numAuthBlocks)
```
jaschrep-msft, line 358: @amnguye this is the one test that needs the unsafe tag to be added, and only for `useRealEncryption == true`. I'm not sure just removing a test is a good idea, but I'll admit it becomes more of an integration test than a unit test when we use GCM here. Perhaps we need a broader effort to classify tests as unit/integration/smoke. This would likely cut down on bulky test classes and packages in general.

sdk/storage/Azure.Storage.Queues/src/Azure.Storage.Queues.csproj
```diff
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFrameworks>$(RequiredTargetFrameworks)</TargetFrameworks>
```
kasobol-msft, line 3: should this get compiled to netstandard2.1?

sdk/storage/Azure.Storage.Blobs/src/BlobBaseClient.cs
```diff
@@ -1424,11 +1459,14 @@ private void SetNameFieldsIfNull()
...
+                    if (UsingClientSideEncryption && !ignoreClientSideEncryption)
                     {
                         // TODO #27253
                         //options = BlobDownloadOptions.CloneOrDefault(options) ?? new BlobDownloadOptions();
-                        range = BlobClientSideDecryptor.GetEncryptedBlobRange(range);
+                        if ((await GetPropertiesInternal(conditions, async, cancellationToken).ConfigureAwait(false)).Value.Metadata.TryGetValue(Constants.ClientSideEncryption.EncryptionDataKey, out string rawEncryptiondata))
+                        {
+                            range = BlobClientSideDecryptor.GetEncryptedBlobRange(range, rawEncryptiondata);
+                        }
                     }
```
kasobol-msft, line 1470: is there other reason than get properties call for `ignoreClientSideEncryption` ? 
It seems that we're pushing up implementation to callers where properties are already available.

Would passing optional properties/metadata to this method eliminate need for  `ignoreClientSideEncryption` flag? I.e. call getProperties if needed?

jaschrep-msft: It ignores overlapping ranges for something like a partitioned download, and also just skips decryption as a whole. This was trivial in v1 when ranges were extended a handful of bytes, but now we are extending to ~4MB bounds, which may not play nice with transfer options. While we could pass in optional prefetched properties, we also ignore potential boundary extension waste by just shifting decryption back to the partitioned downloader.

jaschrep-msft: I do see how just adding bool flags is unfun, though. Perhaps it's better for the "real" DownloadStreamingInternal method as it exists today to not have a concept of client-side encryption. Instead, the public ranged download APIs go through a "DirectCalledDownloadStreamingInternal" which applies the appropriate transformations. All customer-called ranged downloads go through there, and our partitioned downloader and open-read don't. This way every public API goes through some function that applies client-side encryption, but we don't have flagged branching logic in our low-level REST download. 

Thoughts?

sdk/storage/Azure.Storage.Blobs/tests/PartitionedDownloaderTests.cs
```diff
@@ -205,32 +201,60 @@ private void AssertContent(int expectedLength, MemoryStream stream)
...
+            //        .Returns<HttpRange, BlobRequestConditions, bool, CancellationToken>((range, conditions, rangeGetHash, cancellation) =>
+            //            dataSource.GetStream(
+            //                range,
+            //                conditions,
+            //                rangeGetHash,
+            //                progress: default,
+            //                cancellation));
+            //    blockClient.Setup(c => c.DownloadStreaming(It.IsAny<HttpRange>(), It.IsAny<BlobRequestConditions>(), false, It.IsAny<IProgress<long>>(), s_cancellationToken))
+            //        .Returns<HttpRange, BlobRequestConditions, bool, IProgress<long>, CancellationToken>(dataSource.GetStream);
+            //}
```
jaschrep-msft, line 257: Remove

sdk/storage/Azure.Storage.Blobs/src/BlobBaseClient.cs
kasobol-msft: yeah, that'd be my next suggestion. `UsingClientSideEncryption` and `ignoreClientSideEncryption` in same condition looks odd and smells like spaghetti.

Pablo1q (approved):
Good

sdk/storage/Azure.Storage.Common/tests/AuthenticatedRegionCryptoStreamTest.cs
jaschrep-msft: I have moved it to the blob encryption tests, which already need to be configured for something like this. I can take a work item later to revisit our testing organization.

sdk/storage/tests.yml
```diff
@@ -6,13 +6,13 @@ extends:
...
-    Location: canadacentral
+    Location: canadacentral
     CloudConfig:
       Preview:
-        SubscriptionConfiguration: $(sub-config-azure-cloud-test-resources-preview)
-      PrivatePreview:
+        SubscriptionConfiguration: $(sub-config-azure-cloud-test-resources-preview)
+      PrivatePreview:
         SubscriptionConfiguration: $(sub-config-storage-test-resources)
-    Clouds: Preview
```
jaschrep-msft, line 15: line ending normalization from editor

sdk/storage/tests.yml
```diff
@@ -26,5 +26,9 @@ extends:
     ${{ if not(contains(variables['Build.DefinitionName'], 'tests-weekly')) }}:
       EnvVars:
         AZURE_ONLY_TEST_LATEST_SERVICE_VERSION: true
+    PreSteps:
+      - bash: brew install openssl@1.1
+        displayName: (MacOS) Grant access to ~/.Azure
```
weshaggard, line 31: Probably should update to display name. 
```suggestion
        displayName: (MacOS) Install OpenSSL
```

sdk/storage/ci.yml
```diff
@@ -53,3 +55,5 @@ extends:
       safeName: AzureResourceManagerStorage
     TestSetupSteps:
     - template: /sdk/storage/tests-install-azurite.yml
+    - ${{ contains(variables['OSVmImage'], 'mac') }}:
+        bash: brew install openssl@3
```
weshaggard, line 59:
```suggestion
    - bash: brew install openssl@3
      displayName: (MacOS) Install OpenSSL
      condition: contains(variables['OSVmImage'], 'mac')
    
```

sdk/storage/Azure.Storage.Common/src/Shared/UriExtensions.cs
```diff
@@ -113,7 +113,7 @@ public static string GetAccountNameFromDomain(string host, string serviceSubDoma
         /// <returns>True if using IP Endpoint style.</returns>
         public static bool IsHostIPEndPointStyle(this Uri uri) =>
             (!string.IsNullOrEmpty(uri.Host) &&
-            uri.Host.IndexOf(".", StringComparison.InvariantCulture) >= 0 &&
+            uri.Host.Contains(".") &&
```
tg-msft, line 116: Why was this change necessary?  Can we pass the invariant culture into `Contains` at least?

sdk/storage/Azure.Storage.Common/src/ClientSideEncryptionVersion.cs
```diff
@@ -12,7 +12,12 @@ public enum ClientSideEncryptionVersion
         /// <summary>
         /// 1.0
         /// </summary>
-        V1_0 = 1
+        V1_0 = 1,
```
tg-msft, line 15: Should we add a comment about which version to prefer?

sdk/storage/Azure.Storage.Common/src/Shared/ClientsideEncryption/AuthenticatedRegionCryptoStream.cs
```diff
@@ -0,0 +1,290 @@
...
+        private readonly IAuthenticatedCryptographicTransform _transform;
+        private bool _flushedFinal;
+
+        private readonly byte[] _buffer;
+        private int _bufferPos;
+        // in read mode, innerStream content length may not allign with buffer size
+        // need to record how much data in buffer is legitimate
+        private int _bufferPopulatedLength;
+
+        private readonly int _tempRefilBufferSize;
```
tg-msft, line 27: Nit - Refill typo?

sdk/storage/Azure.Storage.Common/src/Shared/ClientsideEncryption/ClientSideEncryptorV2_0.cs
```diff
@@ -0,0 +1,199 @@
...
+            _keyEncryptionKey = options.KeyEncryptionKey;
+            _keyWrapAlgorithm = options.KeyWrapAlgorithm;
+        }
+
+        private void ValidateMembers()
+        {
+            if (_keyEncryptionKey == default || _keyWrapAlgorithm == default)
+            {
+                throw Errors.ClientSideEncryption.MissingRequiredEncryptionResources(nameof(_keyEncryptionKey), nameof(_keyWrapAlgorithm));
+            }
```
tg-msft, line 39: Can we do this validation just once in the .ctor?

sdk/storage/Azure.Storage.Common/src/Shared/ClientsideEncryption/IAuthenticatedCryptographicTransform.cs
```diff
@@ -0,0 +1,35 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Azure.Storage.Cryptography.Models;
+
+namespace Azure.Storage.Cryptography
+{
+    internal interface IAuthenticatedCryptographicTransform : IDisposable
```
tg-msft, line 13: Could we get away with making these `abstract class`es instead?  It probably doesn't matter as much for shared source, but versioning is always more flexible with an `abstract class`.

tg-msft (approved):
Looks great!  I left a few nits, but nothing critical that needs to be addressed immediately.

sdk/storage/Azure.Storage.Common/src/Shared/UriExtensions.cs
jaschrep-msft: Netstandard2.1 does not like `IndexOf()` and warns. I can try and put invariant back in there.

jaschrep-msft:
@tg-msft Good catches! All internals though that will need to wait for GA. I am already writing my own list of work to track and I will add these to the list. Issue will be posted before merge.

seanmcc-msft (approved):

sdk/storage/Azure.Storage.Common/src/Shared/ClientsideEncryption/AuthenticatedRegionCryptoStream.cs
```diff
@@ -0,0 +1,290 @@
...
+            {
+                bufferSize = transform.NonceLength + regionDataSize + transform.TagLength;
+                _tempRefilBufferSize = regionDataSize;
+            }
+            else
+            {
+                throw Errors.InvalidArgument(nameof(transform));
+            }
+
+            _buffer = new byte[bufferSize];
```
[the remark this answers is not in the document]
jaschrep-msft: This particular buffer is long-living (life of the stream) and we need it to be exact size (not guaranteed by array pool) or the math gets even more complex. But there are several other temporary arrays throughout this class that can be taken from the array pool.

sdk/storage/Azure.Storage.Common/src/Shared/ClientsideEncryption/IAuthenticatedCryptographicTransform.cs
jaschrep-msft: I'm actually a bit afraid of flexibility here for this shared source. This is delicate code and I wouldn't want any assumptions to be made when adding a new piece of functionality.

PR #28885, opened 2022-05-19
Encryption Aaed Gcm base

amnguye:
Looking to support Encrypting and Decrypting using the AAED GCM in the Storage Client SDK. This is just the layer to encrypt and decrypt, not to expose externally. No public API changes. All code is copied from the dotnet runtime library.

https://github.com/dotnet/runtime/blob/main/src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/AesGcm.Windows.cs
https://github.com/dotnet/runtime/blob/main/src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/AeadCommon.Windows.cs

This would only be called in the case that a lower framework version that can't use the System.Security.Cryptography.AesGcm dll provided by the dotnet runtime.


amnguye:
https://github.com/dotnet/runtime/blob/main/src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/AesGcm.Windows.cs
https://github.com/dotnet/runtime/blob/main/src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/AeadCommon.Windows.cs

sdk/storage/Azure.Storage.Common/src/Azure.Storage.Common.csproj
```diff
@@ -16,6 +16,7 @@
       Microsoft Azure Storage quickstarts and tutorials - https://docs.microsoft.com/en-us/azure/storage/
       Microsoft Azure Storage REST API Reference - https://docs.microsoft.com/en-us/rest/api/storageservices/
     </Description>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
```
jaschrep-msft, line 19: We should probably be adding this code to the ignore list for compiling common, instead explicitly compiling it into Blobs and Queues, since that is where it will end up.

We can probably also compile it into Common.Tests if need be.

sdk/storage/Azure.Storage.Common/src/Shared/AesGcm/AesGcmWindows.cs
```diff
@@ -0,0 +1,146 @@
...
+            ReadOnlySpan<byte> tag)
+        {
+            if (plaintext.Length != ciphertext.Length)
+                throw new ArgumentException("SR.Cryptography_PlaintextCiphertextLengthMismatch");
+
+            if (!nonce.Length.IsLegalSize(NonceByteSizes))
+                throw new ArgumentException("SR.Cryptography_InvalidNonceLength", nameof(nonce));
+
+            if (!tag.Length.IsLegalSize(TagByteSizes))
+                throw new ArgumentException("SR.Cryptography_InvalidTagLength", nameof(tag));
```
bartonjs, line 96: For many of these exception messages you should be able to get away with asserts, as it's non-public API in your copy.  But things like a tag mismatch (data tampering, or a wrong key or wrong nonce) you might want to actually populate.

All of the `SR.something` values are defined in the same file. 
 This one is at https://github.com/dotnet/runtime/blob/032a7dcbe1056493e8bab51e6b5b9503de727273/src/libraries/System.Security.Cryptography/src/Resources/Strings.resx#L417-L419, but they're all in there... generally alphabetically.

sdk/storage/Azure.Storage.Common/src/Azure.Storage.Common.csproj
```diff
@@ -30,6 +30,7 @@
   <ItemGroup>
     <Compile Remove="Shared\*.cs" />
     <Compile Remove="Shared\ClientsideEncryption\*.cs;Shared\ClientsideEncryption\Models\*.cs" />
+    <Compile Remove="Shared\AesGcm\*.cs;Shared\AesGcm\AllowUnsafeBlocks\*.cs" />
```
kasobol-msft, line 33: nit. would `Shared\AesGcm\**\*.cs` work?

kasobol-msft (commented):
lgtm

sdk/storage/Azure.Storage.Common/tests/AesGcmTests/Azure.Storage.Common.Tests.AesGcm.csproj
```diff
@@ -0,0 +1,37 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFrameworks>$(RequiredTargetFrameworks)</TargetFrameworks>
```
amnguye, line 3: I meant to only put net461, since the other target frameworks are not meant to be added. But we're not allowed to compile without the other ones. If someone can point me to only targetting certain frameworks let me know.

sdk/storage/Azure.Storage.Common/tests/AesGcmTests/Azure.Storage.Common.Tests.AesGcm.csproj
```diff
@@ -0,0 +1,37 @@
...
+  </PropertyGroup>
+  <PropertyGroup>
+    <AssemblyTitle>Microsoft Azure.Storage.Common client library tests</AssemblyTitle>
+    <IsPackable>false</IsPackable>
+    <RootNamespace>Azure.Storage.Tests</RootNamespace>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+  </PropertyGroup>
+  <ItemGroup>
+    <ProjectReference Include="$(MSBuildThisFileDirectory)..\..\src\Azure.Storage.Common.csproj" />
+    <ProjectReference Include="..\..\..\..\core\Azure.Core.TestFramework\src\Azure.Core.TestFramework.csproj" />
```
amnguye, line 13: fix the path for this

sdk/storage/Azure.Storage.Common/src/Shared/AesGcm/AllowUnsafeBlocks/AeadCommon.cs
```diff
@@ -0,0 +1,130 @@
...
+                    ciphertextBytes,
+                    ciphertext.Length,
+                    out int ciphertextBytesWritten,
+                    0);
+
+                Debug.Assert(plaintext.Length == ciphertextBytesWritten);
+
+                if (ntStatus != Interop.BCrypt.NTSTATUS.STATUS_SUCCESS)
+                {
+                    throw Interop.BCrypt.CreateCryptographicException(ntStatus);
```
amnguye, line 58: Couldn't find an exception for this, I'm guessing if we get here it's unexpected?

sdk/storage/Azure.Storage.Common/src/Shared/AesGcm/AllowUnsafeBlocks/AeadCommon.cs
```diff
@@ -0,0 +1,130 @@
...
+                        if (clearPlaintextOnFailure)
+                        {
+                            plaintext.Clear();
+                            //CryptographicOperations.ZeroMemory(plaintext);
+                        }
+                        Errors.CryptographyAuthTagMismatch();
+                        return;
+                    default:
+                        //throw CreateCryptographicException(ntStatus);
+                        throw new CryptographicException($"Error: 0x{(int)ntStatus:X8}");
```
amnguye, line 115: Same here

bartonjs: Yeah, both here and Encrypt represent "we got an error code that we don't understand".  For Encrypt, the documented return codes are https://docs.microsoft.com/en-us/windows/win32/api/bcrypt/nf-bcrypt-bcryptencrypt#return-value

All of the documented failures were checked by us in preconditions to make for nicer exceptions.  Anything else is a surprise.

What we do in the framework is call into FormatMessage to ask the OS for the text associated with the error code.  It's probably good enough to share just the error code, and if people see them, get confused by them, and report them, then maybe there are some more returns worth handling.

terrajobst:
Which frameworks are you targeting? Looks like there is an MSBuild property for that but I can't seem to find the logic that defines it.

amnguye:
> Which frameworks are you targeting? Looks like there is an MSBuild property for that but I can't seem to find the logic that defines it.

Right now we're targets .NET Core 3.1,  .NET 4.6.1, .NET 6.0, but my aim was to only do .NET 4.6.1. Unfortunately we can't build unless targetting all of them.

sdk/storage/Azure.Storage.Common/tests/AesGcmTests/Azure.Storage.Common.Tests.AesGcm.csproj
KrzysztofCwalina: It might be good to run tests on runtimes that don't support this feature and simply verify that we throw the right exceptions.

KrzysztofCwalina:
It would be good to provide a description of this PR.

sdk/storage/Azure.Storage.Common/src/Shared/AesGcm/AesGcmWindows.cs
```diff
@@ -98,6 +96,14 @@ internal void Decrypt(byte[] nonce, byte[] ciphertext, byte[] tag, byte[] plaint
...
         private static void ThrowIfNotSupported()
         {
+            // Added to prevent using if framework version can use the provided and supported
+            // AES by System.Security.Cryptography library instead
+#if NETSTANDARD2_1_OR_GREATER || NETCOREAPP3_0_OR_GREATER
+            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                throw Errors.CryptographyAlgorithmNotSupported();
+            }
+#endif
```
kasobol-msft, line 106: is this correct ?

Should this always throw on netstandard2.1+ and core 3.0+ ?
And then only throw on non-windows otherwise ?

Would it make sense to compile types in the Shared/AesGcm only if we target netstandard2.0 and carve them out if we compile netstandard2.1+ or core3+? I.e. don't even include these classes in dll that targets platform which has aesgcm built in.
See https://github.com/Azure/azure-sdk-for-net/blob/789a2416e4d7c346f43e51487c0097333c07812a/sdk/core/Azure.Core/src/Pipeline/HttpWebRequestTransport.cs#L20 for example.

amnguye: The intent was to prevent the user from calling into this and use the dotnet runtime AES GCM (or linux preferred version of GCM algorithm).

But you're right, we should just prevent it from compiling in netstandard2.1+ and core3, so they don't have a chance of calling into this code if they have a newer version.

Give me a bit, I'm going to remove this and target the respective platforms.

sdk/storage/Azure.Storage.Common/src/Shared/Errors.cs
```diff
@@ -65,5 +66,57 @@ internal static void VerifyStreamPosition(Stream stream, string streamName)
...
+            {
+                throw new ArgumentNullException($"Param: \"{paramName}\" is null");
+            }
+        }
+
+        internal static void CheckCryptKeySize(int keySizeInBytes)
+        {
+            if (keySizeInBytes != (128 / 8) && keySizeInBytes != (192 / 8) && keySizeInBytes != (256 / 8))
+            {
+                throw new CryptographicException("SR.Cryptography_InvalidKeySize");
```
bartonjs, line 82: Still says "SR."

bartonjs (approved):
The important tests are present (known success cases, and that on failure cases the unverified plaintext is not released), and the code has generally the right shape.

Since it's just supposed to be a copy of what we have in dotnet/runtime, I didn't look too deeply.

kasobol-msft (approved):

sdk/storage/Azure.Storage.Common/src/Shared/AesGcm/AesGcmWindows.cs
```diff
@@ -0,0 +1,143 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+using System;
+using System.Security.Cryptography;
+
+namespace Azure.Storage
+{
+    /// <summary>
+    /// Taken from
+    /// https://github.com/dotnet/runtime/blob/main/src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/AeadCommon.Windows.cs
```
tg-msft, line 12: I don't see the `AesGcmWindows` class at this link.  Also, can we use a specific SHA (you can easily get this by hitting the Y key on github's site to update the browser's path from `main`)?  For example, it gives me something more like https://github.com/dotnet/runtime/blob/89cee6e3fbd21ad79be00a735a017f0c932c1501/src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/AeadCommon.Windows.cs for this path.

sdk/storage/Azure.Storage.Common/src/Shared/AesGcm/AesGcmWindows.cs
```diff
@@ -0,0 +1,143 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+using System;
+using System.Security.Cryptography;
+
+namespace Azure.Storage
```
tg-msft, line 7: Should we put this all in a common namespace more like `Azure.Storage.Shared.AesGcm` or something so it's easier to lump all these pieces together via intellisense when browsing code in the future?

sdk/storage/Azure.Storage.Common/src/Shared/AesGcm/AesGcmWindows.cs
tg-msft: Also, does this comment apply to the class or the entire file?  We might consider moving it to the top of the file if it applies to everything.

tg-msft (commented):
Sorry I'm a little late, but I wanted to add a few suggestions to make this easier to maintain.

sdk/storage/Azure.Storage.Common/src/Shared/AesGcm/AesGcmWindows.cs
amnguye: I'm going to get this in one quick PR, the other comments will be in another PR.

amnguye: https://github.com/Azure/azure-sdk-for-net/pull/29067

sdk/storage/Azure.Storage.Common/src/Shared/AesGcm/AesGcmWindows.cs
```diff
@@ -0,0 +1,146 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+using System;
+using System.Runtime.InteropServices;
+using System.Security.Cryptography;
+
+namespace Azure.Storage
+{
+#if !(NETSTANDARD2_1_OR_GREATER || NETCOREAPP3_0_OR_GREATER)
```
weshaggard, line 9: Did we plan to explicitly target netstandard2.1 as well? I don't see use adding that target to any of the libraries.

weshaggard:
>Right now we're targets .NET Core 3.1, .NET 4.6.1, .NET 6.0, but my aim was to only do .NET 4.6.1. Unfortunately we can't build unless targetting all of them.

Those are the target frameworks we test against but it doesn't look like you added any new target frameworks in the libraries so that means we only end up with a netstandard2.0 implementation that includes this code. From the email it seems like we also wanted to add a netstandard2.1 target in the storage libraries that need this. For reference here are our default target frameworks, https://github.com/Azure/azure-sdk-for-net/blob/main/eng/Directory.Build.Common.props#L119 which for client libraries is just netstandard2.0.

sdk/storage/Azure.Storage.Common/src/Shared/AesGcm/AesGcmWindows.cs
terrajobst: My view is that you probably should. Otherwise you'll limit GCM for platforms implementing .NET Standard 2.1 to Windows only (which basically disables it for Xamarin) and would only work in Mono and Unity when running on Windows. Whether or not that's acceptable is up to you; adding a target definitely increases package size.

sdk/storage/Azure.Storage.Common/src/Azure.Storage.Common.csproj
```diff
@@ -30,6 +30,7 @@
   <ItemGroup>
     <Compile Remove="Shared\*.cs" />
     <Compile Remove="Shared\ClientsideEncryption\*.cs;Shared\ClientsideEncryption\Models\*.cs" />
+    <Compile Remove="Shared\AesGcm\**\*.cs" />
```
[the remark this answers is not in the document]
amnguye: All Shared classes within the Azure.Storage namespace are not included in building the Azure.Storage package. Not every class is required for every storage package. Also most of these shared classes are internal (to prevent the user from being able to call them directly), so we have to directly include the respective shared common classes in each storage package that requires it anyways.

sdk/storage/Azure.Storage.Common/src/Shared/AesGcm/AesGcmWindows.cs
amnguye: @jaschrep-msft

jaschrep-msft: This statement is fine regardless of what our final targeting story is. Regardless of our targeting story, this prevents us from accidentally compiling this into any future splits, standard or official implementation.

weshaggard: I agree that this condition is correct but right now it is a noop because we aren't targeting any of those higher platforms.

sdk/storage/Azure.Storage.Common/src/Azure.Storage.Common.csproj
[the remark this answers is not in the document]
weshaggard: Ok so we haven't done the work yet to pull these into blobs and queues yet.

Issue #28895, opened 2022-05-20
[Core] Build & Test (all tests for net6.0) failing in nightly runs


v-yilinhu:
Core nightly test runs are failing with:

> Error message
> Starting test execution, please wait...
> A total of 1 test files matched the specified pattern.
> NUnit Adapter 4.2.0.0: Test execution started
> Running selected tests in /mnt/vss/_work/1/s/artifacts/bin/Microsoft.Azure.Core.Spatial.NewtonsoftJson.Tests/Debug/net6.0/Microsoft.Azure.Core.Spatial.NewtonsoftJson.Tests.dll
>    NUnit3TestExecutor discovered 20 of 20 NUnit test cases using Current Discovery mode, Non-Explicit run
>   Passed CanConvert [23 ms]
>   Passed ReadBadJson("[]","Deserialization failed: Expected token 'StartObject'.") [20 ms]
>   Passed ReadBadJson("{}","Deserialization failed: Required GeoJson property 'type' not found.") [< 1 ms]
>   Passed ReadBadJson("{\"type\":\"point\"}","Deserialization failed: GeoJson property 'type' values are case sensitive. Use 'Point' instead.") [1 ms]
>   Passed ReadBadJson("{\"type\":\"Polygon\"}","Deserialization failed: Required GeoJson property 'coordinates' not found.") [< 1 ms]
>   Passed ReadBadJson("{\"Type\":\"Point\"}","Deserialization failed: Required GeoJson property 'type' not found.") [< 1 ms]
>   Passed ReadBadJson("{\"type\":\"Point\",\"Coordinates\":[-121.726906,46.879967,2541.118]}","Deserialization failed: Required GeoJson property 'coordinates' not found.") [< 1 ms]
>   Passed ReadBadJson("{\"type\":\"Point\",\"coordinates\":-121.726906}","Deserialization failed: Expected token 'StartArray'.") [< 1 ms]
>   Passed ReadBadJson("{\"type\":\"Point\",\"coordinates\":[]}","Deserialization failed: Expected token 'Integer' or 'Float'.") [< 1 ms]
>   Passed ReadBadJson("{\"type\":\"Point\",\"coordinates\":[\"foo\"]}","Deserialization failed: Expected token 'Integer' or 'Float'.") [< 1 ms]
[...]

PR #28900, opened 2022-05-20
Fix IgnoreServiceErrorAttributeTests

JoshLove-msft:
Previously the ignore service error attribute would only be applied during Playback, which was not the intent.

Addresses https://github.com/Azure/azure-sdk-for-net/issues/28895

### DIFF
--- a/sdk/core/Azure.Core.TestFramework/src/RecordedTestAttribute.cs
+++ b/sdk/core/Azure.Core.TestFramework/src/RecordedTestAttribute.cs
@@ -125,27 +125,25 @@ namespace Azure.Core.TestFramework
 
         private class RecordedTestAttributeCommand : DelegatingTestCommand
         {
-            private readonly List<IgnoreServiceErrorAttribute> _ignoreServiceErrorAttributes;
             public RecordedTestAttributeCommand(TestCommand innerCommand) : base(innerCommand)
             {
-                Test test = innerCommand.Test;
-
-                // Check if there are any service errors we should ignore.
-                _ignoreServiceErrorAttributes = innerCommand.Test.GetCustomAttributes<IgnoreServiceErrorAttribute>(true).ToList();
-
-                // Check parents for service errors to ignore.
-                while (test.Parent is Test t)
-                {
-                    _ignoreServiceErrorAttributes.AddRange(t.GetCustomAttributes<IgnoreServiceErrorAttribute>(true));
-                    test = t;
-                }
             }
 
             public override TestResult Execute(TestExecutionContext context)
             {
                 if (IsTestFailed(context))
                 {
-                    foreach (IgnoreServiceErrorAttribute attr in _ignoreServiceErrorAttributes)
+                    // Check if there are any service errors we should ignore.
+                    var ignoreServiceErrorAttributes = innerCommand.Test.GetCustomAttributes<IgnoreServiceErrorAttribute>(true).ToList();
+
+                    // Check parents for service errors to ignore.
+                    var test = Test;
+                    while (test.Parent is Test t)
+                    {
+                        ignoreServiceErrorAttributes.AddRange(t.GetCustomAttributes<IgnoreServiceErrorAttribute>(true));
+                        test = t;
+                    }
+                    foreach (IgnoreServiceErrorAttribute attr in ignoreServiceErrorAttributes)
                     {
                         if (attr.Matches(context.CurrentResult.Message))
                         {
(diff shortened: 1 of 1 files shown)


sdk/core/Azure.Core.TestFramework/src/RecordedTestAttribute.cs
```diff
@@ -26,20 +26,30 @@ public TestCommand Wrap(TestCommand command)
             {
                 test = test.Parent;
             }
-            if (test.Fixture is RecordedTestBase fixture && fixture.Mode == RecordedTestMode.Playback)
+
+            if (test.Fixture is RecordedTestBase fixture)
             {
-                return new FallbackCommand(command);
+                return fixture.Mode == RecordedTestMode.Playback ? new PlaybackCommand(command) : new RecordedTestAttributeCommand(command);
```
christothes, line 32: This naming is a bit confusing. What do you think about something like:
```suggestion
                return fixture.Mode == RecordedTestMode.Playback ? new AutoReRecordingCommand(command) : new IgnoreServiceErrorsCommand(command);
```

I realize this may not make sense in the future if the capabilities of the IgnoreServiceErrorsCommand expands its functionality, but for now it makes things very clear.

JoshLove-msft: As for PlaybackCommand, I think we would only have at most 1 command per mode.

JoshLove-msft: What about RecordedTestAttributePlaybackCommand and RecordedTestAttributeCommand?

christothes: I was just thinking that commands should be discrete and named for their functionality rather than be defined by the recording mode they apply to.

christothes: I think that happens to be true today, but not because it must be true of all command functionality.

christothes (approved):

sdk/core/Azure.Core.TestFramework/src/RecordedTestAttribute.cs
JoshLove-msft: Totally agree - in the future it may be the case that this proposed strategy of using separate modes per command won't actually work, and we would need to re-think it at that time. But my hope is that we would add additional playback functionality into the Playback command.

sdk/core/Azure.Core.TestFramework/src/RecordedTestAttribute.cs
```diff
@@ -102,32 +111,44 @@ public override TestResult Execute(TestExecutionContext context)
...
                     {
-                        // Check if there are any service errors we should ignore.
-                        List<IgnoreServiceErrorAttribute> attributes = Test.GetCustomAttributes<IgnoreServiceErrorAttribute>(true).ToList();
+                        CheckForIgnoredServiceErrors(context);
+                    }
+                }
+                else
+                {
+                    CheckForIgnoredServiceErrors(context);
+                }
```
christothes, line 120: Could this be simplified to a single unconditional `CheckForIgnoredServiceErrors(context);` at the end?

JoshLove-msft: today is not my day

JoshLove-msft: Wait actually no, because we don't want to apply this if we fall into one of the other error blocks.

christothes: Would it make sense to return immediately from those other error blocks?

sdk/core/Azure.Core.TestFramework/src/RecordedTestAttribute.cs
```diff
@@ -14,7 +14,7 @@ namespace Azure.Core.TestFramework
 {
     [AttributeUsage(AttributeTargets.Method)]
     /// <summary>
-    /// This attribute replaces the [Test] attribute and will dynamically re-record recorded tests on failure.
+    /// This attribute replaces the [Test] attribute and will dynamically re-record recorded tests on failure when run in Playback mode.
```
jsquire, line 17: What happens if this is being run in Playback mode and there aren't test resources deployed to re-record?   I understand they'll fail, but would that potentially cause issues with the session records if someone blindly copies the failed result over?

JoshLove-msft: The session will not be saved into the session file if the test failed (unless the [SaveDebugRecordingsOnFailure](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core.TestFramework/src/RecordedTestBase.cs#L108) flag is set).

PR #28901, opened 2022-05-20
[TA] Add audience

### LINKED ISSUE #28674: Add **Audience property to allow the user to select the Azure cloud where the resource is located at
See issue https://github.com/Azure/azure-sdk-for-net/issues/18520

maririos:
Same as https://github.com/Azure/azure-sdk-for-net/pull/26583 it adds a `DocumentTranslationAudience` a user to select which Azure cloud the resource is located at.

Fixes https://github.com/Azure/azure-sdk-for-net/issues/28674

### DIFF
--- a/sdk/translation/Azure.AI.Translation.Document/CHANGELOG.md
+++ b/sdk/translation/Azure.AI.Translation.Document/CHANGELOG.md
@@ -5,11 +5,12 @@
 ### Features Added
 - Added optional parameter `categoryId` to the `DocumentTranslationInput.AddTarget`.
 - Added property `Ascending` to type `DocumentFilterOrder` and `TranslationFilterOrder`.
-- Renamed parameter `asc` to `ascending` in `TranslationFilterOrder` constructor.
+- `DocumentTranslationAudience` has been added to allow the user to select the Azure cloud where the resource is located.
 
 ### Breaking Changes
 - Renamed type `StorageInputType` to `StorageInputUriKind`.
 - Renamed property `StorageType` to `StorageUriKind` in `DocumentTranslationInput`.
+- Renamed parameter `asc` to `ascending` in `TranslationFilterOrder` constructor.
 
 ### Bugs Fixed
 
--- a/sdk/translation/Azure.AI.Translation.Document/src/DocumentTranslationClientOptions.cs
+++ b/sdk/translation/Azure.AI.Translation.Document/src/DocumentTranslationClientOptions.cs
@@ -31,9 +31,6 @@ namespace Azure.AI.Translation.Document
         {
             Version = version;
             AddLoggedHeadersAndQueryParameters();
-
-            //Default Audience to Azure Public Cloud
-            Audience ??= DocumentTranslationAudience.AzurePublicCloud;
         }
 
         internal string GetVersionString()
--- a/sdk/translation/Azure.AI.Translation.Document/src/DocumentTranslationClientOptions.cs
+++ b/sdk/translation/Azure.AI.Translation.Document/src/DocumentTranslationClientOptions.cs
@@ -31,6 +31,9 @@ namespace Azure.AI.Translation.Document
         {
             Version = version;
             AddLoggedHeadersAndQueryParameters();
+
+            //Default Audience to Azure Public Cloud
+            Audience ??= DocumentTranslationAudience.AzurePublicCloud;
         }
 
         internal string GetVersionString()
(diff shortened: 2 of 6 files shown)


kristapratico (approved):
LGTM. Just wondering - do DT tests currently run weekly in gov/mooncake? I'm not sure we've set that up in Python yet.

maririos:
> do DT tests currently run weekly in gov/mooncake?

We haven't turned them on although the infrastructure is there.

sdk/translation/Azure.AI.Translation.Document/src/DocumentTranslationClientOptions.cs
```diff
@@ -31,6 +31,9 @@ public DocumentTranslationClientOptions(ServiceVersion version = LatestVersion)
         {
             Version = version;
             AddLoggedHeadersAndQueryParameters();
+
+            //Default Audience to Azure Public Cloud
+            Audience ??= DocumentTranslationAudience.AzurePublicCloud;
```
kinelski, line 36: I know this probably comes from the implementation in other services, but this looks a bit weird to me.

Firstly, `Audience` is guaranteed to be null here since it has never been initialized, so the null check is not doing anything.

Secondly, in the `Audience` property docstrings, we state that we'll assume public cloud if null, so my guess is that we intended to keep it null and not initialize it in the constructor.

It seems that in TA we keep it null while in FR we didn't. They have not GA-ed yet, so we still have time to unify.

pallavit: Actually this pattern is present across SDKs not just cognitive space. For example ACR also has this.

@maririos I agree with @kinelski here. We can let this be null. we already have code in the Client to default to AzureCloud instead. Options values should be true to what the customer chose. 

Let me know if we did this based on the archboard feedback, in which case we should potentially change other SDK implementations too.

sdk/translation/Azure.AI.Translation.Document/CHANGELOG.md
```diff
@@ -5,11 +5,12 @@
 ### Features Added
 - Added optional parameter `categoryId` to the `DocumentTranslationInput.AddTarget`.
 - Added property `Ascending` to type `DocumentFilterOrder` and `TranslationFilterOrder`.
-- Renamed parameter `asc` to `ascending` in `TranslationFilterOrder` constructor.
+- `DocumentTranslationAudience` has been added to allow the user to select the Azure cloud where the resource is located.
```
pallavit, line 8: Do we have any pipelines enabled for national cloud testing is this something we should add to the backlog?

sdk/translation/Azure.AI.Translation.Document/src/DocumentTranslationClientOptions.cs
kinelski: Just found the original discussion in the FR PR: https://github.com/Azure/azure-sdk-for-net/pull/26741#discussion_r797788186

Per Jesse's comment:
> We want the options to default to null; the client is applying the default.

I think we forgot to update the code in some places. I'll open a GH issue to track the work in FR and we can align with TA (default `Audience` to null). What do you think?

pallavit: I agree. @maririos let us know if you think otherwise

sdk/translation/Azure.AI.Translation.Document/CHANGELOG.md
pallavit: Just saw Krista had asked the same question.

sdk/translation/Azure.AI.Translation.Document/src/DocumentTranslationClientOptions.cs
maririos: Thank you for the context! I will update

sdk/translation/Azure.AI.Translation.Document/CHANGELOG.md
maririos: https://github.com/Azure/azure-sdk-for-net/issues/29000

sdk/translation/Azure.AI.Translation.Document/src/DocumentTranslationClientOptions.cs
maririos: Updated

pallavit (approved):

PR #28909, opened 2022-05-21
Fix tests that depend on network/compute

fengzhou-msft:


sdk/network/Azure.ResourceManager.Network/src/Generated/Models/EndpointServiceResult.cs
```diff
@@ -19,10 +19,17 @@ public EndpointServiceResult()
...
+        internal EndpointServiceResult(ResourceIdentifier id, string name, string resourceType) : base(id)
         {
+            Name = name;
+            ResourceType = resourceType;
         }
+
+        /// <summary> Name of the endpoint service. </summary>
+        public string Name { get; }
+        /// <summary> Type of the endpoint service. </summary>
+        public string ResourceType { get; }
```
fengzhou-msft, line 33: Network has different combinations of the read-only attributes for `name` and `type` (id is always not read-only), this is the only case that both name and resourceType are read-only. I'm not going to create a XXResourceData base type just for this single case. We already have `NetworkResourceData` (name is not readonly, type is readonly), `NetworkWritableResourceData` (name and type are both not read-only) and they both have mutilple cases.

sdk/appconfiguration/ci.yml
```diff
@@ -21,6 +21,7 @@ pr:
     include:
     - sdk/appconfiguration/
     - sdk/resourcemanager/
+    - sdk/network/
```
fengzhou-msft, line 24: We can remove this after we GA network and change the reference in tests to PackageReference.

JoshLove-msft (approved):

Yao725 (approved):

PR #28925, opened 2022-05-23
[Event Hubs] Log Event Processor Handler

jsquire:
# Summary

The focus of these changes is to add logging around the invocation of the event processing handler when a batch of events is dispatched for processing.

### DIFF
--- a/sdk/eventhub/Azure.Messaging.EventHubs/src/Diagnostics/EventHubsEventSource.cs
+++ b/sdk/eventhub/Azure.Messaging.EventHubs/src/Diagnostics/EventHubsEventSource.cs
@@ -2527,6 +2527,81 @@ namespace Azure.Messaging.EventHubs.Diagnostics
         }
 
         /// <summary>
+        ///   Indicates that an <see cref="EventProcessor{TPartition}" /> instance has dispatched a batch of events to the processing handler.
+        /// </summary>
+        ///
+        /// <param name="partitionId">The identifier of the Event Hub partition associated with the event batch.</param>
+        /// <param name="identifier">A unique name used to identify the event processor.</param>
+        /// <param name="eventHubName">The name of the Event Hub that the processor is associated with.</param>
+        /// <param name="consumerGroup">The name of the consumer group that the processor is associated with.</param>
+        /// <param name="operationId">An artificial identifier for the handler invocation.</param>
+        /// <param name="eventCount">The number of events in the batch that was passed to the processing handler.</param>
+        ///
+        [Event(123, Level = EventLevel.Verbose, Message = "Started dispatching events to the processing handler for partition '{0}' by processor instance with identifier '{1}' for Event Hub: {2}, Consumer Group: {3}, and Operation Id: '{4}'.  Event Count: '{5}'")]
+        public virtual void EventProcessorProcessingHandlerStart(string partitionId,
+                                                                 string identifier,
+                                                                 string eventHubName,
+                                                                 string consumerGroup,
+                                                                 string operationId,
+                                                                 int eventCount)
+        {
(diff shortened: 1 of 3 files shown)


sdk/eventhub/Azure.Messaging.EventHubs/src/Diagnostics/EventHubsEventSource.cs
```diff
@@ -2852,6 +2927,112 @@ public void UnexpectedException(string errorMessage)
...
+        /// <param name="partitionId">The identifier of the Event Hub partition associated with the event batch.</param>
+        /// <param name="identifier">A unique name used to identify the event processor.</param>
+        /// <param name="eventHubName">The name of the Event Hub that the processor is associated with.</param>
+        /// <param name="consumerGroup">The name of the consumer group that the processor is associated with.</param>
+        /// <param name="operationId">An artificial identifier for the handler invocation.</param>
+        /// <param name="eventCount">The number of events in the batch that was passed to the processing handler.</param>
+        ///
+        [NonEvent]
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private unsafe void EventProcessorProcessingHandlerStartCore(int eventId,
```
JoshLove-msft, line 2945: Just to play devil's advocate, why not just define the core event with all string parameters and have it call directly into WriteEvent. I realize now that this same comment could be equally applied to the Service Bus events.

JoshLove-msft: Eh, I guess this could have a negative impact for someone wanting to consume the individual typed parameters rather than just look at logs. Carry on!

sdk/eventhub/Azure.Messaging.EventHubs/tests/Primitives/EventProcessorTests.PartitionProcessing.cs
```diff
@@ -119,6 +119,81 @@ public async Task ProcessEventBatchAsyncDoesNotInvokeTheHandlerWhenEmptyBatchesS
                 .Verify("OnProcessingEventBatchAsync", Times.Never(), EmptyBatch, partition, cancellationSource.Token);
         }
 
+        /// <summary>
+        ///   Verifies functionality of the <see cref="EventProcessor{TPartition}.ProcessEventBatchAsync" />
+        ///   method.
+        /// </summary>
+        ///
+        [Test]
+        public async Task ProcessEventBatchAsyncLogsWhenBatchesAreDispatched()
```
JoshLove-msft, line 128: Is it worth using the TestEventListener (whether in live tests or unit tests) to verify that the events are being logged as expected?

JoshLove-msft (approved):

sdk/eventhub/Azure.Messaging.EventHubs/tests/Primitives/EventProcessorTests.PartitionProcessing.cs
jsquire: Currently all of the verification that we're doing is unit tests only.  I've got an issue open to take a look at integrating `TestEventListener` for all of our logs across the project as part of Live coverage.  I think there's value in doing it as a whole, but not so much for just covering this change.

sdk/eventhub/Azure.Messaging.EventHubs/src/Diagnostics/EventHubsEventSource.cs
jsquire: Truthfully, I'm not sure.   That would seem to be a reasonable question.   I guess the best answer is that I just didn't think of it.

PR #28931, opened 2022-05-23
[Event Hubs] Processor Sample Tweak

jsquire:
# Summary

Adjusting some code samples and text to better illustrate recommended patterns for error handling.

### DIFF
--- a/sdk/eventhub/Azure.Messaging.EventHubs.Processor/tests/Snippets/Sample03_EventProcessorHandlersLiveTests.cs
+++ b/sdk/eventhub/Azure.Messaging.EventHubs.Processor/tests/Snippets/Sample03_EventProcessorHandlersLiveTests.cs
@@ -61,13 +61,16 @@ namespace Azure.Messaging.EventHubs.Tests.Snippets
             {
                 try
                 {
-                    // Process the event.
+                    // TODO:
+                    //   Process the event according to application needs.
                 }
                 catch
                 {
-                    // Take action to handle the exception.
-                    // It is important that all exceptions are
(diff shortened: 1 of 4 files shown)


sdk/eventhub/Azure.Messaging.EventHubs.Processor/tests/Snippets/Sample03_EventProcessorHandlersLiveTests.cs
```diff
@@ -432,19 +449,10 @@ async Task processErrorHandler(ProcessErrorEventArgs args)
                         return;
```
JoshLove-msft, line 449: Interesting, is OOM something that we expect users to explicitly handle?

JoshLove-msft (approved):

sdk/eventhub/Azure.Messaging.EventHubs.Processor/tests/Snippets/Sample03_EventProcessorHandlersLiveTests.cs
jsquire: Not as a general rule; if they hit an OOM condition, it's not likely that we're going to even be healthy enough to ping the handler.    I don't recall why this exists this way, other than to just demonstrate the types of things that can be done.   I'll take a look and see if it makes sense to keep it here.

jsquire: I don't think that it makes sense in context;  removed.

PR #28938, opened 2022-05-24
Add Archboard review page

pshao25:
Add Archboard review page

### DIFF
--- a/doc/DataPlaneCodeGeneration/EvolutionArchBoard.md
+++ b/doc/DataPlaneCodeGeneration/EvolutionArchBoard.md
@@ -1,0 +1,630 @@
+# DPG .NET Arch Board Review
+
+## Evolving the API
+
+### Basic REST Calls
+
+#### Takes RequestContent and returns Response
+
+**MetricsAdvisorAdministrationClient.CreateDataFeed**
+
+<table>
+<tr>
+<td> Protocol Method </td>
+</tr>
+<tr>
+<td>
+
+```csharp
+public virtual Response CreateDataFeed(RequestContent content, RequestContext context = null);
+```
+</td>
+</tr>
+<tr>
+<td> Convenience Method </td>
+</tr>
+<td>
+
+```csharp
+public virtual Response<DataFeed> CreateDataFeed(DataFeed dataFeed, CancellationToken cancellationToken = default)
+```
+</td>
+</tr>
+</table>
+
+[APIView](https://apiview.dev/Assemblies/Review/a9812a1691e54ab4bf52540c4e54d433?diffRevisionId=9260bdbf10e346b9ba8dac7399ea885b&doc=False&diffOnly=False&revisionId=66cd7343ae6c47988e3f8b3caf52617b#Azure.AI.MetricsAdvisor.Administration.MetricsAdvisorAdministrationClient), Lines 205-208.
+
+##### Customer UX: Calling CreateDataFeed
+
+<table>
+<tr>
+<td> Protocol Method </td>
+</tr>
+<tr>
+<td>
+  
+```c#
+MetricsAdvisorAdministrationClient adminClient = GetMetricsAdvisorAdministrationClient();
+  
+var body = new
+{
+    dataFeedName = "dataFeedName",
+    dataSourceType = "AzureApplicationInsights"
+    dataStartFrom = "2021-11-17T16:29:56.5770502+00:00"
+    granularityName = "Weekly"
+    metrics = new [] {new {metricName="metricNme"}}
+};   
+Response response = await adminClient.CreateDataFeedAsync( RequestContent.Create(body), new RequestContext());
+
+JsonElement createBodyJson = JsonDocument.Parse(response.Content).RootElement;
+var name = createBodyJson.GetProperty("dataFeedName").GetString();
+
+```
+</td>
+</tr>
+<tr>
+<td> Convenience Method </td>
+</tr>
+<td>
+  
+```c#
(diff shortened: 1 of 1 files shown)


doc/DataPlaneCodeGeneration/EvolutionArchBoard.md
```diff
@@ -0,0 +1,630 @@
+# DPG .NET Arch Board Review
+
+## Evolving the API
+
+### Basic REST Calls
```
chunyu3, line 5: Do you use the content from Anne's gist for the Basic and Paging block? It is not same as what we show in the review.

doc/DataPlaneCodeGeneration/EvolutionArchBoard.md
```diff
@@ -0,0 +1,630 @@
...
+```C#
+public virtual Response DeleteDeployment(string deploymentId, string groupId = null, RequestContext context = null);
+```
+Breaking scenario in v<sub>i+1</sub>
+```C#
+Response response = DeleteDeployment(groupId, deploymentId);
+```
+Solution in v<sub>i+1</sub>
+
+Solve it manually.
```
lirenhe, line 610: As i know, there is a feedback in the review that we should report error instead of ask people to solve it manually, do I understand correctly?

pshao25: It is correct we should not solve it manually. But the solution still needs discussion. Report error is a possible solution. 
Do we need to update this page with the feedback? There are many things still need to discuss.

doc/DataPlaneCodeGeneration/EvolutionArchBoard.md
```diff
@@ -4,183 +4,567 @@
...
 
-```csharp
+#### Scenario: no required parameters
+
+If your convenience method has the same parameter list as that of the protocol method, the initial version of convenience method could have ambiguity with the protocol method. To resolve that issue we will append suffix (such as `Value` or `Values`) to the initial method name as the convenience method name.
+
+- Protocol Method
+
+```c#
+public virtual Response GetConstitution(RequestContext context = null)
```
annelo-msft, line 30: I don't think we want to show the implementations for these, only the method signatures.  I'm concerned we might later want to change the implementation.  We would not be able to change the method signature because that would be breaking.

doc/DataPlaneCodeGeneration/EvolutionArchBoard.md
```diff
@@ -4,183 +4,567 @@
...
+}
+
+```
+
+- Convenience Method
+
+```c#
+public virtual Response<LedgerConstitution> GetConstitutionValue(CancellationToken cancellationToken = default)
+{
+    Response response = GetConstitution(new RequestContext { CancellationToken = cancellationToken });
```
annelo-msft, line 68: For example, this implementation is incorrect, because we would want to add a diagnostic scope that uses the public method name and suppress the inner scope.

doc/DataPlaneCodeGeneration/EvolutionArchBoard.md
```diff
@@ -4,183 +4,567 @@
...
+
+    LedgerConstitution constitution = LedgerConstitution.FromResponse(response);
+
+    return Response.FromValue(constitution, response);
+}
+
+public virtual async Task<Response<LedgerConstitution>> GetConstitutionValueAsync(CancellationToken cancellationToken = default)
+{
+    Response response = await GetConstitutionAsync(new RequestContext { CancellationToken = cancellationToken }).ConfigureAwait(false);
```
annelo-msft, line 77: This may not be the way we want to create a RequestContext instance going forward -- we will probably want to change it so that there are no allocations.

annelo-msft (commented):
@lirenhe, @pshao25, I'm not sure I understand the goal of putting the arch board review page in our DPG docs folder.  This folder is for docs to help partner teams author DPG libraries.  This may be useful in that, but if so, it would need an introduction telling people how to use it.  Would it instead make sense to add this to the comments of the arch board meeting, where it will be in context?

lirenhe:
> @lirenhe, @pshao25, I'm not sure I understand the goal of putting the arch board review page in our DPG docs folder. This folder is for docs to help partner teams author DPG libraries. This may be useful in that, but if so, it would need an introduction telling people how to use it. Would it instead make sense to add this to the comments of the arch board meeting, where it will be in context?

@annelo-msft, the information we prepared for review is great and I don't want to lost them. Currently we put all information in the gist which has the following limitation:
1. difficult to discover, I need to remember that link to get the content. 
2. Contribute to the existing content is difficult. Submit PR in gist to update the content is not allowed.
3. In general, we should put materials in company storage instead of under our personal account.

That is why I want to put them in this place for easy modification in the future and easy to find for others' reference. 

I understand your concern but I am not sure whether adding this to the comments of the arch board meeting would solve the above limitation.
As it is for mainly our internal usage,  how about creating a word doc and put it into our team OneNote?

pshao25:
@lirenhe @annelo-msft How about putting at internal [DevOps wiki page](https://dev.azure.com/azure-sdk/internal/_wiki/wikis). We could contribute, comment, check the history and organize.

annelo-msft:
Yes, I like the idea of being able to submit PRs and see the modification history.

Ideally, it would still be accessible from the arch board review issue, as that is the natural place that people would want to discover it from, I think.  @tg-msft, is there a precedent for how arch board reviewed materials are typically stored?  @srnagar, @iscai-msft, @joheredi, where did you store your materials for the RLC/DPG arch boards?

lirenhe:
Let's hold the merge of this PR and leverage it for discussion to make sure the content are correct and up-to-date.

PR #28947, opened 2022-05-24
Update developer-driven evolution doc with guidance on distributed tracing

### LINKED ISSUE #2115: Update developer-driven evolution doc with guidance on distributed tracing
In grow-up scenarios, convenience layer will call protocol method. Both of them might need to have a diagnostic scope. We need to decide what the scope should be like. See whole conversation here:
https://github.com/Azure/azure-sdk-for-net/pull/27363#discussion_r822152875

pshao25:
Fix https://github.com/Azure/autorest.csharp/issues/2115


doc/DataPlaneCodeGeneration/DeveloperDrivenEvolution.md
```diff
@@ -97,7 +97,28 @@ public virtual async Task<Response<MetricFeedback>> GetMetricFeedbackValuesAsync
...
+In the implementation, you should only have one `DiagnosticScope` inside instead of creating `DiagnosticScope` for both protocol method and convenience method. Reasons are:
+- We want to reduce the telemetry costs.
+- Trace should correspond to logical operations and not implementations
+
+To achieve this, our guidance is:
+- Don't open a scope in the grow-up method if it is just a simple wrapper (i.e. if the only thing in the convenience API is serialization and deserialization and the protocol method and convenience API have the same name). Logging will help developers debug any exceptions thrown from the convenience method.
+- If the grow-up method diverges much from the protocol method (i.e. if they have different names or code in the convenience method could take much time), do open a scope in the convenience API and turn on suppression of inner scopes.
+- If you want to add attributes or otherwise modify scopes, do that in the convenience method.
+
+To turn on the suppression, pass `ture` to `ClientDiagnostics` in the client constructor.
```
lirenhe, line 111: typo?  true.

doc/DataPlaneCodeGeneration/DeveloperDrivenEvolution.md
```diff
@@ -97,7 +97,28 @@ public virtual async Task<Response<MetricFeedback>> GetMetricFeedbackValuesAsync
...
+```C#
+public MetricsAdvisorClient(Uri endpoint, MetricsAdvisorKeyCredential credential, MetricsAdvisorClientsOptions options)
+{
+    ...
+    ClientDiagnostics = new ClientDiagnostics(options, true);
```
lirenhe, line 116: Don't quite understand about this.  As MetricsAdvisorClient is generated, do you mean we need to modify the generated code?

doc/DataPlaneCodeGeneration/DeveloperDrivenEvolution.md
```diff
@@ -109,6 +130,7 @@ namespace Azure.AI.MetricsAdvisor
     {
         public virtual async Task<Response> GetMetricFeedbackAsync(Guid feedbackId, RequestContext context = null)
         {
+            using var scope = ClientDiagnostics.CreateScope("MetricsAdvisorClient.GetMetricFeedback");
```
lirenhe, line 133: Please indicate there are other logic in this method instead of just 1 line of code.

doc/DataPlaneCodeGeneration/DeveloperDrivenEvolution.md
pshao25: It is generated. Updated.

doc/DataPlaneCodeGeneration/DeveloperDrivenEvolution.md
```diff
@@ -97,7 +97,28 @@ public virtual async Task<Response<MetricFeedback>> GetMetricFeedbackValuesAsync
 If the suggested rename does not apply to your scenario, you could pick the best name fitting your API.
 
 ## Implement the method
-You could call the protocol method inside your convenience method and map the returned raw response to model. See different scenarios and corresponding examples below. 
+You could call the protocol method inside your convenience method and map the returned raw response to model.
```
annelo-msft, line 100: Let's start this section with a line:

Inside a convenience method, you will do the following:
1. Optionally open a `DiagnosticScope`
2. Serialize any input model to `RequestContent`
3. Call the protocol method
4. Deserialize any response body to Response<T>

annelo-msft: Let's then break up the guidance that follows into sections that are numbered for each of these steps.

doc/DataPlaneCodeGeneration/DeveloperDrivenEvolution.md
```diff
@@ -97,7 +97,28 @@ public virtual async Task<Response<MetricFeedback>> GetMetricFeedbackValuesAsync
...
+
+In the implementation, you should only have one `DiagnosticScope` inside instead of creating `DiagnosticScope` for both protocol method and convenience method. Reasons are:
```
annelo-msft, line 102: Let's change this to something like:

The protocol method opens a diagnostic scope for the operation.  This can be used by the convenience method in many cases.   Only open a `DiagnosticScope` if:
- The convenience method has a different name than the protocol method.
- There is significant logic in the convenience method beyond model serialization/deserialization.
- If you need to add attributes or otherwise modify the diagnostic scope.

If a diagnostic scope is added in the convenience method, it should be suppressed in the inner protocol method so that there is only one distributed tracing span corresponding to the public method invocation.  This should happen by default in DPG clients, where `suppressNestedClientActivities` is set to true in the `ClientDiagnostics` constructor.

doc/DataPlaneCodeGeneration/DeveloperDrivenEvolution.md
annelo-msft: If the code is generated, we probably don't need to show it in this document, since this document is primarily focused on convenience methods.  I think mentioning that this is the default without showing the code is sufficient.

doc/DataPlaneCodeGeneration/DeveloperDrivenEvolution.md
```diff
@@ -122,6 +146,9 @@ namespace Azure.AI.MetricsAdvisor
         // Suggest appending value to method name
         public virtual async Task<Response<MetricFeedback>> GetMetricFeedbackValueAsync(Guid feedbackId, CancellationToken cancellationToken = default)
         {
+            // Open a new scope here and the inner scope will be suppressed
```
annelo-msft, line 149: It would be good to mention in the comment that we open the scope in the convenience method because it has a different name.

doc/DataPlaneCodeGeneration/DeveloperDrivenEvolution.md
```diff
@@ -157,6 +187,7 @@ namespace Azure.AI.MetricsAdvisor
         // Can use same method name here
         public virtual async Task<Response<MetricFeedback>> CreateMetricFeedbackAsync(MetricFeedback feedback, CancellationToken cancellationToken = default)
         {
+            // We don't open a new scope here because it is just a wrapper
```
annelo-msft, line 190: Instead of "just a wrapper," let's say it's because 1) it has the same name as the protocol method, 2) it does not add any logic beyond serialization/deserialization and 3) no attributes are added to the span.

doc/DataPlaneCodeGeneration/DeveloperDrivenEvolution.md
```diff
@@ -196,8 +228,8 @@ namespace Azure.AI.MetricsAdvisor
         // Suggest appending values to a plural method name
         public virtual AsyncPageable<MetricFeedback> GetMetricFeedbacksValuesAsync(Guid feedbackId, CancellationToken cancellationToken = default)
         {
-            // Call protocol method
-            AsyncPageable<BinaryData> pageableBinaryData = GetMetricFeedbacksAsync(feedbackId, new RequestContext() { CancellationToken = cancellationToken });
+            // Call internal paging implementation by passing the new scope name, the inner scope will be suppressed
```
annelo-msft, line 231: Lets also add here that we need to pass this because the convenience method has a different name from the protocol method.

annelo-msft (commented):
Where we show the convenience method implementations in this doc, where we are creating new RequestContext values and setting the CancellationToken to the input ct, could we instead do:

`RequestContext context = (ct == CancellationToken.None) ? null : new { CancellationToken = ct };`
I have filed https://github.com/Azure/azure-sdk-for-net/issues/29015 to track adding a helper to RequestContext to standardize this instead.

@JoshLove-msft FYI

doc/DataPlaneCodeGeneration/DeveloperDrivenEvolution.md
```diff
@@ -97,7 +97,48 @@ public virtual async Task<Response<MetricFeedback>> GetMetricFeedbackValuesAsync
...
+ClientDiagnostics = new ClientDiagnostics(options, true);
+```
+which means the `DiagnosticScope` who is not the outmost one will be suppressed, and only the outmost one will be left in the telemetry.
+
+The generated protocol method opens a diagnostic scope for each operation. This scope can be used by the convenience method in many cases. Open a new `DiagnosticScope` only when:
+- The convenience method has a different name than the protocol method.
+- There is significant logic in the convenience method beyond model serialization/deserialization.
+- If you need to add attributes or otherwise modify the diagnostic scope.
+
+If a new diagnostic scope is added in the convenience method, the one in the protocol method should be suppressed so that there is only one distributed tracing span corresponding to the public method invocation. This should happen by default in generated clients, where `suppressNestedClientActivities` is set to `true` in the `ClientDiagnostics` constructor.
```
lirenhe, line 119: Will you put the sample code here to be consistent with other section?

using var scope = ClientDiagnostics.CreateScope("MetricsAdvisorClient.GetMetricFeedbackValue");

lirenhe (approved):
Left one comment, other part looks good to me.  Thanks

doc/DataPlaneCodeGeneration/DeveloperDrivenEvolution.md
pshao25: This code should be in the method, but here we are talking about the constructor. The code in the constructor should be like Line 110: ClientDiagnostics = new ClientDiagnostics(options, true);

pshao25:
Don't close

pshao25:
don't close

PR #28989, opened 2022-05-26
Fix some flakiness in SB extension tests

JoshLove-msft:


sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/tests/ServiceBusSessionsEndToEndTests.cs
```diff
@@ -792,22 +792,6 @@ public class DrainModeTestJobTopicBatch
...
-            {
-                // Wait until the drain operation begins, signalled by the cancellation token
-                int elapsedTimeMills = 0;
-                while (elapsedTimeMills < DrainWaitTimeoutMills && !cancellationToken.IsCancellationRequested)
-                {
-                    await Task.Delay(elapsedTimeMills += 500);
-                }
-
-                // Allow some time for the Service Bus SDK to start draining before returning
-                await Task.Delay(DrainSleepMills);
```
JoshLove-msft, line 807: Moved this entire class up into the base class, but this delay was removed as there is no need to wait for the SDK once the token is signaled - the SDK will wait for the function to finish.

jsquire (approved):

PR #29010, opened 2022-05-27
[Network] Address API review comments

fengzhou-msft:

### DIFF
--- a/sdk/network/Azure.ResourceManager.Network/api/Azure.ResourceManager.Network.netstandard2.0.cs
+++ b/sdk/network/Azure.ResourceManager.Network/api/Azure.ResourceManager.Network.netstandard2.0.cs
@@ -25,7 +25,7 @@ namespace Azure.ResourceManager.Network
         public System.Collections.Generic.IList<Azure.ResourceManager.Network.Models.ApplicationGatewayCustomError> CustomErrorConfigurations { get { throw null; } }
         public bool? EnableFips { get { throw null; } set { } }
         public bool? EnableHttp2 { get { throw null; } set { } }
-        public string Etag { get { throw null; } }
+        public Azure.ETag? Etag { get { throw null; } }
         public Azure.Core.ResourceIdentifier FirewallPolicyId { get { throw null; } set { } }
         public bool? ForceFirewallPolicyAssociation { get { throw null; } set { } }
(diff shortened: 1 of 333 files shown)


sdk/network/Azure.ResourceManager.Network/api/Azure.ResourceManager.Network.netstandard2.0.cs
```diff
@@ -868,14 +868,13 @@ public partial class ExpressRouteConnectionCollection : Azure.ResourceManager.Ar
         System.Collections.Generic.IEnumerator<Azure.ResourceManager.Network.ExpressRouteConnectionResource> System.Collections.Generic.IEnumerable<Azure.ResourceManager.Network.ExpressRouteConnectionResource>.GetEnumerator() { throw null; }
         System.Collections.IEnumerator System.Collections.IEnumerable.GetEnumerator() { throw null; }
     }
-    public partial class ExpressRouteConnectionData : Azure.ResourceManager.Network.Models.NetworkSubResource
+    public partial class ExpressRouteConnectionData : Azure.ResourceManager.Network.Models.NetworkResourceData
     {
-        public ExpressRouteConnectionData(string name) { }
```
Yao725, line 873: I'm wondering if this change has any side affect. Removing `name` from the ctor might make user miss the property in setting, but it's required in swagger definition. Can this API call succeed without `name` property?

fengzhou-msft: Good catch. The model is used in input body parameter and name was required before I transform the [swagger](https://github.com/Azure/azure-rest-api-specs/blob/ef8024500a08004364d3e0ff30091fc8b3786629/specification/network/resource-manager/Microsoft.Network/stable/2021-02-01/virtualWan.json#L4346-L4361). But there is already a connectionName path parameter. I will double check if name is really required.

fengzhou-msft: Found some live traffic and shows that `name` is actually not required in input: https://portal.azure-devex-tools.com/issues/completeness/detail?errorId=C785D1E5-C9FC-475C-9444-1551C32248DB

Yao725 (approved):

PR #28559, opened 2022-05-05
[FormRecognizer] Applying the IgnoreServiceError property

kinelski:
This PR makes our tests ignore a couple of service-side errors we have been facing for months. GitHub issues have been created to track those.

Fixes #26956.
Fixes #27498.

PR #29014, opened 2022-05-27
[FormRecognizer] Added IgnoreServiceError attribute to FormRecognizerSamples

kinelski:
Missed this scenario in the [previous PR](https://github.com/Azure/azure-sdk-for-net/pull/28559).


maririos (approved):
For a moment I was all confused on why to suppress it for the mock sample and then I realized the samples are all under the same class 😮

Issue #21884, opened 2021-06-15
[FEATURE REQ] Azure Functions Service Bus output - support setting message metadata


anthonychu:
**Library or service name.**
Microsoft.Azure.WebJobs.Extensions.ServiceBus
 
**Is your feature request related to a problem? Please describe.**

For out of process languages, including the .NET isolated worker, it's not possible to set metadata on output messages such as correlation ID. Provide a way to optionally set [these properties](https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.servicebus.message?view=azure-dotnet#properties) on output messages.

https://github.com/Azure/azure-functions-dotnet-worker/issues/474

Dave-Kiwi:
Should the library you quote be Microsoft.Azure.Functions.Worker.Extensions.ServiceBus in the case of the .Net Isolated Worker?

JoshLove-msft:
> Should the library you quote be Microsoft.Azure.Functions.Worker.Extensions.ServiceBus in the case of the .Net Isolated Worker?

I think so. This is being tracked in https://github.com/Azure/azure-functions-dotnet-worker/issues/496

anthonychu:
> Should the library you quote be Microsoft.Azure.Functions.Worker.Extensions.ServiceBus in the case of the .Net Isolated Worker?

No. This request is for the WebJobs extension to support metadata properties on output messages; this is language worker agnostic. Once this ability is there, each language can add support in a way that makes sense for their language.

JoshLove-msft:
So this is asking for a way to set these properties outside of the strong model type (similar to how you can bind to string versions of the properties for the trigger)?

anthonychu:
I think so. This a problem across several bindings and they should probably be addressed in a consistent manner. See also: 
* Queues: #21887
* Event Hubs: https://github.com/Azure/azure-functions-python-worker/issues/654

For example, one possible solution would be to add a prop that indicates that the worker is returning a message envelope instead of just the data itself.

```json
{
  "name": "$return",
  "type": "serviceBus",
  "queueName": "testqueue",
  "connection": "MyServiceBusConnection",
  "includeMetadata": true,
  "direction": "out"
}
```

JavaScript: 

```javascript
return {
  body: "my message",
  correlationId: "foo",
  userProperties: { /* ... */ }
}
```

 /cc @fabiocav @ankitkumarr @brettsam

fabiocav:
[...]

PR #29022, opened 2022-05-28
Support outputting the entire message envelope for out of proc functions

JoshLove-msft:
Contributes to https://github.com/Azure/azure-sdk-for-net/issues/21884

The attribute will need to be implemented in the out of process worker to light things up end-to-end.

### DIFF
--- a/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/ServiceBusAttribute.cs
+++ b/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/ServiceBusAttribute.cs
@@ -6,6 +6,8 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using Microsoft.Azure.WebJobs.Description;
 using Microsoft.Azure.WebJobs.ServiceBus;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Serialization;
 
 namespace Microsoft.Azure.WebJobs
 {
@@ -41,6 +43,14 @@ namespace Microsoft.Azure.WebJobs
             EntityType = entityType;
         }
 
+        // for testing of the internal IncludeMetadata property. Internal properties cannot be set on attributes outside of the constructor.
+        internal ServiceBusAttribute(string queueOrTopicName, bool includeMetadata, ServiceBusEntityType entityType = ServiceBusEntityType.Queue)
+        {
+            QueueOrTopicName = queueOrTopicName;
+            EntityType = entityType;
+            IncludeMetadata = includeMetadata;
+        }
+
         /// <summary>
         /// Gets the name of the queue or topic to bind to.
         /// </summary>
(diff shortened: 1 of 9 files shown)


sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/ServiceBusAttribute.cs
```diff
@@ -55,5 +65,13 @@ public ServiceBusAttribute(string queueOrTopicName, ServiceBusEntityType entityT
...
         /// </summary>
         public ServiceBusEntityType EntityType { get; set; }
+
+        /// <summary>
+        /// This property indicates whether the string being sent from out of process functions
+        /// includes the message metadata. By default this is false, meaning the string represents
+        /// only the message body.
+        /// </summary>
+        [JsonProperty]
+        internal bool IncludeMetadata { get; set; }
```
JoshLove-msft, line 75: I _think_ this should work even though it is internal. My understanding is that the attributes for out of process .NET and for other languages get serialized using NewtonSoft, sent to the host process, and then deserialized into this model also using NewtonSoft.

/cc @fabiocav @mattchenderson

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/ServiceBusMessageConverter.cs
```diff
@@ -0,0 +1,120 @@
...
+
+using System;
+using System.Globalization;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using Azure.Messaging.ServiceBus;
+
+namespace Microsoft.Azure.WebJobs.ServiceBus
+{
+    internal class ServiceBusMessageConverter : JsonConverter<ServiceBusMessage>
```
jsquire, line 12: It may be helpful to have some unit tests for different scenarios rather than relying on the end-to-end tests.  That may help to improve the dev loop and better isolate a specific regression since the E2E test has so much going on and needs to be run Live.

jsquire (approved):
The serializer and Service Bus bits look good to me, though there seems to be a minor type issue for some platforms.  I didn't see anything that jumped out at me in the bindings portion, but I don't feel like I've got enough context for how that works to be confident saying that it's good to go.     You'll probably want to make sure that one of the Functions folks is able to give it a look.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/ServiceBusAttribute.cs
JoshLove-msft: I dug into this, and the property does need to be public as it is consumed here - https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeConverter.cs#L72

Issue #25991, opened 2021-12-17
[MQ] Suggested Refactorings to unify LROs


annelo-msft:

annelo-msft:
Suggestion in https://github.com/Azure/azure-sdk-for-net/issues/29072

annelo-msft:
Issue in NextLinkOperationImplementation: https://github.com/Azure/azure-sdk-for-net/issues/44640

PR #29033, opened 2022-05-30
Add scope of waitForCompletion in LRO

### LINKED ISSUE #2236: WaitForCompletion scope is not included in the LowLevelFuncOperation
We have below code ([full version](https://github.com/Azure/autorest.csharp/blob/86230daa6868fb0eb63939fff4801fce5efcb4af/test/AutoRest.TestServerLowLevel.Tests/dpg-customization.cs#L101-L105)):
```C#
Operation<BinaryData> result = await new DPGClient(Key, host, null).LroAsync(WaitUntil.Started, "raw");
await result.WaitForCompletionAsync();
```
According to https://github.com/Azure/autorest.csharp/issues/2233#issuecomment-1127819652:

**Expected behavior:** we should have scope `DPGClient.Lro.WaitForCompletion`
**Actual behavior:** now we have scope `DPGClient.Lro.UpdateStatus`

**Root cause:** 
We return LowLevelFuncOperation which has OperationInternalBase inside with only _updateStatusScopeName (no _waitForCompletionScopeName as what ConvertOperation does)

pshao25:
Fix https://github.com/Azure/autorest.csharp/issues/2236

### DIFF
--- a/sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
+++ b/sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
@@ -121,7 +121,7 @@ namespace Azure.Core
         /// <returns>The last HTTP response received from the server, including the final result of the long-running operation.</returns>
         /// <exception cref="RequestFailedException">Thrown if there's been any issues during the connection, or if the operation has completed with failures.</exception>
         public async ValueTask<Response> WaitForCompletionResponseAsync(CancellationToken cancellationToken)
-            => await WaitForCompletionResponseAsync(async: true, null, cancellationToken).ConfigureAwait(false);
+            => await WaitForCompletionResponseAsync(async: true, null, _waitForCompletionResponseScopeName, cancellationToken).ConfigureAwait(false);
 
         /// <summary>
         /// Periodically calls <see cref="UpdateStatusAsync(CancellationToken)"/> until the long-running operation completes. The interval
@@ -142,7 +142,7 @@ namespace Azure.Core
         /// <returns>The last HTTP response received from the server, including the final result of the long-running operation.</returns>
         /// <exception cref="RequestFailedException">Thrown if there's been any issues during the connection, or if the operation has completed with failures.</exception>
         public async ValueTask<Response> WaitForCompletionResponseAsync(TimeSpan pollingInterval, CancellationToken cancellationToken)
-            => await WaitForCompletionResponseAsync(async: true, pollingInterval, cancellationToken).ConfigureAwait(false);
+            => await WaitForCompletionResponseAsync(async: true, pollingInterval, _waitForCompletionResponseScopeName, cancellationToken).ConfigureAwait(false);
 
         /// <summary>
         /// Periodically calls <see cref="UpdateStatus(CancellationToken)"/> until the long-running operation completes.
@@ -162,7 +162,7 @@ namespace Azure.Core
         /// <returns>The last HTTP response received from the server, including the final result of the long-running operation.</returns>
         /// <exception cref="RequestFailedException">Thrown if there's been any issues during the connection, or if the operation has completed with failures.</exception>
         public Response WaitForCompletionResponse(CancellationToken cancellationToken)
-            => WaitForCompletionResponseAsync(async: false, null, cancellationToken).EnsureCompleted();
+            => WaitForCompletionResponseAsync(async: false, null, _waitForCompletionResponseScopeName, cancellationToken).EnsureCompleted();
 
         /// <summary>
         /// Periodically calls <see cref="UpdateStatus(CancellationToken)"/> until the long-running operation completes. The interval
@@ -183,9 +183,9 @@ namespace Azure.Core
         /// <returns>The last HTTP response received from the server, including the final result of the long-running operation.</returns>
         /// <exception cref="RequestFailedException">Thrown if there's been any issues during the connection, or if the operation has completed with failures.</exception>
         public Response WaitForCompletionResponse(TimeSpan pollingInterval, CancellationToken cancellationToken)
-            => WaitForCompletionResponseAsync(async: false, pollingInterval, cancellationToken).EnsureCompleted();
+            => WaitForCompletionResponseAsync(async: false, pollingInterval, _waitForCompletionResponseScopeName, cancellationToken).EnsureCompleted();
 
-        protected async ValueTask<Response> WaitForCompletionResponseAsync(bool async, TimeSpan? pollingInterval, CancellationToken cancellationToken)
+        protected async ValueTask<Response> WaitForCompletionResponseAsync(bool async, TimeSpan? pollingInterval, string scopeName, CancellationToken cancellationToken)
         {
             // If _responseLock has the value, lockOrValue will contain that value, and no lock is acquired.
             // If _responseLock doesn't have the value, GetLockOrValueAsync will acquire the lock that will be released when lockOrValue is disposed
@@ -195,7 +195,7 @@ namespace Azure.Core
                 return lockOrValue.Value;
             }
 
-            using var scope = CreateScope(_waitForCompletionResponseScopeName);
+            using var scope = CreateScope(scopeName);
             try
             {
                 var poller = new OperationPoller(_fallbackStrategy);
--- a/sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
+++ b/sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
@@ -19,12 +19,13 @@ namespace Azure.Core
         private readonly DelayStrategy? _fallbackStrategy;
         private readonly AsyncLockWithValue<Response> _responseLock;
 
-        protected readonly string _operationTypeName;
+        private readonly string? _waitForCompletionResponseScopeName;
+        protected readonly string? _updateStatusScopeName;
+        protected readonly string? _waitForCompletionScopeName;
 
         protected OperationInternalBase(Response rawResponse)
         {
             _diagnostics = new ClientDiagnostics(ClientOptions.Default);
-            _operationTypeName = string.Empty;
             _scopeAttributes = default;
             _fallbackStrategy = default;
             _responseLock = new AsyncLockWithValue<Response>(rawResponse);
@@ -33,7 +34,9 @@ namespace Azure.Core
         protected OperationInternalBase(ClientDiagnostics clientDiagnostics, string operationTypeName, IEnumerable<KeyValuePair<string, string>>? scopeAttributes = null, DelayStrategy? fallbackStrategy = null)
         {
             _diagnostics = clientDiagnostics;
-            _operationTypeName = operationTypeName;
+            _updateStatusScopeName = $"{operationTypeName}.UpdateStatus";
+            _waitForCompletionResponseScopeName = $"{operationTypeName}.WaitForCompletionResponse";
+            _waitForCompletionScopeName = $"{operationTypeName}.WaitForCompletion";
             _scopeAttributes = scopeAttributes?.ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
             _fallbackStrategy = fallbackStrategy;
             _responseLock = new AsyncLockWithValue<Response>();
@@ -179,7 +182,7 @@ namespace Azure.Core
         public Response WaitForCompletionResponse(TimeSpan pollingInterval, CancellationToken cancellationToken)
             => WaitForCompletionResponseAsync(async: false, pollingInterval, cancellationToken).EnsureCompleted();
 
-        private async ValueTask<Response> WaitForCompletionResponseAsync(bool async, TimeSpan? pollingInterval, CancellationToken cancellationToken)
+        protected async ValueTask<Response> WaitForCompletionResponseAsync(bool async, TimeSpan? pollingInterval, CancellationToken cancellationToken)
         {
             // If _responseLock has the value, lockOrValue will contain that value, and no lock is acquired.
             // If _responseLock doesn't have the value, GetLockOrValueAsync will acquire the lock that will be released when lockOrValue is disposed
@@ -189,7 +192,7 @@ namespace Azure.Core
                 return lockOrValue.Value;
             }
 
-            using var scope = CreateScope(string.IsNullOrEmpty(_operationTypeName) ? string.Empty : $"{_operationTypeName}.WaitForCompletionResponse");
+            using var scope = CreateScope(_waitForCompletionResponseScopeName!);
             try
             {
                 var poller = new OperationPoller(_fallbackStrategy);
@@ -228,5 +231,10 @@ namespace Azure.Core
         protected async ValueTask<RequestFailedException> CreateException(bool async, Response response) => async
             ? await _diagnostics.CreateRequestFailedExceptionAsync(response).ConfigureAwait(false)
             : _diagnostics.CreateRequestFailedException(response);
+
+        protected bool TryGetResponseValue(out Response? value)
+        {
+            return _responseLock.TryGetValue(out value);
+        }
     }
 }
--- a/sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
+++ b/sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
@@ -16,6 +16,7 @@ namespace Azure.Core
     {
         private readonly ClientDiagnostics _diagnostics;
         private readonly string _updateStatusScopeName;
+        private readonly string _waitForCompletionScopeName;
         private readonly IReadOnlyDictionary<string, string>? _scopeAttributes;
         private readonly DelayStrategy? _fallbackStrategy;
         private readonly AsyncLockWithValue<Response> _responseLock;
@@ -24,6 +25,7 @@ namespace Azure.Core
         {
             _diagnostics = new ClientDiagnostics(ClientOptions.Default);
             _updateStatusScopeName = string.Empty;
+            _waitForCompletionScopeName = string.Empty;
             _scopeAttributes = default;
             _fallbackStrategy = default;
             _responseLock = new AsyncLockWithValue<Response>(rawResponse);
@@ -33,6 +35,7 @@ namespace Azure.Core
         {
             _diagnostics = clientDiagnostics;
             _updateStatusScopeName = $"{operationTypeName}.UpdateStatus";
+            _waitForCompletionScopeName = $"{operationTypeName}.WaitForCompletion";
             _scopeAttributes = scopeAttributes?.ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
             _fallbackStrategy = fallbackStrategy;
             _responseLock = new AsyncLockWithValue<Response>();
@@ -188,20 +191,29 @@ namespace Azure.Core
                 return lockOrValue.Value;
             }
 
-            var poller = new OperationPoller(_fallbackStrategy);
-            var response = async
-                ? await poller.WaitForCompletionResponseAsync(this, pollingInterval, cancellationToken).ConfigureAwait(false)
-                : poller.WaitForCompletionResponse(this, pollingInterval, cancellationToken);
+            using var scope = CreateScope(false);
+            try
+            {
+                var poller = new OperationPoller(_fallbackStrategy);
+                var response = async
+                    ? await poller.WaitForCompletionResponseAsync(this, pollingInterval, cancellationToken).ConfigureAwait(false)
+                    : poller.WaitForCompletionResponse(this, pollingInterval, cancellationToken);
 
-            lockOrValue.SetValue(response);
-            return response;
+                lockOrValue.SetValue(response);
+                return response;
+            }
+            catch (Exception e)
+            {
+                scope.Failed(e);
+                throw;
+            }
         }
 
         protected abstract ValueTask<Response> UpdateStatusAsync(bool async, CancellationToken cancellationToken);
 
-        protected DiagnosticScope CreateScope()
+        protected DiagnosticScope CreateScope(bool isUpdateStatus)
         {
-            DiagnosticScope scope = _diagnostics.CreateScope(_updateStatusScopeName);
+            DiagnosticScope scope = _diagnostics.CreateScope(isUpdateStatus ? _updateStatusScopeName : _waitForCompletionScopeName);
 
             if (_scopeAttributes != null)
             {
--- a/sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
+++ b/sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
@@ -234,10 +234,5 @@ namespace Azure.Core
         protected async ValueTask<RequestFailedException> CreateException(bool async, Response response) => async
             ? await _diagnostics.CreateRequestFailedExceptionAsync(response).ConfigureAwait(false)
             : _diagnostics.CreateRequestFailedException(response);
-
-        protected bool TryGetResponseValue(out Response? value)
-        {
-            return _responseLock.TryGetValue(out value);
-        }
     }
 }
--- a/sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
+++ b/sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
@@ -34,8 +34,8 @@ namespace Azure.Core
         protected OperationInternalBase(ClientDiagnostics clientDiagnostics, string operationTypeName, IEnumerable<KeyValuePair<string, string>>? scopeAttributes = null, DelayStrategy? fallbackStrategy = null)
         {
             _diagnostics = clientDiagnostics;
-            _updateStatusScopeName = $"{operationTypeName}.UpdateStatus";
-            _waitForCompletionResponseScopeName = $"{operationTypeName}.WaitForCompletionResponse";
+            _updateStatusScopeName = $"{operationTypeName}.{nameof(UpdateStatus)}";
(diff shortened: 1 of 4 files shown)


sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
```diff
@@ -188,20 +191,29 @@ private async ValueTask<Response> WaitForCompletionResponseAsync(bool async, Tim
...
+            {
+                scope.Failed(e);
+                throw;
+            }
         }
 
         protected abstract ValueTask<Response> UpdateStatusAsync(bool async, CancellationToken cancellationToken);
 
-        protected DiagnosticScope CreateScope()
+        protected DiagnosticScope CreateScope(bool isUpdateStatus)
```
lirenhe, line 214: Can we have the default value for isUpdateStatus to make the default bahavior unchanged?

lirenhe, line 214: Can we have the default value for isUpdateStatus to make the default bahavior unchanged?

chunyu3: Use an flag will make it is not readable and a little confuse. Since the purpose of the flag is to select the right scope name. Why not use the scope name as the parameter.
protected DiagnosticScope CreateScope(string scopeName)

chunyu3:
Hi @pshao25 So now Will LRO set two scopes, one is 'WaitForcomplete', the other is 'UpdateStatus'?

pshao25:
> Hi @pshao25 So now Will LRO set two scopes, one is 'WaitForcomplete', the other is 'UpdateStatus'?

Yes

pshao25:
@AlexanderSher @chunyu3 Considering your suggesting about changing flag to string. I would add a check whether the input name is `_updateStatusScopeName` or `_waitForCompletionScopeName` because I really don't want to allow passing in arbitrary name. In that case, we might not put the `_updateStatusScopeName` into `OperationInternal<T>` and I have to make it from private to public.

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
```diff
@@ -188,20 +192,34 @@ private async ValueTask<Response> WaitForCompletionResponseAsync(bool async, Tim
...
         protected abstract ValueTask<Response> UpdateStatusAsync(bool async, CancellationToken cancellationToken);
 
-        protected DiagnosticScope CreateScope()
+        protected DiagnosticScope CreateScope(string scopeName)
         {
-            DiagnosticScope scope = _diagnostics.CreateScope(_updateStatusScopeName);
+            if (scopeName != _updateStatusScopeName && scopeName != _waitForCompletionScopeName)
+            {
+                throw new NotSupportedException($"The scope name is not supported. Allowed value: {_updateStatusScopeName}, {_waitForCompletionScopeName}");
+            }
```
chunyu3, line 220: Why do we need add this check? CreateScope is a function to create a scope, it is not a dedicate function to only create UpdateStatusScope or WaitForCompletionScope.  When you call this function you need to pass the right scope name.

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
```diff
@@ -15,15 +15,18 @@ namespace Azure.Core
...
         private readonly AsyncLockWithValue<Response> _responseLock;
 
+        protected readonly string _updateStatusScopeName;
+        private readonly string _waitForCompletionScopeName;
+
         protected OperationInternalBase(Response rawResponse)
         {
             _diagnostics = new ClientDiagnostics(ClientOptions.Default);
             _updateStatusScopeName = string.Empty;
+            _waitForCompletionScopeName = string.Empty;
```
annelo-msft, line 29: Curious: when do we want to open distributed tracing span with an empty name?

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
```diff
@@ -15,15 +15,18 @@ namespace Azure.Core
     internal abstract class OperationInternalBase
     {
         private readonly ClientDiagnostics _diagnostics;
-        private readonly string _updateStatusScopeName;
         private readonly IReadOnlyDictionary<string, string>? _scopeAttributes;
         private readonly DelayStrategy? _fallbackStrategy;
         private readonly AsyncLockWithValue<Response> _responseLock;
 
+        protected readonly string _updateStatusScopeName;
```
annelo-msft, line 22: Why is this protected and not private?

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
```diff
@@ -33,6 +36,7 @@ protected OperationInternalBase(ClientDiagnostics clientDiagnostics, string oper
         {
             _diagnostics = clientDiagnostics;
             _updateStatusScopeName = $"{operationTypeName}.UpdateStatus";
+            _waitForCompletionScopeName = $"{operationTypeName}.WaitForCompletion";
```
annelo-msft, line 39: Could we instead store a single private field `_operationTypeName` and append `.UpdateStatus`/`.WaitForCompletion` in the appropriate public methods?

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
```diff
@@ -188,20 +192,34 @@ private async ValueTask<Response> WaitForCompletionResponseAsync(bool async, Tim
                 return lockOrValue.Value;
             }
 
-            var poller = new OperationPoller(_fallbackStrategy);
-            var response = async
-                ? await poller.WaitForCompletionResponseAsync(this, pollingInterval, cancellationToken).ConfigureAwait(false)
-                : poller.WaitForCompletionResponse(this, pollingInterval, cancellationToken);
+            using var scope = CreateScope(_waitForCompletionScopeName);
```
annelo-msft, line 195: It looks like this will open a scope called `<OperationTypeName>.WaitForCompletion` but the name of the public method is `WaitForCompletionResponse`.  Please update this so the scope name matches the name of the public method.

annelo-msft (commented):
What tests validate that this functionality is correct?

pshao25:
> What tests validate that this functionality is correct?

We will comment these lines to test this scenario https://github.com/Azure/azure-sdk-for-net/pull/29033#pullrequestreview-995607491
Meanwhile, we will remove the scopes added for grow-up story.

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
pshao25: Updated. But just want to confirm: we don't expect RP who might customize the `operation` will pass in an invalid scope name, right?

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
pshao25: I changed the field to `_operationTypeName` but still protected because its sub class `OperationInternal<T>` needs it.

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
pshao25: Management plane is using this. It looks like for mocking purpose. The usage is like 
https://github.com/Azure/autorest.csharp/blob/0790793118f59b07bfd58047ab0dbbf1501373fe/test/TestProjects/ResourceRename/Generated/LongRunningOperation/ResourceRenameArmOperation.cs#L31
User calls this to create an operation with in a final successful/failed state.

annelo-msft:
> We will comment these lines to test this scenario https://github.com/Azure/autorest.csharp/blob/feature%2Fv3/test/AutoRest.TestServerLowLevel.Tests/dpg-customization.cs#L106-L108

These are changes to Azure.Core files in `azure-sdk-for-net`.  We should have test coverage showing their correctness in `azure-sdk-for-net`.

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
annelo-msft: Oh, I see it was addressed!  Thanks!

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
```diff
@@ -15,15 +15,16 @@ namespace Azure.Core
...
         private readonly DelayStrategy? _fallbackStrategy;
         private readonly AsyncLockWithValue<Response> _responseLock;
 
+        protected readonly string _operationTypeName;
+
         protected OperationInternalBase(Response rawResponse)
         {
             _diagnostics = new ClientDiagnostics(ClientOptions.Default);
-            _updateStatusScopeName = string.Empty;
+            _operationTypeName = string.Empty;
```
annelo-msft, line 27: We would now open a scope called ".UpdateStatus" or ".WaitForCompletionResponse" if someone used this constructor.  We shouldn't rely on the fact that there is only one caller of this constructor (right now) to ensure that we don't get this behavior -- it would be easy for someone to add a usage of this constructor that _does_ open a scope.

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
annelo-msft: I don't think we validate this for scopes elsewhere -- would you be able to confirm this?

@JoshLove-msft/@christothes/@AlexanderSher, is it correct that we validate diagnostic scope names only through tests?  Could you point @pshao25 to documentation/examples of how we do this?

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
annelo-msft: See comment here: https://github.com/Azure/azure-sdk-for-net/pull/29033#discussion_r890410465

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
annelo-msft: I think that is ok for now, we can revisit this when we refactor these types as part of MQ.  @AlexanderSher FYI

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
pshao25: This is never going to happen. If anyone calls this constructor, he has to pass in a rawResponse, which will initiliza a `_responseLock` by storing its value. With this value, we will stop before we create the scopes. We will return [here](https://github.com/Azure/azure-sdk-for-net/blob/1660eec291fea10fb590eee75eacbdeeb6f20675/sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs#L188).

pshao25:
> > We will comment these lines to test this scenario https://github.com/Azure/autorest.csharp/blob/feature%2Fv3/test/AutoRest.TestServerLowLevel.Tests/dpg-customization.cs#L106-L108
> 
> These are changes to Azure.Core files in `azure-sdk-for-net`. We should have test coverage showing their correctness in `azure-sdk-for-net`.

Added.

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
AlexanderSher: It is not for mocking purpose. This constructor is used when operation is in the final state, hence no scope will be created.

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
AlexanderSher: This constructor is for completed operation only. As @pshao25 has correctly pointed out, no scope should be opened for completed operation. If that happens, then whole type is broken

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
```diff
@@ -188,20 +189,29 @@ private async ValueTask<Response> WaitForCompletionResponseAsync(bool async, Tim
                 return lockOrValue.Value;
             }
 
-            var poller = new OperationPoller(_fallbackStrategy);
-            var response = async
-                ? await poller.WaitForCompletionResponseAsync(this, pollingInterval, cancellationToken).ConfigureAwait(false)
-                : poller.WaitForCompletionResponse(this, pollingInterval, cancellationToken);
+            using var scope = CreateScope(string.IsNullOrEmpty(_operationTypeName) ? string.Empty : $"{_operationTypeName}.WaitForCompletionResponse");
```
AlexanderSher, line 192: `_operationTypeName` can be empty only if operation is completed, so result of this check is known upfront. And if it is not empty, we create garbage strings every time scope is created. Considering the time that is required to complete the operation, those strings can end up in Gen2.

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
annelo-msft: I see, thanks for helping me understand!  Per offline conversation, I understand now that we can guarantee that this constructor is only ever called with a Response from a completed operation because:

1. It can only be called by subclasses, because it is protected.
2. Only OperationInternal and OperationInternal\<T\> subclass OperationInternalBase, and so are the only callers of this constructor.
3. The only callers of the protected constructor are on static Succeeded() and Failed() factory methods, that are only ever called after the operation has completed.

Thank you!

Would it make sense to set _operationTypeName to null, then, to indicate that we expect it will never be used in a concatenated string?

annelo-msft: I see, thanks for helping me understand!  Per offline conversation, I understand now that we can guarantee that this constructor is only ever called with a Response from a completed operation because:

1. It can only be called by subclasses, because it is protected.
2. Only OperationInternal and OperationInternal\<T\> subclass OperationInternalBase, and so are the only callers of this constructor.
3. The only callers of the protected constructor are on static Succeeded() and Failed() factory methods, that are only ever called after the operation has completed.

Thank you!

Would it make sense to set _operationTypeName to null, then, to indicate that we expect it will never be used in a concatenated string?

sdk/core/Azure.Core/tests/OperationInternalOfTTests.cs
```diff
@@ -205,6 +218,35 @@ public async Task UpdateStatusPassesTheCancellationTokenToUpdateState([Values(tr
...
+
+            string expectedTypeName = operationTypeName ?? nameof(TestOperation);
+            KeyValuePair<string, string>[] expectedAttributes = { new("key1", "value1"), new("key2", "value2") };
+            var operationInternal = new OperationInternal<int>(ClientDiagnostics, TestOperation.Succeeded(1), InitialResponse, operationTypeName, expectedAttributes);
+
+            _ = async
+                ? await operationInternal.WaitForCompletionResponseAsync(CancellationToken.None)
+                : operationInternal.WaitForCompletionResponse(CancellationToken.None);
+
+            testListener.AssertScope($"{expectedTypeName}.WaitForCompletionResponse", expectedAttributes);
```
annelo-msft, line 234: We should add a test for `WaitForCompletion` for Operation\<T\> as well.  (Operation does not have this method so doesn't need this test.)

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
JoshLove-msft: Yes, when you instrument your client as part of recorded tests, we add an interceptor that will validate that a scope is created that matches the name of the method. There is a gap in that if you are using the ForwardClientCalls attribute and creating the scope from a different internal method, the name is not validated.

annelo-msft (commented):
We should have end-to-end tests for distributed tracing in long-running operations, validating each of the cases described here: https://github.com/Azure/autorest.csharp/issues/2136#issuecomment-1092257699

**Update:** @AlexanderSher let me know the end-to-end tests are here: https://github.com/Azure/autorest.csharp/blob/feature/v3/test/AutoRest.TestServerLowLevel.Tests/dpg-customization.cs#L96.

We should move these into Azure.Core when we migrate these types over.

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
pshao25: Makes sense. @annelo-msft  How about we go back to the original implementation, i.e. having two fields `_updateStatusScopeName` and `_waitForCompletionResponseScopeName`? Then it will only have two strings for an operation instance.

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
pshao25: Agree. Will update after we have a conclusion for the perf issue of https://github.com/Azure/azure-sdk-for-net/pull/29033#discussion_r891517743

sdk/core/Azure.Core/tests/OperationInternalOfTTests.cs
pshao25: `WaitForCompletionResponse` is a public method could be called by operation. Why we don't need this?

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
annelo-msft: @pshao25, I believe Alex is saying that we don't need to do the check because we already know the result.  By not doing the check, we would avoid creating strings we don't need that could also end up in Gen2.  @AlexanderSher, is this correct?  Do you have a recommendation for what Pan should to here instead?

sdk/core/Azure.Core/tests/OperationInternalOfTTests.cs
annelo-msft: > WaitForCompletionResponse is a public method could be called by operation. Why we don't need this?

I am not saying we don't need the test for WaitForCompletionResponse.  I am saying that we need to **add** a test for WaitForCompletion to the OperationInternal\<T\> tests.  OperationInternal\<T\> does have a WaitForCompletion method.  I would like for us to add a test that shows that when `WaitForCompletion` is called, the scope name is `WaitForCompletion` and not `WaitForCompletionResponse`.

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
AlexanderSher: > Would it make sense to set _operationTypeName to null

It would require to add null checks in various places where field is used. So I'd prefer it to be `string.Empty`. Also, I'd like to bring back `_updateStatusScopeName` to avoid creation of garbage strings.

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
AlexanderSher: @pshao25 got it right. We can set values to `_updateStatusScopeName` and `_waitForCompletionResponseScopeName` in constructor or when they are accessed. There is a case when operation is created and not awaited, but that case is rare. The most common case is when both fields are used, so I think it is better to assign these fields in constructor. In this case, `_operationTypeName` won't be required.

annelo-msft: Thank you for the clarification, @AlexanderSher.  I am fine with your recommended implementation and we can come back and change it if needed when we refactor the LRO types in MQ.

sdk/core/Azure.Core/tests/OperationInternalOfTTests.cs
```diff
@@ -205,6 +218,64 @@ public async Task UpdateStatusPassesTheCancellationTokenToUpdateState([Values(tr
...
```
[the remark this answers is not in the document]
pshao25: You are corret. But I found some problem with the suppression flag. I'll update this after I figure out the root cause.

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
pshao25: Do we have further plan to refactor? Could you let me know the context?

annelo-msft: - https://github.com/Azure/autorest.csharp/issues/1706
- https://github.com/Azure/azure-sdk-for-net/issues/25991
- https://github.com/Azure/autorest.csharp/issues/2231

sdk/core/Azure.Core/tests/OperationInternalTests.cs
```diff
@@ -18,7 +18,7 @@ namespace Azure.Core.Tests
     public class OperationInternalTests
     {
         private static readonly string DiagnosticNamespace = "Azure.Core.Tests";
-        private static readonly ClientDiagnostics ClientDiagnostics = new(new TestClientOptions());
+        private static readonly ClientDiagnostics ClientDiagnostics = new(new TestClientOptions(), true);
```
annelo-msft, line 21: I think many of these tests are for HLC OperationInternals, so we shouldn't just update the ClientDiagnostics to only represent DPG.

pshao25: Does that mean we separate these tests out of this test class and put it into another one only for DPG?

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
AlexanderSher: Should we use `{operationTypeName}.{(nameof(WaitForCompletionResponse)}` and `{operationTypeName}.{(nameof(WaitForCompletion)}` here?

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
```diff
@@ -15,15 +15,17 @@ namespace Azure.Core
     internal abstract class OperationInternalBase
     {
         private readonly ClientDiagnostics _diagnostics;
-        private readonly string _updateStatusScopeName;
         private readonly IReadOnlyDictionary<string, string>? _scopeAttributes;
         private readonly DelayStrategy? _fallbackStrategy;
         private readonly AsyncLockWithValue<Response> _responseLock;
 
+        private readonly string? _waitForCompletionResponseScopeName;
```
AlexanderSher, line 22: I don't like that these properties have become nullable. Later, we have to use `!` operator, meaning that we break our own nullability rules.

sdk/core/Azure.Core/src/Shared/OperationInternalOfT.cs
```diff
@@ -244,9 +235,28 @@ public Response<T> WaitForCompletion(CancellationToken cancellationToken)
         /// <returns>The last HTTP response received from the server, including the final result of the long-running operation.</returns>
         /// <exception cref="RequestFailedException">Thrown if there's been any issues during the connection, or if the operation has completed with failures.</exception>
         public Response<T> WaitForCompletion(TimeSpan pollingInterval, CancellationToken cancellationToken)
+            => WaitForCompletionAsync(async: false, pollingInterval, cancellationToken).EnsureCompleted();
+
+        private async ValueTask<Response<T>> WaitForCompletionAsync(bool async, TimeSpan? pollingInterval, CancellationToken cancellationToken)
```
AlexanderSher, line 240: It looks like that if you add `scopeName` parameter to `OperationInternalBase.WaitForCompletionResponseAsync`, this method can be reduced to:
```cs
private async ValueTask<Response<T>> WaitForCompletionAsync(bool async, TimeSpan? pollingInterval, CancellationToken cancellationToken) {
    var rawResponse = await WaitForCompletionResponseAsync(async, pollingInterval, _waitForCompletionScopeName, cancellationToken);
    return Response.FromValue(Value, rawResponse);
}
```

annelo-msft (approved):
All of the tests I've requested are present and pass.  Please address @AlexanderSher's comments before merging.

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
pshao25: This is because of this https://github.com/Azure/azure-sdk-for-net/pull/29033#discussion_r891693000

> Would it make sense to set _operationTypeName to null, then, to indicate that we expect it will never be used in a concatenated string?

@annelo-msft so we revert it back to make the defualt as `string.Empty`?

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
pshao25: We can only add `{operationTypeName}.{(nameof(WaitForCompletionResponse)}`.
`WaitForCompletion` does not exist in `OperationInternalBase`.

sdk/core/Azure.Core/src/Shared/OperationInternalOfT.cs
pshao25: `WaitForCompletionResponseAsync` can only have scope name `_waitForCompletionResponseScopeName`. I don't think it is a good practice to expose this parameter.

Pablo1q:
-  good[]()[]()

sdk/core/Azure.Core/src/Shared/OperationInternalBase.cs
annelo-msft: Yes.  The reason to have made this nullable instead of an empty string was to ensure that we didn't accidentally print an empty string in the diagnostic scope.  Since you have both convinced me that [this isn't possible](https://github.com/Azure/azure-sdk-for-net/pull/29033#discussion_r891693000), I agree that it is preferred not to make them nullable.

annelo-msft (approved):
You can merge the PR with these changes once the CI is passing.  Thanks, @pshao25!

PR #29041, opened 2022-05-31
feat(dpg): improve client summary

archerzz:
part of https://github.com/Azure/autorest.csharp/issues/2195


annelo-msft (commented):
@mikekistler, @tg-msft, @KrzysztofCwalina.  We are adding this string to the DPG client comment to be able to find these clients more easily by grepping through the repo.  Do you have any concerns about using this string?  

A few things I noticed that might be of concern are:
- We no longer say `service client`, but "service" appears in the identifier string sometimes
- We identify that it is a generated client
- We use the same string for service clients and subclients.

This can always be changed post-GA, but since we are now looking more closely at the docs story, this becomes more important to get right for GA.

tg-msft:
If this is only for grepping through the repo, can you just make it a regular `//` comment so it doesn't pollute generated documentation?

archerzz:
continue the work in https://github.com/Azure/azure-sdk-for-net/pull/30505

PR #29068, opened 2022-06-01
[Event Hubs] Recover From Ownership Blob Deletion

jsquire:
# Summary

The focus of these changes is to adjust the approach used for claiming ownership such that the processor can recover gracefully when a blob was deleted while the processor is running.  Previously, this scenario would cause the processor to remain in a bad state where the partition was considered owned locally, but Storage did not reflect that, causing load balancing issues within the cluster.

### DIFF
--- a/sdk/eventhub/Azure.Messaging.EventHubs.Shared/src/BlobCheckpointStore/BlobCheckpointStoreInternal.cs
+++ b/sdk/eventhub/Azure.Messaging.EventHubs.Shared/src/BlobCheckpointStore/BlobCheckpointStoreInternal.cs
@@ -204,6 +204,30 @@ namespace Azure.Messaging.EventHubs.Primitives
                     // Even though documentation states otherwise, we cannot use UploadAsync when the blob already exists in
                     // the current storage SDK.  For this reason, we are using the specified ETag as an indication of what
                     // method to use.
+                    //
+                    // If an ETag is associated with ownership, assume the blob exists and attempt to update it.
+
+                    if (ownership.Version != null)
+                    {
+                        try
+                        {
+                            blobRequestConditions.IfMatch = new ETag(ownership.Version);
+                            infoResponse = await blobClient.SetMetadataAsync(metadata, blobRequestConditions, cancellationToken).ConfigureAwait(false);
+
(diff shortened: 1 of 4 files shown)


sdk/eventhub/Azure.Messaging.EventHubs.Shared/src/BlobCheckpointStore/BlobCheckpointStoreInternal.cs
```diff
@@ -218,7 +242,7 @@ await foreach (BlobItem blob in ContainerClient.GetBlobsAsync(traits: BlobTraits
                         catch (RequestFailedException ex) when (ex.ErrorCode == BlobErrorCode.BlobAlreadyExists)
                         {
                             // A blob could have just been created by another Event Processor that claimed ownership of this
-                            // partition.  In this case, there's no point in retrying because we don't have the correct ETag.
+                            // partition.  In this case, there's no point in retrying because we don't have the correct eTag.
```
jsquire, line 245:
```suggestion
                            // partition.  In this case, there's no point in retrying because we don't have the correct ETag.
```

sdk/eventhub/Azure.Messaging.EventHubs.Shared/src/BlobCheckpointStore/BlobCheckpointStoreInternal.cs
```diff
@@ -204,6 +204,30 @@ await foreach (BlobItem blob in ContainerClient.GetBlobsAsync(traits: BlobTraits
...
+                    {
+                        try
+                        {
+                            blobRequestConditions.IfMatch = new ETag(ownership.Version);
+                            infoResponse = await blobClient.SetMetadataAsync(metadata, blobRequestConditions, cancellationToken).ConfigureAwait(false);
+
+                            ownership.LastModifiedTime = infoResponse.Value.LastModified;
+                            ownership.Version = infoResponse.Value.ETag.ToString();
+                        }
+                        catch (RequestFailedException ex) when (ex.ErrorCode == BlobErrorCode.BlobNotFound)
```
christothes, line 220: Do we handle the case where multiple partition workers encounter this and race to upload the blob?

christothes (approved):

PR #29100, opened 2022-06-03
[FormRecognizer] Rebase feature branch into main

kinelski:
This feature branch contains the new features for 2022-06-30-preview. All changes have already been reviewed.


kinelski:
Current failure is a flaky service error we've been seeing in the pipeline for some time now. Not caused by these changes.

kristapratico (approved):

sdk/formrecognizer/Azure.AI.FormRecognizer/src/AddressValue.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Azure.AI.FormRecognizer.DocumentAnalysis
```
pallavit, line 4: Why create this class at all? The generated model should suffice no?

pallavit:
>     public partial class BoundingRegion

Curious - why do we special case the BoundingPolygon?

---
Refers to: sdk/formrecognizer/Azure.AI.FormRecognizer/src/BoundingRegion.cs:11 in 594f6a6. [](commit_id = 594f6a6c5951c31c80e0559fd179c70da4c6e8e9, deletion_comment = False)

sdk/formrecognizer/Azure.AI.FormRecognizer/src/AddressValue.cs
kinelski: CodeGen generates `public` models, but we added a transform to default everything to `internal` ([autorest.md file](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/formrecognizer/Azure.AI.FormRecognizer/src/autorest.md#make-generated-models-internal-by-default)):
```
directive:
  from: https://github.com/Azure/azure-rest-api-specs/blob/7043b48f4be1fdd40757b9ef372b65f054daf48f/specification/cognitiveservices/data-plane/FormRecognizer/stable/v2.1/FormRecognizer.json
  where: $.definitions.*
  transform: >
    $["x-accessibility"] = "internal"
```

We started doing this because we had way more generated models we wanted to keep hidden than models we wanted to expose. I can't tell for sure if that's still true today (would need to double check).

kinelski: We are adding this file to tell CodeGen this model should be made `public` instead of defaulting to `internal`.

kinelski:
> BoundingRegion.cs

kinelski:
> > ```
> > public partial class BoundingRegion
> > ```
> 
> Curious - why do we create a special type BoundingPolygon and not use BoundingRegion directly from the rest API?
> 
> Refers to: sdk/formrecognizer/Azure.AI.FormRecognizer/src/BoundingRegion.cs:11 in [594f6a6](https://github.com/Azure/azure-sdk-for-net/commit/594f6a6c5951c31c80e0559fd179c70da4c6e8e9). [](commit_id = [594f6a6](https://github.com/Azure/azure-sdk-for-net/commit/594f6a6c5951c31c80e0559fd179c70da4c6e8e9), deletion_comment = False)

`BoundingRegion` and `BoundingPolygon` are different concepts, and we expose both in the SDK. `BoundingRegion` is defined by a `BoundingPolygon` + a page number. This pair uniquely determines a region in a multipage input document.

In addition to this, `BoundingPolygon` is an SDK-only model. The definition exists in the service's swagger, but it's treated only as an array of float coordinates, so CodeGen does not generate a corresponding model ([swagger here](https://github.com/Azure/azure-rest-api-specs/blob/a69db59e981eae7dc554823100220c81b34225cc/specification/cognitiveservices/data-plane/FormRecognizer/preview/2022-06-30-preview/FormRecognizer.json#L2060-L2067)):
```
    "BoundingPolygon": {
      "description": "Bounding polygon, with coordinates specified relative to the top-left of the page. The numbers represent the x, y values of the polygon vertices, clockwise from the left (-180 degrees inclusive) relative to the element orientation.",
      "type": "array",
      "items": {
        "type": "number",
        "minimum": 0
      }
    },
```

We encapsulate this array of floats in the `BoundingPolygon` type, which contains an array of `PointF` instead.

kinelski:
> Current failure is a flaky service error we've been seeing in the pipeline for some time now. Not caused by these changes.

Merging this to unblock the next PR. Feel free to add more comments to this PR and I'll address in the next one.

sdk/formrecognizer/Azure.AI.FormRecognizer/src/AddressValue.cs
maririos: I recently learned that something other libraries have done for classes that are just exposed but don't have any logic to it, is to aggregate them in one file.
We did this in TA https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/textanalytics/Azure.AI.TextAnalytics/src/GeneratorStubs/InternalTypes.cs (although it was to hide things, but same concept) and I think KV has it too? or Search?.

Something to consider for future reference :)

sdk/formrecognizer/Azure.AI.FormRecognizer/tests/FormRecognizerClient/RecognizeContentLiveTests.cs
```diff
@@ -29,6 +29,7 @@ public RecognizeContentLiveTests(bool isAsync, FormRecognizerClientOptions.Servi
         }
 
         [RecordedTest]
+        [Ignore("https://github.com/Azure/azure-sdk-for-net-pr/issues/1726")]
```
pallavit, line 32: Why ignoring these tests? I thought test issue was fixed?

kinelski: The issue that got solved was another one (failure when copying a model). This test is being ignored because AAD is not working in the new preview. There are still a couple of issues to be solved on the service side for this new release (we are tracking them in GH), and the service team has already been notified about those.

pallavit: Sounds good :)

PR #29110, opened 2022-06-03
Debug storage CI failures

### LINKED ISSUE #26881: Blob tests not running on Net461
During the Build & Test step, I see this error:
   NUnit failed to load D:\a\_work\1\s\artifacts\bin\Azure.Storage.Blobs.Tests\Release\net461\Azure.Storage.Blobs.Tests.dll
https://dev.azure.com/azure-sdk/public/_build/results?buildId=1353384&view=logs&j=45f79b3a-2624-51cc-bc09-5c8a7cc7b4c9&t=a5b29a05-b133-5d99-2802-d1ab9b116429&l=120

You can see from the dashboard, that 3k fewer tests are running for Net461:
https://dev.azure.com/azure-sdk/public/_build/results?buildId=1353384&view=ms.vss-test-web.build-test-results-tab

I'd expect that the test run would fail if one of the assemblies failed to load, but it is still marked as passing.

JoshLove-msft:
Attempting to fix https://github.com/Azure/azure-sdk-for-net/issues/26881

### DIFF
--- a/sdk/storage/ci.yml
+++ b/sdk/storage/ci.yml
@@ -53,3 +53,6 @@ extends:
       safeName: AzureResourceManagerStorage
     TestSetupSteps:
     - template: /sdk/storage/tests-install-azurite.yml
+    MatrixReplace:
+      # Use dedicated storage pool in canadacentral with higher memory capacity
+      - Pool=(.*)-general/$1-storage


sdk/storage/ci.yml
```diff
@@ -22,6 +22,7 @@ pr:
     - sdk/storage/
     - sdk/resourcemanager/
     - sdk/compute/
+    - sdk/keyvault/
```
weshaggard, line 25: How does adding this extra trigger path fix that associated issue?

JoshLove-msft: I don't think it does.. I think I'm adding it in the wrong place. Trying to figure out the correct parameter to use.

JoshLove-msft: I need to specify additional Paths for the sparse checkout - something like:
```
  template: ../../eng/common/pipelines/templates/steps/sparse-checkout.yml
  parameters:
    Paths:
      include:
      - sdk/storage/
      - sdk/keyvault/
```

weshaggard: For sparse checkout we include everything except the test recordings. See https://github.com/Azure/azure-sdk-for-net/blob/main/eng/pipelines/templates/jobs/ci.tests.yml#L44

Even looking at the logs I see that we are building the keyvault library successfully in the test job.

weshaggard: See https://dev.azure.com/azure-sdk/public/_build/results?buildId=1619591&view=logs&j=45f79b3a-2624-51cc-bc09-5c8a7cc7b4c9&t=a5b29a05-b133-5d99-2802-d1ab9b116429&l=36 

If keyvault directory didn't exist that would fail to restore/build.

JoshLove-msft: Agreed, but when I turn off sparse checkout the failure goes away, so it seems like it is the cause.

JoshLove-msft: > For sparse checkout we include everything except the test recordings. See https://github.com/Azure/azure-sdk-for-net/blob/main/eng/pipelines/templates/jobs/ci.tests.yml#L44

I suspect that the sparse settings that are used for the dynamic test generation are somehow overriding the settings that are used for the test runs.

JoshLove-msft: This was a false alarm. It was unrelated to sparse checkout.

sdk/storage/ci.yml
```diff
@@ -53,3 +53,6 @@ extends:
       safeName: AzureResourceManagerStorage
     TestSetupSteps:
     - template: /sdk/storage/tests-install-azurite.yml
+    MatrixReplace:
+      # Use dedicated storage pool in canadacentral with higher memory capacity
+      - Pool=(.*)-general/$1-storage
```
benbp, line 58: We can certainly try this, but in CI mode the region of the VM shouldn't matter, and the SKUs between these pools are identical.

You can also invoke a run directly from this branch instead of merging, which I think is preferable, so that we don't accidentally leave tests in that use up allocation in the storage pool.

JoshLove-msft: @weshaggard tracked down the issue - it is related to a T1 mgmt project reference that storage tests were using that was not fully signed.

Issue #29116, opened 2022-06-05
[Service Bus] Add property to ServiceBusSessionReceiver to indicate whether lock is there


JoshLove-msft:
If the lock is lost for a ServiceBusSessionReceiver, it will not be reconnected. The only way to tell if the lock has been lost is to attempt an operation and see if it succeeds or fails. It might be useful if there was a property that could be exposed to let users know if the lock is valid, e.g. `HasSessionLock`. This could be helpful when implementing a factory pattern where SessionReceivers are handed out to callers. The callers may still end up with a "dead" session receiver after some time, and they would thus need to ask the factory for a new instance, but the factory would be able to check the property before attempting to re-accept a session, rather than doing a try/catch on some arbitrary operation on the receiver.

We already have the IsClosed property, but the semantics there are that this is true only when the user closes the receiver themselves. If we were to set this to true when the session lock is lost, this could break user exception handling when they are expecting a SessionLockLost error rather than an ObjectDisposedException (which is what is thrown when IsClosed is true). We could maintain the behavior that we only throw ObjectDisposedException if the user manually closed the receiver though. This might be better than introducing a separate property as the expectation is that if IsClosed is true, you shouldn't use the receiver.

See https://stackoverflow.com/questions/72457902/azure-servicebus-managing-of-session-based-receivers/72510777

PR #29117, opened 2022-06-06
Tweak IsClosed behavior for Session receiver

JoshLove-msft:
Addresses https://github.com/Azure/azure-sdk-for-net/issues/29116

IsClosed on ServiceBusSessionReceiver will now return true if the session link dropped (either due to network flakiness or session lock lost), in addition to returning true if user calls CloseAsync or DisposeAsync. 

The behavior when calling an operation on a receiver that had the session link dropped is **not** changed - SessionLockLostException will still be thrown. ObjectDisposedException is only thrown if user calls CloseAsync or DisposeAsync.

Because this is a subtle behavioral change, will need to run it by the architects. The use case for IsClosed is that it can be checked to determine if the receiver should not be used anymore. Thus the change here is an enhancement to make this property more useful.

### DIFF
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpReceiver.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpReceiver.cs
@@ -33,17 +33,29 @@ namespace Azure.Messaging.ServiceBus.Amqp
     internal class AmqpReceiver : TransportReceiver
 #pragma warning restore CA1001 // Types that own disposable fields should be disposable
     {
-        /// <summary>Indicates whether or not this instance has been closed.</summary>
-        private bool _closed;
-
         /// <summary>
-        /// Indicates whether or not this receiver has been closed.
+        /// Indicates whether or not this receiver has been closed by the user.
         /// </summary>
         ///
         /// <value>
         /// <c>true</c> if the receiver is closed; otherwise, <c>false</c>.
         /// </value>
-        public override bool IsClosed => _closed;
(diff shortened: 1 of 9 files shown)


jsquire:
> IsClosed on ServiceBusSessionReceiver will now return true if the session link dropped (either due to network flakiness or session lock lost), in addition to returning true if user calls CloseAsync or DisposeAsync.

We would still want developers to call `CloseAsync` or `DisposeAsync` in this case, right?   Otherwise, the underlying link and other network goo will live until the idle timeout or the associated `ServiceBusClient` is closed?    Assuming there's no way to recover from the session link error, is there a way that we can detect as it happens and just do a full close flow?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpReceiver.cs
```diff
@@ -33,17 +33,27 @@ namespace Azure.Messaging.ServiceBus.Amqp
...
         /// <summary>
-        /// Indicates whether or not this receiver has been closed.
+        /// Indicates whether or not this receiver has been closed by the user.
         /// </summary>
         ///
         /// <value>
         /// <c>true</c> if the receiver is closed; otherwise, <c>false</c>.
         /// </value>
-        public override bool IsClosed => _closed;
+        public override bool IsReceiverClosedByUser => _receiverClosedByUserByUser;
```
jsquire, line 43: Maybe consider something like `WasClosedExplicitly` or similar?   It's not particularly important since it's an internal property... it just reads a bit oddly to me.

jsquire (approved):

PR #29127, opened 2022-06-06
Fix storage blob tests by enabling public signing for dependencies

### LINKED ISSUE #26881: Blob tests not running on Net461

weshaggard:
Fixes https://github.com/Azure/azure-sdk-for-net/issues/26881

Microsoft.Azure.Management.Storage was delay signed and is referenced by Azure.Storage.Blobs.Tests and caused the tests to fail to run on .NET Framework. In order to fix this we need the track 1 mgmt libraries to be public signed so that they can be ran on .NET framework before real signing.

As a follow-up we will want to enable TreatNoTestsAsError as for our run configs to help catch this type of issue in the future. However before we can enable that we need to fix a number of test projects which are currently either failing to load or have no tests in their projects.

### DIFF
--- a/eng/nunit.runsettings
+++ b/eng/nunit.runsettings
@@ -5,8 +5,12 @@
     </NUnit>
     <RunConfiguration>
         <TargetPlatform>x64</TargetPlatform>
+
+        <!-- Return non-zero exit code if a test run finds no tests -->
+        <!-- disable for now because there are a lot of test projects that don't find tests and would start failing -->
+        <!-- <TreatNoTestsAsError>true</TreatNoTestsAsError> -->
     </RunConfiguration>
-    
+
     <TestRunParameters>
         <!-- Controls which service versions are running in live mode  -->
         <!-- <Parameter name="LiveTestServiceVersions" value="V2019_02_02" /> -->
@@ -16,7 +20,7 @@
 
         <!-- Change the test mode -->
         <!-- <Parameter name="TestMode" value="Record" /> -->
-        
+
         <!-- Enable running Fiddler -->
         <!-- <Parameter name="EnableFiddler" value="true" /> -->
     </TestRunParameters>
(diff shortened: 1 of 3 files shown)


eng/nunit.runsettings
```diff
@@ -5,8 +5,12 @@
     </NUnit>
     <RunConfiguration>
         <TargetPlatform>x64</TargetPlatform>
+
+        <!-- Return non-zero exit code if a test run finds no tests -->
+        <!-- disable for now because there are a lot of test projects that don't find tests and would start failing -->
```
weshaggard, line 10: Issue https://github.com/Azure/azure-sdk-for-net/issues/29139 filed to enable this.

weshaggard:
@JoshLove-msft could you help look to make sure the core test failures look expected as I don't think those are related to my changes.

@seanmcc-msft could you have a look at the storage tests? I verified the blob .NET Framework tests are now running but the other failures look similar to existing failures.

JoshLove-msft:
> @JoshLove-msft could you help look to make sure the core test failures look expected as I don't think those are related to my changes.

Looks like at least one of these is a new flaky test. I will kick off a re-run.
Filed https://github.com/Azure/azure-sdk-for-net/issues/29146

JoshLove-msft (approved):

weshaggard:
Storage and core test failures are known and not because of these changing so moving forward with merging this.

ArthurMa1978:
@JoshLove-msft and @weshaggard  we got same mgmt core failure https://github.com/Azure/azure-sdk-for-net/pull/29153 which happened on Microsoft.Azure.Search.Tests.CustomAnalyzerTests.CanUseAllAnalysisComponentOptions, playback test failed. This sdk has not been changed for long time, is it caused by test proxy change?

PR #29129, opened 2022-06-06
[Event Hubs] Checkpoint Migration Sample Tweak

jsquire:
# Summary

The focus of these changes is to tweak the checkpoint migration sample to account for the potential that a blob exists for ownership purposes but does not have valid checkpoint data.  Previously, the sample would emit a T2 checkpoint for these will potentially invalid data.


JoshLove-msft (approved):

m-redding (approved):
Everything makes sense from my perspective!

PR #29135, opened 2022-06-06
[AzureMonitorExporter] implementing Truncation rules for ExceptionDetails and StackFrame

TimothyMothra:
Continuation of #29094

## Changes
- new class `StringExtensions` to hold `Truncation()` extension method.
- new `partial class StackFrame`
  - custom ctor with truncation rules.
- `partial class TelemetryExceptionDetails`
  - remove use of base ctor. Need to apply truncation rules to props in the base ctor, have to implement this in custom ctor.
  - remove `public const int MaxParsedStackLength`, use `SchemaConstants` instead.
  - moved helper methods to `StackFrame` class.
  - refactor method `SanitizeStackFrame`.
    - old: `Tuple<List<TOutput>, bool> SanitizeStackFrame<TInput, TOutput>(IList<TInput> inputList, Func<TInput, int, TOutput> converter, Func<TOutput, int> lengthGetter)` 
    - new: `List<StackFrame> GetSanitizedStackFrame(Exception exception, out bool hasFullStack)`

### DIFF
--- a/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Customizations/Models/TelemetryExceptionDetails.cs
+++ b/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Customizations/Models/TelemetryExceptionDetails.cs
@@ -49,7 +49,7 @@ namespace Azure.Monitor.OpenTelemetry.Exporter.Models
                 int totalLength = 0;
                 for (int counter = 0; counter < frames.Length; counter++)
                 {
-                    // 1) Check that we don't exceeded MaxFrames.
+                    // 1) Check if we have exceeded MaxFrames.
                     if (orderedStackTrace.Count == SchemaConstants.ExceptionDetails_Stack_MaxFrames)
                     {
                         hasFullStack = false;
--- a/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Customizations/Models/TelemetryExceptionDetails.cs
+++ b/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Customizations/Models/TelemetryExceptionDetails.cs
@@ -2,6 +2,9 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
 
 namespace Azure.Monitor.OpenTelemetry.Exporter.Models
 {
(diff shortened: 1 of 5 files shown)


sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/StringExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// This method will truncate a string if that string exceeds a specified max length.
+        /// </summary>
+        /// <remarks>
+        /// This method is a wrapper around <see cref="string.Substring(int, int)"/>.
+        /// </remarks>
+        /// <param name="input">A string to be evaluated.</param>
+        /// <param name="maxLength">A specified length which is used to evaluate the input string.</param>
+        /// <returns>The input string if less than max length, or a substring that begins at 0.</returns>
+        /// <exception cref="ArgumentNullException"></exception>
+        public static string Truncate(this string input, int maxLength)
```
cijothomas, line 20: https://github.com/open-telemetry/opentelemetry-dotnet/issues/2919 Check this, and consider following that approach.

TimothyMothra: I'm not sure what alternatives would be appropriate here. Do you have any ideas?

> As rule of thumb, I avoid adding extension methods for BCL classes unless these are very generic.
> https://github.com/open-telemetry/opentelemetry-dotnet/pull/2914#discussion_r809758538 

I think this `Truncate` method meets the bar of being very generic.
I've got almost 100 constants defined in [SchemaConstants](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/SchemaConstants.cs) and my intent is to use this extension method for all of these.

cijothomas: Okay if it is generic enough, ( I agree), this is fine.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Customizations/Models/TelemetryExceptionDetails.cs
```diff
@@ -4,124 +4,83 @@
...
+            if (frames != null && frames.Any())
             {
-                int currentParsedStackLength = 0;
-                for (int level = 0; level < inputList.Count; level++)
+                int totalLength = 0;
+                for (int counter = 0; counter < frames.Length; counter++)
                 {
-                    // Skip middle part of the stack
-                    int current = (level % 2 == 0) ? (inputList.Count - 1 - (level / 2)) : (level / 2);
+                    // 1) Check that we don't exceeded MaxFrames.
```
vishweshbankwar, line 52:
```suggestion
                    // Check if we have exceeded MaxFrames.
```

vishweshbankwar (approved):

rajkumar-rangaraj (approved):

Issue #29013, opened 2022-05-27
[FormRecognizer] Improve debugging experience for DocumentField


kinelski:
The value of a `DocumentField` is not parsed until the appropriate parsing method is called. For example:
```cs
DocumentField field = result.Fields["totalTax"]; // value of the field is stored in a private member
double fieldValue = field.AsDouble(); // parsing method must be called to expose the value for users
```

This makes the debugger navigation experience difficult. Users can't easily expand the object and see its value without calling the parsing method. What could we do to improve it? Some ideas:
- Consider exposing the value as a generic `object` property called `Value`. On the other hand, this could make navigation more confusing to users who rely on intellisense.
- Consider parsing the fields into specialized `DocumentField<T>` as soon as we are parsing the JSON response. This would change the whole strategy we use for dealing with fields, and should be analyzed carefully.
- Consider using the [DebuggerTypeProxy attribute](https://docs.microsoft.com/en-us/dotnet/framework/debug-trace-profile/enhancing-debugging-with-the-debugger-display-attributes#using-the-debuggertypeproxy) to add a `Value` property. On the other hand, this could make users confused since this property does not exist in the actual field object.

PR #29138, opened 2022-06-06
[FormRecognizer] Added InternalValue property to DocumentField

kinelski:
Please read #29013 for more context.

Before these changes, users needed to navigate to the correct private member (in the example below, `ValueArray`) to access the value of a `DocumentField` in the debugger:
image

This experience is not ideal and requires knowledge about the internal implementation of the Form Recognizer SDK library.

They could also call the appropriate parsing method (in the example below, `AsList`) to parse the field in the debugger, and then navigate into its value:
image

This, however, requires users to inspect the `ValueType` for each field they want to debug, and then call the appropriate method in the debugger. This can be very cumbersome for nested fields, which makes the debugging experience annoying.

For this reason, we're adding a private `InternalValue` property to the `DocumentField` class that can be seen in the debugger to improve navigation:
image

We're also hiding all the private members with the `DebuggerBrowsable` attribute. This approach has its own drawbacks, but this is just a workaround to improve customer experience for the upcoming preview. It's not a final design.


sdk/formrecognizer/Azure.AI.FormRecognizer/src/DocumentField.cs
```diff
@@ -311,5 +326,23 @@ public AddressValue AsAddress()
 
             return ValueAddress;
         }
+
+        private object InternalValue => ValueType switch
+        {
+            DocumentFieldType.Address => ValueAddress,
+            DocumentFieldType.CountryRegion => ValueCountryRegion,
+            DocumentFieldType.Currency => ValueCurrency,
```
maririos, line 334: if I understand correctly, if the property has a value, the user can see it here.
If it is null, the user will see null right? And then they'll try to access it and then we will throw.
Is this the desired behavior?

kinelski: That's a good point.

After discussing with the service team, they confirmed they can't guarantee the `Value{...}` private properties will always be returned, so we should update our code to reflect the possibility of them being null. This reminds me of some negative feedback we got a couple of times about the conversion methods (`AsDate`, `AsTime`, etc) throwing exceptions customers were not expecting, and about considering implementing them with the `Try` pattern in .NET (such as in `TryConvertToInt64`).

From my perspective, the implementation of `InternalValue` is more aligned to what we want for GA, and conversion methods should be updated somehow. I agree that having them behaving differently can lead to confusion. What do you think about updating `InternalValue` to rely on the current conversion methods so they have the same behavior for now? For example:
```cs
        private object InternalValue => ValueType switch
        {
            DocumentFieldType.Address => AsAddress(),
            DocumentFieldType.CountryRegion => AsCountryRegion(),
            DocumentFieldType.Currency => AsCurrency(),
            // ....
```

sdk/formrecognizer/Azure.AI.FormRecognizer/src/DocumentField.cs
```diff
@@ -31,23 +32,32 @@ internal DocumentField(SelectionMarkState? value, string content, IReadOnlyList<
         [CodeGenMember("Type")]
         public DocumentFieldType ValueType { get; }
```
pallavit, line 33: >DocumentFieldType [](http://example.com/codeflow?start=15&length=17)

You mentioned in the description that this approach has some drawbacks - curious what they are.

sdk/formrecognizer/Azure.AI.FormRecognizer/src/DocumentField.cs
```diff
@@ -311,5 +326,23 @@ public AddressValue AsAddress()
 
             return ValueAddress;
         }
+
+        private object InternalValue => ValueType switch
+        {
+            DocumentFieldType.Address => AsAddress(),
+            DocumentFieldType.CountryRegion => AsCountryRegion(),
```
pallavit, line 333: You mentioned we have seen this pattern in TextAnalytics too. Is this a pattern we should account for by default in our codegen?

i.e. debuggability experience for a type that is essentially being converted from a catch-all type to a strongly typed design in convenience layer will need extra debuggability support.
Is there a swagger check that we can use to avoid this?
If this is a common enough pattern should we consider this in codegen via transform\attributes etc?

sdk/formrecognizer/Azure.AI.FormRecognizer/src/DocumentField.cs
maririos: > What do you think about updating InternalValue to rely on the current conversion methods so they have the same behavior for now

Agree and as you said, we can test this out on the beta release and keep looking into it before GA. It would be good ot have an issue tracking the feedback so that this gets look at before GA

maririos (approved):

sdk/formrecognizer/Azure.AI.FormRecognizer/src/DocumentField.cs
```diff
@@ -311,5 +326,23 @@ public AddressValue AsAddress()
 
             return ValueAddress;
         }
+
+        private object InternalValue => ValueType switch
```
tg-msft, line 330: I worry this doesn't help enough with debugging.  Seeing a magic private field doesn't tell me how to go get the value myself in the way a `ToString` method that rendered as `[DocumentField: ValueType=String, AsString()=>Hello, World]` could.  Should we also make this a `public object GetValue()` method for anyone who wants to handle values generically?

sdk/formrecognizer/Azure.AI.FormRecognizer/src/DocumentField.cs
```diff
@@ -311,5 +326,23 @@ public AddressValue AsAddress()
...
+            DocumentFieldType.Dictionary => AsDictionary(),
+            DocumentFieldType.Double => AsDouble(),
+            DocumentFieldType.Int64 => AsInt64(),
+            DocumentFieldType.List => AsList(),
+            DocumentFieldType.PhoneNumber => AsPhoneNumber(),
+            DocumentFieldType.SelectionMark => AsSelectionMarkState(),
+            DocumentFieldType.Signature => AsSignatureType(),
+            DocumentFieldType.String => AsString(),
+            DocumentFieldType.Time => AsTime(),
+            _ => null
```
tg-msft, line 345: If we remove this final case, will the compiler verify exhaustiveness for us?

tg-msft (approved):

sdk/formrecognizer/Azure.AI.FormRecognizer/src/DocumentField.cs
kinelski: Yes. It gives us a warning (CS8524) if we remove it:
> The switch expression does not handle some values of its input type (it is not exhaustive) involving an unnamed enum value. For example, the pattern '(Azure.AI.FormRecognizer.DocumentAnalysis.DocumentFieldType)13' is not covered.

sdk/formrecognizer/Azure.AI.FormRecognizer/src/DocumentField.cs
kinelski: I think the private property approach improves debugging navigation since users won't need to call the conversion methods to access the field's internal value. On the other hand, I agree it may not solve the problem for users who have no idea on how to extract the value.

If we had a `public object GetValue()` method, wouldn't it be better to expose it as a `public object Value { get; }` directly? From my understanding, we added the conversion methods because users could struggle relying on casting, specially as we add more structured field values (such as `Currency`, `Address`, etc).

sdk/formrecognizer/Azure.AI.FormRecognizer/src/DocumentField.cs
kinelski: We have an open issue for it: https://github.com/Azure/azure-sdk-for-net/issues/29013
I'm not closing it with this PR since I can't say this is a final design yet. `DocumentField` will change a bit until GA.

sdk/formrecognizer/Azure.AI.FormRecognizer/src/DocumentField.cs
kinelski: Regarding the `ToString` implementation, what would be a good message for references? `AsCurrency`, for example, returns a `ValueCurrency`. Does `[DocumentField: ValueType=Currency, AsCurrency()=>Value]` sound reasonable?

sdk/formrecognizer/Azure.AI.FormRecognizer/src/DocumentField.cs
kinelski: My concerns is users would be inclined to attempt extracting the `InternalValue` object somehow. Not noticing it's private, some will try to find it in intellisense and think it's a bug when they can't find it. The reason we don't like exposing it as an `object` directly is because users usually struggle when trying to figure out which type they should cast to. That's why we introduced the `DocumentField` conversion methods (`AsInt64`, `AsDouble`, `AsAddress`, etc).

sdk/formrecognizer/Azure.AI.FormRecognizer/src/DocumentField.cs
kinelski: I actually mentioned we used the [DebuggerTypeProxyAttribute](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.debuggertypeproxyattribute?view=net-6.0) in TA, but the scenario there was a bit different ([TA code](https://github.com/Azure/azure-sdk-for-net/blob/6880fd3930bb5cd7531850a0ec88eab109a61c93/sdk/textanalytics/Azure.AI.TextAnalytics/src/AnalyzeHealthcareEntitiesResultCollection.cs#L45)). In TA, we wanted to expose a property (`Items`) that was hard to find in the debugger because it was nested inside other elements, so we simply moved it to the root level.

Originally I was thinking about using the `DebuggerTypeProxyAttribute` as well, but after some consideration I realized it was an overkill. The approach I'm taking in this PR uses less code, but I can't say it's a final design.

kinelski: > i.e. debuggability experience for a type that is essentially being converted from a catch-all type to a strongly typed design in convenience layer will need extra debuggability support. Is there a swagger check that we can use to avoid this?

From my understanding the swagger does not have any kind of flag to distinguish this scenario from others. Here's the `DocumentField` definition: [swagger](https://github.com/Azure/azure-rest-api-specs/blob/a69db59e981eae7dc554823100220c81b34225cc/specification/cognitiveservices/data-plane/FormRecognizer/preview/2022-06-30-preview/FormRecognizer.json#L1849)

It only defines all of its value properties (`valueString`, `valueNumber`, `valueTime`, etc) as optional, and we trust the service will send back the one that corresponds to the `type` property.

I've only seen this pattern being used in `DocumentField`, and haven't seen it in other libraries.

pallavit: Thanks Caio

pallavit (approved):
:shipit:

sdk/formrecognizer/Azure.AI.FormRecognizer/src/DocumentField.cs
pallavit: I like [DocumentField: ValueType=Currency, AsCurrency()=>Value] - where value is actually the result of the AsCurrency call. This seems like a good start and we can iterate over it as we get more feedback.

kinelski: I was actually thinking of literally writing out `Value`. If we print the actual internal value of the field, I think we could end up with a long string and not being too helpful. For example, for currency:
image

Or even for long strings:
image

pallavit: sounds good.

PR #29148, opened 2022-06-07
Sql test record2022.05

### LINKED ISSUE #28621: Re-record SQL tests
The Azure.Resourcemanager.Sql SDK has updated API versions for some resources but the tests are not re-recorded. It's not caught in CI as the tests are marked with `RunTestFrequency.Manually` and skipped in CI tests. We should re-record the tests before GA.

mcgallan:
Completed Sql Test re-record


Yao725 (dismissed):
Nice work, but I think we should remove the attribute `RunFrequency` in `SqlManagementClientBase.cs`, otherwise the tests will not run in the pipeline.

mcgallan:
@fengzhou-msft I made some changes to this part of the Sql test, on the one hand, because the sql test takes too much time to run, so I implemented playback passing by modifying the json file in SessionRecords, and on the other hand, in the SqlManagementClientBase .cs，The original code 
image
caused an error to be reported，because
image
So I changed SourceType="Microsoft.Sql" to SourceType="/Microsoft.Sql" to ensure that the test case passed, but the type in the json file also changed
image
I wasn't sure if there was a more appropriate way to modify it, so I left this comment

sdk/sqlmanagement/Azure.ResourceManager.Sql/tests/Scenario/ManagedInstanceTests.cs
```diff
@@ -62,7 +62,7 @@ public async Task GlobalSetUp()
                         AddressPrefix = "10.10.2.0/24",
                         Delegations =
                         {
-                            new ServiceDelegation() { ServiceName  = "Microsoft.Sql/managedInstances",Name="Microsoft.Sql/managedInstances" ,ResourceType="Microsoft.Sql"}
+                            new ServiceDelegation() { ServiceName  = "Microsoft.Sql/managedInstances",Name="Microsoft.Sql/managedInstances" ,ResourceType="/Microsoft.Sql"}
```
Yao725, line 65: Is the slash used to bypass the `ResourceType` validation? Based on @ArthurMa1978's investigation, can we change this property value to `Microsoft.Sql/managedInstances`?

Yao725 (commented):
Could you also please resolve the conflicts?

sdk/sqlmanagement/Azure.ResourceManager.Sql/tests/Scenario/ManagedInstanceTests.cs
mcgallan: All the above modifications are complete

fengzhou-msft (approved):

Yao725 (approved):

PR #29059, opened 2022-06-01
Initial Azure.ResourceManager.AppPlatform

ArthurMa1978:

PR #29152, opened 2022-06-07
Fix the path for all ci settings

ArthurMa1978:


jsquire:
@ArthurMa1978:  Can you help me understand why this change is needed?

ArthurMa1978:
> @ArthurMa1978: Can you help me understand why this change is needed?

https://github.com/Azure/azure-sdk-for-net/pull/29059#discussion_r888302856
@benbp suggested to change this.

ArthurMa1978:
> > @benbp suggested to change this.
> 
> Thank you for the context. In that case, I'll defer to the engsys team for approvals.

I will ping @benbp ~

weshaggard (approved):

weshaggard:
Thanks for the fix @ArthurMa1978. For future we need to be careful about changing a ton of ci.yml files across the repo because as you can see it triggers a ton of pipelines and can cause a build storm where our queue gets backed up.

Issue #29125, opened 2022-06-06
[FEATURE REQ] - Completion and Complete on ServiceBusProcessor


Rayzbam:
### Library name

Azure.Messaging.ServiceBus

### Please describe the feature.

I'm quite a big user of the TPL library because it served a lot of my issues and problematics.
Is there any possibility to have any clue on what's happening on the ServiceBusProcess instances ?

I mean, it's possible to call "StopProcessingAsync" to stop receiving messages, but what happen if it still were messages processing asynchronously (MaxConcurrentCall > 1) ? 
When i try to gracefully shutdown my service, what i would do is : 
- Stop Processing (stop receiving messages)
- Wait for processor messages that are still in progress in background
- Dispose Processors resources

In TPL library, what you can do is simply :

```
actionBlock.Complete(); // Equivalent of StopProcessingAsync on ServiceBusProcessor
await actionBlock.Completion;
```

Found an issue, which seems linked to my request : [https://github.com/Azure/azure-sdk-for-net/issues/21869](https://github.com/Azure/azure-sdk-for-net/issues/21869)

Is it embedded in the StopProcessingAsync ? The doc doesn't mention anything about work in progress, i can only see "The method will cause the receivers to stop receiving" not "and finish its work in progress".

Any thoughts ?

JoshLove-msft:
Hi @Rayzbam,
Calling StopProcessingAsync or CloseAsync will allow in-flight processing to complete. The difference between the two methods is that StopProcessingAsync will leave the links open (in case you want to resume processing), so you would still want to call CloseAsync at some point when shutting down.

JoshLove-msft:
Sorry if I wasn't clear, but StopProcessingAsync/CloseAsync will not return until all of your in-flight processing has completed, so there is no need to track the asks yourself.

JoshLove-msft:
> Is that possible to add a few words in the documentation to explain that in-flight processing messages are awaited too ?

Absolutely - I will submit a PR to improve the ref docs for these methods.

PR #29159, opened 2022-06-07
Improve docs for stopping processor methods.

JoshLove-msft:
Addresses https://github.com/Azure/azure-sdk-for-net/issues/29125

### DIFF
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessor.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessor.cs
@@ -723,9 +723,11 @@ namespace Azure.Messaging.ServiceBus
         /// Signals the processor to stop processing messaging. Should this method be
         /// called while the processor is not running, no action is taken. This method
         /// will not close the underlying receivers, but will cause the receivers to stop
-        /// receiving. To close the underlying receivers, <see cref="CloseAsync"/> should be called.
-        /// Any in-flight message handlers will be awaited, and this method will not return
-        /// until all in-flight message handlers have returned.
+        /// receiving. Any in-flight message handlers will be awaited, and this method will not return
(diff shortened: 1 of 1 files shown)


sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessor.cs
```diff
@@ -724,6 +724,8 @@ private void ValidateMessageHandler()
         /// called while the processor is not running, no action is taken. This method
         /// will not close the underlying receivers, but will cause the receivers to stop
         /// receiving. To close the underlying receivers, <see cref="CloseAsync"/> should be called.
+        /// Any in-flight message handlers will be awaited, and this method will not return
```
jsquire, line 727: Would it be worth explicitly calling out that no messages will be processed after this method as returned unless `StartAsync` is called again?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessor.cs
```diff
@@ -999,8 +1001,9 @@ internal void EnsureNotRunningAndInvoke(Action action)
         }
 
         /// <summary>
-        ///   Performs the task needed to clean up resources used by the <see cref="ServiceBusProcessor" />.
-        ///   This is equivalent to calling <see cref="CloseAsync"/>.
+        /// Performs the task needed to clean up resources used by the <see cref="ServiceBusProcessor" />. Any in-flight message handlers will
+        /// be awaited. Once all message handling has completed, the underlying links will be closed. After this point, the method will return.
+        /// This is equivalent to calling <see cref="CloseAsync"/>.
```
jsquire, line 1006: I like this line....  I think I may steal it!

jsquire (approved):

PR #29164, opened 2022-06-07
Remove OperationOrResponseInternals dependencies

AlexanderSher:
Remove remaining dependencies on OperationOrResponseInternals so that type can be removed from autorest.csharp


christothes (approved):

sdk/core/Azure.Core/tests/TestClients/TestResourceOperationOrResponse.cs
```diff
@@ -12,7 +12,7 @@ internal class TestResourceOperationOrResponse : Operation
     {
         private TestResource _value;
         private bool _exceptionOnWait;
-        private OperationOrResponseInternals<TestResource> _operationHelper;
+        private OperationInternal<TestResource> _operationHelper;
         private int _delaySteps = 0;
 
         protected TestResourceOperationOrResponse()
```
annelo-msft, line 18: Should the name of the test class/test methods be changed so that they reflect which type is being tested?

AlexanderSher: We don't test `OperationOrResponseInternals` with this type. This type is a mock.
@m-nash, what do you think about naming?

m-nash: I would imagine if we are consolidating on Operation and getting rid of OperationOrResponse we should get rid of these classes as well and consolidate the tests.

AlexanderSher: We can't get rid of these test classes cause ARM still generates individual `Operation` types (one per project). But I've changed the name.

PR #29166, opened 2022-06-07
Add catch all code owners for the repo

pallavit:
Add the catch all owners for the repo. In some places we depend on this fallback having an owner (example docs or pipeline ownership) for any paths that do not have a natural owner.

https://github.com/Azure/azure-sdk-for-net/pull/28768#discussion_r891756947_

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -13,7 +13,7 @@
 #########
 
 # Catch all
-/sdk/                                           
+/sdk/                                           @pallavit @jsquire
 
 # ######## Core Libraries ########
 


jsquire (changes requested):
I do not want to be a catch-all _(directly?)_ and I'm not sure that you do either.   Discussion in #28768

weshaggard:
>@weshaggard - Can we confirm behavior to be sure that my mental model is correct? My understanding is that this will essentially end up sending alerts for anything in the repository that doesn't have a more specific path. Assuming that I'm correct, I'm not comfortable having myself here (I'll let Pallavi speak for herself)
I'm still actively working in parts of the repository and rely on GH notifications via email for these. The last I spoke with Alex, he received enough spam from the catch-all notifications that all GH emails were noise to him and were dealt with by filtering into an "ignore" area until someone spoke to him directly. That's not a situation that's going to work for me.
Can we define what areas, specifically, need the leads to watch and add explicit ownership for those areas? Alternately, lets consider creating a GH handle and DG that Pallavi and I can subscribe to which I can use for filtering notifications for catch-all from those that I explicitly opt into.

Moving this conversation over here.

Part of the reason we put the lead(s) as the catch-all is to ensure new directories end-up with owners. Ideally the catch-all wouldn't be noisy once there is a more specific owner path in the file. So if you guys are seeing missing paths then the fix is to add a more specific path and identify the correct owners for that path. 

Outside of just being added as code owner in PR reviews other places we use this data is with pipeline failure notifications, and with documentation generation owners. So we really need some way to ensure that this is as complete as possible and we felt the leads would be the best people to help ensure it is complete.

jsquire:
> So we really need some way to ensure that this is as complete as possible and we felt the leads would be the best people to help ensure it is complete.

I'm not opposed, in principle.   I'd prefer that we go down the path of creating a GH handle and DG that Pallavi and I can join rather than having this added directly.   That'll allow me to more reliably set Outlook rules and ensure that I don't miss things that I need to see.

jsquire:
We'll try it for now; if it breaks GitHub workflows, I'll remove myself and we can revisit an alias to allow for outlook rules.

jsquire (approved):

PR #29180, opened 2022-06-08
docs(core): add an example of custom polling

### LINKED ISSUE #29072: [mgmt]: reconsider lro polling strategy to honour user input
## Problem

We already have a customer issue regarding LRO in track1. The root cause is on the server side which doesn't return the correct `Retry-After`. However, retrospecting our track2 LRO implementation, there is one issue found: we totally ignore user input interval, thus in such cases there is no way for customers to work around.

Here is the `RetryAfterDelayStrategy`: https://github.com/Azure/azure-sdk-for-net/blob/8914c0e7076f82a95efa8a1d6cc955dd69b04c05/sdk/core/Azure.Core/src/Shared/RetryAfterDelayStrategy.cs#L41-L58

We choose the max value of `Retry-After` header and the fall-back strategy. And for mgmt, the fall-back strategy is `ExponentialDelayStrategy`: https://github.com/Azure/azure-sdk-for-net/blob/8914c0e7076f82a95efa8a1d6cc955dd69b04c05/sdk/core/Azure.Core/src/Shared/ExponentialDelayStrategy.cs#L40-L44

which ignores the user input. So there is no way for users to impact the polling interval, thus they lose the capability to remediate some cases.

archerzz:
- add a sample case in `OperationSamples`
- add a section `Custom polling for completion of operation` in document

resolve #29072

### DIFF
--- a/sdk/core/Azure.Core/tests/samples/OperationSamples.cs
+++ b/sdk/core/Azure.Core/tests/samples/OperationSamples.cs
@@ -63,18 +63,13 @@ namespace Azure.Core.Samples
             // Start the operation
             DeleteSecretOperation operation = await client.StartDeleteSecretAsync("SecretName");
 
-            while (true)
+            while (operation.HasCompleted)
             {
                 // there could be more complex logic to control polling interval
                 Thread.Sleep(TimeSpan.FromSeconds(60));
 
-                Response state = await operation.UpdateStatusAsync();
-                Console.WriteLine($"Operation status is {state}");
-
-                if (operation.HasCompleted)
-                {
-                    break;
-                }
+                Response statusResponse = await operation.UpdateStatusAsync();
+                // optionally do something with the statusResponse
             }
             #endregion
         }
--- a/sdk/core/Azure.Core/tests/samples/OperationSamples.cs
+++ b/sdk/core/Azure.Core/tests/samples/OperationSamples.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Threading;
 using System.Threading.Tasks;
 using Azure.Identity;
 using Azure.Security.KeyVault.Secrets;
@@ -48,6 +49,33 @@ namespace Azure.Core.Samples
             // HasValue indicated is operation Value is available, for some operations it can return true even when operation
             // hasn't completed yet.
             Console.WriteLine(operation.HasValue);
+            #endregion
+        }
+
+        [Test]
+        [Ignore("Only verifying that the sample builds")]
+        public async Task OperationCustomPollingCompletion()
+        {
+            // create a client
+            SecretClient client = new SecretClient(new Uri("http://example.com"), new DefaultAzureCredential());
+
+            #region Snippet:OperationCustomPollingCompletion
+            // Start the operation
+            DeleteSecretOperation operation = await client.StartDeleteSecretAsync("SecretName");
+
+            while (true)
+            {
+                // there could be more complex logic to control polling interval
+                Thread.Sleep(TimeSpan.FromSeconds(60));
+
+                Response state = await operation.UpdateStatusAsync();
+                Console.WriteLine($"Operation status is {state}");
+
+                if (operation.HasCompleted)
+                {
+                    break;
+                }
+            }
             #endregion
         }
 
--- a/sdk/core/Azure.Core/tests/samples/OperationSamples.cs
+++ b/sdk/core/Azure.Core/tests/samples/OperationSamples.cs
@@ -66,7 +66,7 @@ namespace Azure.Core.Samples
             while (operation.HasCompleted)
             {
                 // there could be more complex logic to control polling interval
-                Thread.Sleep(TimeSpan.FromSeconds(60));
+                await Task.Delay(TimeSpan.FromSeconds(60));
 
                 Response statusResponse = await operation.UpdateStatusAsync();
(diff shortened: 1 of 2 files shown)


archerzz:
We've been contact for several times regarding custom polling. So, it would be better we add an example in our document.

sdk/core/Azure.Core/tests/samples/OperationSamples.cs
```diff
@@ -51,6 +52,33 @@ public async Task OperationUpdateStatus()
...
+                Thread.Sleep(TimeSpan.FromSeconds(60));
+
+                Response state = await operation.UpdateStatusAsync();
+                Console.WriteLine($"Operation status is {state}");
+
+                if (operation.HasCompleted)
+                {
+                    break;
+                }
+            }
```
christothes, line 78: I think we can simplify the loop. Also, I think it may be better not to try and print the response to the console as it would n't provide any information about the response content.

```suggestion
            while (operation.HasCompleted)
            {
                // there could be more complex logic to control polling interval
                Thread.Sleep(TimeSpan.FromSeconds(60));

                Response statusResponse = await operation.UpdateStatusAsync();
                // optionally do something with the statusResponse
            }
```

annelo-msft (commented):
I'm curious how broadly this approach has been agreed upon.  We shouldn't provide a sample to the Azure.Core docs for something we don't recommend for all of Azure.Core.

archerzz:
> I'm curious how broadly this approach has been agreed upon. We shouldn't provide a sample to the Azure.Core docs for something we don't recommend for all of Azure.Core.

@annelo-msft We have been contacted at least twice for customer cases (@m-nash and me respectively). The solution from SDK side is to customize the polling behavior. In my case(#29072), the `Retry-After` header is not correctly set, so basically the service is asking the clients to DDos itself. In current implementation of mgmt track2 polling, the `TimeSpan` parameter in `WaitForCompletion` actually doesn't work. So previously I thought we should make the user input `TimeSpan` work for some cases, so that the customers can input a big value to avoid huge traffic to the server side (it's just a remediation). 

But after discussion, we thought it might be better to let users fully control the polling since we cannot foresee all edge cases.
The capability provided by `Azure.Core` is better than track1 in which it's hard to customize your own polling behavior. I thought it is useful, though not realized that it's not recommended :)

sdk/core/Azure.Core/tests/samples/OperationSamples.cs
```diff
@@ -51,6 +52,28 @@ public async Task OperationUpdateStatus()
...
+        public async Task OperationCustomPollingCompletion()
+        {
+            // create a client
+            SecretClient client = new SecretClient(new Uri("http://example.com"), new DefaultAzureCredential());
+
+            #region Snippet:OperationCustomPollingCompletion
+            // Start the operation
+            DeleteSecretOperation operation = await client.StartDeleteSecretAsync("SecretName");
+
+            while (operation.HasCompleted)
```
christothes, line 66:
```suggestion
            while (!operation.HasCompleted)
```

sdk/core/Azure.Core/tests/samples/OperationSamples.cs
```diff
@@ -51,6 +52,28 @@ public async Task OperationUpdateStatus()
...
+            {
+                // there could be more complex logic to control polling interval
+                Thread.Sleep(TimeSpan.FromSeconds(60));
```
christothes, line 69: If the example is going to be async, we can do this instead.

```suggestion
                await Task.Delay(TimeSpan.FromSeconds(60));
```

nishanthm-msft:
Hi, we are migrating from Track 1 to Track 2 SDK. I need to track a long running operation in an async manner using its operation Id. It would not be feasible to wait synchronously for the LRO to complete. So, we do not want to wait until completion (synchronously) and rather would need to jump to do other tasks meanwhile the LRO is going on and come back and be able to poll the operation status and act accordingly.

We were able to do this in Track 1 SDK using https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/recoveryservices-backup/Microsoft.Azure.Management.RecoveryServices.Backup/src/recoveryservicesbackup/Generated/BackupOperationStatusesOperations.cs#L92 where the LRO status could be polled using its Operation Id. But similar functionality is not available in Track 2 SDK.

This is a useful functionality and can this be addressed in Track 2 SDK?

m-nash:
@nishanthm-msft are you doing this within the same process?  If so you can store the ArmOperation objects you get back and periodically call `UpdateStatus` on them to see if they are finished.

If you are doing this across processes then yes you will need to send the `Id` property to the other process so that it can rehydrate the LRO and pick up where it left off.

nishanthm-msft:
> @nishanthm-msft are you doing this within the same process? If so you can store the ArmOperation objects you get back and periodically call `UpdateStatus` on them to see if they are finished.
> 
> If you are doing this across processes then yes you will need to send the `Id` property to the other process so that it can rehydrate the LRO and pick up where it left off which is the process that is not complete yet.

@m-nash I am doing this across processes. So, the process which executes this operation is actually different from the one tracking it. Can you please share your thoughts on the below approach which I thought could serve the purpose.

Can we pass the reference of the entire ArmOperation object itself as it is (instead of Id alone as rehydration isn't available yet) that we get back from the process executing this LRO to the **_calling process_** and then later poll the status of the LRO when required from the **_calling process_** itself using `UpdateStatus`?

PR #29208, opened 2022-06-09
Fix the plurality naming issue for KeyVault

ArthurMa1978:
Since the [auto-resolve solution](https://github.com/Azure/autorest.csharp/pull/2296) will not be merged soon, here manually fix it to unblock GA


Yao725 (commented):
I think you might miss the change in another place, because I don't find the old plural enum is deleted.

I have double checked, this model is also used in `ManagedHsmNetworkRuleSet`.

ArthurMa1978:
> I think you might miss the change in another place, because I don't find the old plural enum is deleted.
> 
> I have double checked, this model is also used in `ManagedHsmNetworkRuleSet`.

Good catch, fixed

Yao725 (approved):

PR #29220, opened 2022-06-09
[AzureMonitorExporter] implementing Truncation rules for ExceptionData

TimothyMothra:
Continuation of #29094

## Changes
- implement truncation for `ExceptionData`
- `ExceptionData.Properties` is set in `LogHelper`
- refactor `SchemaConstants`
  - added `KVP_MaxKeyLength` and `KVP_MaxValueLength`
    these generic constants are to be used in `LogHelper`.
    the specific constants are still needed to track limits in other classes.

### DIFF
--- a/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/LogsHelper.cs
+++ b/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/LogsHelper.cs
@@ -80,7 +80,7 @@ namespace Azure.Monitor.OpenTelemetry.Exporter
 
             if (!string.IsNullOrEmpty(logRecord.EventId.Name))
             {
-                properties.Add("EventName", logRecord.EventId.Name);
+                properties.Add("EventName", logRecord.EventId.Name.Truncate(SchemaConstants.KVP_MaxValueLength));
             }
 
             return message;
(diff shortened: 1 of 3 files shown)


sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/LogsHelper.cs
```diff
@@ -80,7 +80,7 @@ internal static string GetMessageAndSetProperties(LogRecord logRecord, IDictiona
 
             if (!string.IsNullOrEmpty(logRecord.EventId.Name))
             {
-                properties.Add("EventName", logRecord.EventId.Name);
+                properties.Add("EventName", logRecord.EventId.Name.Truncate(SchemaConstants.KVP_MaxValueLength));
             }
 
             return message;
```
vishweshbankwar, line 86: Truncate message?

vishweshbankwar (approved):
LGTM

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/LogsHelper.cs
TimothyMothra: I'm wary of doing too much truncation in the `LogsHelper` class because this class is referenced by multiple types and it's tough to verify where these values are going.

In this case the `message` gets passed into `LogsHelper.ConvertExceptionTree()` which is a recursive method that can overwrite `message`.

vishweshbankwar: Makes sense

PR #29234, opened 2022-06-10
Preview serialization update

fengzhou-msft:


sdk/network/Azure.ResourceManager.Network/src/Generated/Models/NetworkInterfaceIPConfigurationData.Serialization.cs
```diff
@@ -130,7 +132,8 @@ internal static NetworkInterfaceIPConfigurationData DeserializeNetworkInterfaceI
             Optional<ETag> etag = default;
             Optional<ResourceIdentifier> id = default;
             Optional<string> name = default;
-            Optional<ResourceType> type = default;
+            Optional<ResourceType?> type = default;
```
fengzhou-msft, line 135: This looks problematic.

eng/Packages.Data.props
```diff
@@ -97,7 +97,7 @@
     <PackageReference Update="Azure.Security.KeyVault.Certificates" Version="4.2.0" />
     <PackageReference Update="Azure.Storage.Blobs" Version="12.12.0" />
     <PackageReference Update="Azure.Storage.Queues" Version="12.10.0" />
-    <PackageReference Update="Azure.ResourceManager" Version="1.0.0" />
+    <PackageReference Update="Azure.ResourceManager" Version="1.2.0-alpha.20220609.1" />
```
m-nash, line 100: This is fine as a POC but shouldn't be checked in.

sdk/resources/Azure.ResourceManager.Resources/src/Generated/Models/AzureCliScript.Serialization.cs
```diff
@@ -117,7 +117,7 @@ internal static AzureCliScript DeserializeAzureCliScript(JsonElement element)
             ResourceIdentifier id = default;
             string name = default;
             ResourceType type = default;
-            SystemData systemData = default;
+            Optional<SystemData> systemData = default;
```
m-nash, line 120: why does this result in only serialization changes without changing any base types in resources package?

sdk/resourcemanager/Azure.ResourceManager/src/Common/Custom/Models/WritableResourceData.cs
```diff
@@ -0,0 +1,43 @@
...
+
+#nullable disable
+
+using Azure.Core;
+
+namespace Azure.ResourceManager.Models
+{
+    /// <summary> Common fields that are returned in the response for all Azure Resource Manager resources. </summary>
+    [ReferenceType(new string[]{"SystemData"})]
+    public abstract partial class WritableResourceData
```
m-nash, line 12: Did we run the same analysis to see if we want a WriteableTrackedResourceData as well?

m-nash (changes requested):

sdk/resources/Azure.ResourceManager.Resources/src/Generated/Models/AzureCliScript.Serialization.cs
fengzhou-msft: Previously, during replacement with the common type, we hardcoded `Required` as true for its properties. I have updated generator to fix that. See SystemObjectType code [here](https://github.com/Azure/autorest.csharp/pull/2303/files#diff-32a4ccc7bb0ed76e7043f8ac39b8b80af8dcbf850aeb1add76ccd922b6215c8e).

sdk/resourcemanager/Azure.ResourceManager/src/Common/Custom/Models/WritableResourceData.cs
fengzhou-msft: Yes. It turns out to be not common for id, name or type to be writable in TrackedResourceData. The only valid case I met with is writable ids in network.

PR #29236, opened 2022-06-10
docs(dpg): generate <example> comment

archerzz:
preview only


sdk/confidentialledger/Azure.Security.ConfidentialLedger/src/Generated/ConfidentialLedgerClient.cs
```diff
@@ -46,6 +46,18 @@ public ConfidentialLedgerClient(Uri ledgerUri, TokenCredential credential) : thi
 
         /// <summary> The constitution is a script that assesses and applies proposals from consortium members. </summary>
         /// <param name="context"> The request context, which can override default behaviors on the request on a per-call basis. </param>
+        /// <example>
+        /// This sample shows how to call GetConstitutionAsync.
+        /// <code>
+        /// var credential = new DefaultAzureCredential();
+        /// var endpoint = new Uri("https://my-account-name.azure.com");
+        /// var client = new ConfidentialLedgerClient(endpoint, credential);
```
annelo-msft, line 54: Please add a carriage return after this line.

sdk/confidentialledger/Azure.Security.ConfidentialLedger/src/Generated/ConfidentialLedgerClient.cs
```diff
@@ -46,6 +46,18 @@ public ConfidentialLedgerClient(Uri ledgerUri, TokenCredential credential) : thi
 
         /// <summary> The constitution is a script that assesses and applies proposals from consortium members. </summary>
         /// <param name="context"> The request context, which can override default behaviors on the request on a per-call basis. </param>
+        /// <example>
+        /// This sample shows how to call GetConstitutionAsync.
```
annelo-msft, line 50: Could we add "and parse the result" for cases that we parse a JSON body out of the response?

sdk/confidentialledger/Azure.Security.ConfidentialLedger/src/Generated/ConfidentialLedgerClient.cs
```diff
@@ -272,6 +350,36 @@ public virtual Response GetEnclaveQuotes(RequestContext context = null)
...
         /// <param name="subLedgerId"> The sub-ledger id. </param>
         /// <param name="context"> The request context, which can override default behaviors on the request on a per-call basis. </param>
+        /// <example>
+        /// This sample shows how to call PostLedgerEntryAsync with required parameters and properties.
+        /// <code>
+        /// var credential = new DefaultAzureCredential();
+        /// var endpoint = new Uri("https://my-account-name.azure.com");
+        /// var client = new ConfidentialLedgerClient(endpoint, credential);
+        /// var data = new {
+        ///     contents = "LedgerEntryContents",
```
annelo-msft, line 360: Can you remind me how we decided to set the value of sample inputs?  I'm wondering if there's a way we can indicate it's a sample input and not a real input, let me follow up on this with the .NET architects.

sdk/confidentialledger/Azure.Security.ConfidentialLedger/src/Generated/ConfidentialLedgerClient.cs
```diff
@@ -362,6 +500,29 @@ public virtual Response PostLedgerEntry(RequestContent content, string subLedger
         /// <param name="context"> The request context, which can override default behaviors on the request on a per-call basis. </param>
         /// <exception cref="ArgumentNullException"> <paramref name="transactionId"/> is null. </exception>
         /// <exception cref="ArgumentException"> <paramref name="transactionId"/> is an empty string, and was expected to be non-empty. </exception>
+        /// <example>
+        /// This sample shows how to call GetLedgerEntryAsync with required parameters and properties.
```
annelo-msft, line 504: What are the properties mentioned here?  It looks like the method is being called with a single parameter.

sdk/confidentialledger/Azure.Security.ConfidentialLedger/src/Generated/ConfidentialLedgerClient.cs
```diff
@@ -408,6 +569,29 @@ public virtual async Task<Response> GetLedgerEntryAsync(string transactionId, st
         /// <param name="context"> The request context, which can override default behaviors on the request on a per-call basis. </param>
         /// <exception cref="ArgumentNullException"> <paramref name="transactionId"/> is null. </exception>
         /// <exception cref="ArgumentException"> <paramref name="transactionId"/> is an empty string, and was expected to be non-empty. </exception>
+        /// <example>
+        /// This sample shows how to call GetLedgerEntry with required parameters and properties.
+        /// <code>
+        /// var credential = new DefaultAzureCredential();
+        /// var endpoint = new Uri("https://my-account-name.azure.com");
+        /// var client = new ConfidentialLedgerClient(endpoint, credential);
```
annelo-msft, line 577: Let's put a carriage return after this line as well.

sdk/confidentialledger/Azure.Security.ConfidentialLedger/src/Generated/ConfidentialLedgerClient.cs
```diff
@@ -870,6 +1200,36 @@ public virtual Response GetUser(string userId, RequestContext context = null)
         /// <param name="context"> The request context, which can override default behaviors on the request on a per-call basis. </param>
         /// <exception cref="ArgumentNullException"> <paramref name="userId"/> or <paramref name="content"/> is null. </exception>
         /// <exception cref="ArgumentException"> <paramref name="userId"/> is an empty string, and was expected to be non-empty. </exception>
+        /// <example>
+        /// This sample shows how to call CreateOrUpdateUserAsync with required parameters and properties.
```
annelo-msft, line 1204: Instead of properties, should we say request content?

sdk/confidentialledger/Azure.Security.ConfidentialLedger/src/Generated/ConfidentialLedgerClient.cs
```diff
@@ -870,6 +1200,36 @@ public virtual Response GetUser(string userId, RequestContext context = null)
...
+        /// <code>
+        /// var credential = new DefaultAzureCredential();
+        /// var endpoint = new Uri("https://my-account-name.azure.com");
+        /// var client = new ConfidentialLedgerClient(endpoint, credential);
+        /// var data = new {
+        ///     assignedRole = "Administrator",
```
annelo-msft, line 1210: How did we choose to populate the example value with "Administrator" instead of "AssignedRole"?

sdk/webpubsub/Azure.Messaging.WebPubSub/src/Generated/HealthApiClient.cs
```diff
@@ -73,6 +82,15 @@ public virtual async Task<Response> GetServiceStatusAsync(RequestContext context
 
         /// <summary> Get service health status. </summary>
         /// <param name="context"> The request context, which can override default behaviors on the request on a per-call basis. </param>
+        /// <example>
+        /// This sample shows how to call GetServiceStatus.
+        /// <code>
+        /// var credential = new DefaultAzureCredential();
+        /// var endpoint = new Uri("https://my-account-name.azure.com");
+        /// var client = new HealthApiClient(endpoint, credential);
+        /// Response response = client.GetServiceStatus();
```
annelo-msft, line 91: I'm wondering if we want an example if there is no input content or response body content.

Where is the service status found on the response in this call?  Ideally we would show that.

sdk/webpubsub/Azure.Messaging.WebPubSub/src/Generated/WebPubSubServiceClient.cs
```diff
@@ -345,6 +475,22 @@ internal virtual Response ConnectionExistsImpl(string connectionId, RequestConte
...
+        /// var endpoint = new Uri("https://my-account-name.azure.com");
+        /// var client = new WebPubSubServiceClient(endpoint, credential);
+        /// Response response = await client.CloseConnectionAsync("connectionId");
+        /// </code>
+        /// This sample shows how to call CloseConnectionAsync with all parameters and properties.
+        /// <code>
+        /// var credential = new DefaultAzureCredential();
+        /// var endpoint = new Uri("https://my-account-name.azure.com");
+        /// var client = new WebPubSubServiceClient(endpoint, credential);
+        /// Response response = await client.CloseConnectionAsync("connectionId", null);
```
annelo-msft, line 491: Why are we passing `null` here?

sdk/webpubsub/Azure.Messaging.WebPubSub/src/Generated/WebPubSubServiceClient.cs
```diff
@@ -622,6 +840,22 @@ public virtual async Task<Response> CloseGroupConnectionsAsync(string group, IEn
...
+        /// var endpoint = new Uri("https://my-account-name.azure.com");
+        /// var client = new WebPubSubServiceClient(endpoint, credential);
+        /// Response response = client.CloseGroupConnections("group");
+        /// </code>
+        /// This sample shows how to call CloseGroupConnections with all parameters and properties.
+        /// <code>
+        /// var credential = new DefaultAzureCredential();
+        /// var endpoint = new Uri("https://my-account-name.azure.com");
+        /// var client = new WebPubSubServiceClient(endpoint, credential);
+        /// Response response = client.CloseGroupConnections("group", null, null);
```
annelo-msft, line 856: I think it would be better if we don't pass nulls in the examples, unless this is a common scenario.

sdk/webpubsub/Azure.Messaging.WebPubSub/src/Generated/WebPubSubServiceClient.cs
```diff
@@ -1425,6 +1905,22 @@ internal virtual async Task<Response> CheckPermissionAsync(string permission, st
...
+        /// var endpoint = new Uri("https://my-account-name.azure.com");
+        /// var client = new WebPubSubServiceClient(endpoint, credential);
+        /// Response response = client.CheckPermission("permission", "connectionId");
+        /// </code>
+        /// This sample shows how to call CheckPermission with all parameters and properties.
+        /// <code>
+        /// var credential = new DefaultAzureCredential();
+        /// var endpoint = new Uri("https://my-account-name.azure.com");
+        /// var client = new WebPubSubServiceClient(endpoint, credential);
+        /// Response response = client.CheckPermission("permission", "connectionId", null);
```
annelo-msft, line 1921: Curious what the expectation is for the customer to do with this response to check the permission?

sdk/confidentialledger/Azure.Security.ConfidentialLedger/src/Generated/ConfidentialLedgerClient.cs
archerzz: done.

sdk/confidentialledger/Azure.Security.ConfidentialLedger/src/Generated/ConfidentialLedgerClient.cs
archerzz: done.

sdk/confidentialledger/Azure.Security.ConfidentialLedger/src/Generated/ConfidentialLedgerClient.cs
archerzz: done.

sdk/confidentialledger/Azure.Security.ConfidentialLedger/src/Generated/ConfidentialLedgerClient.cs
archerzz: done

sdk/confidentialledger/Azure.Security.ConfidentialLedger/src/Generated/ConfidentialLedgerClient.cs
archerzz: @annelo-msft Previously the proposal is as follows:
> Use `<outer>-<inner>-<propertyName>` to populate string parameter values.

In the [experimental example doc](https://review.docs.microsoft.com/en-us/dotnet/api/azure.analytics.purview.scanning.purviewscanclient.createorupdate?view=azure-dotnet-preview&branch=ShivangiTestBranchWithoutFiles#examples), the value is composed in the format `test-<propertyName>-<propertyName>xxxx`, like `test-scanResults-parentId`.

During the coding, I found that `Schema.DefaultValue` can be used to as mock values as well. So in current implementation, I'm using the `Schema.DefaultValue`.

sdk/confidentialledger/Azure.Security.ConfidentialLedger/src/Generated/ConfidentialLedgerClient.cs
archerzz: I've split the cases into finer granularity. Now:
- If there are both body parameters and non-body parameters, then we return `parameters and request content`.
- If there are only non-body parameters, then we return `parameters`.
- If there are only body parameters, then we return `request content`.

sdk/confidentialledger/Azure.Security.ConfidentialLedger/src/Generated/ConfidentialLedgerClient.cs
```diff
@@ -639,6 +883,28 @@ public virtual Response GetTransactionStatus(string transactionId, RequestContex
         /// <summary> A sub-ledger id may optionally be specified. </summary>
         /// <param name="subLedgerId"> The sub-ledger id. </param>
         /// <param name="context"> The request context, which can override default behaviors on the request on a per-call basis. </param>
+        /// <example>
+        /// This sample shows how to call GetCurrentLedgerEntryAsync with required parameters and properties.
```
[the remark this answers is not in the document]
archerzz: I fixed the issue, and now only there is required parameter or request content, we will show the corresponding example.

sdk/confidentialledger/Azure.Security.ConfidentialLedger/src/Generated/ConfidentialLedgerClient.cs
archerzz: This is an enum type, see the [swagger definition](https://github.com/Azure/azure-rest-api-specs/blob/f4d9fd5c6a0b8bb34a881728d2e8db28069f297a/specification/confidentialledger/data-plane/Microsoft.ConfidentialLedger/stable/2022-05-13/confidentialledger.json#L663-L675). Here we choose the first value, e.g., `Administrator`.

sdk/webpubsub/Azure.Messaging.WebPubSub/src/Generated/HealthApiClient.cs
archerzz: I think the no input content case should have been handled, see those `GetXXX` operations. For no response body case, I suggest that we should print the `response.Status`, e.g., `Console.WriteLine(response.Status);`

sdk/webpubsub/Azure.Messaging.WebPubSub/src/Generated/WebPubSubServiceClient.cs
```diff
@@ -38,6 +38,26 @@ protected WebPubSubServiceClient()
...
         /// <param name="context"> The request context, which can override default behaviors on the request on a per-call basis. </param>
+        /// <example>
+        /// This sample shows how to call GenerateClientTokenImplAsync with required parameters and properties.
+        /// <code>
+        /// var credential = new DefaultAzureCredential();
+        /// var endpoint = new Uri("https://my-account-name.azure.com");
+        /// var client = new WebPubSubServiceClient(endpoint, credential);
+        /// Response response = await client.GenerateClientTokenImplAsync();
+        /// JsonElement result = JsonDocument.Parse(GetContentFromResponse(response)).RootElement;
+        /// Console.WriteLine(result.ToString());
```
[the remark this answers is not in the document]
archerzz: This is because there is no required property defined in the response, and we're generating an example which only handles required properties.
In original design, our logic is to generate two types of examples
- Compose the required parameter/request content, and only parse required properties from response.
- Compose all parameter/request, and parse all properties from response.


For this specific case, it can be solved by the fix that we won't generate the required property case if no required parameter/request content is defined.

sdk/webpubsub/Azure.Messaging.WebPubSub/src/Generated/WebPubSubServiceClient.cs
```diff
@@ -38,6 +38,26 @@ protected WebPubSubServiceClient()
         /// <param name="role"> Roles that the connection with the generated token will have. </param>
         /// <param name="minutesToExpire"> The expire time of the generated token. </param>
         /// <param name="context"> The request context, which can override default behaviors on the request on a per-call basis. </param>
+        /// <example>
+        /// This sample shows how to call GenerateClientTokenImplAsync with required parameters and properties.
```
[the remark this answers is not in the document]
archerzz: This example will be removed since there is no required parameter defined. We'll only generate the example for all paramter/request content, which will also solve the issue above.

sdk/webpubsub/Azure.Messaging.WebPubSub/src/Generated/WebPubSubServiceClient.cs
archerzz: Previously I used the `DefaultValue`. I've improved the logic to not use `null ` default values.

archerzz:
List of changes so far:
- Support mocking of more types of parameters, instead of `null` value
- Add `Console.WriteLine(response.Status)` when response body is not defined
- Only generate examples for required parameters/request content if they are available
- Tune the sentence:
  - Replace `properties` with `request content`
  - Finer granularity control over words:
    - Print `parameters` when there is parameter to demonstrate
    - Print `request content` when there is request content to demonstrate
    - Print `parse the result` when there is response body to parse
- add new line after client initialization

sdk/confidentialledger/Azure.Security.ConfidentialLedger/src/Generated/ConfidentialLedgerClient.cs
```diff
@@ -453,6 +651,25 @@ public virtual Response GetLedgerEntry(string transactionId, string subLedgerId
         /// <param name="context"> The request context, which can override default behaviors on the request on a per-call basis. </param>
         /// <exception cref="ArgumentNullException"> <paramref name="transactionId"/> is null. </exception>
         /// <exception cref="ArgumentException"> <paramref name="transactionId"/> is an empty string, and was expected to be non-empty. </exception>
+        /// <example>
+        /// This sample shows how to call GetReceiptAsync and parse the result.
```
annelo-msft, line 655: nit/curious: should this say "with required parameters", since we are passing "transactionId" to GetReceiptAsync?  If not, I'd like to understand why since we may need to tune the wording a bit.

sdk/webpubsub/Azure.Messaging.WebPubSub/src/Generated/HealthApiClient.cs
annelo-msft: It looks like that is the right thing for [this API](https://docs.microsoft.com/en-us/rest/api/webpubsub/dataplane/health-api/get-service-status#response), but I wonder if it is correct for all cases.  Let's use this for now and keep an eye out if we find examples in Azure that this doesn't work for.  We can always update generated documentation after GA.

sdk/webpubsub/Azure.Messaging.WebPubSub/src/Generated/WebPubSubServiceClient.cs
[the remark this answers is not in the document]
annelo-msft: > For this specific case, it can be solved by the fix that we won't generate the required property case if no required parameter/request content is defined.

Sounds good, thanks!

sdk/confidentialledger/Azure.Security.ConfidentialLedger/src/Generated/ConfidentialLedgerClient.cs
annelo-msft: Where does Schema.DefaultValue come from?  Is this specified in the swagger or elsewhere?

archerzz: I think it's from M4. I'll investigate later.

archerzz:
List of changes so far:
- exclude read-only properties when composing request body, that normally happens when a schema is used in both request and response
- fix enum value in parameters, replace underlying value (e.g., `1, 2, 3` or `"foo", "bar"`) with enum value (e.g., `Category.A, Category.B`)
- fix binary data in request and response
- add all impacted SDKs after excluding HLC protocol methods

sdk/confidentialledger/Azure.Security.ConfidentialLedger/src/Generated/ConfidentialLedgerClient.cs
archerzz: I've put the conclusion in the comments below.

archerzz:
As for `DefaultValue` of each property, it's from M4 (see [here](https://github.com/Azure/autorest.csharp/blob/15c1e6cee6998c3c0d3088231804e007f5fa3797/src/AutoRest.CSharp/Common/Input/Generated/CodeModel.cs#L61-L63)). I checked the example codes, and it seems **only string properties** have value in `DefaultValue` property.

For string properties, the format is `{ParentSchemaName}{PropertyName}`, very similar with what we proposed(`{outer}-{inner}-{propertyName}`) in [original design](https://github.com/Azure/autorest.csharp/issues/2286), e.g., `CropsName` v.s. `crops-name` (note the in the [corresponding example doc](https://review.docs.microsoft.com/en-us/dotnet/api/azure.analytics.purview.scanning.purviewscanclient.createorupdate?view=azure-dotnet-preview&branch=ShivangiTestBranchWithoutFiles#examples) the value doesn't strictly follow the original design, its value is `test-crops-name` of which the difference is trivial).

Besides, the `DefaultValue` should come from [swagger`default` value](https://swagger.io/blog/api-development/unlocking-the-spec-the-default-keyword/#:~:text=In%20the%20Swagger%20specification%2C%20default%20allows%20you%20to,keyword%20is%20slightly%20unique%20in%20the%20Swagger%20specification.). So, it's still a good candidate for mocking value of all types of properties, though in reality maybe only string properties have value.
> In the Swagger specification, default allows you to set a value to parameters, models (and their properties) or response headers (the container). When those are omitted by either the client or the server, the receiving end should assume the default value.

Last but not least, we could leverage `example` property to mock the input values. See an [example here](https://github.com/Azure/azure-rest-api-specs/blob/683e3f4849ee1d84629d0d0fa17789e80a9cee08/specification/agfood/data-plane/Microsoft.AgFoodPlatform/preview/2021-03-31-preview/agfood.json#L8938-L8941). 
```
        "properties": {
          "description": "A collection of key value pairs that belongs to the resource.\r\nEach pair must not have a key greater than 50 characters\r\nand must not have a value greater than 150 characters.\r\nNote: A maximum of 25 key value pairs can be provided for a resource and only string and numeral values are supported.",
          "type": "object",
          "additionalProperties": {
            "type": "object"
          },
          "example": {
            "key1": "value1",
            "key2": 123.45
          }
        }
```

My suggestion is that we could use the following sequence to mock input value:
- `example`
- `default`
- hard code

sdk/confidentialledger/Azure.Security.ConfidentialLedger/src/Generated/ConfidentialLedgerClient.cs
archerzz: Done. Add corresponding description when there are only required parameter/request content.

sdk/purview/Azure.Analytics.Purview.Scanning/src/Generated/PurviewClassificationRuleClient.cs
```diff
@@ -71,6 +71,20 @@ public PurviewClassificationRuleClient(Uri endpoint, string classificationRuleNa
 
         /// <summary> Get a classification rule. </summary>
         /// <param name="context"> The request context, which can override default behaviors on the request on a per-call basis. </param>
+        /// <example>
+        /// This sample shows how to call GetPropertiesAsync and parse the result.
+        /// <code><![CDATA[
+        /// var credential = new DefaultAzureCredential();
+        /// var endpoint = new Uri("<https://my-account-name.azure.com>");
+        /// var client = new PurviewClassificationRuleClient(endpoint, credential);
```
annelo-msft, line 79: [ClassificationRuleClient is a subclient](https://apiview.dev/Assemblies/Review/705ac828d1604e8881431011883e2f58#Azure.Analytics.Purview.Scanning.PurviewScanningServiceClientOptions) and should be created using

`new PurviewScanningServiceClient(endpoint, credential).GetClassificationRuleClient("<classification-rule-name>");`

sdk/purview/Azure.Analytics.Purview.Scanning/src/Generated/PurviewDataSourceClient.cs
```diff
@@ -72,6 +72,52 @@ public PurviewDataSourceClient(Uri endpoint, string dataSourceName, TokenCredent
         /// <summary> Creates or Updates a data source. </summary>
         /// <param name="content"> The content to send as the body of the request. </param>
         /// <param name="context"> The request context, which can override default behaviors on the request on a per-call basis. </param>
+        /// <example>
+        /// This sample shows how to call CreateOrUpdateAsync with required request content and parse the result.
+        /// <code><![CDATA[
+        /// var credential = new DefaultAzureCredential();
+        /// var endpoint = new Uri("<https://my-account-name.azure.com>");
+        /// var client = new PurviewDataSourceClient(endpoint, credential);
```
annelo-msft, line 80: This is also a subclient.

sdk/purview/Azure.Analytics.Purview.Scanning/src/Generated/PurviewScanningServiceClient.cs
```diff
@@ -67,6 +67,21 @@ public PurviewScanningServiceClient(Uri endpoint, TokenCredential credential, Pu
         /// <param name="context"> The request context, which can override default behaviors on the request on a per-call basis. </param>
         /// <exception cref="ArgumentNullException"> <paramref name="keyVaultName"/> is null. </exception>
         /// <exception cref="ArgumentException"> <paramref name="keyVaultName"/> is an empty string, and was expected to be non-empty. </exception>
+        /// <example>
+        /// This sample shows how to call GetKeyVaultReferenceAsync with required parameters and parse the result.
+        /// <code><![CDATA[
+        /// var credential = new DefaultAzureCredential();
+        /// var endpoint = new Uri("<https://my-account-name.azure.com>");
+        /// var client = new PurviewScanningServiceClient(endpoint, credential);
```
annelo-msft, line 75: 👍  This is the root client.

sdk/purview/Azure.Analytics.Purview.Catalog/src/Generated/PurviewCatalogClient.cs
```diff
@@ -64,6 +64,68 @@ public PurviewCatalogClient(Uri endpoint, TokenCredential credential, PurviewCat
...
+        ///         },
+        ///         facet = new {
+        ///             count = 1234,
+        ///             facet = "<SearchFacetItemFacet>",
+        ///             sort = new {},
+        ///         },
+        ///     },
+        /// };
+        /// 
+        /// Response response = await client.SearchAsync(RequestContent.Create(data));
```
annelo-msft, line 110: Would it be possible to add a carriage return after the client call to separate this from parsing the response?

sdk/purview/Azure.Analytics.Purview.Scanning/src/Generated/PurviewClassificationRuleClient.cs
```diff
@@ -360,6 +452,27 @@ public virtual Response Delete(RequestContext context = null)
...
+        /// 
+        /// Response response = await client.TagVersionAsync(1234, "<action>");
+        /// JsonElement result = JsonDocument.Parse(GetContentFromResponse(response)).RootElement;
+        /// Console.WriteLine(result.GetProperty("scanResultId").ToString());
+        /// Console.WriteLine(result.GetProperty("startTime").ToString());
+        /// Console.WriteLine(result.GetProperty("endTime").ToString());
+        /// Console.WriteLine(result.GetProperty("status").ToString());
+        /// Console.WriteLine(result.GetProperty("error").GetProperty("code").ToString());
```
annelo-msft, line 468: Since we are no longer showing the error response schema in the `<remarks>` block, we should remove parsing error from the examples as well.  In C#, customers get error information via RequestFailedException.

sdk/purview/Azure.Analytics.Purview.Scanning/src/Generated/PurviewClassificationRuleClient.cs
```diff
@@ -71,6 +71,20 @@ public PurviewClassificationRuleClient(Uri endpoint, string classificationRuleNa
 
         /// <summary> Get a classification rule. </summary>
         /// <param name="context"> The request context, which can override default behaviors on the request on a per-call basis. </param>
+        /// <example>
```
annelo-msft, line 74: It looks like this one only has one example, not one for required and one for all properties.  Can you help remind me why?

archerzz: Because in the [original swagger spec](https://github.com/Azure/azure-rest-api-specs/blob/1c7df99f6a84335cfd7bf5be8c800d72c1dddbc2/specification/purview/data-plane/Azure.Analytics.Purview.Scanning/preview/2018-12-01-preview/scanningService.json#L197-L236), this method has one required parameter only, so here we just generate one example. And in this example, we parse all properties of the [response](https://github.com/Azure/azure-rest-api-specs/blob/1c7df99f6a84335cfd7bf5be8c800d72c1dddbc2/specification/purview/data-plane/Azure.Analytics.Purview.Scanning/preview/2018-12-01-preview/scanningService.json#L1926-L1950).

sdk/purview/Azure.Analytics.Purview.Scanning/src/Generated/PurviewClassificationRuleClient.cs
archerzz: This is the `OperationResponse.error` property (see [definition](https://github.com/Azure/azure-rest-api-specs/blob/1c7df99f6a84335cfd7bf5be8c800d72c1dddbc2/specification/purview/data-plane/Azure.Analytics.Purview.Scanning/preview/2018-12-01-preview/scanningService.json#L2019-L2023)) of an LRO [`ClassificationRules_TagClassificationVersion`](https://github.com/Azure/azure-rest-api-specs/blob/1c7df99f6a84335cfd7bf5be8c800d72c1dddbc2/specification/purview/data-plane/Azure.Analytics.Purview.Scanning/preview/2018-12-01-preview/scanningService.json#L427).
It's part of a LRO response, though the operation does define a separate [error response](https://github.com/Azure/azure-rest-api-specs/blob/1c7df99f6a84335cfd7bf5be8c800d72c1dddbc2/specification/purview/data-plane/Azure.Analytics.Purview.Scanning/preview/2018-12-01-preview/scanningService.json#L470-L474).

So, this looks like an issue in the original swagger definition. I'm not sure if we should fix that.

sdk/purview/Azure.Analytics.Purview.Scanning/src/Generated/PurviewScanClient.cs
```diff
@@ -76,6 +76,21 @@ public PurviewScanClient(Uri endpoint, string dataSourceName, string scanName, T
 
         /// <summary> Get a filter. </summary>
         /// <param name="context"> The request context, which can override default behaviors on the request on a per-call basis. </param>
+        /// <example>
+        /// This sample shows how to call GetFilterAsync and parse the result.
+        /// <code><![CDATA[
+        /// var credential = new DefaultAzureCredential();
+        /// var endpoint = new Uri("<https://my-account-name.azure.com>");
+        /// var client = new PurviewScanClient(endpoint, credential);
```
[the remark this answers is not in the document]
archerzz: The root cause is that in the original codes, the sub-client is created using custom codes, not the recommended way (e.g., `CodeGenClient` attribute) in the doc. So, the code model doesn't contain metadata of sub-clients and corresponding factory methods. That's why the initialization codes are wrong here (and below).

I've fixed by using the `CodeGenClient` attribute. The side effect is that `PurviewScanClient.cs` is renamed to `PurviewScan`. In the `PureviewScan`, the example codes have the correct initialization logic.

sdk/purview/Azure.Analytics.Purview.Scanning/src/Generated/PurviewClassificationRuleClient.cs
archerzz: fixed by using `CodeGenClient` attribute, the side effect is that `PurviewClassificiationRuleClient` to `PurviewClassificiationRule`.

sdk/purview/Azure.Analytics.Purview.Scanning/src/Generated/PurviewDataSourceClient.cs
archerzz: fixed by using `CodeGenClient` attribute, the side effect is that `PurviewDataSourceClient` to `PurviewDataSource`.

sdk/purview/Azure.Analytics.Purview.Catalog/src/Generated/PurviewCatalogClient.cs
annelo-msft: Thanks! :)

sdk/purview/Azure.Analytics.Purview.Scanning/src/Generated/PurviewClassificationRuleClient.cs
annelo-msft: Please file an issue to track resolving this in v1.1 or v1.2.  I don't think we need to block GA on this.

sdk/purview/Azure.Analytics.Purview.Scanning/src/Generated/PurviewClassificationRuleClient.cs
annelo-msft: Thank you!

sdk/purview/Azure.Analytics.Purview.Scanning/src/Generated/PurviewScanClient.cs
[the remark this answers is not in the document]
annelo-msft: `PurviewScan` is Resource Client, so, according to the [guidelines for subclients](https://azure.github.io/azure-sdk/dotnet_introduction.html#dotnet-subclient-factory-methods-suffix), it is correct that it does not have the `Client` suffix.

sdk/purview/Azure.Analytics.Purview.Scanning/src/Generated/PurviewScan.cs
```diff
@@ -33,49 +33,48 @@ public partial class PurviewScanClient
...
         }
 
         /// <summary> Get a filter. </summary>
         /// <param name="context"> The request context, which can override default behaviors on the request on a per-call basis. </param>
+        /// <example>
+        /// This sample shows how to call GetFilterAsync and parse the result.
+        /// <code><![CDATA[
+        /// var credential = new DefaultAzureCredential();
+        /// var endpoint = new Uri("<https://my-account-name.azure.com>");
+        /// var client = new PurviewScanningServiceClient(endpoint, credential).GetPurviewDataSourceClient("<dataSourceName>").GetPurviewScanClient("<scanName>");
```
annelo-msft, line 67: There are no methods named `GetPurviewDataSourceClient` or `GetPurviewScanClient`.  

Also, the data source client and scan clients are resource subclients, so the factory method should not have the `Client` suffix [according to the guidelines](https://azure.github.io/azure-sdk/dotnet_introduction.html#dotnet-subclient-factory-methods-suffix).

sdk/purview/Azure.Analytics.Purview.Scanning/src/Generated/PurviewClassificationRuleClient.cs
Pablo1q: Whats happens?

sdk/purview/Azure.Analytics.Purview.Scanning/src/Generated/PurviewScan.cs
ShivangiReja: >Would it be possible for you to confirm that the code compiles as part of your implementation?

For this, can we add these examples in a `.cs` file(maybe in `Generated/Examples` folder) and pull it into the docs from there(maybe use it as snippets)? If that's possible, then we can make sure that these examples compile.

sdk/confidentialledger/Azure.Security.ConfidentialLedger/src/Generated/ConfidentialLedgerClient.cs
```diff
@@ -967,6 +1381,35 @@ public virtual Response CreateOrUpdateUser(string userId, RequestContent content
...
+        /// Console.WriteLine(result.GetProperty("state").ToString());
+        /// Console.WriteLine(result.GetProperty("entries").Item[0].GetProperty("contents").ToString());
+        /// ]]></code>
+        /// This sample shows how to call GetLedgerEntriesAsync with all parameters, and how to parse the result.
+        /// <code><![CDATA[
+        /// var credential = new DefaultAzureCredential();
+        /// var endpoint = new Uri("<https://my-account-name.azure.com>");
+        /// var client = new ConfidentialLedgerClient(endpoint, credential);
+        /// 
+        /// Response response = await client.GetLedgerEntriesAsync("<subLedgerId>", "<fromTransactionId>", "<toTransactionId>");
```
annelo-msft, line 1403: `GetLedgerEntries` is a pageable operation.  The example should show how to use the `Pageable<BinaryData>` return value and parse the result from the `BinaryData` item.

sdk/purview/Azure.Analytics.Purview.Scanning/src/Generated/PurviewClassificationRuleClient.cs
archerzz: Filed an issue: https://github.com/Azure/autorest.csharp/issues/2345

sdk/purview/Azure.Analytics.Purview.Scanning/src/Generated/PurviewScanClient.cs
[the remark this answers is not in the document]
archerzz: @annelo-msft Actually `PurviewScan` is **not** a Resource Client. See the [swagger definition](https://github.com/Azure/azure-rest-api-specs/blob/1c7df99f6a84335cfd7bf5be8c800d72c1dddbc2/specification/purview/data-plane/Azure.Analytics.Purview.Scanning/preview/2018-12-01-preview/scanningService.json#L774). It doesn't have a resource identifier parameter, which is the condition of a Resource Client, according to the generator code [IsResourceClient](https://github.com/archerzz/autorest.csharp/blob/00d2093b0d2d594fc23fdab7d63f1fb8f31cf81a/src/AutoRest.CSharp/LowLevel/Output/LowLevelClient.cs#L64).
In the test project `ResourceClients-LowLevel` of `autorest.csharp`, there is [a resource identifier parameter in the swagger](https://github.com/archerzz/autorest.csharp/blob/78e59af2f193d7b445a21e054702a9416fafa341/test/TestProjects/ResourceClients-LowLevel/ResourceClients-LowLevel.json#L138), so the corresponding client is considered as a Resource Client.

The reason why it doesn't have `Client` suffix is because it's a SubClient. A SubClient doesn't have `Client` suffix, according to the generator code [here](https://github.com/archerzz/autorest.csharp/blob/b20e5ee38dce9d743b5c3390118433e2499ff4c1/src/AutoRest.CSharp/LowLevel/Output/LowLevelOutputLibraryFactory.cs#L88).

sdk/confidentialledger/Azure.Security.ConfidentialLedger/src/Generated/ConfidentialLedgerClient.cs
archerzz: Fixed. Adding support for pageable operation

archerzz:
changes since last update:

- Add support for pageable and LRO
- Fix wrong `WaitForCompletion` method name
- Replace `GetContentFromResponse(response)` with `response.ContentStream`, so that the code can compile

sdk/purview/Azure.Analytics.Purview.Scanning/src/Generated/PurviewClassificationRuleClient.cs
archerzz: @Pablo1q just some name changes on the generated classes. Since the SDK is not GA, it's OK for breaking changes.

archerzz:
changes since last update:
- fix incorrect client constructor invocation

archerzz:
changes since last update:
- fix: some array properties are not parsed in response
- fix incorrect pagination response parsing

archerzz:
changes since last update:

- incorporate other comments improvement
- fix incorrect handling of inline property when composing request
- fix incorrect formatting if the child property is null, when composing request

archerzz:
updated preview doc sites:
- [Azure.Analytics.Purview.Scanning](https://raw.githack.com/archerzz/azure-sdk-for-net/dpg/docs/example-comment/doc/ApiDocGeneration/Azure.Analytics.Purview.Scanning/docfx-output/docfx_project/_site/index.html)
- [Azure.IoT.DeviceUpdate](https://raw.githack.com/archerzz/azure-sdk-for-net/dpg/docs/example-comment/doc/ApiDocGeneration/Azure.IoT.DeviceUpdate/docfx-output/docfx_project/_site/index.html)
- [Azure.Messaging.WebPubSub](https://raw.githack.com/archerzz/azure-sdk-for-net/dpg/docs/example-comment/doc/ApiDocGeneration/Azure.Messaging.WebPubSub/docfx-output/docfx_project/_site/index.html)
- [Azure.Security.ConfidentialLedger](https://raw.githack.com/archerzz/azure-sdk-for-net/dpg/docs/example-comment/doc/ApiDocGeneration/Azure.Security.ConfidentialLedger/docfx-output/docfx_project/_site/index.html)

archerzz:
closing since the change has been merged in https://github.com/Azure/azure-sdk-for-net/pull/30025

PR #29244, opened 2022-06-10
[AzureMonitorExporter] move internal types to subdirectory and normalize nampspaces

TimothyMothra:
This PR does some housekeeping.
Previously all our types were in the root of the Exporter project. 
This was fine when the project was small, but it's getting unwieldly to navigate.

<details><summary>Screenshot Before</summary>

image

</details>

<details><summary>Screenshot After</summary>

image

image

</details>

## Changes
- move all `internal` types to a new subdirectory `Internals`
  - This complies with [Azure SDK Guidelines](https://azure.github.io/azure-sdk/dotnet_introduction.html#:~:text=YOU%20SHOULD%20NOT%20have%20many%20types%20in%20the%20main%20namespace.): 
    > YOU SHOULD NOT have many types in the main namespace. Number of types is directly proportional to the perceived complexity of a library.
  - "Internals" was chosen to hold all `internal` types.
    - Azure.Core uses the name "Internal" [example1](https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/core/Azure.Core/src/Internal) [example2](https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/core/Azure.Core/src/Pipeline/Internal)
    - [.NET Guidelines](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-namespaces#:~:text=CONSIDER%20using%20plural%20namespace%20names%20where%20appropriate) recommends plural namespaces:
       > CONSIDER using plural namespace names where appropriate.

- normalize namespaces across all projects.
  - removed default namespace from Test project.
     This was causing new classes in the Test project to not use the assembly name, requiring manual effort to change.
- Intellisense applied several other misc changes.


sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Azure.Monitor.OpenTelemetry.Exporter.csproj
```diff
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
TimothyMothra, line 1: IntelliSense found a trailing space

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/AzMonList.cs
```diff
@@ -81,7 +81,7 @@ internal static object[] GetTagValues(ref AzMonList list, params string[] tagNam
         {
             int lengthTagNames = tagNames.Length;
             int lengthList = list.Length;
-            object[] values = new object[(int)lengthTagNames];
+            object[] values = new object[lengthTagNames];
```
TimothyMothra, line 84: `lengthTagNames` is already an int and doesn't need the cast. We should invest in an Analyzer to catch things like this.

vishweshbankwar (approved):

PR #29246, opened 2022-06-10
Add the ability to record / playback multiple versions

m-nash:
To support AzureStack we want to be able to certify the SDK with multiple versions without needing to rewrite test cases.  There was already a mechanism that allowed for multiple versions in live mode [here ](https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/core/Azure.Core.TestFramework#support-multi-service-version-testing).  This PR extends that functionality in a few ways.

1. Allows for the version to be either an Enum convertible to Int (previous behavior) OR a string (new behavior)
2. Adds one new parameter to the `ClientTestFixture` constructor to indicate if you want the versions to apply to record / playback (default being false which maintains previous behavior)
3. Adds a new qualifier to the filename to differentiate between which recording belongs to which version


sdk/core/Azure.Core.TestFramework/src/ClientTestFixtureAttribute.cs
```diff
@@ -40,15 +42,23 @@ public class ClientTestFixtureAttribute : NUnitAttribute, IFixtureBuilder2, IPre
...
+        public ClientTestFixtureAttribute(bool recordMultipleVersions, params object[] serviceVersions) : this(serviceVersions: serviceVersions, default, recordMultipleVersions)
+        { }
+
         /// <summary>
         /// Initializes an instance of the <see cref="ClientTestFixtureAttribute"/> accepting additional fixture parameters.
         /// </summary>
         /// <param name="serviceVersions">The set of service versions that will be passed to the test suite.</param>
         /// <param name="additionalParameters">An array of additional parameters that will be passed to the test suite.</param>
-        public ClientTestFixtureAttribute(object[] serviceVersions, object[] additionalParameters)
+        public ClientTestFixtureAttribute(object[] serviceVersions, object[] additionalParameters, bool recordMultipleVersions = false)
```
JoshLove-msft, line 57: nit: we chatted about something like `recordAllVersions`

JoshLove-msft:
Overall approach looks good. Added a couple comments for things we discussed offline.
Here are the list of the test classes/patterns that can be moved to a new Azure.Core.TestFramework.Tests project that @m-nash offered to create:
- ClientTestBase*
- RecordedTestBaseTests
- TestEnvironmentTests
- ManagementRecordedTestBaseTests
- IgnoreServiceErrorAttributeTests

sdk/core/Azure.Core.TestFramework/src/RecordedTestBase.cs
```diff
@@ -177,9 +179,10 @@ public string ReplacementHost
         {
         };
 
-        protected RecordedTestBase(bool isAsync, RecordedTestMode? mode = null) : base(isAsync)
+        protected RecordedTestBase(bool isAsync, RecordedTestMode? mode = null, string apiVersion = null) : base(isAsync)
```
JoshLove-msft, line 182: nit: call the param versionQualifier?

sdk/core/Azure.Core.TestFramework/src/ClientTestFixtureAttribute.cs
```diff
@@ -15,13 +16,49 @@ namespace Azure.Core.TestFramework
 {
     public class ClientTestFixtureAttribute : NUnitAttribute, IFixtureBuilder2, IPreFilter
     {
+        private readonly struct TestVersion : IComparable
+        {
+            public IComparable Comparable { get; }
```
JoshLove-msft, line 21: Do we need this property? Can't we just store the converted data in `Object` and use CompareTo on that.

sdk/core/Azure.Core.TestFramework/src/ClientTestFixtureAttribute.cs
```diff
@@ -37,18 +74,32 @@ public class ClientTestFixtureAttribute : NUnitAttribute, IFixtureBuilder2, IPre
...
+        public ClientTestFixtureAttribute(bool recordAllVersions, params object[] serviceVersions) : this(serviceVersions: serviceVersions, additionalParameters: default, recordAllVersions)
         { }
 
         /// <summary>
         /// Initializes an instance of the <see cref="ClientTestFixtureAttribute"/> accepting additional fixture parameters.
         /// </summary>
         /// <param name="serviceVersions">The set of service versions that will be passed to the test suite.</param>
         /// <param name="additionalParameters">An array of additional parameters that will be passed to the test suite.</param>
-        public ClientTestFixtureAttribute(object[] serviceVersions, object[] additionalParameters)
+        public ClientTestFixtureAttribute(object[] serviceVersions, object[] additionalParameters, bool recordAllVersions = false)
```
JoshLove-msft, line 92: nit: add param docs

sdk/core/Azure.Core.TestFramework/src/RecordedTestBase.cs
JoshLove-msft, line 182: We should also document this constructor and parameters explaining the usage and that it is only currently used for mgmt plane.

sdk/core/Azure.Core.TestFramework/src/VersionString.cs
```diff
@@ -0,0 +1,81 @@
...
+        public VersionString(string version)
+        {
+            Argument.AssertNotNull(version, nameof(version));
+
+            RawVersion = version;
+            ReadOnlySpan<char> chars = version.AsSpan();
+            IsGa = true;
+            int index = -1;
+            PreviewString = null;
+            if (chars.Length > 10)
```
JoshLove-msft, line 24: Why 10?

sdk/core/Azure.Core.TestFramework/src/VersionString.cs
```diff
@@ -0,0 +1,81 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+
+namespace Azure.Core.TestFramework
+{
+    internal readonly struct VersionString : IComparable
```
JoshLove-msft, line 8: Is this still needed?

sdk/core/Azure.Core.TestFramework/src/VersionString.cs
christothes: I think that is the standard length for a version string minus any preview suffixes. `nnnn-nn-nn`

sdk/core/Azure.Core.TestFramework/src/RecordedTestBase.cs
christothes: Should we also consider creating a separate ctor overload that takes this argument, just to make it less visible to data plane test classes?

sdk/core/Azure.Core.TestFramework/src/ClientTestFixtureAttribute.cs
```diff
@@ -37,18 +74,32 @@ public class ClientTestFixtureAttribute : NUnitAttribute, IFixtureBuilder2, IPre
...
         {
             _additionalParameters = additionalParameters ?? new object[] { };
-            _serviceVersions = serviceVersions ?? new object[] { };
+            _serviceVersions = serviceVersions ?? Array.Empty<object>();
+            if (_serviceVersions.Length > 0)
+            {
+                var first = _serviceVersions[0];
+                if (!_serviceVersions.All(x => x.GetType() == first.GetType()))
+                    throw new InvalidOperationException("All service versions must be the same type");
```
christothes, line 100: should we use the same string as above? 
"The items in the serviceVersions array must be an enum convertable to Int32 or a string in date format with an optional preview string"

sdk/core/Azure.Core.TestFramework/src/VersionString.cs
JoshLove-msft: Yes, that is what I assumed, but I am hinting at the need for a constant or a comment.

sdk/core/Azure.Core.TestFramework/src/ClientTestFixtureAttribute.cs
m-nash: Object is actually the only thing needed outside this struct on line 169, but I can make this a private readonly field

sdk/core/Azure.Core.TestFramework/src/RecordedTestBase.cs
m-nash: This is not intended to be a mgmt plane feature, it allows you record / playback specific versions where as before the only functionality was to be able to run live tests on multiple versions.  Which SDKs opt in or out today shouldn't sway our design since that will change over time.

sdk/core/Azure.Core.TestFramework/src/VersionString.cs
m-nash: I need somewhere to encaspulate the parsing / comparison of a version string

sdk/core/Azure.Core.TestFramework/src/ClientTestFixtureAttribute.cs
m-nash: This error is specifically testing for homogenousness, but I can add an additional part which specifies the valid types

sdk/core/Azure.Core.TestFramework/src/VersionString.cs
JoshLove-msft: Can we move it to where it is used as part of TestVersion?

m-nash: Its cleaner to separate this as its own thing.  There is also potential here to share this logic since this isn't the only place we do this, but I am punting on that for now I have a follow up I will track

sdk/core/Azure.Core.TestFramework/src/ClientTestFixtureAttribute.cs
```diff
@@ -18,18 +18,18 @@ public class ClientTestFixtureAttribute : NUnitAttribute, IFixtureBuilder2, IPre
     {
         private readonly struct TestVersion : IComparable
         {
-            public IComparable Comparable { get; }
+            private readonly IComparable _comparable;
```
JoshLove-msft, line 21: Still don't see why we need this at all, as compared to just storing in Object.

m-nash: Object is the original thing that was passed in which is needed on line 170 (used to be 169 in a previous comment but it moved).  _comparable is the "converted thing" that can be used to compare for finding the Min / Max which is of a completely different type.

If I don't keep both in this struct then I will need some sort of dictionary to look up the original object from the comparable after the Max method is called.

JoshLove-msft: Ah, got it.

sdk/core/Azure.Core.TestFramework/src/RecordedTestBase.cs
m-nash: This was resolved when we were fixing a consistency issue between two settings so resolving as no-op at this point.

JoshLove-msft (approved):
LGTM!

PR #29252, opened 2022-06-10
[AzureMonitorExporter] implementing Truncation rules for RemoteDependencyData and RequestData

TimothyMothra:
Continuation of #29094

## Changes
- implement truncation for `RemoteDependencyData` and `RequestData`
- update `SchemaConstants`
  - Duration must be less than 1000 days. 1 tick less than 1000 days is "999.23:59:59.9999999".
    I verified that Ingestion will accept this value (using Application Insights SDK) and added it to the constants.
- remove null check from Truncate() method.
   Some of these fields are optional and could be null.

### DIFF
--- a/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Customizations/Models/RequestData.cs
+++ b/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Customizations/Models/RequestData.cs
@@ -27,7 +27,7 @@ namespace Azure.Monitor.OpenTelemetry.Exporter.Models
                     break;
             }
 
-            Id = activity.Context.SpanId.ToHexString().Truncate(SchemaConstants.RequestData_Id_MaxLength);
+            Id = activity.Context.SpanId.ToHexString();
             Name = TraceHelper.GetOperationName(activity, ref monitorTags.MappedTags).Truncate(SchemaConstants.RequestData_Name_MaxLength);
             Duration = activity.Duration < SchemaConstants.RequestData_Duration_LessThanDays
                 ? activity.Duration.ToString("c", CultureInfo.InvariantCulture)
--- a/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Customizations/Models/RequestData.cs
+++ b/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Customizations/Models/RequestData.cs
@@ -27,12 +27,16 @@ namespace Azure.Monitor.OpenTelemetry.Exporter.Models
                     break;
             }
 
-            Id = activity.Context.SpanId.ToHexString();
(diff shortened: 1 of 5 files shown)


TimothyMothra:
Several tests are failing. I'll look into this on Monday.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Customizations/Models/RequestData.cs
```diff
@@ -27,12 +27,16 @@ public RequestData(int version, Activity activity, ref TagEnumerationState monit
                     break;
             }
 
-            Id = activity.Context.SpanId.ToHexString();
-            Name = TraceHelper.GetOperationName(activity, ref monitorTags.MappedTags);
-            Duration = activity.Duration.ToString("c", CultureInfo.InvariantCulture);
+            Id = activity.Context.SpanId.ToHexString().Truncate(SchemaConstants.RequestData_Id_MaxLength);
```
vishweshbankwar, line 30: Minor: We can skip this truncate. Activity `SpanId` and `TraceId` will be of fixed length.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Customizations/Models/RequestData.cs
```diff
@@ -27,12 +27,16 @@ public RequestData(int version, Activity activity, ref TagEnumerationState monit
...
+            Duration = activity.Duration < SchemaConstants.RequestData_Duration_LessThanDays
+                ? activity.Duration.ToString("c", CultureInfo.InvariantCulture)
+                : SchemaConstants.Duration_MaxValue;
             Success = activity.GetStatus().StatusCode != StatusCode.Error;
-            ResponseCode = AzMonList.GetTagValue(ref monitorTags.MappedTags, SemanticConventions.AttributeHttpStatusCode)?.ToString() ?? "0";
-            Url = url;
+            ResponseCode = AzMonList.GetTagValue(ref monitorTags.MappedTags, SemanticConventions.AttributeHttpStatusCode)
+                ?.ToString().Truncate(SchemaConstants.RequestData_ResponseCode_MaxLength)
+                ?? "0";
+            Url = url.Truncate(SchemaConstants.RequestData_Url_MaxLength);
```
rajkumar-rangaraj, line 39: What would happen if this truncation results in a mal-formed URL?

TimothyMothra: the max length is 2048.
I think it's rare that this would occur, but if it does we would ingest an incomplete url.

I don't see any special instructions in the Mapping doc to handle this differently.

rajkumar-rangaraj (approved):

vishweshbankwar (approved):

Issue #29212, opened 2022-06-09
[BUG][Storage mgmt Track2] RestoreBlobRanges' output TimeToRestore is always 1/1/0001 12:00:00 AM +00:00


yifanz0:
### Library name and version

Azure.ResourceManager.Storage 1.0.0-beta.9

### Describe the bug

In RestoreBlobRanges' output, no matter what the user input of TimeToRestore is, the output TimeToRestore's value is always 1/1/0001 12:00:00 AM +00:00. This happens no matter if we put WaitUtil.Started or WaitUtil.Completed when calling the method.

### Expected behavior

For RestoreBlobRanges, the TimeToRestore in the output BlobRestoreStatus object should be the same as the TimeToRestore value specified in BlobRestoreContent object when calling RestoreBlobRanges. Say the input is 1/1/2022 15:23:03 AM +00:00, then the output should be the same. 

### Actual behavior

RestoreBlobRanges' output TimeToRestore is 1/1/0001 12:00:00 AM +00:00, but the input is not. 

![image](https://user-images.githubusercontent.com/100746763/172806392-55ef1f22-f427-4777-bd87-06450deacf7e.png)

The response returned by the server seems to be correct and has the right time:
```
GET https://management.azure.com/subscriptions/***/providers/Microsoft.Storage/locations/eastus2euap/asyncoperations/97cea899-6954-400d-8ac0-5bbbc547ff10?monitor=true&api-version=2021-08-01 HTTP/1.1
CommandName: Restore-AzStorageBlobRange
ParameterSetName: AccountName
x-ms-client-request-id: ea2a7a0a-d3aa-4912-b4ac-ce1169dc6363
x-ms-return-client-request-id: true
Authorization: Bearer ***
Host: management.azure.com


HTTP/1.1 200 OK
Cache-Control: no-cache
Pragma: no-cache
Content-Length: 277
Content-Type: application/json
Expires: -1
x-ms-client-request-id: ea2a7a0a-d3aa-4912-b4ac-ce1169dc6363
x-ms-request-id: 0818c420-1797-4b42-af4e-c0fd775fc311
Strict-Transport-Security: max-age=31536000; includeSubDomains
x-ms-ratelimit-remaining-subscription-reads: 11990
Server: Microsoft-Azure-Storage-Resource-Provider/1.0,Microsoft-HTTPAPI/2.0 Microsoft-HTTPAPI/2.0
[...]

PR #29263, opened 2022-06-13
[Storage] Fix timeToRestore issue for storage

Yao725:
This one is a workaround for issue #29212, will be removed after the fix in swagger.


sdk/storage/Azure.ResourceManager.Storage/src/Customized/Models/BlobRestoreContent.Serialization.cs
```diff
@@ -0,0 +1,40 @@
...
+// Licensed under the MIT License.
+
+#nullable disable
+
+using System;
+using System.Collections.Generic;
+using System.Text.Json;
+using Azure.Core;
+
+namespace Azure.ResourceManager.Storage.Models
```
ArthurMa1978, line 11: please open another issue to track this customization which will be removed once the `2022-05-01` is merged

ArthurMa1978 (approved):

sdk/storage/Azure.ResourceManager.Storage/src/Customized/Models/BlobRestoreContent.Serialization.cs
Yao725: I use directive instead of customization to solve this problem and this directive has no side impact if this issue is fixed in swagger side.

Issue #28926, opened 2022-05-23
[BUG] EventHubBufferedProducerClient: high CPU usage when no events are being sent


andreminelli:
### Library name and version

Azure.Messaging.EventHubs 5.7.0

### Describe the bug

We have an application with high event producing rate (at least hundreads per second) and after a load test (peak of 25K events in 1 minute) with duration of 3 minutes, we observed that CPU usage keeps almost the same as the event 'processing'.

We compared the very same approach/code/test exchanging to an `EventHubProducerClient` (on same version) and CPU drops seconds after the 'injection' of events finishes. 
In both versions, each queue name has a unique instance of this class/Producer being (re)used to send all events.

### Expected behavior

After the load test has stopped - as the the production of events has stopped too - we expect the CPU usage drops to same levels before the test, after a few seconds.

### Actual behavior

CPU usage keeps at the almost same level after several minutes of the ending of load test. Only after a process restart the CPU usage has got back to normal levels.

### Reproduction Steps

The code is very straighforward (I have edited to remove some internal abstraction details, but nothing relevant):

* Using `EventHubBufferedProducerClient`:

```csharp
    public class SenderQueue<T> : IAsyncDisposable
    {
        private readonly ISerializer<T> _serializer;
        private readonly EventHubBufferedProducerClient _eventHubProducerClient;
        private readonly ILogger _logger;

        public SenderQueue(
            IConfiguration configuration,
            string queueName,
            ISerializer<T> serializer,
            ILogger logger)
        {
            _serializer = serializer;
            _logger = logger;
            _eventHubProducerClient = GetProducerClient(configuration, queueName);
        }

        public virtual async Task EnqueueAsync(T item, CancellationToken cancellationToken = default)
        {
            if (item == null)
                throw new ArgumentNullException(nameof(item));
           var serializedItem = _serializer.Serialize(item);
           await _eventHubProducerClient.EnqueueEventAsync(new EventData(serializedItem), cancellationToken);
        }

        public async ValueTask DisposeAsync()
        {
            _eventHubProducerClient.SendEventBatchSucceededAsync -= SendSuccessfulHandler;
            _eventHubProducerClient.SendEventBatchFailedAsync -= SendFailedHandler;
            await _eventHubProducerClient.DisposeAsync();
        }

        protected EventHubBufferedProducerClient GetProducerClient(IConfiguration configuration, string queueName)
        {
            var eventHubBufferedProducerClient = new EventHubBufferedProducerClient(
[...]

Issue #28840, opened 2022-05-17
Buffered Producer: Consider Blocking when No Events Available


jsquire:
# Summary

In order to avoid a tight loop that consumes CPU resources on the host, the `EventHubBufferedProducerClient` will introduce an expanding delay when no events are available to be published.   During longer periods of low activity, it may be beneficial to instead block on a synchronization primitive within the processing loop and allow the next call to enqueue events to signal it.

# Scope of Work

 - Consider the loop in `RunPublishingAsync`; in the case that no events are available in the buffers and `delay` reaches the maximum threshold, would it be advantageous to block on a synchronization primitive? (_[src](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Azure.Messaging.EventHubs/src/Producer/EventHubBufferedProducerClient.cs#L2213)_)

- If yes:
  - Determine which synchronization primitive can be safely used to block here and be woken by any enqueue call.
  - Determine how to handle the race condition where an event is enqueued after the check but before blocking on the primitive
  - Do the implementaiton

- If no:
  - Update this issue with rationale and close

# Success Criteria

- A determination has been made on whether a synchronization primitive should be introduced when the buffers are empty or not.

**-- EITHER --**
  - The implementation has been updated to make use of the synchronization primitive
  - The tests necessary for sample validation have been created or adjusted and pass reliably.
  - The existing test suite continues to produce deterministic results and pass reliably.

**-- OR --**
- This issue has been updated with the rationale for the decision and closed.

# References and Resources

- [Event Hubs: Buffered Producer Improvements (28508)](https://github.com/Azure/azure-sdk-for-net/pull/28508#discussion_r864087254) _[see discussion]_

PR #29276, opened 2022-06-13
[Event Hubs] Enter Idle State with Empty Buffers

jsquire:
# Summary

The focus of these changes is to fix an issue with the `EventHubBufferedProducerClient` where it failed to detect empty buffers and enter an idle state.  Instead, the background task that manages publishing would spin and consume an unreasonable amount of resources.

# References and Related

- [[BUG] EventHubBufferedProducerClient: high CPU usage when no events are being sent (#28926)](https://github.com/Azure/azure-sdk-for-net/issues/28926)
- [Buffered Producer: Consider Blocking when No Events Available (#28840)](https://github.com/Azure/azure-sdk-for-net/issues/28840)

### DIFF
--- a/sdk/eventhub/Azure.Messaging.EventHubs/src/Diagnostics/EventHubsEventSource.cs
+++ b/sdk/eventhub/Azure.Messaging.EventHubs/src/Diagnostics/EventHubsEventSource.cs
@@ -2631,7 +2631,7 @@ namespace Azure.Messaging.EventHubs.Diagnostics
         /// <param name="operationId">An artificial identifier for the idle operation.</param>
         /// <param name="durationSeconds">The total duration that the idle await took place, in seconds.</param>
         ///
-        [Event(127, Level = EventLevel.Verbose, Message = "Publishing for the buffered producer instance with identifier '{0}' for Event Hub: {1} has completed its idle state; one or more events was enqueued.  Operation Id: '{2},' Duration: '{3:0.00}' seconds.")]
+        [Event(127, Level = EventLevel.Verbose, Message = "Publishing for the buffered producer instance with identifier '{0}' for Event Hub: {1} has exited the idle state; one or more events was enqueued.  Operation Id: '{2},' Duration: '{3:0.00}' seconds.")]
         public virtual void BufferedProducerIdleComplete(string identifier,
                                                                  string eventHubName,
                                                                  string operationId,
--- a/sdk/eventhub/Azure.Messaging.EventHubs/src/Diagnostics/EventHubsEventSource.cs
+++ b/sdk/eventhub/Azure.Messaging.EventHubs/src/Diagnostics/EventHubsEventSource.cs
@@ -2602,6 +2602,48 @@ namespace Azure.Messaging.EventHubs.Diagnostics
         }
 
         /// <summary>
+        ///   Indicates that an <see cref="EventHubBufferedProducerClient" /> instance publishing task
+        ///   detected an empty buffer and is entering an idle state to await an event being enqueued.
+        /// </summary>
+        ///
+        /// <param name="identifier">A unique name used to identify the buffered producer.</param>
+        /// <param name="eventHubName">The name of the Event Hub that the buffered producer is associated with.</param>
+        /// <param name="operationId">An artificial identifier for the idle operation.</param>
+        ///
+        [Event(126, Level = EventLevel.Verbose, Message = "Publishing for the buffered producer instance with identifier '{0}' to Event Hub: {1} has detected an empty buffer and is waiting for an event to be enqueued.  Operation Id: '{2}'")]
+        public virtual void BufferedProducerIdleStart(string identifier,
+                                                      string eventHubName,
+                                                      string operationId)
+        {
+            if (IsEnabled())
+            {
+                WriteEvent(126, identifier ?? string.Empty, eventHubName ?? string.Empty, operationId ?? string.Empty);
+            }
+        }
+
+        /// <summary>
+        ///   Indicates that an <see cref="EventHubBufferedProducerClient" /> instance publishing task
+        ///   has detected an event being enqueued and has exiting an idle state.
+        /// </summary>
+        ///
+        /// <param name="identifier">A unique name used to identify the buffered producer.</param>
+        /// <param name="eventHubName">The name of the Event Hub that the buffered producer is associated with.</param>
+        /// <param name="operationId">An artificial identifier for the idle operation.</param>
+        /// <param name="durationSeconds">The total duration that the idle await took place, in seconds.</param>
+        ///
+        [Event(127, Level = EventLevel.Verbose, Message = "Publishing for the buffered producer instance with identifier '{0}' for Event Hub: {1} has completed its idle state; one or more events was enqueued.  Operation Id: '{2},' Duration: '{3:0.00}' seconds.")]
+        public virtual void BufferedProducerIdleComplete(string identifier,
+                                                                 string eventHubName,
+                                                                 string operationId,
+                                                                 double durationSeconds)
+        {
+            if (IsEnabled())
+            {
+                BufferedProducerIdleCompleteCore(127, identifier ?? string.Empty, eventHubName ?? string.Empty, operationId ?? string.Empty, durationSeconds);
+            }
+        }
+
+        /// <summary>
         ///   Indicates that the publishing of events has completed, writing into a stack allocated
         ///   <see cref="EventSource.EventData"/> struct to avoid the parameter array allocation on the WriteEvent methods.
         /// </summary>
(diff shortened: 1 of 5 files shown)


sdk/eventhub/Azure.Messaging.EventHubs/src/Producer/EventHubBufferedProducerClient.cs
```diff
@@ -2204,21 +2202,40 @@ private async Task RunProducerManagementAsync(CancellationToken cancellationToke
...
                                 activeTasks.Add(PublishBatchToPartition(partitionState, releaseGuard: true, activeOperationCancellationSource.Token));
-                                consecutiveEmptyBufferCount = 0;
                             }
 
-                            // If there are no publishing tasks active, introduce a small delay to avoid a tight loop.  If no events are
-                            // available in the buffers, use an increasing delay to avoid wasting resources during periods of low activity.
+                            // If there are no events in the buffer, avoid a tight loop by blocking to wait for events to be enqueued
+                            // after a small delay.
 
-                            if (activeTasks.Count == 0)
```
jsquire, line 2213: This is the root of the issue; we deferred removing tasks until we hit the concurrency limit, rather than sweeping for completed tasks each iteration.   As a result, the criteria was not met.

sdk/eventhub/Azure.Messaging.EventHubs/src/Diagnostics/EventHubsEventSource.cs
```diff
@@ -2601,6 +2601,48 @@ public void UnexpectedException(string errorMessage)
...
+        ///   has detected an event being enqueued and has exiting an idle state.
+        /// </summary>
+        ///
+        /// <param name="identifier">A unique name used to identify the buffered producer.</param>
+        /// <param name="eventHubName">The name of the Event Hub that the buffered producer is associated with.</param>
+        /// <param name="operationId">An artificial identifier for the idle operation.</param>
+        /// <param name="durationSeconds">The total duration that the idle await took place, in seconds.</param>
+        ///
+        [Event(127, Level = EventLevel.Verbose, Message = "Publishing for the buffered producer instance with identifier '{0}' for Event Hub: {1} has completed its idle state; one or more events was enqueued.  Operation Id: '{2},' Duration: '{3:0.00}' seconds.")]
+        public virtual void BufferedProducerIdleComplete(string identifier,
```
christothes, line 2635: nit: `BufferedProducer Entered/Exited IdleState` seems more natural.

sdk/eventhub/Azure.Messaging.EventHubs/src/Producer/EventHubBufferedProducerClient.cs
```diff
@@ -103,6 +97,9 @@ public class EventHubBufferedProducerClient : IAsyncDisposable
         /// <summary>A <see cref="CancellationTokenSource"/> instance to signal that any active publishing operations, including those in-flight, should terminate immediately.</summary>
         private CancellationTokenSource _activeSendOperationsCancellationSource;
 
+        /// <summary>A completion source that be awaited when publishing would like to pause and wait for an event to be enqueued.</summary>
+        private TaskCompletionSource<bool> _eventEnqueuedCompletionSource;
```
christothes, line 101: It seems that the `bool` is not really used in the current implementation. Could this just be a `TaskCompletionSource` or are there reasons we care when it is set to `true` versus `false`?

jsquire: There's no non-generic `TaskCompletionSource` for `netstandard2.0` unfortunately.    _([ref](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskcompletionsource.trysetresult?view=net-6.0&viewFallbackFrom=netstandard-2.0#system-threading-tasks-taskcompletionsource-trysetresult))_

sdk/eventhub/Azure.Messaging.EventHubs/src/Diagnostics/EventHubsEventSource.cs
jsquire: I don't disagree, but Start/Complete better matches the pattern used for the other events.   I'd rather keep the name as-is, but I'll adjust the message for better readability.

sdk/eventhub/Azure.Messaging.EventHubs/src/Producer/EventHubBufferedProducerClient.cs
```diff
@@ -2204,21 +2202,40 @@ private async Task RunProducerManagementAsync(CancellationToken cancellationToke
...
 
-                            if (activeTasks.Count == 0)
+                            if (_totalBufferedEventCount == 0)
                             {
-                                var delay = _totalBufferedEventCount switch
+                                // If completion source doesn't exist or was already set, then swap in a new completion source to be
+                                // set when an event is enqueued.  Allow the publishing loop to tick for one additional check of the
+                                // buffers, to ensure that no events were enqueued during the swap.
+
+                                if ((_eventEnqueuedCompletionSource == null) || (_eventEnqueuedCompletionSource.Task.IsCompleted))
```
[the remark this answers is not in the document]
jsquire: I did.   There's no async version of a ManualResetEvent, which is really the pattern that I'd like here.   Things just didn't work well with a semaphore unless we're either good with dealing with a potential exception on each enqueue call or we want to synchronize each enqueue.

I chose this approach to try and keep enqueue as efficient as possible.   I'd rather pay for the memory barrier and allocation in the background task when the buffers are empty than optimize that path by paying for it when enqueuing.   

That said - if you've got ideas, I'd love to hear them.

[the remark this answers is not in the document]
jsquire: Discussed with Chris offline; we weren't able to come up with a better approach for now.

sdk/eventhub/Azure.Messaging.EventHubs/src/Producer/EventHubBufferedProducerClient.cs
```diff
@@ -2204,21 +2202,40 @@ private async Task RunProducerManagementAsync(CancellationToken cancellationToke
                                 // Responsibility for releasing the guard semaphore is passed to the task.
 
                                 activeTasks.Add(PublishBatchToPartition(partitionState, releaseGuard: true, activeOperationCancellationSource.Token));
-                                consecutiveEmptyBufferCount = 0;
                             }
 
-                            // If there are no publishing tasks active, introduce a small delay to avoid a tight loop.  If no events are
-                            // available in the buffers, use an increasing delay to avoid wasting resources during periods of low activity.
+                            // If there are no events in the buffer, avoid a tight loop by blocking to wait for events to be enqueued
+                            // after a small delay.
```
m-redding, line 2208: Just to confirm - small delay referring to the extra tick of the loop?

sdk/eventhub/Azure.Messaging.EventHubs/tests/Producer/EventHubBufferedProducerClientLiveTests.cs
```diff
@@ -654,6 +656,134 @@ await foreach (var partitionEvent in consumerClient.ReadEventsAsync(cancellation
             }
         }
 
+        /// <summary>
+        ///   Verifies that the <see cref="EventHubBufferedProducerClient" /> is able to publish events.
+        /// </summary>
+        ///
+        [Test]
+        public async Task ProducerCanPublisAfterIdle()
```
m-redding, line 664:
```suggestion
        public async Task ProducerCanPublishAfterIdle()
```

m-redding (commented):
Would it be useful to have a test to make sure no events get left in the queue while idle? Maybe targeting the case where going into idle is happening while an event is being enqueued? Not sure if it's worth the effort or is even possible to make deterministic, but just an idea.

sdk/eventhub/Azure.Messaging.EventHubs/src/Producer/EventHubBufferedProducerClient.cs
jsquire: It's an extra tick after we perform a 100ms delay; this is intended to (hopefully) offset the race between going idle and enqueuing an event by allowing a yield point.

JoshLove-msft (approved):

jsquire:
> Would it be useful to have a test to make sure no events get left in the queue while idle? Maybe targeting the case where going into idle is happening while an event is being enqueued? Not sure if it's worth the effort or is even possible to make deterministic, but just an idea.

It's a good idea, but I don't know that it's feasible to trigger that exact case.   Basically, what we'd need to do is go through the first "make a new completion source" block and then be able to complete an async enqueue call at exactly the moment when we hit the idle check again.   That flow would either see the task as completed and avoid the await or enter the await and get triggered by enqueue.

At that point, we'd kind of be testing that the completion source was thread-safe.   Our tests do something pretty similar but with a bit more determinism to ensure that we're fully idle before we spin back up.

PR #29281, opened 2022-06-14
Simplify OperationInternal

AlexanderSher:
Requires https://github.com/Azure/autorest.csharp/pull/2311

With introduction of `VoidValue` and removal of `OperationInternals`, `OperationInternals<T>`, `OperationOrResponseInternals` and `OperationOrResponseInternals<T>`, it is now possible to remove `OperationInternalBase` type and move all the logic into `OperationInternal<T>`. `OperationInternal` becomes a special case of the `OperationInternal<VoidValue>` that uses `IOperation` instead of `IOperation<T>`.


pshao25 (commented):
Just two small suggestions

sdk/core/Azure.Core/tests/OperationPollerTests.cs
```diff
@@ -49,7 +49,7 @@ public void WaitForCompletionResponseAsyncCancelled()
         {
             var cts = new CancellationTokenSource();
             var poller = new OperationPoller(new TestDelayStrategy(TimeSpan.FromSeconds(10), TimeSpan.FromSeconds(1), cts));
-            var operation = new OperationInternal(new ClientDiagnostics(ClientOptions.Default), new EndlessOperation(), new MockResponse(200));
+            var operation = new OperationInternal<int>(new ClientDiagnostics(ClientOptions.Default), new EndlessOperation(), new MockResponse(200));
```
annelo-msft, line 52: Is it possible to provide test coverage for OperationInternal still?

sdk/core/Azure.Core/src/Shared/OperationInternalOfT.cs
```diff
@@ -99,26 +105,54 @@ internal class OperationInternal<T> : OperationInternalBase
...
+            _updateStatusScopeName = $"{operationTypeName}.UpdateStatus";
         }
 
         private OperationInternal(OperationState<T> finalState)
-            : base(finalState.RawResponse)
         {
             // FinalOperation represents operation that is in final state and can't be updated.
             // It implements IOperation<T> and throws exception when UpdateStateAsync is called.
             _operation = new FinalOperation();
+            _diagnostics = new ClientDiagnostics(ClientOptions.Default);
```
annelo-msft, line 126: Is it correct to use the default client options here?  If the client was instantiated with sanitizers, shouldn't we be using them for LRO messages?

sdk/core/Azure.Core/src/Shared/OperationInternalOfT.cs
```diff
@@ -249,7 +283,145 @@ public Response<T> WaitForCompletion(TimeSpan pollingInterval, CancellationToken
...
+        /// </code>
+        /// </example>
+        /// </summary>
+        /// <param name="cancellationToken">A <see cref="CancellationToken"/> controlling the request lifetime.</param>
+        /// <returns>The HTTP response received from the server.</returns>
+        /// <remarks>
+        /// After a successful run, this method will update <see cref="RawResponse"/> and might update <see cref="HasCompleted"/>.
+        /// </remarks>
+        /// <exception cref="RequestFailedException">Thrown if there's been any issues during the connection, or if the operation has completed with failures.</exception>
+        public async ValueTask<Response> UpdateStatusAsync(CancellationToken cancellationToken) =>
```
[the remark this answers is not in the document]
annelo-msft: I think C# style guidelines recommend keeping operators at the end of lines?  Somewhere we have tooling to enforce style guidelines.

sdk/core/Azure.Core/src/Shared/OperationInternalOfT.cs
```diff
@@ -295,6 +470,22 @@ private static Response GetResponseFromState(OperationState<T> state)
...
+
+            if (_scopeAttributes != null)
+            {
+                foreach (KeyValuePair<string, string> attribute in _scopeAttributes)
+                {
+                    scope.AddAttribute(attribute.Key, attribute.Value);
+                }
+            }
+
+            scope.Start();
```
annelo-msft, line 485: nit: I'm not wild about this factoring, since it separates the start method from the close method, which makes the code that calls this harder to read and verify at a glance.

sdk/core/Azure.Core/src/Shared/OperationPoller.cs
```diff
@@ -28,10 +28,10 @@ public ValueTask<Response> WaitForCompletionResponseAsync(Operation operation, T
         public Response WaitForCompletionResponse(Operation operation, TimeSpan? suggestedInterval, CancellationToken cancellationToken)
             => WaitForCompletionAsync(false, operation, suggestedInterval, cancellationToken).EnsureCompleted();
 
-        public ValueTask<Response> WaitForCompletionResponseAsync(OperationInternalBase operation, TimeSpan? suggestedInterval, CancellationToken cancellationToken)
+        public ValueTask<Response<T>> WaitForCompletionResponseAsync<T>(OperationInternal<T> operation, TimeSpan? suggestedInterval, CancellationToken cancellationToken)
```
annelo-msft, line 31: Operation.WaitForCompletionResponse returns Response, where WaitForCompletion is what returns Response\<T\>.  Having a WaitForCompletionResponse\<T\> seems incorrect to me, and at a minimum can easily lead to misunderstanding of the Operation/Operation\<T\> APIs by anyone reading/maintaining this code.

sdk/core/Azure.Core/tests/OperationPollerTests.cs
AlexanderSher: This is a test for `OperationPoller`. There is no signature that accepts `OperationInternal` anymore.

sdk/core/Azure.Core/src/Shared/OperationPoller.cs
AlexanderSher: Good point. Will fix it.

PR #29295, opened 2022-06-14
hide two classes and one enumeration in `Azure.ResourceManager`

### LINKED ISSUE #29294: Hide `KeyvaultProperties` and `EncryptionProperties` in ResourceManager
These 2 models should not be in Azure.ResourceManager lib, need to be marded as EBNever and remove `PropertyReferenceType` tag.

ArcturusZhang:
Fixes https://github.com/Azure/azure-sdk-for-net/issues/29294


sdk/resourcemanager/Azure.ResourceManager/src/autorest.md
```diff
@@ -14,8 +14,8 @@ mgmt-debug:
   show-request-path: true
 batch:
   - tag: package-common-type-2022-04
-  - tag: package-resources-2022-04
-  - tag: package-management-2022-04
+#   - tag: package-resources-2022-04
```
ArthurMa1978, line 17: consider to remove comment out lines

sdk/resourcemanager/Azure.ResourceManager/src/Common/Custom/Models/EncryptionProperties.cs
```diff
@@ -0,0 +1,16 @@
...
+
+using System.ComponentModel;
+
+namespace Azure.ResourceManager.Models
+{
+    // this class here is to keep the class public and add the EditorBrowsableNever attribute
+    // this class is exposed in resourcemanager by accident, now we hide it in resourcemanager
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public partial class EncryptionProperties
+    {
```
ArthurMa1978, line 14: If empty, then my understanding is a breaking change

sdk/resourcemanager/Azure.ResourceManager/src/autorest.md
ArcturusZhang: Sorry, I forgot this. This is only to accelerate my local generation

sdk/resourcemanager/Azure.ResourceManager/src/Common/Generated/Models/KeyVaultProperties.cs
```diff
@@ -5,24 +5,19 @@
 
 #nullable disable
 
-using Azure.Core;
-
 namespace Azure.ResourceManager.Models
 {
     /// <summary> The KeyVaultProperties. </summary>
-    [PropertyReferenceType]
```
m-nash, line 13: Lets leave these attributes in place and add a provision in autorest.csharp to ignore reference types marked with Obsolete

sdk/resourcemanager/Azure.ResourceManager/src/Common/Custom/Models/EncryptionProperties.cs
```diff
@@ -0,0 +1,16 @@
...
+
+#nullable disable
+
+using System.ComponentModel;
+
+namespace Azure.ResourceManager.Models
+{
+    // this class here is to keep the class public and add the EditorBrowsableNever attribute
+    // this class is exposed in resourcemanager by accident, now we hide it in resourcemanager
+    [EditorBrowsable(EditorBrowsableState.Never)]
```
m-nash, line 12: Add the Obsolete as warning attribute here as well.
`[Obsolete("This type is obsolete and will be removed in a future release.", false)]`

m-nash (changes requested):

sdk/resourcemanager/Azure.ResourceManager/src/autorest.md
```diff
@@ -52,56 +52,90 @@ rename-rules:
...
+      $["x-ms-mgmt-propertyReferenceType"] = true;
+      $["x-namespace"] = "Azure.ResourceManager.Models";
+      $["x-accessibility"] = "public";
+      $["x-csharp-formats"] = "json";
+      $["x-csharp-usage"] = "model,input,output";
   - from: types.json
     where: $.definitions.systemData.properties.*
     transform: >
-      $["readOnly"] = true
+      $["readOnly"] = true;
```
fengzhou-msft, line 100: I think we can update the `x-csharp-usage` to be `model,output`, then we can remove the below workaround to set each property in `systemData` as readOnly.

m-nash (approved):
Approved from Apiview perspective

sdk/resourcemanager/Azure.ResourceManager/src/autorest.md
ArcturusZhang: I will have a try

ArcturusZhang: well, it works, but it also changes a constructor from public to internal, which is a breaking change. I would like to keep the previous implementation...

fengzhou-msft (approved):

PR #29303, opened 2022-06-14
Convert to DPG

### LINKED ISSUE #28590: Generate CLU SDK for 2022-05-01

### LINKED ISSUE #28964: Remove `ConversationAnalysisClient.AnalyzeConversation` and related models
Per discussion, remove `AnalyzeConversation` convenience methods (sync and async) from `ConversationAnalysisClient` and define only an LLC-like method with raw request and response models. Update samples.

heaths:
Resolves #28590 and resolves #28964


heaths:
Merging so I can work on a separate PR for #29331 without jumping through hoops later. The public API was reviewed, and this is mostly a DPG anyway.

heaths:
@annelo-msft or @tg-msft would either of you mind at least looking at the DPG conversion so I can merge this? I'd like to do a separate PR for all the archboard feedback so it's easier to compare gaps with Python.

/cc @kristapratico @mshaban-msft

sdk/cognitivelanguage/test-resources.json
```diff
@@ -24,41 +24,41 @@
         // Developers should define the following environment variables to run tests locally.
```
kristapratico, line 24: just want to check - are the changes in this file intentional?

kristapratico (approved):

sdk/cognitivelanguage/test-resources.json
heaths: Yes. Our test framework tries a number of environment variables beginning with a service directory-prefixed name like `COGNITIVELANUAGE_{name}` then just `{name}` then `AZURE_{name}` for backward compatibility with our older system. Because of that, and so I could easily share _test-resources.json_ with (or from) you folks, I put this back to creating (eventually, when we have dynamic deployment) `AZURE_{name}` variables.

/cc @mshaban-msft

PR #29308, opened 2022-06-15
[Cdn] Resolve arch board review comments

Yao725:

### DIFF
--- a/sdk/cdn/Azure.ResourceManager.Cdn/api/Azure.ResourceManager.Cdn.netstandard2.0.cs
+++ b/sdk/cdn/Azure.ResourceManager.Cdn/api/Azure.ResourceManager.Cdn.netstandard2.0.cs
@@ -506,7 +506,7 @@ namespace Azure.ResourceManager.Cdn
     {
         public FrontDoorRouteData() { }
         public Azure.ResourceManager.Cdn.Models.FrontDoorRouteCacheConfiguration CacheConfiguration { get { throw null; } set { } }
-        public System.Collections.Generic.IList<Azure.ResourceManager.Cdn.Models.FrontDoorActivatedResourceData> CustomDomains { get { throw null; } }
+        public System.Collections.Generic.IList<Azure.ResourceManager.Cdn.Models.FrontDoorActivatedResourceInfo> CustomDomains { get { throw null; } }
         public Azure.ResourceManager.Cdn.Models.FrontDoorDeploymentStatus? DeploymentStatus { get { throw null; } }
         public Azure.ResourceManager.Cdn.Models.EnabledState? EnabledState { get { throw null; } set { } }
         public string EndpointName { get { throw null; } }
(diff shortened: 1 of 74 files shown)


sdk/cdn/Azure.ResourceManager.Cdn/src/autorest.md
```diff
@@ -89,6 +89,9 @@ directive:
                         if(key.endsWith('ActionParameters')) {
                              $[key]['x-ms-client-name'] = newKey + 'Properties';
                         }
+                        if(key.endsWith('ConditionParameters')) {
+                            $[key].properties.operator['x-ms-client-name'] = 'op';
```
ArcturusZhang, line 93: Is this change from comments of arch board? this name reads weird

sdk/cdn/Azure.ResourceManager.Cdn/api/Azure.ResourceManager.Cdn.netstandard2.0.cs
```diff
@@ -2360,7 +2360,7 @@ public partial class LoadContent
     public partial class ManagedCertificateProperties : Azure.ResourceManager.Cdn.Models.SecretProperties
     {
         public ManagedCertificateProperties() { }
-        public string ExpirationDate { get { throw null; } }
+        public System.DateTimeOffset? ExpirationOn { get { throw null; } }
```
ArcturusZhang, line 2363: The name change of this property is introduced by hard-coded logic in our generator?
Should this be `ExpiredOn`? Do we have similar cases in other RPs?

sdk/cdn/Azure.ResourceManager.Cdn/src/autorest.md
Yao725: Yes, the main reason of this change is that arch board want to avoid the occurrence of `@operator` in the method parameters, here is the [comment](https://apiview.dev/Assemblies/Conversation/43b0983f23fe4e07b9acdd88d99492c6#517929f6ed4d4b1c80f9f5c50c8c0ad7). Unfortunately I didn't find a way to only change the parameter name in model ctor, and I also don't want to change generator to make exemption for this edge case, so I just renamed it.

sdk/cdn/Azure.ResourceManager.Cdn/api/Azure.ResourceManager.Cdn.netstandard2.0.cs
Yao725: I think so, the logic is [here](https://github.com/Azure/autorest.csharp/blob/feature/v3/src/AutoRest.CSharp/Mgmt/Decorator/Transformer/RenameTimeToOn.cs). We should have the similar cases in other RPs, for example, we renamed `updatedTime` in ResourceManager to `updatedOn`.

ArcturusZhang: Therefore I believe from grammar's perspective, we should have `expiredOn` and `expirationDate`. `expirationOn` does not make much sense.

Yao725: I renamed the property to `ExpiresOn` manually according to the sync meeting discussion, and create an [issue](https://github.com/Azure/autorest.csharp/issues/2316) to make sure we implement it in our generator.

ArthurMa1978 (approved):

PR #29320, opened 2022-06-15
Resolve archboard review comments for KeyVault

ArthurMa1978:


ArcturusZhang (commented):
Can we rename all the `Vault` thing to `KeyVault`? KeyVault is the name that we usually reference these resources

ArcturusZhang (approved):

PR #29339, opened 2022-06-16
Update AutoRest C# version 3.0.0-beta.20220621.1

ArthurMa1978:
https://github.com/Azure/autorest.csharp/pull/2318

Here also add a new script Generate-Mgmt-Sdk.ps1 which can easily re-gen all sdk by default `dotnet build` or special local build of autorest


m-nash (changes requested):
Overall I like the change as it gives us a bulk update capability but I left a comment here https://apiview.dev/Assemblies/Review/ac5c88f4ea2941609517815a69afa7ec/06c44c9f5dba441eb8f2187b6542596c?diffRevisionId=4173d69a16984391b855ef0bd55eea0c&doc=False&diffOnly=True#75a63573255a4a7e9102700b61b8cd2b asking how confident we are that all of these are of type `ResourceType`.   More importantly can we define what action we would take if we found out one of these was really just a string after we already hit GA?

m-nash (approved):
Approved

eng/scripts/Generate-Mgmt-Sdk.ps1
```diff
@@ -0,0 +1,32 @@
...
+    if ($file.Contains("Azure.ResourceManager")) {
+        $workPath = Split-Path -Path $file -Parent
+        Set-Location -Path $workPath
+        
+        if ([string]::IsNullOrWhitespace($AutorestPath)) { 
+            dotnet build /t:GenerateCode
+        }
+        else {
+            autorest .\autorest.md --use:$AutorestPath
+        }
```
Yao725, line 29: Nit: can we also update the api here?

fengzhou-msft:
Consolidated changes from #29234

eng/scripts/Generate-Mgmt-Sdk.ps1
ArthurMa1978: api already have script, which can run it after run this script

PR #29353, opened 2022-06-17
[perf] Add pipeline template and storage pipelines

mikeharder:


sdk/storage/Azure.Storage.Files.Shares/perf.yml
```diff
@@ -0,0 +1,32 @@
...
+    displayName: Iterations (times to run each test)
+    type: number
+    default: '5'
+  - name: additionalArguments
+    displayName: AdditionalArguments (passed to PerfAutomation)
+    type: string
+    default: ' '
+
+extends:
+  template: ../../../eng/pipelines/templates/jobs/perf.yml
```
weshaggard, line 24:
```suggestion
  template: /eng/pipelines/templates/jobs/perf.yml
```

eng/pipelines/templates/jobs/perf.yml
```diff
@@ -0,0 +1,111 @@
...
+    value: 'azsdk-pool-mms-ubuntu-2004-perf'
+  - name: OSVmImage
+    value: 'MMSUbuntu20.04'
+
+resources:
+  repositories:
+  - repository: azure-sdk-tools
+    type: github
+    endpoint: Azure
+    name: Azure/azure-sdk-tools
```
weshaggard, line 42: We should include a ref, usually a release tag.

eng/pipelines/templates/jobs/perf.yml
```diff
@@ -0,0 +1,111 @@
...
+    default: '5'
+  - name: envVars
+    type: object
+    default: {}
+  - name: additionalArguments
+    type: string
+    default: ''
+
+variables:
+  - name: NugetSecurityAnalysisWarningLevel
```
weshaggard, line 28: We should include the globals.yml file to get these common properties.

eng/pipelines/templates/jobs/perf.yml
```diff
@@ -0,0 +1,111 @@
...
+
+    - template: /eng/common/pipelines/templates/steps/verify-agent-os.yml
+      parameters:
+        AgentImage: $(OSVmImage)
+
+    - template: ../steps/install-dotnet.yml
+      parameters:
+        EnableNuGetCache: false
+
+    - template: ../../../common/TestResources/deploy-test-resources.yml
```
weshaggard, line 65:
```suggestion
    - template: /eng/common/TestResources/deploy-test-resources.yml
```

weshaggard: These are usually best to absolute paths from the root of the repo.

eng/pipelines/templates/jobs/perf.yml
mikeharder: For now we are using main.  I will add an explicit reference to main.  I know this means the build is unstable, but while the tooling is still being actively developed, I don't want to require PRs to every language every time we update the tooling.

eng/pipelines/templates/jobs/perf.yml
```diff
@@ -0,0 +1,109 @@
...
+        get-content results.json
+      workingDirectory: azure-sdk-tools/tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/results
+      displayName: Print results.json
+
+    - template: /eng/common/TestResources/remove-test-resources.yml
+      parameters:
+        ServiceDirectory: ${{ parameters.serviceDirectory }}
+        ResourceType: perf
+
+    - task: PublishPipelineArtifact@1
```
benbp, line 106: You may want to put this above `remove-test-resources.yml` so you still publish results in case the resource removal script fails for some reason.

benbp (approved):

Issue #26514, opened 2022-01-20
Enable implementors of TypeBinder<TExchange> to serialize properties from explicitly implemented interfaces


christothes:
### Library name

Azure.Core

### Please describe the feature.

Currently `TypeBinder<TExchange>` doesn't support serialization of properties that are explicitly implemented from an interface.

Ideas for this change include:
- addition of method `protected virtual bool ShouldSerialize(MemberInfo info)`
  - Allows for implementors to have find grained control over how the TExchange type's properties are considered for serialization.

Areas of interest in the code include:
https://github.com/Azure/azure-sdk-for-net/blob/dcac296df717dbb75fcc50648c2132d47cd413f2/sdk/core/Azure.Core/src/Shared/TypeBinder.cs#L74
https://github.com/Azure/azure-sdk-for-net/blob/dcac296df717dbb75fcc50648c2132d47cd413f2/sdk/core/Azure.Core/src/Shared/TypeBinder.cs#L170-L171

related to #26363

jsquire:
It's worth noting that the current lack of support for explicit interface implementation creates a poor developer experience when using F# with `Azure.Data.Tables`.   Because F# lacks implicit interface implementation, it is necessary to create a shadow for each property of `ITableEntity` when creating  table entitties in F#.   For example:

```fsharp
type Customer() =

    interface ITableEntity with
        member val PartitionKey = "" with get, set
        member val RowKey = "" with get, set
        member val Timestamp = Nullable () with get, set
        member val ETag = ETag () with get, set

    member val PartitionKey = "" with get, set
    member val RowKey = "" with get, set
    member val Timestamp = Nullable() with get, set
    member val ETag = ETag() with get, set

let customer = Customer()
customer.PartitionKey <- "SomePartitionKey"
customer.RowKey <- "SomeRowKey"

table.AddEntity customer |> ignore
```

More context can be found in [F# Azure Table Storage sample is broken (#28828)](https://github.com/dotnet/docs/issues/28828).

isaacabraham:
Indeed, the error you get from this is difficult to diagnose and will confuse basically every F# developer who tries this.

marklio:
[...]

PR #29364, opened 2022-06-18
Support explicit implementations of ITableEntity in the TypeBinder.

marklio:
* Add support to TypeBinder for binding ITableEntity properties when:
  * the behavior is requested
  * The type implements ITableEntity
  * The type doesn't provide implementations of properties with the same names
* Opt TablesTypeBinder into this new behavior

### DIFF
--- a/sdk/core/Azure.Core/src/Shared/TypeBinder.cs
+++ b/sdk/core/Azure.Core/src/Shared/TypeBinder.cs
@@ -7,7 +7,6 @@ using System.Collections.Generic;
 using System.Linq.Expressions;
 using System.Reflection;
 using System.Runtime.Serialization;
-using Azure.Data.Tables;
 
 namespace Azure.Core
 {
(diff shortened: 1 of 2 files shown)


marklio:
This is offered as an approach to address #26514 for the specific case of the ITableEntity interface.

marklio:
Due to the separation of components, I reworked it so that TypeBinder doesn't need to know specifically about ITableEntity, but TabesTypeBinder passes in ITableEntity as an interface of interest for additional binding of properties if they aren't provided implicitly.

marklio:
I think it's totally reasonable to think about the general case as a broader feature. I will point out that older SDKs did not have this broader feature. They did, however, honor implicit implementations of ITableEntity (due to ITableEntity playing a larger role in serialization). As a result, the lack of ITableEntity support here is a serious blocker for moving from old SDKs, and so I think it should have a much higher priority than a generalized feature that older SDKs did not have.

marklio:
Oh sure, no doubt. I just wanted to put the right priority on the narrow case.

marklio:
@christothes Since the bot is flagging this PR as stale... How is resolution on coming #26514?

christothes:
Thanks for contributing your thoughts here @marklio - this PR helped inform the change in #30990

marklio:
> this PR helped inform the change in #30990

That is great! Thank you so much for addressing this! I'll close this PR

PR #29365, opened 2022-06-19
Always populate the sequence number if present when checkpointing

JoshLove-msft:


sdk/eventhub/Azure.Messaging.EventHubs.Shared/src/BlobCheckpointStore/BlobCheckpointStoreInternal.cs
```diff
@@ -441,10 +441,10 @@ await foreach (BlobItem blob in ContainerClient.GetBlobsAsync(traits: BlobTraits
                 offset = result;
                 startingPosition = EventPosition.FromOffset(result, false);
             }
-            else if (metadata.TryGetValue(BlobMetadataKey.SequenceNumber, out str) && long.TryParse(str, NumberStyles.Integer, CultureInfo.InvariantCulture, out result))
+            if (metadata.TryGetValue(BlobMetadataKey.SequenceNumber, out str) && long.TryParse(str, NumberStyles.Integer, CultureInfo.InvariantCulture, out result))
             {
                 sequenceNumber = result;
```
JoshLove-msft, line 446: The sequenceNumber is not exposed publicly, but it is used in the internal BlobStorageCheckpoint which is used by the EH extensions.

sdk/eventhub/Azure.Messaging.EventHubs.Shared/src/BlobCheckpointStore/BlobCheckpointStoreInternal.cs
```diff
@@ -441,10 +441,10 @@ await foreach (BlobItem blob in ContainerClient.GetBlobsAsync(traits: BlobTraits
...
-                startingPosition = EventPosition.FromSequenceNumber(result, false);
+                startingPosition ??= EventPosition.FromSequenceNumber(result, false);
```
JoshLove-msft, line 447: This maintains the existing behavior where offset takes precedence over sequence number when creating the position.

jsquire: This is consistent with the expected behavior of the Event Processor as well.

sdk/eventhub/Azure.Messaging.EventHubs.Shared/src/BlobCheckpointStore/BlobCheckpointStoreInternal.cs
jsquire: Since this is shared code, I double checked and confirmed that this is safe for the processor as well.   The metadata for a checkpoint is not considered, only the starting position.  _([ref](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/EventProcessor.cs#L1724))_

sdk/eventhub/Azure.Messaging.EventHubs.Shared/tests/BlobCheckpointStore/BlobsCheckpointStoreInternalTests.cs
```diff
@@ -413,6 +413,11 @@ public async Task GetCheckpointUsesSequenceNumberAsTheStartingPositionWhenNoOffs
 
             Assert.That(checkpoint, Is.Not.Null, "A checkpoint should have been returned.");
             Assert.That(checkpoint.StartingPosition, Is.EqualTo(expectedStartingPosition));
+
+            Assert.That(checkpoint, Is.InstanceOf<BlobCheckpointStoreInternal.BlobStorageCheckpoint>());
+            var blobCheckpoint = (BlobCheckpointStoreInternal.BlobStorageCheckpoint)checkpoint;
+            Assert.IsNull(blobCheckpoint.Offset);
```
jsquire, line 419: Could we please stick to the `Assert.That` format and include a failure message since there are multiple asserts in this test?

sdk/eventhub/Azure.Messaging.EventHubs.Shared/tests/BlobCheckpointStore/BlobsCheckpointStoreInternalTests.cs
```diff
@@ -1202,6 +1207,11 @@ public async Task GetCheckpointPreferredNewCheckpointOverLegacy()
             Assert.That(checkpoint, Is.Not.Null, "A single checkpoint should have been returned.");
             Assert.That(checkpoint.StartingPosition, Is.EqualTo(EventPosition.FromOffset(14, false)));
             Assert.That(checkpoint.PartitionId, Is.EqualTo("0"));
+
+            Assert.That(checkpoint, Is.InstanceOf<BlobCheckpointStoreInternal.BlobStorageCheckpoint>());
+            var blobCheckpoint = (BlobCheckpointStoreInternal.BlobStorageCheckpoint)checkpoint;
+            Assert.AreEqual(14, blobCheckpoint.Offset);
```
jsquire, line 1213: Could we please stick to the `Assert.That` format and include a failure message since there are multiple asserts in this test?

jsquire (approved):

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Microsoft.Azure.WebJobs.Extensions.EventHubs.csproj
```diff
@@ -3,9 +3,9 @@
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0</TargetFrameworks>
     <Description>Microsoft Azure WebJobs SDK EventHubs Extension</Description>
-    <Version>5.1.0-beta.1</Version>
+    <Version>5.1.1</Version>
     <!--The ApiCompatVersion is managed automatically and should not generally be modified manually.-->
-    <ApiCompatVersion>5.0.1</ApiCompatVersion>
+    <ApiCompatVersion>5.1.0</ApiCompatVersion>
```
JoshLove-msft, line 8: This did not get updated because 5.1.0 was shipped out of a feature branch and not merged back to main.

PR #29379, opened 2022-06-20
Change to dynamically use the TagResource API

### LINKED ISSUE #2525: The generator should not generate tags operation methods when it does not a put or patch
Example: VirtualMachineExtensionImageResource in compute

m-nash:
Fix https://github.com/Azure/autorest.csharp/issues/2525
There are some environments like AzureStack which do not have the TagResource deployed.  In these cases we need to fall back to use the PATCH/PUT methods for updating the tags for a given resource.  The benefit to TagResource is it utilizes either the TagResource permission OR the write permission on the resource where as PATCH/PUT only works with write permission on the resource.

This change dynamically checks to see if the TagResource API is deployed in the current environment and switches which implementation it uses for manipulating tags.  I also added an override in the options, if we don't want this we will need to find a way to override this at runtime in the test cases to test both cases.


ArthurMa1978 (approved):

weshaggard:
@m-nash this change is killing our machine pool because it is touching so many files which trigger all the pipelines. Every time a new commit is pushed we are requesting about 1000 agents which is backing up everything else trying to run. Can you please avoid pushing a lot of commits and only push when everything is complete? Also in the future if you are going to have a large change that touches all these files please reach out to EngSys first so we can come up with a plan to best handle this load.

m-nash:
@weshaggard spoke offline to come up with strategies on how to remove duplicate runs given the base library resourcemanager now runs all tests from all dependencies.

@ArthurMa1978  FYI

archerzz (approved):

m-nash:
changes were merged in here https://github.com/Azure/azure-sdk-for-net/commit/49d50b33ccb348eb665cb3e27428ad095e09b449

Issue #29398, opened 2022-06-21
[FormRecognizer] Add Length property to BoundingPolygon


kinelski:
See https://github.com/Azure/azure-sdk-for-net/issues/28972#issuecomment-1162143753 for more context.

`BoundingPolygon` instances can have a varying number of points. However, we only expose them through an indexer and we don't provide a `Length` property. Users don't know until which index they need to iterate to get all the points.

Action items:

1. Create a `BoundingPolygon.Length` property.
2. Update our samples to start using this new property. Currently what we do is iterating until `BoundingPolygon.Points.Length`, but the `Points` property is internal and can't be used by customers.

PR #29399, opened 2022-06-21
[FormRecognizer] Updated samples to hardcode 4 points length in bounding polygon

kinelski:
Related to https://github.com/Azure/azure-sdk-for-net/issues/29398.

Our samples iterate over bounding polygons using the `BoundingPolygon.Points.Length` property, which is not possible for customers since the `Points` property is marked as internal. Ideally, we want to expose a public `BoundingPolygon.Length` property.

Since our next release will not happen before August, I didn't want to start using the `BoundingPolygon.Length` property in samples right now given that it will only confuse costumers. We're hardcoding the polygon length to 4 to make sure users can compile the samples and they work as intended.

### DIFF
--- a/sdk/formrecognizer/Azure.AI.FormRecognizer/MigrationGuide.md
+++ b/sdk/formrecognizer/Azure.AI.FormRecognizer/MigrationGuide.md
@@ -404,7 +404,7 @@ foreach (DocumentPage page in result.Pages)
 
         Console.WriteLine($"    Its bounding polygon (points ordered clockwise):");
 
-        for (int j = 0; j < line.BoundingPolygon.Points.Length; j++)
+        for (int j = 0; j < 4; j++)
         {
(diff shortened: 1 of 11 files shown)


sdk/formrecognizer/Azure.AI.FormRecognizer/MigrationGuide.md
```diff
@@ -404,7 +404,7 @@ foreach (DocumentPage page in result.Pages)
 
         Console.WriteLine($"    Its bounding polygon (points ordered clockwise):");
 
-        for (int j = 0; j < line.BoundingPolygon.Points.Length; j++)
+        for (int j = 0; j < 4; j++)
```
kinelski, line 407: All bounding polygons have a fixed length of 4. The service team wants to increase the number of points to better fit the visual elements in the future. Hardcoding to 4 works until we expose a `Length` property in the next release.

pallavit (approved):

PR #29422, opened 2022-06-22
Adding test coverage for account SAS generated in different resource …

amnguye:
…types, permission types, and service types

Found a test gap where we weren't testing for ensuring that the SDK was not altering or manipulating the SAS in any form or shape. Since the SDK will generate a SAS with a fixed resource type, service type and permissions type, it was difficult to test the many different ways a customer could possibly generate a SAS with different orders of the resource type, permission types and service types.

According to the documentation for Account SAS, the storage service does not care about the order of resource type, permission types and service types.
See: https://docs.microsoft.com/en-us/rest/api/storageservices/create-account-sas#specify-the-account-sas-parameters

Not to be confused with Service SAS which does require the ordering to be correct for permissions.

### DIFF
--- a/sdk/storage/Azure.Storage.Common/tests/Sas/TestAccountSasBuilder.cs
+++ b/sdk/storage/Azure.Storage.Common/tests/Sas/TestAccountSasBuilder.cs
@@ -2,8 +2,8 @@
 // Licensed under the MIT License.
 using System;
 using System.Collections.Generic;
+using System.Text;
 using Azure.Storage.Sas;
-using Castle.Core.Internal;
 
 namespace Azure.Storage.Test.Shared
 {
@@ -114,7 +114,7 @@ namespace Azure.Storage.Test.Shared
         internal static void ValidateAndRawTypes(string rawString,
             List<char> validCharacters)
         {
-            if (rawString.IsNullOrEmpty())
+            if (string.IsNullOrEmpty(rawString))
             {
                 throw new ArgumentException($"{rawString} is empty or null, pass valid type");
             }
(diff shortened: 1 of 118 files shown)


sdk/storage/Azure.Storage.Common/tests/Sas/TestAccountSasBuilder.cs
```diff
@@ -0,0 +1,233 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+using System;
+using System.Collections.Generic;
+using Azure.Storage.Sas;
+using Castle.Core.Internal;
```
amnguye, line 6: TODO remove this dependency

jaschrep-msft (commented):
I'm not seeing much different between the tests besides making a client and attempting a GetProperties(). Is it possible to reduce the code copying via test-inheritance or some shared-code setup methods?

sdk/storage/Azure.Storage.Files.DataLake/tests/ServiceClientTests.cs
```diff
@@ -976,31 +976,6 @@ public void GenerateAccountSas_Builder()
             expectedUri.Query += sasBuilder.ToSasQueryParameters(constants.Sas.SharedKeyCredential).ToString();
             Assert.AreEqual(expectedUri.Uri, sasUri);
         }
-
-        [RecordedTest]
```
amnguye, line 980: TODO: add back.

jaschrep-msft (approved):

Issue #29267, opened 2022-06-13
[FEATURE REQ] Enable support to GZip a request payload


pallavit:
### Library name

Azure.core

### Please describe the feature.

The Azure Monitor Logs Ingestion library (see https://github.com/Azure/azure-sdk-for-net/issues/29207) needs a way to Gzip a request payload before sending it to the service. Even though this can be built in the convenience layer doing so will have both performance and logging implications and so doing it in core would be the best approach here.

Also - this is something we would like to support per request.

PR #29426, opened 2022-06-22
POC for GZipUtf8JsonRequestContent

christothes:
#29267


sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -22,5 +24,183 @@ public partial class LogsIngestionClient
...
+                    WriteMemory(stream, BinaryData.FromString("]").ToMemory());
+                    stream.Position = 0; // set Position to 0 to return everything from beginning of stream
+                    yield return (BinaryData.FromStream(stream));
+                    stream = new MemoryStream(_singleUploadThreshold); // reset stream
+                }
+                else
+                {
+                    WriteMemory(stream, memory);
+                    WriteMemory(stream, BinaryData.FromString(",").ToMemory());
+                    if ((entryCount + 1) == logEntries.Count())
```
tg-msft, line 91: `logEntries.Count()` might be waaaaay too expensive to do on each iteration.  We should just double this logic up after the loop too.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -22,5 +24,183 @@ public partial class LogsIngestionClient
...
+        /// Hidden method for batching data
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="logEntries"></param>
+        /// <returns></returns>
+        internal static IEnumerable<BinaryData> Batching<T>(IEnumerable<T> logEntries)
+        {
+            //TODO: use Array pool instead
+            MemoryStream stream = new MemoryStream(_singleUploadThreshold);
+            WriteMemory(stream, BinaryData.FromString("[").ToMemory());
```
tg-msft, line 73: We should probably just create constants for these three characters with something like:
```C#
        private const byte JsonArrayOpenUtf8 = 0x5B; // '['
        private const byte JsonArrayCloseUtf8 = 0x5D; // ']'
        private const byte JsonArraySeparatorUtf8 = 0x2C; // ','
```
The code you have now will allocate a lot of little arrays we can avoid.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -22,5 +24,183 @@ public partial class LogsIngestionClient
...
+                        yield return (BinaryData.FromStream(stream));
+                    }
+                }
+                entryCount++;
+            }
+        }
+
+        private static void WriteMemory(MemoryStream stream, ReadOnlyMemory<byte> memory)
+        {
+            stream.Write(memory.ToArray(), 0, memory.Length); //TODO: fix ToArray
```
tg-msft, line 105: I think `memory.ToArray()` is allocating here.  Can you try `memory.AsSpan()` instead?

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -22,5 +24,183 @@ public partial class LogsIngestionClient
         protected LogsIngestionClient()
         {
         }
+
+        // The size we use to determine whether to upload as a single PUT BLOB
+        // request or stage as multiple blocks.
+        private const int _singleUploadThreshold = 1000000; // 1 Mb in byte format
```
tg-msft, line 30: Nit - we typically use PascalCase for constants.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -22,5 +24,183 @@ public partial class LogsIngestionClient
...
+        /// <summary>
+        /// Hidden method for batching data
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="logEntries"></param>
+        /// <returns></returns>
+        internal static IEnumerable<BinaryData> Batching<T>(IEnumerable<T> logEntries)
+        {
+            //TODO: use Array pool instead
+            MemoryStream stream = new MemoryStream(_singleUploadThreshold);
```
tg-msft, line 72: We should back this with something from an ArrayPool and then wrap this in a GZipStream ourselves to avoid extra data copies.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -22,5 +24,183 @@ public partial class LogsIngestionClient
...
+            foreach (var log in logEntries)
+            {
+                BinaryData entry = log is BinaryData d ? d : BinaryData.FromObjectAsJson(log);
+
+                var memory = entry.ToMemory();
+                if ((stream.Length + memory.Length + 1) >= _singleUploadThreshold) // if adding this entry makes stream > 1 Mb send current stream now
+                {
+                    WriteMemory(stream, BinaryData.FromString("]").ToMemory());
+                    stream.Position = 0; // set Position to 0 to return everything from beginning of stream
+                    yield return (BinaryData.FromStream(stream));
```
tg-msft, line 84: We should probably just return the stream itself to avoid an extra memory copy into BinaryData.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -22,5 +24,183 @@ public partial class LogsIngestionClient
...
+        {
+            Argument.AssertNotNullOrEmpty(ruleId, nameof(ruleId));
+            Argument.AssertNotNullOrEmpty(streamName, nameof(streamName));
+            Argument.AssertNotNullOrEmpty(logEntries, nameof(logEntries));
+
+            using var scope = ClientDiagnostics.CreateScope("LogsIngestionClient.Upload");
+            scope.Start();
+            try
+            {
+                foreach (var partition in Batching(logEntries))
```
tg-msft, line 143: We should add an `ObjectSerializer` property on your `ClientOptions` and pass it in here so customers can use STJ by default, Newtonsoft if they want, etc.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -22,5 +24,183 @@ public partial class LogsIngestionClient
...
+                {
+                    //TODO: catch errors and correlate with Batching.start
+                    return Upload(ruleId, streamName, partition, "gzip", new RequestContext() { CancellationToken = cancellationToken });
```
tg-msft, line 146: Nit - create a constant for `gzip`

Nit - pull the `RequestContext` out so you don't allocate a new one for every partition

Ditto for the Async method

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -22,5 +24,183 @@ public partial class LogsIngestionClient
...
+        /// <returns> The response returned from the service. </returns>
+        /// <example>
+        /// This sample shows how to call Upload with required parameters and request content.
+        /// <code><![CDATA[
+        /// var credential = new DefaultAzureCredential();
+        /// var endpoint = new Uri("<https://my-account-name.azure.com>");
+        /// var client = new LogsIngestionClient(endpoint, credential);
+        ///
+        /// var data = new[] {
+        ///     new {}
```
tg-msft, line 125: Is this supposed to be coming from a snippet?

tg-msft (approved):
This is looking really good overall.  I left some suggestions on how we might be able to allocate/copy a little less though.

sdk/core/Azure.Core/src/Shared/GZipUtf8JsonRequestContent.cs
```diff
@@ -0,0 +1,78 @@
...
+
+namespace Azure.Core
+{
+    internal class GZipUtf8JsonRequestContent: RequestContent
+    {
+        private GZipStream _gzip;
+        private readonly MemoryStream _stream;
+        private readonly RequestContent _content;
+
+        public Utf8JsonWriter JsonWriter { get; }
```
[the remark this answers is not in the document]
christothes: This is modeled after [UTF8JsonRequestContent](https://github.com/Azure/autorest.csharp/blob/6b849d88a08151603d29684d2234d294c310ff55/src/assets/Generator.Shared/Utf8JsonRequestContent.cs)

Currently, [the generator](https://grep.app/search?q=content.JsonWriter.&filter[repo][0]=Azure/azure-sdk-for-net) uses this property to write Json payloads directly.

sdk/core/Azure.Core/src/Shared/GZipUtf8JsonRequestContent.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        public GZipUtf8JsonRequestContent(RequestContent content)
+        {
+            _stream = new MemoryStream();
+            _gzip = new GZipStream(_stream, CompressionMode.Compress, true);
+            _content = Create(_stream);
+            JsonWriter = new Utf8JsonWriter(_gzip);
+            content.WriteTo(_gzip, default);
```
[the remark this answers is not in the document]
christothes: This may not be needed by clients, but it allows for scenarios where the payload has already been built for some reason without being GZipped. We can probably remove for v1.

sdk/core/Azure.Core/src/Shared/GZipUtf8JsonRequestContent.cs
```diff
@@ -0,0 +1,78 @@
...
+    {
+        private GZipStream _gzip;
+        private readonly MemoryStream _stream;
+        private readonly RequestContent _content;
+
+        public Utf8JsonWriter JsonWriter { get; }
+
+        public GZipUtf8JsonRequestContent(RequestContent content)
+        {
+            _stream = new MemoryStream();
```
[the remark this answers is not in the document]
christothes: It's worth looking into. I'm not sure if there is a reason UTF8JsonRequestContent doesn't do this either. Perhaps @AlexanderSher has some thoughts there.

sdk/core/Azure.Core/src/Shared/GZipUtf8JsonRequestContent.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        public override async Task WriteToAsync(Stream stream, CancellationToken cancellation)
+        {
+            await FlushAsync().ConfigureAwait(false);
+            await _content.WriteToAsync(stream, cancellation).ConfigureAwait(false);
+        }
+
+        public override void WriteTo(Stream stream, CancellationToken cancellation)
+        {
+#if NETFRAMEWORK
```
JoshLove-msft, line 50: Why is this needed?

nisha-bhatia:
LogsIngestionClient and LogsIngestionClientLive tests will be addressed in a separate PR

sdk/core/Azure.Core/src/Shared/GZipUtf8JsonRequestContent.cs
nisha-bhatia: I think this because in .NET framework, we have to flush before we dispose the gzip stream. @christothes

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -22,5 +24,190 @@ public partial class LogsIngestionClient
...
+            Argument.AssertNotNullOrEmpty(logEntries, nameof(logEntries));
+
+            using var scope = ClientDiagnostics.CreateScope("LogsIngestionClient.Upload");
+            scope.Start();
+            try
+            {
+                foreach (BinaryData partition in Batch(logEntries))
+                {
+                    //TODO: catch errors and correlate with Batch.start
+                    return Upload(ruleId, streamName, partition, "gzip", new RequestContext() { CancellationToken = cancellationToken });
```
JoshLove-msft, line 153: Hmm, does this mean we will only upload the first partition in the batch? Is this just for prototyping?

sdk/core/Azure.Core/src/Shared/GZipUtf8JsonRequestContent.cs
christothes: yeah - this explains it https://github.com/dotnet/runtime/issues/15371

JoshLove-msft: Interesting! Would it be worth linking to the issue in a comment?

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
nisha-bhatia: this has been updated

sdk/core/Azure.Core/src/Shared/GZipUtf8JsonRequestContent.cs
```diff
@@ -0,0 +1,115 @@
...
+#pragma warning restore CA2213 // Disposable fields should be disposed
+        private readonly RequestContent _content;
+
+        public Utf8JsonWriter JsonWriter { get; }
+
+        public GZipUtf8JsonRequestContent(RequestContent content)
+        {
+            _stream = new MemoryStream();
+            _gzip = new GZipStream(_stream, CompressionMode.Compress, true);
+            _content = Create(_stream);
```
KrzysztofCwalina, line 28: It seems like we are doing too many copies with this implementation. This create method creates StreamRequestContent, which stores a stream and then it's written to the network, it reads the stored stream to an array and then writes it to the destination. I think this filed (_content) should be simply MemoryStream and not RequestContent, and then we would avoid some of the unnecessary copies.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -137,28 +152,51 @@ private static void WriteMemory(MemoryStream stream, ReadOnlyMemory<byte> memory
...
 
             using var scope = ClientDiagnostics.CreateScope("LogsIngestionClient.Upload");
             scope.Start();
+
+            var requestContext = new RequestContext() { CancellationToken = cancellationToken };
+            requestContext.AddClassifier(500, false);
+            requestContext.AddClassifier(403, false);
+            requestContext.AddClassifier(413, false);
+            requestContext.AddClassifier(429, false);
+            requestContext.AddClassifier(503, false);
```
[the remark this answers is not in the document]
nisha-bhatia: I extracted this code into a helper method but I'm not sure if we should re-use it across Upload calls. I wouldn't want the RequestContent to get modified somehow

sdk/core/Azure.Core/src/Shared/GZipUtf8JsonRequestContent.cs
christothes: The copying is primarily to read the uncompressed input `RequestContext` into the newly created `RequestContext` (`_context`) which is required to compress it into the gzip stream.  I'm not sure we can avoid these copies to accomplish this. But I could be missing something.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -30,6 +30,18 @@ protected LogsIngestionClient()
         // request or stage as multiple blocks.
         private const int SingleUploadThreshold = 1000000; // 1 Mb in byte format
 
+        internal readonly struct BatchedLogs <T>
+        {
+            public BatchedLogs(List<T> X, BinaryData Y)
```
christothes, line 35: nit: 
```suggestion
            public BatchedLogs(List<T> logList, BinaryData logData)
```

sdk/core/Azure.Core/src/Shared/GZipUtf8JsonRequestContent.cs
nisha-bhatia: https://github.com/Azure/azure-sdk-for-net/issues/30691

sdk/core/Azure.Core/src/Shared/GZipUtf8JsonRequestContent.cs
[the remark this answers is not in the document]
nisha-bhatia: https://github.com/Azure/azure-sdk-for-net/issues/30693

PR #29441, opened 2022-06-23
Eventhub Stress Read Tests

### LINKED ISSUE #22436: Event Hubs: Legitimize the stress test suite
# Summary

While effective, the current suite of stress tests for Event Hubs is largely ad-hoc and lives in a private repository.  With a common approach to long-running stress tests now agreed upon, the suite should be revised for a more structured approach, reworked to fit the common approach, integrated with the shared hosting assets, and moved into the official repository.

This issue is primarily intended for personal thoughts and notes;  it is not expected that it be a clear and comprehensive work item that could be fully understood and implemented by others.  Please expect some gaps in descriptions and a non-trivial amount of assumed knowledge.

# Scope of Work

- Refactor the existing ad-hoc stress tests to remove code duplication, follow a modular architecture, and follow the conventions used for the other Event Hubs areas.  Like the performance suite, the stress tests should make use of the Event Hubs shared code project for common items.

- Integrate metrics collection with Application Insights, following the common conventions agreed upon.  _(coordinate with Ben and Richard)_

- Integrate chaos injection with the current platform toolset.  _(coordinate with Ben and Richard)_

- Adjust the architecture to support the concept of "roles" for the stress testing, allowing scenarios to be focused without taking on additional resource overhead.  For example, tests for the event processor should not include an embedded producer; that should exist as an independent role that may be deployed separately and then coordinated with.   Roles should probably include:
    - Publisher
    - Processor
    - Consumer 
    - Partition Receiver

- Allow for the concept of an orchestrator to control scenarios.  The orchestrator will be responsible for deploying roles needed and coordinating between the roles for a scenario.   For example, the orchestrator should be able to run a processor-based scenario that:
    - Deploys a set of publishers
    - Deploys a set of processors
    - Configures both roles to shared some form of key/value store to coordinate around what events have been published
    - Can simulate processor scale up/down
    - Configures both roles to publish metrics to some common location.

- Roles should have the ability to react to configuration that is specific to the scenario that they're running in.  For example, a publisher should be able to be told to collect metrics or not, allowing noise to be removed from consuming scenarios.

- Stress tests should use an approach to bootstrap consistent with Test Resources and the dynamic creation of child entities.  It should be possible to run multiple scenarios concurrently without them interfering with one another and resources should be cleaned up after the run.

# Success Criteria

- The stress tests have been integrated into the repository under the `Azure.Messaging.EventHubs` package, in a dedicated `stress` folder.

- The stress tests can be run locally with reasonable configuration/effort needed.

- The stress tests can be deployed to AKS, using the common platform elements.

- Stress tests have their own ReadMe that details how to run different scenarios, highlighting local runs.

# Reference and Resources

- [Getting Started guide](https://github.com/Azure/azure-sdk-tools/tree/master/tools/stress-cluster/chaos)
- [Richard's work for JS](https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/servicebus/service-bus/test/stress)
- [Event Hubs metrics discussion](https://github.com/Azure/azure-sdk/issues/3200)
- [Event Hubs quick-n-dirty stress harness](https://github.com/kinelski/EventHubStressTest)
- [Stress wiki](https://aka.ms/azsdk/stress) **(Microsoft Internal)**
- Teams channel for stress  **(Microsoft Internal)**

m-redding:
The changes I am including here are mostly centered around adding in processing and consuming tests to the existing stress test suite. I also did a lot of code cleanup and refactoring to make it easier to add more test scenarios in and to more effectively share code between tests.

#### Brief context around the processing and consuming tests:
The processor and consumer clients do not have a way to communicate directly with the producer client. In order to work around this, there is one producer per Event Hub partition which augments events with a publisher-defined index number so that the processors and consumers can detect out of order and missing events. The publisher also includes the hashed event body and a unique Id to help the consumers/processors understand if something goes wrong.

See #22436 


sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Configurations/TestConfiguration.cs
```diff
@@ -45,4 +45,46 @@ internal class TestConfiguration
...
+        return partitionCount;
+    }
+
+    /// <summary>
+    ///   Gets the list of partitions from the Event Hub being used for this test run.
+    /// </summary>
+    ///
+    /// <returns>An array holding the partition IDs associated with the Event Hub used for this test.>/returns>
+    ///
+    public async Task<string[]> GetEventHubPartitionKeysAsync()
```
jsquire, line 78:
```suggestion
    public async Task<string[]> GetEventHubPartitionsAsync()
```

We probably want to be cautious here; there's a big difference between a "partition key" and a "partition id" conceptually.

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Configurations/TestConfiguration.cs
```diff
@@ -45,4 +45,46 @@ internal class TestConfiguration
...
+    public async Task<int> GetEventHubPartitionCountAsync()
+    {
+        int partitionCount;
+
+        if (string.IsNullOrEmpty(EventHubsConnectionString) || string.IsNullOrEmpty(EventHub))
+        {
+            return 0;
+        }
+
+        await using (var producerClient = new EventHubProducerClient(EventHubsConnectionString, EventHub))
```
jsquire, line 64: Should we consider removing this method?   The value can be derived from the set of partition identifiers pulled by the other method.

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Infrastructure/EventTracking.cs
```diff
@@ -0,0 +1,246 @@
...
+
+    /// <summary>
+    ///   Gets the assigned partitions for the <see cref="PartitionPublisher"/>. This allows each partition publisher instance to publish
+    ///   to distinct and separate partitions. This is necessary so that missing and duplicate events can be detected by the <see cref="Role.Consumer"/>
+    ///   role without needing access to the <see cref="PartitionPublisher"/> instance.
+    /// </summary>
+    ///
+    /// <param name="partitionCount"></param>
+    /// <param name="roleIndex"></param>
+    /// <param name="partitionIds"></param>
```
jsquire, line 124: cheater!

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Infrastructure/EventTracking.cs
```diff
@@ -0,0 +1,246 @@
...
+    ///
+    /// <returns>The array of partitions that a given <see cref="PartitionPublisher"/> should send to.</returns>
```
jsquire, line 126:
```suggestion
    /// <returns>The set of partitions that a given <see cref="PartitionPublisher"/> should send to.</returns>
```

it's not actually an array...

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Infrastructure/EventTracking.cs
```diff
@@ -0,0 +1,246 @@
...
+    /// <param name="metrics"></param>
+    ///
+    /// <returns>The publisher assigned index number of the event that was just read, or -1 if this event was unkown or a duplicate.</returns>
+    ///
+    /// <remarks>
+    ///   One instance should be created for each processor or consumer role. This instance will be able to keep track of all the events the
+    ///   processor or consumer has received for each partition it is reading or processing from.
+    /// </remarks>
+    ///
+    private int _checkEvent(EventData eventData, string partitionReceivedFrom, Metrics metrics)
```
jsquire, line 179:
```suggestion
    private int CheckEvent(EventData eventData, string partitionReceivedFrom, Metrics metrics)
```

nit:  The underscore convention is only used for class-level variables.

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Infrastructure/EventTracking.cs
```diff
@@ -0,0 +1,246 @@
...
+    {
+        // Id Checks
+        var hasID = eventData.Properties.TryGetValue(IdPropertyName, out var eventIdProperty);
```
jsquire, line 182:
```suggestion
        var hasId = eventData.Properties.TryGetValue(IdPropertyName, out var eventIdProperty);
```

jsquire: C# conventions discourage consecutive capitals.

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Infrastructure/EventTracking.cs
```diff
@@ -0,0 +1,246 @@
...
+        var publisherSetPartitionProperty = partitionProperty?.ToString();;
+
+        if (partitionReceivedFrom != publisherSetPartitionProperty)
+        {
+            metrics.Client.GetMetric(Metrics.EventReceivedFromWrongPartition).TrackValue(1); //want an event here? or exception? This doesn't give much info
+        }
+
+        // Index Value checks
+        var hasIndex = eventData.Properties.TryGetValue(IndexNumberPropertyName, out var IndexProperty);
+        int.TryParse(IndexProperty.ToString(), out var indexNumber);
```
jsquire, line 208: We should consider what happens if this isn't present or isn't numeric here.    That will default to 0, which may track as valid.     Likewise, if `hasIndex` is `false` this line will throw.

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Infrastructure/EventTracking.cs
```diff
@@ -0,0 +1,246 @@
...
+    /// <param name="roleIndex"></param>
+    /// <param name="partitionIds"></param>
+    ///
+    /// <returns>The array of partitions that a given <see cref="PartitionPublisher"/> should send to.</returns>
+    ///
+    public static List<string> GetAssignedPartitions(int partitionCount, int roleIndex, string[] partitionIds, Role[] roles)
+    {
+        var roleList = new List<Role>(roles);
+
+        var numPublishers = roleList.Where(role => (role == Role.Publisher || role == Role.BufferedPublisher || role == Role.PartitionPublisher)).Count();
```
jsquire, line 132:
```suggestion
        var numPublishers = roleList.Count(role => (role == Role.Publisher || role == Role.BufferedPublisher || role == Role.PartitionPublisher));
```

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Infrastructure/EventTracking.cs
```diff
@@ -0,0 +1,246 @@
...
+        var thisPublisherIndex = (roleList.GetRange(0,roleIndex)).Where(role => (role == Role.Publisher || role == Role.BufferedPublisher || role == Role.PartitionPublisher)).Count();
```
jsquire, line 133:
```suggestion
        var thisPublisherIndex = (roleList.GetRange(0,roleIndex)).Count(role => (role == Role.Publisher || role == Role.BufferedPublisher || role == Role.PartitionPublisher));
```

jsquire: Wouldn't this produce a 1-based index?

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Infrastructure/EventTracking.cs
```diff
@@ -0,0 +1,246 @@
...
+            var eventProperties = new Dictionary<string,string>();
+            eventProperties.Add(Metrics.PublisherAssignedIndex, indexNumber.ToString());
+            eventProperties.Add(Metrics.EventBody, eventData.EventBody.ToString());
+            metrics.Client.TrackEvent(Metrics.MissingOrOutOfOrderEvent, eventProperties);
+        }
+
+        LastReadPartitionSequence.AddOrUpdate(partitionReceivedFrom, _ => indexNumber, (k,v) => Math.Max(v, indexNumber));
+
+        // Hashed event body checks
+        using (SHA256 sha256Hash = SHA256.Create())
```
jsquire, line 225: I wonder if we would be better off making this class disposable and creating the hasher at the class level.   That would prevent us from having to allocate/destroy it multiple times.

jsquire: for that matter... making it `static` and just letting it live for the duration of the test run (and not bothering to dispose) is probably ok too...

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Program.cs
```diff
@@ -79,118 +68,88 @@ private static async Task RunOptions(Options opts)
...
-
-            roleConfiguration.TestScenarioRoles.TryGetValue(RoleConfiguration.EventProcessorTest, out var roleList);
-            testScenarioTasks.Add(RunScenario(roleList, testConfiguration, roleConfiguration, metrics, opts.Role, RoleConfiguration.BurstBufferedProducerTest, cancellationSource.Token));
-        }
+        var metrics = new Metrics(appInsightsKey);
 
-        // Wait for all scenario runs to finish before returning.
-        await Task.WhenAll(testScenarioTasks).ConfigureAwait(false);
-    }
+        using var azureEventListener = new AzureEventSourceListener((args, level) => metrics.Client.TrackTrace($"EventWritten: {args.ToString()} Level: {level}."), EventLevel.Warning);
```
jsquire, line 74: Following up on thoughts in the `EventTracking` class - you could also consider creating the hasher here and passing it into the places it was used.

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Program.cs
```diff
@@ -79,118 +68,88 @@ private static async Task RunOptions(Options opts)
...
-                                           string roleString,
-                                           string testName,
-                                           CancellationToken cancellationToken)
-    {
-        var scenarioTasks = new List<Task>();
         try
         {
-            // This means the job index environment variable was not set, so run all roles for this scenario in parallel.
-            if (string.IsNullOrEmpty(roleString))
+            foreach (TestScenario t in testsToRun)
```
jsquire, line 77: If we could avoid single-letter variable names, that would be much appreciated.

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Program.cs
```diff
@@ -215,71 +174,40 @@ private static async Task RunOptions(Options opts)
...
-            var publisherConfiguration = new BufferedPublisherConfiguration();
-            publisherConfiguration.ProducerPublishingDelay = TimeSpan.FromMinutes(25);
-            var publisher = new BufferedPublisher(testConfiguration, publisherConfiguration, metrics, testName);
-            await publisher.Start(cancellationToken);
-        }
-
-        //return Task.CompletedTask;
-        await Task.Delay(TimeSpan.FromSeconds(1));
-    }
+        TestScenario.BufferedProducerTest => "BufferedProducerTest",
```
jsquire, line 186: Have we considered just using `ToString` here?  With the current mapping it seems like it would produce the same result.

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Program.cs
```diff
@@ -215,71 +174,40 @@ private static async Task RunOptions(Options opts)
...
     {
-        metrics.Client.GetMetric(Metrics.GenerationZeroCollections).TrackValue(GC.CollectionCount(0));
-        metrics.Client.GetMetric(Metrics.GenerationOneCollections).TrackValue(GC.CollectionCount(1));
-        metrics.Client.GetMetric(Metrics.GenerationTwoCollections).TrackValue(GC.CollectionCount(2));
-    }
+        "BufferedProducerTest" or "BuffProd"=> TestScenario.BufferedProducerTest,
+        "BurstBufferedProducerTest" or "BurstBuffProd" => TestScenario.BurstBufferedProducerTest,
+        "EventProducerTest" or "EventProd"=> TestScenario.EventProducerTest,
+        "ProcessorTest" or "Processor"=> TestScenario.ProcessorTest,
+        "ConsumerTest" or "Consumer" => TestScenario.ConsumerTest,
```
jsquire, line 208:
```suggestion
        "BufferedProducerTest" or "BuffProd" => TestScenario.BufferedProducerTest,
        "BurstBufferedProducerTest" or "BurstBuffProd" => TestScenario.BurstBufferedProducerTest,
        "EventProducerTest" or "EventProd" => TestScenario.EventProducerTest,
        "ProcessorTest" or "Processor" => TestScenario.ProcessorTest,
        "ConsumerTest" or "Consumer" => TestScenario.ConsumerTest,
```

jsquire (commented):
I only got about 1/2 way through but wanted to leave some feedback before I head out.   I'll loop back first thing tomorrow and finish up my first pass.

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Configurations/PartitionPublisherConfiguration.cs
```diff
@@ -0,0 +1,56 @@
...
+    ///   The percentage of generated events for each send that have large bodies in bytes
+    /// </summary>
+    ///
+    public int LargeMessageRandomFactorPercent = 50;
+
+    /// <summary>
+    ///   The amount of time to wait between enqueuing <see cref="PublishBatchSize"/> events.
+    /// </summary>
+    ///
+    public TimeSpan? ProducerPublishingDelay = TimeSpan.FromMilliseconds(4000);
```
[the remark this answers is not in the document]
m-redding: I will test this shortened again but initially I had made it long because otherwise the test got a little crazy in terms with throttling/cpu.

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Infrastructure/EventTracking.cs
m-redding: I could be wrong but I don't think so, GetRange (int index, int count) should look at everything up to but not including the current publisher when you do (0, roleIndex).

Like if you had [x,x,publisher,publisher] with roleIndex 2 that would do GetRange(index: 0, count:2) so that would look at x and x and return 0.

jsquire (commented):
Looks good!   

Most of my comments are just formatting nits.   There's a couple of things that we probably want to talk though, but they should be mostly small adjustments.

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Configurations/PartitionPublisherConfiguration.cs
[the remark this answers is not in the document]
jsquire: To be clear, I wasn't asking that we change the value - just that we initialize it in a more direct/clear way.    4000ms == 4s

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Roles/Consumer.cs
```diff
@@ -0,0 +1,128 @@
...
+    ///
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> instance to signal the request to cancel the operation.</param>
+    ///
+    public async Task StartAsync(CancellationToken cancellationToken)
+    {
+        var eventTracking = new EventTracking();
+        var consumerTasks = new Dictionary<string, Task>();
+        var partitionIds = await _testConfiguration.GetEventHubPartitionKeysAsync();
+
+        foreach (var partitionId in partitionIds)
```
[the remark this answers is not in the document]
jsquire: Discussed offline.  We're going to leave as-is for now and consider spawning into child pods if we need to scale out.

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Roles/PartitionPublisher.cs
```diff
@@ -0,0 +1,223 @@
...
+    {
+        var producersRunning = new List<Task>();
+
+        while (!cancellationToken.IsCancellationRequested)
+        {
+            using var backgroundCancellationSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+
+            foreach (var partition in _assignedPartitions)
+            {
+                producersRunning.Add(RunPartitionSpecificProducerAsync(partition, backgroundCancellationSource.Token));
```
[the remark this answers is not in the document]
jsquire: Discussed offline; we're going to leave things as-is for now and evaluate distributing the work to child pods if needed in the future.

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Scenarios/BufferedProducerTest.cs
```diff
@@ -0,0 +1,92 @@
...
+    /// <param name="cancellationToken">A <see cref="CancellationToken"/> instance to signal the request to cancel the operation.</param>
+    ///
+    public async Task RunRoleAsync(Role role, CancellationToken cancellationToken)
+    {
+        switch (role)
+        {
+            case Role.BufferedPublisher:
+                var publisherConfiguration = new BufferedPublisherConfiguration();
+                var publisher = new BufferedPublisher(_testConfiguration, publisherConfiguration, _metrics);
+                await publisher.StartAsync(cancellationToken).ConfigureAwait(false);
```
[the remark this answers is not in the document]
jsquire: Pragmatically, it probably won't be that drastic in our case, but it can sometimes take a bit for the first yield.   Not a blocker, but worth considering.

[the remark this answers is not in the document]
m-redding: Ah I see that makes sense. Thank you!

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Configurations/TestConfiguration.cs
```diff
@@ -45,4 +46,25 @@ internal class TestConfiguration
     /// </summary>
     ///
     public int DurationInHours = 150;
+
+    public SHA256 Sha256Hash = default;
```
jsquire, line 50: nit:  Doc comments.   :smile:


We should consider making this class disposable and having it own responsibility for cleaning up the hasher.

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Infrastructure/EventTracking.cs
```diff
@@ -0,0 +1,273 @@
...
+    ///
+    public static void AugmentEvent(EventData eventData,
+                                    int indexNumber,
+                                    string partition=null)
+    {
+        eventData.Properties.Add(IndexNumberPropertyName, indexNumber);
+        eventData.Properties.Add(PublishTimePropertyName, DateTimeOffset.UtcNow);
+        eventData.Properties.Add(IdPropertyName, Guid.NewGuid().ToString());
+
+        using (SHA256 sha256Hash = SHA256.Create())
```
jsquire, line 73: Should this be using the `TestConfiguration.Sha256Hash` member?   Maybe by passing it in like `CheckEvent`?

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Infrastructure/EventTracking.cs
```diff
@@ -0,0 +1,273 @@
...
+        return assignedPartitions;
+    }
+
+    /// <summary>
+    ///   Validates an event. This instance checks if the event has an id, if the event has already been seen before by this instance, that the
+    ///   intended partition was the one the event was received from, that the event was received in order, and that none were missed, and that
+    ///   the body is valid.
+    /// </summary>
+    ///
+    /// <param name="eventData"></param>
```
jsquire, line 175: Please add info for the `<param>` tags.

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Infrastructure/EventTracking.cs
```diff
@@ -0,0 +1,273 @@
...
+    ///   One instance should be created for each processor or consumer role. This instance will be able to keep track of all the events the
+    ///   processor or consumer has received for each partition it is reading or processing from.
+    /// </remarks>
+    ///
+    private static int CheckEvent(EventData eventData,
+                           SHA256 sha256Hash,
+                           string partitionReceivedFrom,
+                           Metrics metrics,
+                           ConcurrentDictionary<string, int> lastReadPartitionSequence,
+                           ConcurrentDictionary<string, byte> readEvents)
```
jsquire, line 191:
```suggestion
    private static int CheckEvent(EventData eventData,
                                  SHA256 sha256Hash,
                                  string partitionReceivedFrom,
                                  Metrics metrics,
                                  ConcurrentDictionary<string, int> lastReadPartitionSequence,
                                  ConcurrentDictionary<string, byte> readEvents)
```

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Infrastructure/Metrics.cs
```diff
@@ -199,6 +155,82 @@ internal class Metrics
     ///
     public const string EventsProcessed = "EventsProcessed";
 
+    /// <summary>
+    ///   This is the metric name used to collect metrics on the number of duplicate events processed by the processor.
+    /// </summary>
+    ///
+    public const string DuplicateEventsDiscarded = "DuplicateEventsDiscarded";
```
jsquire, line 162: Not important, but if you wanted to cheat and make sure that any updates to names are captured as they happen, you could do:

```csharp
public static readonly string DuplicateEventsDiscarded = nameof(DuplicateEventsDiscarded);
```

That said, I don't think it's worth doing here since we aren't likely to refactor often.   Just thought I'd mention it for future considerations.

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Roles/Consumer.cs
```diff
@@ -0,0 +1,141 @@
...
+    /// </summary>
+    ////
+    /// <param name="partitionId">The Id of the partition to read events from.</param>
+    /// <param name="eventTracking">The <see cref="EventTracking"/> instance used to validate events.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> instance to signal the request to cancel the operation.</param>
+    ///
+    private async Task ConsumePartitionAsync(string partitionId,
+                                             CancellationToken cancellationToken)
+    {
+        using var sha256Hash = SHA256.Create();
```
jsquire, line 95: Should we be using the `TestConfiguration` hasher here?

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Scenarios/BufferedProducerTest.cs
```diff
@@ -0,0 +1,94 @@
...
+
+    /// <summary>
+    ///   Creates a role instance and runs that role.
+    /// </summary>
+    ///
+    /// <param name="role">The <see cref="Role"/> to run.</param>
+    /// <param name="cancellationToken">A <see cref="CancellationToken"/> instance to signal the request to cancel the operation.</param>
+    ///
+    private Task RunRoleAsync(Role role,
+                                   CancellationToken cancellationToken)
```
jsquire, line 81:
```suggestion
    private Task RunRoleAsync(Role role,
                              CancellationToken cancellationToken)
```
```suggestion
    private Task RunRoleAsync(Role role,
                                   CancellationToken cancellationToken)
```

jsquire (commented):
Looking great!   We're down to a couple of minor nits and formatting suggestions.

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Scenarios/BufferedProducerTest.cs
jsquire: Not sure if there's a push pending, but this is still off...

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Scenarios/TestParameters.cs
```diff
@@ -0,0 +1,85 @@
...
+        {
+            return (await producerClient.GetPartitionIdsAsync().ConfigureAwait(false));
+        }
+    }
+
+    public void Dispose()
+    {
+        if (!_disposed)
+        {
+            Sha256Hash.Dispose();
```
jsquire, line 82:
```suggestion
            Sha256Hash.Dispose();
            _disposed
```

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Scenarios/TestParameters.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        await using (var producerClient = new EventHubProducerClient(EventHubsConnectionString, EventHub))
+        {
+            return (await producerClient.GetPartitionIdsAsync().ConfigureAwait(false));
+        }
+    }
+
+    public void Dispose()
+    {
+        if (!_disposed)
```
jsquire, line 80: No need to check a flag; `Dispose` is idempotent and safe to call multiple times.   I've no objection to keeping it - but if you do, you should probably set it to `true` at the end of the `if` block.   😄

sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Scenarios/TestParameters.cs
```diff
@@ -0,0 +1,85 @@
...
+using Azure.Messaging.EventHubs.Producer;
+using System.Security.Cryptography;
+
+namespace Azure.Messaging.EventHubs.Stress;
+
+/// <summary>
+///   The set of configurations to use for a test scenario run.
+/// </summary>
+///
+public class TestParameters : IDisposable
```
jsquire, line 15: Nice.... I love the rename!

jsquire (approved):

PR #29448, opened 2022-06-23
Upload crash dumps for test crashes on .NET 6

JoshLove-msft:
Upload crash dumps on .NET 6, e.g. https://dev.azure.com/azure-sdk/public/_build/results?buildId=1668526&view=artifacts&pathAsName=false&type=publishedArtifacts

### DIFF
--- a/eng/pipelines/templates/jobs/ci.tests.yml
+++ b/eng/pipelines/templates/jobs/ci.tests.yml
@@ -65,7 +65,7 @@ jobs:
           NuGetCacheKey: Test
       - script: >-
           dotnet test eng/service.proj --filter "(TestCategory!=Manually) & (TestCategory!=Live)" --framework $(TestTargetFramework)
-          --logger "trx;LogFileName=$(TestTargetFramework).trx" --logger:"console;verbosity=normal"
+          --logger "trx;LogFileName=$(TestTargetFramework).trx" --logger:"console;verbosity=normal" --blame-crash
           /p:SDKType=${{ parameters.SDKType }}
           /p:ServiceDirectory=${{ parameters.ServiceToTest }}
           /p:IncludeSrc=false /p:IncludeSamples=false /p:IncludePerf=false /p:IncludeStress=false
--- a/eng/pipelines/templates/jobs/ci.tests.yml
+++ b/eng/pipelines/templates/jobs/ci.tests.yml
@@ -64,7 +64,7 @@ jobs:
         parameters:
           NuGetCacheKey: Test
       - script: >-
-          dotnet tool install --global dotnet-dump
+          dotnet tool install dotnet-dump --global
         displayName: "Install dotnet-dump"
       - script: >-
           dotnet test eng/service.proj --filter "(TestCategory!=Manually) & (TestCategory!=Live)" --framework $(TestTargetFramework)
--- a/eng/pipelines/templates/jobs/ci.tests.yml
+++ b/eng/pipelines/templates/jobs/ci.tests.yml
@@ -64,6 +64,9 @@ jobs:
         parameters:
           NuGetCacheKey: Test
       - script: >-
+          dotnet tool install --global dotnet-dump
+        displayName: "Install dotnet-dump"
+      - script: >-
           dotnet test eng/service.proj --filter "(TestCategory!=Manually) & (TestCategory!=Live)" --framework $(TestTargetFramework)
           --logger "trx;LogFileName=$(TestTargetFramework).trx" --logger:"console;verbosity=normal"
           --blame-crash --blame-crash-dump-type full
--- a/eng/pipelines/templates/jobs/ci.tests.yml
+++ b/eng/pipelines/templates/jobs/ci.tests.yml
@@ -66,7 +66,7 @@ jobs:
       - script: >-
           dotnet test eng/service.proj --filter "(TestCategory!=Manually) & (TestCategory!=Live)" --framework $(TestTargetFramework)
           --logger "trx;LogFileName=$(TestTargetFramework).trx" --logger:"console;verbosity=normal"
-          --blame-crash-dump-type full
+          --blame-crash-dump-type full --blame-hang-dump-type full
           /p:SDKType=${{ parameters.SDKType }}
           /p:ServiceDirectory=${{ parameters.ServiceToTest }}
           /p:IncludeSrc=false /p:IncludeSamples=false /p:IncludePerf=false /p:IncludeStress=false
--- a/eng/pipelines/templates/jobs/ci.tests.yml
+++ b/eng/pipelines/templates/jobs/ci.tests.yml
@@ -64,12 +64,9 @@ jobs:
         parameters:
           NuGetCacheKey: Test
       - script: >-
-          dotnet tool install dotnet-dump --global
-        displayName: "Install dotnet-dump"
-      - script: >-
           dotnet test eng/service.proj --filter "(TestCategory!=Manually) & (TestCategory!=Live)" --framework $(TestTargetFramework)
           --logger "trx;LogFileName=$(TestTargetFramework).trx" --logger:"console;verbosity=normal"
-          --blame-crash --blame-crash-dump-type full
+          --blame-crash-dump-type full
           /p:SDKType=${{ parameters.SDKType }}
           /p:ServiceDirectory=${{ parameters.ServiceToTest }}
           /p:IncludeSrc=false /p:IncludeSamples=false /p:IncludePerf=false /p:IncludeStress=false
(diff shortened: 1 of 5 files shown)


eng/pipelines/templates/jobs/ci.tests.yml
```diff
@@ -65,7 +65,7 @@ jobs:
           NuGetCacheKey: Test
       - script: >-
           dotnet test eng/service.proj --filter "(TestCategory!=Manually) & (TestCategory!=Live)" --framework $(TestTargetFramework)
-          --logger "trx;LogFileName=$(TestTargetFramework).trx" --logger:"console;verbosity=normal"
+          --logger "trx;LogFileName=$(TestTargetFramework).trx" --logger:"console;verbosity=normal" --blame-crash
```
weshaggard, line 68: If this is helpful we might want to add it to live test command as well. Also nit but I'd suggest putting it on its own line.

christothes, line 68: Should we also specify `--blame-crash-dump-type` to get a full dump?

JoshLove-msft: Not sure. The docs indicate that blame-crash will already collect the dump.

christothes: My guess is that it's a mini dump by default. We should get a full if we can.

JoshLove-msft: I guess we will see!

JoshLove-msft: Wouldn't we want to use a minidump if it proves sufficient?

christothes: I'd just rather get all the info the first time, just in case :) The test logs already give us a basic stack, but I'm guessing we'll want to poke around a bit more.

eng/pipelines/templates/steps/install-dotnet.yml
```diff
@@ -17,15 +17,6 @@ steps:
     inputs:
       useGlobalJson: true
       performMultiLevelLookup: true
-  - task: UseDotNet@2
```
JoshLove-msft, line 20: Trying this because I couldn't get the dump to generate. From the docs, it seems that this feature requires .NET 5.0 SDK. We install both 6.0 and 3.1, but I'm not sure what the behavior is when both are installed side by side.

The blame-crash flag did at least generate a file that shows the offending test. Each time so far they have been different recorded tests from the Tables Extensions test project.

JoshLove-msft: Didn't help.

christothes: I'm still wondering if this is because it is linux.

JoshLove-msft: It seems like it should be possible. There may be other undocumented args that need to be set - https://github.com/microsoft/vstest/issues/3085#issuecomment-961085082

JoshLove-msft: I think we may need to install dotnet-dump as well:
dotnet tool install --global dotnet-dump

eng/pipelines/templates/jobs/ci.tests.yml
```diff
@@ -63,9 +63,14 @@ jobs:
       - template: /eng/pipelines/templates/steps/install-dotnet.yml
         parameters:
           NuGetCacheKey: Test
+      - script: >-
+          dotnet tool install dotnet-dump --global
```
JoshLove-msft, line 67: Trying to install this tool, which is apparently required to use the --blame-test-crash flag for dotnet test.
I keep getting this message:
"Specify which project file to use because this '/mnt/vss/_work/1/s' contains more than one project file."

JoshLove-msft: adding working directory seems to have fixed this.

eng/pipelines/templates/jobs/ci.tests.yml
```diff
@@ -66,6 +66,7 @@ jobs:
       - script: >-
           dotnet test eng/service.proj --filter "(TestCategory!=Manually) & (TestCategory!=Live)" --framework $(TestTargetFramework)
           --logger "trx;LogFileName=$(TestTargetFramework).trx" --logger:"console;verbosity=normal"
+          --blame-crash-dump-type full
```
JoshLove-msft, line 69: I couldn't get a crash dump to generate successfully on Linux. I'm unsure if it will work on other OSes, but at least this flag generates a sequence file when there is a crash which makes it easier to pinpoint the test that was running at the time of the crash.

sdk/tables/Microsoft.Azure.WebJobs.Extensions.Tables/tests/TablesLiveTestBase.cs
```diff
@@ -100,7 +100,7 @@ protected async Task<T> CallAsync<T>(string methodName = null, object arguments
         protected async Task<object> CallAsync(Type funcType, string methodName = null, object arguments = null, Action<HostBuilder> configure = null)
         {
             var instance = Activator.CreateInstance(funcType);
-            var (host, jobHost) = CreateHost(funcType, configure, instance);
+            using var jobHost = CreateHost(funcType, configure, instance);
```
JoshLove-msft, line 103: I eventually reproed this on WSL and was able to collect a crash dump. I saw there was an OOM and StackOverflow exception. I didn't pinpoint the exact code, but using that clue I looked more closely at the code and realized we were not disposing the job host.

christothes: Do we want to keep the dump changes, or just the fix? If not, perhaps we could put these dump changes behind a variable config so that it's easy to turn on as needed.

JoshLove-msft: I think it is worth keeping the dump collection on. Do you have any concerns with it?
/cc @weshaggard

eng/pipelines/templates/jobs/ci.tests.yml
```diff
@@ -83,6 +84,21 @@ jobs:
...
           testResultsFormat: "VSTest"
           mergeTestResults: true
+      - pwsh: |
+          $dump = Get-ChildItem -Path $(Build.SourcesDirectory) -Include *.dmp -Recurse |Select -First 1
+          if ($dump)
+          {
+            Write-Host "##vso[task.setvariable variable=uploadDump]true"
+            Write-Host "##vso[task.setvariable variable=dumpPath]$dump"
+          }
+        condition: failed()
```
weshaggard, line 94: Will this always be a failure case? I know some of the ones we were seeing failed but we didn't get a failed step.

sdk/tables/Microsoft.Azure.WebJobs.Extensions.Tables/tests/TablesLiveTestBase.cs
weshaggard: I like keeping the dump code as well.

eng/pipelines/templates/jobs/ci.tests.yml
JoshLove-msft: > Also do we know how largest these dmp files are? Would it make sense to zip them up? For reference the java build is zipping them but I'm not sure if that is necessary or not. https://github.com/Azure/azure-sdk-for-java/blob/main/eng/pipelines/templates/steps/retain-heap-dump-hprofs.yml

They can go up to 1GB.

JoshLove-msft: > It is probably worth refactoring this into a template similarly so we can share this and we also need a unique artifact name if we ever need to retry this on a failed build.

Do you have a suggestion for a variable that can be used that would be unique to each retry?

weshaggard: See the example at https://github.com/Azure/azure-sdk-for-java/blob/main/eng/pipelines/templates/steps/retain-heap-dump-hprofs.yml#L16 That takes retries into account (System.JobAttempt)

christothes (approved):

eng/pipelines/templates/jobs/ci.tests.yml
JoshLove-msft: Ah wow.. this would have saved me a bunch of time 😆

JoshLove-msft: It might even be worth putting this in common - I expect the pattern to always be "given an extension, upload all files of that extension".

christothes: If we go down that road, why not go ahead and make it an arbitrary Filter pattern :)

eng/pipelines/templates/steps/upload-dumps.yml
```diff
@@ -0,0 +1,16 @@
...
+        $fileName = $dumpFile.Name
+        Move-Item -Path $fileFullName -Destination $(Build.ArtifactStagingDirectory)/dumps/$fileName -ErrorAction SilentlyContinue
+      }
+      [System.IO.Compression.ZipFile]::CreateFromDirectory("$(Build.ArtifactStagingDirectory)/dumps","$(Build.ArtifactStagingDirectory)/dumps.zip")
+    condition: always()
+    displayName: Copy dumps to staging directory
+  - publish: $(Build.ArtifactStagingDirectory)/dumps.zip
+    condition: always()
+    displayName: Upload any dumps
+    artifact: "Crash Dump - $(OSName) $(TestTargetFramework) - $(System.JobName)-$(System.JobAttempt)"
```
weshaggard, line 16: I think OSName and TestTargetFramework are already covered by the JobName.

weshaggard (approved):
Looks reasonable. Can you please add a test that forces a crash just so we can see this actually work?

eng/pipelines/templates/jobs/ci.tests.yml
JoshLove-msft: Thanks @alzimmermsft!

PR #29456, opened 2022-06-24
Fix regional authority for `ClientCertificateCredential`

### LINKED ISSUE #29454: [BUG] Regional Authority is not used with latest version of `ClientCertificateCredential`
### Library name and version

Azure.Identity 1.6.0

### Describe the bug

The regional authority is not being used by the `ClientCertificateCredential` constructor even when it is passed in with `options` since it no longer passes it onto the `MsalConfidentialClient` ctor: [link](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/identity/Azure.Identity/src/Credentials/ClientCertificateCredential.cs#L162)

We're trying to use the regional endpoints and this was working fine in the older versions:
- 1.5.0 - :white_check_mark: [Works fine](https://github.com/Azure/azure-sdk-for-net/blob/Azure.Identity_1.5.0/sdk/identity/Azure.Identity/src/ClientCertificateCredential.cs#L163)
- 1.6.0-beta1 - :white_check_mark: [Works fine](https://github.com/Azure/azure-sdk-for-net/blob/Azure.Identity_1.6.0-beta.1/sdk/identity/Azure.Identity/src/ClientCertificateCredential.cs#L163)
- 1.6.0 - :x: [Broken](https://github.com/Azure/azure-sdk-for-net/blob/Azure.Identity_1.6.0/sdk/identity/Azure.Identity/src/Credentials/ClientCertificateCredential.cs#L162)

It accidentally got [removed here](https://github.com/Azure/azure-sdk-for-net/commit/2a009f1f5414d44bb29b200d199357128e9e83c5#diff-dd3f1b93fd92b1250e2d16929dcf03fc20898ef67fbb604a0d9d10b567c33c76) during a refactor while the other credentials that support Regional Authority i.e `ClientSecretCredential` and `OnBehalfOfCredential` still have support for RA.

Our service can't rollback to the previous versions because we need to be on 1.6.0 which has the fix for #27679 (OBOCredentials not passing x5c)

### Expected behavior

Regional Authority should work with `ClientCertificateCredential` just like it does with `ClientSecretCredential` and `OnBehalfOfCredential` types.

### Actual behavior

Regional Authority doesn't work and calls are sent to global AAD authority instead.

### Reproduction Steps

1. Create a new ClientCertificateCredential and try to pass a valid Regional Authority
2. Calls go out to global AAD instead of regional authority

### Environment

_No response_

degant:
Fixing support for regional authority in `ClientCertificateCredential` along with necessary unit tests.

Fixes #29454

### DIFF
--- a/sdk/identity/Azure.Identity/src/Credentials/ClientCertificateCredential.cs
+++ b/sdk/identity/Azure.Identity/src/Credentials/ClientCertificateCredential.cs
@@ -159,7 +159,8 @@ namespace Azure.Identity
                          clientId,
                          certificateProvider,
                          certCredOptions?.SendCertificateChain ?? false,
-                         options);
+                         options,
+                         certCredOptions?.RegionalAuthority);
         }
 
         /// <summary>
--- a/sdk/identity/Azure.Identity/tests/ClientCertificateCredentialTests.cs
+++ b/sdk/identity/Azure.Identity/tests/ClientCertificateCredentialTests.cs
@@ -196,7 +196,7 @@ namespace Azure.Identity.Tests
         {
             TestSetup();
             var _transport = Createx5cValidatingTransport(sendCertChain);
-            var _pipeline = new HttpPipeline(_transport, new[] {new BearerTokenAuthenticationPolicy(new MockCredential(), "scope")});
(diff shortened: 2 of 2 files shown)


snakefoot:
Superseeded by #29612 ?

degant:
@snakefoot yes that is already merged. We also verified that the fix is working as expected so I'm going to abandon this

PR #29483, opened 2022-06-24
Fix build break

JoshLove-msft:


heaths (approved):
To fix the tests and unblock other PRs this is fine, but @m-nash this also falls in line with your recent comment in APIView about using "KeyVault" as opposed to "Vault". That said, I also appreciate we want to do minimal changes for DPGs to scale.

PR #29486, opened 2022-06-25
[Service Linker] Polish for archboard review

ArthurMa1978:


sdk/servicelinker/Azure.ResourceManager.ServiceLinker/src/autorest.md
```diff
@@ -8,18 +8,40 @@ azure-arm: true
...
+  ConfluentSchemaRegistry: ConfluentSchemaRegistryInfo
+  AzureResourcePropertiesBase: AzureResourceBaseProperties
+  SecretInfoBase: SecretBaseInfo
+  ValueSecretInfo: RawValueSecretInfo
+  AuthInfoBase: AuthBaseInfo
+  AuthType: LinkerAuthType
+  SecretType: LinkerSecretType
+  SecretStore: LinkerSecretStore
+  ClientType: LinkerClientType
+  ValidationResultItem: ValidationResultItemData
```
Yao725, line 32: Why did we use the `Data` suffix here? It looks like that this model doesn't inherits from any `Data` type in ResourcceManager.

Yao725 (approved):

sdk/servicelinker/Azure.ResourceManager.ServiceLinker/src/autorest.md
ArthurMa1978: In the `KeyVault` SDK Ted considered to avoid use `Item` suffix, so append `Data`, so here do the same rename

PR #29497, opened 2022-06-27
clientside encryption perf with existing tests

jaschrep-msft:
Adds support for client-side encryption options to be set on clients.

- Separates transfer options and client options into interfaces.
- Current overarching concrete options class inherits both the above interfaces, renamed class to scope it to SDK-partitioned transfers.
- Blobs ServiceTest does not require options implement client options interface but will honor it if it does.
- Created options folder based on namespacing of existing classes.


mikeharder (approved):

sdk/storage/Azure.Storage.Blobs/perf/Azure.Storage.Blobs.Perf/Options/PartitionedTransferOptions.cs
```diff
@@ -0,0 +1,86 @@
...
+            get => _maximumConcurrency;
+            set
+            {
+                _maximumConcurrency = value;
+                UpdateStorageTransferOptions();
+            }
+        }
+
+        [Option("clientEncryptionVersion")]
+        public string EncryptionVersionString { get; set; }
```
kasobol-msft, line 39: I'd check if the CommandLineParser can handle enums.
See https://github.com/commandlineparser/commandline/wiki/Enum-Flags

kasobol-msft (approved):

sdk/storage/Azure.Storage.Blobs/perf/Azure.Storage.Blobs.Perf/Options/IBlobClientOptionsOptions.cs
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Azure.Storage.Blobs.Perf.Options
+{
+    public interface IBlobClientOptionsOptions
```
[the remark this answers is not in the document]
jaschrep-msft: Seems good. Was following existing pattern but this definitely feels better.

sdk/storage/Azure.Storage.Blobs/perf/Azure.Storage.Blobs.Perf/Options/PartitionedTransferOptions.cs
jaschrep-msft: It seems this is only possible if I provide the string name of the enum value. However, those values in here are `V1_0` and `V2_0`, which are not the values we want to provide (`1.0` and `2.0`). So I have to do the conversion myself it seems.

The actual storage library has internal extension methods that do the conversion.

PR #29501, opened 2022-06-27
Pass in Scope to the mgmt credscan step

weshaggard:
mgmt doesn't work against ServiceDirectory it uses Scope so we need to pass that in instead.

### DIFF
--- a/eng/pipelines/templates/jobs/ci.mgmt.yml
+++ b/eng/pipelines/templates/jobs/ci.mgmt.yml
@@ -116,4 +116,3 @@ jobs:
       - template: /eng/common/pipelines/templates/steps/credscan.yml
         parameters:
           BaselineFilePath: $(Build.sourcesdirectory)/eng/dotnet.gdnbaselines
-          ServiceDirectory: ${{ parameters.Scope }}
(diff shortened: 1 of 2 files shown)


JoshLove-msft:
> Azure Pipelines could not run because the pipeline triggers exclude this branch/path.

This highlights the problem with changing this ci file - it requires manual testing.

weshaggard:
>This highlights the problem with changing this ci file - it requires manual testing.

Yep just added the template file to the set of triggers for the pipeline.

JoshLove-msft (approved):

eng/pipelines/templates/jobs/ci.mgmt.yml
```diff
@@ -116,4 +116,4 @@ jobs:
       - template: /eng/common/pipelines/templates/steps/credscan.yml
         parameters:
           BaselineFilePath: $(Build.sourcesdirectory)/eng/dotnet.gdnbaselines
-          ServiceDirectory: ${{ parameters.ServiceDirectory }}
\ No newline at end of file
+          ServiceDirectory: ${{ parameters.Scope }}
```
weshaggard, line 119:
```suggestion
```

PR #29523, opened 2022-06-28
Refactor Azure.Storage.Blobs.Perf

jaschrep-msft:
Refactoring of the test definition inheritance structure. Gives base classes responsibilities that align with their naming. Has concrete test classes inherit from the semantically sensible class.

### DIFF
--- a/sdk/storage/Azure.Storage.Blobs/perf/Azure.Storage.Blobs.Perf/Infrastructure/BlobTest.cs
+++ b/sdk/storage/Azure.Storage.Blobs/perf/Azure.Storage.Blobs.Perf/Infrastructure/BlobTest.cs
@@ -2,21 +2,55 @@
 // Licensed under the MIT License.
 
 using System;
-using Azure.Storage.Blobs.Specialized;
+using System.IO;
+using System.Threading.Tasks;
 using Azure.Test.Perf;
 
 namespace Azure.Storage.Blobs.Perf
 {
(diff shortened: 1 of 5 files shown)


mikeharder:
> @mikeharder Is storage CI already testing changes I make here? Is there a test suite somewhere I can run manually to validate these changes?

The "net - storage - ci" pipeline should compile the perf test project to check for compile errors.  But this doesn't seem to be working.  Tracking issue: https://github.com/Azure/azure-sdk-for-net/issues/29543

You can run the actual perf tests using command `/azp run net - storage-blob - perf`.  This will run all the perf tests which takes about 2 hours.  You can also manually trigger the `net - storage-blob - perf` pipeline using the DevOps UI, and select a subset of tests which will run much faster.  For example, you could just run one iteration of the the 10KB upload and download tests, which only takes a few minutes.

Another issue I found is the perf pipelines may "pass" even though the perf tests cannot be run.  Tracking issue: https://github.com/Azure/azure-sdk-tools/issues/3518

amnguye (approved):

mikeharder (approved):
I agree this is better than the previous code.  Your perf run got some unexpectedly low results for 10KB and 10MB upload.  I'm almost certain this is noise and unrelated to your PR, but I am running those tests again to make certain:

https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1675478&view=results

mikeharder:
Re-run of upload tests looks good, so this is ready to merge by me.

PR #29533, opened 2022-06-28
[GitHub] Add priority to Core feature request

jsquire:
# Summary

The focus of these changes is to add a priority field into the Azure.Core feature request template.

### DIFF
--- a/.github/ISSUE_TEMPLATE/04_core_feature_request.yml
+++ b/.github/ISSUE_TEMPLATE/04_core_feature_request.yml
@@ -2,7 +2,7 @@ name: Core Feature Request
 description: Create a feature request for Azure.Core
 title: "[Core] "
 labels: ["Azure.Core", "feature-request"]
-assignees: ["christothes","annelo-msft","JoshLove-msft"]
+assignees: ["christothes","annelo-msft","JoshLove-msft", "jsquire"]
 
 body:
   - type: dropdown
@@ -43,6 +43,16 @@ body:
         ```
     validations:
       required: true
+  - type: dropdown
+    id: priority
+    attributes:
+      label: Priority
+      description: What is the priority of this feature with respect to your intended use?
+      options:
+        - "P0 (blocking)"
+        - "P1 (needed-urgent)"
+        - "P2 (needed)"
+        - "P3 (nice-to-have)"
   - type: textarea
     id: constraints
     attributes:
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/04_core_feature_request.yml
```diff
@@ -2,7 +2,7 @@ name: Core Feature Request
 description: Create a feature request for Azure.Core
 title: "[Core] "
 labels: ["Azure.Core", "feature-request"]
```
jsquire, line 4:
```suggestion
labels: ["Azure.Core", "Client", "feature-request"]
```

.github/ISSUE_TEMPLATE/04_core_feature_request.yml
```diff
@@ -43,6 +43,16 @@ body:
...
     validations:
       required: true
+  - type: dropdown
+    id: priority
+    attributes:
+      label: Priority
+      description: What is the priority of this feature with respect to your intended use?
+      options:
+        - "P0 (blocking)"
+        - "P1 (needed-urgent)"
```
JoshLove-msft, line 53: What is the difference between needed-urgent and needed? The urgency is probably represented by the time constraints section.

JoshLove-msft (approved):

.github/ISSUE_TEMPLATE/04_core_feature_request.yml
christothes: That was my same thought - an aggressive time constraint seems equivalent to "urgent" and blocking would also be expressed in the dependencies section. This field does make it a bit more explicit, but my only concern is that it could be a shortcut to expressing this in the next section.

jsquire: In a plan language:

  - This blocks me; I cannot make forward progress at all without it
  - I need this as soon as possible
  - I need this, but have some timing flexibility
  - This would be cool to have at some point

My intent was to allow requesters to be explicit about their priorities rather than us inferring them; these are the priorities that I'd like to align across Core and CodeGen internally, even if we're not including them here.

jsquire: To be clear, I'm fine dropping the template change if we collectively feel like it's not helpful or redundant.

christothes: I'm fine with it as long as it's not a substitute for the long form in the next section.

jsquire:
Going to close this out for now; we'll continue to set priority by discretion after evaluating the form.

PR #29550, opened 2022-06-29
[EventHubs] Remove the dependency on storage and update tests

Yao725:
This PR removed the project reference of Azure.ResourceManage.Resources and Azure.ResourceManage.Storage, which means that we don't need to include the `sdk/resources` and `sdk/storage` in the pr path of ci.mgmt.yml.

### DIFF
--- a/sdk/eventhub/Azure.ResourceManager.EventHubs/tests/Azure.ResourceManager.EventHubs.Tests.csproj
+++ b/sdk/eventhub/Azure.ResourceManager.EventHubs/tests/Azure.ResourceManager.EventHubs.Tests.csproj
@@ -1,11 +1,10 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <ItemGroup>
     <ProjectReference Include="..\src\Azure.ResourceManager.EventHubs.csproj" />
-    <ProjectReference Include="..\..\..\storage\Azure.ResourceManager.Storage\src\Azure.ResourceManager.Storage.csproj" />
-    <ProjectReference Include="..\..\..\resources\Azure.ResourceManager.Resources\src\Azure.ResourceManager.Resources.csproj" />
     <ProjectReference Include="..\..\..\network\Azure.ResourceManager.Network\src\Azure.ResourceManager.Network.csproj" />
   </ItemGroup>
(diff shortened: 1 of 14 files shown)


sdk/eventhub/Azure.ResourceManager.EventHubs/tests/Tests/ConsumerGroupTests.cs
```diff
@@ -36,22 +36,6 @@ public async Task CreateNamespaceAndGetEventhubCollection()
             _consumerGroupCollection = _eventHub.GetEventHubsConsumerGroups();
         }
 
-        [TearDown]
```
Yao725, line 39: Removed all this kind of tear down cleanup code as we already have the resource group cleanup setting.

ArthurMa1978 (approved):

sdk/eventhub/Azure.ResourceManager.EventHubs/tests/Azure.ResourceManager.EventHubs.Tests.csproj
```diff
@@ -1,8 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <ItemGroup>
     <ProjectReference Include="..\src\Azure.ResourceManager.EventHubs.csproj" />
-    <ProjectReference Include="..\..\..\storage\Azure.ResourceManager.Storage\src\Azure.ResourceManager.Storage.csproj" />
-    <ProjectReference Include="..\..\..\resources\Azure.ResourceManager.Resources\src\Azure.ResourceManager.Resources.csproj" />
     <ProjectReference Include="..\..\..\network\Azure.ResourceManager.Network\src\Azure.ResourceManager.Network.csproj" />
```
weshaggard, line 4: Is there plans to remove the networking dependency as well? As was discussed we generally should avoid having any project references to packages in other service directories longer term.

Yao725: Yes, we are working on releasing the stable version of network. Once this is done, the project reference will be removed.

PR #29558, opened 2022-06-29
Update baselines

benbp:
Updating credscan baselines file to correspond with recent updates to messaging library recordings.


JoshLove-msft (approved):

jsquire:
@benbp:  Not sure if you were trying to validate the management libraries against the baseline, but they're in a different set of pipelines:

- `net - eventhub - mgmt`
- `net - servicebus - mgmt`

Want me to kick those?

Issue #24043, opened 2021-09-16
When can we expect to set normalizers using the SimpleFieldAttribute properties?


dloukola:
**Query/Question**
When can we expect to set normalizers using the SimpleFieldAttribute properties?
I see that normalizers are in Azure.Search.Documents 11.3.0-beta.1
https://github.com/Azure/azure-sdk-for-net/blob/Azure.Search.Documents_11.3.0/sdk/search/Azure.Search.Documents/CHANGELOG.md
But there doesn't seem to be support for adding the normalizer via an attribute yet.
I expect that a normalizer would be set like any other analyzer. example:
```
[SimpleField(NormalizerName = LexicalNormalizerName.Standard)]
[SearchableField(NormalizerName = LexicalNormalizerName.Standard)]
```
It would be really useful to get a quick answer on IF this is in the backlog and if it is in the backlog, WHEN it would be expected. Otherwise, we'll have to write some workaround for this. Thanks!

**Environment:**
 - Name and version of the Library package used: Azure.Search.Documents 11.3.0-beta.1
 - Hosting platform or OS and .NET runtime version: Windows 10
 - IDE and version : Visual Studio 2019 16.10.4

dloukola:
Any idea if this will be worked on within the next couple of months? otherwise I'll need to figure something else out.

Mohit-Chakraborty:
Hi, We will look into adding this support for our December release of the Search package.
Does that timeline work for you?

dloukola:
Hi @Mohit-Chakraborty , yes Dec works for us. Thank you for your support!

dloukola:
Hi @Mohit-Chakraborty is there any update on this? It's now the end of January and we don't see any normalizer available in the production or even preview versions of the Azure.Search.Documents package.

Mohit-Chakraborty:
We did not release an update in December, team has been focusing on improving the documentation and test coverage. We should have it ready for the March release.

dloukola:
Thank you for the quick reply. We'll have to put a workaround in, but we look forward to the March release.

dloukola:
Should we be able to use normalizers directly without the attributes today? My co-worker tried the solution as shown here:
https://docs.microsoft.com/en-us/rest/api/searchservice/preview-api/test-analyzer
with the REST API and the preview version of the search SDK and neither worked.
Do we have to upgrade our search instance somehow or something?
/cc: @Mohit-Chakraborty

[...]

PR #29568, opened 2022-06-30
Add Support for Normalizer in SimpleField and SearchableField

ShivangiReja:
Adds support for normalizers in `SimpleField` and `SearchableField`

Issue: https://github.com/Azure/azure-sdk-for-net/issues/24043

### DIFF
--- a/sdk/search/Azure.Search.Documents/src/Indexes/Models/LexicalNormalizerName.cs
+++ b/sdk/search/Azure.Search.Documents/src/Indexes/Models/LexicalNormalizerName.cs
@@ -1,0 +1,28 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Azure.Search.Documents.Indexes.Models
+{
+    public readonly partial struct LexicalNormalizerName
+    {
+#pragma warning disable CA1034 // Nested types should not be visible
+        /// <summary>
+        /// The values of all declared <see cref="LexicalNormalizerName"/> properties as string constants.
+        /// These can be used in <see cref="SimpleFieldAttribute"/>, <see cref="SearchableFieldAttribute"/> and anywhere else constants are required.
+        /// </summary>
+        public static class Values
+        {
+            /// <summary> Converts alphabetic, numeric, and symbolic Unicode characters which are not in the first 127 ASCII characters (the &quot;Basic Latin&quot; Unicode block) into their ASCII equivalents, if such equivalents exist. See http://lucene.apache.org/core/4_10_3/analyzers-common/org/apache/lucene/analysis/miscellaneous/ASCIIFoldingFilter.html. </summary>
+            public const string AsciiFolding = LexicalNormalizerName.AsciiFoldingValue;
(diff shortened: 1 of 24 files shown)


sdk/search/Azure.Search.Documents/src/Indexes/Models/LexicalNormalizerName.cs
```diff
@@ -0,0 +1,28 @@
...
+namespace Azure.Search.Documents.Indexes.Models
+{
+    public readonly partial struct LexicalNormalizerName
+    {
+#pragma warning disable CA1034 // Nested types should not be visible
+        /// <summary>
+        /// The values of all declared <see cref="LexicalNormalizerName"/> properties as string constants.
+        /// These can be used in <see cref="SimpleFieldAttribute"/>, <see cref="SearchableFieldAttribute"/> and anywhere else constants are required.
+        /// </summary>
+        public static class Values
```
jsquire, line 13: This is an interesting pattern.   Can you help me understand what the intention is for it?

ShivangiReja: We are using `LexicalNormalizerName` constant values for `SimpleFieldAttribute` and `SearchableFieldAttribute`. You can see an example [here](https://github.com/Azure/azure-sdk-for-net/blob/5293a0c02298c3ad359962a0caf2395c2bf56986/sdk/search/Azure.Search.Documents/tests/Models/ReflectableModel.cs#L96).

I tried using an [existing enum type](https://github.com/Azure/azure-sdk-for-net/blob/5293a0c02298c3ad359962a0caf2395c2bf56986/sdk/search/Azure.Search.Documents/src/Generated/Models/LexicalNormalizerName.cs#L14) for the attributes [here](https://github.com/Azure/azure-sdk-for-net/blob/5293a0c02298c3ad359962a0caf2395c2bf56986/sdk/search/Azure.Search.Documents/src/Indexes/SimpleFieldAttribute.cs#L58) but compiler gives `LexicalNormalizerName is not a valid attribute parameter type` error:

image

So I used the existing pattern for `LexicalNormalizerName` that we using for [LexicalAnalyzerName](https://github.com/Azure/azure-sdk-for-net/blob/5293a0c02298c3ad359962a0caf2395c2bf56986/sdk/search/Azure.Search.Documents/src/Indexes/Models/LexicalAnalyzerName.cs#L13) and it works well for attributes.

jsquire: I was poking more at "why are we wrapping values in an inner static class rather than just exposing the members".

ShivangiReja: >@jsquire - why are we wrapping values in an inner static class rather than just exposing the members

Because those members already exist in the generated type [here](https://github.com/Azure/azure-sdk-for-net/blob/5293a0c02298c3ad359962a0caf2395c2bf56986/sdk/search/Azure.Search.Documents/src/Generated/Models/LexicalNormalizerName.cs#L32-L40).

So compiler gives this error:
image

heaths: This was the suggested work around from @tg-msft and/or @KrzysztofCwalina at the time since, yes, you can't use the "extensible enum" struct as an attribute value (not constant).

pallavit (approved):

PR #29625, opened 2022-07-02
Changes in azure-sdk-for-net for PR #2405 in autorest.csharp

AlexanderSher:
This is a draft PR that would contain all changes required for https://github.com/Azure/autorest.csharp/pull/2405


AlexanderSher:
DO NOT MERGE until https://github.com/Azure/autorest.csharp/pull/2405 is merged.

eng/Packages.Data.props
```diff
@@ -155,7 +155,7 @@
     All should have PrivateAssets="All" set so they don't become package dependencies
   -->
   <ItemGroup>
-    <PackageReference Update="Microsoft.Azure.AutoRest.CSharp" Version="3.0.0-beta.20220701.3" PrivateAssets="All" />
+    <PackageReference Update="Microsoft.Azure.AutoRest.CSharp" Version="3.0.0-beta.20220701.2" PrivateAssets="All" />
```
jsquire, line 158: This seems like a down-grade, unless versioning in Autorest differs.  This seems like something we shouldn't do...

sdk/textanalytics/Azure.AI.TextAnalytics.Legacy.Shared/src/Generated/TextAnalyticsRestClient.cs
```diff
@@ -75,7 +75,7 @@ public async Task<ResponseWithHeaders<TextAnalyticsAnalyzeHeaders>> AnalyzeAsync
                 case 202:
                     return ResponseWithHeaders.FromValue(headers, message.Response);
                 default:
-                    throw await ClientDiagnostics.CreateRequestFailedExceptionAsync(message.Response).ConfigureAwait(false);
```
jsquire, line 78: I wonder if this change means that we should revisit passing around `ClientDiagnostics` through the TA layers.   IIRC, we do this only because of this use. 
@maririos - thoughts?

heaths (approved):
CLU, KV changes LGTM

ShivangiReja (approved):
Search changes LGTM

amnguye (approved):
LGTM - Storage

sdk/textanalytics/Azure.AI.TextAnalytics.Legacy.Shared/src/Generated/TextAnalyticsRestClient.cs
maririos: we would still need it for the LRO classes as we manage throwing exceptions there, no?

christothes: The updated pattern doesn't require ClientDiagnostics. I just updated `LanguageServiceClient` with the new approach [here](https://github.com/Azure/azure-sdk-for-net/pull/29625/commits/deef24801d5fd27fe6da328158673599ac8cdfa6)

kinelski (approved):
Form Recognizer and Metrics Advisor LGTM.

PR #29641, opened 2022-07-05
Attach Statsbeat [Part1] - Set up

vishweshbankwar:
This is a initial PR for Statsbeat collection. The idea is to use OTel metric SDK with Azure Monitor exporter to export statsbeat metrics to Application Insights. There are two types of statsbeat that will be collected at interval of 24 hrs and 15 mins. This PR covers the overall structure with attach statsbeat which has a collection interval of 24 hrs. For the metrics that needs to be exported at 15 mins interval, another instance of metricreader will be added.


sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/StatsBeatHelper.cs
```diff
@@ -0,0 +1,25 @@
...
+using Azure.Monitor.OpenTelemetry.Exporter.Internals.ConnectionString;
+
+namespace Azure.Monitor.OpenTelemetry.Exporter.Internals
+{
+    internal class StatsBeatHelper
+    {
+        internal static void InitializeStatsBeat(string connectionString)
+        {
+            // initialize statsbeat by accessing singleton instance
+            _ = StatsBeat.StatsBeatInstance;
```
rajkumar-rangaraj, line 13: Verify if this initialization is needed.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/StatsBeat.cs
```diff
@@ -0,0 +1,59 @@
...
+            })
+            .Build();
+        }
+
+        public static StatsBeat StatsBeatInstance { get; } = new();
+
+        private static Measurement<int> GetAttachStatsBeat()
+        {
+            // TODO: Add additional properties required for statbeat
+            return new((int)1, new("ikey", s_ikey), new("language", "DotNet"), new("os", s_os));
```
rajkumar-rangaraj, line 51: ikey->c_ikey. Change s_ikey to c_ikey.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/StatsBeat.cs
```diff
@@ -0,0 +1,59 @@
...
+{
+    internal sealed class StatsBeat : IDisposable
+    {
+        private readonly IDisposable _meterProvider;
+
+        private static readonly Meter s_myMeter = new("AttachStatsBeatMeter", "1.0");
+
+        private const string ConnectionString = "<ConnectionString>";
+
+        internal static string s_roleName;
```
rajkumar-rangaraj, line 19: These should be a property as we are assigning outside of this class.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/AzureMonitorStatsbeatExporter.cs
```diff
@@ -0,0 +1,51 @@
...
+
+using System;
+using System.Threading;
+using Azure.Core.Pipeline;
+using Azure.Monitor.OpenTelemetry.Exporter.Internals;
+
+using OpenTelemetry;
+using OpenTelemetry.Metrics;
+
+namespace Azure.Monitor.OpenTelemetry.Exporter
```
TimothyMothra, line 12: nit: consider moving these 4 new files into a new subdirectory: "Internals/Statsbeat"

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/AzureMonitorExporterOptions.cs
```diff
@@ -51,5 +51,7 @@ public enum ServiceVersion
         public string StorageDirectory { get; set; }
 
         public bool DisableOfflineStorage { get; set; }
+
+        internal int StatsbeatInterval { get; set; }
```
TimothyMothra, line 55: Consider making this a `TimeSpan`. That would eliminate confusion about the unit here (milliseconds, minutes, etc).

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/StatsBeat.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        private const string ConnectionString = "<ConnectionString>";
+
+        private static string s_roleName;
+
+        private static string s_roleInstance;
+
+        private static string s_ikey;
+
+        internal static string s_os = StorageHelper.IsWindowsOS() ? "windows" : "linux";
```
TimothyMothra, line 26: Seems odd to me that this comes from the StorageHelper.
We might want to consider moving this to a different class (RuntimeHelper ?).

The `RuntimeInformation.IsOSPlatform` actually supports Windows, Linux, and OSX. 
Does Statsbeat want to record OSX here?

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/Statsbeat/Statsbeat.cs
```diff
@@ -0,0 +1,49 @@
...
+{
+    internal static class Statsbeat
+    {
+        private static readonly Meter s_myMeter = new("AttachStatsBeatMeter", "1.0");
+
+        private const string StatsBeat_ConnectionString = "<StatsBeat_ConnectionString>";
+
+        // TODO: Move IsWindowsOS() to new class
+        // Do we need to support OSX?
+        internal static string s_OS = StorageHelper.IsWindowsOS() ? "windows" : "linux";
```
cijothomas, line 20: if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux)) why not this!

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/Statsbeat/Statsbeat.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        static Statsbeat()
+        {
+            s_myMeter.CreateObservableGauge("AttachStatsBeat", () => GetAttachStatsBeat());
+
+            // Configure for attach statsbeat which has collection
+            // schedule of 24 hrs == 86400000 milliseconds.
+            Sdk.CreateMeterProviderBuilder()
+            .AddMeter("AttachStatsBeatMeter")
+            .AddStatsbeatExporter(o =>
```
cijothomas, line 32: can we not use AzMonExporter itself?

vishweshbankwar: We might be able to but that will need some changes on AzMonExporter side like special casing for resources. For statsbeat I introduced new properties on `ResourceParser` and AzMonexporter does not allow export interval to be configurable right now. Also, if we need to make any changes to just statsbeat, having separate exporter might help better in that case.

cijothomas: > like special casing for resources

Not clear why. Can you elaborate?

cijothomas: > AzMonexporter does not allow export interval

Can you modify the Gauge callback to deal with it? Return measurement only once every 15 mins, else return empty?

vishweshbankwar: > > AzMonexporter does not allow export interval
> 
> Can you modify the Gauge callback to deal with it? Return measurement only once every 15 mins, else return empty?

I am not sure if I understood this - do you have an example?

cijothomas: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs#L65-L74

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/Statsbeat/Statsbeat.cs
vishweshbankwar: Removing this to keep PR short. will revisit this in later PRs

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/Statsbeat/Statsbeat.cs
vishweshbankwar: Added internal option to deal with this.

rajkumar-rangaraj (approved):

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/AzureMonitorExporterMetricExtensions.cs
```diff
@@ -31,7 +31,7 @@ public static MeterProviderBuilder AddAzureMonitorMetricExporter(this MeterProvi
 
             var exporter = new AzureMonitorMetricExporter(options);
 
-            return builder.AddReader(new PeriodicExportingMetricReader(new AzureMonitorMetricExporter(options))
+            return builder.AddReader(new PeriodicExportingMetricReader(new AzureMonitorMetricExporter(options), options.PeriodicReaderExportIntervalInMilliseconds)
```
cijothomas, line 34: we can achieve this without touching existing exporter path.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/Statsbeat/Statsbeat.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        static Statsbeat()
+        {
+            s_myMeter.CreateObservableGauge("AttachStatsBeat", () => GetAttachStatsBeat());
+
+            // Configure for attach statsbeat which has collection
+            // schedule of 24 hrs == 86400000 milliseconds.
+            Sdk.CreateMeterProviderBuilder()
+            .AddMeter("AttachStatsBeatMeter")
+            .AddAzureMonitorMetricExporter(o =>
```
cijothomas, line 26: no need of using this extension method. can we not add the reader ourself by instantiating AzureMonitormetricExporter?

vishweshbankwar: updated

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/Statsbeat/Statsbeat.cs
```diff
@@ -0,0 +1,43 @@
...
+            s_myMeter.CreateObservableGauge("AttachStatsBeat", () => GetAttachStatsBeat());
+
+            // Configure for attach statsbeat which has collection
+            // schedule of 24 hrs == 86400000 milliseconds.
+            var exporterOptions = new AzureMonitorExporterOptions();
+            exporterOptions.DisableOfflineStorage = true;
+            exporterOptions.ConnectionString = StatsBeat_ConnectionString;
+
+            Sdk.CreateMeterProviderBuilder()
+            .AddMeter("AttachStatsBeatMeter")
```
reyang, line 29:
```suggestion
                .AddMeter("AttachStatsBeatMeter")
```

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/Statsbeat/Statsbeat.cs
```diff
@@ -0,0 +1,43 @@
...
+        private const string StatsBeat_ConnectionString = "<StatsBeat_ConnectionString>";
+
+        internal const int AttachStatsBeatInterval = 86400000;
+
+        static Statsbeat()
+        {
+            s_myMeter.CreateObservableGauge("AttachStatsBeat", () => GetAttachStatsBeat());
+
+            // Configure for attach statsbeat which has collection
+            // schedule of 24 hrs == 86400000 milliseconds.
```
reyang, line 23: Do we have a rough idea regarding the percentage of processes that are going to live above 24 hours?

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/Statsbeat/StatsbeatHelper.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Azure.Monitor.OpenTelemetry.Exporter.Internals.ConnectionString;
+
+namespace Azure.Monitor.OpenTelemetry.Exporter.Internals
+{
+    internal class StatsbeatHelper
+    {
+        internal static void InitializeStatsbeat(string connectionString)
```
cijothomas, line 10: who calls this?

vishweshbankwar: This will need to be added to all 3 signals.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/Statsbeat/Statsbeat.cs
vishweshbankwar: not at the moment. I will follow up to see if we can get this information from our rp integrations.

cijothomas: maybe call forceflush at startup, so we ensure stats is sent once. and then if the process lives long enough, then every 24 hours ..

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/Statsbeat/Statsbeat.cs
```diff
@@ -0,0 +1,43 @@
...
+            var exporterOptions = new AzureMonitorExporterOptions();
+            exporterOptions.DisableOfflineStorage = true;
+            exporterOptions.ConnectionString = StatsBeat_ConnectionString;
+
+            Sdk.CreateMeterProviderBuilder()
```
cijothomas, line 28: save the provider to a variable, and possibly call dispose at shutdown

rajkumar-rangaraj: Out of curiosity, what would happen if the meter provider is not disposed at shutdown?

cijothomas: nothing.
if we run for 23 hours and shut down, dispose will ensure we emit one metric. If the apps never run for 24hours, then dispose ensures metric is emitted atleast once in process lifetime, when it is about to die.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/Statsbeat/Statsbeat.cs
vishweshbankwar: I have added a todo for now to follow up with the spec on behavior in case of app exiting before 24hr duration

PR #29651, opened 2022-07-05
[Event Hubs] Processor Validation Method Rename

jsquire:
# Summary

The focus of these changes is to remove an unintended parameter and rename the new protected method responsible for validating preconditions of the processor when it is started.   This is intended to ensure that the processor fails fast for configuration and permissions issues that would prevent it from making forward progress.


jsquire:

christothes (approved):

PR #29660, opened 2022-07-06
resolve name issues for `Azure.ResourceManager.Storage`

ArcturusZhang:


sdk/storage/Azure.ResourceManager.Storage/src/autorest.md
```diff
@@ -94,12 +167,35 @@ directive:
...
+    where: $.definitions.DeletedAccountProperties.properties.storageAccountResourceId
+    transform: $["x-ms-format"] = "arm-id";
+  - from: swagger-document
+    where: $.definitions.DeletedAccountProperties.properties.creationTime
+    transform: $["format"] = "date-time";
+  - from: swagger-document
+    where: $.definitions.DeletedAccountProperties.properties.deletionTime
+    transform: $["format"] = "date-time";
+  - from: swagger-document
+    where: $.definitions.StorageAccountProperties.properties.primaryLocation
```
ArthurMa1978, line 184: Use `format-by-name-rules`, '*location'

sdk/storage/Azure.ResourceManager.Storage/src/Generated/FileShareData.cs
```diff
@@ -65,7 +65,7 @@ internal FileShareData(ResourceIdentifier id, string name, ResourceType resource
             ShareUsageBytes = shareUsageBytes;
             LeaseStatus = leaseStatus;
             LeaseState = leaseState;
-            LeaseDuration = leaseDuration;
+            LeaseDurationType = leaseDurationType;
```
blueww, line 68: The same property calls "LeaseDuration" in dataplane. We should keep terminology alignment between dataplane and management plane.
https://github.com/Azure/azure-sdk-for-net/blob/3b0aed1451566beef9f7b57158d0b4025af28496/sdk/storage/Azure.Storage.Files.Shares/src/Models/ShareProperties.cs#L93

sdk/storage/Azure.ResourceManager.Storage/src/autorest.md
ArcturusZhang: I would not like to do that since I do not have solid confirmation that others are also locations

sdk/storage/Azure.ResourceManager.Storage/src/Generated/FileShareData.cs
blueww: The type name change should be OK.

ArthurMa1978 (approved):

PR #29676, opened 2022-07-06
Export activity exception event

vishweshbankwar:
**End To End view - AI SDK** 
image

**End To End view - OTel exporter**
image

**Exception Details - AI SDK**
image
image

**Exception Details - OTel exporter**
image


sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TraceHelper.cs
```diff
@@ -27,7 +27,7 @@ internal static List<TelemetryItem> OtelToAzureMonitorTrace(Batch<Activity> batc
             foreach (var activity in batchActivity)
             {
                 var monitorTags = EnumerateActivityTags(activity);
-                telemetryItem = new TelemetryItem(activity, ref monitorTags, roleName, roleInstance, instrumentationKey);
+                telemetryItem = new TelemetryItem(TelemetryItem.s_telemetryItem_Name_Mapping[activity.GetTelemetryType()], activity, ref monitorTags, roleName, roleInstance, instrumentationKey);
```
rajkumar-rangaraj, line 30: Should we move `s_telemetryItem_Name_Mapping` to this file and rename it to a shorter name?

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TraceHelper.cs
```diff
@@ -168,5 +171,68 @@ internal static string GetOperationName(Activity activity, ref AzMonList MappedT
 
             return activity.DisplayName;
         }
+
+        private static void AddExceptionTelemetryFromActivityExceptionEvents(Activity activity, ref TagEnumerationState monitorTags, string roleName, string roleInstance, string instrumentationKey, List<TelemetryItem> telemetryItems)
+        {
+            foreach (var evnt in activity.Events)
```
rajkumar-rangaraj, line 177: We could avoid a foreach allocation, if we check for activity.Events.Count() > 0

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TraceHelper.cs
```diff
@@ -168,5 +177,71 @@ internal static string GetOperationName(Activity activity, ref AzMonList MappedT
...
+
+            TelemetryExceptionDetails exceptionDetails = new(exceptionMessage);
+
+            exceptionDetails.Stack = exceptionStackTrace;
+            exceptionDetails.TypeName = exceptionType;
+
+            List<TelemetryExceptionDetails> exceptions = new List<TelemetryExceptionDetails>();
+            exceptions.Add(exceptionDetails);
+
+            TelemetryExceptionData exceptionData = new TelemetryExceptionData(2, exceptions);
```
TimothyMothra, line 238: What does `2` mean here?
Consider using named parameters.

TimothyMothra:
The implementation looks good to me. 
Can you add any new tests that verify the Exception Telemetry?

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TraceHelper.cs
```diff
@@ -168,5 +177,71 @@ internal static string GetOperationName(Activity activity, ref AzMonList MappedT
 
             return activity.DisplayName;
         }
+
+        private static void AddExceptionTelemetryFromActivityExceptionEvents(Activity activity, ref TagEnumerationState monitorTags, string roleName, string roleInstance, string instrumentationKey, List<TelemetryItem> telemetryItems)
+        {
+            if (activity.Events.Any())
```
rajkumar-rangaraj, line 183: Nesting makes it difficult to read the code. How about the below code?

```
if (!activity.Events.Any())
    return;
```

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TraceHelper.cs
```diff
@@ -19,6 +19,12 @@ internal static class TraceHelper
         private const int Version = 2;
         private const int MaxlinksAllowed = 100;
 
+        internal static readonly IReadOnlyDictionary<TelemetryType, string> s_telemetryItem_Name = new Dictionary<TelemetryType, string>
```
rajkumar-rangaraj, line 22: s_telemetryItem_Name => s_telemetryName?

cijothomas, line 22: a simple if else would be much better than a Dictionary to achieve this.

rajkumar-rangaraj: Yes, this list earlier had may values. If else would be better.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TraceHelper.cs
```diff
@@ -168,5 +177,71 @@ internal static string GetOperationName(Activity activity, ref AzMonList MappedT
...
             return activity.DisplayName;
         }
+
+        private static void AddExceptionTelemetryFromActivityExceptionEvents(Activity activity, ref TagEnumerationState monitorTags, string roleName, string roleInstance, string instrumentationKey, List<TelemetryItem> telemetryItems)
+        {
+            if (activity.Events.Any())
+            {
+                foreach (var evnt in activity.Events)
+                {
+                    if (evnt.Name == SemanticConventions.AttributeExceptionEventName)
```
cijothomas, line 187: can we exit the for loop once we fine the exception event? Do you want to support the case where user has added more than exception to events?

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TraceHelper.cs
```diff
@@ -168,5 +177,71 @@ internal static string GetOperationName(Activity activity, ref AzMonList MappedT
...
+            string exceptionType = null;
+            string exceptionStackTrace = null;
+            string exceptionMessage = null;
+
+            // TODO: update to use perf improvements in .NET7.0
+            foreach (var tag in activityEventTags)
+            {
+                if (tag.Key == SemanticConventions.AttributeExceptionType)
+                {
+                    exceptionType = tag.Value.ToString();
```
cijothomas, line 215: a dictionary to cache this string would be nice. else we end up with new strings ...

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TraceHelper.cs
vishweshbankwar: updated

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TraceHelper.cs
vishweshbankwar: Yes - I was considering the case where user has called `activity.RecordException(ex)` multiple times

vishweshbankwar:
> The implementation looks good to me. Can you add any new tests that verify the Exception Telemetry?

Added one simple test

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TraceHelper.cs
```diff
@@ -168,5 +172,73 @@ internal static string GetOperationName(Activity activity, ref AzMonList MappedT
...
+                    continue;
+                }
+                if (tag.Key == SemanticConventions.AttributeExceptionStacktrace)
+                {
+                    exceptionStackTrace = tag.Value.ToString();
+                    continue;
+                }
+            }
+
+            TelemetryExceptionDetails exceptionDetails = new(exceptionMessage);
```
cijothomas, line 227: is there a guarantee that exceptionMessage,Type,Stack will be non null here?

vishweshbankwar: No. type and stack can be null. Message cannot be otherwise it will throw. Refactored the code a bit to save extra work.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TraceHelper.cs
```diff
@@ -168,5 +172,86 @@ internal static string GetOperationName(Activity activity, ref AzMonList MappedT
...
+                if (evnt.Name == SemanticConventions.AttributeExceptionEventName)
+                {
+                    try
+                    {
+                        var exceptionData = GetExceptionDataDetailsOnTelemetryItem(evnt.Tags);
+                        if (exceptionData == null)
+                        {
+                            continue;
+                        }
+                        else
```
rajkumar-rangaraj, line 194: nit: else may not be needed here.

rajkumar-rangaraj (approved):

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TraceHelper.cs
```diff
@@ -168,5 +172,84 @@ internal static string GetOperationName(Activity activity, ref AzMonList MappedT
 
             return activity.DisplayName;
         }
+
+        private static void AddExceptionTelemetryFromActivityExceptionEvents(Activity activity, ref TagEnumerationState monitorTags, string roleName, string roleInstance, string instrumentationKey, List<TelemetryItem> telemetryItems)
+        {
+            // Export exceptions from server spans only
+            // TODO: how do we handler consumer/producer spans?
+            if (!activity.Events.Any() || activity.Kind != ActivityKind.Server)
```
cijothomas, line 180: Exporter shouldn't ignore exceptions based on Kind.

cijothomas (dismissed):
See https://github.com/Azure/azure-sdk-for-net/pull/29676/files#r965232534

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TraceHelper.cs
vishweshbankwar: This is something I am planning to validate with Osvaldo. There is no client side exception thing in AI SDK. As an example if there is a network error while making Http call, AI SDK will not export that exception as a separate exception telemetry if it was handled by user using try catch. It will just export the dependency telemetry with resultCode as `Faulted`.

In case of exporter we do not have a way to know whether it was handled or not Because we simply look at the activity event and export it (HttpClient fires exception event and instrumentation library records it). 

This could be a behavior change between AI SDK and OTEL but it needs to be called out as we will export more data than we do currently with AISDK. That is the reason to limit the exception export to `Kind == Server` for now. If we agree with the behavior change, will remove the condition.

cijothomas: > As an example if there is a network error while making Http call, AI SDK will not export that exception as a separate exception telemetry if it was handled by user using try catch. It will just export the dependency telemetry with resultCode as Faulted.

^ Same with Asp.net Core reporting RequestTelemetry (and not ExceptionTeleemtry)

vishweshbankwar: Removed the check as per offline discussion.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TraceHelper.cs
```diff
@@ -168,5 +172,84 @@ internal static string GetOperationName(Activity activity, ref AzMonList MappedT
 
             return activity.DisplayName;
         }
+
+        private static void AddExceptionTelemetryFromActivityExceptionEvents(Activity activity, ref TagEnumerationState monitorTags, string roleName, string roleInstance, string instrumentationKey, List<TelemetryItem> telemetryItems)
+        {
+            // Export exceptions from server spans only
```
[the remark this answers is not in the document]
vishweshbankwar: No - removed it.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TraceHelper.cs
vishweshbankwar: added a todo for this to follow up with.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TraceHelper.cs
```diff
@@ -168,5 +172,83 @@ internal static string GetOperationName(Activity activity, ref AzMonList MappedT
 
             return activity.DisplayName;
         }
+
+        private static void AddExceptionTelemetryFromActivityExceptionEvents(Activity activity, ref TagEnumerationState monitorTags, string roleName, string roleInstance, string instrumentationKey, List<TelemetryItem> telemetryItems)
+        {
+            if (!activity.Events.Any())
```
cijothomas, line 178: nit: could we check this even before calling this method?

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TraceHelper.cs
```diff
@@ -168,5 +172,83 @@ internal static string GetOperationName(Activity activity, ref AzMonList MappedT
...
+                        if (exceptionData != null)
+                        {
+                            var exceptionTelemetryItem = new TelemetryItem("Exception", activity, ref monitorTags, roleName, roleInstance, instrumentationKey);
+                            exceptionTelemetryItem.Data = exceptionData;
+                            telemetryItems.Add(exceptionTelemetryItem);
+                        }
+                    }
+                    catch (Exception ex)
+                    {
+                        AzureMonitorExporterEventSource.Log.WriteError("FailedToExtractExceptionFromActivityEvent", ex);
```
cijothomas, line 199: Do we need Error level? A warning would be more apt?

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TraceHelper.cs
```diff
@@ -168,5 +172,83 @@ internal static string GetOperationName(Activity activity, ref AzMonList MappedT
...
+
+            if (exceptionMessage == null)
+            {
+                return null;
+            }
+
+            TelemetryExceptionDetails exceptionDetails = new(exceptionMessage);
+
+            exceptionDetails.Stack = exceptionStackTrace;
+            exceptionDetails.TypeName = exceptionType;
```
cijothomas, line 240: are Stack, TypeName optional?

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TraceHelper.cs
```diff
@@ -168,5 +172,83 @@ internal static string GetOperationName(Activity activity, ref AzMonList MappedT
...
+
+            List<TelemetryExceptionDetails> exceptions = new List<TelemetryExceptionDetails>();
```
cijothomas, line 242: a new List is created just to store a single exception. Not much we can without a schema level fix. But you can create the list with size 1, to avoid wasted space even for a moment.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/TraceHelperTests.cs
```diff
@@ -217,6 +218,30 @@ public void LinksAreNotTruncatedWhenCanBeFitInMaxLength(string telemetryType)
...
 
+        [Fact]
+        public void ActivityWithExceptionEventCreatesExceptionTelemetry()
+        {
+            using ActivitySource activitySource = new ActivitySource(ActivitySourceName);
+            using var activity = activitySource.StartActivity(
+               ActivityName,
+               ActivityKind.Server);
+
+            activity.RecordException(new Exception("Exception Message"));
```
cijothomas, line 229: Move "Exception Message" to variable above.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/TraceHelperTests.cs
```diff
@@ -217,6 +218,30 @@ public void LinksAreNotTruncatedWhenCanBeFitInMaxLength(string telemetryType)
...
+            Batch<Activity> batch = new Batch<Activity>(activityList, 1);
+
+            var telemetryItems = TraceHelper.OtelToAzureMonitorTrace(batch, "roleName", "roleInstance", "00000000 - 0000 - 0000 - 0000 - 000000000000");
+
+            Assert.Equal(2, telemetryItems.Count());
+            Assert.Equal("Request", (IEnumerable<char>)telemetryItems[0].Name);
+            Assert.Equal("Exception", (IEnumerable<char>)telemetryItems[1].Name);
+            Assert.Equal("Exception Message", (IEnumerable<char>)(telemetryItems[1].Data.BaseData as TelemetryExceptionData).Exceptions.First().Message);
+            Assert.Equal("System.Exception", (IEnumerable<char>)(telemetryItems[1].Data.BaseData as TelemetryExceptionData).Exceptions.First().TypeName);
+            Assert.Equal("System.Exception: Exception Message", (IEnumerable<char>)(telemetryItems[1].Data.BaseData as TelemetryExceptionData).Exceptions.First().Stack);
```
cijothomas, line 242: please add test for scenarios involving missing Stack, TypeName, Message.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TraceHelper.cs
vishweshbankwar: As per schema, yes.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Customizations/Models/TelemetryItem.cs
```diff
@@ -16,16 +16,15 @@ namespace Azure.Monitor.OpenTelemetry.Exporter.Models
...
-            [TelemetryType.Dependency] = "RemoteDependency",
-        };
 
-        public TelemetryItem(Activity activity, ref TagEnumerationState monitorTags, string roleName, string roleInstance, string instrumentationKey) :
-            this(s_telemetryItem_Name_Mapping[activity.GetTelemetryType()], FormatUtcTimestamp(activity.StartTimeUtc))
+        public TelemetryItem(string name, Activity activity, ref TagEnumerationState monitorTags, string roleName, string roleInstance, string instrumentationKey) :
+            this(name, FormatUtcTimestamp(activity.StartTimeUtc))
         {
-            if (activity.ParentSpanId != default)
+            if (name == "Exception")
```
TimothyMothra, line 23: Can the values for `name` be either an Enum or string constants?

This is a red flag for me. Inline strings have a risk of typos, which could affect behavior in if statements.

vishweshbankwar: Yes - this needs to be fixed in other places as well (Request and Dependency). I will add a todo for this.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TraceHelper.cs
vishweshbankwar: I did the test without type / stack. type is required, ingestion reject the record without it. stack is optional. We need to fix the swagger schema as it allows exceptionDetails to be created without type right now. @rajkumar-rangaraj  FYI.

cijothomas (approved):
LGTM. There are some inefficient schema related stuff. which can be optimized later.
Also add a changelog entry too.

PR #29677, opened 2022-07-06
Setting MaxDepth for JsonReaderSettings

amnguye:
Setting MaxDepth for JsonReaderSettings to 128 and making JsonSeralizerSettings publicly settable.


sdk/storage/Microsoft.Azure.WebJobs.Extensions.Storage.Common/src/Shared/Constants.cs
```diff
@@ -6,5 +6,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.Storage.Common
     internal static class Constants
     {
         public const string DateTimeFormatString = "yyyy'-'MM'-'dd'T'HH':'mm':'ss'.'fffK";
+        public const int MaxDepthDefault = 128;
```
heaths, line 9: This is too high. The default was 64 and leading to StackOverflowExceptions.

heaths (changes requested):

sdk/storage/Microsoft.Azure.WebJobs.Extensions.Storage.Common/src/Shared/Constants.cs
amnguye: Done

sdk/storage/Microsoft.Azure.WebJobs.Extensions.Storage.Common/src/Shared/Constants.cs
```diff
@@ -6,6 +6,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.Storage.Common
     internal static class Constants
     {
         public const string DateTimeFormatString = "yyyy'-'MM'-'dd'T'HH':'mm':'ss'.'fffK";
-        public const int MaxDepthDefault = 128;
+        public const int MaxDepthDefault = 4;
```
heaths, line 9: Is this large enough now? I only mentioned "4" in email - with a note saying implying it was entirely random - as an example. Make sure it's large enough for your needs if you know the shape of the response JSON.

Unblocking since I'm OOF starting tomorrow, but make sure this value works for you - whatever it may be.

heaths (approved):

sdk/storage/Microsoft.Azure.WebJobs.Extensions.Storage.Common/src/Shared/Constants.cs
amnguye: We can talk more offline about this value, I think we're still going back and forth about what the best value is. I put it back to 128 cause that's the default for the NewtonJson package.

sdk/storage/Microsoft.Azure.WebJobs.Extensions.Storage.Common/src/Shared/Constants.cs
heaths: Still 128, which is too high. See my original comment.

sdk/storage/Microsoft.Azure.WebJobs.Extensions.Storage.Common/src/Shared/JsonSerialization.cs
```diff
@@ -18,9 +20,27 @@ internal static class JsonSerialization
...
+            string envrMaxDepth = Environment.GetEnvironmentVariable(Constants.MaxDepthEnvironmentVariable);
+            if (!string.IsNullOrEmpty(envrMaxDepth))
+            {
+                if (int.TryParse(envrMaxDepth, NumberStyles.None, CultureInfo.InvariantCulture, out int maxDepth))
+                {
+                    return maxDepth;
+                }
+                else
+                {
+                    throw new ArgumentException($"Cannot parse the value for the Json Serializer Max Depth from the environment variable:\"{Constants.MaxDepthEnvironmentVariable}\"\n" +
```
heaths, line 37:
```suggestion
                    throw new ArgumentException($"Cannot parse the value for the JSON serializer max depth from the environment variable:\"{Constants.MaxDepthEnvironmentVariable}\"\n" +
```

heaths (changes requested):

sdk/storage/Microsoft.Azure.WebJobs.Extensions.Storage.Queues/src/QueueCausalityManager.cs
```diff
@@ -28,10 +29,12 @@ internal class QueueCausalityManager
         private const string ParentGuidFieldName = "$AzureWebJobsParentId";
 
         private readonly ILogger<QueueCausalityManager> _logger;
+        internal readonly JsonSerializerSettings jsonSerializerSettings;
 
-        public QueueCausalityManager(ILoggerFactory loggerFactory)
+        public QueueCausalityManager(ILoggerFactory loggerFactory, JsonSerializerSettings settings = default)
```
heaths, line 34: We should not be exposing Newtonsoft.Json directly from public APIs. Use an abstraction of only settings we care about instead to marshal values between types.

heaths (changes requested):

sdk/storage/Microsoft.Azure.WebJobs.Extensions.Storage.Common/src/Shared/Queues/QueuesOptions.cs
```diff
@@ -201,6 +203,21 @@ public QueueMessageEncoding MessageEncoding
             }
         }
 
+        /// <summary>
+        /// Gets or sets the JSON serialization settings to use when binding to POCOs.
+        /// </summary>
+#pragma warning disable AZC0014 // Avoid using banned types in public API
```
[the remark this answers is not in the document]
amnguye: I'm pattern matching what we have exposed in the ServiceBusOptions in the WebJobs Extensions for Service Bus
https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusOptions.cs#L150
which is how we're currently exposing the Json Serializer Settings.

I can do something different here where we have our own object for the Json serializer settings.

Did we want to open an issue to change the existing WebJobs ServiceBus Json Serializer settings?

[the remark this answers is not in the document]
heaths: Are these track 1 packages, or track 2 extension packages? It's hard to tell from the name, unfortunately. If track 1, then exposing `JsonSerializerSettings` directly is fine.

sdk/storage/Microsoft.Azure.WebJobs.Extensions.Storage.Queues/src/QueueCausalityManager.cs
heaths: Same response to this.

sdk/storage/Microsoft.Azure.WebJobs.Extensions.Storage.Common/src/Shared/Queues/QueuesOptions.cs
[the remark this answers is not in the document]
amnguye: This PR is for the track 2 extensions for WebJobs Extensions for Storage.

@JoshLove-msft Do you know if the [this ServiceBusOptions](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusOptions.cs#L150) is apart of Track 1 or Track 2?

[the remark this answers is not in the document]
amnguye: @heaths It's track 2.

If we go about making our own object for the JsonSerializerSettings is it okay to be referencing the many Newtonsoft.Json enums that are used in that class for many of the properties?
https://www.newtonsoft.com/json/help/html/T_Newtonsoft_Json_JsonSerializerSettings.htm
e.g.  [JsonSerializerSettings.ConstructorHandling](https://www.newtonsoft.com/json/help/html/P_Newtonsoft_Json_JsonSerializerSettings_ConstructorHandling.htm) uses [Newtonsoft.Json.ConstuctorHandling](https://www.newtonsoft.com/json/help/html/T_Newtonsoft_Json_ConstructorHandling.htm) and so on.

It would look like this
```
using Newtonsoft.Json;
using Newtonsoft.Json.Serialization;

namespace Microsoft.Azure.WebJobs.Extensions.Storage.Common
{
 internal class JsonSerializationSettings
    {
        /// <summary>
        /// Using the Newtonsoft.Json.JsonSerializerSettings.
        /// </summary>
        private JsonSerializerSettings _internalSettings;

        public ReferenceLoopHandling ReferenceLoopHandling
        {
            get
            {
                return _internalSettings.ReferenceLoopHandling;
            }
            set
            {
                _internalSettings.ReferenceLoopHandling = value;
            }
        }
        public MissingMemberHandling MissingMemberHandling
        {
            get
            {
                return _internalSettings.MissingMemberHandling;
            }
            set
            {
                _internalSettings.MissingMemberHandling = value;
            }
        }
        public ObjectCreationHandling ObjectCreationHandling
        {
            get
            {
                return _internalSettings.ObjectCreationHandling;
            }
            set
            {
                _internalSettings.ObjectCreationHandling = value;
            }
        }
        public int? MaxDepth
        {
            get
            {
                return _internalSettings.MaxDepth;
            }
            set
            {
                _internalSettings.MaxDepth = value;
            }
        }
        /// And so on
    }
}
```

Is that okay?

sdk/storage/Microsoft.Azure.WebJobs.Extensions.Storage.Queues/src/QueueCausalityManager.cs
JoshLove-msft: We already have precedence for this in the Service Bus extensions. We should probably ask the architects.

sdk/storage/Microsoft.Azure.WebJobs.Extensions.Storage.Common/src/Shared/Queues/QueuesOptions.cs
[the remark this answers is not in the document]
heaths: If the extension is already referencing and exposing Newtonsoft.Json you can probably do so again here. But, if not, we need an abstraction for all types.

PR #29688, opened 2022-07-07
[Event Hubs] Processor Release Prep

jsquire:
# Summary

The focus of these changes is to prepare the Processor package for its July release.


christothes (approved):

jsquire:
Merging with Core failure; the only impactful change was to a package version referenced only by the Event Hubs libraries.   Core Analyze step is unrelated.

PR #29719, opened 2022-07-08
Traffic Manager library based on azure track 2 SDK

oldukhno:
Generated classes by the Traffic Manager swagger API definition
Worked out multiple customizations of the generated classes to overcome various issues
Covered all Collections and Resources with unit-tests


jsquire (changes requested):
Please rebase on `main` and ensure that the PR does not include files unrelated to the change that you're intending to make.

PR #29720, opened 2022-07-08
Change the format of `DefaultCustomBlockResponseBody` to `BinaryData`

ArthurMa1978:


sdk/cdn/Azure.ResourceManager.Cdn/api/Azure.ResourceManager.Cdn.netstandard2.0.cs
```diff
@@ -4266,7 +4266,7 @@ public partial class WafPolicyManagedRuleSet
     public partial class WafPolicySettings
     {
         public WafPolicySettings() { }
-        public string DefaultCustomBlockResponseBody { get { throw null; } set { } }
+        public System.BinaryData DefaultCustomBlockResponseBody { get { throw null; } set { } }
```
Yao725, line 4269: Do we have a new rule that the property ends with `Body` need to become `BinaryData`?

Yao725 (approved):

sdk/cdn/Azure.ResourceManager.Cdn/api/Azure.ResourceManager.Cdn.netstandard2.0.cs
ArthurMa1978: no, manual is ok

ArcturusZhang (commented):
Why do we not have the change in WafPolicySettings.cs but only the changes in the serialization and api?

PR #29730, opened 2022-07-08
Update Message.cs

jbjordan:
Updated comments for EnqueuedSequenceNumber


EldertGrootenboer (approved):

sdk/servicebus/Microsoft.Azure.ServiceBus/src/Message.cs
```diff
@@ -467,8 +467,12 @@ internal short PartitionId
             /// <summary>Gets or sets the original sequence number of the message.</summary>
```
jsquire, line 467:
```suggestion
            /// <summary>Gets the original sequence number of the message.</summary>
```

If this is intended to be read-only, as the remarks detail, we may want to consider clarifying in the summary as well, since that's the typical area that gets read by developers using code-completion.

jsquire (approved):
@jbjordan:  Let me know if you need me to merge on your behalf and please feel free to reach out if you'd like me to point you at the docs for requesting direct permissions.

EldertGrootenboer (approved):

PR #29731, opened 2022-07-08
remove apicompat condition

jaschrep-msft:
workaround no longer necessary


jaschrep-msft:
Addressed in a separate PR on a fork. This is duplicate.

PR #29733, opened 2022-07-08
Traffic Manager library based on azure track 2 SDK

oldukhno:
- Generated classes by the Traffic Manager swagger API definition
- Worked out multiple customizations of the generated classes to overcome various issues
- Covered all Collections and Resources with unit-tests


Danieladu:
> Thank you for your contribution @oldukhno! We will review the pull request and get back to you soon.

how it going now~Thanks

sdk/trafficmanager/Azure.ResourceManager.TrafficManager/tests/TrafficManagerManagementTestEnvironment.cs
```diff
@@ -0,0 +1,11 @@
...
+// Licensed under the MIT License.
+
+using Azure.Core.TestFramework;
+
+namespace Azure.ResourceManager.TrafficManager.Tests
+{
+    public class TrafficManagerManagementTestEnvironment : TestEnvironment
+    {
+    }
+}
```
m-nash, line 11: nit: newline

sdk/trafficmanager/Azure.ResourceManager.TrafficManager/src/Customization/EndpointCollection.cs
```diff
@@ -0,0 +1,54 @@
...
+            Argument.AssertNull(profileData, nameof(profileData));
+
+            this._profileData = profileData;
+        }
+
+        /// <summary>
+        /// Asynchroniosly lists all Traffic Manager endpoints within a profile.
+        /// </summary>
+        /// <returns> A collection of <see cref="EndpointResource" /> that may take multiple service requests to iterate over. </returns>
+        public virtual AsyncPageable<EndpointData> GetAllAsync()
```
m-nash, line 34: this is marked in the md as list exception yet we are adding it here?  Without digging in more I think we should remove that exception and simply leave this custom implementation if that is the difference that way the collection will have the property interfaces added like `IEnumerable`

sdk/trafficmanager/Azure.ResourceManager.TrafficManager/src/Customization/TrafficManagerUserMetricsKeysRestOperations.cs
```diff
@@ -0,0 +1,120 @@
...
+namespace Azure.ResourceManager.TrafficManager
+{
+    /// <summary>
+    /// Class to overcome issue when the original resource is generated by swagger has "default" suffix in it. E.g.
+    /// "/subscriptions/{subscriptionId}/providers/Microsoft.Network/trafficManagerUserMetricsKeys/default"
+    /// which generates HTTP 400 and simply does not work.
+    /// </summary>
+    internal partial class TrafficManagerUserMetricsKeysRestOperations
+    {
+        internal HttpMessage CreateGetRequest(string subscriptionId)
```
m-nash, line 19: very rare to have customizations in rest operations lets discuss this

sdk/trafficmanager/Azure.ResourceManager.TrafficManager/src/Customization/HeatmapType.cs
```diff
@@ -0,0 +1,15 @@
...
+
+using Azure.Core;
+
+namespace Azure.ResourceManager.TrafficManager.Models
+{
+    /// <summary>
+    /// Renaming Enum8 enum appropriately.
+    /// </summary>
+    [CodeGenModel("Enum8")]
+    public readonly partial struct HeatMapType
```
m-nash, line 12: lets do this with a rename in the md instead of this customization file.  An [example can be found here](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/resourcemanager/Azure.ResourceManager/src/autorest.md) searching for enum.

looks like this
```
  - from: resources.json
    where: $.definitions.AliasPathMetadata.properties.attributes["x-ms-enum"]
    transform:
      $["name"] = "ResourceTypeAliasPathAttributes";
```

sdk/trafficmanager/Azure.ResourceManager.TrafficManager/src/Customization/EndpointResource.cs
```diff
@@ -0,0 +1,23 @@
...
+namespace Azure.ResourceManager.TrafficManager
+{
+    /// <summary>
+    /// Customizes validation of the reousce name since default resource name of the Endpoint has a template parameter baked in.
+    /// </summary>
+    public partial class EndpointResource : ArmResource
+    {
+        internal static void ValidateResourceId(ResourceIdentifier id)
+        {
+            string expectedResourceType = ResourceType.ToString().Replace("{endpointType}", id.ResourceType.GetLastType());
```
m-nash, line 17: it looks like this simply ignores the Type and compares only the Namespace?  That would be a faster implementation if we swap that.

sdk/trafficmanager/Azure.ResourceManager.TrafficManager/src/Customization/ProfileListResult.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Azure.ResourceManager.TrafficManager.Models
+{
+    /// <summary>
+    /// Changing access modifier from internal to public since it's a result from public API.
+    /// </summary>
+    public partial class ProfileListResult
```
m-nash, line 9: is this still needed?

sdk/trafficmanager/Azure.ResourceManager.TrafficManager/src/Customization/EndpointRestOperations.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// <param name="profileName"> The name of the Traffic Manager profile. </param>
+        /// <param name="endpointType"> The type of the Traffic Manager endpoint to be deleted. </param>
+        /// <param name="endpointName"> The name of the Traffic Manager endpoint to be deleted. </param>
+        /// <param name="cancellationToken"> The cancellation token to use. </param>
+        /// <exception cref="ArgumentNullException"> <paramref name="subscriptionId"/>, <paramref name="resourceGroupName"/>, <paramref name="profileName"/>, <paramref name="endpointType"/> or <paramref name="endpointName"/> is null. </exception>
+        /// <exception cref="ArgumentException"> <paramref name="subscriptionId"/>, <paramref name="resourceGroupName"/>, <paramref name="profileName"/>, <paramref name="endpointType"/> or <paramref name="endpointName"/> is an empty string, and was expected to be non-empty. </exception>
+        public async Task<Response<DeleteOperationResult>> DeleteAsync(string subscriptionId, string resourceGroupName, string profileName, string endpointType, string endpointName, CancellationToken cancellationToken = default)
```
m-nash, line 26: rare to have customizations on rest operations lets discuss

sdk/trafficmanager/Azure.ResourceManager.TrafficManager/src/Generated/UserMetricsModelResource.cs
```diff
@@ -0,0 +1,189 @@
...
+        }
+
+        /// <summary>
+        /// Delete a subscription-level key used for Real User Metrics collection.
+        /// Request Path: /subscriptions/{subscriptionId}/providers/Microsoft.Network/trafficManagerUserMetricsKeys/default
+        /// Operation Id: TrafficManagerUserMetricsKeys_Delete
+        /// </summary>
+        /// <param name="waitUntil"> <see cref="WaitUntil.Completed"/> if the method should wait to return until the long-running operation has completed on the service; <see cref="WaitUntil.Started"/> if it should return after starting the operation. For more information on long-running operations, please see <see href="https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/samples/LongRunningOperations.md"> Azure.Core Long-Running Operation samples</see>. </param>
+        /// <param name="cancellationToken"> The cancellation token to use. </param>
+        public virtual async Task<ArmOperation<DeleteOperationResult>> DeleteAsync(WaitUntil waitUntil, CancellationToken cancellationToken = default)
```
m-nash, line 92: so the trailing /default is required for the delete methods but will break if you use it in the get methods?

m-nash (changes requested):

bmwaechter:
@oldukhno you still working on this PR? I would love to leverage the work in a project im currently working on.

oldukhno:
> @oldukhno you still working on this PR? I would love to leverage the work in a project im currently working on.

Hi, yes, I'm working on that as a back burner. The plan to have a discussion with [m-nash](https://github.com/m-nash) on his comments and commit this to master somewhere next week.

m-nash:
@oldukhno we are also missing the ci.yml file in the Azure.ResourceManager.TrafficManager folder, see an [exmaple here](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/resourcemanager/ci.yml).

sdk/trafficmanager/Azure.ResourceManager.TrafficManager/src/Customization/EndpointCollection.cs
oldukhno: Unfortunately, the exception and the customization were added on purpose and without the exception code generation does not succeed.

This is what node returns when we remove the exception from md.

OpenAPI spec has a security definition but autorest security config is defined. Security config from autorest will be used.
fatal   | The ResourceCollection EndpointCollection (RequestPath: /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Network/trafficmanagerprofiles/{profileName}/{endpointType}/{endpointName}) does not have a GetAll method

sdk/trafficmanager/Azure.ResourceManager.TrafficManager/src/Customization/HeatmapType.cs
oldukhno: Sure, let's do it. But for me reference, what's the benefit of doing it in the md vs class customization?

oldukhno: After scratching the surface, I do not think the advice is applicable here. 

The Enum8 gets generated automatically from the following definition (and is not x-ms-enum definition as in the example provided):

"operationId": "HeatMap_Get",
        "description": "Gets latest heatmap for Traffic Manager profile.",
        "parameters": [
....
{
            "name": "heatMapType",
            "in": "path",
            "required": true,
            "type": "string",
            "enum": [
              "default"
            ],
            "description": "The type of HeatMap for the Traffic Manager profile."
          }
...
]

sdk/trafficmanager/Azure.ResourceManager.TrafficManager/src/Customization/EndpointResource.cs
oldukhno: It substitutes the template with the actual value relevant to the endpoint resource.
E.g. "Microsoft.Network/trafficmanagerprofiles/{endpointType}"

sdk/trafficmanager/Azure.ResourceManager.TrafficManager/src/Customization/ProfileListResult.cs
oldukhno: yes, it changes default generated modifier from internal to public. The code does not compile without this customization.

sdk/trafficmanager/Azure.ResourceManager.TrafficManager/src/Generated/UserMetricsModelResource.cs
oldukhno: exactly

sdk/trafficmanager/Azure.ResourceManager.TrafficManager/tests/Scenario/HeatMapModelTests.cs
```diff
@@ -0,0 +1,104 @@
...
+namespace Azure.ResourceManager.TrafficManager.Tests
+{
+    public sealed class HeatMapModelTests : TrafficManagerManagementTestBase
+    {
+        private const string ProfileWithHeatmapResrouceGroupName = "dialtone-traffic-manager";
+        private const string ProfileWithHeatmapName = "az-int-int-msftmetrics";
+
+        private SubscriptionResource _subscription;
+
+        public HeatMapModelTests(bool isAsync) : base(isAsync, RecordedTestMode.Record)
```
m-nash, line 19: comment out record

sdk/trafficmanager/Azure.ResourceManager.TrafficManager/tests/Scenario/ProfileTestBase.cs
```diff
@@ -0,0 +1,99 @@
...
+        protected internal const string EndpointName2 = "endpoint2";
+        protected internal const string EndpointTypeName = "externalEndpoints";
+        protected internal const string EndpointType = "Microsoft.Network/trafficManagerProfiles/" + EndpointTypeName;
+
+        protected internal SubscriptionResource _subscription;
+        protected internal ProfileCollection _profileCollection;
+        protected internal ResourceGroupResource _resourceGroup;
+        protected internal string _profileName;
+
+        public ProfileTestBase(bool isAsync, RecordedTestMode mode = RecordedTestMode.Playback) : base(isAsync, mode)
```
m-nash, line 27: lets not hard code playback

m-nash: If we want to exclude a test from running in live use the `PlaybackOnlyAttribute`

sdk/trafficmanager/Azure.ResourceManager.TrafficManager/src/Customization/EndpointCollection.cs
m-nash: Lets add the interfaces to the customized class in that case

sdk/trafficmanager/Azure.ResourceManager.TrafficManager/tests/Scenario/UserMetricsModelTests.cs
```diff
@@ -0,0 +1,86 @@
...
+            UserMetricsModelResource userMetricsModelResource = _subscription.GetUserMetricsModel();
+            userMetricsModelResource = await userMetricsModelResource.GetAsync();
+
+            Assert.IsNotEmpty(userMetricsModelResource.Data.Key);
+
+            await Delete();
+
+            userMetricsModelResource = await userMetricsModelResource.GetAsync();
+
+            Assert.IsEmpty(userMetricsModelResource.Data.Key);
```
m-nash, line 43: validate that using a non existant resource results in `RequestFailedException` with a 404.

```
var badResource = Client.GetUserMetricsModelResource(new ResourceIdentifier(userMetricsModelResource.Id + "foo"));
var exception = Assert.ThrowsAsync<RequestFailedException>(async () => await badResource.DeleteAsync());
Assert.AreEqual(404, exception.Status);
```

sdk/trafficmanager/Azure.ResourceManager.TrafficManager/src/Customization/TrafficManagerUserMetricsKeysRestOperations.cs
```diff
@@ -0,0 +1,120 @@
...
+
+        /// <summary> Delete a subscription-level key used for Real User Metrics collection. </summary>
+        /// <param name="subscriptionId"> Gets subscription credentials which uniquely identify Microsoft Azure subscription. The subscription ID forms part of the URI for every service call. </param>
+        /// <param name="cancellationToken"> The cancellation token to use. </param>
+        /// <exception cref="ArgumentNullException"> <paramref name="subscriptionId"/> is null. </exception>
+        /// <exception cref="ArgumentException"> <paramref name="subscriptionId"/> is an empty string, and was expected to be non-empty. </exception>
+        /// <remarks>
+        /// Overcoming situation when the rest delete request results with HTTP 200 but does not provide any json content in the response.
+        /// </remarks>
+        public async Task<Response<DeleteOperationResult>> DeleteAsync(string subscriptionId, CancellationToken cancellationToken = default)
```
m-nash, line 78: instead of making this customization lets remove the [schema from the 200 response](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/trafficmanager/resource-manager/Microsoft.Network/stable/2018-08-01/trafficmanager.json#L937) since that matches reality in the autorest.md

sdk/trafficmanager/Azure.ResourceManager.TrafficManager/src/Customization/EndpointResource.cs
m-nash: Since this path only executes in debug mode lets change this to a nit, you can skip if you want

sdk/trafficmanager/Azure.ResourceManager.TrafficManager/src/Customization/ProfileListResult.cs
m-nash: These objects shouldn't be exposed on public endpoints, lets try changing it back to internal and find out where the leak is and address that.

sdk/trafficmanager/Azure.ResourceManager.TrafficManager/src/Customization/EndpointCollection.cs
```diff
@@ -0,0 +1,66 @@
...
+        public virtual Pageable<EndpointData> GetAll()
+        {
+            if (this._profileData == null)
+                throw new InvalidOperationException("The method can work only when the profileData is provided over ctor.");
+
+            return PageableHelpers.CreateEnumerable(_ => Page.FromValues(this._profileData.Endpoints, null, null), null, null);
+        }
+
+        /// <inheritdoc />
+        public IEnumerator<EndpointData> GetEnumerator() => GetAll().GetEnumerator();
```
m-nash, line 58: These should be implemented as explicit implementations so the API doesn't show up on the public surface.  Just copy paste [this](https://github.com/Azure/azure-sdk-for-net/pull/29733/files#diff-f2bbb58f384717f0c18f3fe1bfdcb6ecd2730447e49bcf285a7e5cce9e2e39d6R289-R302).

sdk/trafficmanager/Azure.ResourceManager.TrafficManager/src/Customization/TrafficManagerUserMetricsKeysRestOperations.cs
```diff
@@ -0,0 +1,70 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Azure.Core;
+using Azure.ResourceManager.TrafficManager.Models;
+
+namespace Azure.ResourceManager.TrafficManager
```
m-nash, line 10: lets mirror the folder structure in the `Generated` folder.  This fille should go in a subfolder called `RestOperations`

m-nash (changes requested):

m-nash (approved):

PR #29734, opened 2022-07-08
Enable file case duplication check in .NET

### LINKED ISSUE #3109: Block PRs with case-insensitive file collisions
A case-insensitive file collision was recently introduced into the .NET repo in https://github.com/Azure/azure-sdk-for-net/pull/27961.  When you clone the repo on Windows or Mac with default settings (case-insensitive file system), you get this warning:
 
```
Updating files: 100% (68668/68668), done.
warning: the following paths have collided (e.g. case-sensitive paths
on a case-insensitive filesystem) and only one from the same
colliding group is in the working tree:
  'sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Generated/Models/OCIManifest.cs'
  'sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Generated/Models/OciManifest.cs'
```

This has happened a few times now, and it's usually easy to fix once detected, but it can go days until someone reports it.

We should investigate if it's possible to detect this during PR validation and block on failures.  Ideally we could block this at the Git or GitHub level, but I didn't see any settings for this in GitHub, and a Git-level solution might be client-dependent.

sima-zhu:

Add check step to avoid check in the duplicated files only differ in case.

Resolve: https://github.com/Azure/azure-sdk-tools/issues/3109

Testing on dup:
image

### DIFF
--- a/eng/pipelines/templates/jobs/ci.yml
+++ b/eng/pipelines/templates/jobs/ci.yml
@@ -151,6 +151,7 @@ jobs:
               Paths:
                 - "/*"
                 - "!SessionRecords"
+      - template: /eng/common/pipelines/templates/steps/validate-filename.yml
       - template: /eng/common/pipelines/templates/steps/check-spelling.yml
       - task: UsePythonVersion@0
         displayName: "Use Python 3.6"


eng/pipelines/templates/jobs/ci.yml
```diff
@@ -151,6 +151,7 @@ jobs:
               Paths:
                 - "/*"
                 - "!SessionRecords"
+      - template: /eng/common/pipelines/templates/steps/validate-filename.yml
```
weshaggard, line 154: Have you done any testing of this when using sparse checkout? Probably worth validating the git commands we are using work correctly in this case.

weshaggard (approved):
Change looks fine once validation is complete.

PR #29740, opened 2022-07-08
[Event Hubs] Flaky Test Fix

jsquire:
# Summary

The focus of these changes is to fix a test believed to be flaky due to timing issues around callback execution and ignoring of a cancellation token.

### DIFF
--- a/sdk/eventhub/Azure.Messaging.EventHubs/tests/Primitives/EventProcessorTests.StartStop.cs
+++ b/sdk/eventhub/Azure.Messaging.EventHubs/tests/Primitives/EventProcessorTests.StartStop.cs
@@ -637,7 +637,11 @@ namespace Azure.Messaging.EventHubs.Tests
 
             mockConsumer
                 .Setup(consumer => consumer.ReceiveAsync(It.IsAny<int>(), It.IsAny<TimeSpan?>(), It.IsAny<CancellationToken>()))
-                .ReturnsAsync(new List<EventData> { new EventData(new BinaryData(Array.Empty<byte>())), new EventData(new BinaryData(Array.Empty<byte>())) });
+                .Returns<int, object, CancellationToken>((_, _, cancelToken) =>
+                {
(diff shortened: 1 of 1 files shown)


sdk/eventhub/Azure.Messaging.EventHubs/tests/Primitives/EventProcessorTests.StartStop.cs
```diff
@@ -654,7 +657,8 @@ public async Task StopProcessingStopsProcessing(bool async)
                     }
 
                     completionSource.TrySetResult(true);
-                });
+                })
+                .Returns(Task.CompletedTask);
```
jsquire, line 661: I believe this may have been the cause of the flakiness.   Because the callback was invoked after the call returned, the processor would have already gotten unblocked on moving forward and may have set `stopInvoked` before the callback was executed.

sdk/eventhub/Azure.Messaging.EventHubs/tests/Primitives/EventProcessorTests.StartStop.cs
```diff
@@ -637,15 +637,18 @@ public async Task StopProcessingStopsProcessing(bool async)
 
             mockConsumer
                 .Setup(consumer => consumer.ReceiveAsync(It.IsAny<int>(), It.IsAny<TimeSpan?>(), It.IsAny<CancellationToken>()))
-                .ReturnsAsync(new List<EventData> { new EventData(new BinaryData(Array.Empty<byte>())), new EventData(new BinaryData(Array.Empty<byte>())) });
+                .Returns<int, object, CancellationToken>((_, _, cancelToken) =>
+                {
+                    cancelToken.ThrowIfCancellationRequested<TaskCanceledException>();
```
jsquire, line 642: This more closely mimics the actual behavior.

JoshLove-msft (approved):

PR #29747, opened 2022-07-09
Address apiview comments for Compute, Network and Cdn

ArthurMa1978:
[Cdn] SslProtocol -> DeliveryRuleSslProtocol
[Cdn] SslProtocolMatchCondition -> DeliveryRuleSslProtocolMatchCondition
[Cdn] CacheLevel -> CdnCacheLevel
[Network] HttpConfiguration -> NetworkHttpConfiguration
[Network] HttpConfigurationMethod -> NetworkHttpConfigurationMethod
[Network] HttpMethod -> NetworkWatcherHttpMethod
[Network] IPConfigurationProfile -> NetworkIPConfigurationProfile
[Network] IPConfigurationBgpPeeringAddress -> NetworkIPConfigurationBgpPeeringAddress
[Network] IpAllocationType -> NetworkIPAllocationType
[Network] IpAllocationListResult -> NetworkIPAllocationListResult


sdk/compute/Azure.ResourceManager.Compute/src/autorest.md
```diff
@@ -191,6 +191,9 @@ rename-mapping:
   UpgradePolicy: VirtualMachineScaleSetUpgradePolicy
   ResourceSkuCapabilities: ComputeResourceSkuCapabilities
   ProtocolTypes: WinRMListenerProtocolType
+  VMGuestPatchRebootBehavior: VirtualMachineGuestPatchRebootBehavior
+  VMGuestPatchRebootSetting: VirtualMachineGuestPatchRebootSetting
+  VMGuestPatchRebootStatus: VirtualMachineGuestPatchRebootStatus
```
ArcturusZhang, line 196: I did not change these three because I think VmGuest is kind of a fixed phrase, it looks really weird we have both VmGuest (see above) and VirtualMachineGuest.

ArcturusZhang (approved):

sdk/cdn/Azure.ResourceManager.Cdn/api/Azure.ResourceManager.Cdn.netstandard2.0.cs
```diff
@@ -862,20 +862,20 @@ public partial class CacheKeyQueryStringActionProperties
         public override string ToString() { throw null; }
     }
     [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
-    public readonly partial struct CacheLevel : System.IEquatable<Azure.ResourceManager.Cdn.Models.CacheLevel>
+    public readonly partial struct CdnCacheLevel : System.IEquatable<Azure.ResourceManager.Cdn.Models.CdnCacheLevel>
```
Yao725, line 865: Can we leave it as it is? Prepending the RP prefix here doesn't feel right. We can keep it as we follow the guide that first come first serve.

sdk/compute/Azure.ResourceManager.Compute/src/autorest.md
ArthurMa1978: According to the description from the swagger spec, we can remove `Guest` here

sdk/cdn/Azure.ResourceManager.Cdn/api/Azure.ResourceManager.Cdn.netstandard2.0.cs
ArthurMa1978: This is cdn cache, so I think `Cdn` prefix is more accurate

sdk/compute/Azure.ResourceManager.Compute/src/autorest.md
```diff
@@ -172,8 +172,6 @@ rename-mapping:
   VMDiskSecurityProfile: VirtualMachineDiskSecurityProfile
   VmDiskTypes: VirtualMachineDiskType
   VMGalleryApplication: VirtualMachineGalleryApplication
-  VMGuestPatchClassificationLinux: LinuxVmGuestPatchClassification
-  VMGuestPatchClassificationWindows: WindowsVmGuestPatchClassification
```
ArcturusZhang: We should not remove these two, or we need to add the later to the list of preserve plural enum names - because the windows suffix is changed to window here which makes it no longer make sense

PR #29748, opened 2022-07-09
[Sql] Resource renaming

fengzhou-msft:


ArthurMa1978 (approved):

TianchengGuo:
1. “your username will have to be 9 characters long” --> "your username will have to be less than 9 characters long". This change will be easier to understand.
2. "you will then need to first delete directory" --> "you need to first delete directory." Clearer expression and adding a full stop.
3. "See (README file)[(https://github.com/Azure/azure-sdk-for-net/blob/main/eng/templates/README.md)]." It does not be translated into proper .md format.
4. "```C# Snippet:<snippetName>```". It also has .md format problem.
5. "Projects within the solution will use the feed defined in that file." --> "Projects within the solution will firstly use the feed defined in that file." Give a clearer concept of priority.
6. "<service_directory>" --> "<SerivceDirectory>". Should be same expression format.
7. "< ServiceName >" and "< Resource Provider Name >". I think there will be not space after "<" and before ">". It is better to have same format as previous, like "<PackageName>".
8. "$(RequiredTargetFrameworks) (defined in eng/Directory.Build.Data.props) ". I think this maybe another .md format problem?
9. "and /or" --> "and/or". Delete a space here.
10. "this step requires Node 14" --> "this step requires NodeJS 14".
11. The picture after "Request Changes" is somehow annoying.
12. "Client Library Tested OSs and .NET Versions" session. Maybe provide some versions of SDK that meet the requirements are better.

PR #29787, opened 2022-07-11
[Event Hubs] Update Scale Monitor

### LINKED ISSUE #29810: [QUERY] Microsoft.Azure.WebJobs EventHubTrigger PartitionContext empty except PartitionId
### Library name and version

Microsoft.Azure.WebJobs.Extensions.EventHubs 5.1.1

### Query/Question

Porting from 4.0.1

Previously, PartitionContext had even methods, but now it is properties only.  Method sig is simplified:

`EventData[] events, PartitionContext PartitionContext`

What I get though is this:
image

Basically everything empty except for PartitionId - is this supposed to be the case? Esp. EventHubName and ConsumerGroup is very surprising to be empty (and we rely on that in our code). One internal var seems to have most information... any help appreciated.

### Environment

_No response_

jsquire:
# Summary

The focus of these changes is to update the scale monitor for the Event Hubs Function extension to correct some left over T1 patterns that are no longer applicable.  For example, it is valid to have null checkpoints for a partition, but not possible to have a checkpoint with null offset. Additionally, correcting the assumption that the beginning sequence number for a partition is always 0.  This may not be true if the partition has been previously used and events have aged out, such as for very low volume or DR/fail-over scenarios.

Riding along is a fix for how the `PartitionContext` passed around is populated; the extended set of properties added in v5.7.0 had been ignored by the extension.

### DIFF
--- a/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubsScaleMonitor.cs
+++ b/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubsScaleMonitor.cs
@@ -89,8 +89,9 @@ namespace Microsoft.Azure.WebJobs.EventHubs.Listeners
 
             try
             {
-                checkpoints = await Task.WhenAll(checkpointTasks)
-                    .ConfigureAwait(false);
+                checkpoints = (await Task.WhenAll(checkpointTasks).ConfigureAwait(false))
+                    .Where(c => c != null)
+                    .ToArray();
             }
             catch
             {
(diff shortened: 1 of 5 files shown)


jsquire:

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubsScaleMonitor.cs
```diff
@@ -89,8 +89,9 @@ public async Task<EventHubsTriggerMetrics> GetMetricsAsync()
 
             try
             {
-                checkpoints = await Task.WhenAll(checkpointTasks)
-                    .ConfigureAwait(false);
+                checkpoints = (await Task.WhenAll(checkpointTasks).ConfigureAwait(false))
+                    .Where(c => c != null)
+                    .ToArray();
```
JoshLove-msft, line 94: Should we avoid allocating and just add the condition into the SingleOrDefault clause?

JoshLove-msft (approved):

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubsScaleMonitor.cs
```diff
@@ -89,8 +89,8 @@ public async Task<EventHubsTriggerMetrics> GetMetricsAsync()
 
             try
             {
-                checkpoints = await Task.WhenAll(checkpointTasks)
-                    .ConfigureAwait(false);
+                checkpoints = (await Task.WhenAll(checkpointTasks).ConfigureAwait(false))
```
JoshLove-msft, line 92: Hmm.. could Task.WhenAll actually return null when awaited?

jsquire: It can, but only in the case where the partition information reports no partitions.   Under normal circumstances, this should not occur but is possible in some odd corner cases.

jsquire: actually... I'm incorrect.   Even when there are no tasks to await, it seems that it emits an empty array.   I'll update to remove the guard.

PR #29798, opened 2022-07-12
Apply 2388: Put swagger summary into comment summary

pshao25:
Apply https://github.com/Azure/autorest.csharp/pull/2425


ArcturusZhang (approved):

archerzz (commented):
The only problem I can see is that if `summary` and `remarks` are somehow the same, we'll duplicate the contents, e.g. we will generate a `<remarks>` section with the same contents in the final document.

lirenhe (approved):
LGTM.  Please also confirm with Alexander. Thanks

sdk/attestation/Azure.Security.Attestation/src/Generated/PolicyCertificatesRestClient.cs
```diff
@@ -54,6 +54,7 @@ internal HttpMessage CreateGetRequest()
 
         /// <summary> Retrieves the set of certificates used to express policy for the current tenant. </summary>
         /// <param name="cancellationToken"> The cancellation token to use. </param>
+        /// <remarks> Retrieves the set of certificates used to express policy for the current tenant. </remarks>
```
jsquire, line 57: This doesn't look right to me.   The summary and remarks shouldn't be repeated.   Not an issue for an internal client type such as this, but it would be a concern for a pure DPG client.

jsquire (approved):

sdk/anomalydetector/Azure.AI.AnomalyDetector/src/Generated/AnomalyDetectorClient.cs
```diff
@@ -88,9 +88,10 @@ internal AnomalyDetectorClient(ClientDiagnostics clientDiagnostics, HttpPipeline
             _pipeline = pipeline;
         }
 
-        /// <summary> This operation generates a model with an entire series, each point is detected with the same model. With this method, points before and after a certain point are used to determine whether it is an anomaly. The entire detection can give user an overall status of the time series. </summary>
+        /// <summary> Detect anomalies for the entire series in batch. </summary>
```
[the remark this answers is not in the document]
jsquire: I'd call this non-blocking here, as the impacted types are internal.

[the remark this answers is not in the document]
pshao25: I totally agree with that, but this is indeed what we intend to achieve for this issue.

pshao25:
> > The only problem I can see is that if `summary` and `remarks` are somehow the same, we'll duplicate the contents, e.g. we will generate a `<remarks>` section with the same contents in the final document.
> 
> I strongly agree with this. Since the impacted files here are `internal`, I don't want to block this, but I do believe that we need a follow-up to fix the duplication.

As we discussed with @lirenhe, it's better to have duplicate contents than losing information from users. Because of the M4 issue discussed [here](https://github.com/Azure/azure-sdk-for-net/pull/29798#discussion_r923367416),  we don't want to lose description information if it really has.

sdk/attestation/Azure.Security.Attestation/src/Generated/PolicyCertificatesRestClient.cs
jsquire: If we change the M4 behavior, it would potentially be a breaking change with a wider impact.   While I, personally, disagree with the M4 behavior, it's a safer change to include this downstream since these are new changes.

jsquire:
> we don't want to lose description information if it really has (i.e., user input the same text in both "description" and "summary").

I disagree; if the same exact text appears in both places, there is no loss of information rendering in a single element.

AlexanderSher:
> > we don't want to lose description information if it really has (i.e., user input the same text in both "description" and "summary").
> 
> I disagree; if the same exact text appears in both places, there is no loss of information rendering in a single element.

Agree. There is no reason to write the same text twice. cc: @annelo-msft

sdk/confidentialledger/Azure.Security.ConfidentialLedger/src/Generated/ConfidentialLedgerCertificateClient.cs
```diff
@@ -42,6 +42,8 @@ protected ConfidentialLedgerCertificateClient()
         /// <exception cref="RequestFailedException"> Service returned a non-success status code. </exception>
         /// <returns> The response returned from the service. Details of the response body schema are in the Remarks section below. </returns>
         /// <remarks>
+        /// Gets identity information for a Confidential Ledger instance.
```
christothes, line 45: What's the value of adding this text that is already in the summary section? Do we have an example of where they are different?

sdk/webpubsub/Azure.Messaging.WebPubSub/src/Generated/HealthApiClient.cs
```diff
@@ -57,6 +57,7 @@ public HealthApiClient(string endpoint, WebPubSubServiceClientOptions options)
         /// <param name="context"> The request context, which can override default behaviors of the client pipeline on a per-call basis. </param>
         /// <exception cref="RequestFailedException"> Service returned a non-success status code. </exception>
         /// <returns> The response returned from the service. </returns>
+        /// <remarks> Get service health status. </remarks>
```
annelo-msft, line 60: Here is an instance of a public API where `remarks` is being added because of duplicated text.  I agree with @jsquire and @AlexanderSher that if the text is identical, there is no reason to add the `remarks` block and no loss of information.

pshao25:
Thank you all! I will merge this PR and fix the duplicate issue in https://github.com/Azure/autorest.csharp/issues/2435 to unblock others.

PR #29799, opened 2022-07-12
fix(sdk): incorrect `ServiceVresion` in `Azure.MixedReality.Authentication`

archerzz:
Per SDK guideline, the variable name should be pascal case. So `V2019_02_28_preview` should be `V2019_02_28_Preview` (with upper case `P`). Since `Azure.MixedReality.Authentication` is already GAed, we cannot simply update the enum definition.

This commit will fix the case problem and keep backward compatibility by:

- add the correct enum definition `V2019_02_28_Preview` with the same value `1`
- hide the problematic enum definition `V2019_02_28_preview` using attribute `[EditorBrowsable(EditorBrowsableState.Never)]`
- change the default value and service version selection in `MixedRealityStsClientOptions()`

part of https://github.com/Azure/autorest.csharp/issues/1524

### DIFF
--- a/sdk/mixedreality/Azure.MixedReality.Authentication/src/MixedRealityStsClientOptions.cs
+++ b/sdk/mixedreality/Azure.MixedReality.Authentication/src/MixedRealityStsClientOptions.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
+using System.ComponentModel;
 using Azure.Core;
 
 namespace Azure.MixedReality.Authentication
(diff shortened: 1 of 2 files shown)


sdk/mixedreality/Azure.MixedReality.Authentication/src/MixedRealityStsClientOptions.cs
```diff
@@ -19,11 +20,11 @@ public class MixedRealityStsClientOptions : ClientOptions
         /// Initializes a new instance of the <see cref="MixedRealityStsClientOptions"/> class.
         /// </summary>
         /// <param name="version">The version.</param>
-        public MixedRealityStsClientOptions(ServiceVersion version = ServiceVersion.V2019_02_28_preview)
+        public MixedRealityStsClientOptions(ServiceVersion version = ServiceVersion.V2019_02_28_Preview)
         {
             Version = version switch
             {
-                ServiceVersion.V2019_02_28_preview => "2019-02-28-preview",
+                ServiceVersion.V2019_02_28_Preview => "2019-02-28-preview",
```
jsquire, line 27: We're missing the second case.   `V2019_02_28_preview` needs to have a mapping.

jsquire (changes requested):

sdk/mixedreality/Azure.MixedReality.Authentication/src/MixedRealityStsClientOptions.cs
jsquire: I stand corrected; apparently, this is doing a value-based match and works.

jsquire (approved):

PR #29802, opened 2022-07-12
verify test generation in pipeline

changlong-liu:
Fixing https://github.com/Azure/autorest.csharp/issues/2380

This PR adds test generation verification in pipeline.
After this PR, the pipeline can check that the generated tests are latest like what we do now for source code (for any packages enabled test generation).


eng/service.proj
```diff
@@ -51,6 +51,14 @@
              SkipNonexistentTargets="true" />
   </Target>
 
+  <Target Name="GenerateTest">
```
heaths, line 54: Nit: "GenerateTests" (plural) sounds more apt. Your call to `Write-Host` above even uses the plural form since we'll certainly generate numerous tests.

heaths (approved):

eng/scripts/CodeChecks.ps1
```diff
@@ -84,6 +84,11 @@ try {
         Invoke-Block {
             & dotnet msbuild $PSScriptRoot\..\service.proj /restore /t:GenerateCode /p:SDKType=$SDKType /p:ServiceDirectory=$ServiceDirectory
         }
+
+        Write-Host "Re-generating tests"
+        Invoke-Block {
+            & dotnet msbuild $PSScriptRoot\..\service.proj /restore /t:GenerateTest /p:SDKType=$SDKType /p:ServiceDirectory=$ServiceDirectory
```
benbp, line 90:
```suggestion
            & dotnet msbuild $PSScriptRoot/../service.proj /restore /t:GenerateTest /p:SDKType=$SDKType /p:ServiceDirectory=$ServiceDirectory
```

eng/service.proj
changlong-liu: done

eng/scripts/CodeChecks.ps1
changlong-liu: done

m-nash (approved):

eng/service.proj
```diff
@@ -51,6 +51,14 @@
              SkipNonexistentTargets="true" />
   </Target>
 
+  <Target Name="GenerateTests">
+    <MSBuild Projects="@(ProjectReference)"
+             Targets="GenerateTest"
```
weshaggard, line 56: Is there already a GenerateTest target in the library targets somewhere? A quick search didn't show anything.

changlong-liu: Yes @weshaggard, there already be a GenerateTest target. It searches "autorest.tests.md" in tests folder and generate test. The file is expected to be like this: https://github.com/Azure/azure-sdk-for-net/blob/2a79d3ba5592bdac5456d8b0fd7a8fa1e6433e1c/sdk/servicebus/Azure.ResourceManager.ServiceBus/tests/autorest.tests.md

For now, we didn't commit "autorest.tests.md" to any RPs yet, so it generates nothing currently. It's expected to be enabled after we finish scenario test generation development.

PR #29814, opened 2022-07-12
Consolidate SystemAssignedServiceIdentity into ManagedServiceIdentity

fengzhou-msft:


sdk/core/Azure.Core/src/Shared/PropertyReferenceTypeAttribute.cs
```diff
@@ -14,23 +14,23 @@ internal class PropertyReferenceTypeAttribute : Attribute
         /// <summary>
         /// Instatiate a new reference type attribute.
         /// </summary>
-        /// <param name="skipTypes"> An array of types to skip for this reference type. </param>
-        public PropertyReferenceTypeAttribute(Type[] skipTypes)
+        /// <param name="optionalProperties"> An array of property names that are optional when comparing the type. </param>
+        public PropertyReferenceTypeAttribute(string[] optionalProperties)
```
fengzhou-msft, line 18: This is a breaking change, but nothing actually uses `skipTypes`.

sdk/resourcemanager/Azure.ResourceManager/src/Azure.ResourceManager.csproj
```diff
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <Version>1.3.0-beta.1</Version>
     <!--The ApiCompatVersion is managed automatically and should not generally be modified manually.-->
-    <ApiCompatVersion>1.2.0</ApiCompatVersion>
+    <!-- <ApiCompatVersion>1.2.0</ApiCompatVersion> -->
```
fengzhou-msft, line 6: Uncomment this to fix all CIs first. Need to decide how to proceed later.

sdk/core/Azure.Core/src/Shared/PropertyReferenceTypeAttribute.cs
fengzhou-msft: It seems this is an internal class change and not breaking the ApiCompatVersion check.

sdk/resourcemanager/Azure.ResourceManager/src/Resources/Custom/PolicyAssignmentData.cs
```diff
@@ -57,8 +57,19 @@ internal PolicyAssignmentData(ResourceIdentifier id, string name, ResourceType r
...
         /// <summary> The location of the policy assignment. Only required when utilizing managed identity. </summary>
         public AzureLocation? Location { get; set; }
+
+        private SystemAssignedServiceIdentity _identity;
         /// <summary> The managed identity associated with the policy assignment. </summary>
-        public SystemAssignedServiceIdentity Identity { get; set; }
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public SystemAssignedServiceIdentity Identity
+        {
+            get => _identity ??= ManagedIdentity == null ? null : new SystemAssignedServiceIdentity(ManagedIdentity.PrincipalId, ManagedIdentity.TenantId, ManagedIdentity.ManagedServiceIdentityType.ToString());
```
m-nash, line 66: I feel like we need some rigorous test cases around these conversions to make sure there are no errors and they persist.

If we are making a new one every time wouldn't I lose my changes after I did a set?  Also what happens if ManagedIdentity is null?

Lets add the test cases to answer these questions.

fengzhou-msft: With more tests, I found it very difficult to always keep Identity and ManagedIdentity data in sync with the way to update identity with `data.Identity.SystemAssignedServiceIdentityType = ...` as it's beyond the reach of setters in PolicyAssignmentData. Therefore, I updated the code to assume a user will only either use Identity or ManagedIdentity which I think is a reasonable assumption and added tests for different scenarios.

sdk/resourcemanager/Azure.ResourceManager/src/Resources/Custom/PolicyAssignmentData.cs
```diff
@@ -57,8 +57,42 @@ internal PolicyAssignmentData(ResourceIdentifier id, string name, ResourceType r
...
+        {
+            get => _managedIdentity;
+            set
+            {
+                _managedIdentity = value;
+                if (value == null)
+                    _identity = null;
+                else
+                {
+                    if (_identity == null)
```
m-nash, line 88: can't we just have line 90?  Since we just need to change the internal memory the reference of SystemAssignedServiceIdentity doesn't need to change here.

m-nash (approved):
Please validate my last comment before merging otherwise this looks good.

sdk/resourcemanager/Azure.ResourceManager/src/Resources/Custom/PolicyAssignmentData.cs
fengzhou-msft: In this codnition branch when we set a non-null value to ManagedIdentity, _identity could initially be null, therefore we need to create a new instance for it if it's null before we can update its internal memory of ManagedServiceIdentity.

PR #29822, opened 2022-07-12
MR Auth: Updated dependencies

craigktreasure:
- This change updates the common dependency version for `System.IdentityModel.Tokens.Jwt` to `6.5.0`, which is the first version to remove the dependency on `Newtonsoft.Json`.

### DIFF
--- a/eng/Packages.Data.props
+++ b/eng/Packages.Data.props
@@ -108,7 +108,7 @@
     <PackageReference Update="Microsoft.Identity.Client.Extensions.Msal" Version="2.19.3" />
 
     <!-- TODO: Make sure this package is arch-board approved -->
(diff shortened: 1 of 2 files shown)


heaths (approved):
I've confirmed that the only track 2 shipping package affected is `Azure.MixedReality.Authentication` so this should be fine. @KrzysztofCwalina @tg-msft @schaabs see any problems with updating this dependency, which itself removes any explicit dependency on `Newtonsoft.Json` (there's a problem with < 13.0.1).

craigktreasure:
@KrzysztofCwalina, @tg-msft, @schaabs, any concerns here?

KrzysztofCwalina:
I don't have any concerns with the dependency change.

Issue #27143, opened 2022-02-22
Custom endpoint for connection establishment 


Joseandresc:
### Library name

Azure.Messaging.ServiceBus

### Please describe the feature.

Currently there is no way to specify an endpoint other than the FQDN of the namespace, specificying a custom endpoint is useful in scenarios where customers might not to connect through a custom domain that points to a Firewall, proxy, waf, etc. 
This feature was implemented in the Event Hubs SDK https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Azure.Messaging.EventHubs/src/EventHubConnectionOptions.cs#L108, would be great to also implement such feature within the Service Bus SDK.

jsquire:
//fyi:  @ramya-rao-a

jsquire:
Thank you for your feedback.  Tagging and routing to the team member best able to assist.

JoshLove-msft:
This has been addressed and will be available in an upcoming release.

PR #18014, opened 2021-01-15
[Event Hubs Client] Custom Endpoint Support

jsquire:
# Summary

The focus of these changes is to add an option for specifying a custom address to use for establishing the connection to the Event Hubs service; the namespace address will continue to be used for communication with the service once the connection is established when the custom endpoint is configured.

**Note:** These changes have discussed with, and approved, the .NET language architect. Marking with the corresponding tag for tracking purposes.

# Last Upstream Rebase

Friday, January 15, 2:56pm (EST)

# References and Related

- [Board Review: Event Hubs - Custom Endpoint Support (#1367)](https://github.com/Azure/azure-sdk/issues/1367)
- [Design: Event Hubs Custom Endpoint Support](https://gist.github.com/jsquire/f0948c9a5ffe2741b03d312dcf8d899e)
- [Custom Endpoint Address Support (#12555)](https://github.com/Azure/azure-sdk-for-net/issues/136712555)

PR #29823, opened 2022-07-12
Service Bus Custom Endpoint Support

m-redding:
The intention of these changes is to provide support for Custom Endpoint use when establishing a connection with the Service Bus service. The fully qualified namespace is still required after the initial connection, but this allows for easier connection in circumstances such as when routing through an application gateway or using a proxy.

See https://github.com/Azure/azure-sdk-for-net/issues/27143

### DIFF
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpClient.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpClient.cs
@@ -50,7 +50,7 @@ namespace Azure.Messaging.ServiceBus.Amqp
         public override Uri ServiceEndpoint { get; }
 
         /// <summary>
-        ///   The endpoint for the Service Bus service to be used when establising the connection.
+        ///   The endpoint for the Service Bus service to be used when establishing the connection.
         /// </summary>
         ///
         public Uri ConnectionEndpoint { get; }
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpClient.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpClient.cs
@@ -50,6 +50,12 @@ namespace Azure.Messaging.ServiceBus.Amqp
         public override Uri ServiceEndpoint { get; }
 
         /// <summary>
+        ///   The endpoint for the Service Bus service to be used when establising the connection.
+        /// </summary>
+        ///
+        public Uri ConnectionEndpoint { get; }
+
+        /// <summary>
         ///   Gets the credential to use for authorization with the Service Bus service.
         /// </summary>
         ///
@@ -95,6 +101,12 @@ namespace Azure.Messaging.ServiceBus.Amqp
                 Host = host
             }.Uri;
 
+            ConnectionEndpoint = (options.CustomEndpointAddress == null) ? ServiceEndpoint : new UriBuilder
+            {
+                Scheme = ServiceEndpoint.Scheme,
+                Host = options.CustomEndpointAddress.Host
+            }.Uri;
+
             Credential = credential;
             if (options.EnableTransportMetrics)
             {
@@ -102,6 +114,7 @@ namespace Azure.Messaging.ServiceBus.Amqp
             }
             ConnectionScope = new AmqpConnectionScope(
                 ServiceEndpoint,
+                ConnectionEndpoint,
                 credential,
                 options.TransportType,
                 options.WebProxy,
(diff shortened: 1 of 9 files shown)


sdk/servicebus/Azure.Messaging.ServiceBus/tests/Samples/Sample01_HelloWorld.cs
```diff
@@ -264,6 +264,25 @@ public void AuthenticateWithConnectionString()
...
 
+        /// <summary>
+        /// Connect to the service using a custom endpoint address/>.
+        /// </summary>
+        public void ConnectUsingCustomEndpoint()
+        {
+            #region Snippet:ServiceBusCustomEndpoint
+            // Connect to the service using a custom endpoint
+            string connectionString = "<connection_string>";
+            string customEndpoint = "<custom_endpoint";
```
JoshLove-msft, line 275:
```suggestion
            string customEndpoint = "<custom_endpoint>";
```

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpClient.cs
```diff
@@ -49,6 +49,12 @@ internal class AmqpClient : TransportClient
         ///
         public override Uri ServiceEndpoint { get; }
 
+        /// <summary>
+        ///   The endpoint for the Service Bus service to be used when establising the connection.
```
JoshLove-msft, line 53:
```suggestion
        ///   The endpoint for the Service Bus service to be used when establishing the connection.
```

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Client/ServiceBusClientTests.cs
```diff
@@ -245,6 +246,7 @@ public void ConstructorValidatesConnectionStringForDuplicateAuthorization(string
             Assert.That(options, Is.Not.SameAs(constructorOptions), $"The { constructorDescription } constructor should have cloned the options.");
             Assert.That(options.TransportType, Is.EqualTo(constructorOptions.TransportType), $"The { constructorDescription } constructor should have the correct connection type.");
             Assert.That(options.WebProxy, Is.EqualTo(constructorOptions.WebProxy), $"The { constructorDescription } constructor should have the correct proxy.");
+            Assert.That(options.CustomEndpointAddress, Is.EqualTo(constructorOptions.CustomEndpointAddress), $"The {constructorDescription} constructor should have the correct custom endpoint.");
```
JoshLove-msft, line 249: Hmm.. do we need to update the test cases to set the endpoint address?

JoshLove-msft (approved):
LGTM!

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Client/ServiceBusClientTests.cs
```diff
@@ -226,6 +226,7 @@ public void ConstructorValidatesConnectionStringForDuplicateAuthorization(string
             Assert.That(options, Is.Not.SameAs(defaultOptions), $"The { constructorDescription } constructor should not have the same options instance.");
```
jsquire, line 226: We should also consider bringing these over if we don't have equivalents.... these were added as part of the custom endpoint address work in Event Hubs:   
https://github.com/Azure/azure-sdk-for-net/pull/18014/files#diff-d317b71217720aff41196242e8f784de2e9cc76d78f0952785198e2a13ede7f0

jsquire (approved):
Nicely done!

jsquire:
Looks like analyze is cranky about code gen.   Running the snippets script should unblock.

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Client/ServiceBusClientTests.cs
```diff
@@ -226,6 +226,7 @@ public void ConstructorValidatesConnectionStringForDuplicateAuthorization(string
...
             Assert.That(options.TransportType, Is.EqualTo(defaultOptions.TransportType), $"The { constructorDescription } constructor should have the correct connection type.");
             Assert.That(options.WebProxy, Is.EqualTo(defaultOptions.WebProxy), $"The { constructorDescription } constructor should have the correct proxy.");
+            Assert.That(options.CustomEndpointAddress, Is.EqualTo(defaultOptions.CustomEndpointAddress), $"The {constructorDescription} constructor should have the correct custom endpoint.");
```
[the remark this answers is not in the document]
m-redding: @JoshLove-msft I'm not sure, it didn't look like there were any explicit options being set there so I think just checking that the default is being set would work, but is there something else that you were thinking of for this case?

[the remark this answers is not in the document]
JoshLove-msft: Nevermind, I see this test is just checking the defaults, so it makes sense as is!

m-redding:
Resolves #28350

PR #29857, opened 2022-07-13
Minwoolee/callingserverv3 fixclients

minwoolee-msft:
From Feedbacks and suggestion, Here are the latest updates for Server Calling: 

- Have multiple Clients for each groups of methods: CallConnectionClient, CallRecordingClient, CallContentclient
- New Clients can be fetched from Get[ClientName], from root service client
- Updated other download helpers to work with CallRecordingClient
- CallConnectionProperties is now separate object that only contains data of the call. All methods are moved into CallConnectionClient
- AnswerCall, MakeCall, GetCall now returns CallConnectionProperties instead of CallConnectionClient
- Removed reason as input from TerminateCall (this was never used)
- GetParticipants now returns participants in IReadOnlyList
- Various Request objects are now internal, hidden from contoso (i.e. RejectCallRequest)
- Dropped unnecessary suffix, like Dto and Model (i.e.  CallConnectionStateModel ->  CallConnectionState)
- Options (such as AddParticipantsOptions) constructor now only takes optional inputs. All other options updated to be similar
- CallLocator now updated to be Abstract class. Derived classes (GroupCallLocator & ServerCallLocator) Added.
- Moved most of Models into correct namespace
- Deleted unused models
- Various small typeo fixes


sdk/communication/Azure.Communication.CallingServer/src/CallingServerClient.cs
```diff
@@ -425,245 +408,15 @@ public virtual Response<CallConnection> GetCall(string callConnectionId, Cancell
...
-        }
-
-        /// <summary>
-        /// resume recording of the call.
-        /// </summary>
-        /// <param name="recordingId">The recording id to resume.</param>
-        /// <param name="cancellationToken">The cancellation token.</param>
-        public virtual Response ResumeRecording(string recordingId, CancellationToken cancellationToken = default)
+        /// <summary> Initializes a new instance of CallConnectionClient. <see cref="GetCallConnectionClient"/>.</summary>
+        /// <param name="callConnectionId"> The thread id for the ChatThreadClient instance. </param>
```
yyw-msft, line 412: update callConnectionId description?

sdk/communication/Azure.Communication.CallingServer/src/CallingServerClient.cs
```diff
@@ -672,237 +425,14 @@ public virtual Response ResumeRecording(string recordingId, CancellationToken ca
...
-        /// </param>
-        /// <returns>
-        /// A <see cref="Response"/> describing the operation.
-        /// </returns>
-        /// <remarks>
-        /// A <see cref="RequestFailedException"/> will be thrown if
-        /// a failure occurs.
-        /// </remarks>
-        public virtual Response DeleteRecording(Uri deleteEndpoint, CancellationToken cancellationToken = default)
+        /// <summary> Initializes a new instance of CallConnectionClient. <see cref="CallRecordingClient"/>.</summary>
```
yyw-msft, line 428: Initialize new CallRecordingClient?

yyw-msft (approved):

PR #29878, opened 2022-07-13
Changes after autorest.csharp PR 2424 is merged

AlexanderSher:


sdk/deviceupdate/Azure.IoT.DeviceUpdate/src/Generated/DeviceUpdateClient.cs
```diff
@@ -1055,7 +1055,7 @@ public virtual async Task<Operation> ImportUpdateAsync(WaitUntil waitUntil, Requ
         /// 
         /// Request Body:
         /// 
-        /// Schema for <c>ImportUpdateInput</c>:
+        /// Schema for <c>ImportUpdateInputItem</c>:
```
archerzz, line 1058: I think here the schema should be `ImportUpdateInput` which is an array of `ImportUpdateInputItem`. See the original [swagger definition](https://github.com/Azure/azure-rest-api-specs/blob/a9e895ccfe29d0646795f7ff1cb78e185bd09529/specification/deviceupdate/data-plane/Microsoft.DeviceUpdate/preview/2022-07-01-preview/deviceupdate.json#L2323) here.

archerzz: Look around the changes, it seems we delve into the array type request body, get the element type as the request schema type.

AlexanderSher: Yes, as you can see, the code below hasn't changed. So right now the name of the type is correct, however the `remarks` part as a whole is invalid. But it was invalid before.

PR #29882, opened 2022-07-13
[For prototype]Propose new way to write lines to file

sima-zhu:
`PropertyGroup` cannot serialize the array correctly. 
Use `WriteLinesToFiles` directly, and use their default utf-8 encoding.

The command line for this:
```
dotnet msbuild /t:ProjectDependsOn ./eng/service.proj /p:ServiceDirectory=* /p:TestTargetDependency="Azure.Core" /p:IncludeSrc=false /p:IncludeStress=false /p:IncludePerf=false
```


eng/service.proj
```diff
@@ -21,7 +21,7 @@
     <StressProjects Include="..\sdk\$(ServiceDirectory)\$(Project)\stress\**\*.csproj" />
     <SampleApplications Include="..\samples\**\*.csproj" />
     <SrcProjects Include="..\sdk\$(ServiceDirectory)\$(Project)\src\**\*.csproj" Exclude="@(TestProjects);@(SamplesProjects);@(PerfProjects);@(StressProjects)" />
-    <ProjectReference Include="@(TestProjects)" Exclude="@(MgmtExcludePaths)" Condition="'$(IncludeTests)' == 'true'" />
+    <ProjectReference Include="@(TestProjects)" Exclude="@(MgmtExcludePaths)" Condition="'$(IncludeTests)' == 'true' and '$(IsClientLibrary)' == 'true'" />
```
weshaggard, line 24: This isn't the correct place for this condition.

weshaggard: You will want to add it to the target in common which seems to be missing in your current changes.

eng/service.proj
```diff
@@ -11,17 +11,20 @@
     <IncludeStress Condition="'$(IncludeStress)' == ''">true</IncludeStress>
     <IncludeSamplesApplications Condition="'$(IncludeSamplesApplications)' == ''">true</IncludeSamplesApplications>
     <IncludeSamplesApplications Condition="'$(ServiceDirectory)' != '*' or '$(IncludeSamples)' == 'false'">false</IncludeSamplesApplications>
+    <IsClientLibrary Condition="'$(IsClientLibrary)' == ''">true</IsClientLibrary>
```
weshaggard, line 14: This is not going to be the correct place to set this property.

eng/service.proj
```diff
@@ -11,17 +11,20 @@
...
   </PropertyGroup>
 
   <ItemGroup>
     <MgmtExcludePaths Include="$(MSBuildThisFileDirectory)..\sdk\*\Microsoft.*.Management.*\**\*.csproj;$(MSBuildThisFileDirectory)..\sdk\*mgmt*\**\*.csproj" />
     <TestProjects Include="..\sdk\$(ServiceDirectory)\$(Project)\tests\**\*.csproj" />
+    <ClientTestProjects Include="..\sdk\$(ServiceDirectory)\$(Project)\tests\**\Azure.*csproj" />
```
weshaggard, line 20: We want to try and avoid this globbing because we want to support client libraries that don't start with Azure as well.

PR #29884, opened 2022-07-13
Address apiview comments for CosmosDB

ArthurMa1978:


sdk/cosmosdb/Azure.ResourceManager.CosmosDB/src/Generated/Models/CassandraCommandPostBody.cs
```diff
@@ -43,6 +43,6 @@ public CassandraCommandPostBody(string command, string host)
         /// <summary> If true, stops cassandra before executing the command and then start it again. </summary>
         public bool? CassandraStopStart { get; set; }
         /// <summary> If true, allows the command to *write* to the cassandra directory, otherwise read-only. </summary>
-        public bool? Readwrite { get; set; }
+        public bool? IsReadwrite { get; set; }
```
fengzhou-msft, line 46: Based on the description, `AllowReadWrite` may be a better name?

ArthurMa1978: Good suggestion, renamed to `AllowWrite`

ArcturusZhang (approved):
It would be nice if we could move those `rename-model` into `rename-mapping`

sdk/cosmosdb/Azure.ResourceManager.CosmosDB/src/Generated/Models/CosmosDBServerVersion.cs
```diff
@@ -22,16 +22,10 @@ public CosmosDBServerVersion(string value)
...
-        /// <summary> 3.2. </summary>
-        public static CosmosDBServerVersion Three2 { get; } = new CosmosDBServerVersion(Three2Value);
-        /// <summary> 3.6. </summary>
-        public static CosmosDBServerVersion Three6 { get; } = new CosmosDBServerVersion(Three6Value);
-        /// <summary> 4.0. </summary>
-        public static CosmosDBServerVersion Four0 { get; } = new CosmosDBServerVersion(Four0Value);
+        private const string V3_2Value = "3.2";
+        private const string V3_6Value = "3.6";
+        private const string V4_0Value = "4.0";
+        private const string V4_2Value = "4.2";
```
Yao725, line 28: Interesting, why we don't have `4.2` in the previous generated code?

Yao725 (approved):

sdk/cosmosdb/Azure.ResourceManager.CosmosDB/src/Generated/Models/CosmosDBServerVersion.cs
ArcturusZhang: Arthur updated the swagger commit

fengzhou-msft (approved):

sdk/cosmosdb/Azure.ResourceManager.CosmosDB/src/Generated/Models/CosmosDBServerVersion.cs
ArthurMa1978: That's the problem that we not use `readme.md` under spec repo

Issue #29889, opened 2022-07-14
[BUG] When using ServiceBusSessionProcessor If OnSessionInitializingAsync throws an exception messages are still processed


asanjabi:
### Library name and version

Azure.Messaging.ServiceBus 7.9.0

### Describe the bug

If user code throws an exception in `OnSessionInitializingAsync`, the processor still continues with processing messages even though initialization failed.

### Expected behavior

If user code throws an exception during initialization seems like the right thing to do is to cancel processing that session.

### Actual behavior

During the [session initialization step](https://github.com/Azure/azure-sdk-for-net/blob/5aee4c5f0b72d3bc3ff0bf36f0c194e2cd800465/sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/SessionReceiverManager.cs#L124), if the code throws an exception, it is caught at the [end of ReceiveAndProcessMessagesAsync ](https://github.com/Azure/azure-sdk-for-net/blob/5aee4c5f0b72d3bc3ff0bf36f0c194e2cd800465/sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/SessionReceiverManager.cs#L311) which only logs the event.

Unfortunately, the finally clause only cleans up if the call [was successful?](https://github.com/Azure/azure-sdk-for-net/blob/5aee4c5f0b72d3bc3ff0bf36f0c194e2cd800465/sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/SessionReceiverManager.cs#L341)  `CloseReceiverIfNeeded` won't clean this up properly anyhow since `_threadCount `only gets incremented if call [to initialize succeeds](https://github.com/Azure/azure-sdk-for-net/blob/5aee4c5f0b72d3bc3ff0bf36f0c194e2cd800465/sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/SessionReceiverManager.cs#L85), which gets skipped and during [clean-up it will be -1](https://github.com/Azure/azure-sdk-for-net/blob/5aee4c5f0b72d3bc3ff0bf36f0c194e2cd800465/sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/SessionReceiverManager.cs#L183).

[...]

PR #29908, opened 2022-07-14
Add ability to release/renew session from SessionInitialization handler

JoshLove-msft:
Addresses https://github.com/Azure/azure-sdk-for-net/issues/29889

### DIFF
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ProcessSessionEventArgs.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ProcessSessionEventArgs.cs
@@ -24,6 +24,8 @@ namespace Azure.Messaging.ServiceBus
         /// </summary>
         private readonly ServiceBusSessionReceiver _sessionReceiver;
 
+        private readonly SessionReceiverManager _manager;
+
         /// <summary>
         /// The Session Id associated with the session being processed.
         /// </summary>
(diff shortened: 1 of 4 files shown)


sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ProcessSessionEventArgs.cs
```diff
@@ -85,5 +97,21 @@ public class ProcessSessionEventArgs : EventArgs
...
+        /// session will be closed.
+        /// The session may end up being reopened for processing immediately after closing if there are messages remaining in the session (
+        /// This depends on what other session messages may be in the queue or subscription).
+        /// </summary>
+        public virtual void ReleaseSession() =>
+            // manager will be null if instance created using the public constructor which is exposed for testing purposes
+            _manager?.CancelSession();
+
+        ///<inheritdoc cref="ServiceBusSessionReceiver.RenewSessionLockAsync(CancellationToken)"/>
+        public virtual async Task RenewSessionLockAsync(CancellationToken cancellationToken = default)
```
jsquire, line 112: Interesting...  I guess this is for potentially long-running initializations?

jsquire (approved):

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ProcessSessionEventArgs.cs
JoshLove-msft: Yeah, I think it makes sense to include as well. Basically these handlers should have all non-message operations.

Issue #29871, opened 2022-07-13
SearchModelFactory.FacetResult parameter not documented


julyvee:
The parameter additionalProperties does not have any description. A list of possible properties, how they are used or at least some examples would be nice.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 23772ac0-d70d-4cec-cb13-c9c48a31c2a0
* Version Independent ID: da083860-3dd7-2650-3b35-28210b0cd827
* Content: [SearchModelFactory.FacetResult Method (Azure.Search.Documents.Models) - Azure for .NET Developers](https://docs.microsoft.com/en-us/dotnet/api/azure.search.documents.models.searchmodelfactory.facetresult?view=azure-dotnet)
* Content Source: [xml/Azure.Search.Documents.Models/SearchModelFactory.xml](https://github.com/Azure/azure-docs-sdk-dotnet/blob/master/xml/Azure.Search.Documents.Models/SearchModelFactory.xml)
* Service: **azure**
* GitHub Login: @rloutlaw
* Microsoft Alias: **routlaw**

jsquire:
Thank you for your feedback.  Tagging and routing to the team member best able to assist.

ShivangiReja:
Hi @julyvee, Thanks for filing this issue. For context [FacetResult](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/search/Azure.Search.Documents/src/Generated/Models/FacetResult.cs) is an output model(read- only model) is an output model(read-only model) that's why we provide [SearchModelFactory.FacetResult](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/search/Azure.Search.Documents/src/Models/SearchModelFactory.cs#L324) to mock [FacetResult](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/search/Azure.Search.Documents/src/Generated/Models/FacetResult.cs) type for testing. We provide Model factory for all read-only models to support mocking to enable non-live testing of service clients by customers.

Since [FacetResult](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/search/Azure.Search.Documents/src/Generated/Models/FacetResult.cs) is an output model(read- only model), user can not pass `additionalProperties`, the values comes from the service.

[...]

PR #29909, opened 2022-07-14
Use generated code for SearchModelFactory.FacetResult and remove manually added code

ShivangiReja:
Issue - https://github.com/Azure/azure-sdk-for-net/issues/29871


jsquire (approved):

sdk/search/Azure.Search.Documents/api/Azure.Search.Documents.netstandard2.0.cs
```diff
@@ -2954,7 +2954,7 @@ public static partial class SearchModelFactory
         public static Azure.Search.Documents.Indexes.Models.CognitiveServicesAccount CognitiveServicesAccount(string oDataType, string description) { throw null; }
         public static Azure.Search.Documents.Indexes.Models.DataChangeDetectionPolicy DataChangeDetectionPolicy(string oDataType) { throw null; }
         public static Azure.Search.Documents.Indexes.Models.DataDeletionDetectionPolicy DataDeletionDetectionPolicy(string oDataType) { throw null; }
-        public static Azure.Search.Documents.Models.FacetResult FacetResult(long? count, System.Collections.Generic.IReadOnlyDictionary<string, object> additionalProperties) { throw null; }
+        public static Azure.Search.Documents.Models.FacetResult FacetResult(long? count = default(long?), System.Collections.Generic.IReadOnlyDictionary<string, object> additionalProperties = null) { throw null; }
```
pallavit, line 2957: Is this not considered a breaking change?

ShivangiReja: Since model factory is used only for mocking to enable non-live testing of service clients by customers, I'm not sure, if this is a breaking change.

ShivangiReja: I don't think it's a breaking change, we are just providing default values for the all the parameters.

ShivangiReja: I don't think it will break customer's existing code, since we are just providing default values for all the parameters.

Also, we are not changing public client. We are just updating model factory which is used for mocking to enable non-live testing of service clients by customers.

ShivangiReja: @tg-msft could you please confirm if this is considered as a breaking change or not?

jsquire: I couldn't come up with a case where existing code would fail to resolve the parameters or bind to a different overload.   @pallavit - was there a scenario that you had in mind?

tg-msft: The compiler only includes the default value at the call site, so

```C#
Test();
void Test(int x = 42) { }
```

compiles into 

```diff
.method private hidebysig static void '<Main>$' (string[] args) cil managed 
{
    .maxstack 8
    .entrypoint
    ldc.i4.s 42
    call void '<Program>$'::'<<Main>$>g__Test|0_0'(int32)
    ret
}

.method assembly hidebysig static void '<<Main>$>g__Test|0_0' (int32 x) cil managed 
{
    .maxstack 8
    ret
}
```

That luckily means we don't have to worry about binary breaking changes!  It is possible for tweaking default values to introduce source breaking changes though.  For example, doing something like adding an overload to switch to an options bag:

```diff
- void Test(int x = 42, string y = "hello")
+ void Test(int x, string y)
+ void Test(TestOptions options = default)
```

would cause a compile error like `error CS1503: Argument 1: cannot convert from 'int' to 'TestOptions'` for anyone using `Test(21)` today.

We try really hard to avoid those, but sometimes let them through.  Default values in C# unfortunately version worse than most other language features.  We've had a lot of discussion with the .NET team about this and their view is we're using overloads incorrectly/too aggressively.  :smile:

All that said, this particular example isn't going to break anyone.

pallavit: Thank you everyone! Yes my conclusion based on all the breaking change conversations about optional arguments was that everything regarding optional arguments is breaking :P. 

Apologies for replying on a merged thread but I was curious what we do in these scenarios :)

PR #29914, opened 2022-07-14
[FormRecognizer] Update LROs to use new pattern

kinelski:
Based on architect feedback, we're updating LROs in Form Recognizer to use a new pattern with the `WaitUntil` enum. Ideally, new SDKs will start following the new pattern from now on.

LRO usage with the old pattern:
```cs
var operation = client.StartAnalyzeDocument(<parameters>);
await operation.WaitForCompletionAsync();

var value = operation.Value;
```

LRO usage with the new pattern:
```cs
// "Start" prefix removed from the method name.
// WaitUntil parameter added (supported values: Completed or Started).
var operation = client.AnalyzeDocument(WaitUntil.Completed, <parameters>);

var value = operation.Value;
```


sdk/formrecognizer/Azure.AI.FormRecognizer/tests/DocumentAnalysisClient/DocumentAnalysisClientLiveTests.cs
```diff
@@ -925,6 +904,7 @@ public async Task StartAnalyzeDocumentPopulatesExtractedInvoiceJpg(bool useStrea
         [RecordedTest]
         [TestCase(true)]
         [TestCase(false)]
+        [Ignore("https://github.com/Azure/azure-sdk-for-net/issues/29966")]
```
kinelski, line 907: Copying the description of https://github.com/Azure/azure-sdk-for-net/issues/29966:
> Our Test Framework has some kind of logic to speed up the polling process of LROs in playback mode. For some reason, this logic does not seem to work with the new LRO pattern in Form Recognizer. This is causing the `StartBuildModelWithNeuralBuildMode` and `StartAnalyzeDocumentCanParseMultipageInvoice` tests to time out, so we disabled them for the time being.

sdk/formrecognizer/Azure.AI.FormRecognizer/README.md
```diff
@@ -159,7 +159,7 @@ Please note that models can also be built using a graphical user interface such
...
 
-For long running operations in the Azure SDK, the client exposes a `Start<operation-name>` method that returns an `Operation<T>`.  You can use the extension method `WaitForCompletionAsync()` to wait for the operation to complete and obtain its result.  A sample code snippet is provided to illustrate using long-running operations [below](#extract-layout).
+For long running operations in the Azure SDK, the client exposes a method that returns an `Operation<T>` object. You can set its parameter `waitUntil` to `WaitUntil.Completed` to wait for the operation to complete and obtain its result; or set it to `WaitUntil.Started` if you just want to start the operation and consume the result later. A sample code snippet is provided to illustrate using long-running operations [below](#extract-layout).
```
pallavit, line 162: Is this a new pattern we have created for LRO? If so, how is the pattern in FA different from other LRO patterns? Is this common across cognitive space so say TA would benefit from this too? 

Also, do we need to publicize this pattern in our design guidelines?

sdk/formrecognizer/Azure.AI.FormRecognizer/README.md
```diff
@@ -287,10 +284,7 @@ Analyze text, selection marks, table structures, styles, paragraphs, and key-val
 ```C# Snippet:FormRecognizerAnalyzePrebuiltDocumentFromUriAsync
 Uri fileUri = new Uri("<fileUri>");
 
-AnalyzeDocumentOperation operation = await client.StartAnalyzeDocumentFromUriAsync("prebuilt-document", fileUri);
-
-await operation.WaitForCompletionAsync();
-
+AnalyzeDocumentOperation operation = await client.AnalyzeDocumentFromUriAsync(WaitUntil.Completed, "prebuilt-document", fileUri);
```
pallavit, line 287: >AnalyzeDocumentFromUriAsync [](http://example.com/codeflow?start=50&length=27)

All our code snippets use the WaitUntil.Complete pattern. Should we add code snippets or perhaps a sample for the other option here?

sdk/formrecognizer/Azure.AI.FormRecognizer/README.md
pallavit: Looks like this is an existing pattern. WaitUntil is a type that already exists :)

pallavit (approved):

sdk/formrecognizer/Azure.AI.FormRecognizer/README.md
kinelski: That's right. It's a new pattern architects want libraries to follow from now on. From my understanding, libraries that already implement the old pattern should keep doing so for consistency.

The pattern is not described in our guidelines yet. Krzysztof said we should remind him to update it.

PR #29915, opened 2022-07-14
Purview Share SDK

yamanwahsheh:


doc/ApiDocGeneration/Generate-Api-Docs.ps1
```diff
@@ -182,4 +182,13 @@ $mutatedContent = $mutatedContent -replace $hrefRegex, '"./$1"'
 Set-Content -Path $baseUrl -Value $mutatedContent -NoNewline
 
 Write-Verbose "Compress and copy HTML into the staging Area"
+Write-Verbose "Path to Archive"
```
pallavit, line 185: @benbp , @chidozieononiwu , @weshaggard Could you please help review the change to this file. It LGTM but want to make sure you get a chance to review as you likely have more context here.

benbp: Changes look fine to me as well. @sima-zhu do you have any thoughts?

sima-zhu: The reason why we failed this is because in our `Build and Package` step, we create the folder over there. 
image

There is no source code for `Azure.Analytics.Purview.Share`, so the `Generate-API-Docs.ps1` failed to upload docs. 
For artifacts, we usually contain both source package and docs package instead of having docs only. That's why we do not create folder if not exist. Could you understand why the share package does not show on build step?

Expected to see (something like storage):
image

Actually seen:
image

yamanwahsheh: Thank you @sima-zhu , as discussed over teams, the name was incorrect, it was Azure.Analytics.Share which was supposed to be Azure.Analytics.Purview.Share, that was fixed, the powershell scripts is undone now. The pipeline passed as the correct folder name exists.

sdk/purview/Azure.Analytics.Purview.Share/README.md
```diff
@@ -0,0 +1,51 @@
+# Azure Share client library for .NET
```
anniruddh, line 1: Please fill up the ReadMe.

yamanwahsheh: Readme file updated

anniruddh (approved):

pallavit (approved):
Approving the check-in to allow the partner team to make progress on their SDK.

PR #29916, opened 2022-07-15
Add storage test pipeline with static config for virtualized resource

### LINKED ISSUE #3580: Support multiple ARM template selection in test pipelines
We need a way to support selecting different ARM templates within a single test pipeline. Currently the storage team needs this as a way of testing one-off resource types. A couple ways I can think of to do this:

- Support custom arm template paths via matrix variables (e.g. similar to how we support the special field `armTemplateParameters`) or via the `MatrixConfigs` yaml configuration.
- Give guidance on using bicep modules with a top level bicep file that uses if conditionals to select a module, based on `armTemplateParameters` set in the matrix config. It may be complicated to figure out hwo to standardize deployment outputs and skip/run test-resources-post scripts.

CC @seanmcc-msft

benbp:
Resolves https://github.com/Azure/azure-sdk-tools/issues/3580

This PR adds support for storage live tests to download a static test configuration file instead of constructing it dynamically via `test-resources-post.ps1` from ARM template output variables. Specifically this will light up a new virtualized test pipeline with non-ARM deployable resources.

CC @seanmcc-msft @kasobol-msft

### DIFF
--- a/eng/pipelines/templates/jobs/live.tests.yml
+++ b/eng/pipelines/templates/jobs/live.tests.yml
@@ -32,6 +32,9 @@ parameters:
 - name: TestSetupSteps
   type: stepList
   default: []
+- name: DeployTestResources
+  type: boolean
+  default: true
 - name: CloudConfig
   type: object
   default: {}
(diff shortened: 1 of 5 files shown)


weshaggard (approved):
This looks reasonable to me.

seanmcc-msft (approved):

PR #29925, opened 2022-07-15
Update ACL .NET client

darracott:
This PR updates Azure Confidential Ledger's .NET Client code and documentation. Currently running the code snippets in the .NET "Create a Client" section [here](https://docs.microsoft.com/en-gb/dotnet/api/overview/azure/security.confidentialledger-readme-pre?WT.mc_id=Portal-fx#create-a-client) fails. This PR fixes that failure and streamlines the tutorial docs. 

- ~~Tutorial code currently creates a Ledger Client with custom `CertValidationCheck` which happens to be the same as the default it is overwriting. The issue is that the Ledger Client constructor instantiates an Identity Service Client and passes it the same custom `CertValidationCheck`, but the two clients have different needs. In the tutorial's case we end up overwriting the Identity Service's default Cert Validation with the Ledger Client's `CertValidationCheck`. As the two clients have different cert validation requirements this causes the IdentityService CLient to fail to connect to the Identity Service, blocking the creation of the Ledger Client. (_The Identity Service server returns a certificate which is signed by a standard CA, not by our Ledger's Root Cert. The `CertValidationCheck` intended for the Ledger Client rejects server certs that are not signed by the Ledger's Root Cert._)~~

~~This makes it awkward for anyone wanting to implement a custom `CertValidationCheck` for their Ledger Client, as currently this check will be picked up by both the Ledger Client and the Identity Service Client. This PR prevents the Identity Client from picking up options intended for the Ledger Client. This leaves the Identity client always using its defaults. It would be good to discuss this change.~~ After discussion we decided to live with this as it is unlikely anyone will ever edit the `CertValidationCheck` and if they do they can add complexity to deal with the initial Identity Client connections.

- Additionally, this PR streamlines the "Create a Client" documentation as it is confusing for the tutorial to reimplement the `CertValidationCheck` which is already built into the Ledger Client by its constructor by default. 
- This PR also tweaks the default Ledger Client cert validation code. The `CertValidationCheck` callback for `ConfidentialLedgerClient` instances now checks the final certificate in the server's certificate chain matches the trusted TLS certificate. Previously this callback checked if the thumbprint of the trusted TLS certificate was present anywhere in the server's certificate chain.

### DIFF
--- a/sdk/confidentialledger/Azure.Security.ConfidentialLedger/CHANGELOG.md
+++ b/sdk/confidentialledger/Azure.Security.ConfidentialLedger/CHANGELOG.md
@@ -8,8 +8,6 @@
 
 ### Bugs Fixed
 
-- Custom client `options` used when instantiating a `ConfidentialLedgerClient` are no longer also passed to the `ConfidentialLedgerIdentityServiceClient` instance used during construction of the `ConfidentialLedgerClient`, as these two clients have different requirements, in particular on certificate validation. 
-
 ### Other Changes
 
 - The `CertValidationCheck` callback for `ConfidentialLedgerClient` instances now checks the final certificate in the server's certificate chain matches the trusted TLS certificate. Previously this callback checked if the thumbprint of the trusted TLS certificate was present anywhere in the server's certificate chain. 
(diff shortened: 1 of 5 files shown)


christothes:
Thanks for this! I had another PR in progress so just incorporated the relevant changes here: https://github.com/Azure/azure-sdk-for-net/pull/29929

PR #29946, opened 2022-07-18
[ServiceBus] No longer create a copy of the array in the message converter for the default and stream case

danielmarbach:
This gets rid of the extra allocations and CPU overhead when calling `ToArray`. For the default and memory stream case that is quite a bit of allocations and memory copying operations that are no longer necessary.

- [x] Update Change Log
- [x] Settle discussions about disposing and `TryGetBuffer` vs `GetBuffer`

### DIFF
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
@@ -159,14 +159,22 @@ namespace Azure.Messaging.ServiceBus.Amqp
                 {
                     using var memStreamCopy = new MemoryStream((int)(memStreamSource.Length - memStreamSource.Position));
                     memStreamSource.CopyTo(memStreamCopy, StreamBufferSizeInBytes);
-                    return new ArraySegment<byte>(memStreamCopy.ToArray());
+                    if (!memStreamCopy.TryGetBuffer(out ArraySegment<byte> segment))
+                    {
+                        segment = new ArraySegment<byte>(memStreamCopy.ToArray());
+                    }
+                    return segment;
                 }
 
                 default:
                 {
-                    using var memStream = new MemoryStream(StreamBufferSizeInBytes);
-                    stream.CopyTo(memStream, StreamBufferSizeInBytes);
-                    return new ArraySegment<byte>(memStream.ToArray());
+                    using var memStreamCopy = new MemoryStream(StreamBufferSizeInBytes);
+                    stream.CopyTo(memStreamCopy, StreamBufferSizeInBytes);
+                    if (!memStreamCopy.TryGetBuffer(out ArraySegment<byte> segment))
+                    {
+                        segment = new ArraySegment<byte>(memStreamCopy.ToArray());
+                    }
+                    return segment;
                 }
             }
         }
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
@@ -850,14 +850,6 @@ namespace Azure.Messaging.ServiceBus.Amqp
             return buffer;
         }
 
-        private static Data ToData(AmqpMessage message)
-        {
-            ArraySegment<byte>[] payload = message.GetPayload();
-            var buffer = new BufferListStream(payload);
-            ArraySegment<byte> value = buffer.ReadBytes((int)buffer.Length);
-            return new Data { Value = value };
-        }
-
         internal static AmqpMap GetSqlRuleFilterMap(SqlRuleFilter sqlRuleFilter)
         {
             var amqpFilterMap = new AmqpMap
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
@@ -145,13 +145,11 @@ namespace Azure.Messaging.ServiceBus.Amqp
         ///
         private static ArraySegment<byte> ReadStreamToArraySegment(Stream stream)
         {
-            if (stream == null)
-            {
-                return new ArraySegment<byte>();
-            }
-
             switch (stream)
             {
+                case { Length: < 1 }:
+                    return default;
+
                 case BufferListStream bufferListStream:
                     return bufferListStream.ReadBytes((int)stream.Length);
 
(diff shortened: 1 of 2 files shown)


sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
```diff
@@ -159,14 +159,22 @@ private static ArraySegment<byte> ReadStreamToArraySegment(Stream stream)
                 {
                     using var memStreamCopy = new MemoryStream((int)(memStreamSource.Length - memStreamSource.Position));
```
danielmarbach, line 160: It is actually not necessary to dispose `MemoryStreams` see https://docs.microsoft.com/en-us/dotnet/api/system.io.memorystream?view=netstandard-2.0#remarks but I figured I leave it as is currently since there is no negative impact on the code when we still do it. 

One could argue though that it could be "confusing" to the reader of the code to see that we pass the buffer beyond the scope of the lifetime of the memory stream. Yet not disposing the stream could trigger some kind of analysis warning.

Raising it here as a discussion point to see what you prefer.

danielmarbach:
Will apply the same changes in the EventHubs client once the discussion has settled here

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
jsquire: Personally, I'd keep the `using` to appease the analyzers and keep with the "dispose it if its disposable" heuristic.   Maybe drop a comment that the buffer is safe to outlive the dispose call.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
```diff
@@ -159,14 +159,22 @@ private static ArraySegment<byte> ReadStreamToArraySegment(Stream stream)
...
                     memStreamSource.CopyTo(memStreamCopy, StreamBufferSizeInBytes);
-                    return new ArraySegment<byte>(memStreamCopy.ToArray());
+                    if (!memStreamCopy.TryGetBuffer(out ArraySegment<byte> segment))
```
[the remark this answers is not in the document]
jsquire: I like the defensive approach, personally.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
```diff
@@ -159,14 +159,22 @@ private static ArraySegment<byte> ReadStreamToArraySegment(Stream stream)
                 {
                     using var memStreamCopy = new MemoryStream((int)(memStreamSource.Length - memStreamSource.Position));
                     memStreamSource.CopyTo(memStreamCopy, StreamBufferSizeInBytes);
```
jsquire, line 161: Since the buffer can outlive the stream, do we need to create the copy before reading it, since this case is known to be a memory stream?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
```diff
@@ -831,24 +839,17 @@ private static ArraySegment<byte> StreamToBytes(Stream stream)
             }
             else
             {
-                using (var memoryStream = new MemoryStream(512))
+                using var memoryStream = new MemoryStream(StreamBufferSizeInBytes);
```
jsquire, line 842: If we can get away with extracting the buffer without a copy for a `MemoryStream`, should we consider a switch here and only allocate a new one if it's not already a memory stream?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
jsquire: Fair enough...  nor am I.   I'll ask around internally, but in the absence of information - the safer approach is definitely better.  😄

JoshLove-msft:
> @JoshLove-msft any thoughts on this one?

I agree with @jsquire's suggestions, if you can consider implementing them.

danielmarbach:
@JoshLove-msft you lost me there ;) @jsquire upvoted my current approach in this PR so unless I missed something this should be good to go after adding something to the changelog.

Or are you referring to

> Since the buffer can outlive the stream, do we need to create the copy before reading it, since this case is known to be a memory stream?

which I think I mentioned

> I'm not entirely aware of all the edge cases here, so I went with the safest approach for now. But in case you and Josh have some more insights I'd be happy to change it.

So this PR here seems to be already a good improvement and I'm happy to think about that additional potentially "more dangerous" optimization in a dedicated PR once I find some more time

JoshLove-msft:
> @JoshLove-msft you lost me there ;) @jsquire upvoted my current approach in this PR so unless I missed something this should be good to go after adding something to the changelog.
> 
> Or are you referring to
> 
> > Since the buffer can outlive the stream, do we need to create the copy before reading it, since this case is known to be a memory stream?
> 
> which I think I mentioned
> 
> > I'm not entirely aware of all the edge cases here, so I went with the safest approach for now. But in case you and Josh have some more insights I'd be happy to change it.
> 
> So this PR here seems to be already a good improvement and I'm happy to think about that additional potentially "more dangerous" optimization in a dedicated PR once I find some more time

I was referring to this comment: https://github.com/Azure/azure-sdk-for-net/pull/29946#discussion_r923430929

danielmarbach:
@JoshLove-msft your wish is my command. Let's be brave then and go one step further.

danielmarbach:
I think we can revert the last commit. Since Amqp Message uses the following Memory Stream ctor `MemoryStream(Byte[], Int32, Int32)` `TryGetBuffer` will return `false`.

See 

> This constructor does not expose the underlying stream. GetBuffer throws UnauthorizedAccessException. However, you can write to the stream.

https://docs.microsoft.com/en-us/dotnet/api/system.io.memorystream.-ctor?view=net-6.0#system-io-memorystream-ctor(system-byte()-system-int32-system-int32)

JoshLove-msft:
> I think we can revert the last commit. Since Amqp Message uses the following Memory Stream ctor `MemoryStream(Byte[], Int32, Int32)` `TryGetBuffer` will return `false`.
> 
> See
> 
> > This constructor does not expose the underlying stream. GetBuffer throws UnauthorizedAccessException. However, you can write to the stream.
> 
> https://docs.microsoft.com/en-us/dotnet/api/system.io.memorystream.-ctor?view=net-6.0#system-io-memorystream-ctor(system-byte()-system-int32-system-int32)

Sounds good - sorry for the run around!

danielmarbach:
It was a good turnaround after all. I think I managed to collapse the two code paths nicely, which also cleans up the code quite a bit

danielmarbach:
I think this is good to go now.

danielmarbach:
Greeeeeen 💪👍😎

JoshLove-msft (approved):

danielmarbach:
Will try to find time to do the same for EventHubs

danielmarbach:
Here we go https://github.com/Azure/azure-sdk-for-net/pull/30073

PR #29952, opened 2022-07-18
[Event Hub] Small improvements & refactorings

danielmarbach:

### DIFF
--- a/sdk/eventhub/Azure.Messaging.EventHubs/src/Core/TransportProducerPool.cs
+++ b/sdk/eventhub/Azure.Messaging.EventHubs/src/Core/TransportProducerPool.cs
@@ -157,7 +157,7 @@ namespace Azure.Messaging.EventHubs.Core
         {
            if (Pool.TryRemove(partitionId, out var poolItem))
            {
-               if ((poolItem.ActiveInstances.IsEmpty) || (forceClose))
+               if (poolItem.ActiveInstances.IsEmpty || forceClose)
                {
                    await poolItem.PartitionProducer.CloseAsync(CancellationToken.None).ConfigureAwait(false);
                }
--- a/sdk/eventhub/Azure.Messaging.EventHubs/src/Core/TransportProducerPool.cs
+++ b/sdk/eventhub/Azure.Messaging.EventHubs/src/Core/TransportProducerPool.cs
@@ -348,15 +348,8 @@ namespace Azure.Messaging.EventHubs.Core
             ///
             /// <returns>A task to be resolved on when the operation has completed.</returns>
             ///
-            public virtual ValueTask DisposeAsync()
-            {
-                if (CleanUp != default)
-                {
-                    return new ValueTask(CleanUp(TransportProducer));
-                }
-
-                return new ValueTask(Task.CompletedTask);
-            }
+            public virtual ValueTask DisposeAsync() =>
+                CleanUp != default ? new ValueTask(CleanUp(TransportProducer)) : new ValueTask();
         }
     }
 }
--- a/sdk/eventhub/Azure.Messaging.EventHubs/src/Core/TransportProducerPool.cs
+++ b/sdk/eventhub/Azure.Messaging.EventHubs/src/Core/TransportProducerPool.cs
@@ -101,8 +101,8 @@ namespace Azure.Messaging.EventHubs.Core
                 return new PooledProducer(EventHubProducer);
             }
 
-            var identifier = Guid.NewGuid().ToString();
-            var item = Pool.GetOrAdd(partitionId, id => new PoolItem(partitionId, TransportProducerFactory(id), removeAfterDuration));
+            var identifier = Guid.NewGuid();
+            var item = Pool.GetOrAdd(partitionId, id => new PoolItem(id, TransportProducerFactory(id), removeAfterDuration));
 
             // A race condition at this point may end with CloseAsync called on
             // the returned PoolItem if it had expired. The probability is very low and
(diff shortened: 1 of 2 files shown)


danielmarbach:
@jsquire I noticed `GetPooledProducer` might call `TransportProducerFactory` multiple times because `GetOrAdd` might call the value factory multiple times. Maybe it is not a problem I just made me suspicious because the transport producer is a resource that needs to be closed if I understand the code correctly.

sdk/eventhub/Azure.Messaging.EventHubs/src/Core/TransportProducerPool.cs
```diff
@@ -101,17 +101,17 @@ internal TransportProducerPool()
                 return new PooledProducer(EventHubProducer);
             }
 
-            var identifier = Guid.NewGuid().ToString();
-            var item = Pool.GetOrAdd(partitionId, id => new PoolItem(partitionId, TransportProducerFactory(id), removeAfterDuration));
+            var identifier = Guid.NewGuid();
+            var item = Pool.GetOrAdd(partitionId, id => new PoolItem(id, TransportProducerFactory(id), removeAfterDuration));
```
danielmarbach, line 105: It would be nice to be able to properly #ifdef here to that the `GetOrAdd` method could be used that allows to pass the argument. Then the closure allocation would be gone because the pool could pass itself as an argument to the value factory. It would require to slightly change the remove after duration value propagation but since the code currently always calls ExtendRemoveAfter it would be easy to rename that method and bring in the constructor logic that checks for the default value.

jsquire: not sure that I'm following this one...

sdk/eventhub/Azure.Messaging.EventHubs/src/Core/TransportProducerPool.cs
```diff
@@ -348,15 +348,8 @@ internal class PooledProducer : IAsyncDisposable
...
-            {
-                if (CleanUp != default)
-                {
-                    return new ValueTask(CleanUp(TransportProducer));
-                }
-
-                return new ValueTask(Task.CompletedTask);
-            }
+            public virtual ValueTask DisposeAsync() =>
+                CleanUp != default ? new ValueTask(CleanUp(TransportProducer)) : new ValueTask();
```
jsquire, line 352: I didn't think there was a value task overload without args...

sdk/eventhub/Azure.Messaging.EventHubs/src/Core/TransportProducerPool.cs
jsquire, line 105: Doesn't changing the first arg from `partitionId` to `id` here also change the semantics?   That makes me a bit nervous.   Then again, looking at the constructor for `PoolItem` it would seem they need to be the same value so that the factory and item agree on what partition they're for...  why in the world do we have two of them?

danielmarbach: I'll try to type some up pseudo code later.

danielmarbach: The partition key is used as the key to the dictionary which is also what is being passed as an argument into the value factory method. In an ideal world the value factory method would not need to reference anything outside the closure so I did a first step to align things here since partionKey and id are the same

danielmarbach:
Please also see https://github.com/Azure/azure-sdk-for-net/pull/29952#issuecomment-1187588797

danielmarbach:
Ah I think these comments



```
// A race condition at this point may end with CloseAsync called on
            // the returned PoolItem if it had expired. The probability is very low and
            // possible exceptions should be handled by the invoking methods.

// If TryRemove returned false the PoolItem would not be closed deterministically
                // and the ExpirationTimer callback would eventually remove it from the
                // Pool leaving to the Garbage Collector the responsibility of closing
                // the TransportProducer and the AMQP link.
```

explains things

danielmarbach:
I have another design related question. I wonder if it would make sense to add the necessary state and references to `PooledProducer` and then make the `DisposeAsync` method do the cleaning. With that the pooled producer could have a reference to the PooledItem and piggy back the identifier and what is needed for the cleanup. While this class is a bit sneaky how it reaches into the parent state it would get rid of the unnecessary closure.

sdk/eventhub/Azure.Messaging.EventHubs/src/Core/TransportProducerPool.cs
jsquire: ahh...  yeah.   Reading is hard.

jsquire: hrmm...   this was a contribution, so I don't know all the ins-and-outs as well as I'd like, but I don't see a reason why we couldn't remove the duration for the initial construction.    We'd end up with a default value (10 minutes) and, to your point, we do explicitly extend it again once we get an instance.  

We'd probably want to tweak the default duration to 5 minutes to match what the producer uses.   Not sure if that throws off the tests at all, but I'm hoping not.

danielmarbach: These changes would be good if we manage to bring in the GetOrAdd overload for TFMs that support it. Are other parts of the SDK doing more TFM specific optimizations? Is that something OK doing?

jsquire:
> I have another design related question. I wonder if it would make sense to add the necessary state and references to `PooledProducer` and then make the `DisposeAsync` method do the cleaning. With that the pooled producer could have a reference to the PooledItem and piggy back the identifier and what is needed for the cleanup. While this class is a bit sneaky how it reaches into the parent state it would get rid of the unnecessary closure.

I generally have a strong preference to keep object graphs flowing in a single direction, but this is an internal construct of very tightly coupled types, so I'm not totally opposed in this specific instance.  I do know that we originally ran into a lot of issues with concurrent use and non-deterministic clean-up that we had to guard against.   _(the two comments that you highlighted call some out)_

Do you think we'd be opening up additional corner cases to guard against?

sdk/eventhub/Azure.Messaging.EventHubs/src/Core/TransportProducerPool.cs
danielmarbach: No you are absolutely right and I agree at this stage it is probably not worth the hazzle.

danielmarbach:
I guess this is good to go then

jsquire (approved):

jsquire:
> I guess this is good to go then

Much appreciated!

PR #29953, opened 2022-07-18
Updates suggested from Azure Board SDK review

minwoolee-msft:
- CallConnectionClient and CallRecordingClient suffix 'Client' removed
- Azure.Communication.CallingServer.Models namespace removed
- CallConnectionClient and CallContentClient combined
- Hangup and Terminate combined. Hangup now expects boolean to either hangup, or terminate the call
- GetCallConnectionProperties now under CallConnectionClient
- GetCallConnectionProperties renamed to GetProperties
- ...Response suffix updated to be ...Result suffix (i.e. TransferCallToParticipantResponse -> TransferCallToParticipantResult)
- AlternateCallerId renamed
- All Options constructor doesn't take optional value as input
- GetParticipants now accepts participantMri string
- OperationResponse removed
- PlayMedia updated to accept PlaySource's derived class (FileSource)
- Added many Internal models to hide generated models that contoso do not need
- AnswerCall and CreateCall now returns Result. This object now contains both CallProperties & CallConnectionClient
- CreateCall now accepts subject instead of option that contains only 1 value
- ...Type suffix now dropped. (i.e. RecordingContentType -> RecordingContent)
- ServerCallingModelFactory to include all read-only models
- Autorest regenerated with latest swagger
- Tests updated to accomodate latest changes
- Summaries of each methods updated to accomodate latest changes
- README.md updated to accomodate latest changes


cochi2 (approved):

jasonshave (approved):

sdk/communication/Azure.Communication.CallingServer/src/Models/PlaySource.cs
```diff
@@ -5,22 +5,11 @@
 using System.Collections.Generic;
 using System.Text;
 
-namespace Azure.Communication.CallingServer.Models
+namespace Azure.Communication.CallingServer
 {
     /// <summary> The PlaySource. </summary>
-    public class PlaySource
+    public abstract class PlaySource
```
williamzhao87, line 11: This code seems to have already been updated in master branch, such that it's not an abstract class anymore @arupdutta-msft

sdk/communication/Azure.Communication.CallingServer/src/CallConnection.cs
```diff
@@ -82,19 +84,30 @@ public virtual Response Hangup(CancellationToken cancellationToken = default)
             }
         }
 
-        /// <summary> Terminates the conversation for all participants in the call. </summary>
+        /// <summary> Disconnect the current caller in a group-call or end a p2p-call.</summary>
+        /// <param name="forEveryone"> If true, this will terminate the call and hang up on all participants in this call. </param>
         /// <param name="cancellationToken"> The cancellation token. </param>
         /// <exception cref="RequestFailedException">The server returned an error. See <see cref="Exception.Message"/> for details returned from the server.</exception>
-        public virtual async Task<Response> TerminateCallAsync(CancellationToken cancellationToken = default)
+        public virtual async Task<Response> HangupAsync(bool forEveryone, CancellationToken cancellationToken = default)
```
williamzhao87, line 91: Should forEveryone have a default of false?

williamzhao87: I guess it depends on how frequently we expect these methods to be used as well.  If majority of time the user would just hangup their own callConnection, I feel having default value would be useful so we don't always have to repeatedly pass in the same value, except in exceptional case where you want to actually terminate a call.  If we expect fairly equal usage for both hangup/terminate, then mandatory property makes most sense to me.

sdk/communication/Azure.Communication.CallingServer/src/Models/AnswerCallResult.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Azure.Communication.CallingServer
+{
+    /// <summary> The result from creating a call. </summary>
+    public class AnswerCallResult
+    {
+        internal AnswerCallResult(CallConnection callConnection, CallConnectionProperties callProperties)
```
williamzhao87, line 9: Why is the type called CallConnectionProperties, but the variable name callProperties?

sdk/communication/Azure.Communication.CallingServer/src/CallConnection.cs
minwoolee-msft: I still kind of like keeping this mandatory, but I can bring this topic up during the meeting as one of the discussion topics

sdk/communication/Azure.Communication.CallingServer/src/Models/AnswerCallResult.cs
jasonshave: I would tend to agree here but the type has a "long name" and the rule of thumb is that input parameter names match their type names, aside from primitive types of course. So maybe consider changing `CallConnectionProperties` to `CallProperties` with a matching `callProperties` parameter name or make it match the other way.

Issue #30007, opened 2022-07-19
Record vs Playback Race Condition - Storage Client-side Encryption Tests


jaschrep-msft:
Some blob client-side encryption tests for partitioned transfer are encountering a race condition, preventing them from being converted from LiveOnly to recordable. Helps address #26037

Issue #26037, opened 2021-12-22
Investigate better maintenance of client-side encryption tests


jaschrep-msft:
We're currently forced to have live-only tests for client-side encryption requests that hit the service. This maintenance issue is growing over time and requiring many very simple tests to be live-only, creating lengthy PR processes.

PR #30010, opened 2022-07-19
Record encryption tests

jaschrep-msft:
Eliminates LiveOnly dependency of client-side cryptography tests (mostly). Addresses #26037

- Makes encryption region data size configurable internally for smaller recording size
- Tests mess with function pointers to replace cryptographically secure RNG with recordable RNG, without leaving any holes in published package for a PR to accidentally undermine security
- Moves client-side encryption tests to new test package with unsafe code allowed (to accommodate above point)
- Some tests have a race condition currently preventing them from being played back appropriately. They have been left LiveOnly. #30007 to track


jaschrep-msft:
Confirmed credscan was a false positive. Recordings had 64 byte reads, whose base64 values in the recording file were interpreted as 512-bit keys stored in code/config. Test definitions altered to avoid this exact read size.

jaschrep-msft:
Blocked on #30065. Necessary test configurations that were possible in live testing are not possible in CI at the moment.

jaschrep-msft:
work blocked

jaschrep-msft:
Still blocked on mentioned feature request.

PR #30038, opened 2022-07-20
[EventHubs] ClientLibraryInformation serializes the current information once and then no longer uses reflection

danielmarbach:
Same as https://github.com/Azure/azure-sdk-for-net/pull/29949 but for EventHubs as promised


jsquire (approved):

danielmarbach:
Anything required for the change log? I was a bit hesitant because it is an internal only change. I haven't measured it but since we did the reflection and materialization for every connection scope this should give some noticeable boost.

jsquire:
> Anything required for the change log? I was a bit hesitant because it is an internal only change. I haven't measured it but since we did the reflection and materialization for every connection scope this should give some noticeable boost.

I don't think anything specific is needed;  I was planning on lumping these wonderful things under a general item for "performance enhancements and efficiency improvements" with attribution.

PR #30039, opened 2022-07-20
[Media Composition] Break down Update into more granular functions

minnieliu:


sdk/communication/Azure.Communication.MediaComposition/api/Azure.Communication.MediaComposition.netstandard2.0.cs
```diff
@@ -145,18 +145,26 @@ public partial class MediaCompositionClient
...
         public virtual System.Threading.Tasks.Task<Azure.Response> DeleteAsync(string mediaCompositionId, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
         public virtual Azure.Response<Azure.Communication.MediaComposition.MediaCompositionBody> Get(string mediaCompositionId, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
         public virtual System.Threading.Tasks.Task<Azure.Response<Azure.Communication.MediaComposition.MediaCompositionBody>> GetAsync(string mediaCompositionId, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
+        public virtual Azure.Response<Azure.Communication.MediaComposition.MediaCompositionBody> RemoveInputs(string mediaCompositionId, System.Collections.Generic.IEnumerable<string> inputIds, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
```
DominikMe, line 158: It would be nicer to rename `MediaCompositionBody` to `MediaComposition` in the swagger.

sdk/communication/Azure.Communication.MediaComposition/api/Azure.Communication.MediaComposition.netstandard2.0.cs
```diff
@@ -145,18 +145,26 @@ public partial class MediaCompositionClient
         public MediaCompositionClient(string connectionString, Azure.Communication.MediaComposition.MediaCompositionClientOptions options) { }
         public MediaCompositionClient(System.Uri endpoint, Azure.AzureKeyCredential keyCredential, Azure.Communication.MediaComposition.MediaCompositionClientOptions options = null) { }
         public MediaCompositionClient(System.Uri endpoint, Azure.Core.TokenCredential tokenCredential, Azure.Communication.MediaComposition.MediaCompositionClientOptions options = null) { }
+        public virtual Azure.Response<Azure.Communication.MediaComposition.MediaCompositionBody> AddInputs(string mediaCompositionId, System.Collections.Generic.IDictionary<string, Azure.Communication.MediaComposition.MediaInput> inputs, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
```
DominikMe, line 148: How do we think about Updates?
We could add a new `UpdateInputs` method, or we change `AddInputs` to `UpsertInputs`, or we could replace Add/Remove with a single `UpdateInputs(IEnumerable<MediaInput> upsert, IEnumerable<string> remove)`.

sdk/communication/Azure.Communication.MediaComposition/api/Azure.Communication.MediaComposition.netstandard2.0.cs
```diff
@@ -145,18 +145,26 @@ public partial class MediaCompositionClient
...
+        public virtual System.Threading.Tasks.Task<Azure.Response<Azure.Communication.MediaComposition.MediaCompositionBody>> AddInputsAsync(string mediaCompositionId, System.Collections.Generic.IDictionary<string, Azure.Communication.MediaComposition.MediaInput> inputs, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
```
DominikMe, line 149: In [Chat](https://github.com/Azure/azure-sdk-for-net/blob/cbf8799dfbb1222be660fa904a8f31e165ea97dc/sdk/communication/Azure.Communication.Chat/api/Azure.Communication.Chat.netstandard2.0.cs#L114) adding a participant just returns an empty response. Our service works differently and returns the full resource. The [SDK guidelines](https://azure.github.io/azure-sdk/general_design.html#client-interface) suggest to return the added item when a single item gets added, but I don't know if that rule can be easily applied here.

Pro of returning the full composition is that the developer potentially saves another GET request. Con is that an empty response might be more idiomatic (at least for in-memory types like a dictionary, but here we have a REST resource where the manipulation happens on the server and not locally). We can ask the Azure board.

sdk/communication/Azure.Communication.MediaComposition/api/Azure.Communication.MediaComposition.netstandard2.0.cs
minnieliu: That's a good point, Dom. I like the option of `UpsertInputs`, we still have the granularity but Update + Add fits together well and reflects merge+patch. What are your preferences?

sdk/communication/Azure.Communication.MediaComposition/api/Azure.Communication.MediaComposition.netstandard2.0.cs
minnieliu: Sounds good, let's check with the Azure board for this. Since the different parts of MC (layouts, inputs, outputs) have some relationship with each other, it's nice to see the full composition after an update has been made.

sdk/communication/Azure.Communication.MediaComposition/api/Azure.Communication.MediaComposition.netstandard2.0.cs
minnieliu: Discussed offline, will go with `UpsertInputs` for now

sdk/communication/Azure.Communication.MediaComposition/api/Azure.Communication.MediaComposition.netstandard2.0.cs
minnieliu: Hmm I tried doing this but it's a bit tricky because it conflicts with the namespace. 
I'll address this in the next PR where we will have a bunch of swagger changes and think more about renaming this

sdk/communication/Azure.Communication.MediaComposition/api/Azure.Communication.MediaComposition.netstandard2.0.cs
DominikMe: I don't think we have a place yet, the current thread is with the REST API Stewardship board. We could ask an architect directly or wait until we have a SDK review with the Azure SDK board. We can also start by asking that question in our ACS API & Design Discussion channel and then Mikhail/Prathmesh/Petr/Rodrigo can weigh in.

DominikMe (approved):

PR #30058, opened 2022-07-21
Address apiview comments for IoT, StoragePool 0721

ArthurMa1978:


sdk/iothub/Azure.ResourceManager.IotHub/src/autorest.md
```diff
@@ -95,6 +102,7 @@ format-by-name-rules:
   '*Uris': 'Uri'
   'thumbprint': 'any'
   'certificate': 'any'
+  'UserAssignedIdentity': 'arm-id'
```
ArcturusZhang, line 105: I did not see a change related with this, maybe it does not take effect

ArthurMa1978: updated, forget to regen the api file

ArcturusZhang (approved):

PR #30061, opened 2022-07-21
[Messaging] Client Lifetime Docs

jsquire:
# Summary

The focus of these changes is to add the remarks for client lifetime to the Service Bus clients and tune the verbiage for the Event Hubs versions.


jsquire:
@JoshLove-msft:  I intentionally didn't add this to the `ServiceBusSessionReceiver`, as my understanding is that should be scoped to just a single session and then closed.   Please let me know if that isn't the case.

JoshLove-msft (approved):

PR #30067, opened 2022-07-21
Respect SessionIdleTimeout for batch functions

### LINKED ISSUE #30043: [Service Bus Extensions] SessionIdleTimeout not being applied for batch functions
This option should also work for batch functions.

JoshLove-msft:
Fixes https://github.com/Azure/azure-sdk-for-net/issues/30043

### DIFF
--- a/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusClientFactory.cs
+++ b/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusClientFactory.cs
@@ -35,7 +35,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.ServiceBus.Config
             logForwarder.Start();
         }
 
-        internal ServiceBusClient CreateClientFromSetting(string connection)
+        internal virtual ServiceBusClient CreateClientFromSetting(string connection)
         {
             var connectionInfo = ResolveConnectionInformation(connection);
 
--- a/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusClientFactory.cs
+++ b/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusClientFactory.cs
@@ -35,7 +35,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.ServiceBus.Config
             logForwarder.Start();
         }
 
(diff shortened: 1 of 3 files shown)


sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusListener.cs
```diff
@@ -390,10 +393,13 @@ private async Task RunBatchReceiveLoopAsync(CancellationToken cancellationToken)
                         }
                     }
 
-                    IReadOnlyList<ServiceBusReceivedMessage> messages =
-                        await receiver.ReceiveMessagesAsync(
-                            _serviceBusOptions.MaxMessageBatchSize,
-                            cancellationToken: cancellationToken).AwaitWithCancellation(cancellationToken);
```
JoshLove-msft, line 396: AwaitWithCancellation is no longer needed as underlying SDK handles cancelling in progress receives now.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusClientFactory.cs
```diff
@@ -35,7 +35,7 @@ internal class ServiceBusClientFactory
             logForwarder.Start();
         }
 
-        internal ServiceBusClient CreateClientFromSetting(string connection)
+        internal virtual ServiceBusClient CreateClientFromSetting(string connection)
```
JoshLove-msft, line 38: For testing

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusListener.cs
```diff
@@ -179,19 +179,22 @@ public async Task StopAsync(CancellationToken cancellationToken)
                 // CloseAsync method stop new messages from being processed while allowing in-flight messages to be processed.
                 if (_singleDispatch)
                 {
-                    if (_isSessionsEnabled)
+                    if (_sessionMessageProcessor.IsValueCreated)
```
JoshLove-msft, line 182: This might be introducing a subtle race condition if the processor somehow is created after this check. It would still potentially get cleaned up in Dispose, but that would again be a race condition. Probably safest to just use the isSessionsEnabled check.

jsquire (approved):

PR #30069, opened 2022-07-21
bug fix - default location

vishweshbankwar:
The default location %temp% should be tried if %localappdata% one fails or not found[this was not covered before]. Will do a follow-up PR for tests.


rajkumar-rangaraj (approved):
We could also avoid a nesting here on else. We could cover in a different PR.

```
if (s_defaultStorageDirectory != null)
            {
                return s_defaultStorageDirectory;
            }
            else
            {

```

PR #30073, opened 2022-07-22
[Event Hub] No longer create a copy of the array in the message converter for the default and stream case

danielmarbach:
Similar to https://github.com/Azure/azure-sdk-for-net/pull/29946 to make @jsquire happy

### DIFF
--- a/sdk/eventhub/Azure.Messaging.EventHubs/CHANGELOG.md
+++ b/sdk/eventhub/Azure.Messaging.EventHubs/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## 5.8.0-beta.1 (Unreleased)
 
+### Acknowledgments
+
+Thank you to our developer community members who helped to make the Event Hubs client libraries better with their contributions to this release:
+
+- Daniel Marbach _([GitHub](https://github.com/danielmarbach))_
+
 ### Features Added
 
 ### Breaking Changes
--- a/sdk/eventhub/Azure.Messaging.EventHubs/CHANGELOG.md
+++ b/sdk/eventhub/Azure.Messaging.EventHubs/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ### Other Changes
 
+- Reduced memory allocations when converting messages into the underlying AMQP primitives. _(A community contribution, courtesy of [danielmarbach](https://github.com/danielmarbach))_
+
 ## 5.7.1 (2022-07-07)
 
 ### Acknowledgments
@@ -24,7 +26,7 @@ Thank you to our developer community members who helped to make the Event Hubs c
 
 ### Bugs Fixed
 
-- Fixed an issue with the `EventHubBufferedProducerClient` where it was not properly identifying when buffers were empty and should enter an idle state; this caused the background task that manages publishing to spin and consume an unreasonable amount of resources. 
+- Fixed an issue with the `EventHubBufferedProducerClient` where it was not properly identifying when buffers were empty and should enter an idle state; this caused the background task that manages publishing to spin and consume an unreasonable amount of resources.
 
 - Fixed an issue with event processor startup validation where an invalid consumer group was not properly detected.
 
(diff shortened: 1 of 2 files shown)


danielmarbach:
The failing test is passing for me locally. Could this be a race and a problem of the closure capturing of the bool in these tests?

sdk/eventhub/Azure.Messaging.EventHubs/src/Amqp/AmqpMessageConverter.cs
```diff
@@ -1208,28 +1208,34 @@ private static bool TryGetValueBody(AmqpMessage source, out AmqpMessageBody valu
...
                 {
                     using var memStreamCopy = new MemoryStream((int)(memStreamSource.Length - memStreamSource.Position));
                     memStreamSource.CopyTo(memStreamCopy, StreamBufferSizeInBytes);
-                    return new ArraySegment<byte>(memStreamCopy.ToArray());
+                    if (!memStreamCopy.TryGetBuffer(out ArraySegment<byte> segment))
+                    {
+                        segment = new ArraySegment<byte>(memStreamCopy.ToArray());
+                    }
+                    return segment;
                 }
```
jsquire, line 1228:
```suggestion
                    if (!memStreamSource.TryGetBuffer(out ArraySegment<byte> segment))
                    {
                        using var memStreamCopy = new MemoryStream((int)(memStreamSource.Length - memStreamSource.Position));
                        memStreamSource.CopyTo(memStreamCopy, StreamBufferSizeInBytes);
                        
                        segment = new ArraySegment<byte>(memStreamCopy.ToArray());
                    }
                    return segment;
                }
```

If you're feeling like trying something, I'm very curious if this would work.   From the docs that you shared on the SB changes, I **_think_** that it should, since we know its a memory stream and has an internal buffer.    If we pass the live suite with this, I'm inclined to take it and push it through our stress tests for a few days to see what happens.

Thoughs?

jsquire:
> closure capturing of the bool in these tests?

It's possible.   That test was not easy to get the order of operations and timing right.   I thought that I had it stabilized, but I may need to try a volatile read in the closure.

sdk/eventhub/Azure.Messaging.EventHubs/CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
 
 ### Other Changes
```
jsquire, line 12: Don't forget to add this sucker before the "Features Added" header:

```markdown
### Acknowledgments

Thank you to our developer community members who helped to make the Event Hubs client libraries better with their contributions to this release:

- Daniel Marbach _([GitHub](https://github.com/danielmarbach))_
```

It's not much but we can at least highlight your efforts rather than burying them inline.  😄

jsquire (commented):
LGTM;  holding off clicking the approval only to hear your thoughts on the proposed experiment.   
_(and a live test run once we decide that we're done)_

sdk/eventhub/Azure.Messaging.EventHubs/src/Amqp/AmqpMessageConverter.cs
danielmarbach: I think the problem is the following: I assume the most likely source here is AmqpMessage. AmqpMessage uses [`MemoryStream(Byte[], Int32, Int32)`](https://docs.microsoft.com/en-us/dotnet/api/system.io.memorystream.-ctor?view=net-6.0#system-io-memorystream-ctor(system-byte()-system-int32-system-int32)) which means `TryGetBuffer` will always return `false`and then we would fall back creating a memory stream with a predefined capacity (of which `TryGetBuffer` would return `true`) only then to write to it and then copy the internal buffer again into the segment because we are calling `ToArray` (which is the unnecessary double allocation this PR gets rid off). 

So what would we gain from this experiment under that promise? Or am I misunderstanding that the sources are most likely AmqpMessages?

What would help is to do a double `TryGetBuffer`. First on the passed stream, then if it returns false allocate the memory stream with the length, copy stuff into it and then `TryGetBuffer` to return it to avoid the `ToArray`. But then again how much value would that add in case we are always getting MemoryStreams managed by AmqpMessage and their internal buffering (still assuming that is indeed the case)?

jsquire: That's where my disconnect is;  the "only works with some constructors".   I think the majority of the time, we'll hit the `BufferListStream` case anyway, coming from `AmqpMessage`, just looking over the [ToStream](https://github.com/Azure/azure-amqp/blob/hotfix/Microsoft.Azure.Amqp/Amqp/AmqpMessage.cs) implementations.    The only other case where we'll hit this method is when someone puts a `Stream` inside the application properties - which is a very rare occurrence.

I agree, not worth the experiment and though the double try-get would guard, unlikely worth the complexity.

jsquire (approved):
LGTM; when you're ready to marge, give a yell.

sdk/eventhub/Azure.Messaging.EventHubs/src/Amqp/AmqpMessageConverter.cs
danielmarbach: How about we merge this as is and I spend some more time after my vacation to think about these scenarios in more depth and then send follow up PRs?

danielmarbach:
From my standpoint this is already a nice improvement and as such valuable and good to go

sdk/eventhub/Azure.Messaging.EventHubs/src/Amqp/AmqpMessageConverter.cs
jsquire: Agreed; not sure how much more we can optimize and we're hitting the majority cases.   Thank you, once again, for all of your work on this!

PR #30075, opened 2022-07-22
[Azure.Communication.Common] - Overridden operators for the CommunicationIdentifier

petrsvihlik:
- related design: https://skype.visualstudio.com/SPOOL/_wiki/wikis/SPOOL.wiki/31075/Design-change-request-RawId-support-for-(de)serialization
- related azure board review item: https://github.com/Azure/azure-sdk/issues/4473


AikoBB (approved):

sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
```diff
@@ -27,6 +27,31 @@ public abstract class CommunicationIdentifier : IEquatable<CommunicationIdentifi
...
 
+        /// <summary>
+        /// Overrides the equality operator.
+        /// </summary>
+        /// <param name="left">The first identifier to compare.</param>
+        /// <param name="right">The second identifier to compare.</param>
+        /// <returns>True if the types and <see cref="RawId"/> match.</returns>
+        public static bool operator ==(CommunicationIdentifier left, CommunicationIdentifier right)
+        {
+            return Equals(left, right) && left.GetType() == right.GetType();
```
DominikMe, line 38: nit: you can use an expression body here and elsewhere =>

sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
```diff
@@ -27,6 +27,31 @@ public abstract class CommunicationIdentifier : IEquatable<CommunicationIdentifi
...
+
+        /// <summary>
+        /// Overrides the non-equality operator.
+        /// </summary>
+        /// <param name="left">The first identifier to compare.</param>
+        /// <param name="right">The second identifier to compare.</param>
+        /// <returns>True if the types or <see cref="RawId"/> values are different.</returns>
+        public static bool operator !=(CommunicationIdentifier left, CommunicationIdentifier right)
+        {
+            return !Equals(left, right) || left.GetType() != right.GetType();
```
DominikMe, line 49: I suspect that `GetType` is the cheaper check so with short-circuiting in mind I would change the order to 
```suggestion
            return left.GetType() != right.GetType() || !Equals(left, right);
```
Same for `==`

sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
```diff
@@ -27,6 +27,31 @@ public abstract class CommunicationIdentifier : IEquatable<CommunicationIdentifi
...
+        }
+
+        /// <summary> Converts a string to a <see cref="CommunicationIdentifier"/>. </summary>
+        public static implicit operator CommunicationIdentifier(string value) => FromRawId(value);
```
DominikMe, line 53: Implicit cast operators can be quite surprising, do we want that you can pass any string to methods that take a CommunicationIdentifier? I don't know if there was any feedback that prompted you to add this conversion.

I would be less concerned if this was an `explicit` cast operator, and the C# docs also say that conversions that can throw shouldn't be implicit: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/user-defined-conversion-operators

DominikMe (commented):
Thanks! 
Nits and I have concerns about the implicit cast.

sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
petrsvihlik: 👍

sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
petrsvihlik: Yes, there was one specific thing that prompted me to add the implicit operator and that was the `CommunicationCloudEnvironment` which also overrides the equality operators. 

https://github.com/Azure/azure-sdk-for-net/blob/aed1cf1cc8f335afd393ca73cec6b0627ce5d382/sdk/communication/Azure.Communication.Common/src/CommunicationCloudEnvironment.cs#L34

I agree that we should stick to the guidelines and use an `explicit` operator. 

Pity that changing it for the `CommunicationCloudEnvironment` would be a breaking change...as it can throw as well.

sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
petrsvihlik: I ran a quick benchmark, and it seems to be the case. well spotted! :)

maximrytych-ms (approved):

DominikMe (approved):

DominikMe:
@petrsvihlik I'm okay with the explicit cast. In documentation do you think we should promote to use `(CommunicationIdentifier) 8:orgid:cf2d38b4-d46e-48cd-b95d-dbe64d99b99d` or `CommunicationIdentifier.FromRawId("8:orgid:cf2d38b4-d46e-48cd-b95d-dbe64d99b99d")?

petrsvihlik:
> @petrsvihlik I'm okay with the explicit cast. In documentation do you think we should promote to use `(CommunicationIdentifier) 8:orgid:cf2d38b4-d46e-48cd-b95d-dbe64d99b99d` or `CommunicationIdentifier.FromRawId("8:orgid:cf2d38b4-d46e-48cd-b95d-dbe64d99b99d")`?

Depending on how the docs will be structured, I'd try to follow the pattern that we usually apply when we use abbreviations - first we use the full form, and the subsequent occurrences use the shortened form.
In this case, I'd probably add a comment to the first occurrence of the explicit cast, explaining that the `FromRawId()` method is used behind the scenes. WDYT?

petrsvihlik:
https://apiview.dev/Assemblies/Review/0d13d0a5ceaa494f8a920f01b59d206d?diffRevisionId=0561b06141784030aa0b3c05529400f4&doc=False&diffOnly=False&revisionId=e5d38220943a4fc5a4a4597f3190b29e

PR #30090, opened 2022-07-22
Feature/calling server v3 events

yyw-msft:
- events added with serialization methods
- abstract participants related events into internal events and public facing events
- added EventParser class to help parse calling server call back events


sdk/communication/Azure.Communication.CallingServer/src/Models/Events/AcsEventType.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Azure.Communication.CallingServer
+{
+    public partial struct AcsEventType
```
minwoolee-msft, line 6: This is overwriting generated AcsEventType right? it is difficult to tell which is overwriting, and which is standalone model. Can you add [CodeGenModel("??")] for all overwriting models?

sdk/communication/Azure.Communication.CallingServer/src/Models/Events/CallingServerEventBase.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Azure.Communication.CallingServer
+{
+    /// <summary>
+     /// The CallingServer Event Base.
+     /// </summary>
+    public partial class CallingServerEventBase
```
minwoolee-msft, line 9: All Events seems to have 1 common property - AcsEventType - is it possible to include that in here in base class?
the benefit is that the SDK user can easily use base classe's EventType to figure out which event it parsed into

jasonshave:
@yyw-msft could you update the description with a code sample of how a developer would use these enhancements? Also, in one of the branches for handling the publishing of events we have a struct called `AcsCallData` which wraps the three main properties we always give back. Here is an example:

```csharp
    public class CallConnectedEvent : AcsCallbackEvent
    {
        public CallConnectedEvent(AcsCallData acsCallData) 
            : base(acsCallData)
        {
        }
    }
```

The `AcsCallbackEvent` in this case consumes the immutable  `AcsCallData` and assigns the properties to the base class. Give me a ring if you want to discuss.

sdk/communication/Azure.Communication.CallingServer/src/Models/Events/AcsEventType.cs
yyw-msft: added!

sdk/communication/Azure.Communication.CallingServer/src/Models/Events/CallingServerEventBase.cs
yyw-msft: Added AcsEventType to the EventBase class. I had the field non-nullable instead of using `AcsEventType?` do u foresee any problem with that? My reasoning is that we should always construct an Event with an AcsEventType. Do you see this being too restrictive?

minwoolee-msft: This is good - one of ACS review board mentioned, considering making this non-nullable. Thank you for pointing out.

sdk/communication/Azure.Communication.CallingServer/src/Models/Events/AddParticipantsFailedEvent.cs
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        /// <summary> Operation context. </summary>
+        public string OperationContext { get; }
+        /// <summary> Gets the result info. </summary>
+        public ResultInformation ResultInfo { get; }
+        /// <summary> Participants failed to be added. </summary>
+        public IReadOnlyList<CommunicationIdentifier> Participants { get; }
+        /// <summary> Gets the type. </summary>
+        public AcsEventType? Type { get; }
```
minwoolee-msft, line 42: Because now we have Base abstract class has EventType, this should go away. All derived events to remove Type.

yyw-msft: updated all the derived event classes! thx for catching it.

yyw-msft:
> @yyw-msft could you update the description with a code sample of how a developer would use these enhancements? Also, in one of the branches for handling the publishing of events we have a struct called `AcsCallData` which wraps the three main properties we always give back. Here is an example:
> 
> ```cs
>   public class CallConnectedEvent : AcsCallbackEvent
>   {
>       public CallConnectedEvent(AcsCallData acsCallData) 
>           : base(acsCallData)
>       {
>       }
>   }
> ```
> 
> The `AcsCallbackEvent` in this case consumes the immutable `AcsCallData` and assigns the properties to the base class. Give me a ring if you want to discuss.

@jasonshave Per our conversation offline, I have removed "event" from all the event classes name and added string manipulation on CloudEvent.Type.

minwoolee-msft (approved):

PR #30112, opened 2022-07-25
Handle tests that call LRO with WaitUntil.Completed

### LINKED ISSUE #29966: [TestFramework][FormRecognizer] Speed up LRO polling for new LRO pattern
## Description

Check https://github.com/Azure/azure-sdk-for-net/pull/29914 for more information about the new LRO pattern that uses the `WaitUntil` enum.

Our Test Framework has some kind of logic to speed up the polling process of LROs in playback mode. For some reason, this logic does not seem to work with the new LRO pattern in Form Recognizer. This is causing the `StartBuildModelWithNeuralBuildMode` and `StartAnalyzeDocumentCanParseMultipageInvoice` tests to time out, so we disabled them for the time being.

Multiple tests were affected in the Form Recognizer test project, but currently the tests mentioned above are the only ones timing out.

## Action Items

- Investigate and fix the issue so that LRO polling takes no extra time with the new pattern.
- Re-enable `StartBuildModelWithNeuralBuildMode` and `StartAnalyzeDocumentCanParseMultipageInvoice`.
- Ensure Form Recognizer ci stays green.

JoshLove-msft:
Fixes https://github.com/Azure/azure-sdk-for-net/issues/29966

### DIFF
--- a/sdk/confidentialledger/Azure.Security.ConfidentialLedger/tests/ConfidentialLedgerClientTests.cs
+++ b/sdk/confidentialledger/Azure.Security.ConfidentialLedger/tests/ConfidentialLedgerClientTests.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Threading.Tasks;
 using Azure.Core;
 using Azure.Core.TestFramework;
 using Azure.Security.ConfidentialLedger.Certificate;
@@ -15,7 +16,7 @@ namespace Azure.Security.ConfidentialLedger.Tests
         private const string transactionId = "1234";
 
         [Test]
-        public void FailedTransaction()
+        public async Task FailedTransaction()
         {
             var client = InstrumentClient(
                 new ConfidentialLedgerClient(
@@ -54,9 +55,9 @@ namespace Azure.Security.ConfidentialLedger.Tests
                             })
                     }
                     ));
-
+            var operation = await client.PostLedgerEntryAsync(WaitUntil.Started, RequestContent.Create(new { contents = "test" }), null, default);
             var ex = Assert.ThrowsAsync<RequestFailedException>(
-                async () => await client.PostLedgerEntryAsync(waitUntil: WaitUntil.Completed, RequestContent.Create(new { contents = "test" }), null, default));
+                async () => await operation.WaitForCompletionResponseAsync());
             Assert.That(ex.Message, Does.Contain(transactionId));
         }
     }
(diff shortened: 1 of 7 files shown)


sdk/core/Azure.Core.TestFramework/src/TestProxy.cs
```diff
@@ -94,7 +94,7 @@ private TestProxy(string proxyPath, bool debugMode = false)
             if (debugMode)
             {
                 _proxyPortHttp = 5000;
-                _proxyPortHttp = 5001;
+                _proxyPortHttps = 5001;
```
JoshLove-msft, line 97: Fixed a couple issues with the new debugMode flag.

sdk/core/Azure.Core.TestFramework/src/Instrumentation/ManagementInterceptor.cs
```diff
@@ -24,50 +24,24 @@ public ManagementInterceptor(ClientTestBase testBase)
 
         public void Intercept(IInvocation invocation)
         {
-            bool modifiedAskToWait = false;
-
-            if (_testMode == RecordedTestMode.Playback && IsLro(invocation.Method.ReturnType))
```
m-nash, line 29: I think we still need this otherwise we will apply zero wait during recording won't we?

JoshLove-msft: Good catch!

m-nash (approved):

kinelski (approved):
LGTM. Thanks for investigating this!

sdk/confidentialledger/Azure.Security.ConfidentialLedger/tests/ConfidentialLedgerClientTests.cs
```diff
@@ -54,9 +55,9 @@ public void FailedTransaction()
                             })
                     }
                     ));
-
+            var operation = await client.PostLedgerEntryAsync(WaitUntil.Started, RequestContent.Create(new { contents = "test" }), null, default);
```
JoshLove-msft, line 58: @christothes I changed the implementation of the test to pass WaitUntil.Started. Otherwise InstrumentResultInterceptor would swap out Completed for Started and then after invoking the method would call waitForCompletion(as per the changes in this PR), but we would then hit a failure in the Diagnostic Scope validation because the scope would not be marked as failed. This points to a more general issue where this swapping mechanism doesn't play nice with tests that expect a failure. Since this was the only such test, and it is not actually even a live test, I figured it was probably fine to just change the test rather than trying to figure out a way to support the test in its original implementation.

Issue #28360, opened 2022-04-24
[Service Bus] Receivers used by Processor should include processor identifier in logs


JoshLove-msft:
The underlying receiver is not exposed to the user, but currently we log these receiver events with a different Identifier than the Processor that created them. The Identifiers are not exposed to customers as a property on either type, but they are emitted in logs. 

We should either include the processor Identifier in the receiver logs, or just make them have the same identifier. The downside to making them have the same identifier would be that it would make it harder to distinguish between receivers when using the session processor for multiple sessions. So it would probably be safer to just add the additional processor Identifier to the receiver logs.

PR #29508, opened 2022-06-27
Add EntityPath and FullyQualifiedNamespace properties to processor args

JoshLove-msft:
Addresses https://github.com/Azure/azure-sdk-for-net/issues/28769 and https://github.com/Azure/azure-sdk-for-net/issues/23582

PR #30136, opened 2022-07-26
Service Bus Client Identifier

m-redding:
The changes here expose the client identifiers for each of the different client types in Service Bus (`ServiceBusClient`, `ServiceBusSender`, `ServiceBusReceiver`, and `ServiceBusProcessor`). They also allow for custom identifiers to be supplied to the clients through their respective options types.

### DIFF
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/api/Azure.Messaging.ServiceBus.netstandard2.0.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/api/Azure.Messaging.ServiceBus.netstandard2.0.cs
@@ -13,6 +13,7 @@ namespace Azure.Messaging.ServiceBus
     }
     public sealed partial class ProcessErrorEventArgs : System.EventArgs
     {
+        public ProcessErrorEventArgs(System.Exception exception, Azure.Messaging.ServiceBus.ServiceBusErrorSource errorSource, string fullyQualifiedNamespace, string entityPath, string identifier, System.Threading.CancellationToken cancellationToken) { }
         public ProcessErrorEventArgs(System.Exception exception, Azure.Messaging.ServiceBus.ServiceBusErrorSource errorSource, string fullyQualifiedNamespace, string entityPath, System.Threading.CancellationToken cancellationToken) { }
         public System.Threading.CancellationToken CancellationToken { get { throw null; } }
         public string EntityPath { get { throw null; } }
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/api/Azure.Messaging.ServiceBus.netstandard2.0.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/api/Azure.Messaging.ServiceBus.netstandard2.0.cs
@@ -96,7 +96,7 @@ namespace Azure.Messaging.ServiceBus
         public ServiceBusClient(string fullyQualifiedNamespace, Azure.Core.TokenCredential credential, Azure.Messaging.ServiceBus.ServiceBusClientOptions options) { }
         public ServiceBusClient(string connectionString, Azure.Messaging.ServiceBus.ServiceBusClientOptions options) { }
         public virtual string FullyQualifiedNamespace { get { throw null; } }
-        public string Identifier { get { throw null; } }
+        public virtual string Identifier { get { throw null; } }
         public virtual bool IsClosed { get { throw null; } }
         public Azure.Messaging.ServiceBus.ServiceBusTransportType TransportType { get { throw null; } }
         public virtual System.Threading.Tasks.Task<Azure.Messaging.ServiceBus.ServiceBusSessionReceiver> AcceptNextSessionAsync(string queueName, Azure.Messaging.ServiceBus.ServiceBusSessionReceiverOptions options = null, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
@@ -252,7 +252,7 @@ namespace Azure.Messaging.ServiceBus
         public virtual bool AutoCompleteMessages { get { throw null; } }
         public virtual string EntityPath { get { throw null; } }
         public virtual string FullyQualifiedNamespace { get { throw null; } }
-        public string Identifier { get { throw null; } }
+        public virtual string Identifier { get { throw null; } }
         public virtual bool IsClosed { get { throw null; } }
         public virtual bool IsProcessing { get { throw null; } }
         public virtual System.TimeSpan MaxAutoLockRenewalDuration { get { throw null; } }
@@ -335,7 +335,7 @@ namespace Azure.Messaging.ServiceBus
         protected ServiceBusReceiver(Azure.Messaging.ServiceBus.ServiceBusClient client, string topicName, string subscriptionName, Azure.Messaging.ServiceBus.ServiceBusReceiverOptions options) { }
         public virtual string EntityPath { get { throw null; } }
         public virtual string FullyQualifiedNamespace { get { throw null; } }
-        public string Identifier { get { throw null; } }
+        public virtual string Identifier { get { throw null; } }
         public virtual bool IsClosed { get { throw null; } }
         public virtual int PrefetchCount { get { throw null; } }
         public virtual Azure.Messaging.ServiceBus.ServiceBusReceiveMode ReceiveMode { get { throw null; } }
@@ -428,7 +428,7 @@ namespace Azure.Messaging.ServiceBus
         protected ServiceBusSender(Azure.Messaging.ServiceBus.ServiceBusClient client, string queueOrTopicName, Azure.Messaging.ServiceBus.ServiceBusSenderOptions options) { }
         public virtual string EntityPath { get { throw null; } }
         public virtual string FullyQualifiedNamespace { get { throw null; } }
-        public string Identifier { get { throw null; } }
+        public virtual string Identifier { get { throw null; } }
         public virtual bool IsClosed { get { throw null; } }
         public virtual System.Threading.Tasks.Task CancelScheduledMessageAsync(long sequenceNumber, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
         public virtual System.Threading.Tasks.Task CancelScheduledMessagesAsync(System.Collections.Generic.IEnumerable<long> sequenceNumbers, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
@@ -467,7 +467,7 @@ namespace Azure.Messaging.ServiceBus
         public virtual bool AutoCompleteMessages { get { throw null; } }
         public virtual string EntityPath { get { throw null; } }
         public virtual string FullyQualifiedNamespace { get { throw null; } }
-        public string Identifier { get { throw null; } }
+        public virtual string Identifier { get { throw null; } }
         protected internal virtual Azure.Messaging.ServiceBus.ServiceBusProcessor InnerProcessor { get { throw null; } }
         public virtual bool IsClosed { get { throw null; } }
         public virtual bool IsProcessing { get { throw null; } }
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/api/Azure.Messaging.ServiceBus.netstandard2.0.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/api/Azure.Messaging.ServiceBus.netstandard2.0.cs
@@ -84,6 +84,7 @@ namespace Azure.Messaging.ServiceBus
         public ServiceBusClient(string fullyQualifiedNamespace, Azure.Core.TokenCredential credential, Azure.Messaging.ServiceBus.ServiceBusClientOptions options) { }
         public ServiceBusClient(string connectionString, Azure.Messaging.ServiceBus.ServiceBusClientOptions options) { }
         public virtual string FullyQualifiedNamespace { get { throw null; } }
+        public string Identifier { get { throw null; } }
         public virtual bool IsClosed { get { throw null; } }
         public Azure.Messaging.ServiceBus.ServiceBusTransportType TransportType { get { throw null; } }
         public virtual System.Threading.Tasks.Task<Azure.Messaging.ServiceBus.ServiceBusSessionReceiver> AcceptNextSessionAsync(string queueName, Azure.Messaging.ServiceBus.ServiceBusSessionReceiverOptions options = null, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
@@ -110,6 +111,7 @@ namespace Azure.Messaging.ServiceBus
         public ServiceBusClientOptions() { }
         public System.Uri CustomEndpointAddress { get { throw null; } set { } }
         public bool EnableCrossEntityTransactions { get { throw null; } set { } }
+        public string Identifier { get { throw null; } }
         public Azure.Messaging.ServiceBus.ServiceBusRetryOptions RetryOptions { get { throw null; } set { } }
         public Azure.Messaging.ServiceBus.ServiceBusTransportType TransportType { get { throw null; } set { } }
         public System.Net.IWebProxy WebProxy { get { throw null; } set { } }
@@ -238,6 +240,7 @@ namespace Azure.Messaging.ServiceBus
         public virtual bool AutoCompleteMessages { get { throw null; } }
         public virtual string EntityPath { get { throw null; } }
         public virtual string FullyQualifiedNamespace { get { throw null; } }
+        public string Identifier { get { throw null; } }
         public virtual bool IsClosed { get { throw null; } }
         public virtual bool IsProcessing { get { throw null; } }
         public virtual System.TimeSpan MaxAutoLockRenewalDuration { get { throw null; } }
@@ -264,6 +267,7 @@ namespace Azure.Messaging.ServiceBus
     {
         public ServiceBusProcessorOptions() { }
         public bool AutoCompleteMessages { get { throw null; } set { } }
+        public string Identifier { get { throw null; } set { } }
         public System.TimeSpan MaxAutoLockRenewalDuration { get { throw null; } set { } }
         public int MaxConcurrentCalls { get { throw null; } set { } }
         public int PrefetchCount { get { throw null; } set { } }
@@ -319,6 +323,7 @@ namespace Azure.Messaging.ServiceBus
         protected ServiceBusReceiver(Azure.Messaging.ServiceBus.ServiceBusClient client, string topicName, string subscriptionName, Azure.Messaging.ServiceBus.ServiceBusReceiverOptions options) { }
         public virtual string EntityPath { get { throw null; } }
         public virtual string FullyQualifiedNamespace { get { throw null; } }
+        public string Identifier { get { throw null; } }
         public virtual bool IsClosed { get { throw null; } }
         public virtual int PrefetchCount { get { throw null; } }
         public virtual Azure.Messaging.ServiceBus.ServiceBusReceiveMode ReceiveMode { get { throw null; } }
@@ -347,6 +352,7 @@ namespace Azure.Messaging.ServiceBus
     public partial class ServiceBusReceiverOptions
     {
         public ServiceBusReceiverOptions() { }
+        public string Identifier { get { throw null; } set { } }
         public int PrefetchCount { get { throw null; } set { } }
         public Azure.Messaging.ServiceBus.ServiceBusReceiveMode ReceiveMode { get { throw null; } set { } }
         public Azure.Messaging.ServiceBus.SubQueue SubQueue { get { throw null; } set { } }
@@ -405,10 +411,11 @@ namespace Azure.Messaging.ServiceBus
     }
     public partial class ServiceBusSender : System.IAsyncDisposable
     {
-        protected ServiceBusSender() { }
+        protected ServiceBusSender(Azure.Messaging.ServiceBus.ServiceBusClient client, string queueOrTopicName) { }
         protected ServiceBusSender(Azure.Messaging.ServiceBus.ServiceBusClient client, string queueOrTopicName, Azure.Messaging.ServiceBus.ServiceBusSenderOptions options) { }
         public virtual string EntityPath { get { throw null; } }
         public virtual string FullyQualifiedNamespace { get { throw null; } }
+        public string Identifier { get { throw null; } }
         public virtual bool IsClosed { get { throw null; } }
         public virtual System.Threading.Tasks.Task CancelScheduledMessageAsync(long sequenceNumber, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
         public virtual System.Threading.Tasks.Task CancelScheduledMessagesAsync(System.Collections.Generic.IEnumerable<long> sequenceNumbers, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/api/Azure.Messaging.ServiceBus.netstandard2.0.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/api/Azure.Messaging.ServiceBus.netstandard2.0.cs
@@ -19,6 +19,7 @@ namespace Azure.Messaging.ServiceBus
         public Azure.Messaging.ServiceBus.ServiceBusErrorSource ErrorSource { get { throw null; } }
         public System.Exception Exception { get { throw null; } }
         public string FullyQualifiedNamespace { get { throw null; } }
+        public string Identifier { get { throw null; } }
     }
     public partial class ProcessMessageEventArgs : System.EventArgs
     {
@@ -26,6 +27,7 @@ namespace Azure.Messaging.ServiceBus
         public System.Threading.CancellationToken CancellationToken { get { throw null; } }
         public string EntityPath { get { throw null; } }
         public string FullyQualifiedNamespace { get { throw null; } }
+        public string Identifier { get { throw null; } }
         public Azure.Messaging.ServiceBus.ServiceBusReceivedMessage Message { get { throw null; } }
         public virtual System.Threading.Tasks.Task AbandonMessageAsync(Azure.Messaging.ServiceBus.ServiceBusReceivedMessage message, System.Collections.Generic.IDictionary<string, object> propertiesToModify = null, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
         public virtual System.Threading.Tasks.Task CompleteMessageAsync(Azure.Messaging.ServiceBus.ServiceBusReceivedMessage message, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
@@ -47,6 +49,7 @@ namespace Azure.Messaging.ServiceBus
         public System.Threading.CancellationToken CancellationToken { get { throw null; } }
         public string EntityPath { get { throw null; } }
         public string FullyQualifiedNamespace { get { throw null; } }
+        public string Identifier { get { throw null; } }
         public string SessionId { get { throw null; } }
         public System.DateTimeOffset SessionLockedUntil { get { throw null; } }
         public virtual System.Threading.Tasks.Task<System.BinaryData> GetSessionStateAsync(System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
@@ -60,6 +63,7 @@ namespace Azure.Messaging.ServiceBus
         public System.Threading.CancellationToken CancellationToken { get { throw null; } }
         public string EntityPath { get { throw null; } }
         public string FullyQualifiedNamespace { get { throw null; } }
+        public string Identifier { get { throw null; } }
         public Azure.Messaging.ServiceBus.ServiceBusReceivedMessage Message { get { throw null; } }
         public string SessionId { get { throw null; } }
         public System.DateTimeOffset SessionLockedUntil { get { throw null; } }
@@ -517,6 +521,7 @@ namespace Azure.Messaging.ServiceBus
     public partial class ServiceBusSessionReceiverOptions
     {
         public ServiceBusSessionReceiverOptions() { }
+        public string Identifier { get { throw null; } set { } }
         public int PrefetchCount { get { throw null; } set { } }
         public Azure.Messaging.ServiceBus.ServiceBusReceiveMode ReceiveMode { get { throw null; } set { } }
         [System.ComponentModel.EditorBrowsableAttribute(System.ComponentModel.EditorBrowsableState.Never)]
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/api/Azure.Messaging.ServiceBus.netstandard2.0.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/api/Azure.Messaging.ServiceBus.netstandard2.0.cs
@@ -455,6 +455,7 @@ namespace Azure.Messaging.ServiceBus
         public virtual bool AutoCompleteMessages { get { throw null; } }
         public virtual string EntityPath { get { throw null; } }
         public virtual string FullyQualifiedNamespace { get { throw null; } }
+        public string Identifier { get { throw null; } }
         protected internal virtual Azure.Messaging.ServiceBus.ServiceBusProcessor InnerProcessor { get { throw null; } }
         public virtual bool IsClosed { get { throw null; } }
         public virtual bool IsProcessing { get { throw null; } }
@@ -488,6 +489,7 @@ namespace Azure.Messaging.ServiceBus
     {
         public ServiceBusSessionProcessorOptions() { }
         public bool AutoCompleteMessages { get { throw null; } set { } }
+        public string Identifier { get { throw null; } set { } }
         public System.TimeSpan MaxAutoLockRenewalDuration { get { throw null; } set { } }
         public int MaxConcurrentCallsPerSession { get { throw null; } set { } }
         public int MaxConcurrentSessions { get { throw null; } set { } }
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/api/Azure.Messaging.ServiceBus.netstandard2.0.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/api/Azure.Messaging.ServiceBus.netstandard2.0.cs
@@ -14,6 +14,7 @@ namespace Azure.Messaging.ServiceBus
     public sealed partial class ProcessErrorEventArgs : System.EventArgs
     {
         public ProcessErrorEventArgs(System.Exception exception, Azure.Messaging.ServiceBus.ServiceBusErrorSource errorSource, string fullyQualifiedNamespace, string entityPath, string identifier, System.Threading.CancellationToken cancellationToken) { }
+        [System.ComponentModel.EditorBrowsableAttribute(System.ComponentModel.EditorBrowsableState.Never)]
         public ProcessErrorEventArgs(System.Exception exception, Azure.Messaging.ServiceBus.ServiceBusErrorSource errorSource, string fullyQualifiedNamespace, string entityPath, System.Threading.CancellationToken cancellationToken) { }
         public System.Threading.CancellationToken CancellationToken { get { throw null; } }
         public string EntityPath { get { throw null; } }
@@ -24,6 +25,8 @@ namespace Azure.Messaging.ServiceBus
     }
     public partial class ProcessMessageEventArgs : System.EventArgs
     {
+        public ProcessMessageEventArgs(Azure.Messaging.ServiceBus.ServiceBusReceivedMessage message, Azure.Messaging.ServiceBus.ServiceBusReceiver receiver, string identifier, System.Threading.CancellationToken cancellationToken) { }
+        [System.ComponentModel.EditorBrowsableAttribute(System.ComponentModel.EditorBrowsableState.Never)]
         public ProcessMessageEventArgs(Azure.Messaging.ServiceBus.ServiceBusReceivedMessage message, Azure.Messaging.ServiceBus.ServiceBusReceiver receiver, System.Threading.CancellationToken cancellationToken) { }
         public System.Threading.CancellationToken CancellationToken { get { throw null; } }
         public string EntityPath { get { throw null; } }
@@ -46,6 +49,8 @@ namespace Azure.Messaging.ServiceBus
     }
     public partial class ProcessSessionEventArgs : System.EventArgs
     {
+        public ProcessSessionEventArgs(Azure.Messaging.ServiceBus.ServiceBusSessionReceiver receiver, string identifier, System.Threading.CancellationToken cancellationToken) { }
+        [System.ComponentModel.EditorBrowsableAttribute(System.ComponentModel.EditorBrowsableState.Never)]
         public ProcessSessionEventArgs(Azure.Messaging.ServiceBus.ServiceBusSessionReceiver receiver, System.Threading.CancellationToken cancellationToken) { }
         public System.Threading.CancellationToken CancellationToken { get { throw null; } }
         public string EntityPath { get { throw null; } }
@@ -60,6 +65,8 @@ namespace Azure.Messaging.ServiceBus
     }
     public partial class ProcessSessionMessageEventArgs : System.EventArgs
     {
+        public ProcessSessionMessageEventArgs(Azure.Messaging.ServiceBus.ServiceBusReceivedMessage message, Azure.Messaging.ServiceBus.ServiceBusSessionReceiver receiver, string identifier, System.Threading.CancellationToken cancellationToken) { }
+        [System.ComponentModel.EditorBrowsableAttribute(System.ComponentModel.EditorBrowsableState.Never)]
         public ProcessSessionMessageEventArgs(Azure.Messaging.ServiceBus.ServiceBusReceivedMessage message, Azure.Messaging.ServiceBus.ServiceBusSessionReceiver receiver, System.Threading.CancellationToken cancellationToken) { }
         public System.Threading.CancellationToken CancellationToken { get { throw null; } }
         public string EntityPath { get { throw null; } }
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/api/Azure.Messaging.ServiceBus.netstandard2.0.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/api/Azure.Messaging.ServiceBus.netstandard2.0.cs
@@ -100,6 +100,7 @@ namespace Azure.Messaging.ServiceBus
         public virtual Azure.Messaging.ServiceBus.ServiceBusReceiver CreateReceiver(string topicName, string subscriptionName, Azure.Messaging.ServiceBus.ServiceBusReceiverOptions options) { throw null; }
         public virtual Azure.Messaging.ServiceBus.ServiceBusRuleManager CreateRuleManager(string topicName, string subscriptionName) { throw null; }
         public virtual Azure.Messaging.ServiceBus.ServiceBusSender CreateSender(string queueOrTopicName) { throw null; }
+        public virtual Azure.Messaging.ServiceBus.ServiceBusSender CreateSender(string queueOrTopicName, Azure.Messaging.ServiceBus.ServiceBusSenderOptions options) { throw null; }
         public virtual Azure.Messaging.ServiceBus.ServiceBusSessionProcessor CreateSessionProcessor(string queueName, Azure.Messaging.ServiceBus.ServiceBusSessionProcessorOptions options = null) { throw null; }
         public virtual Azure.Messaging.ServiceBus.ServiceBusSessionProcessor CreateSessionProcessor(string topicName, string subscriptionName, Azure.Messaging.ServiceBus.ServiceBusSessionProcessorOptions options = null) { throw null; }
         public virtual System.Threading.Tasks.ValueTask DisposeAsync() { throw null; }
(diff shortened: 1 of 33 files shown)


m-redding:
Resolves #28832

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessor.cs
```diff
@@ -71,7 +71,7 @@ public class ServiceBusProcessor : IAsyncDisposable
         /// Gets the ID to identify this processor. This can be used to correlate logs and exceptions.
         /// </summary>
         /// <remarks>Every new processor has a unique ID.</remarks>
-        internal string Identifier { get; }
+        public string Identifier { get; }
```
JoshLove-msft, line 74: Should we consider plumbing the Identifier through to the receivers that the processor uses? Most of the logs generated when using a processor will come from various receivers. See https://github.com/Azure/azure-sdk-for-net/issues/28360. I can see us doing this only if someone sets the identifier (to avoid breaking anyone). Another option would be to add an additional configuration option to indicate whether the processor identifier should be used for all clients that the processor uses.
/cc @jsquire

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessorOptions.cs
```diff
@@ -135,6 +135,12 @@ public int MaxConcurrentCalls
         /// </value>
         public SubQueue SubQueue { get; set; } = SubQueue.None;
 
+        /// <summary>
+        /// Gets the ID to identify this client. This can be used to correlate logs and exceptions.
```
JoshLove-msft, line 139:
```suggestion
        /// Gets the ID to identify this processor. This can be used to correlate logs and exceptions.
```

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessorOptions.cs
```diff
@@ -135,6 +135,12 @@ public int MaxConcurrentCalls
...
+        /// </summary>
+        /// <remarks>Every new client has a unique ID.</remarks>
```
JoshLove-msft, line 141: Do we actually enforce this? What if a user uses the same identifier for multiple processors? Would that cause any issues?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpReceiver.cs
```diff
@@ -55,6 +55,11 @@ internal class AmqpReceiver : TransportReceiver
         /// </value>
         public override bool IsSessionLinkClosed => _isSessionReceiver && LinkException != null;
 
+        /// <summary>
+        /// The identifier for the sender.
```
JoshLove-msft, line 59:
```suggestion
        /// The identifier for the receiver.
```

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Processor/ProcessorLiveTests.cs
```diff
@@ -94,6 +94,83 @@ async Task ProcessMessage(ProcessMessageEventArgs args)
...
+                .Range(0, numThreads)
+                .Select(index => new TaskCompletionSource<bool>(TaskCreationOptions.RunContinuationsAsynchronously))
+                .ToArray();
+                var completionSourceIndex = -1;
+
+                processor.ProcessMessageAsync += ProcessMessage;
+                processor.ProcessErrorAsync += ServiceBusTestUtilities.ExceptionHandler;
+                await processor.StartProcessingAsync();
+
+                async Task ProcessMessage(ProcessMessageEventArgs args)
```
JoshLove-msft, line 139: I think we should probably thread the Identifier through into the various processor message args types - ProcessMessageEventArgs, ProcessErrorEventArgs, ProcessSessionEventArgs, and ProcessSessionMessageEventArgs. We did something very similar recently - https://github.com/Azure/azure-sdk-for-net/pull/29508

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpConnectionScope.cs
```diff
@@ -818,7 +820,7 @@ private async Task<AmqpSession> CreateAndOpenSessionAsync(AmqpConnection connect
                 {
                     Role = false,
                     InitialDeliveryCount = 0,
-                    Source = new Source { Address = Guid.NewGuid().ToString() },
+                    Source = new Source { Address = identifier },
```
JoshLove-msft, line 823: I'm not sure we will leverage this feature yet for Durable Terminus, but my understanding is that link recovery is based on link name. It looks like the LinkName is set on L831 and includes the identifier, but also additional pieces. Should the linkanem just be the identifier? I suppose this could result in uniqueness issues. We may want to confirm with service folks here. /cc @jsquire

JoshLove-msft (commented):
Looks great! I have a few questions/comments about how we envision this setting working with the processor, and with the upcoming durable terminus feature.

sdk/servicebus/Azure.Messaging.ServiceBus/api/Azure.Messaging.ServiceBus.netstandard2.0.cs
```diff
@@ -263,6 +267,7 @@ public partial class ServiceBusProcessorOptions
     {
         public ServiceBusProcessorOptions() { }
         public bool AutoCompleteMessages { get { throw null; } set { } }
+        public string Identifier { get { throw null; } set { } }
```
JoshLove-msft, line 270: I think we will also want to update ServiceBusSessionProcessorOptions and ServiceBusSessionProcessor.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessor.cs
jsquire: I believe that we should.   In Event Hubs, we use the processor's identifier as the basis for the consumers for each partition with the pattern "P{partitionId}-{processor.identifier}";  I think we should consider something similar here.   

@JoshLove-msft:  Does the Service Bus processor have any kind of pattern where it uses multiple receivers outside of sessions where we'd need a pattern?   Should we consider something like "{processor.identifier}-{sessionId}" or similar for sessions?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpConnectionScope.cs
jsquire: We've been using the Address in Event Hubs and in Service Bus for the other languages; I don't **_think_** this is the link name they were talking about for durable terminus, but if it is - yeah, we'll want to make this artificial until we decide to enable the resume feature.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpSender.cs
```diff
@@ -47,15 +47,15 @@ internal class AmqpSender : TransportSender
         public override bool IsClosed => _closed;
 
         /// <summary>
-        ///   The name of the Service Bus entity to which the sender is bound.
+        /// The identifier for the sender.
```
jsquire, line 50:
```suggestion
        /// The identifier of the sender.
```

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessorOptions.cs
jsquire: We don't _(and we probably  probably should be careful about stating it this way)_ - and it should not cause problems outside of log readability or durable terminus state _(eventually)_.

jsquire, line 141:
```suggestion
        /// <summary>
        ///   A unique name used to identify the processor.  If <c>null</c> or empty, a random unique value will be will be used.
        /// </summary>
        ///
```

sdk/servicebus/Azure.Messaging.ServiceBus/src/Receiver/ServiceBusReceiver.cs
```diff
@@ -66,7 +66,7 @@ public class ServiceBusReceiver : IAsyncDisposable
         /// Gets the ID to identify this client. This can be used to correlate logs and exceptions.
         /// </summary>
         /// <remarks>Every new client has a unique ID.</remarks>
```
jsquire, line 68: Here, again, we should be cautious about making this guarantee, since the value is not under our control.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Sender/ServiceBusSender.cs
```diff
@@ -124,6 +129,18 @@ public virtual bool IsClosed
             Logger.ClientCreateComplete(typeof(ServiceBusSender), Identifier);
         }
 
+        /// <summary>
+        ///   Initializes a new instance of the <see cref="ServiceBusSender"/> class.
+        /// </summary>
+        /// <param name="entityPath">The entity path to send the message to.</param>
+        /// <param name="connection">The connection for the sender.</param>
+        internal ServiceBusSender(
```
jsquire, line 137: This may just be the diff looking odd, but did we add additional constructors?

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Amqp/AmqpConnectionScopeTests.cs
```diff
@@ -130,6 +131,130 @@ public async Task OpenAmqpObjectAsyncTranslatesInvalidStateExceptions(Type excep
             mockScope.MockConnection.VerifyAll();
         }
 
+        /// <summary>
+        ///   Verifies functionality of the <see cref="AmqpConnectionScope.OpenSenderLinkAsync(string, string, TimeSpan, CancellationToken)" />
+        ///   method.
+        /// </summary>
+        ///
+        [Test]
+        public async Task OpenLinkAssociatesSendLinkSource()
```
jsquire, line 140: nice!

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Processor/ProcessorLiveTests.cs
jsquire: That's an interesting one.   I hadn't thought of that, but it does make sense in context.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Sender/ServiceBusSender.cs
m-redding: Yes I added a new constructor that takes in the options as well. I was getting API compat errors when just adding a default parameter.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessor.cs
JoshLove-msft: Currently, we don't have any sort of pattern like that. I think this would work though and it would address the concern I mentioned in the issue about customers potentially wanting separate identifiers for the individual receivers, while still allowing them to pattern match on the common processor prefix.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Sender/ServiceBusSender.cs
jsquire: Interesting.   Internal constructors on a public type still flag the analyzer.   I guess there's no real harm in including it rather than suppressing the analyzer.     @JoshLove-msft:  Any concerns?

JoshLove-msft: Hmm, I'm not sure why Api Compat would care about internal constructors. I don't think it matters either way, but I am slightly curious as to what the error was.

m-redding: I believe the error was that the ServiceBusSender(path, connection) didn't exist in implementation but was in the contract

jsquire: oh... API Compat... I thought this was the analyzers.   Yeah, API Compat shouldn't care about the `internal` members.   But we have to be careful with `protected` ones.   If you want a hand looking at the API Compat thing, let me know.

JoshLove-msft: Ah yes, it makes sense that we'd need a new protected ctor, but should be able to reuse the existing internal one.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpConnectionScope.cs
JoshLove-msft: The service team has said that the link recovery is based on the LinkName + ConnectionId.

Right now the Identifier influences a piece of the LinkName. Is the plan that if we need to support link recovery through the LinkIdentifier, that we would add a different setting to specify the link recovery link? Or would we just take durable terminus being enabled to mean that Identifier should be used as the LinkName + ConnectionId?

jsquire: My vote would be the latter.   I don't see a reason to have multiple identifiers for clients with slightly different purposes.    Personally, I'd like to see the ConnectionId stay random today and the LinkName get set to the Identifier.    That way, if we light up terminus recovery, we only have to pipe the `ServiceBusClient.Identifier` into the connection id.

JoshLove-msft: Makes sense to me, but does that mean the Identifier behavior would differ between SB and EH, and .NET SB and other language SB?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessor.cs
m-redding: @JoshLove-msft  that makes sense to me for the `SessionReceiverManager` receivers, but for the `ReceiverManager`s would those just propagate the processor Id and ignore the duplication issue?

sdk/servicebus/Azure.Messaging.ServiceBus/api/Azure.Messaging.ServiceBus.netstandard2.0.cs
m-redding: @JoshLove-msft should I also do the same for the receiver and sender? `ServiceBusSessionReceiver`, `ServiceBusSessionReceiverOptions`, etc?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpConnectionScope.cs
m-redding: @JoshLove-msft @jsquire does it make more sense for the sender link to be associated with the sender identifier or the general service bus client identifier? As it's written these are independent

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpConnectionScope.cs
```diff
@@ -818,15 +820,15 @@ private async Task<AmqpSession> CreateAndOpenSessionAsync(AmqpConnection connect
...
                     Target = new Target { Address = destinationEndpoint.AbsolutePath },
-                    OperationTimeout = _operationTimeout
+                    OperationTimeout = _operationTimeout,
                 };
 
                 linkSettings.AddProperty(AmqpClientConstants.TimeoutName, (uint)timeout.CalculateRemaining(stopWatch.GetElapsedTime()).TotalMilliseconds);
 
                 link = new SendingAmqpLink(linkSettings);
-                linkSettings.LinkName = $"{ Id };{ connection.Identifier }:{ session.Identifier }:{ link.Identifier }";
+                linkSettings.LinkName = Id;
```
m-redding, line 831: Is losing the additional information here a potential issue?

sdk/servicebus/Azure.Messaging.ServiceBus/api/Azure.Messaging.ServiceBus.netstandard2.0.cs
jsquire: If they don't inherit directly from the non-session types (where they'd get the Identifier for free), we should include them.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Client/ServiceBusClient.cs
```diff
@@ -259,7 +259,33 @@ public virtual ServiceBusSender CreateSender(string queueOrTopicName)
 
             return new ServiceBusSender(
                 entityPath: queueOrTopicName,
```
jsquire, line 261: To avoid duplication, should we consider delegating the implementation to the overload and just pass `null`  or `new()` for the `options`?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Client/ServiceBusClientOptions.cs
```diff
@@ -36,6 +36,13 @@ public class ServiceBusClientOptions
         ///
         public IWebProxy WebProxy { get; set; }
 
+        /// <summary>
+        /// Gets the ID to identify this client. This can be used to correlate logs and exceptions.
+        /// </summary>
+        /// <remarks>Every new client has a unique ID.</remarks>
```
jsquire, line 42: nit:  We should consider updating this block to remove the implied guarantee about uniqueness, since we don't have control over the value in all cases.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ProcessErrorEventArgs.cs
```diff
@@ -48,6 +48,9 @@ public sealed class ProcessErrorEventArgs : EventArgs
         /// This is likely to be similar to <c>{yournamespace}.servicebus.windows.net</c>.</summary>
         public string FullyQualifiedNamespace { get; }
 
+        /// <summary>Gets the client identifier associated with the error event.</summary>
+        public string Identifier { get; }
```
jsquire, line 52: Do we need to update/overload the constructor to populate this?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ProcessErrorEventArgs.cs
```diff
@@ -48,6 +48,9 @@ public sealed class ProcessErrorEventArgs : EventArgs
         /// This is likely to be similar to <c>{yournamespace}.servicebus.windows.net</c>.</summary>
         public string FullyQualifiedNamespace { get; }
 
+        /// <summary>Gets the client identifier associated with the error event.</summary>
```
jsquire, line 51:
```suggestion
        /// <summary>Gets the identifier of the processor that raised this event.</summary>
```

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ProcessMessageEventArgs.cs
```diff
@@ -31,6 +31,11 @@ public class ProcessMessageEventArgs : EventArgs
         /// </summary>
         public string EntityPath => _receiver.EntityPath;
 
+        /// <summary>
+        /// The identifier of the Service Bus entity that the message was received from.
```
jsquire, line 35:
```suggestion
        /// The identifier of the processor that raised this event.
```

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusSessionProcessor.cs
```diff
@@ -30,10 +30,9 @@ public class ServiceBusSessionProcessor : IAsyncDisposable
         public virtual string EntityPath => InnerProcessor.EntityPath;
 
         /// <summary>
-        /// Gets the ID to identify this processor. This can be used to correlate logs and exceptions.
+        /// A unique name used to identify the processor.  If <c>null</c> or empty, a random unique value will be will be used.
```
jsquire, line 33: nit:  We should consider updating this block to remove the implied guarantee about uniqueness, since we don't have control over the value in all cases.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Receiver/ServiceBusReceiver.cs
```diff
@@ -63,10 +63,9 @@ public class ServiceBusReceiver : IAsyncDisposable
         public virtual int PrefetchCount { get; }
 
         /// <summary>
-        /// Gets the ID to identify this client. This can be used to correlate logs and exceptions.
+        /// A unique name used to identify the processor.  If <c>null</c> or empty, a random unique value will be will be used.
```
jsquire, line 66: nit:  We should consider updating this block to remove the implied guarantee about uniqueness, since we don't have control over the value in all cases.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Receiver/ServiceBusReceiverOptions.cs
```diff
@@ -40,6 +40,13 @@ public int PrefetchCount
         /// </summary>
         public ServiceBusReceiveMode ReceiveMode { get; set; } = ServiceBusReceiveMode.PeekLock;
 
+        /// <summary>
+        /// Gets the ID to identify this client. This can be used to correlate logs and exceptions.
```
jsquire, line 44: nit:  We should consider updating this block to remove the implied guarantee about uniqueness, since we don't have control over the value in all cases.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Sender/ServiceBusSender.cs
```diff
@@ -66,7 +66,8 @@ public virtual bool IsClosed
         /// Gets the ID to identify this client. This can be used to correlate logs and exceptions.
         /// </summary>
         /// <remarks>Every new client has a unique ID.</remarks>
-        internal string Identifier { get; private set; }
+        ///
```
jsquire, line 69: nit:  We should consider updating this block to remove the implied guarantee about uniqueness, since we don't have control over the value in all cases.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Sender/ServiceBusSenderOptions.cs
```diff
@@ -0,0 +1,60 @@
...
+namespace Azure.Messaging.ServiceBus
+{
+    /// <summary>
+    /// The set of options that can be specified when creating a <see cref="ServiceBusSender"/>
+    /// to configure its behavior.
+    /// </summary>
+    public class ServiceBusSenderOptions
+    {
+        /// <summary>
+        /// A property used to set the <see cref="ServiceBusSender"/> ID to identify the client. This can be used to correlate logs
```
jsquire, line 15: nit:  We should consider updating this block to remove the implied guarantee about uniqueness, since we don't have control over the value in all cases.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpConnectionScope.cs
jsquire: We should use the sender/receiver identifiers with the links.   Eventually, we'll take the `ServiceBusClient` identifier and associate it with the connection... but, for now, we want to keep the connection's name generated so that we don't unintentionally enable a feature.  :)

jsquire: > Makes sense to me, but does that mean the Identifier behavior would differ between SB and EH, and .NET SB and other language SB?

It shouldn't.   Event Hubs uses the identifier as the Source/Target address currently.   That was also in the issue for the other languages to do in SB.    

The Link Name pattern will need to change a bit to make it predictable for durable terminus recovery - so we should probably drop the session/link parts.   That will deviate from EH, but should have the same pattern for SB in other languages after they do the durable terminus stuff.   That said, I've no objection to leaving the link name as-is for now and considering changes as part of the durable terminus story later.    Your call, @JoshLove-msft, I support whichever direction you decide is best.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpConnectionScope.cs
jsquire: Depending on Josh's decision on the direction, we may want to leave this as-is or potentially trim it to:

```csharp
linkSettings.LinkName = $"{ Id };{ connection.Identifier }";
```

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpConnectionScope.cs
JoshLove-msft: As long as there are no concerns from the service side for dropping the session/link part from the session name, I would vote to do it now just in case we end up needing this for durable terminus.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpConnectionScope.cs
JoshLove-msft: Here was the feedback from service team:

> link recovery happens based on linkIdentifier, which is a combination of LinkName coming from linkSettings and also the ConnectionId which comes from the connection

To me, this says that the linkname would need to contain only the Identifier and not any connection information.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Client/ServiceBusClient.cs
```diff
@@ -231,7 +231,7 @@ public ServiceBusClient(string connectionString, ServiceBusClientOptions options
         {
             Logger.ClientCreateStart(typeof(ServiceBusClient), fullyQualifiedNamespace);
             _options = options?.Clone() ?? new ServiceBusClientOptions();
-            Identifier = DiagnosticUtilities.GenerateIdentifier(fullyQualifiedNamespace);
+            Identifier = _options.Identifier ?? DiagnosticUtilities.GenerateIdentifier(fullyQualifiedNamespace);
```
JoshLove-msft, line 234: Should the client identifier be threaded through to anything on the connection?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpConnectionScope.cs
JoshLove-msft: Actually, I think we should just leave as-is for now. We can always modify the behavior based on whether durable terminus receive mode is being used. And there is a good chance we will never enable this part of the durable terminus story either.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Client/ServiceBusClient.cs
m-redding: Custom identifiers would be passed through down to the connection scope through the options, just made a change so the generated one would be as well.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpConnectionScope.cs
m-redding: Sounds good. Switched it back.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessor.cs
JoshLove-msft: Good point, I think for non-session we can just append "-Receiver"

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/SessionReceiverManager.cs
```diff
@@ -52,6 +52,7 @@ internal class SessionReceiverManager : ReceiverManager
             {
                 ReceiveMode = sessionProcessor.InnerProcessor.Options.ReceiveMode,
                 PrefetchCount = sessionProcessor.InnerProcessor.Options.PrefetchCount,
+                Identifier = $"S{sessionId}-{sessionProcessor.Identifier}"
```
JoshLove-msft, line 55: I don't think we can add sessionId here since this could be null for the case where users just want the next available session. Instead we will have to set the Identifier after the ServiceBusSessionReceiver is returned to us on Line 138 based on receiver.SessionId.

JoshLove-msft: I'd also suggest we format as: Identifier-S{sessionId}, so that it is easier to match on the beginning of logs.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Client/ServiceBusClient.cs
JoshLove-msft: Interesting.. I don't have a strong opinion here and would be inclined to leave as-is for non-custom identifier case. Is this the same for EH?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ProcessMessageEventArgs.cs
```diff
@@ -31,6 +31,11 @@ public class ProcessMessageEventArgs : EventArgs
         /// </summary>
         public string EntityPath => _receiver.EntityPath;
 
+        /// <summary>
+        /// The identifier of the processor that raised this event.
+        /// </summary>
+        public string Identifier => _receiver.Identifier;
```
JoshLove-msft, line 37: I think we should probably use the Processor Identifier here.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ProcessErrorEventArgs.cs
```diff
@@ -12,6 +12,33 @@ namespace Azure.Messaging.ServiceBus
...
+        /// <param name="identifier">The identifier of the client entity that reported this error.</param>
+        /// <param name="cancellationToken">The processor's <see cref="System.Threading.CancellationToken"/> instance which will be cancelled
+        /// in the event that <see cref="ServiceBusProcessor.StopProcessingAsync"/> is called.</param>
+        public ProcessErrorEventArgs(
+            Exception exception,
+            ServiceBusErrorSource errorSource,
+            string fullyQualifiedNamespace,
+            string entityPath,
+            string identifier,
+            CancellationToken cancellationToken)
```
JoshLove-msft, line 32: nit: call into the other ctor, and set the identifier in the body of this one.
We should probably hide the other ctor as well with EditorBrowsableNever.

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Processor/ProcessorLiveTests.cs
```diff
@@ -94,6 +94,83 @@ async Task ProcessMessage(ProcessMessageEventArgs args)
...
+                {
+                    Assert.AreEqual(processor.EntityPath, args.EntityPath);
+                    Assert.AreEqual(processor.FullyQualifiedNamespace, args.FullyQualifiedNamespace);
```
JoshLove-msft, line 142: We should validate that the Identifier matches.

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Client/ServiceBusClientLiveTests.cs
```diff
@@ -112,6 +112,42 @@ await using (var client = new ServiceBusClient(TestEnvironment.FullyQualifiedNam
             }
         }
 
+        /// <summary>
+        ///   Verifies that the <see cref="EventHubConnection" /> is able to
```
JoshLove-msft, line 116:
```suggestion
        ///   Verifies that the <see cref="ServiceBusClient" /> is able to
```

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ProcessMessageEventArgs.cs
JoshLove-msft: This also means that we will need a new ctor. Similar to the comment for the Error args, I think we should hide the old ctor.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ProcessSessionEventArgs.cs
```diff
@@ -31,6 +31,11 @@ public class ProcessSessionEventArgs : EventArgs
         /// </summary>
         public string SessionId => _sessionReceiver.SessionId;
 
+        /// <summary>
+        /// The identifier of the Service Bus entity that the message was received from.
+        /// </summary>
+        public string Identifier => _sessionReceiver.Identifier;
```
JoshLove-msft, line 37: Use processor identifier and add new ctor. We might consider just taking the Processor in the new ctor rather than the Identifier so we don't have to keep adding new ctors in the future.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Client/ServiceBusClient.cs
m-redding: It looks like the `AmqpConnectionScope` identifier in EH is also generated randomly independent of the identifiers of the clients

jsquire: There's no equivalent of `ServiceBusClient` in Event Hubs to own the connection uniquely, and no real benefit from setting it to something known.    For Service Bus, we should probably leave it as-is for now and push the `ServiceBusClient` identifier into the connection if/when we choose to allow durable recovery.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/SessionReceiverManager.cs
m-redding: Based on the current pattern the identifier can only be set during creation through the options, because `ServiceBusSessionReceiver.Identifier` is readonly, which is the same as EH, we could do an internal setter on that property or just do something like identifier-sessionreceiver here too if sessionId is null?

JoshLove-msft: Good call - I think the internal setter is a good idea.

sdk/servicebus/Azure.Messaging.ServiceBus/api/Azure.Messaging.ServiceBus.netstandard2.0.cs
```diff
@@ -13,19 +13,22 @@ public partial class CreateMessageBatchOptions
...
     public sealed partial class ProcessErrorEventArgs : System.EventArgs
     {
+        public ProcessErrorEventArgs(System.Exception exception, Azure.Messaging.ServiceBus.ServiceBusErrorSource errorSource, string fullyQualifiedNamespace, string entityPath, string identifier, System.Threading.CancellationToken cancellationToken) { }
         public ProcessErrorEventArgs(System.Exception exception, Azure.Messaging.ServiceBus.ServiceBusErrorSource errorSource, string fullyQualifiedNamespace, string entityPath, System.Threading.CancellationToken cancellationToken) { }
         public System.Threading.CancellationToken CancellationToken { get { throw null; } }
         public string EntityPath { get { throw null; } }
         public Azure.Messaging.ServiceBus.ServiceBusErrorSource ErrorSource { get { throw null; } }
         public System.Exception Exception { get { throw null; } }
         public string FullyQualifiedNamespace { get { throw null; } }
+        public string Identifier { get { throw null; } }
```
jeremymeng, line 23: I got a comment from JS architect about whether `clientId` is better as `clientId` is used in most REST APIs.  Any thoughts? @JoshLove-msft @jsquire

JoshLove-msft: I'd prefer Identifier because we intentionally didn't call most of our types clients, only the ServiceBusClient has client in its name.

jeremymeng: Got! It looks that I have more work to do. I didn't read the description carefully and thought that we only had one identifier on the client that is passing down to all sub types

sdk/servicebus/Azure.Messaging.ServiceBus/src/Client/ServiceBusClient.cs
m-redding: @jsquire should this be true for the custom identifier as well? i.e. not setting connection identifiers based on the options.Identifier? Right now I am doing this in `AmqpConnectionScope`

JoshLove-msft: It may be worth checking to see what was done in the already GA'ed Service Bus libraries (I think Python). If we don't include the identifier in the connection settings, then the only way it is being used is in logs for creating and disposing the client.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Receiver/ServiceBusReceiverOptions.cs
```diff
@@ -40,6 +40,12 @@ public int PrefetchCount
         /// </summary>
         public ServiceBusReceiveMode ReceiveMode { get; set; } = ServiceBusReceiveMode.PeekLock;
 
+        /// <summary>
+        /// A property used to set the <see cref="ServiceBusReceiver"/> ID to identify the client. This can be used to correlate logs
+        /// and exceptions. If <c>null</c> or empty, a random unique value will be used.
+        /// </summary>
```
JoshLove-msft, line 46: Should we also add a remarks section that describes how this is used in the LinkName?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusSessionProcessor.cs
```diff
@@ -30,10 +30,9 @@ public class ServiceBusSessionProcessor : IAsyncDisposable
         public virtual string EntityPath => InnerProcessor.EntityPath;
 
         /// <summary>
-        /// Gets the ID to identify this processor. This can be used to correlate logs and exceptions.
+        /// Gets the Identifier used to identify this processor client.  If <c>null</c> or empty, a random unique value will be will be used.
```
JoshLove-msft, line 33: Seems like we only generate the random ID if it is null, not if it is empty. I'd be okay with updating the behavior to match the docs or vice versa.

jsquire: Event Hubs is testing null or empty:   _([src](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Azure.Messaging.EventHubs/src/Producer/EventHubProducerClient.cs#L245))_

```csharp
Identifier = string.IsNullOrEmpty(clientOptions.Identifier)
    ? Guid.NewGuid().ToString()
    : clientOptions.Identifier;
```

sdk/servicebus/Azure.Messaging.ServiceBus/src/Client/ServiceBusClient.cs
jsquire: In T0, there was a property on the client types for an identifier.   We started adding these back because we've seen a low volume of customer requests along the lines of "I'm migrating and used these properties in my application to know what client this is, since I have several that do different things."

We want the custom identifier on `ServiceBusClient` to be public, but from discussion elsewhere, we don't be using it to set the connection properties just yet.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Receiver/ServiceBusReceiverOptions.cs
jsquire: What benefit would that provide customers?  Personally, I see that as leaking implementation details, but I defer to your judgement.

sdk/servicebus/Azure.Messaging.ServiceBus/api/Azure.Messaging.ServiceBus.netstandard2.0.cs
jsquire: @jeremymeng: There's an established pattern in Event Hubs using `Identifier`; we should look to keep Service Bus and Event Hubs aligned to the same pattern within each language.

I don't see any identifier-like members implemented for Event Hubs JS, so I have no objections to them being called `clientId` there, so long as Brian is comfortable with the drift between languages.

JoshLove-msft: @jsquire - The consideration for not using "client" in the names of the various Service Bus types apart from ServiceBusClient is relevant for all languages, not just .NET.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Receiver/ServiceBusReceiverOptions.cs
JoshLove-msft: Fair point, I withdraw this comment 😆

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpConnectionScope.cs
```diff
@@ -188,7 +189,10 @@ public override bool IsDisposed
             IWebProxy proxy,
             bool useSingleSession,
             TimeSpan operationTimeout,
-            ServiceBusTransportMetrics metrics)
+            ServiceBusTransportMetrics metrics,
+#pragma warning disable CA1801 // Review unused parameters. This parameter will be used for future planned features.
+            string identifier = default)
```
JoshLove-msft, line 194: I would just remove this until we actually need it. Who knows - that day may never actually come and we would be left with this odd parameter.

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Processor/ProcessorLiveTests.cs
JoshLove-msft: Can we add a live test to SessionProcessorLiveTests as well?

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Processor/ProcessorTests.cs
```diff
@@ -125,6 +125,7 @@ public void ProcessorOptionsSetOnClient()
             var fullyQualifiedNamespace = new UriBuilder($"{account}.servicebus.windows.net/").Host;
             var connString = $"Endpoint=sb://{fullyQualifiedNamespace};SharedAccessKeyName=RootManageSharedAccessKey;SharedAccessKey={Encoding.Default.GetString(ServiceBusTestUtilities.GetRandomBuffer(64))}";
             var client = new ServiceBusClient(connString);
+            var identifier = "MyProcessor";
```
JoshLove-msft, line 128: Let's do an equivalent update to SessionProcessorTests.

JoshLove-msft (approved):
This is awesome! It turned out to be way more involved than I initially anticipated but you quickly figured out how to implement everything, and had great design ideas along the way.
One thing that would be really cool, though I don't consider it blocking, would be to add a new test to EventSourceLiveTests that uses and asserts custom identifiers.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Client/ServiceBusClient.cs
JoshLove-msft: I'd suggest that we add it to the connection if/when we actually need it to avoid confusion.

JoshLove-msft:
This PR fixes https://github.com/Azure/azure-sdk-for-net/issues/28360

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Processor/ProcessorTests.cs
m-redding: Sounds good! I'm working on adding in a bunch of tests with all of the new updates now and fixing the pipelines as well.

jsquire (approved):

sdk/servicebus/Azure.Messaging.ServiceBus/api/Azure.Messaging.ServiceBus.netstandard2.0.cs
```diff
@@ -237,6 +252,7 @@ public partial class ServiceBusProcessor : System.IAsyncDisposable
         public virtual bool AutoCompleteMessages { get { throw null; } }
         public virtual string EntityPath { get { throw null; } }
         public virtual string FullyQualifiedNamespace { get { throw null; } }
+        public string Identifier { get { throw null; } }
```
JoshLove-msft, line 255: I think we should probably make this virtual, if for no other reason than to be consistent with the other properties. Not all of them needed to be virtual in the first place, but I think it will be more annoying for devs if some things can be mocked and others can't.

jsquire: I disagree with the need; this property can be directly set by passing options.   Generally we make things virtual when callers can't directly control them.    That said, I don't feel strongly about it, just seems to cut against the normal pattern.

JoshLove-msft: Yeah, as I mentioned the other properties didn't need to be virtual in the first place, but I think it will be a bad user experience if only some of the properties, but not all of them, can be mocked.

PR #30139, opened 2022-07-27
Merge Single- and MultiLabelClassifyResult

### LINKED ISSUE #30089: Merge SingleClassify and MultiClassify response models into a single type

heaths:
Resolves #30089


kristapratico (approved):
LGTM. Were there no instances on the README.md to update?

heaths:
> LGTM. Were there no instances on the README.md to update?

No samples in a README that needed updating, no. We don't want to make READMEs to large. Linking to all the samples at the top like we did with CLU might be a better option for a less-overwhelming README, IMO.

kristapratico:
> > LGTM. Were there no instances on the README.md to update?
> 
> No samples in a README that needed updating, no. We don't want to make READMEs to large. Linking to all the samples at the top like we did with CLU might be a better option for a less-overwhelming README, IMO.

Yeah I think all languages intentionally linked out to the samples in the README to not overcrowd it, but IIRC we had a few instances of calling it "category" instead of "label" on the README.

PR #30155, opened 2022-07-27
Changes that will happen after autorest.csharp PR applied

AlexanderSher:


jsquire:
The associated AutoRest PR has merged; closing.

PR #30160, opened 2022-07-27
Api feedback hashing

jaschrep-msft:
Addresses feedback from https://github.com/Azure/azure-sdk/issues/4495.

Reshapes public API for using transfer validation.


sdk/storage/Azure.Storage.Common/src/Shared/ContentHasher.cs
```diff
@@ -89,17 +89,15 @@ private static void AssertResponseHashMatch(GetHashResult computedHash, Validati
...
-                case ValidationAlgorithm.MD5:
-                    if (!Enumerable.SequenceEqual(
-                        computedHash.MD5,
+                case StorageChecksumAlgorithm.MD5:
+                    if (!computedHash.MD5.Span.SequenceEqual(
                         response.Headers.TryGetValue("Content-MD5", out byte[] md5) ? md5 : default))
                     {
                         throw Errors.HashMismatch("Content-MD5");
                     }
                     break;
```
jaschrep-msft, line 98: @KrzysztofCwalina I'm not the most familiar with Span<T> vs Memory<T> but it seems like all the utility methods are on Span<T> and developers are expected to convert a Memory<T> to a Span<T> to use them. Is this correct?

seanmcc-msft (approved):

sdk/storage/Azure.Storage.Common/src/StorageChecksumAlgorithm.cs
```diff
@@ -15,18 +15,18 @@ public enum ValidationAlgorithm
...
         /// <summary>
-        /// Azure Storage custom 64 bit CRC.
+        /// Standard MD5 hash algorithm.
         /// </summary>
-        StorageCrc64 = 2,
+        MD5 = 2,
 
         /// <summary>
-        /// Standard MD5 hash algorithm.
+        /// Azure Storage custom 64 bit CRC.
```
tg-msft, line 28: Can we add a link for people to learn more?

sdk/storage/Azure.Storage.Common/src/TransferValidationOptions.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// <summary>
+    /// Options for additional content integrity checks on transfer.
+    /// </summary>
+    public class TransferValidationOptions
+    {
+        /// <summary>
+        /// Checksum algorithm to use. If left unset (<see cref="StorageChecksumAlgorithm.Auto"/>),
+        /// the library will pick for you.
+        /// </summary>
+        public UploadTransferValidationOptions Upload { get; } = new();
```
tg-msft, line 17: I think the doc comments need to be updated

sdk/storage/Azure.Storage.Common/src/TransferValidationOptions.cs
```diff
@@ -0,0 +1,27 @@
...
+
+        /// <summary>
+        /// Optional. Can only be specified on specific operations. An existing checksum of
+        /// the data to be uploaded. Not all upload APIs can use this value, and will throw
+        /// if one is provided. Please check documentation on specific APIs for whether this
+        /// can be used.
+        /// </summary>
+        public DownloadTransferValidationOptions Download { get; } = new();
```
tg-msft, line 25: I think the doc comments need to be updated

sdk/storage/Azure.Storage.Files.Shares/src/ShareErrors.cs
```diff
@@ -17,18 +17,18 @@ internal static class ShareErrors
...
-                case ValidationAlgorithm.MD5:
+                case StorageChecksumAlgorithm.None:
+                case StorageChecksumAlgorithm.MD5:
                     return;
-                case ValidationAlgorithm.StorageCrc64:
+                case StorageChecksumAlgorithm.StorageCrc64:
                     throw new ArgumentException("Azure File Shares do not support CRC-64.");
                 default:
-                    throw new ArgumentException($"SDK does not support ValidationAlgorithm value {Enum.GetName(typeof(ValidationAlgorithm), resolved)}.");
+                    throw new ArgumentException($"SDK does not support ValidationAlgorithm value {Enum.GetName(typeof(StorageChecksumAlgorithm), resolved)}.");
```
tg-msft, line 31: Nit - avoid using the term SDK in error messages.  I'm not sure the message is helpful as-is because if I pass a bogus value it won't be able to print anything.  Maybe we want something more like:
```suggestion
                    throw new ArgumentException($"{nameof(StorageChecksumAlgorithm)} does not support value {Enum.GetName(typeof(StorageChecksumAlgorithm), resolved) ?? resolved}.");
```

tg-msft (approved):
I skimmed all the API changes and they look correct.  I believe we'll be able to follow this same pattern when we want to add client-wide StorageTransferOptions.

sdk/storage/Azure.Storage.Common/src/StorageChecksumAlgorithm.cs
jaschrep-msft: There is not currently public documentation. Since we use a custom polynomial, there is an expectation that anyone using this field is letting the client calculate and set it instead of reimplementing 64bit crc themselves.

Issue #29613, opened 2022-07-01
Azure.AI.Personalizer is using a newer version of System.Text.Json


heaths:
_sdk/personalizer/Azure.AI.Personalizer/src/Azure.AI.Personalizer.csproj_ is using a `@VersionOverride` for _System.Text.Json_ which can prevent customers from using this on older versions of .NET. Note that our [guidelines state you must target `netstandard2.0`](https://azure.github.io/azure-sdk/dotnet_introduction.html#dotnet-build-net-standard), and using _System.Text.Json_ 6.0.1 can hinder this.

Looking at the build errors if you remove `@VersionOverride`, it seems you're relying on a new attribute property that should be unnecessary because collection properties should never be null:

* [❌ DO NOT return null values from collection properties or from methods returning collections. Return an empty collection or an empty array instead](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/guidelines-for-collections#:~:text=DO%20NOT%20return%20null%20values%20from%20collection%20properties)

And if you haven't GA'd yet, note that we also have a guideline to use interfaces for collection properties:

* [☑️ YOU SHOULD use the following collection types for properties of model types ...](https://azure.github.io/azure-sdk/dotnet_introduction.html#dotnet-service-return-model-collections)

We have helper classes we use internally for generated code - which we generally recommend you use anyway - but you can use these manually as well. See the following source files for an example:

* https://github.com/Azure/azure-sdk-for-net/blob/350e1ce4fbdf59af14d6df7a991d71ac79666d19/sdk/keyvault/Azure.Security.KeyVault.Administration/src/Generated/Models/KeyVaultRoleDefinition.cs#L57
* https://github.com/Azure/azure-sdk-for-net/blob/350e1ce4fbdf59af14d6df7a991d71ac79666d19/sdk/keyvault/Azure.Security.KeyVault.Administration/src/Generated/Models/KeyVaultRoleDefinition.Serialization.cs#L36-L45

pallavit:
This issue has been fixed - https://github.com/Azure/azure-sdk-for-net/blob/64386052196f8f8293367c8b24dc42ae48d8a7fc/sdk/personalizer/Azure.AI.Personalizer/src/Azure.AI.Personalizer.csproj#L25. Please feel free to reopen if you think otherwise.

PR #30167, opened 2022-07-27
Remove System.Text.Json 6.0.0 override to address #29613

Tparuchuri:
…azure-sdk-for-net/issues/29613 issue

https://msazure.visualstudio.com/One/_backlogs/backlog/DecisionService/Backlog%20items/?workitem=15124797 will make sure object is not used in public API and will be addressed in upcoming sprints at personalizer end


tyclintw (approved):

sdk/personalizer/Azure.AI.Personalizer/src/Models/DecisionContext.cs
```diff
@@ -53,25 +53,22 @@ public DecisionContext(PersonalizerRankMultiSlotOptions rankRequest, Dictionary<
 
         /// <summary> Properties from url </summary>
         [JsonPropertyName("FromUrl")]
-        [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
-        [JsonConverter(typeof(JsonRawStringListConverter))]
-        public List<string> ContextFeatures { get; }
+        public IList<object> ContextFeatures { get; }
```
heaths, line 56: As just an `object`? How would a caller know what to cast this to or use it in general? This doesn't seem like a good UX. /cc @tg-msft

heaths (changes requested):
You can't have something more specific and caller-friendly than `object`?

sdk/personalizer/Azure.AI.Personalizer/src/Models/DecisionContext.cs
Tparuchuri: ContextFeatures comes as a part of PersnalizerRankOptions or PersonalizerRankMultiSlotOptions. In both it's of type object, there is no caller-friendly way to define that. So this should be fine. Even earlier, whatever passed was JSON serialized and assigned to ContextFeatures.

tg-msft: If it's raw JSON, we should expose that as `BinaryData` so you can get a strongly typed C# object, a C# dynamic, a dictionary, ...

Here are some examples:
```C#
using System.Dynamic;
using System.Text.Json;
using System.Text.Json.Serialization;
using Azure;

BinaryData data;
IDictionary<string, object> dict;
StrongType strong;
// anonymous
JsonElement json;
dynamic expando;

//
// Inputs (all print {"value":42})
//

// Dict => BinaryData
data = BinaryData.FromObjectAsJson(new Dictionary<string, object> { { "value", 42 } });
Console.WriteLine(data);

// C# type => BinaryData
data = BinaryData.FromObjectAsJson(new StrongType { Value = 42 });
Console.WriteLine(data);

// Anonmymous C# type => BinaryData
data = BinaryData.FromObjectAsJson(new { value = 42 });
Console.WriteLine(data);

// JsonElement => BinaryData
data = BinaryData.FromObjectAsJson(JsonDocument.Parse("{ \"value\": 42 }").RootElement);
Console.WriteLine(data);

// dynamic => BinaryData
expando = new ExpandoObject();
expando.value = 42;
data = BinaryData.FromObjectAsJson(expando);
Console.WriteLine(data);

//
// Outputs (all print 42)
//

// BinaryData => Dict
dict = data.ToObjectFromJson() as IDictionary<string, object>;
Console.WriteLine(dict["value"]);

// BinaryData => C# type
strong = data.ToObjectFromJson<StrongType>();
Console.WriteLine(strong.Value);

// BinaryData => JsonElement
json = data.ToObjectFromJson<JsonElement>();
Console.WriteLine(json.GetProperty("value").GetInt32());

// BinaryData => dynamic
expando = data.ToObjectFromJson<ExpandoObject>();
Console.WriteLine(expando.value);

class StrongType
{
    [JsonPropertyName("value")]
    public int Value { get; set; }
}
```

Tparuchuri: used BinaryData instead of object

sdk/personalizer/Azure.AI.Personalizer/src/Models/DecisionContext.cs
```diff
@@ -21,11 +25,11 @@ public DecisionContext()
         /// <param name="rankableActions"> Rankable actions </param>
         public DecisionContext(IEnumerable<object> contextFeatures, List<PersonalizerRankableAction> rankableActions)
```
heaths, line 22: The PR mostly looks good, but the idea here was not to ship a public API that used simply `object`, which is what gets passed to this constructor. That should be a list of `BinaryData`. You haven't GA'd yet, so it's acceptable to break the API - preferred, in fact, to make sure we ship good APIs. /cc @tg-msft

Tparuchuri: changed from IEnumerable<object> to IList<BinaryData>

sdk/personalizer/Azure.AI.Personalizer/src/Models/RlObjectConverter.cs
```diff
@@ -17,7 +17,7 @@ internal static class RlObjectConverter
         /// </summary>
         public static string ConvertToContextJson(IEnumerable<object> contextFeatures, List<PersonalizerRankableAction> rankableActions)
```
heaths, line 18: What's passing in this `IEnumerable<object>`? That's what has to change. The public API should not expose just `object` anywhere.

Tparuchuri: modified to BinaryData list

heaths (changes requested):
You still haven't changed the public API. This should be `IEnumerable<BinaryData>` in the public API and passed all the way down:

https://github.com/Azure/azure-sdk-for-net/blob/74268504535ee85f0517a7fc3aff4ed760f16050/sdk/personalizer/Azure.AI.Personalizer/src/Models/PersonalizerClient.cs#L187

I'd also avoid using `ToList()` if you don't need it. In most places, you're just using LINQ i.e., `Enumerable.Select` so collecting items into a list is wasteful and a performance hit.

We should not be exposing `object` in a public API.

Note that when I say "public", I mean in a fully-public API. The members you've changed so far are declared `public`, yes, but the classes they are in are `internal` so they don't actually have public accessibility and are not part of the public API. Basically, anything in your `_api/Azure.AI.Personalizer.netstandard2.0.cs` that gets updated when you run `eng/scripts/CodeChecks.ps1 personalizer` from the repo root (which you'll also need to do or the PR checks will fail).

sdk/personalizer/Azure.AI.Personalizer/src/Models/DecisionContext.cs
heaths: Not in the public API surface, but we've talked about this offline so resolving.

heaths:
Given our offline conversation and the linked bug in the original description, I'm willing to sign off but would prefer either @KrzysztofCwalina or @tg-msft sign off and/or merge this PR if they agree with the resolution.

heaths (approved):

tg-msft (approved):
Yes, we should definitely change the public API surface to `BinaryData` instead of `object` before this GA's.  I've left a comment at https://apiview.dev/Assemblies/Review/35ddf25e338f4bb88d114ab43933c003 to block final approval so we don't forget about it.

PR #30176, opened 2022-07-27
[For Discussion] Draft initial generated client design requirements doc

annelo-msft:
Initial design proposal for model generation.

The goal of this document is to record design decisions that are made for .NET DPG generated clients as a reference for the C# code generation team.


doc/design/GeneratedClientDesign.md
```diff
@@ -0,0 +1,133 @@
...
+## Client Convenience Operations
+-->
+
+## Model Generation
+
+Models are defined as partial classes with one or more constructors, properties, and methods that read from/write to wire formats.
+
+### Model Namespace
+
+Generated models will be defined in a `.Models` namespace.
```
annelo-msft, line 25: Not having the namespace is the most common case, to use as a default.

Do other languages care?  C# only?
@tg-msft, can you chime in with your thoughts? @KrzysztofCwalina to drive closure.



Only add this if specified in an API review.

Will need Cadl support for specifying to use the namespace.

doc/design/GeneratedClientDesign.md
```diff
@@ -0,0 +1,133 @@
...
+------------------- | -------- | -- | -- | -------------
+string | string | string | [autorest.csharp #2337](https://github.com/Azure/autorest.csharp/issues/2337) | [PrimitivePropertyModel.cs](https://github.com/annelo-msft/azure-sdk-for-net/blob/cadl-model-primitive-properties/sdk/template/Azure.Template/src/Generated/Models/PrimitivePropertyModel.cs) [PrimitivePropertyModel.Serialization.cs](https://github.com/annelo-msft/azure-sdk-for-net/blob/cadl-model-primitive-properties/sdk/template/Azure.Template/src/Generated/Models/PrimitivePropertyModel.Serialization.cs)
+bytes | bytes[] | type: string, format: byte | [autorest.csharp #2337](https://github.com/Azure/autorest.csharp/issues/2337) |
+int32  | int | type: integer, format: int32 | [autorest.csharp #2337](https://github.com/Azure/autorest.csharp/issues/2337) |
+int64  | long | type: integer, format: int64 | [autorest.csharp #2337](https://github.com/Azure/autorest.csharp/issues/2337) |
+safeint  | long | n/a | [autorest.csharp #2337](https://github.com/Azure/autorest.csharp/issues/2337) |
+float32  | float | type: number, format: float | [autorest.csharp #2337](https://github.com/Azure/autorest.csharp/issues/2337) |
```
annelo-msft, line 38: Only negative of using double is if these are stored in an array.

If float32 is on a class, make it a double.

Decision: keep as a float.

doc/design/GeneratedClientDesign.md
```diff
@@ -0,0 +1,133 @@
...
+Generated models will be defined in a `.Models` namespace.
+
+### Type mappings from Cadl to .NET
+
+The following table shows a mapping from Cadl built-in types to the corresponding types used in .NET models.
+
+Cadl Type | .NET Type | OpenAPI Type | GitHub Issue | Notes
+------------------- | -------- | -- | -- | -------------
+string | string | string | [autorest.csharp #2337](https://github.com/Azure/autorest.csharp/issues/2337) | [PrimitivePropertyModel.cs](https://github.com/annelo-msft/azure-sdk-for-net/blob/cadl-model-primitive-properties/sdk/template/Azure.Template/src/Generated/Models/PrimitivePropertyModel.cs) [PrimitivePropertyModel.Serialization.cs](https://github.com/annelo-msft/azure-sdk-for-net/blob/cadl-model-primitive-properties/sdk/template/Azure.Template/src/Generated/Models/PrimitivePropertyModel.Serialization.cs)
+bytes | bytes[] | type: string, format: byte | [autorest.csharp #2337](https://github.com/Azure/autorest.csharp/issues/2337) |
```
annelo-msft, line 34: Should it be Memory, ReadOnlyMemory\<T\>?

Need exact size byte array, may not know the exact size.

ReadOnlyMemory, allows slicing down.

Issue 2:
byte array is mutable.
ReadOnlyMemory doesn't allow changes.

Handle differently based on input/output/roundtrip/parameter.

Negatives of ReadOnlyMemory (need to know how used to decide):
- If BCL APIs want byte[], this is a negative.
- In apps, use ReadOnlyMemory
- If in a library, .NET Standard, limitations on how consumed.

When we manually design the APIs, use BinaryData.
Decision: Default to BinaryData, may find cases where this doesn't work.  We'll learn from these going forward.

annelo-msft: Question: extra work for serialization/deserialization logic - @AlexanderSher  to consider and make a recommendation.

doc/design/GeneratedClientDesign.md
```diff
@@ -0,0 +1,133 @@
...
+
+Model APIs will differ depending on whether the model is an input model, an output model, or a round-trip model (both input and output).
+
+The following describes the model shapes for different cases of properties.
+
+ Item |  Input | Output | Round-Trip | GitHub Issue
+-- | -------| ------------ | -------- | -- |
+Constructor Accessibility | public | internal | public |
+Required Property | get-only | get-only | get/set | [autorest.csharp #2463](https://github.com/Azure/autorest.csharp/issues/2463)
+Optional Property | get/set | get-only | get/set | [autorest.csharp #2339](https://github.com/Azure/autorest.csharp/issues/2339)
```
annelo-msft, line 55: Add info about nullable when we address this attribute.

doc/design/GeneratedClientDesign.md
```diff
@@ -0,0 +1,133 @@
...
+## Client Protocol Operations
+
+TBD
+
+## Client Convenience Operations
+-->
+
+## Model Generation
+
+Models are defined as partial classes with one or more constructors, properties, and methods that read from/write to wire formats.
```
annelo-msft, line 21: Don't use C# Records in public APIs.

doc/design/GeneratedClientDesign.md
```diff
@@ -0,0 +1,133 @@
...
+Model APIs will differ depending on whether the model is an input model, an output model, or a round-trip model (both input and output).
+
+The following describes the model shapes for different cases of properties.
+
+ Item |  Input | Output | Round-Trip | GitHub Issue
+-- | -------| ------------ | -------- | -- |
+Constructor Accessibility | public | internal | public |
+Required Property | get-only | get-only | get/set | [autorest.csharp #2463](https://github.com/Azure/autorest.csharp/issues/2463)
+Optional Property | get/set | get-only | get/set | [autorest.csharp #2339](https://github.com/Azure/autorest.csharp/issues/2339)
+Collection Property | `IList<T>` get-only | `IReadOnlyList<T>` get-only | `IList<T>` get-only | [autorest.csharp #2471](https://github.com/Azure/autorest.csharp/issues/2471)
```
annelo-msft, line 56: Proposal - Input models have get/set IList

Negative of settable collection properties
- may be a collection that's not a known type, e.g. could not retrieve items
- easier to modify the collection from a different thread
- could lose observability

We should decide this as a team, should we change from what we've done in the past?
- @annelo-msft to file an issue to track this in the GA timeframe

doc/design/GeneratedClientDesign.md
```diff
@@ -0,0 +1,133 @@
...
+### Model Shape
+
+Model APIs will differ depending on whether the model is an input model, an output model, or a round-trip model (both input and output).
+
+The following describes the model shapes for different cases of properties.
+
+ Item |  Input | Output | Round-Trip | GitHub Issue
+-- | -------| ------------ | -------- | -- |
+Constructor Accessibility | public | internal | public |
+Required Property | get-only | get-only | get/set | [autorest.csharp #2463](https://github.com/Azure/autorest.csharp/issues/2463)
```
annelo-msft, line 54: Check for null on roundtrip/set?
- Based on whether it's nullable or not.

What does required mean in Cadl?

doc/design/GeneratedClientDesign.md
```diff
@@ -0,0 +1,133 @@
...
+### Model Constructors
+
+Models have one or more constructors as follows.
+
+#### Main Constructor
+
+- Accessibility is specified in model shape table above
+- Takes required parameters and does not take optional parameters
+- Validates required reference type parameters for null using `Argument.AssertNotNull`
+- Takes list properties as `IEnumerable<T>` parameters
```
annelo-msft, line 67: Concerns about IEnumerable<int>, or other value types.
Is Cadl expressive enough?

- Could suggest adding decorators to indicate in .NET whether to use collection or array?

For now, let's leave it as IEnumerable, and look for problems and adjust later.

doc/design/GeneratedClientDesign.md
```diff
@@ -0,0 +1,133 @@
...
+- Initializes lists from `IEnumerable` parameter using System.Linq `.ToList()`
```
annelo-msft, line 68: @KrzysztofCwalina will think about this further wrt collection conversation in comment above.

doc/design/GeneratedClientDesign.md
```diff
@@ -0,0 +1,133 @@
...
+
+#### Serialization Constructor
+
+- Internal accessibility
+- Generated for Output and Round-trip types
+- Takes required and optional parameters
+- Takes list properties as `IReadOnlyList<T>` parameters for Output models and `IList<T>` parameters for Round-trip models
```
annelo-msft, line 75: Semantics of first and second constructor are very different.  One makes a copy, the other is a reference so see changes.

This ok because it's an internal constructor used by serializer.

doc/design/GeneratedClientDesign.md
```diff
@@ -0,0 +1,133 @@
...
+  - [RoundTripModel.Serialization.cs](https://github.com/annelo-msft/azure-sdk-for-net/blob/cadl-models-roundtrip-basic/sdk/template/Azure.Template/src/Generated/Models/RoundTripModel.Serialization.cs)
+- Collection properties:
+  - [RoundTripModel.cs](https://github.com/annelo-msft/azure-sdk-for-net/blob/cadl-models-collections-basic/sdk/template/Azure.Template/src/Generated/Models/RoundTripModel.cs)
+  - Collection properties [RoundTripModel.Serialization.cs](https://github.com/annelo-msft/azure-sdk-for-net/blob/cadl-models-collections-basic/sdk/template/Azure.Template/src/Generated/Models/RoundTripModel.Serialization.cs)
+
+### Enums
+
+Enums may be generated as a CLR enum or a C# "Strongly-typed string" as follows.
+
+- Enums are generated in the `.Models` namespace
```
annelo-msft, line 91: Enums follow same pattern as models.

doc/design/GeneratedClientDesign.md
```diff
@@ -0,0 +1,133 @@
...
+- A Cadl `enum` not found in a corresponding  `@knownValues` decorator is generated as a C# `enum`
```
annelo-msft, line 92: @KrzysztofCwalina will follow up on syntax.

doc/design/GeneratedClientDesign.md
```diff
@@ -0,0 +1,133 @@
...
+- Generated enum has corresponding internal extensions file providing conversions to/from string.
+- A Cadl string model with a `@knownValues` decorator is generated as a C# "Strongly-typed string" (sometimes called "extensible enum") and the corresponding Cadl `enum` is not generated as a C# `enum`
+
+GitHub issue: [autorest.csharp #2477](https://github.com/Azure/autorest.csharp/issues/2477)
+
+See examples:
+
+- CLR enum:
+  - [DayOfTheWeek.cs](https://github.com/annelo-msft/azure-sdk-for-net/blob/cadl-models-enum-properties/sdk/template/Azure.Template/src/Generated/Models/DayOfTheWeek.cs)
```
annelo-msft, line 101: Should we assign int values to these?

Would Cadl preserve these?  @annelo-msft to follow up with the Cadl team on this.

If we can't guarantee it, wouldn't want to give a false sense of security.

doc/design/GeneratedClientDesign.md
```diff
@@ -0,0 +1,133 @@
...
+  - [DayOfTheWeek.Serialization.cs](https://github.com/annelo-msft/azure-sdk-for-net/blob/cadl-models-enum-properties/sdk/template/Azure.Template/src/Generated/Models/DayOfTheWeek.Serialization.cs)
```
annelo-msft, line 102: Add a deserialization methoid that takes what Json Dom returns.

doc/design/GeneratedClientDesign.md
```diff
@@ -0,0 +1,133 @@
...
+- Models that only appear as properties of input models are input models
+- Models that only appear as properties of output models are output models
+- Models that only appear as properties of round-trip models are round-trip models
+- Models that appear as properties of more than one model shape are round-trip models
+
+### Model Reference Documentation
+
+### Miscellaneous
+
+- Models should be generated with the `#nullable disable` because Azure SDK libraries have not yet exposed nullable reference types outside of Azure.Core.
```
annelo-msft, line 133: This is because we don't have tools to ensure that APIs would stay compatible in this regard.

doc/design/GeneratedClientDesign.md
tg-msft: Java ALWAYS wants a `.Models` even if it's only got one type.  TS has never done `-models` because they've never had great namespacing options ([though that might be changing?](https://gist.github.com/mpodwysocki/321a24404bd38be4980c331d8da9537c)).  Python uses `-models` sometimes, but not too often from what I can tell.

I would personally lean toward `.Models` as the default for .NET because it leaves more room for growth.  We should explicitly choose not to have a namespace only if we've discussed it with a service and they have high confidence there won't be significant expansion.

doc/design/GeneratedClientDesign.md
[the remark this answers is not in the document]
annelo-msft: Yep!

doc/design/GeneratedClientDesign.md
annelo-msft: Filed https://github.com/Azure/cadl-azure/issues/1787 to track the ask on the Cadl side.

doc/design/GeneratedClientDesign.md
annelo-msft: Tracking with https://github.com/Azure/autorest.csharp/issues/2340

doc/design/GeneratedClientDesign.md
annelo-msft: Tracking with https://github.com/Azure/autorest.csharp/issues/2340

annelo-msft:
@m-nash, could you help us identify if there are any major divergences from management plane's model generation?

doc/design/GeneratedClientDesign.md
annelo-msft: Filed https://github.com/Azure/cadl-azure/issues/1789 to track on the Cadl side.

doc/design/GeneratedClientDesign.md
annelo-msft: See https://github.com/Azure/autorest.csharp/issues/2477#issuecomment-1198696539

doc/design/GeneratedClientDesign.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Model Constructors
+
+Models have one or more constructors as follows.
+
+#### Main Constructor
+
+- Accessibility is specified in model shape table above
+- Takes required parameters and does not take optional parameters
+- Validates required reference type parameters for null using `Argument.AssertNotNull`
```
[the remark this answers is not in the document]
annelo-msft: Tracking with https://github.com/Azure/autorest.csharp/issues/2468.

doc/design/GeneratedClientDesign.md
annelo-msft: Filed https://github.com/Azure/autorest.csharp/issues/2513 to track on the .NET side.

doc/design/GeneratedClientDesign.md
annelo-msft: https://github.com/Azure/autorest.csharp/issues/2338

jsquire (approved):

PR #30177, opened 2022-07-28
Allow diagnostic scopes names without dot

lmolkova:
ServiceBus message scope name consists of one word - `Message`:

https://github.com/azure/azure-sdk-for-net/blob/main/sdk/servicebus/Azure.Messaging.ServiceBus/src/Diagnostics/DiagnosticProperty.cs#L31

However, ActivitySource (OTel) path ignores scopes that don't contain dot, i.e. servicebus tracing does not work in OTel.
We can't change scope name in ServiceBus because it'll affect the stable  (AI SDK) path, so let's just allow such names as it seems to be harmless.

### DIFF
--- a/sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
+++ b/sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
@@ -61,13 +61,12 @@ namespace Azure.Core.Pipeline
                 return null;
             }
 
+            string clientName = ns;
             int indexOfDot = name.IndexOf(".", StringComparison.OrdinalIgnoreCase);
-            if (indexOfDot == -1)
+            if (indexOfDot != -1)
             {
-                return null;
+                clientName += "." + name.Substring(0, indexOfDot);
             }
-
-            string clientName = ns + "." + name.Substring(0, indexOfDot);
 
             return ActivitySources.GetOrAdd(clientName, static n => ActivityExtensions.CreateActivitySource(n));
         }
--- a/sdk/core/Azure.Core/tests/ClientDiagnosticsTests.Net50.cs
+++ b/sdk/core/Azure.Core/tests/ClientDiagnosticsTests.Net50.cs
@@ -98,6 +98,41 @@ namespace Azure.Core.Tests
             }
         }
 
+        [Test]
+        [NonParallelizable]
+        public void StartsActivityShortNameSourceActivity()
+        {
+            using var _ = SetAppConfigSwitch();
+
+            // Bug: there is no way to set activity type to W3C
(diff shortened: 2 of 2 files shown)


sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
```diff
@@ -61,14 +61,13 @@ internal DiagnosticScope(string scopeName, DiagnosticListener source, object? di
                 return null;
             }
 
+            string clientName = ns;
             int indexOfDot = name.IndexOf(".", StringComparison.OrdinalIgnoreCase);
-            if (indexOfDot == -1)
+            if (indexOfDot != -1)
             {
-                return null;
+                clientName += "." + name.Substring(0, indexOfDot);
```
JoshLove-msft, line 68: Don't we still need to include the name in clientName even if it didn't have a dot?

lmolkova: `clientName` becomes ActivitySource name, `name` will go to the Activity Name which will be created bu this source

JoshLove-msft: So, for the example from the ref docs:
ns: Azure.Storage.Blobs
name: BlobClient.DownloadTo
activity source name: Azure.Storage.Blobs.BlobClient
activity name: BlobClient.DownloadTo?

But if the name does not contain a dot:
ns: Azure.Messaging.ServiceBus
name: Message
activity source name: Azure.Messaging.ServiceBus
activity name: Message?

Is this what we want, e.g. not including the name in the activity source name only if there no dot? We repeat the first part of the name in both activity source name and activity name for the dot case, so this strikes me as somewhat inconsistent.

lmolkova: > So, for the example from the ref docs:
ns: Azure.Storage.Blobs
name: BlobClient.DownloadTo
activity source name: Azure.Storage.Blobs.BlobClient
activity name: BlobClient.DownloadTo?

ns: namespace
name == ActivityName: class.method
activitySourceName: namespace.class // logic with dot does it.

For ServiceBus `Message` there is no class. So assuming class is optional, I don't see any inconsistency. I can totally miss something. Can you explain what you see?

JoshLove-msft: Interesting, I was thinking of Message as being the Class but I guess it is more like the operation (or method). Thanks for the explanation!

JoshLove-msft (approved):

PR #30186, opened 2022-07-28
Regenerate SDKs for datetime renaming and list method fix

fengzhou-msft:


sdk/mysql/Azure.ResourceManager.MySql/api/Azure.ResourceManager.MySql.netstandard2.0.cs
```diff
@@ -497,8 +497,8 @@ public partial class MySqlServerResource : Azure.ResourceManager.ArmResource
...
-        public virtual Azure.ResourceManager.ArmOperation<Azure.ResourceManager.MySql.Models.MySqlConfigurationListResult> UpdateConfigurations(Azure.WaitUntil waitUntil, Azure.ResourceManager.MySql.Models.MySqlConfigurationListContent content, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
-        public virtual System.Threading.Tasks.Task<Azure.ResourceManager.ArmOperation<Azure.ResourceManager.MySql.Models.MySqlConfigurationListResult>> UpdateConfigurationsAsync(Azure.WaitUntil waitUntil, Azure.ResourceManager.MySql.Models.MySqlConfigurationListContent content, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
+        public virtual Azure.ResourceManager.ArmOperation<Azure.ResourceManager.MySql.Models.MySqlConfigurationListResult> UpdateConfigurations(Azure.WaitUntil waitUntil, Azure.ResourceManager.MySql.Models.MySqlConfigurationListResult value, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
```
Yao725, line 500: Use `MySqlConfigurationListResult` as the input parameter is weird.

sdk/postgresql/Azure.ResourceManager.PostgreSql/api/Azure.ResourceManager.PostgreSql.netstandard2.0.cs
```diff
@@ -342,8 +342,8 @@ public partial class PostgreSqlServerResource : Azure.ResourceManager.ArmResourc
...
-        public virtual Azure.ResourceManager.ArmOperation<Azure.ResourceManager.PostgreSql.Models.PostgreSqlConfigurationListResult> UpdateConfigurations(Azure.WaitUntil waitUntil, Azure.ResourceManager.PostgreSql.Models.PostgreSqlConfigurationListContent content, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
-        public virtual System.Threading.Tasks.Task<Azure.ResourceManager.ArmOperation<Azure.ResourceManager.PostgreSql.Models.PostgreSqlConfigurationListResult>> UpdateConfigurationsAsync(Azure.WaitUntil waitUntil, Azure.ResourceManager.PostgreSql.Models.PostgreSqlConfigurationListContent content, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
+        public virtual Azure.ResourceManager.ArmOperation<Azure.ResourceManager.PostgreSql.Models.PostgreSqlConfigurationListResult> UpdateConfigurations(Azure.WaitUntil waitUntil, Azure.ResourceManager.PostgreSql.Models.PostgreSqlConfigurationListResult value, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
```
Yao725, line 345: Same as above.

sdk/mysql/Azure.ResourceManager.MySql/api/Azure.ResourceManager.MySql.netstandard2.0.cs
ArcturusZhang: Why we have this change? Which change in the generator introduces this?

sdk/mysql/Azure.ResourceManager.MySql/src/autorest.md
```diff
@@ -161,27 +159,6 @@ directive:
...
-            }
-          },
-          "description": "A list of server configurations."
-        };
-      $.ConfigurationListResult.properties.value.readOnly = true;
-    reason: The generator will not treat the model as the schema for a list method without value being a IReadOnlyList. Need to have separate models for input and output.
-  - from: mysql.json
-    where: $.paths['/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.DBforMySQL/servers/{serverName}/updateConfigurations'].post.parameters[?(@.name === 'value')]
-    transform: >
-      $.schema['$ref'] = $.schema['$ref'].replace('ConfigurationListResult', 'ConfigurationListContent');
```
ArcturusZhang, line 184: OK removal of these configurations is changing the input parameter from XXXContent to a XXXListResult, which is very weird

fengzhou-msft: Previously, the ConfigurationListResult with a settable IList caused GetAll method not recognized. The ConfigurationListContent is a workaround to make it work, but it has a drawback that the input and output of UpdateConfigurations/GetAll method are different, users  need to convert the models when they get a ConfigurationListResult and want to change it a bit, then send it back to update configurations. Therefore, I reverted the change after fix in generator. I will update the model since it can be input as well.

A question to you is that currently UpdateConfigurations (/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.DBforMySQL/servers/{serverName}/updateConfigurations) is put under MySqlServerResource, I think it makes more sense to put it under MySqlConfigurationCollection. Is there a way to do it through autorest.md configurations?

sdk/mysql/Azure.ResourceManager.MySql/api/Azure.ResourceManager.MySql.netstandard2.0.cs
fengzhou-msft: Updated the name.

sdk/postgresql/Azure.ResourceManager.PostgreSql/api/Azure.ResourceManager.PostgreSql.netstandard2.0.cs
fengzhou-msft: Updated.

ArthurMa1978 (approved):

PR #30195, opened 2022-07-28
Remove preview suffix from service version

craigktreasure:
There is a concern that we're not properly communicating that the API is GA since the service version includes a "preview" suffix. Unfortunately, the service GA'd with a preview API version and will not be changing that. This change simply removes the "Preview" suffix in the `ServiceVersion` enum and will continue to route to the preview service API in order to better indicate to consumers that they're calling a GA endpoint.

### DIFF
--- a/sdk/mixedreality/Azure.MixedReality.Authentication/src/MixedRealityStsClientOptions.cs
+++ b/sdk/mixedreality/Azure.MixedReality.Authentication/src/MixedRealityStsClientOptions.cs
@@ -20,11 +20,11 @@ namespace Azure.MixedReality.Authentication
         /// Initializes a new instance of the <see cref="MixedRealityStsClientOptions"/> class.
         /// </summary>
         /// <param name="version">The version.</param>
-        public MixedRealityStsClientOptions(ServiceVersion version = ServiceVersion.V2019_02_28_Preview)
+        public MixedRealityStsClientOptions(ServiceVersion version = ServiceVersion.V2019_02_28)
         {
             Version = version switch
             {
-                ServiceVersion.V2019_02_28_Preview => "2019-02-28-preview",
+                ServiceVersion.V2019_02_28 => "2019-02-28-preview",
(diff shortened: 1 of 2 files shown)


sdk/mixedreality/Azure.MixedReality.Authentication/src/MixedRealityStsClientOptions.cs
```diff
@@ -20,11 +20,11 @@ public class MixedRealityStsClientOptions : ClientOptions
         /// Initializes a new instance of the <see cref="MixedRealityStsClientOptions"/> class.
         /// </summary>
         /// <param name="version">The version.</param>
-        public MixedRealityStsClientOptions(ServiceVersion version = ServiceVersion.V2019_02_28_Preview)
+        public MixedRealityStsClientOptions(ServiceVersion version = ServiceVersion.V2019_02_28)
         {
             Version = version switch
             {
-                ServiceVersion.V2019_02_28_Preview => "2019-02-28-preview",
+                ServiceVersion.V2019_02_28 => "2019-02-28-preview",
```
KrzysztofCwalina, line 27: V2019_02_28_preview should also map to "2019-02-28-preview", otherwise it's a breaking change.

KrzysztofCwalina (changes requested):

sdk/mixedreality/Azure.MixedReality.Authentication/src/MixedRealityStsClientOptions.cs
craigktreasure: Both enum values are both mapped to the value of `1`, so the pattern matching treats them the same. In fact, you get an error if you try to handle them both.

KrzysztofCwalina: Ah, I missed that. Mapping two enum values to the same numeric value is a bit unusual, but I understand why you are doing this.

KrzysztofCwalina (approved):

PR #30199, opened 2022-07-28
AOA Conversion: Update Azure.MixedReality.Authentication

craigktreasure:
This change updates Azure.MixedReality.Authentication and prepares for release.

### DIFF
--- a/sdk/objectanchors/Azure.MixedReality.ObjectAnchors.Conversion/CHANGELOG.md
+++ b/sdk/objectanchors/Azure.MixedReality.ObjectAnchors.Conversion/CHANGELOG.md
@@ -1,14 +1,10 @@
 # Release History
 
-## 0.3.0-beta.4 (Unreleased)
-
-### Features Added
+## 0.3.0-beta.4 (2022-07-29)
 
 ### Breaking Changes
 
-### Bugs Fixed
-
-### Other Changes
+- `ServiceVersion` enum values are now pascal case.
 
 ## 0.3.0-beta.3 (2022-05-11)
 
--- a/sdk/objectanchors/Azure.MixedReality.ObjectAnchors.Conversion/src/Azure.MixedReality.ObjectAnchors.Conversion.csproj
+++ b/sdk/objectanchors/Azure.MixedReality.ObjectAnchors.Conversion/src/Azure.MixedReality.ObjectAnchors.Conversion.csproj
@@ -12,7 +12,7 @@
 
   <ItemGroup>
     <PackageReference Include="Azure.Core" />
-    <PackageReference Include="Azure.MixedReality.Authentication" VersionOverride="1.0.1" />
+    <PackageReference Include="Azure.MixedReality.Authentication" VersionOverride="1.1.0" />
     <PackageReference Include="System.Text.Json" />
   </ItemGroup>
 


sdk/objectanchors/Azure.MixedReality.ObjectAnchors.Conversion/src/Azure.MixedReality.ObjectAnchors.Conversion.csproj
```diff
@@ -12,7 +12,7 @@
 
   <ItemGroup>
     <PackageReference Include="Azure.Core" />
-    <PackageReference Include="Azure.MixedReality.Authentication" VersionOverride="1.0.1" />
+    <PackageReference Include="Azure.MixedReality.Authentication" VersionOverride="1.1.0" />
```
jsquire, line 15: Why do we need to override?  Does something prevent us from bumping the central package version?

sdk/objectanchors/Azure.MixedReality.ObjectAnchors.Conversion/CHANGELOG.md
```diff
@@ -1,15 +1,16 @@
...
+## 0.3.0-beta.4 (2022-07-29)
 
 ### Breaking Changes
 
-### Bugs Fixed
+- `ServiceVersion` enum values are now pascal case.
 
 ### Other Changes
 
+- Updated `Azure.Core` to `1.25.0`.
```
jsquire, line 11: This is an unusual pattern that we don't typically see.  To me, this implies that you'll include all transitive dependency updates in your changelog going forward.   But, that also opens up the complaint that we're not tracking dependencies of dependencies.

In my opinion, it'd be better to not include this unless we have a specific need/reason for doing so.

craigktreasure: Sure. I can remove them. They're required in Java land, which is why I added them.

craigktreasure: Also, these aren't transitive dependencies. These are direct dependencies.

craigktreasure: @jsquire What's the recommendation on direct dependencies?

sdk/objectanchors/Azure.MixedReality.ObjectAnchors.Conversion/src/Azure.MixedReality.ObjectAnchors.Conversion.csproj
craigktreasure: There are repo policies that require you either set the version centrally or you have to override. This library is consumed by a few other Azure SDKs and I imagine this was the solution so that each could update independently.

sdk/objectanchors/Azure.MixedReality.ObjectAnchors.Conversion/CHANGELOG.md
jsquire: The same; we do not typically detail version bumps unless there's some special scenario that would cause customers to have to take action because of the change.  (for example, breaking changes in a dependency)

sdk/objectanchors/Azure.MixedReality.ObjectAnchors.Conversion/src/Azure.MixedReality.ObjectAnchors.Conversion.csproj
jsquire: We very strongly discourage overrides outside of specific temporary needs.   To ensure that the packages in the Azure SDK work together as seamlessly as possible, we want to manage versions centrally and have packages use them.

sdk/objectanchors/Azure.MixedReality.ObjectAnchors.Conversion/CHANGELOG.md
craigktreasure: Thanks for the clarification. I've removed those notes.

jsquire:
> @jsquire Can I get a thumbs up or thumbs down from you?

I'm not comfortable approving with the version override in place as a strategy.   If that's something that you feel strongly about, I'd like to loop in the architects and folks from EngSys for additional thoughts.

craigktreasure:
> > @jsquire Can I get a thumbs up or thumbs down from you?
> 
> I'm not comfortable approving with the version override in place as a strategy. If that's something that you feel strongly about, I'd like to loop in the architects and folks from EngSys for additional thoughts.

That's how it works today. I'm open to the discussion, but I don't see a need for that to block this PR.

jsquire:
> That's how it works today. I'm open to the discussion, but I don't see a need for that to block this PR.

Fair enough; I'll approve and loop in others for the discussion.

jsquire (approved):

jsquire:
@KrzysztofCwalina, @tg-msft, @weshaggard:  I'd appreciate your thoughts on the approach used for package version override as a long-term strategy here.

PR #30201, opened 2022-07-29
Support additional dotnet test filter statements in matrix

benbp:
Support injecting additional dotnet test filter conditions via matrix configs and/or pipeline runtime variables.

### DIFF
--- a/eng/pipelines/templates/variables/globals.yml
+++ b/eng/pipelines/templates/variables/globals.yml
@@ -17,3 +17,7 @@ variables:
   DOTNET_NOLOGO: 1
   Package.EnableSBOMSigning: true
   Codeql.SkipTaskAutoInjection: true
+  # Define optional AdditionalTestFilters via pipeline runtime variables or matrix config variables
+  # Variable value must start with either & or | to join conditions.
+  # See https://docs.microsoft.com/en-us/dotnet/core/testing/selective-unit-tests?pivots=nunit
+  AdditionalTestFilters: ""
--- a/eng/pipelines/templates/variables/globals.yml
+++ b/eng/pipelines/templates/variables/globals.yml
@@ -18,6 +18,5 @@ variables:
   Package.EnableSBOMSigning: true
   Codeql.SkipTaskAutoInjection: true
   # Define optional AdditionalTestFilters via pipeline runtime variables or matrix config variables
-  # Variable value must start with either & or | to join conditions.
-  # See https://docs.microsoft.com/en-us/dotnet/core/testing/selective-unit-tests?pivots=nunit
-  AdditionalTestFilters: ""
+  # See https://docs.microsoft.com/dotnet/core/testing/selective-unit-tests?pivots=nunit
+  AdditionalTestFilters: "Placeholder!=DefaultIgnoreMe"
--- a/eng/pipelines/templates/jobs/ci.tests.yml
+++ b/eng/pipelines/templates/jobs/ci.tests.yml
@@ -64,7 +64,9 @@ jobs:
         parameters:
           NuGetCacheKey: Test
       - script: >-
-          dotnet test eng/service.proj --filter "(TestCategory!=Manually) & (TestCategory!=Live)" --framework $(TestTargetFramework)
+          dotnet test eng/service.proj
+          --filter "(TestCategory!=Manually) & (TestCategory!=Live) $(AdditionalTestFilters)"
+          --framework $(TestTargetFramework)
           --logger "trx;LogFileName=$(TestTargetFramework).trx" --logger:"console;verbosity=normal"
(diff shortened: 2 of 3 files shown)


eng/pipelines/templates/variables/globals.yml
```diff
@@ -17,3 +17,7 @@ variables:
   DOTNET_NOLOGO: 1
   Package.EnableSBOMSigning: true
   Codeql.SkipTaskAutoInjection: true
+  # Define optional AdditionalTestFilters via pipeline runtime variables or matrix config variables
+  # Variable value must start with either & or | to join conditions.
```
jsquire, line 21: This may be a bit obscure and difficult to trip over if the need arises.   Any chance that we've got a wiki or something with pipeline variables that we could add this to?

jsquire (approved):

eng/pipelines/templates/variables/globals.yml
benbp: Yeah I've been trying to think of a good discoverable doc location, if you have ideas. I can create language specific test config docs underneath the broader https://dev.azure.com/azure-sdk/internal/_wiki/wikis/internal.wiki/389/Configuring-Tests doc? The storage virtualized matrix will be adding some example usage of this variable as well soon.

Ideally you wouldn't have to know to prefix with an operator, but I haven't thought of a clean way to chain statements yet either without possibly breaking out into a script and doing some more parsing. But that adds complexity of a different nature as well.

benbp: I'm not sure if this is the best approach but thought I would put it up in case anyone else had alternative ideas.

jsquire: > Yeah I've been trying to think of a good discoverable doc location, if you have ideas. 

Unfortunately, this is one thing that we don't seem to have... but the DevOps wiki seems to align well with most of the EngSys content, so I'd agree that it's probably the best place.

benbp: Ok I added some preliminary docs [here](https://dev.azure.com/azure-sdk/internal/_wiki/wikis/internal.wiki/742/azure-sdk-for-net). As we get more usage examples I'll link them in the docs. I'm hesitant to include a link to the internal docs in the source code though, unless you think I should?

jsquire: No, I agree... probably not.

eng/pipelines/templates/variables/globals.yml
```diff
@@ -17,3 +17,7 @@ variables:
   DOTNET_NOLOGO: 1
   Package.EnableSBOMSigning: true
   Codeql.SkipTaskAutoInjection: true
```
weshaggard, line 19: @benbp did you decide against having some properties similar to java that allow for setting the DefaultTestFilter and then just allow folks that want a custom TestFilter to create the full expression including the DefaultTestFilter (or perhaps excluding it if they don't want that)?

benbp: Sorry for the wall of text but here is my current thinking on this:

I'm still worried about allowing an override of the default, as excluding or modifying it will change the nature of the pipeline, e.g. the only thing that really defines a live test pipeline is resource deployment and the test filter. If it's changeable in a text config, people can see what the default is and extend it. But since it's a variable it would also be accessible at pipeline runtime. I think we would see issues where people forgot to put the full statement in at runtime and/or do not know where to look to find the default, or we would have to set the variable default via pipeline generator somehow.

Also if this variable ends up getting used more widely, I don't want to have to refactor a ton of config files if we ever change the base filters for whatever reason. So for these reasons I think it's better to keep it "private."

We could consider having a powershell processing step to join the values like `(TestCategory==Live) & $(AdditionalTestFilters)` via the vso set variable command, that way users don't need to prefix their statement with a boolean operator? I guess currently it's still possible to ignore the inner test filter by using `|` instead of `&`.

I also tried a noop like `(TestCategory==Live) & ()` which sadly didn't work, but we could consider having the default variable value be something like `(TestCategory==Live) & (TestCategory!=PlaceholderValue)` that way we can hardcode the `&` into the inner value.

weshaggard: Ok I can buy this argument. 

To solve the starts with `&` argument maybe just always have the `&` in the expression and default AddtionalTestFilters to some expression that is always true instead of empty.

eng/pipelines/templates/variables/globals.yml
weshaggard, line 21: The comment about starting with & can be removed now.

weshaggard (approved):
You can clean up the one comment now but otherwise looks good.

eng/pipelines/templates/variables/globals.yml
benbp: Whoops, I updated the wiki doc but missed this one.

PR #30210, opened 2022-07-29
[Reservations] Self review comments

Yao725:


sdk/reservations/Azure.ResourceManager.Reservations/src/autorest.md
```diff
@@ -23,6 +23,16 @@ format-by-name-rules:
...
   '*Uris': 'Uri'
+  "SessionId": "uuid"
+  "DestinationTenantId": "uuid"
+  "BillingScopeId": "arm-id"
+  "SubRequestId": "uuid"
+
+override-operation-name:
+  Reservation_AvailableScopes: GetAvailableScopes
+  CalculateExchange_Post: CalculateReservationExchange
+  Exchange_Post: Exchange
```
ArthurMa1978, line 34: Is single word ok to this method?

ArthurMa1978 (approved):

sdk/reservations/Azure.ResourceManager.Reservations/src/autorest.md
Yao725: Well, I think it's ok because I don't have a better proposal here and we also have single word method name like `Get`, `Update`. Let's use this name and see how the arch board will say.

Issue #29482, opened 2022-06-24
[FormRecognizer] Cross-language consistency renames


kinelski:
After discussing with other languages, .NET has to adopt the following changes to keep consistency:
- [x] Rename any occurrence of `modelDescription` to `description` only. This is a parameter passed when building, copying, or composing models, and could be returned when fetching model info as well.
- [x] Consider renaming the `trainingFilesUri` parameter in the `StartBuildModel` method. Swagger calls it `source`. Do we intend to support other types of source in the future? Could we rename it to `containerUri`?
- [x] Consider renaming `AccountProperties` to `ResourceInfo` (and rename the method `GetAccountProperties` accordingly). This is also beneficial to .NET since we already have an `AccountProperties` class in the `FormRecognizer.Models` namespace. We need to check if any swagger updates are necessary.
- [x] We need to double-check whether the `BoundingPolygon` class is relevant. .NET is the only language that exposes it as a model. Is there any reason for doing so? We could make it an `IReadOnlyList<PointF>`.
- [x] Rename `CopyAuthorization.ExperationDateTime` to `ExpiresOn`.
- [x] Rename `DocumentModelInfo` to `DocumentModelSummary`.

PR #30217, opened 2022-07-29
[FormRecognizer] Flattened DocumentModelOperationDetails

kinelski:
### Changes

- Removed the inheritance between `DocumentModelOperationDetails` and `DocumentModelOperationSummary`. Based on the swagger specification, CodeGen generates `Details` as a subclass of `Summary`. We want to avoid the inheritance in the SDK. If the service ever decides to evolve these differently in the future, we could end up breaking users. An example would be adding a new property to `Summary` but not to `Details`.
- Updated `ResourceLocation` property from string to `Uri`, since that's how we represent URIs in the .NET SDK. This property is defined in both `DocumentModelOperationDetails` and `DocumentModelOperationSummary`.
- Renamed `BuildModel` parameter from `trainingFilesUri` to `blobContainerUri`. This is being changed for consistency with other languages.
- Re-recorded `StartBuildModelWithNeuralBuildMode`. Test was timing out sometimes, so we are increasing the polling intervals, which reduces recording size.

### Strategy for flattening `DocumentModelOperationDetails`

In the swagger, `DocumentModelOperationDetails` and `DocumentModelOperationSummary` are called `GetOperationResponse` and `OperationInfo`, respectively. The only thing the SDK did before this PR was renaming these models in code:

```cs
[CodeGenModel("OperationInfo")]
public class DocumentModelOperationSummary {}

[CodeGenModel("GetOperationResponse")]
public class DocumentModelOperationDetails : DocumentModelOperationSummary {}
```

Since we **cannot** prevent CodeGen from adding the inheritance, we need to create a whole new `DocumentModelOperationDetails` class. The generated one will be hidden from the public API:

```cs
[CodeGenModel("OperationInfo")]
public class DocumentModelOperationSummary {}

// Keep it internal.
[CodeGenModel("GetOperationResponse")]
internal class GetOperationResponse : DocumentModelOperationSummary {}

// NOT generated. Written manually without inheritance.
public class DocumentModelOperationDetails {}
```

Part of #29482.


sdk/formrecognizer/Azure.AI.FormRecognizer/api/Azure.AI.FormRecognizer.netstandard2.0.cs
```diff
@@ -586,11 +586,19 @@ public partial class DocumentModelDetails : Azure.AI.FormRecognizer.DocumentAnal
...
+        public System.DateTimeOffset CreatedOn { get { throw null; } }
         public Azure.ResponseError Error { get { throw null; } }
+        public Azure.AI.FormRecognizer.DocumentAnalysis.DocumentOperationKind Kind { get { throw null; } }
+        public System.DateTimeOffset LastUpdatedOn { get { throw null; } }
+        public string OperationId { get { throw null; } }
+        public int? PercentCompleted { get { throw null; } }
+        public System.Uri ResourceLocation { get { throw null; } }
         public Azure.AI.FormRecognizer.DocumentAnalysis.DocumentModelDetails Result { get { throw null; } }
+        public Azure.AI.FormRecognizer.DocumentAnalysis.DocumentOperationStatus Status { get { throw null; } }
+        public System.Collections.Generic.IReadOnlyDictionary<string, string> Tags { get { throw null; } }
```
kinelski, line 601: These properties already existed in `DocumentModelOperationDetails`, but were inherited from the parent class.

sdk/formrecognizer/Azure.AI.FormRecognizer/src/CopyModelOperation.cs
```diff
@@ -173,9 +173,9 @@ protected CopyModelOperation()
 
         async ValueTask<OperationState<DocumentModelDetails>> IOperation<DocumentModelDetails>.UpdateStateAsync(bool async, CancellationToken cancellationToken)
         {
-            Response<DocumentModelOperationDetails> response = async
-                    ? await _serviceClient.GetOperationAsync(Id, cancellationToken).ConfigureAwait(false)
-                    : _serviceClient.GetOperation(Id, cancellationToken);
+            Response<GetOperationResponse> response = async
+                ? await _serviceClient.GetOperationAsync(Id, cancellationToken).ConfigureAwait(false)
+                : _serviceClient.GetOperation(Id, cancellationToken);
```
kinelski, line 178: Originally we were just renaming the swagger-defined `GetOperationResponse` to `DocumentModelOperationDetails` and exposing it as part of public API.

Since we want to remove the inheritance between it and its base class, we need to create a whole new `DocumentModelOperationDetails` class and hide the original `GetOperationResponse` from the public API.

sdk/formrecognizer/Azure.AI.FormRecognizer/src/DocumentModelOperationDetails.cs
```diff
@@ -4,38 +4,82 @@
...
-    [CodeGenModel("GetOperationResponse")]
-    public partial class DocumentModelOperationDetails
+    /// <summary>
+    /// Details about a document model long-running operation.
+    /// </summary>
+    public class DocumentModelOperationDetails
     {
-        // This property is set by the DocumentAnalysisModelFactory when mocking this class.
-        private readonly ResponseError _mockError;
+        internal DocumentModelOperationDetails(GetOperationResponse response)
```
kinelski, line 15: Convenient constructor that converts from the generated `GetOperationResponse` to the manually written `DocumentModelOperationDetails`.

sdk/formrecognizer/Azure.AI.FormRecognizer/src/DocumentModelOperationSummary.cs
```diff
@@ -23,9 +23,9 @@ public partial class DocumentModelOperationSummary
         public DateTimeOffset LastUpdatedOn { get; }
 
         /// <summary>
-        /// A list of user-defined key-value tag attributes associated with the model.
+        /// URI of the resource targeted by this operation.
         /// </summary>
-        public IReadOnlyDictionary<string, string> Tags { get; }
+        public Uri ResourceLocation { get; }
```
kinelski, line 28: We don't need the `Tags` property in this file. We moved it to the generated part of this class.

sdk/formrecognizer/Azure.AI.FormRecognizer/tests/DocumentModelAdministrationClient/DocumentModelAdministrationClientLiveTests.cs
```diff
@@ -93,7 +93,9 @@ public async Task StartBuildModelWithNeuralBuildMode()
             var trainingFilesUri = new Uri(TestEnvironment.BlobContainerSasUrl);
             var modelId = Recording.GenerateId();
 
-            BuildModelOperation operation = await client.BuildModelAsync(WaitUntil.Completed, trainingFilesUri, DocumentBuildMode.Neural, modelId);
+            BuildModelOperation operation = await client.BuildModelAsync(WaitUntil.Started, trainingFilesUri, DocumentBuildMode.Neural, modelId);
+
+            await operation.WaitForCompletionAsync(TimeSpan.FromMinutes(1));
```
kinelski, line 98: Test was timing out sometimes. This is probably happening because our recordings had around 2000 poll attempts for this LRO. I re-recorded it with a larger poll interval (1 minute), so we have way less attempts now.

sdk/formrecognizer/Azure.AI.FormRecognizer/tests/DocumentModelAdministrationClient/DocumentModelAdministrationClientLiveTests.cs
```diff
@@ -259,17 +261,7 @@ public async Task GetAndListOperations(bool useTokenCredential)
 
             DocumentModelOperationDetails operationDetails = await client.GetOperationAsync(modelOperationFromList.FirstOrDefault().OperationId);
 
-            ValidateOperationSummary(operationDetails);
```
kinelski, line 262: We can't use `ValidateOperationSummary` anymore for this object since we lost the inheritance between `Details` and `Summary`. For this reason I wrote a new `ValidateOperationDetails` method.

sdk/formrecognizer/Azure.AI.FormRecognizer/src/DocumentModelOperationDetails.cs
```diff
@@ -3,39 +3,81 @@
...
+            ResourceLocation = resourceLocation;
+            CreatedOn = createdOn;
+            LastUpdatedOn = lastUpdatedOn;
+            Tags = tags;
+            Error = error;
             Result = result;
         }
 
-        /// <summary> Operation result upon success. </summary>
-        //TODO service is looking into fixing this so it has different return types that we can adapt.
```
kinelski, line 27: Currently investigating what this TODO message meant, but not related to this PR.

kinelski: The `Result` property is actually defined as a generic `object` in the swagger specification. We are, however, forcibly parsing it to a `DocumentModelDetails` reference in the SDK. Even though this is the only type of result currently returned by the service, this could change in the future and break the SDK.

Created issue to track this here: https://github.com/Azure/azure-sdk-for-net/issues/30226

pallavit (approved):
:shipit:

Issue #28215, opened 2022-04-14
[QUERY] Move samples from track 1 SDK to track 2 SDK


EldertGrootenboer:
### Library name and version

Azure.Messaging.ServiceBus

### Query/Question

There are a lot of samples for the [track 1 SDK](https://github.com/Azure/azure-service-bus/tree/master/samples/DotNet/Microsoft.Azure.ServiceBus) which are not available for the [track 2 SDK](https://docs.microsoft.com/en-us/samples/azure/azure-sdk-for-net/azuremessagingservicebus-samples/).
We should convert these, as it will help customers getting onboarded easier on the new SDK.
Below an overview of all the samples, if they have an equivelant in the track 2 samples, and if not, my opinion on priority.

| Track 1                                                                                                                                                                                | Track 2                                                                                                                                                                          | Priority |
| -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------- |
| [TopicFilters](https://github.com/Azure/azure-service-bus/tree/master/samples/DotNet/Microsoft.Azure.ServiceBus/TopicFilters "TopicFilters")                                           |                                                                                                                                                                                  | 0        |
| [TransactionsAndSendVia](https://github.com/Azure/azure-service-bus/tree/master/samples/DotNet/Microsoft.Azure.ServiceBus/TransactionsAndSendVia "TransactionsAndSendVia")             |                                                                                                                                                                                  | 0        |
| [ReceiveLoop](https://github.com/Azure/azure-service-bus/tree/master/samples/DotNet/Microsoft.Azure.ServiceBus/ReceiveLoop "ReceiveLoop")                                              |                                                                                                                                                                                  | 1        |
| [RoleBasedAccessControl](https://github.com/Azure/azure-service-bus/tree/master/samples/DotNet/Microsoft.Azure.ServiceBus/RoleBasedAccessControl "RoleBasedAccessControl")             |                                                                                                                                                                                  | 1        |
| [SendersReceiversWithTopics](https://github.com/Azure/azure-service-bus/tree/master/samples/DotNet/Microsoft.Azure.ServiceBus/SendersReceiversWithTopics "SendersReceiversWithTopics") |                                                                                                                                                                                  | 1        |
[...]

PR #30225, opened 2022-07-29
Add additional topic and sessions samples

JoshLove-msft:
Contributes to https://github.com/Azure/azure-sdk-for-net/issues/28215


sdk/servicebus/Azure.Messaging.ServiceBus/samples/Sample01_HelloWorld.md
```diff
@@ -32,6 +32,37 @@ string body = receivedMessage.Body.ToString();
 Console.WriteLine(body);
 ```
 
+### Send and receive a message using topics and subscriptions
+
+As discussed in the [Key concepts section](https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/servicebus/Azure.Messaging.ServiceBus#key-concepts) of the README, Service Bus supports topics and subscriptions that are useful in pub/sub scenarios. When using topics and subscriptions, each subscription will get its own copy of each message that is delivered to the topic. So completing a message in one subscription won't impact what happens to the message in a different subscription. The types that we use to send and receive messages are the same as the types used to send and receive messages using queues. However, the way we construct them is slightly different. When constructing a `ServiceBusSender` we pass the topic name to the `CreateSender` method. When constructing a `ServiceBusReceiver`, we have to pass both the topic name *and* the subscription name.
```
jsquire, line 37:
```suggestion
As discussed in the [Key concepts section](https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/servicebus/Azure.Messaging.ServiceBus/README.md#key-concepts) of the README, Service Bus supports topics and subscriptions that are useful in pub/sub scenarios. When using topics and subscriptions, each subscription will get its own copy of each message that is delivered to the topic. So completing a message in one subscription won't impact what happens to the message in a different subscription. The types that we use to send and receive messages are the same as the types used to send and receive messages using queues. However, the way we construct them is slightly different. When constructing a `ServiceBusSender` we pass the topic name to the `CreateSender` method. When constructing a `ServiceBusReceiver`, we have to pass both the topic name *and* the subscription name.
```

jsquire, line 37: Should we mention that subscriptions won't see messages in the topic that exist before the subscription was created - or point to service does?   That seems to be a common confusion in SO questions.

sdk/servicebus/Azure.Messaging.ServiceBus/samples/Sample02_MessageSettlement.md
```diff
@@ -1,6 +1,6 @@
...
+This sample demonstrates how to [settle](https://docs.microsoft.com/azure/service-bus-messaging/message-transfers-locks-settlement#settling-receive-operations) received messages. Message settlement can only be used when using a receiver in [PeekLock](https://docs.microsoft.com/azure/service-bus-messaging/message-transfers-locks-settlement#peeklock) mode, which is the default behavior. In order for the settlement operation to be successful, the message must be locked. By default, received messages will be locked for 30 seconds. This can be configured via the portal or when creating the queue or subscription using the `ServiceBusAdministrationClient` or by using the [Azure Resource Managemer API](https://www.nuget.org/packages/Azure.ResourceManager.ServiceBus). Additionally, it is possible to extend the lock for an already received message by using the `RenewMessageLockAsync` method.
```
jsquire, line 3:
```suggestion
This sample demonstrates how to [settle](https://docs.microsoft.com/azure/service-bus-messaging/message-transfers-locks-settlement#settling-receive-operations) received messages. Message settlement can only be used when using a receiver in [PeekLock](https://docs.microsoft.com/azure/service-bus-messaging/message-transfers-locks-settlement#peeklock) mode, which is the default behavior. In order for the settlement operation to be successful, the message must be locked. By default, received messages will be locked for 30 seconds. This can be configured via the portal or when creating the queue or subscription using the `ServiceBusAdministrationClient` or by using the [Azure Resource Manager API](https://www.nuget.org/packages/Azure.ResourceManager.ServiceBus). Additionally, it is possible to extend the lock for an already received message by using the `RenewMessageLockAsync` method.
```

jsquire (approved):

m-redding (approved):
The samples here make sense to me.

PR #30235, opened 2022-08-01
Regen for abstract base class

Yao725:
Part of https://github.com/Azure/autorest.csharp/pull/2520


sdk/search/Azure.Search.Documents/src/Generated/Models/KnowledgeStoreProjectionSelector.cs
```diff
@@ -11,7 +11,7 @@
 namespace Azure.Search.Documents.Indexes.Models
 {
     /// <summary> Abstract class to share properties between concrete selectors. </summary>
-    public partial class KnowledgeStoreProjectionSelector
```
Yao725, line 14: This kind of change in `Azure.Search.Documents` comes from that I unified the judgement condition of abstract class in [ModelWriter](https://github.com/Azure/autorest.csharp/blob/0eec8c79d213dbabe38de654f6c9566113b3425b/src/AutoRest.CSharp/Common/Generation/Writers/ModelWriter.cs#L300) and [SerializationWriter](https://github.com/Azure/autorest.csharp/blob/0eec8c79d213dbabe38de654f6c9566113b3425b/src/AutoRest.CSharp/Common/Generation/Writers/SerializationWriter.cs#L188) of autorest.csharp.

sdk/metricsadvisor/Azure.AI.MetricsAdvisor/src/Generated/Models/DataSourceCredentialEntity.cs
```diff
@@ -11,7 +11,7 @@
 namespace Azure.AI.MetricsAdvisor.Administration
 {
     /// <summary> The DataSourceCredential. </summary>
-    public partial class DataSourceCredentialEntity
+    public abstract partial class DataSourceCredentialEntity
```
kinelski, line 14: While I agree this should have been made `abstract` during development, this class has already been shipped as part of the Metrics Advisor public API, so we can't change it without architect review. Could we somehow skip these changes for Metrics Advisor?

The same reasoning is true for the `MetricFeedback` and the `NotificationHook` classes.

kinelski (changes requested):

sdk/metricsadvisor/Azure.AI.MetricsAdvisor/src/Generated/Models/DataSourceCredentialEntity.cs
Yao725: Reverted this changes for Metrics Advisor.

kinelski (dismissed):
Metrics Advisor LGTM.

PR #30254, opened 2022-08-02
Respect timeout when no cancellation token

JoshLove-msft:
This addresses potential race condition when a SB session is locked right before the client times out. If the client closes the AMQP session, the link will be cleaned up on the service side and the SB session lock will be released.


sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpConnectionScope.cs
```diff
@@ -1118,20 +1124,29 @@ private void StopTrackingLinkAsActive(AmqpObject link, Timer authorizationRefres
...
-            CancellationToken cancellationToken,
-            string entityPath = default)
+            string entityPath = default,
+            TimeSpan? timeout = default,
+            CancellationToken cancellationToken = default)
         {
             try
             {
-                await target.OpenAsync(cancellationToken).ConfigureAwait(false);
+                // Prefer the cancellation token, falling back to the timeout only when
```
JoshLove-msft, line 1139: Copied this from EH

PR #30255, opened 2022-08-02
Respect timeout when no cancellation token passed to OpenAmqpObject

JoshLove-msft:
CancellationToken should be used when opening links, as the link timeout is already set in the link settings which is what the service will use. Using the same timeout on the client could result in race conditions where the client times out first. We could increase the client timeout, but instead we will just rely on the service timeout here.
For opening AMQP session/connection, we use a client timeout because there is no corresponding service level timeout.

### DIFF
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpConnectionScope.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpConnectionScope.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Globalization;
 using System.Net;
 using System.Runtime.ExceptionServices;
@@ -305,7 +306,7 @@ namespace Azure.Messaging.ServiceBus.Amqp
                     timeout.CalculateRemaining(stopWatch.GetElapsedTime()),
                     cancellationToken).ConfigureAwait(false);
 
-                await OpenAmqpObjectAsync(link, entityPath, cancellationToken: cancellationToken).ConfigureAwait(false);
+                await OpenAmqpLinkAsync(link, entityPath, cancellationToken: cancellationToken).ConfigureAwait(false);
                 cancellationToken.ThrowIfCancellationRequested<TaskCanceledException>();
                 ServiceBusEventSource.Log.CreateManagementLinkComplete(identifier);
                 return link;
@@ -346,7 +347,6 @@ namespace Azure.Messaging.ServiceBus.Amqp
             var receiverEndpoint = new Uri(ServiceEndpoint, entityPath);
 
             var connection = await ActiveConnection.GetOrCreateAsync(timeout, cancellationToken).ConfigureAwait(false);
-            cancellationToken.ThrowIfCancellationRequested<TaskCanceledException>();
 
             ReceivingAmqpLink link = await CreateReceivingLinkAsync(
                 entityPath: entityPath,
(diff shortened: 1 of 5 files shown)


jsquire (approved):

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpConnectionScope.cs
```diff
@@ -365,8 +363,7 @@ protected AmqpConnectionScope()
 
             cancellationToken.ThrowIfCancellationRequested<TaskCanceledException>();
 
-            await OpenAmqpObjectAsync(link, timeout.CalculateRemaining(stopWatch.GetElapsedTime()), cancellationToken, entityPath).ConfigureAwait(false);
-            cancellationToken.ThrowIfCancellationRequested<TaskCanceledException>();
```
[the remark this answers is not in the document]
JoshLove-msft: Hah I think this was a holdover from before the cancellation support in AMQP library.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpConnectionScope.cs
```diff
@@ -535,7 +532,7 @@ public override void Dispose()
                 var sessionSettings = new AmqpSessionSettings { Properties = new Fields() };
                 session = connection.CreateSession(sessionSettings);
 
-                await OpenAmqpObjectAsync(session, timeout, cancellationToken, entityPath).ConfigureAwait(false);
+                await OpenAmqpObjectAsync(session, entityPath, timeout, cancellationToken).ConfigureAwait(false);
                 cancellationToken.ThrowIfCancellationRequested<TaskCanceledException>();
```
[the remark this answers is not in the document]
JoshLove-msft: I know the feeling..

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpConnectionScope.cs
```diff
@@ -1115,23 +1101,64 @@ private void StopTrackingLinkAsActive(AmqpObject link, Timer authorizationRefres
...
             AmqpObject target,
-            TimeSpan timeout,
-            CancellationToken cancellationToken,
-            string entityPath = default)
+            string entityPath = default,
+            TimeSpan? timeout = default,
+            CancellationToken? cancellationToken = default)
         {
+            // only one of timeout or cancellation token should be set
+            Debug.Assert(timeout.HasValue ^ cancellationToken.HasValue);
```
m-redding, line 1151: Just curious, why Debug.Assert instead of Assert?

m-redding (approved):

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpConnectionScope.cs
jsquire: Why an assert here at all?

JoshLove-msft: This is meant to check for coding errors so by using Debug.Assert it wouldn't actually run in a release build but we would still be able to catch any mistakes in our debug builds.

JoshLove-msft: > Why an assert here at all?

I suppose it is probably not needed now that I refactored this to be a private method. Initially I had all parameters in the public method. Still it probably doesn't hurt in documenting the intent, and in case we add a new method that has both params.

therealrss:
@JoshLove-msft Quick question about a behavior prior to this fix (SDK 7.7). Would having say an idle queue with many receiver (1k+) where client timeouts before the service, could lead to hitting the concurrent connections limits and trigger requests throttling ?

JoshLove-msft:
> @JoshLove-msft Quick question about a behavior prior to this fix (SDK 7.7). Would having say an idle queue with many receiver (1k+) where client timeouts before the service, could lead to hitting the concurrent connections limits and trigger requests throttling ?

This would only impact the operation of establishing the connection with the service. So long as you are still disposing the ServiceBusClient, it shouldn't cause excess connections.

PR #30256, opened 2022-08-02
Support cancelling AnalyzeActionsOperation

### LINKED ISSUE #26528: [Text Analytics] Re-enable RecognizeHealthcareEntitiesBatchWithCancellation test
The `RecognizeHealthcareEntitiesBatchWithCancellation` test works as follows:
1. Firstly, it sends a request to the service to start a long-running operation.
2. It immediately tries to cancel the operation that has just started.
3. It repeats the process above a few times until cancellation succeeds (pass), or until we exhaust the maximum number of attempts (fail).

We recently moved to a static resource in Text Analytics in order to test new features that use custom models. Because of that, we have started getting multiple 429 (Too Many Requests) errors, and we increased the maximum number of retries in our retry policy.

Even with the higher number of retries, the `RecognizeHealthcareEntitiesBatchWithCancellation` test sometimes fails. Even if we're able to ignore the 429 errors, if we take too long to send the cancellation request to the service, the operation ends up finishing successfully before we can cancel it.

### LINKED ISSUE #28116: [Text Analytics] Support Cancellation of an Analyze Job
# Summary

A recent update to the service added support for cancellation of analyze jobs that have been started.  This should be supported by the client library.

# Scope of Work

- Coordinate with the feature team for a design of the cancellation API

- Implement the API for cancellation in `TextAnalyticsClient`

# Success Criteria

- Cancellation support has been added to the `TextAnalyticsClient` in the agreed upon API format.

- The tests necessary for validation of cancelling a job have been created or adjusted and pass reliably.

- The existing test suite continues to produce deterministic results and pass reliably.

# References and Resources

- [Text Analytics Swagger Update](https://github.com/Azure/azure-rest-api-specs/pull/17281)

heaths:
Resolves #28116 and resolves #26528 by using a mock instead of a recording.


sdk/textanalytics/Azure.AI.TextAnalytics/src/AnalyzeActionsOperation.cs
```diff
@@ -281,33 +298,26 @@ async ValueTask<OperationState<AsyncPageable<AnalyzeActionsResult>>> IOperation<
...
-                    return OperationState<AsyncPageable<AnalyzeActionsResult>>.Failure(rawResponse, requestFailedException);
-                }
-            }
-
-            if (response.Value.Status == TextAnalyticsOperationStatus.Succeeded ||
-                response.Value.Status == TextAnalyticsOperationStatus.Failed)
+            if (response.Value.Status == TextAnalyticsOperationStatus.Succeeded)
             {
                 string nextLink = response.Value.NextLink;
                 _firstPage = Page.FromValues(new List<AnalyzeActionsResult>() { response.Value.Result }, nextLink, rawResponse);
```
pallavit, line 304: >_firstPage [](http://example.com/codeflow?start=16&length=10)

Why do we set this as part of this function call? We seem to be accessing _firstPage as part of GetVaules however it is getting updated here which is a bit confusing.

sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/LegacyServiceClient.cs
```diff
@@ -1428,6 +1428,16 @@ public override Response<Models.AnalyzeTextJobStatusResult> AnalyzeStatusNextPag
             }
         }
 
+        public override Task CancelAnalyzeActionsJobAsync(string jobId, CancellationToken cancellationToken = default)
```
pallavit, line 1431: Is it only .NET that has a Legacy and the new client? Does Java follow the same pattern?

pallavit (approved):
:shipit:

sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/LegacyServiceClient.cs
heaths: Yes, Java follows the same pattern but I don't know if they've done this yet. Python does, though.

sdk/textanalytics/Azure.AI.TextAnalytics/src/AnalyzeActionsOperation.cs
heaths: `GetValues` uses `_firstPage` that gets set here because `UpdateStatus/Async` is designed to set everything - even the final value if the LRO is successful.

Issue #24955, opened 2021-10-27
[BUG]BlobUriBuilder strips trailing slash from blob name, making SDK 11's "directories" unreachable


Luaancz:
**Describe the bug**
Trying to access a SDK 11-style directory blob is impossible using e.g. `BlobContainerClient.GetBlobClient`. This goes all the way to `BlobUriBuilder`, which strips the trailing slash from the blob name, resulting in the wrong Uri. Bypassing `BlobUriBuilder` by constructing the Uri manually and using `new BlobClient(new Uri("..."), credentials)` works, but we lose the ability to use the container client configuration.

**Expected behavior**
The Uri for a "directory" blob should have a trailing slash.

**Actual behavior (include Exception or Stack Trace)**
The "directory" blob Uri doesn't have a trailing slash, resulting in a 404 from the service.

**To Reproduce**

1. Create a blob with name like "foo/bar/xyz"
2. Call `BlobContainerClient.GetBlobClient("foo/bar/")`. The client's Uri doesn't contain a trailing slash, and running e.g. `client.Exists()` returns false.

**Environment:**
 - Azure.Storage.Blobs 12.10.0
 - Windows 10, .NET Core 5.0.301
 - Visual Studio 16.10.3

jaschrep-msft:
Hi, @Luaancz. Apologies for such a late response.

Your reproduction steps confuse me. The way you use quotations for "directory" implies to me that you understand that blob containers don't truly have a directory structure, but are just flat listings with names that can be interpreted as such, leading to the concept of a "virtual directory:" a "directory" that exists in theory through names of other blobs, but for which no resource exists.

But then your reproduction steps indicate that you expect such a virtual directory to actually be created on the service; you seem to expect creating the blob `foo/bar/xyz` to create a directory `foo/bar` on the service. Clients in v12 purposefully removed the magic of blob "directories" where there were special objects to interpret them from the other blob resources. While you can still do hierarchical listings, we do not give you clients to obfuscate the actual state of the container.

If I've misunderstood your description, please let me know here.

Luaancz:
[...]

PR #30271, opened 2022-08-02
BlobUriBuilder preserve slashes

jaschrep-msft:
Addresses #24955
Addresses #30507

BlobUriBuilder strips leading and trailing slashes from blob names when constructing the Uri. Creating a client to a blob with a leading or trailing slash in blob name requires invoking the constructor directly (`BlobClient(Uri, Credential, BlobClientOptions`), which forces a developer to keep references to their credentials wherever they may need to create a blob client. While this and other modern Storage libraries make it very difficult to create a blob with a trailing slash in the name, it is still possible. Additionally, other libraries for interacting with Storage may be creating blobs with this behavior.  Clients should be capable of accessing these blobs in our recommended workflows.

Changes:
- BlobUriBuilder has an additional constructor which takes a flag on whether to preserve leading and trailing slashes on blob name.
- BlobClientOptions has a bool property `PreserveBlobNameSlashes` determining what flag to construct BlobUriBuilder with. Default is false.


sdk/storage/Azure.Storage.Blobs/src/BlobUriBuilder.cs
```diff
@@ -34,6 +34,19 @@ public class BlobUriBuilder
...
+        /// </summary>
+        public bool PreserveTrailingSlash { get; }
+
+        /// <summary>
+        /// Default value for <see cref="PreserveTrailingSlash"/> to construct an instance
+        /// of <see cref="BlobUriBuilder"/> with when unspecified. This is also the value
+        /// used by <see cref="BlobContainerClient.GetBlobClient(string)"/> and similar
+        /// methods. Defaults to false.
+        /// </summary>
+        public static bool PreserveTrailingSlashDefault { get; set; }
```
jaschrep-msft, line 48: @JoshLove-msft @tg-msft @kasobol-msft Looking for feedback regarding a static property. A static property is not the way we generally want to add these hooks. However, we do want a more centralized access point to change these defaults, as this class is used under the hood in many different places, and it seemed this behavior did not make sense as a "client option."

Do we have best practices for how to expose such a hook for a customer to change default behavior, instead of being forced to call new overloads for every `BlobContainerClient.GetBlobClient()` call in their entire code base? Is client options still the best answer?

jaschrep-msft:

sdk/storage/Azure.Storage.Blobs/src/BlobUriBuilder.cs
jaschrep-msft: Moved back to client options after testing concerns. Still happy to discuss whether this is really a "client option" but it is in clientoptions for now.

seanmcc-msft (approved):

PR #30282, opened 2022-08-02
Do not record fix for playback

JoshLove-msft:


sdk/core/Azure.Core.TestFramework/src/ProxyTransport.cs
```diff
@@ -66,6 +66,11 @@ public ProxyTransport(TestProxy proxy, HttpPipelineTransport transport, TestReco
 
         private async Task ProcessAsyncInternalAsync(HttpMessage message, bool async)
         {
+            if (_recording.Mode == RecordedTestMode.Playback && _filter() == EntryRecordModel.DoNotRecord)
```
JoshLove-msft, line 69: Actually, I'm not sure if this will help as I expect there will be no response on the message so clients will still throw. This check probably needs to be done upstream in the individual tests that need it.

JoshLove-msft: We could at least throw here to make the issue easier to diagnose. /cc @heaths

sdk/core/Azure.Core.TestFramework/src/ProxyTransport.cs
```diff
@@ -66,6 +66,11 @@ public ProxyTransport(TestProxy proxy, HttpPipelineTransport transport, TestReco
...
+            {
+                throw new InvalidOperationException("Operations that are enclosed in a Do Not Record scope should not be executed in Playback mode." +
```
heaths, line 71: "in a Do Not Record scope" reads a little funny. Maybe just replace "Do Not Record scope" with the actual scope name i.e., `TestRecording.DisableRecordingScope` e.g., "Operations that are enclosed in a TestRecording.DisableRecordingScope should not ...". Or could even mention the methods you'd call in case people are using `var`.

heaths (approved):
Just one nit, but otherwise looks great and will hopefully save some time for someone else down the road. 😆

Issue #26564, opened 2022-01-24
The migration document for ServiceBus lacks information on Dead Letter handling


Prinsn:
Very simply, the documentation (https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/servicebus/Azure.Messaging.ServiceBus/MigrationGuide.md) is incomplete for performing migrations

```cs
        public async Task ErrorHandlingWrapper(ServiceBusReceiver messageReceiver, string lockToken, Func<Task> func) {
                    var props = new Dictionary<string, object>()
                    {
                        { Message.DeadLetterReasonHeader, "An Exception has occurred" },
                        { Message.DeadLetterErrorDescriptionHeader, ex.Message }
                    };

                    if (ex.InnerException != null)
                    {
                        props.Add("Exeption With Inner Message", ex.Message + BuildInnerException(ex));
                    }

                    await messageReceiver.DeadLetterAsync(lockToken, props).ConfigureAwait(false);
         }
```

There's no indication of how to migrate away from this.  SeviceBusMessage does not contain the same constant values, and ServiceBusReceiver.DeadLetterMessageAsync has a completely different signature from MessageReceiver.DeadLetterAsync.

There are similar properties on ServiceBusReceivedMessage but they're read only and are inaccessible via the initializer pattern.

jsquire:
Thank you for your feedback.  Tagging and routing to the team member best able to assist.

JoshLove-msft:
The dead letter reason and description headers are not publicly exposed in Azure.Messaging.ServiceBus, but there is now a dedicated overload that allows specifying the reason and description. We can update the sample to demonstrate this.

PR #30299, opened 2022-08-03
Improve deadletter sample

JoshLove-msft:
Addresses https://github.com/Azure/azure-sdk-for-net/issues/26564


jsquire (approved):
Can we add something to the migration guide that details the strings that the legacy `Message.DeadLetterReasonHeader` and `Message.DeadLetterErrorDescriptionHeader` properties resolve to?

Issue #28866, opened 2022-05-18
Cache AMQP messages when adding to the AmqpMessageBatch


jsquire:
# Summary

Because an event batch is size-constrained and ensures that it only accepts events that will fit into the batch, when events are added to the batch using `TryAdd` they are first measured.   Because the size of an event depends on how it is represented for a network request, the transport batch is responsible for transforming the event to its wire format before measuring its size.

In the case of the `AmqpMessageBatch`, this means that an `ServiceBusMessage` instance is transformed into an `AmqpMessage`, which is a disposable resource.  In order to make efficient use of resources, the `AmqpMessageBatch` was intended to cache the `AmqpMessage` for each event that was accepted into the batch so that the transformation from `ServiceBusMessage` did not need to be repeated when a batch was sent.  

Because an AMQP message cannot be sent more than once, whether the send was successful or not, each time sending of a batch is requested, a new message was necessary.  Our initial attempts to clone the message for each send attempt was unsucessful; the AMQP transport viewed the new message as disposed.    This issue was temporarily worked around by caching the non-volatile `ServiceBusMessage` instances within the `AmqpMessageBatch` and perform the full translation to the corresponding AMQP format on send, in order to remove the possibility of premature disposal.  

Now that an approach for successfully cloning AMQP messages has been validated in Event Hubs, it should be ported to Service Bus.

# Scope of Work

- Implement the approach used by Event Hubs and restore `AmqpMessageBatch` functionality to caching the AMQP message rather than the `ServiceBusMessage` instance.

- Add or adjust tests to cover the new functionality.

# Success Criteria

- The implementation has been completed; `AmqpMessageBatch` performs serialization once and caches the `AmqpMessage` instance.

- The tests necessary for validation have been created or adjusted and pass reliably.

- The existing test suite continues to produce deterministic results and pass reliably.

# References

- [[Event Hubs] Event Batch Enhancements (#28055)](https://github.com/Azure/azure-sdk-for-net/pull/28055)
- [Revisit the ability to cache AMQP messages when adding events to the AmqpEventBatch (#9569)](https://github.com/Azure/azure-sdk-for-net/issues/9569)

jsquire:
Closed via #30305.

PR #30305, opened 2022-08-04
[Service Bus] Caching AMQP messages within AmqpMessageBatch

m-redding:
The main focus of the changes here is to implement the changes in `AmqpMessageBatch` in Service Bus that have already been made to the `EventDataBatch` in Event Hubs. Rather than serializing each `ServiceBusMessage` twice before send (once to measure size, and once to actually send), the improved `AmqpMessageBatch` simply serializes them and clones the `AmqpMessage` instance, allowing for a more efficient batching experience.

See #28866


sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageBatch.cs
```diff
@@ -156,12 +170,18 @@ public override void Clear()
         ///
         public override IReadOnlyCollection<T> AsReadOnly<T>()
         {
-            if (typeof(T) != typeof(ServiceBusMessage))
+            if (typeof(T) == typeof(AmqpMessage))
+            {
+                return (IReadOnlyCollection<T>)BatchMessages;
+            }
+            else if (typeof(T) == typeof(ServiceBusMessage))
```
m-redding, line 178: Because of `DiagnosticScope` and needing the first `ServiceBusMessage` to populate the batch envelope, I wasn't sure how to make these changes work without keeping the `ServiceBusMessage` references in the `AmqpMessageBatch`. I am open to other suggestions in this regard. 

I considered changing `DiagnosticScope` (but at first investigation it seems like it would be a pretty extensive change) and `AmqpMessageBatch` (but since there are so many properties held in the `ServiceBusMessage`, a lot would need to be propagated from the first message, unlike Event Hubs which just needs the partition key).

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
```diff
@@ -20,7 +20,7 @@
 
 namespace Azure.Messaging.ServiceBus.Amqp
 {
-    internal static class AmqpMessageConverter
+    internal class AmqpMessageConverter
```
m-redding, line 23: I needed to make this class non-static in order to mock it in test cases. If another approach is preferred here, I am open to suggestions.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
```diff
@@ -76,6 +76,22 @@ private static AmqpMessage BuildAmqpBatchFromMessage(IReadOnlyCollection<Service
...
+        ///   optionally propagating the custom properties.
+        /// </summary>
+        ///
+        /// <param name="source">The set of messages to use as the body of the batch message.</param>
+        /// <param name="firstMessage">The first message being sent in the batch.</param>
+        /// <param name="forceBatch">Set to true to force creating as a batch even when only one message.</param>
+        ///
+        /// <returns>The batch <see cref="AmqpMessage" /> containing the source messages.</returns>
+        ///
+        public virtual AmqpMessage BuildAmqpBatchFromMessages(IReadOnlyCollection<AmqpMessage> source, ServiceBusMessage firstMessage, bool forceBatch)
```
m-redding, line 90: As above, not sure how to work this without requiring the `ServiceBusMessage` first message.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpClient.cs
```diff
@@ -67,6 +67,11 @@ internal class AmqpClient : TransportClient
         ///
         private AmqpConnectionScope ConnectionScope { get; }
 
+        /// <summary>
+        ///    The converter to use for translating <see cref="ServiceBusMessage" /> into an AMQP-specific message.
+        /// </summary>
+        private readonly AmqpMessageConverter MessageConverter;
```
JoshLove-msft, line 73:
```suggestion
        private readonly AmqpMessageConverter _messageConverter;
```

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
jsquire: I think we'll need @JoshLove-msft's thought... but since we have a collection, couldn't we just grab the message at index 0?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
```diff
@@ -76,6 +76,22 @@ private static AmqpMessage BuildAmqpBatchFromMessage(IReadOnlyCollection<Service
...
+        {
+            return BuildAmqpBatchFromMessages(source.ToList<AmqpMessage>(), firstMessage, forceBatch);
```
jsquire, line 92: I don't think we need to force list here and eat the allocation.   Can we either update the parameter to use `IReadOnlyCollection<AmqpMessage>` here?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageBatch.cs
```diff
@@ -59,25 +59,41 @@ internal class AmqpMessageBatch : TransportMessageBatch
...
+        private readonly AmqpMessageConverter MessageConverter;
+
         /// <summary>
         ///   The set of messages that have been added to the batch.
         /// </summary>
         ///
-        private List<ServiceBusMessage> BatchMessages { get; } = new List<ServiceBusMessage>();
+        private List<AmqpMessage> BatchMessages { get; } = new List<AmqpMessage>();
+
+        /// <summary>
```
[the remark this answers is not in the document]
jsquire: Based on Krzysztof's potential "back-compat" idea.... we may actually need to keep this.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Core/TransportMessageBatch.cs
```diff
@@ -53,8 +53,7 @@ internal abstract class TransportMessageBatch : IDisposable
         public abstract void Clear();
 
         /// <summary>
-        ///   Represents the batch as an enumerable set of transport-specific
-        ///   representations of a message.
+        ///   Represents the batch as a set of the AMQP-specific representations of an event.
```
jsquire, line 56:
```suggestion
        ///   Represents the batch as a set of the AMQP-specific representations of a message.
```

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
m-redding: I wasn't sure how to get the right properties off the Amqp message, maybe convert it back to a ServiceBusMessage? Initially, I just put all the converted messages in an AMQP batch and didn't populate the batch envelope properties (see L133) because I couldn't get them from the AmqpMessage but I was getting an error (like [this](https://github.com/Azure/azure-amqp/issues/118#issuecomment-418546828)) that I fixed by adding the messageId back in using the first message.

Right now we just have AmqpMessage -> ReceivedServiceBusMessage, just add an AmqpMessage -> ServiceBusMessage?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
```diff
@@ -53,7 +53,7 @@ public static AmqpMessage BatchSBMessagesAsAmqpMessage(IReadOnlyCollection<Servi
         ///
         /// <returns>The batch <see cref="AmqpMessage" /> containing the source messages.</returns>
         ///
-        private static AmqpMessage BuildAmqpBatchFromMessage(IReadOnlyCollection<ServiceBusMessage> source, bool forceBatch)
+        public virtual AmqpMessage BuildAmqpBatchFromMessage(IReadOnlyCollection<ServiceBusMessage> source, bool forceBatch)
```
jsquire, line 56: I don't think we need an explicit `firstMessage` reference - we don't seem to use it.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
```diff
@@ -73,62 +73,80 @@ private static AmqpMessage BuildAmqpBatchFromMessage(IReadOnlyCollection<Service
...
         /// </summary>
         ///
         /// <param name="batchMessages">The set of messages to use as the body of the batch message.</param>
-        /// <param name="firstMessage">The first message being sent in the batch.</param>
         /// <param name="forceBatch">Set to true to force creating as a batch even when only one message.</param>
         ///
         /// <returns>The batch <see cref="AmqpMessage" /> containing the source messages.</returns>
         ///
         private static AmqpMessage BuildAmqpBatchFromMessages(
             List<AmqpMessage> batchMessages,
```
jsquire, line 104: Can we take `IReadOnlyCollection<AmqpMessage>` here?   That would allow us to remove the overload and avoid allocating a new list copy.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/ServiceBusModelFactory.cs
```diff
@@ -408,6 +410,11 @@ public static class ServiceBusModelFactory
         ///
         /// <returns>The <see cref="Azure.Messaging.ServiceBus.ServiceBusMessageBatch" /> instance that was created.</returns>
         ///
+        /// <remarks>
+        ///  The batch instance keeps an internal copy of events successfully added to the batch through <see cref="ServiceBusMessageBatch.TryAddMessage(ServiceBusMessage)"/>, meaning that any
+        ///  changes made to <paramref name="batchMessageStore"/> after adding the messages to the batch will not be reflected.
+        /// </remarks>
+        ///
         public static ServiceBusMessageBatch ServiceBusMessageBatch(long batchSizeBytes,
```
jsquire, line 418: nit:  Indentation is off

sdk/servicebus/Azure.Messaging.ServiceBus/src/Sender/ServiceBusMessageBatch.cs
```diff
@@ -51,6 +51,27 @@ public sealed class ServiceBusMessageBatch : IDisposable
...
+        /// </summary>
+        ///
+        private readonly string _fullyQualifiedNamespace;
+
+        /// <summary>
+        ///   The list of diagnostic identifiers of messages added to this batch.  To be used during
+        ///   instrumentation.
+        /// </summary>
+        ///
+        private List<string> _messageDiagnosticIdentifiers { get; } = new List<string>();
```
jsquire, line 73: Should this be `readonly`?

jsquire (commented):
LGTM; I'll defer to Josh for the authoritative approval.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageBatch.cs
```diff
@@ -59,25 +59,35 @@ internal class AmqpMessageBatch : TransportMessageBatch
...
+        ///    The converter to use for translating <see cref="ServiceBusMessage" /> into an AMQP-specific message.
+        /// </summary>
+        private readonly AmqpMessageConverter MessageConverter;
+
         /// <summary>
         ///   The set of messages that have been added to the batch.
         /// </summary>
         ///
-        private List<ServiceBusMessage> BatchMessages { get; } = new List<ServiceBusMessage>();
+        private List<AmqpMessage> BatchMessages { get; } = new List<AmqpMessage>();
```
JoshLove-msft, line 71: nit: _batchMessages

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpReceiver.cs
```diff
@@ -104,6 +104,11 @@ internal class AmqpReceiver : TransportReceiver
 
         private Exception LinkException { get; set; }
 
+        /// <summary>
+        ///    The converter to use for translating <see cref="ServiceBusMessage" /> into an AMQP-specific message.
+        /// </summary>
+        private readonly AmqpMessageConverter MessageConverter;
```
JoshLove-msft, line 110: nit: _messageConverter

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
```diff
@@ -551,7 +548,7 @@ public static ServiceBusReceivedMessage AmqpMessageToSBMessage(AmqpMessage amqpM
             return sbMessage;
         }
 
-        public static AmqpMap GetRuleDescriptionMap(RuleProperties description)
+        public virtual AmqpMap GetRuleDescriptionMap(RuleProperties description)
```
JoshLove-msft, line 551: Does this one need to be made virtual?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpSender.cs
```diff
@@ -184,7 +192,7 @@ internal class AmqpSender : TransportSender
             options.MaxSizeInBytes ??= MaxMessageSize;
 
             Argument.AssertInRange(options.MaxSizeInBytes.Value, ServiceBusSender.MinimumBatchSizeLimit, MaxMessageSize.Value, nameof(options.MaxSizeInBytes));
-            return new AmqpMessageBatch(options);
+            return new AmqpMessageBatch(new AmqpMessageConverter(), options);
```
JoshLove-msft, line 195: Should we consider having a static property on AmqpMessageConverter that gives us an AmqpMessageConverter instance? I think we are only making the class non-static for testing, but there isn't actually any state that needs to be maintained in the instance.
/cc @jsquire

sdk/servicebus/Azure.Messaging.ServiceBus/src/Sender/ServiceBusSender.cs
```diff
@@ -463,8 +471,16 @@ private DiagnosticScope CreateDiagnosticScope(IReadOnlyCollection<ServiceBusMess
                 readOnlyCollection.Count,
                 scheduledEnqueueTime.ToString(CultureInfo.InvariantCulture));
 
+            var diagnosticIdentifiers = new List<string>();
+
+            foreach (var message in messages)
+            {
+                var identifier = EntityScopeFactory.InstrumentMessage(message, EntityPath, FullyQualifiedNamespace);
+                diagnosticIdentifiers.Add(identifier);
```
JoshLove-msft, line 479: I'm curious why we need this change?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Sender/ServiceBusMessageBatch.cs
```diff
@@ -147,6 +177,14 @@ internal void Clear()
         ///
         internal IReadOnlyCollection<T> AsReadOnly<T>() => _innerBatch.AsReadOnly<T>();
 
+        /// <summary>
+        ///   Gets the list of diagnostic identifiers of messages added to this batch.
+        /// </summary>
+        ///
+        /// <returns>A read-only list of diagnostic identifiers.</returns>
+        ///
+        internal IReadOnlyList<string> GetMessageDiagnosticIdentifiers() => _messageDiagnosticIdentifiers;
```
JoshLove-msft, line 186: Why is this needed?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Sender/ServiceBusMessageBatch.cs
```diff
@@ -105,8 +128,15 @@ public bool TryAddMessage(ServiceBusMessage message)
             {
                 AssertNotLocked();
 
-                _scopeFactory.InstrumentMessage(message);
-                return _innerBatch.TryAddMessage(message);
+                var identifier = EntityScopeFactory.InstrumentMessage(message, _entityPath, _fullyQualifiedNamespace);
```
JoshLove-msft, line 131: Same question here about why we need to do the business with storing off the diagnostic identifiers.

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Amqp/AmqpSenderTests.cs
```diff
@@ -0,0 +1,145 @@
...
+using NUnit.Framework;
+
+namespace Azure.Messaging.ServiceBus.Tests.Amqp
+{
+    /// <summary>
+    ///   The suite of tests for the <see cref="AmqpSender" />
+    ///   class.
+    /// </summary>
+    ///
+    [TestFixture]
```
JoshLove-msft, line 26: nit: I think this isn't needed.

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Amqp/AmqpSenderTests.cs
```diff
@@ -0,0 +1,145 @@
...
+    public class AmqpSenderTests
+    {
+        /// <summary>
+        ///   Verifies functionality of the <see cref="AmqpProducer.CreateBatchAsync" />
```
JoshLove-msft, line 30:
```suggestion
        ///   Verifies functionality of the <see cref="AmqpSender.CreateBatchAsync" />
```

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Primitives/ServiceBusModelFactoryTests.cs
```diff
@@ -0,0 +1,66 @@
...
+
+namespace Azure.Messaging.ServiceBus.Tests.Amqp
+{
+    /// <summary>
+    ///   The suite of tests for the <see cref="ServiceBusModelFactory" />
+    ///   class.
+    /// </summary>
+    ///
+    [TestFixture]
+    public class ServiceBusModelFactoryTests
```
JoshLove-msft, line 18: nit: add these to MessageBatchTests

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Receiver/ReceiverLiveTests.cs
```diff
@@ -280,7 +282,7 @@ await using (var scope = await ServiceBusScope.CreateWithQueue(enablePartitionin
 
                 ServiceBusSender sender = client.CreateSender(scope.QueueName);
                 using ServiceBusMessageBatch batch = await sender.CreateMessageBatchAsync();
-                IEnumerable<ServiceBusMessage> messages = ServiceBusTestUtilities.AddMessages(batch, messageCount).AsReadOnly<ServiceBusMessage>();
+                IEnumerable<AmqpMessage> messages = ServiceBusTestUtilities.AddMessages(batch, messageCount).AsReadOnly<AmqpMessage>();
```
JoshLove-msft, line 285: Hmm, why do we want to operate on AmqpMessage for this test?

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Receiver/SessionReceiverLiveTests.cs
```diff
@@ -218,7 +218,7 @@ await using (var scope = await ServiceBusScope.CreateWithQueue(enablePartitionin
                 var sessionId = "sessionId1";
                 ServiceBusSender sender = client.CreateSender(scope.QueueName);
                 using ServiceBusMessageBatch batch = await sender.CreateMessageBatchAsync();
-                IEnumerable<ServiceBusMessage> messages = ServiceBusTestUtilities.AddMessages(batch, messageCount, sessionId).AsReadOnly<ServiceBusMessage>();
+                IEnumerable<AmqpMessage> messages = ServiceBusTestUtilities.AddMessages(batch, messageCount, sessionId).AsReadOnly<AmqpMessage>();
```
JoshLove-msft, line 221: I would just leave these as using ServiceBusMessage. I think we can cover this scenario with a single test in ReceiverLiveTests.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/ServiceBusModelFactory.cs
```diff
@@ -448,9 +455,15 @@ private sealed class ListTransportBatch : TransportMessageBatch
             /// <summary>The backing store for storing events in the batch.</summary>
             private readonly IList<ServiceBusMessage> _backingStore;
 
+            /// <summary>The set of events that have been added to the batch, in their <see cref="AmqpMessage" /> serialized format.</summary>
```
JoshLove-msft, line 458:
```suggestion
            /// <summary>The set of messages that have been added to the batch, in their <see cref="AmqpMessage" /> serialized format.</summary>
```

sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/ServiceBusModelFactory.cs
```diff
@@ -448,9 +455,15 @@ private sealed class ListTransportBatch : TransportMessageBatch
...
+            private List<AmqpMessage> _batchMessages;
+
             /// <summary>A callback to be invoked when an adding an event via <see cref="TryAddMessage"/></summary>
```
JoshLove-msft, line 461:
```suggestion
            /// <summary>A callback to be invoked when an adding a message via <see cref="TryAddMessage"/></summary>
```

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Diagnostics/EventSourceLiveTests.cs
```diff
@@ -48,7 +49,7 @@ await using (var scope = await ServiceBusScope.CreateWithQueue(enablePartitionin
                 _listener.SingleEventById(ServiceBusEventSource.CreateMessageBatchStartEvent, e => e.Payload.Contains(sender.Identifier));
                 _listener.SingleEventById(ServiceBusEventSource.CreateMessageBatchCompleteEvent, e => e.Payload.Contains(sender.Identifier));
 
-                IEnumerable<ServiceBusMessage> messages = ServiceBusTestUtilities.AddMessages(batch, messageCount).AsReadOnly<ServiceBusMessage>();
+                IEnumerable<AmqpMessage> messages = ServiceBusTestUtilities.AddMessages(batch, messageCount).AsReadOnly<AmqpMessage>();
```
JoshLove-msft, line 52: Suggest leaving this as is and just having a single test for this scenario.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Diagnostics/EntityScopeFactory.cs
```diff
@@ -17,10 +18,14 @@ internal class EntityScopeFactory
         /// <summary>The namespace used for the Service Bus diagnostic scope.</summary>
         public const string DiagnosticNamespace = "Azure.Messaging.ServiceBus";
 
+        /// <summary> The path to the entity used for the Service Bus instance.</summary>
+        public readonly string EntityPath;
```
JoshLove-msft, line 22: Why does this need to be public?

JoshLove-msft (commented):
Looks good! I left a few comments. The main thing I'm unsure about is why we need to refactor all of the diagnostic scope stuff.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpSender.cs
jsquire: Depends on the usage.  I believe we still intentionally pass an instance of the converter around in the infrastructure in a few places to enable testing.  If those places start to refer to the static property rather than accepting an instance via constructor injection, we're likely to have some test issues and, ultimately, a less testable surface.

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Amqp/AmqpSenderTests.cs
jsquire: We're not deriving from any base that would inherit it.   I think it should be here.

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Receiver/ReceiverLiveTests.cs
jsquire: The batch no longer stores `ServiceBusMessage` instances, only `AmqpMessage`.   Unless you're using a special mocking batch, there are no models to return.

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Receiver/SessionReceiverLiveTests.cs
jsquire: Unless you've got a mocked batch with different behavior, only `AmqpMessage` is being stored by the batch.

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Diagnostics/EventSourceLiveTests.cs
jsquire: Same here.   Not holding onto `ServiceBusMessage` instances in the batch is kind of the entire point of this change.  😄

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Amqp/AmqpSenderTests.cs
JoshLove-msft: It should only be necessary if a parameter is used - https://docs.nunit.org/articles/nunit/writing-tests/attributes/testfixture.html:
`Beginning with NUnit 2.5, the TestFixture attribute is optional for non-parameterized, non-generic fixtures. So long as the class contains at least one method marked with the Test, TestCase or TestCaseSource attribute, it will be treated as a test fixture.`

We are on version 3.13 - https://github.com/Azure/azure-sdk-for-net/blob/main/eng/Packages.Data.props#L247

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Receiver/ReceiverLiveTests.cs
JoshLove-msft: Ah I see. Rather than leaking AmqpMessage throughout all the tests, I think we should just use the SendMessagesAsync overload that takes ServiceBusMessage. Alternatively, we could add a utility method that creates a batch from a set of ServiceBusMessages.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpSender.cs
JoshLove-msft: It's fine to still pass the instance in everywhere, but it would be more efficient if we could use a single static instance.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Sender/ServiceBusMessageBatch.cs
m-redding: I have a getter that returns a read only list defined [here](https://github.com/Azure/azure-sdk-for-net/blob/9e1632fcd59960b1dcbe90ad1079e14fdf609465/sdk/servicebus/Azure.Messaging.ServiceBus/src/Sender/ServiceBusMessageBatch.cs#L186). Would it preferable to declare it `readonly` instead?

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Diagnostics/EventSourceLiveTests.cs
m-redding: I'm not sure how to leave this as is. The batch is only able to do AsReadOnly<AmqpMessage> with the new changes. Should I have written a way for the batch to be able to return SBmessages?

JoshLove-msft: Good point 😄 I would say we either add a test method that returns back the SBMessages or we just use the regular SendMessagesAsync overload in all of these tests.

JoshLove-msft: I think we can use this [method](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/servicebus/Azure.Messaging.ServiceBus/tests/Infrastructure/ServiceBusTestUtilities.cs#L19) for any test that isn't really testing batches.

JoshLove-msft: So, in this case the test is explicitly testing batches, so we'd need to either add a new utility message that returns the SBMessages or just leave what you have. I'm fine with leaving as is for the tests that are specifically testing batches. I was just opposed to changing every test to use `AmqpMessage`.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Sender/ServiceBusMessageBatch.cs
JoshLove-msft: I understand now - we need to output the identifiers so that we can link them to the send activity. We used to rely on the SBMessages being cached in the batch, but this is no longer available. 

We have two options:
1) Do what you have here where we store off the identifiers
2) Update to pull the identifiers off of the AmqpMessage.UserProperties.

I'm fine with either approach.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
JoshLove-msft: Does this need to be public?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
```diff
@@ -53,82 +52,80 @@ public static AmqpMessage BatchSBMessagesAsAmqpMessage(IReadOnlyCollection<Servi
...
+                var data = new List<Data>(source.Count);
+                var messageNum = 0;
+
+                foreach (var message in source)
                 {
                     message.Batchable = true;
                     using var messageStream = message.ToStream();
                     data.Add(new Data { Value = ReadStreamToArraySegment(messageStream) });
+
+                    ++messageNum;
```
JoshLove-msft, line 104: Why do we need this?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
```diff
@@ -580,7 +577,7 @@ public static AmqpMap GetRuleDescriptionMap(RuleProperties description)
             return ruleDescriptionMap;
         }
 
-        public static RuleProperties GetRuleDescription(AmqpRuleDescriptionCodec amqpDescription)
+        public virtual RuleProperties GetRuleDescription(AmqpRuleDescriptionCodec amqpDescription)
```
JoshLove-msft, line 580: Can we move this and GetFilter into AmqpRuleManager? I think it didn't belong in this class.

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Receiver/ReceiverLiveTests.cs
```diff
@@ -619,7 +621,7 @@ await using (var scope = await ServiceBusScope.CreateWithQueue(enablePartitionin
                 Assert.AreEqual(messageList.Count, deferredMessages.Count);
                 for (int i = 0; i < messageList.Count; i++)
                 {
-                    Assert.AreEqual(messageList[i].MessageId, deferredMessages[i].MessageId);
+                    Assert.AreEqual(messageList[i].MessageId.ToString(), deferredMessages[i].MessageId);
```
JoshLove-msft, line 624: nit: ToString no longer needed

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Receiver/ReceiverLiveTests.cs
```diff
@@ -724,7 +726,7 @@ await using (var scope = await ServiceBusScope.CreateWithQueue(enablePartitionin
 
                 ServiceBusSender sender = client.CreateSender(scope.QueueName);
                 using ServiceBusMessageBatch batch = await sender.CreateMessageBatchAsync();
-                IEnumerable<ServiceBusMessage> messages = ServiceBusTestUtilities.AddMessages(batch, messageCount).AsReadOnly<ServiceBusMessage>();
+                IEnumerable<AmqpMessage> messages = ServiceBusTestUtilities.AddMessages(batch, messageCount).AsReadOnly<AmqpMessage>();
```
JoshLove-msft, line 729: nit: update to use new utility method

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Receiver/SessionReceiverLiveTests.cs
JoshLove-msft: nit: update to use new util method

JoshLove-msft (commented):
Looks good! Just a few remaining comments. Also, a couple tests failing in CI.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpSender.cs
JoshLove-msft: Seems like it would be an easy perf win if we could hook everything up.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Sender/ServiceBusSender.cs
```diff
@@ -268,6 +269,24 @@ private DiagnosticScope CreateDiagnosticScope(IReadOnlyCollection<ServiceBusMess
...
             scope.SetMessageData(messages);
+
+            return scope;
+        }
+
+        private DiagnosticScope CreateDiagnosticScope(IReadOnlyCollection<AmqpMessage> messages, string activityName)
+        {
+            foreach (AmqpMessage message in messages)
+            {
+                _scopeFactory.InstrumentMessage(message);
```
JoshLove-msft, line 280: I think these would already be instrumented when they are added to the batch.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Sender/ServiceBusSender.cs
```diff
@@ -268,6 +269,24 @@ private DiagnosticScope CreateDiagnosticScope(IReadOnlyCollection<ServiceBusMess
                 DiagnosticScope.ActivityKind.Client);
 
             scope.SetMessageData(messages);
+
+            return scope;
+        }
+
+        private DiagnosticScope CreateDiagnosticScope(IReadOnlyCollection<AmqpMessage> messages, string activityName)
```
JoshLove-msft, line 276: nit: suggest taking in the ServiceBusMessageBatch here to help make it easier to grok why this method is needed as compared to the SBMessage one.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
m-redding: Accidentally left over from a previous version.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
```diff
@@ -53,54 +52,49 @@ public static AmqpMessage BatchSBMessagesAsAmqpMessage(IReadOnlyCollection<Servi
...
-        /// <param name="firstMessage">The first message being sent in the batch.</param>
+        /// <param name="source">The set of messages to use as the body of the batch message.</param>
         /// <param name="forceBatch">Set to true to force creating as a batch even when only one message.</param>
         ///
         /// <returns>The batch <see cref="AmqpMessage" /> containing the source messages.</returns>
         ///
-        private static AmqpMessage BuildAmqpBatchFromMessages(
-            List<AmqpMessage> batchMessages,
-            ServiceBusMessage firstMessage,
+        public virtual AmqpMessage BuildAmqpBatchFromMessages(
```
JoshLove-msft, line 74: Does this need to be public? Can we instead call into the wrapper?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
```diff
@@ -551,119 +545,6 @@ public static ServiceBusReceivedMessage AmqpMessageToSBMessage(AmqpMessage amqpM
             return sbMessage;
         }
 
-        public static AmqpMap GetRuleDescriptionMap(RuleProperties description)
```
JoshLove-msft, line 554: Thank you!!

sdk/servicebus/Azure.Messaging.ServiceBus/src/Diagnostics/EntityScopeFactory.cs
```diff
@@ -80,6 +82,28 @@ public static bool TryExtractDiagnosticId(IDictionary<string, object> properties
...
+        /// <param name="properties">The properties holding the diagnostic id.</param>
+        /// <param name="id">The value of the diagnostics identifier assigned to the event. </param>
+        ///
+        /// <returns><c>true</c> if the event was contained the diagnostic id; otherwise, <c>false</c>.</returns>
+        ///
+        public static bool TryExtractDiagnosticId(PropertiesMap properties, out string id)
+        {
+            id = null;
+
+            if (properties.TryGetValue<string>(DiagnosticProperty.DiagnosticIdAttribute, out var objectId) && objectId is string stringId)
```
JoshLove-msft, line 98: It looks like TryGetValue will handle the type cast to string, so I think we can just do
```suggestion
            if (properties.TryGetValue<string>(DiagnosticProperty.DiagnosticIdAttribute, out string stringId))
```

sdk/servicebus/Azure.Messaging.ServiceBus/src/Sender/ServiceBusSender.cs
```diff
@@ -268,6 +269,23 @@ private DiagnosticScope CreateDiagnosticScope(IReadOnlyCollection<ServiceBusMess
...
 
             scope.SetMessageData(messages);
+
+            return scope;
+        }
+
+        private DiagnosticScope CreateDiagnosticScope(ServiceBusMessageBatch messageBatch, string activityName)
+        {
+            // Messages in a batch have already been instrumented when
+            // they are added to the batch.
```
JoshLove-msft, line 279:
```suggestion
            // they are added to the batch so we don't need to instrument them here.
```

JoshLove-msft (commented):
Looks great! Just a few remaining comments.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
JoshLove-msft: Same question here about whether this needs to be public.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
m-redding: The reason I made this public was just because I needed a way to pass the AMQP messages around instead of the Service Bus Messages. I think it could work to add a wrapper that takes in a `ServiceBusMessageBatch` and then make this method internal, would that be better?

m-redding: But otherwise, the only other wrappers just take 1+ `ServiceBusMessage`'s

JoshLove-msft: Ah I see. I guess we do need this one to be public. But what about the other one that takes a ServiceBusMessage - https://github.com/Azure/azure-sdk-for-net/pull/30305/files/16bad73f62f1503c888b6a068d214c316a0f8198#diff-a71c7b0a7a2b2fe832ac51ebab6b7aff0f760f124d542984cde6f7b8e566b9f8R55

JoshLove-msft (approved):

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
m-redding: No I don't think this does, switched it to private.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
m-redding: That one can be private. It just can't be static

JoshLove-msft: That's fine if it is not static.

PR #30307, opened 2022-08-04
Excluded tests that should not be included in virtualized test pipeline

seanmcc-msft:


jaschrep-msft (approved):

benbp (approved):

eng/pipelines/templates/jobs/ci.tests.yml
```diff
@@ -63,11 +63,11 @@ jobs:
       - template: /eng/pipelines/templates/steps/install-dotnet.yml
         parameters:
           NuGetCacheKey: Test
-      - script: >-
+      - pwsh: >-
```
weshaggard, line 66: @benbp @seanmcc-msft is there a reason to change this from script to pwsh? I mostly ask because that change removes the echo'ed logging of the command to the logs. For whatever reason script echos and pwsh does not. If we don't need pwsh we should revert this back to script.

benbp: IIRC there may have been some quote handling behavior specifics with pwsh vs. bash/cmd. But I agree it's not good if the test command isn't printed, because you can't try to run it yourself. I can put up a quick test.

benbp: @weshaggard I'm not seeing the echoed logging of the command in older versions with the `script` task?

weshaggard: I think there is some weirdness between windows and linux. I believe it does show up on windows machines.

weshaggard: Actually looking an older build I'm seeing it show up on both windows and linux https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1746701&view=logs&j=90d6a2f0-c5b3-54db-32f0-18fb6ac025ff&t=91147b3f-f662-545c-bcb2-93244f07606f

benbp: I was looking at the old storage builds, which also don't have it even with `script`: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1741972&view=logs&j=2c475c7c-5397-5d99-fddd-441d521d1b89&t=9d12005b-1138-5c80-7fde-34e6ad0b25b1

Odd...

benbp: I also tried reverting it and running the virtualized pipeline, and it's still missing: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1876176&view=logs&j=8a9def8f-4e47-5e01-67bc-38cc3c35450b&t=d003dd4f-4ebf-57ce-4847-66e3b44e9693

weshaggard: @benbp those are the live tests pipelines which is still using pwsh and we should also look to change that to script https://github.com/Azure/azure-sdk-for-net/blob/main/eng/pipelines/templates/jobs/live.tests.yml#L123.

PR #30312, opened 2022-08-04
[EventHubs]: Adding Testcases to EventHub and ServiceBus Track2 SDK's

damodaravadhani:


ArthurMa1978:
@damodaravadhani  can you fix the test failures?

ArthurMa1978 (approved):

sdk/eventhub/Azure.ResourceManager.EventHubs/tests/Azure.ResourceManager.EventHubs.Tests.csproj
```diff
@@ -1,9 +1,11 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
   <ItemGroup>
     <ProjectReference Include="..\src\Azure.ResourceManager.EventHubs.csproj" />
     <PackageReference Include="Azure.ResourceManager.Network" />
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="Azure.Identity" />
+    <PackageReference Include="Azure.ResourceManager.KeyVault" VersionOverride="1.0.0" />
```
weshaggard, line 8: Why do you need to override these versions?

damodaravadhani: Because the default version is a different one (I do not exactly remember the file that has default RP SDK versions). But I faced issues with it and hence had to use this version.

ArthurMa1978: Only one stable version for KeyVault there, not sure which is the default version before, or it's project reference?

damodaravadhani:
Thanks for the merge @ArthurMa1978

sdk/eventhub/Azure.ResourceManager.EventHubs/tests/Azure.ResourceManager.EventHubs.Tests.csproj
weshaggard: We should track getting these updated such that we don't override the version and if you need a newer version, it should be updated in our shared Packages.data.props file.

m-nash: @ArthurMa1978 we definitely shouldn't have package version overrides in any of our csproj files.  I have added this issue to track turning on the validation for this in test projects.  It is already on for our main projects. https://github.com/Azure/azure-sdk-for-net/issues/32033

Issue #30105, opened 2022-07-25
[Feature Request] Add overload of AddAzureClientsCore that allows enabling logging


scottaddie:
### Library name and version

Azure.Core 1.25.0

### Describe the bug

Azure.Core logs aren't printed to the console. I'm either missing a step, or this is a bug.

### Expected behavior

Azure.Core logs should be printed to the console. I expected to see the HTTP request and response bodies there. Instead, I saw no logs.

### Actual behavior

Azure.Core logs aren't printed to the console.

### Reproduction Steps

1. Create a new ASP.NET Core 6.0 Razor Pages app via the project template in VS 2022.
1. Add the following code to *Program.cs*:

    ```csharp
    builder.Services.AddAzureClientsCore();
    ```
1. In the `OnGet` method of *Index.cshtml.cs*, add Azure SDK library client code that initiates an HTTP request. Ensure the options object's `Diagnostics.IsLoggingContentEnabled` property is set to `true`. For example, using v1.1.0 of the Monitor Query library:

    ```csharp
                var client = new LogsQueryClient(
                    new DefaultAzureCredential(),
                    new LogsQueryClientOptions
                    {
                        Diagnostics = { IsLoggingContentEnabled = true }
                    });

                var startDate = DateTime.Parse("2022-07-12 00:00:00");
                var endDate = DateTime.Parse("2022-07-14 23:59:59");
                var metricType = "HeartbeatState";
                var metricNamespace = "netstandard2.0";

                var query = $"AppMetrics | where Name == '{metricType}' and Properties has '{metricNamespace}'";
                Response<LogsQueryResult> response = await client.QueryWorkspaceAsync(
                    "<log_analytics_workspace_id>", query, new QueryTimeRange(startDate, endDate));

                LogsTable table = response.Value.Table;
                Console.WriteLine(table.Rows.Count);
    ```

1. In *appsettings.Development.json*, add the following inside of `Logging:LogLevel`:

    ```json
    "Azure.Core": "Debug",
    ```

1. Run the app using the default launch profile in VS 2022.

### Environment

**VS info:**

Microsoft Visual Studio Enterprise 2022
Version 17.3.0 Preview 4.0
VisualStudio.17.Preview/17.3.0-pre.4.0+32714.290
Microsoft .NET Framework
Version 4.8.04161

Installed Version: Enterprise

**dotnet --info output:**

.NET SDK (reflecting any global.json):ddie/app-insights-query]> do
[...]

PR #30317, opened 2022-08-04
Add AzureClientsCore overload with enableLogging parameter

JoshLove-msft:
Addresses https://github.com/Azure/azure-sdk-for-net/issues/30105

### DIFF
--- a/sdk/extensions/Microsoft.Extensions.Azure/src/AzureClientServiceCollectionExtensions.cs
+++ b/sdk/extensions/Microsoft.Extensions.Azure/src/AzureClientServiceCollectionExtensions.cs
@@ -29,9 +29,9 @@ namespace Microsoft.Extensions.Azure
 
         /// <summary>
         /// Adds the minimum essential Azure SDK interop services like <see cref="AzureEventSourceLogForwarder"/> and <see cref="AzureComponentFactory"/> to the specified <see cref="IServiceCollection"/> without registering any client types.
-        /// Azure SDK logging will not be enabled by default, but can be enabled by calling the <see cref="AzureEventSourceLogForwarder.Start"/> method.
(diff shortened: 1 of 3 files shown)


sdk/extensions/Microsoft.Extensions.Azure/src/AzureClientServiceCollectionExtensions.cs
```diff
@@ -28,12 +29,36 @@ public static void AddAzureClients(this IServiceCollection collection, Action<Az
...
+            collection.AddAzureClientsCore(false);
+        }
+
+        /// <summary>
+        /// Adds the minimum essential Azure SDK interop services like <see cref="AzureEventSourceLogForwarder"/> and <see cref="AzureComponentFactory"/> to the specified <see cref="IServiceCollection"/> without registering any client types.
+        /// </summary>
+        /// <param name="collection">The <see cref="IServiceCollection"/>.</param>
+        /// <param name="enableLogging">Whether or not to enable Azure SDK logging. Even if this is set to <value>false</value>, logging can be enabled
+        /// by calling the <see cref="AzureEventSourceLogForwarder.Start"/> method.</param>
+        public static void AddAzureClientsCore(this IServiceCollection collection, bool enableLogging)
```
jsquire, line 48: Should we consider `startLogging`?   Isn't logging "enabled" either way - this is just about whether we automatically start the forwarder or require the someone do it manually, no?

JoshLove-msft: The logging would only be enabled if someone manually starts the log forwarder. Otherwise they will not get any logs.

JoshLove-msft: What about enableLogForwarding?

jsquire: oohh... I like that!

jsquire (approved):

PR #30322, opened 2022-08-04
Make sure all TestProjects* are deleted for QA

heaths:
Missed a couple cases in PR #30276


kristapratico (approved):

heaths:
One of the tasks is hanging, but this PR is test-only, timely, and shouldn't adversely impact that task or even the job containing it.

PR #30327, opened 2022-08-04
Update TA autorest.md to correct swaggers

heaths:
The reference commits were incorrect.


kristapratico (approved):

heaths:
Will override check-enforcer because the SBOM task is only necessary when publishing packages, which we're not doing with this package (more to come).

/check-enforcer override

PR #30365, opened 2022-08-05
[Media Composition] Introducing polymorphic models

minnieliu:


sdk/communication/Azure.Communication.MediaComposition/src/MediaCompositionClient.cs
```diff
@@ -130,13 +130,13 @@ public virtual Response<MediaCompositionBody> Get(string mediaCompositionId, Can
...
         /// <param name="cancellationToken"> The cancellation token to use. </param>
-        public virtual async Task<Response<MediaCompositionBody>> CreateAsync(string mediaCompositionId, MediaCompositionLayout layout, IDictionary<string, MediaInput> inputs, IDictionary<string, MediaOutput> outputs, CancellationToken cancellationToken = default)
+        public virtual async Task<Response<MediaComposition>> CreateAsync(string mediaCompositionId, MediaCompositionLayout layout, IDictionary<string, MediaInput> inputs, IDictionary<string, MediaOutput> outputs, CancellationToken cancellationToken = default)
         {
             using var scope = _clientDiagnostics.CreateScope("MediaCompositionClient.Create");
             scope.Start();
             try
             {
-                return await RestClient.CreateAsync(mediaCompositionId, mediaCompositionId, layout, inputs, outputs, CompositionStreamState.NotStarted, cancellationToken).ConfigureAwait(false);
+                return await RestClient.CreateAsync(mediaCompositionId, mediaCompositionId, layout, inputs, outputs, new CompositionStreamState(StreamStatus.NotStarted), cancellationToken).ConfigureAwait(false);
```
DominikMe, line 139: Does our service allow to set the StreamStatus? I think we should make that property read-only for clients and only settable via the :start, :stop POST actions

sdk/communication/Azure.Communication.MediaComposition/src/autorest.md
```diff
@@ -7,11 +7,19 @@ Run `dotnet msbuild /t:GenerateCode` to generate code.
...
-tag: package-2022-07-16-preview1
-generation1-convenience-client: true
+tag: package-2022-07-16-preview
 model-namespace: false
+generation1-convenience-client: true
 payload-flattening-threshold: 10
-require:
-    - https://raw.githubusercontent.com/Azure/azure-rest-api-specs/e2de3bd42679ab281a61c52b4a24252d5394449a/specification/communication/data-plane/MediaComposition/readme.md
+input-file:
+    - mediaComposition.json
```
DominikMe, line 15: Instead of having a local copy, can we just push a new commit to https://github.com/Azure/azure-rest-api-specs/tree/feature/communication-media-composition?

sdk/communication/Azure.Communication.MediaComposition/src/autorest.md
```diff
@@ -7,11 +7,19 @@ Run `dotnet msbuild /t:GenerateCode` to generate code.
...
+```
 
+``` yaml
+directive:
+- from: swagger-document
+  where: $.definitions.*
+  transform: >
+    $["x-csharp-usage"] = "model,input,output";
+    $["x-csharp-formats"] = "json";
```
DominikMe, line 24: interesting, what's the reason that we need these transforms and what are they doing?

sdk/communication/Azure.Communication.MediaComposition/src/common.json
```diff
@@ -0,0 +1,172 @@
+{
+    "swagger": "2.0",
+    "info": {
+        "title": "Communication",
+        "description": "Communication Services API common types.",
+        "version": "2022-07-13"
+    },
+    "paths": {},
+    "definitions": {
+        "CommunicationErrorResponse": {
```
DominikMe, line 10: CommunicationErrorResponse and CommunicationIdentifierModel should use a relative path to reference the common.json from the media composition swagger.

DominikMe (commented):
This is awesome and looks much nicer. I'd prefer to keep referencing a swagger url instead of the local copy.

sdk/communication/Azure.Communication.MediaComposition/src/autorest.md
minnieliu: We had some issues with all the models generating as `internal`. I started a thread on the autorest channel and this was what I found. I couldn't find a lot of information on what the transforms do but they seem to be overriding the accessibility.
https://teams.microsoft.com/l/message/19:acbd512e57bd475198ea6bf4564599e3@thread.skype/1658189263863?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&parentMessageId=1658189263863&teamName=Azure%20SDK&channelName=AutoRest%20-%20CSharp&createdTime=1658189263863&allowXTenantAccess=false

DominikMe: Ah, I remember that thread. Not an intuitive fix but a fix nonetheless.

DominikMe (approved):
Swagger url update will happen in a follow up PR, so lgtm!

sdk/communication/Azure.Communication.MediaComposition/src/MediaCompositionClient.cs
minnieliu: Yea, I think you're right Dom. It is already set as a read-only property, and even if we do set on the client, it will not be respected. I can modify to null to avoid confusions

sdk/communication/Azure.Communication.MediaComposition/src/common.json
minnieliu: Yes, will do that as soon as the I link to the swagger in the specs repo

sdk/communication/Azure.Communication.MediaComposition/src/autorest.md
minnieliu: Will do in another PR!

Issue #21445, opened 2021-05-28
[FEATURE REQ] [Storage] Implement SecondaryThenPrimary  and SecondaryOnly policies


kasobol-msft:
Opening on behalf of customer.

V11 Storage had option to specify SecondaryThenPrimary  or SecondaryOnly for geo redundant accounts. See https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.storage.blob.blobrequestoptions.locationmode?view=azure-dotnet-legacy

We should offer similar functionality in V12.

Samples from V11:

```
// The PrimaryOnly LocationMode directs the request and all potential retries to go to the primary endpoint.
BlobRequestOptions locationModeRequestOptions = new BlobRequestOptions() { LocationMode = RetryPolicies.LocationMode.PrimaryOnly };
byteCount = blob.DownloadToByteArray(destinationArray, index: 0, accessCondition: null, options: locationModeRequestOptions);

// The PrimaryThenSecondary LocationMode directs the first request to go to the primary location.
// If this request fails with a retryable error, the retry will next hit the secondary location.
// Retries will switch back and forth between primary and secondary until the request succeeds, 
// or retry attempts have been exhausted.
locationModeRequestOptions = new BlobRequestOptions() { LocationMode = RetryPolicies.LocationMode.PrimaryThenSecondary };
byteCount = blob.DownloadToByteArray(destinationArray, index: 0, accessCondition: null, options: locationModeRequestOptions);

// The SecondaryOnly LocationMode directs the request and all potential retries to go to the secondary endpoint.
locationModeRequestOptions = new BlobRequestOptions() { LocationMode = RetryPolicies.LocationMode.SecondaryOnly };
byteCount = blob.DownloadToByteArray(destinationArray, index: 0, accessCondition: null, options: locationModeRequestOptions);

// The SecondaryThenPrimary LocationMode directs the first request to go to the secondary location.
// If this request fails with a retryable error, the retry will next hit the primary location.
// Retries will switch back and forth between secondary and primary until the request succeeds, or retry attempts
// have been exhausted.
locationModeRequestOptions = new BlobRequestOptions() { LocationMode = RetryPolicies.LocationMode.SecondaryThenPrimary };
byteCount = blob.DownloadToByteArray(destinationArray, index: 0, accessCondition: null, options: locationModeRequestOptions);
```

kasobol-msft:
As interim workaround we suggest to use custom policy:
```
using System;
[...]

PR #30368, opened 2022-08-05
GeoRedundantReadMode

jaschrep-msft:
Addresses #21445

Adds GeoRedundantReadMode enum to Azure.Storage.Common and adds a field for it to relevant packages' ClientOptions types.

Adds a check to Azure.Core.Argument (shared source) for if given enum value is defined in the type.


jaschrep-msft:

jaschrep-msft:
TODO: changelog after version bump from upcoming release

sdk/core/Azure.Core/src/Shared/Argument.cs
```diff
@@ -180,6 +180,21 @@ public static void AssertNotNullOrWhiteSpace(string value, string name)
...
+        /// </summary>
+        /// <param name="enumType">The type to validate against.</param>
+        /// <param name="value">The value to validate.</param>
+        /// <param name="name">The name of the parameter.</param>
+        /// <exception cref="ArgumentException"></exception>
+        public static void AssertEnumDefined(Type enumType, object value, string name)
+        {
+            if (!Enum.IsDefined(enumType, value))
+            {
+                throw new ArgumentException($"Value not defined in enum {value.GetType().FullName}.", name);
```
JoshLove-msft, line 194:
```suggestion
                throw new ArgumentException($"Value not defined in enum {enumType.FullName}.", name);
```

sdk/storage/Azure.Storage.Common/src/GeoRedundantReadMode.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        /// <summary>
+        /// Altnernate between primary and secondary endpoints, attempting secondary first.
+        /// </summary>
+        SecondaryThenPrimary = 1,
+
+        /// <summary>
+        /// Only target the primary endpoint.
+        /// </summary>
+        PrimaryOnly = 2,
```
JoshLove-msft, line 24: What would be the point of setting PrimaryOnly? Isn't this equivalent to leaving the secondary Uri null?

jaschrep-msft: Really just a parity inclusion. Not needed. Technically, secondary only isn't needed either, as the clients could just be instantiated with secondary from the start. We could get away with just the first two enum values.

They were both included as the original feature request was to port the enum and its functionality from v11, which had these values. But also v11 had a full `StorageUri` class that contained primary and secondary together, so it made a bit more sense back then.

seanmcc-msft (approved):

amnguye (approved):

sdk/storage/Azure.Storage.Common/src/GeoRedundantReadMode.cs
jaschrep-msft: Discussed separately with team. Will be removing PrimaryOnly and SecondaryOnly modes, as they are possible by just creating the client with the desired host. Adding support later, if necessary, would not be a breaking change.

sdk/storage/Azure.Storage.Common/src/GeoRedundantReadMode.cs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Azure.Storage
+{
+    /// <summary>
+    /// Pattern for retrying a read request between geo-redundant primary and secondary endpoints.
+    /// </summary>
+    public enum GeoRedundantReadMode
```
seanmcc-msft, line 9: Would it make any sense to include `None` as a default value?

seanmcc-msft (commented):
Does secondary endpoint only apply for the blob, DFS, and queue endpoints?

sdk/storage/Azure.Storage.Common/src/GeoRedundantReadMode.cs
jaschrep-msft: The SDK is already in PrimaryThenSecondary mode today, and just doesn't bother if the secondary endpoint wasn't provided. I don't know what None would help convey here.

jaschrep-msft:
> Does secondary endpoint only apply for the blob, DFS, and queue endpoints?

Yes. File Shares do not have secondary endpoints.

seanmcc-msft (approved):

jaschrep-msft:
@JoshLove-msft or @jsquire can I get an approval from one of you regarding the azure-core helper method being added?

JoshLove-msft (approved):

Issue #13545, opened 2020-07-17
ShareFileClient UploadRangeAsync timeout or reset on large file


dspeidel:
**Query/Question**
Using `Azure.Storage.Files.Shares` v12.2.3. I'm attempting to upload json files. It appears that small files work without issue, but large files (~100MB) seem to "reset" after 90 seconds or so. I wired into the `IProgress<long>` progressUpdate and I'm logging every 5%. You can see it chugs along to 20% then it seems to reset.

Log excerpt:
```
[7/16/2020 11:48:37 PM] Uploading file: [2020-07-15.json]
[7/16/2020 11:48:38 PM] ... 0% [0/101494994]
[7/16/2020 11:48:58 PM] ... 5% [5111808/101494994]
[7/16/2020 11:49:19 PM] ... 10% [10223616/101494994]
[7/16/2020 11:49:40 PM] ... 15% [15335424/101494994]
[7/16/2020 11:50:00 PM] ... 20% [20316160/101494994]
[7/16/2020 11:50:19 PM] ... 0% [0/101494994]
[7/16/2020 11:50:39 PM] ... 5% [5111808/101494994]
```
Is there a way to set a higher timeout or ensure it keeps going or some configuration I need to check/set/enable or...?

**Environment:**
`Azure.Storage.Files.Shares` v12.2.3.

Project properties
```
<PropertyGroup>
    <TargetFramework>netstandard2.0</TargetFramework>
    <AzureFunctionsVersion>v2</AzureFunctionsVersion>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="AsyncEnumerator" Version="4.0.2" />
    <PackageReference Include="Azure.Storage.Files.Shares" Version="12.2.3" />
    <PackageReference Include="IQ.Vanilla" Version="4.1.0" />
    <PackageReference Include="Microsoft.NET.Sdk.Functions" Version="1.0.29" />
    <PackageReference Include="System.Data.SqlClient" Version="4.5.1" />
  </ItemGroup>
```

Microsoft Visual Studio Enterprise 2019
Version 16.6.3

from `dotnet --info`
.NET Core SDK (reflecting any global.json):
 Version:   3.1.301
 Commit:    7feb845744

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.18362
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\3.1.301\

Host (useful for support):
[...]

PR #30398, opened 2022-08-08
Share upload StorageTransferOptions

jaschrep-msft:
Addresses #13545

Adds `StorageTransferOptions` support to `ShareFileClient.UploadAsync()`.

Note: service does not support parallel ranged uploads. Method applies a default value of 1 to `StorageTransferOptions.MaximumConcurrency` and will throw if value is manually set higher.


jaschrep-msft:

jaschrep-msft:
TODO: changelog after version bump from upcoming release

seanmcc-msft (approved):

PR #30399, opened 2022-08-08
Add option to configure Certificate Endpoint and fix recorded tests

christothes:


deer-prudence:
Thank you for making these changes!!

sdk/confidentialledger/Azure.Security.ConfidentialLedger/tests/ConfidentialLedgerClientLiveTests.cs
```diff
@@ -212,6 +214,7 @@ public async Task CreateAndGetAndDeleteUser()
         }
 
         [RecordedTest]
+        [LiveOnly]
```
JoshLove-msft, line 217: Why is this live only?

JoshLove-msft (approved):

sdk/confidentialledger/Azure.Security.ConfidentialLedger/tests/ConfidentialLedgerClientLiveTests.cs
christothes: There is some weirdness with the Proxy config in this scenario, since this test doesn't need the custom transport config but all the others do. Instead of special casing it I just marked it LiveOnly for now.

PR #30400, opened 2022-08-08
Fix issue related to receiver/sender caching

JoshLove-msft:
Fixes issue seen in https://portal.microsofticm.com/imp/v3/incidents/details/322426378/home

### DIFF
--- a/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Primitives/MessagingProvider.cs
+++ b/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Primitives/MessagingProvider.cs
@@ -139,7 +139,7 @@ namespace Microsoft.Azure.WebJobs.ServiceBus
             Argument.AssertNotNull(client, nameof(client));
             Argument.AssertNotNullOrEmpty(entityPath, nameof(entityPath));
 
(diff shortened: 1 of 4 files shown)


sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Primitives/MessagingProvider.cs
```diff
@@ -212,5 +212,10 @@ protected internal virtual ServiceBusSessionProcessor CreateSessionProcessor(Ser
             processor.SessionClosingAsync += Options.SessionClosingHandler;
             return processor;
         }
+
+        private static string GenerateCacheKey(string fullyQualifiedNamespace, string entityPath)
+        {
+            return $"{fullyQualifiedNamespace}/{entityPath}";
```
jsquire, line 218: Should we consider normalizing case and/or trimming spaces?

jsquire (approved):

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Primitives/MessagingProvider.cs
JoshLove-msft: Hmm, I think the namespace should already be normalized. As for the entity path, we could normalize but I'd prefer to err on the side of keeping it really simple and only using the cache for exact matches.

PR #30409, opened 2022-08-09
Update XML docs for AddAzureClients and AddAzureClientsCore

scottaddie:
Edit pass on the XML docs updates in https://github.com/Azure/azure-sdk-for-net/pull/30317

### DIFF
--- a/sdk/extensions/Microsoft.Extensions.Azure/src/AzureClientServiceCollectionExtensions.cs
+++ b/sdk/extensions/Microsoft.Extensions.Azure/src/AzureClientServiceCollectionExtensions.cs
@@ -29,7 +29,7 @@ namespace Microsoft.Extensions.Azure
 
         /// <summary>
         /// Adds the minimum essential Azure SDK interop services like <see cref="AzureEventSourceLogForwarder"/> and <see cref="AzureComponentFactory"/> to the specified <see cref="IServiceCollection"/> without registering any client types.
-        /// Azure SDK log forwarding to <see cref="ILogger"/> will not be enabled by default, but can be enabled by calling the <see cref="AzureEventSourceLogForwarder.Start"/> method.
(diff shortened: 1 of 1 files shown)


JoshLove-msft (approved):

sdk/extensions/Microsoft.Extensions.Azure/src/AzureClientServiceCollectionExtensions.cs
```diff
@@ -29,7 +29,7 @@ public static void AddAzureClients(this IServiceCollection collection, Action<Az
 
         /// <summary>
         /// Adds the minimum essential Azure SDK interop services like <see cref="AzureEventSourceLogForwarder"/> and <see cref="AzureComponentFactory"/> to the specified <see cref="IServiceCollection"/> without registering any client types.
-        /// Azure SDK log forwarding to to <see cref="ILogger"/> will not be enabled by default, but can be enabled by calling the <see cref="AzureEventSourceLogForwarder.Start"/> method.
+        /// Azure SDK log forwarding to <see cref="ILogger"/> is disabled by default. It can be enabled by calling the <see cref="AzureEventSourceLogForwarder.Start"/> method.
```
jsquire, line 32: Should we consider including an `<example>` snippet here to demonstrate how to get at the forwarder to start it?

jsquire (approved):

sdk/extensions/Microsoft.Extensions.Azure/src/AzureClientServiceCollectionExtensions.cs
scottaddie: I think that example snippet would feel out-of-place, since it's not a snippet showing how to call the `AddAzureClientsCore` overload. I'll add an example of this when I address https://github.com/Azure/azure-sdk-for-net/pull/30409.

PR #30430, opened 2022-08-09
Additional samples

JoshLove-msft:
Contributes to https://github.com/Azure/azure-sdk-for-net/issues/28215 
Also fixes bug where Footer was not being included on received message.

### DIFF
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/README.md
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/README.md
@@ -152,8 +152,6 @@ We guarantee that all client instance methods are thread-safe and independent of
 * [Dead letter a message](#dead-letter-a-message)
 * [Using the processor](#using-the-processor)
 * [Authenticating with Azure.Identity](#authenticating-with-azureidentity)
-* [Initiating the connection with a custom endpoint](#initiating-the-connection-with-a-custom-endpoint)
-* [Working with sessions](#working-with-sessions)
 * [More samples](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/servicebus/Azure.Messaging.ServiceBus/samples/README.md)
 
 ### Send and receive a message
@@ -417,23 +415,6 @@ The [Azure Identity library](https://github.com/Azure/azure-sdk-for-net/tree/mai
 // Create a ServiceBusClient that will authenticate through Active Directory
 string fullyQualifiedNamespace = "yournamespace.servicebus.windows.net";
 ServiceBusClient client = new ServiceBusClient(fullyQualifiedNamespace, new DefaultAzureCredential());
-```
-
-### Initiating the connection with a custom endpoint
-
-If an alternative host name is needed to establish the connection to the service, a custom endpoint address can be provided through the `ServiceBusClientOptions`. The client will use this endpoint to open the initial connection, and then will use the default endpoint provided by the Service Bus service for all following operations and validation.
-
-```C# Snippet:ServiceBusCustomEndpoint
-// Connect to the service using a custom endpoint
-string connectionString = "<connection_string>";
-string customEndpoint = "<custom_endpoint>";
-
-var options = new ServiceBusClientOptions
-{
-    CustomEndpointAddress = new Uri(customEndpoint)
-};
-
-ServiceBusClient client = new ServiceBusClient(connectionString, options);
 ```
 
 ### Working with Sessions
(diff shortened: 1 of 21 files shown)


sdk/servicebus/Azure.Messaging.ServiceBus/README.md
```diff
@@ -419,23 +418,6 @@ string fullyQualifiedNamespace = "yournamespace.servicebus.windows.net";
 ServiceBusClient client = new ServiceBusClient(fullyQualifiedNamespace, new DefaultAzureCredential());
 ```
 
-### Initiating the connection with a custom endpoint
```
JoshLove-msft, line 422: Moved this into the Advanced Config sample.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
```diff
@@ -535,6 +535,19 @@ public static ServiceBusReceivedMessage AmqpMessageToSBMessage(AmqpMessage amqpM
 
             // footer
 
+            if (amqpMessage.Footer != null)
```
JoshLove-msft, line 538: Looks like this needs to be fixed in EH as well.

JoshLove-msft: Uh nevermind - https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Azure.Messaging.EventHubs/src/Amqp/AmqpMessageConverter.cs#L794

m-redding (approved):
These changes lgtm

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
```diff
@@ -535,6 +535,19 @@ public static ServiceBusReceivedMessage AmqpMessageToSBMessage(AmqpMessage amqpM
 
             // footer
 
+            if ((sections & SectionFlag.Footer) != 0)
```
jsquire, line 538: whoops!

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Samples/Sample13_AdvancedConfiguration.cs
```diff
@@ -0,0 +1,59 @@
...
+// Licensed under the MIT License.
+
+using System;
+using System.Net;
+
+namespace Azure.Messaging.ServiceBus.Tests.Samples
+{
+    public class Sample13_AdvancedConfiguration : ServiceBusLiveTestBase
+    {
+        public void ConfigureProxy()
```
jsquire, line 11: These be tagged with `[Test]` so that they're executed nightly.

jsquire (approved):

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Samples/Sample13_AdvancedConfiguration.cs
JoshLove-msft: These aren't asserting anything. They will still be built as part of CI, but there is no need to execute them.

jsquire: My understanding is that the goal of snippets was to try to make sure that the code compiles and can be run - acting as smoke tests.   I don't feel particularly strongly in this case, but it seems to drift from the intention.

JoshLove-msft: I think compiles is the main thing. The run part necessitates conditionals to set the connection string appropriately. IMO it is not worth the added friction for small snippets like this. I also don't feel too strongly either way, but I'd prefer to err on the side of making it easier to create samples.

JoshLove-msft: I'd also add that we have coverage for all of these scenarios in unit and live tests.

JoshLove-msft: Ended up switching over to tests.

PR #30436, opened 2022-08-09
Add bespoke methods for Text Analytics

### LINKED ISSUE #28067: [Text Analytics] Add Missing Bespoke Methods
# Summary

Each analysis task that can be performed should have a set of bespoke methods so that the API surface is consistent, and some tasks are not hidden from developers.  The missing bespoke methods are for the extractive summarization and custom text tasks (custom entities, single/multi classification). 

# References and Resources

- [Python API analysis](https://gist.github.com/kristapratico/b70f6513e25d4f4af5b9866cf278d240#adding-bespoke-methods-on-client)

heaths:
Resolves #28067

### DIFF
--- a/sdk/textanalytics/Azure.AI.TextAnalytics/src/ClassifyDocumentOperation.cs
+++ b/sdk/textanalytics/Azure.AI.TextAnalytics/src/ClassifyDocumentOperation.cs
@@ -1,0 +1,292 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Azure.AI.TextAnalytics.Models;
+using Azure.AI.TextAnalytics.ServiceClients;
+using Azure.Core;
+using Azure.Core.Pipeline;
+
+namespace Azure.AI.TextAnalytics
+{
+    /// <summary> Pageable operation class for classifying single and multiple labels from multiple documents using long running operation. </summary>
+    public class ClassifyDocumentOperation : PageableOperation<ClassifyDocumentResultCollection>, IOperation<AsyncPageable<ClassifyDocumentResultCollection>>
+    {
+        internal readonly IDictionary<string, int> _idToIndexMap;
+
+        private readonly bool? _showStats;
+        private readonly string _jobId;
+        private readonly ServiceClient _serviceClient;
+        private readonly ClientDiagnostics _diagnostics;
+        private readonly OperationInternal<AsyncPageable<ClassifyDocumentResultCollection>> _operationInternal;
+
+        private TextAnalyticsOperationStatus _status;
+        private DateTimeOffset? _expiresOn;
+        private DateTimeOffset _lastModified;
+        private DateTimeOffset _createdOn;
+        private Page<ClassifyDocumentResultCollection> _firstPage;
+
+        /// <summary>
+        /// Gets an ID representing the operation that can be used to poll for the status
+        /// of the long-running operation.
+        /// </summary>
+        public override string Id { get; }
+
+        /// <summary>
+        /// Time when the operation was created on.
+        /// </summary>
+        public virtual DateTimeOffset CreatedOn => _createdOn;
+
+        /// <summary>
+        /// Time when the operation will expire.
+        /// </summary>
+        public virtual DateTimeOffset? ExpiresOn => _expiresOn;
+
+        /// <summary>
+        /// Time when the operation was last modified on
+        /// </summary>
+        public virtual DateTimeOffset LastModified => _lastModified;
+
+        /// <summary>
+        /// Gets the status of the operation.
+        /// </summary>
+        public virtual TextAnalyticsOperationStatus Status => _status;
+
+        /// <summary>
+        /// Gets the final result of the long-running operation asynchronously.
+        /// </summary>
+        /// <remarks>
+        /// This property can be accessed only after the operation completes successfully (HasValue is true).
+        /// </remarks>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public override AsyncPageable<ClassifyDocumentResultCollection> Value => _operationInternal.Value;
+
+        /// <summary>
+        /// Returns true if the long-running operation completed.
+        /// </summary>
+        public override bool HasCompleted => _operationInternal.HasCompleted;
+
+        /// <summary>
+        /// Returns true if the long-running operation completed successfully and has produced final result (accessible by Value property).
+        /// </summary>
+        public override bool HasValue => _operationInternal.HasValue;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ClassifyDocumentOperation"/> class.
+        /// </summary>
+        /// <param name="operationId">The ID of this operation.</param>
+        /// <param name="client">The client used to check for completion.</param>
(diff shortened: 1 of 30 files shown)


sdk/textanalytics/Azure.AI.TextAnalytics/src/ClassifyDocumentOperation.cs
```diff
@@ -0,0 +1,293 @@
...
+                var response = await _serviceClient.AnalyzeTextJobStatusNextPageAsync(nextLink, pageSizeHint, _idToIndexMap, cancellationToken).ConfigureAwait(false);
+                return Page.FromValues(new List<ClassifyDocumentResultCollection>() { Transforms.ConvertClassifyDocumentResultCollection(response.Value, _idToIndexMap) }, response.Value.NextLink, response.GetRawResponse());
+            }
+
+            return PageableHelpers.CreateAsyncEnumerable(_ => Task.FromResult(_firstPage), NextPageFunc);
+        }
+
+        async ValueTask<OperationState<AsyncPageable<ClassifyDocumentResultCollection>>> IOperation<AsyncPageable<ClassifyDocumentResultCollection>>.UpdateStateAsync(bool async, CancellationToken cancellationToken)
+        {
+            // BUGBUG
```
kristapratico, line 257: is there actually a bug here?

sdk/textanalytics/Azure.AI.TextAnalytics/src/ClassifyDocumentOperation.cs
```diff
@@ -0,0 +1,293 @@
...
+                    .ConfigureAwait(false);
+
+                return OperationState<AsyncPageable<ClassifyDocumentResultCollection>>.Failure(rawResponse, requestFailedException);
+            }
+            else if (response.Value.Status == TextAnalyticsOperationStatus.Cancelled)
+            {
+                return OperationState<AsyncPageable<ClassifyDocumentResultCollection>>.Failure(rawResponse,
+                    new RequestFailedException("The operation was canceled so no value is available."));
+            }
+
```
kristapratico, line 289: ignore if it's handled somewhere else and I don't see it, but what if partially completed is returned?

https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cognitiveservices/data-plane/Language/stable/2022-05-01/analyzetext.json#L2200

sdk/textanalytics/Azure.AI.TextAnalytics/src/MultiLabelClassifyOptions.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Azure.AI.TextAnalytics
+{
+    /// <summary>
+    /// Options that allow callers to specify details about how the operation
+    /// is run and what information is returned from it by the service.
+    /// </summary>
+    public class MultiLabelClassifyOptions : TextAnalyticsRequestOptions
```
kristapratico, line 10: It looks like TextAnalyticsRequestOptions includes modelVersion which isn't supported by MultiLabelClassify

heaths: I noticed that as well, but wasn't sure if we need to "hide" it. I will see if we could skip the derivation and either just define it standalone, or break up the hierarchy a bit. IIRC, Java has completely separate, non-derived options bags classes to avoid this.

sdk/textanalytics/Azure.AI.TextAnalytics/src/ClassifyDocumentOperation.cs
heaths: IIRC, it was my understanding that it shouldn't happen in practice and is usually a service issue, though maybe we should handle it. That said, I was looking at how you handle it in Python, and - even using the symbols it's defined as e.g. `PARTIALLY_COMPLETED` or used in e.g. `_SUCCEEDED` I don't see it used. What do you do with such a result?

sdk/textanalytics/Azure.AI.TextAnalytics/src/ClassifyDocumentOperation.cs
heaths: Nope. Forgot to remove that.

sdk/textanalytics/Azure.AI.TextAnalytics/src/ClassifyDocumentOperation.cs
heaths: I don't see "partiallyCompleted" in Java as well, but do see "partiallySucceeded", but seems Java TA just treats it as another success: https://cs.github.com/Azure/azure-sdk-for-java?q=partiallySucceeded+path%3Asdk%2Ftextanalytics

/cc @mssfang

kristapratico: It's rare, but I still see it occur in live tests every now and then. Generally one action fails with an InternalServerError and the LRO is reported as "partiallyCompleted". It's important in Python because if we don't account for it, the LRO will hang forever and never get marked as finished (our azure-core impl of the poller doesn't include "partiallyCompleted" since it's not a standard LRO status per the REST guidelines). 

It's used in _FINISHED in that same file you were looking at - it stops the polling so we can raise an exception. If the service ever decides to return an error target like we discussed before, maybe we'd return partial results, but that's currently not the case.

heaths: I see - so at the very least I should make sure it's terminal but maybe not much more than that.

kristapratico: Yeah I think that's enough

mssfang: Yeah. In TA java, it is tread as the LRO as completed.

sdk/textanalytics/Azure.AI.TextAnalytics/src/MultiLabelClassifyOptions.cs
mssfang: In java, we try to not extend from TextAnalyticsRequestOptions

mssfang (approved):

sdk/textanalytics/Azure.AI.TextAnalytics/src/ClassifyDocumentOperation.cs
heaths: @mssfang were you going to add "partiallyCompleted" support? From what I see, you only have "partiallySucceeded" but as @kristapratico [pointed out](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cognitiveservices/data-plane/Language/stable/2022-05-01/analyzetext.json#L2200), the swagger specifies "partiallyCompleted".

heaths: We had some internal discussion as to whether it should be a successful or failed completion. Because Java and .NET return different result types in separate collection properties, we can actually treat these as success; conversely, JavaScript and Python return all results in the original order, and the service does not provide any sort of reference to know which results we need to offset thus JS and Python treat this condition as failure.

IMO, we should have consistent behavior.

I'm going to add support to terminate `partiallyCompleted` as successful completion for this beta release, but will start an internal thread with the service team to get their thoughts on this.

kristapratico (approved):
🚢

PR #30191, opened 2022-07-28
Track 2 Migration for Guest Configuration

anineela:
# Contributing to the Azure SDK

Our latest spec is already merged to main: https://github.com/Azure/azure-rest-api-specs/tree/main/specification/guestconfiguration/resource-manager/Microsoft.GuestConfiguration/stable/2022-01-25

Build succeeds:
image

Tests:
image

Currently, the "GetAllAsync" list tests are disabled due to a bug fix being done by Yao. Brought this to Arthur Ma and Yao Kou's attention.

PR #30440, opened 2022-08-10
[Guestconfiguration] Update atuorest.md to ensure most of the methods could be generated.

Yao725:
This is just a temporary work around.

Related to https://github.com/Azure/azure-sdk-for-net/pull/30191


sdk/guestconfiguration/Azure.ResourceManager.GuestConfiguration/src/autorest.md
```diff
@@ -54,4 +59,10 @@ rename-rules:
   URI: Uri
   Etag: ETag|etag
 
+directive:
+  - remove-operation: GuestConfigurationAssignmentsVMSS_List
```
ArthurMa1978, line 63: Should create an issue of why the list method not work properly here

ArthurMa1978 (approved):

sdk/guestconfiguration/Azure.ResourceManager.GuestConfiguration/src/autorest.md
Yao725: Created: https://github.com/Azure/autorest.csharp/issues/2560

PR #30478, opened 2022-08-11
[Communication-Common] fix CommunicationIdentifier equality operator override

DominikMe:
Fixes a bug in the `CommunicationIdentifier`'s `==` operator override.

```c#
if (identifier == null)
``` 
would always throw, thanks @minnieliu for discovering this.

No customer has been impacted as the operator override is new, and package hasn't been released yet.

### DIFF
--- a/sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
+++ b/sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
@@ -33,7 +33,8 @@ namespace Azure.Communication
         /// <param name="left">The first identifier to compare.</param>
         /// <param name="right">The second identifier to compare.</param>
         /// <returns>True if the types and <see cref="RawId"/> match.</returns>
-        public static bool operator ==(CommunicationIdentifier left, CommunicationIdentifier right) => left.GetType() == right.GetType() && Equals(left, right);
+        public static bool operator ==(CommunicationIdentifier left, CommunicationIdentifier right)
+            => ReferenceEquals(left, right) || left is not null && right is not null && Equals(left, right);
 
         /// <summary>
         /// Overrides the non-equality operator.
@@ -41,7 +42,7 @@ namespace Azure.Communication
         /// <param name="left">The first identifier to compare.</param>
         /// <param name="right">The second identifier to compare.</param>
         /// <returns>True if the types or <see cref="RawId"/> values are different.</returns>
-        public static bool operator !=(CommunicationIdentifier left, CommunicationIdentifier right) => left.GetType() != right.GetType() || !Equals(left, right);
+        public static bool operator !=(CommunicationIdentifier left, CommunicationIdentifier right) => !(left == right);
 
         /// <summary>
         /// Creates a <see cref="CommunicationIdentifier"/> from a given rawId.
--- a/sdk/communication/Azure.Communication.Common/tests/CommunicationIdentifierTest.cs
+++ b/sdk/communication/Azure.Communication.Common/tests/CommunicationIdentifierTest.cs
@@ -119,11 +119,21 @@ namespace Azure.Communication
             var sameIdentifier = new CommunicationUserIdentifier("123");
             var otherIdentifier = new CommunicationUserIdentifier("124");
             var otherTypeIdentifier = new MicrosoftTeamsUserIdentifier("123");
+            CommunicationIdentifier? nullIdentifier = null;
 
             Assert.False(identifier == null);
             Assert.False(null == identifier);
             Assert.True(identifier != null);
             Assert.True(null != identifier);
+
+            Assert.True(null as CommunicationIdentifier == null as CommunicationIdentifier);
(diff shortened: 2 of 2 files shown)


paolamvhz (approved):

sdk/communication/Azure.Communication.Common/tests/CommunicationIdentifierTest.cs
```diff
@@ -111,5 +111,31 @@ public void RawIdIsOverriddenBySubTypes()
                 Assert.AreNotEqual(baseType, implementation.GetProperty(nameof(CommunicationIdentifier.RawId))?.DeclaringType);
             }
         }
+
+        [Test]
+        public void EqualityOperatorOverrideDoesntThrow()
+        {
+            var identifier = new CommunicationUserIdentifier("123");
```
mikehang-msft, line 118: not sure if this is the case but we can try with a scenario where identifier is null (identifier = null). This should be passed as well.

sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
```diff
@@ -33,15 +33,16 @@ public override bool Equals(object obj)
         /// <param name="left">The first identifier to compare.</param>
         /// <param name="right">The second identifier to compare.</param>
         /// <returns>True if the types and <see cref="RawId"/> match.</returns>
-        public static bool operator ==(CommunicationIdentifier left, CommunicationIdentifier right) => left.GetType() == right.GetType() && Equals(left, right);
+        public static bool operator ==(CommunicationIdentifier left, CommunicationIdentifier right)
+            => ReferenceEquals(left, right) || left is not null && right is not null && Equals(left, right);
```
mikehang-msft, line 37: Not with this fix, but I notice a few classes accessing left.Equals() where left can be null itself. I think this case can cause an exception too.

DominikMe: Do you have an example where `left.Equals` is being called?

minnieliu:
Thanks for making the fix Dom :)

minnieliu (approved):

sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
DominikMe: @mikehang-msft Resolving to complete this PR, let me know if you encounter any unsafe occurrence and we can follow up.

mikehang-msft: Sorry for a late response. Yes, we should resolve this PR and tackle other issues later. 

I did a quick search on Azure.Communication.MediaComposition project (there could be more in the solution). The following classes have left.Equals() calls. They are all auto-gen and implementing IEquatable<> interface.

        public static bool operator ==(AudiencePosition left, AudiencePosition right) => left.Equals(right);	 mikehang-msft\azure-sdk-for-net\sdk\communication\Azure.Communication.MediaComposition\src\Generated\Models\AudiencePosition.cs	39	28
        public static bool operator ==(CompositionStreamState left, CompositionStreamState right) => left.Equals(right);	 mikehang-msft\azure-sdk-for-net\sdk\communication\Azure.Communication.MediaComposition\src\Generated\Models\CompositionStreamState.cs	36	28
        public static bool operator ==(LayerVisibility left, LayerVisibility right) => left.Equals(right);	 mikehang-msft\azure-sdk-for-net\sdk\communication\Azure.Communication.MediaComposition\src\Generated\Models\LayerVisibility.cs	33	28
        public static bool operator ==(LayoutType left, LayoutType right) => left.Equals(right);	 mikehang-msft\azure-sdk-for-net\sdk\communication\Azure.Communication.MediaComposition\src\Generated\Models\LayoutType.cs	42	28
        public static bool operator ==(MediaInputType left, MediaInputType right) => left.Equals(right);	 mikehang-msft\azure-sdk-for-net\sdk\communication\Azure.Communication.MediaComposition\src\Generated\Models\MediaInputType.cs	57	28
        public static bool operator ==(MediaOutputType left, MediaOutputType right) => left.Equals(right);	 mikehang-msft\azure-sdk-for-net\sdk\communication\Azure.Communication.MediaComposition\src\Generated\Models\MediaOutputType.cs	42	28
        public static bool operator ==(RtmpMode left, RtmpMode right) => left.Equals(right);	 mikehang-msft\azure-sdk-for-net\sdk\communication\Azure.Communication.MediaComposition\src\Generated\Models\RtmpMode.cs	33	28
        public static bool operator ==(SupportPosition left, SupportPosition right) => left.Equals(right);	 mikehang-msft\azure-sdk-for-net\sdk\communication\Azure.Communication.MediaComposition\src\Generated\Models\SupportPosition.cs	39	28

petrsvihlik:
thanks for catching and fixing this @minnieliu & @DominikMe !

PR #30486, opened 2022-08-12
Solve the pageable comments for MySql

ArthurMa1978:


sdk/mysql/Azure.ResourceManager.MySql/src/MySqlFlexibleServers/Generated/MySqlFlexibleServerConfigurationCollection.cs
```diff
@@ -127,30 +127,15 @@ async Task<Page<MySqlFlexibleServerConfigurationResource>> FirstPageFunc(int? pa
...
-                    return Page.FromValues(response.Value.Value.Select(value => new MySqlFlexibleServerConfigurationResource(Client, value)), response.Value.NextLink, response.GetRawResponse());
-                }
-                catch (Exception e)
-                {
-                    scope.Failed(e);
-                    throw;
-                }
-            }
-            return PageableHelpers.CreateAsyncEnumerable(FirstPageFunc, NextPageFunc);
+            return PageableHelpers.CreateAsyncEnumerable(FirstPageFunc, null);
```
fengzhou-msft, line 138: After the change does it mean the method will only get the first page even if there's more?

ArthurMa1978: fixed

fengzhou-msft (approved):

PR #30487, opened 2022-08-12
update eventParser, added tests

yyw-msft:
- reanamed EventParser to CallAutomationEventParser for better visibility
- removed some of the overloading methods for Parse and ParseMany
- added `Parse(string eventData, string eventType)`
- added tests


minwoolee-msft:
Can we add BinaryData parser back? 
EventGrid's parse/parsemany takes Binary data for their parsing (https://docs.microsoft.com/en-us/dotnet/api/azure.messaging.eventgrid.eventgridevent?view=azure-dotnet#methods). It would make sense for us to also support this syntax.

yyw-msft:
> Can we add BinaryData parser back? EventGrid's parse/parsemany takes Binary data for their parsing (https://docs.microsoft.com/en-us/dotnet/api/azure.messaging.eventgrid.eventgridevent?view=azure-dotnet#methods). It would make sense for us to also support this syntax.

yes, makes sense. people using event grid event  (IncomingCall Event) may already been providing binarydata. Added back. thx!

minwoolee-msft (approved):

PR #30501, opened 2022-08-12
[Azure.Communication.JobRouter] Fix several bugs

williamzhao87:
Fixes https://skype.visualstudio.com/SPOOL/_workitems/edit/2950692
Fixes https://skype.visualstudio.com/SPOOL/_workitems/edit/2951190
Fixes bug with several models using TimeSpan trimming off time longer than 60 seconds due to taking the Seconds property rather than TotalSeconds

### DIFF
--- a/sdk/communication/Azure.Communication.JobRouter/src/Models/UpdateJobOptions.cs
+++ b/sdk/communication/Azure.Communication.JobRouter/src/Models/UpdateJobOptions.cs
@@ -28,7 +28,7 @@ namespace Azure.Communication.JobRouter
         /// <summary>
         /// Id of the job.
         /// </summary>
-        public string JobId { get; set; }
+        public string JobId { get; }
 
         /// <summary>
         /// A set of key/value pairs that are identifying attributes used by the rules engines to make decisions.
@@ -50,7 +50,7 @@ namespace Azure.Communication.JobRouter
         public string ChannelId { get; set; }
 
         /// <summary> The priority of this job. </summary>
-        public int Priority { get; set; }
+        public int? Priority { get; set; }
 
         /// <summary> Reason code for cancelled or closed jobs. </summary>
         public string DispositionCode { get; set; }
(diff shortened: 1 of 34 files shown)


sdk/communication/Azure.Communication.JobRouter/src/Models/UpdateJobOptions.cs
```diff
@@ -50,7 +50,7 @@ public UpdateJobOptions(string jobId)
         public string ChannelId { get; set; }
 
         /// <summary> The priority of this job. </summary>
-        public int Priority { get; set; }
+        public int? Priority { get; set; }
```
sarkar-rajarshi, line 53: It was recommended to not have nullable types in public models. Should we set a default value of -1 instead?

sarkar-rajarshi: This would be applicable for all `int?`

sdk/communication/Azure.Communication.JobRouter/src/Models/UpdateWorkerOptions.cs
```diff
@@ -29,13 +29,13 @@ public UpdateWorkerOptions(string workerId)
...
-        public string WorkerId { get; set; }
+        public string WorkerId { get; }
 
         /// <summary> The total capacity score this worker has to manage multiple concurrent jobs. </summary>
-        public int TotalCapacity { get; set; }
+        public int? TotalCapacity { get; set; }
 
         /// <summary> A flag indicating this worker is open to receive offers or not. </summary>
-        public bool AvailableForOffers { get; set; }
+        public bool? AvailableForOffers { get; set; }
```
sarkar-rajarshi, line 38: Same here.

sarkar-rajarshi: Maybe a generic wrapper of `Optional<T>` needs to be created.

sdk/communication/Azure.Communication.JobRouter/src/Models/UpdateJobOptions.cs
williamzhao87: As discussed offline, according to SDK review board, nullable value types are ok since, there's no backwards compatibility issue

sdk/communication/Azure.Communication.JobRouter/src/Models/UpdateWorkerOptions.cs
williamzhao87: Same here

sarkar-rajarshi (approved):

PR #30506, opened 2022-08-15
Prototype for base resources

ArcturusZhang:

### DIFF
--- a/sdk/resourcemanager/Azure.ResourceManager/src/Resources/Generated/BasePolicyDefinitionResource.cs
+++ b/sdk/resourcemanager/Azure.ResourceManager/src/Resources/Generated/BasePolicyDefinitionResource.cs
@@ -1,0 +1,115 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Azure.Core;
+using Azure.ResourceManager.ManagementGroups;
+
+namespace Azure.ResourceManager.Resources
+{
+    /// <summary>
+    /// the base resource of policy definition
+    /// </summary>
+    public abstract class BasePolicyDefinitionResource : ArmResource
+    {
+        private readonly PolicyDefinitionData _data;
+
+        /// <summary>
+        /// ctor for mock
+        /// </summary>
+        protected BasePolicyDefinitionResource()
+        { }
+
+        internal BasePolicyDefinitionResource(ArmClient client, PolicyDefinitionData data) : base(client, data.Id)
+        {
+            HasData = true;
+            _data = data;
+        }
+
+        internal BasePolicyDefinitionResource(ArmClient client, ResourceIdentifier id) : base(client, id)
+        { }
+
+        /// <summary> Gets whether or not the current instance has data. </summary>
+        public virtual bool HasData { get; }
+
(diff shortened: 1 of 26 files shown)


sdk/resourcemanager/Azure.ResourceManager/src/Resources/Generated/BasePolicyDefinitionResource.cs
```diff
@@ -0,0 +1,115 @@
...
+using System.Threading.Tasks;
+using Azure.Core;
+using Azure.ResourceManager.ManagementGroups;
+
+namespace Azure.ResourceManager.Resources
+{
+    /// <summary>
+    /// the base resource of policy definition
+    /// </summary>
+    public abstract class BasePolicyDefinitionResource : ArmResource
```
m-nash, line 15: I think the naming convention that we want should match the common Data so in this case it would be `PolicyDefinitionResource`

sdk/resourcemanager/Azure.ResourceManager/src/Resources/Generated/BasePolicyDefinitionResource.cs
```diff
@@ -0,0 +1,115 @@
...
+        protected virtual Response<T> GetCore<T>(CancellationToken cancellationToken = default) where T : BasePolicyDefinitionResource
+        {
+            return Response.FromValue(default(T), default);
+        }
+
+        /// <summary>
+        /// base implementation
+        /// </summary>
+        /// <param name="cancellationToken"> The cancellation token to use. </param>
+        public virtual async Task<Response<BasePolicyDefinitionResource>> GetAsync(CancellationToken cancellationToken = default)
```
m-nash, line 75: I am pretty sure having virtual here will not play nicely with `new` on the dervied class

sdk/websites/Azure.ResourceManager.AppService/tests/TestsCase/SiteCollectionTests.cs
```diff
@@ -23,6 +23,37 @@ private async Task<WebSiteCollection> GetSiteCollectionAsync()
...
+            var name = Recording.GenerateAssetName("testSite");
+            var input = ResourceDataHelper.GetBasicSiteData(DefaultLocation);
+            var lro = await container.CreateOrUpdateAsync(WaitUntil.Completed, name, input);
+            var site = lro.Value;
+
+            int ftp = 0;
+            int scm = 0;
+            await foreach (var policy in site.GetBasicPublishingCredentialsPoliciesAsync())
+            {
+                // unfortunately all the element here is of type of BaseWebSitePublishingCredentialsPolicyResourceProxy
```
m-nash, line 40: most likely need to use a dynamic type here https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core.TestFramework/src/Instrumentation/AsyncPageableInterceptor.cs#L22.

probably `_inner.Current.GetType()` instead of `typeof(T)`

sdk/resourcemanager/Azure.ResourceManager/src/Resources/Generated/BasePolicyDefinitionResource.cs
ArcturusZhang: I will apply this naming convention in the actual implementation

PR #30526, opened 2022-08-16
resolve some issues in azure.resourcemanager.appservice

ArcturusZhang:


sdk/websites/Azure.ResourceManager.AppService/api/Azure.ResourceManager.AppService.netstandard2.0.cs
```diff
@@ -3269,6 +3141,29 @@ public partial class WebSiteConfigResource : Azure.ResourceManager.ArmResource
...
+    }
+    public partial class WebSiteConnectionStringConfigResource : Azure.ResourceManager.ArmResource
+    {
+        public static readonly Azure.Core.ResourceType ResourceType;
+        protected WebSiteConnectionStringConfigResource() { }
+        public virtual Azure.ResourceManager.AppService.ApiKeyVaultReferenceData Data { get { throw null; } }
+        public virtual bool HasData { get { throw null; } }
+        public static Azure.Core.ResourceIdentifier CreateResourceIdentifier(string subscriptionId, string resourceGroupName, string name, string connectionStringKey) { throw null; }
+        public virtual Azure.Response<Azure.ResourceManager.AppService.WebSiteConnectionStringConfigResource> Get(System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
+        public virtual System.Threading.Tasks.Task<Azure.Response<Azure.ResourceManager.AppService.WebSiteConnectionStringConfigResource>> GetAsync(System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
```
fengzhou-msft, line 3165: It feels like this resource could be configured as nonresource as the only operation on the Resource class is to get itself.

ArcturusZhang: Yes, we have tons of this kind of configuration resources in appservice.
I will have some discussion with Elendil to decide whether we should keep these as resources or non-resources.

PR #30575, opened 2022-08-18
add generated samples to `Azure.ResourceManager.Compute`

ArcturusZhang:


sdk/compute/Azure.ResourceManager.Compute/tests/Generated/Samples/Sample_CapacityReservationCollection.cs
```diff
@@ -0,0 +1,166 @@
...
+            // this example assumes you already have this CapacityReservationGroupResource created on azure
+            // for more information of creating CapacityReservationGroupResource, please refer to the document of CapacityReservationGroupResource
+            string subscriptionId = "{subscription-id}";
+            string resourceGroupName = "myResourceGroup";
+            string capacityReservationGroupName = "myCapacityReservationGroup";
+            ResourceIdentifier capacityReservationGroupResourceId = Compute.CapacityReservationGroupResource.CreateResourceIdentifier(subscriptionId, resourceGroupName, capacityReservationGroupName);
+            Compute.CapacityReservationGroupResource capacityReservationGroup = client.GetCapacityReservationGroupResource(capacityReservationGroupResourceId);
+
+            // get the collection of this CapacityReservationResource
+            Compute.CapacityReservationCollection collection = capacityReservationGroup.GetCapacityReservations();
```
fengzhou-msft, line 40: We cannot get rid of the `Compute.` namesapce prefix?

ArcturusZhang: The change is ongoing in the auto-PR. Will update this soon

ArcturusZhang: Just merge from main and regenerate. Now the `Compute.` prefix is removed

sdk/compute/Azure.ResourceManager.Compute/tests/Generated/Samples/Sample_AvailabilitySetCollection.cs
```diff
@@ -0,0 +1,246 @@
...
+            string subscriptionId = "{subscription-id}";
+            string resourceGroupName = "rgcompute";
+            ResourceIdentifier resourceGroupResourceId = ResourceGroupResource.CreateResourceIdentifier(subscriptionId, resourceGroupName);
+            ResourceGroupResource resourceGroupResource = client.GetResourceGroupResource(resourceGroupResourceId);
+
+            // get the collection of this AvailabilitySetResource
+            AvailabilitySetCollection collection = resourceGroupResource.GetAvailabilitySets();
+
+            // invoke the operation
+            string availabilitySetName = "aaaaaaaaaaaaaaaaaaaa";
```
fengzhou-msft, line 140: The main difference between `Get_AvailabilitySetsGetMinimumSetGen` and  `Get_AvailabilitySetsGetMaximumSetGen` is  the value of `availabilitySetName`, other code are basically the same. The original swagger samples probably have differences in the responses based on the example names. Can we config to suppress one of the two samples if we cannot check the duplication automatically?

sdk/compute/Azure.ResourceManager.Compute/tests/Generated/Samples/Sample_AvailabilitySetCollection.cs
```diff
@@ -0,0 +1,246 @@
...
+            string subscriptionId = "{subscription-id}";
+            string resourceGroupName = "rgcompute";
+            ResourceIdentifier resourceGroupResourceId = ResourceGroupResource.CreateResourceIdentifier(subscriptionId, resourceGroupName);
+            ResourceGroupResource resourceGroupResource = client.GetResourceGroupResource(resourceGroupResourceId);
+
+            // get the collection of this AvailabilitySetResource
+            AvailabilitySetCollection collection = resourceGroupResource.GetAvailabilitySets();
+
+            // invoke the operation and iterate over the result
+            await foreach (AvailabilitySetResource item in collection.GetAllAsync())
```
fengzhou-msft, line 200: Do we want to show the convenient way to do this with `await foreach (AvailabilitySetResource item in collection)`?

sdk/compute/Azure.ResourceManager.Compute/tests/Generated/Samples/Sample_AvailabilitySetCollection.cs
```diff
@@ -0,0 +1,246 @@
...
+                Console.WriteLine($"Succeeded on id: {resourceData.Id}");
+            }
+
+            Console.WriteLine($"Succeeded");
+        }
+
+        // AvailabilitySets_List_MinimumSet_Gen
+        [NUnit.Framework.Test]
+        [NUnit.Framework.Ignore("Only verifying that the sample builds")]
+        public async Task GetAll_AvailabilitySetsListMinimumSetGen()
```
fengzhou-msft, line 215: This example is identical to `GetAll_AvailabilitySetsListMaximumSetGen`. Similar comments with the Get methods examples to remove duplication with config or automatically if possible.

sdk/compute/Azure.ResourceManager.Compute/tests/Generated/Samples/Sample_AvailabilitySetCollection.cs
ArcturusZhang: I prefer we do not check or suppress samples.
Samples will be show in the REST API doc page with the swagger example - when we have a swagger example, we should have a corresponding dotnet sample, otherwise the page viewer might have questions, like, why this example does not have its corresponding dotnet sample while other languages have?

sdk/compute/Azure.ResourceManager.Compute/tests/Generated/Samples/Sample_AvailabilitySetCollection.cs
ArcturusZhang: Well, I kind of do not want a special case here - because some examples might have optional parameters with values in the example, such like `collection.GetAllAsync(top: 3)`

fengzhou-msft (approved):

PR #30614, opened 2022-08-18
Init Azure.ResourceManager.EventGrid

ArthurMa1978:


sdk/eventgrid/Azure.ResourceManager.EventGrid/tests/EventGridManagementTestBase.cs
```diff
@@ -0,0 +1,41 @@
...
+        public void CreateCommonClient()
+        {
+            Client = GetArmClient();
+        }
+
+        protected async Task<ResourceGroupResource> CreateResourceGroup(SubscriptionResource subscription, string rgNamePrefix, AzureLocation location)
+        {
+            string rgName = Recording.GenerateAssetName(rgNamePrefix);
+            ResourceGroupData input = new ResourceGroupData(location);
+            var lro = await subscription.GetResourceGroups().CreateOrUpdateAsync(WaitUntil.Completed, rgName, input);
```
benbp, line 37: There are no live tests in this PR, but in case they will get added (similar to other mgmt libraries), is deletion handled in the tests?

ArthurMa1978: We are still digging that the framework can handle all clean up

Yao725 (approved):

PR #30618, opened 2022-08-18
Init Azure.ResourceManager.StorageCache

ArthurMa1978:


Yao725 (approved):

sdk/storagecache/Azure.ResourceManager.StorageCache/src/autorest.md
```diff
@@ -0,0 +1,64 @@
...
+      to: StorageTargets_Flush
+  - rename-operation:
+      from: StorageTarget_Suspend
+      to: StorageTargets_Suspend
+  - rename-operation:
+      from: StorageTarget_Resume
+      to: StorageTargets_Resume
+  - rename-operation:
+      from: StorageTarget_Invalidate
+      to: StorageTargets_Invalidate
```
ArcturusZhang, line 63: I am OK with the renamings here, but why do we have to do this?
As far as I know, even if we do not have these renamings, the generation should be fine. But this will change the information in our operations' summary to show the incorrect operationId.
If possible we could remove this..

ArcturusZhang (approved):

PR #30636, opened 2022-08-19
[BOT RULES] Adding Mention

ronniegeraghty:
Adding Mention for Container Registry bot rules to coincide with CODEOWNERS file change here: #30634

### DIFF
--- a/.github/fabricbot.json
+++ b/.github/fabricbot.json
@@ -1709,7 +1709,8 @@
           ],
           "mentionees": [
             "toddysm",
-            "northtyphoon"
+            "northtyphoon",
+            "ShivangiReja"
           ]
         },
         {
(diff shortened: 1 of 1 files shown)


.github/fabricbot.json
```diff
@@ -1709,7 +1709,8 @@
           ],
           "mentionees": [
             "toddysm",
-            "northtyphoon"
+            "northtyphoon",
+            "ShivangiReja"
```
jsquire, line 1713: Shivangi is not part of the service team and should not be mentioned for "Service Attention" issues.

jsquire (changes requested):

.github/fabricbot.json
ronniegeraghty: Removed Shivangi

jsquire (approved):

Issue #29536, opened 2022-06-28
[FEATURE REQ] Implement ScaleController's Burst Scale Mode Feature


Tonewall:
### Library name

Microsoft.Azure.WebJobs.Extensions.ServiceBus

### Please describe the feature.

ICM regarding slow scale out: https://portal.microsofticm.com/imp/v3/incidents/details/316361790/home
Need burst scale mode feature implemented to runtime scale monitoring. The ScaleController is already capable of this [ScaleController Logic](https://msazure.visualstudio.com/One/_git/AAPT-Antares-ScaleController?path=/src/Microsoft.Azure.AppService.ScaleController.Extension/ScalerBase.cs&version=GBv2.x&line=290&lineEnd=318&lineStartColumn=12&lineEndColumn=14&lineStyle=plain&_a=contents).
Just need the above logic to be implemented in the ServiceBusScaleMonitor.cs [GetScaleStatusCore()](https://github.com/Azure/azure-sdk-for-net/blob/fcee313b5ecf26d5d3de37dd56d32fd115ee498f/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusScaleMonitor.cs#L206)

jsquire:
@JoshLove-msft :  Would you please triage to add costing and share your thoughts?

JoshLove-msft:
@Tonewall can you please point us to the burst scale mode implementation that you reference?

JoshLove-msft:
> @Tonewall can you please point us to the burst scale mode implementation that you reference?

Nevermind, I see it is in the summary!

JoshLove-msft:
This seems like it should be fairly straightforward. We are already collecting the QueueTime. It would just be a matter of checking the last TriggerMetric instances QueueTime like they do on the host.

danielmarbach:
I'd be happy to give this a go but I would require a bit more insights. Based on the current description that also links to some internal repo that I don't have access to, I'm not sure if there is much I can do

jsquire:
With the target-based scaling improvements and (forthcoming?) efforts to move the scale controller into the extension entirely, I don't believe this is relevant any longer.   I'm going to close this out.   @alrod, @chiangvincent, if you're still planning on this, please reopen.

PR #30639, opened 2022-08-19
Make ServiceBusScaleMonitor Public for Use in Scale Controller

chiangvincent:

# Overview
Currently, code in the Scale Controller and Extensions is duplicated (code for `GetMetricsAsync` and `GetScaleStatus` exist in the ActivationTask and ScalerBase classes in Scale Controller, respectively). We are currently redesigning the scale controller, and the first phase is to re-platform the scale controller to load the ServiceBusSDK. This allows us to directly instantiate + call into `ServiceBusScaleMonitor` methods. This also unifies any logical implementations that have diverged over time between runtime driven scaling (handled by extensions), and regular scaling (handled by scale controller). For example, a we had a recent CRI where `ServiceBusScaleMonitor` did not support burst scaling, but scale controller did. See [open issue.](https://github.com/Azure/azure-sdk-for-net/issues/29536)

# Testing
After making the methods public, we generated local nuget packages which we referenced in our redesign (see redesign repo [here](https://github.com/chiangvincent/ScaleController/tree/main/ScaleController/nuget))

### DIFF
--- a/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/api/Microsoft.Azure.WebJobs.Extensions.ServiceBus.netstandard2.0.cs
+++ b/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/api/Microsoft.Azure.WebJobs.Extensions.ServiceBus.netstandard2.0.cs
@@ -116,18 +116,18 @@ namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
     public partial class ServiceBusScaleMonitor : Microsoft.Azure.WebJobs.Host.Scale.IScaleMonitor, Microsoft.Azure.WebJobs.Host.Scale.IScaleMonitor<Microsoft.Azure.WebJobs.ServiceBus.Listeners.ServiceBusTriggerMetrics>
     {
         public ServiceBusScaleMonitor(string functionId, Microsoft.Azure.WebJobs.ServiceBus.ServiceBusEntityType serviceBusEntityType, string entityPath, string connection, Microsoft.Extensions.Logging.ILoggerFactory loggerFactory, Microsoft.Extensions.Configuration.IConfiguration configuration, Microsoft.Extensions.Azure.AzureComponentFactory componentFactory, Microsoft.Azure.WebJobs.ServiceBus.MessagingProvider messagingProvider, Microsoft.Extensions.Azure.AzureEventSourceLogForwarder logForwarder, Microsoft.Extensions.Options.IOptions<Microsoft.Azure.WebJobs.ServiceBus.ServiceBusOptions> options) { }
-        public Microsoft.Azure.WebJobs.Host.Scale.ScaleMonitorDescriptor Descriptor { get { throw null; } }
-        public System.Threading.Tasks.Task<Microsoft.Azure.WebJobs.ServiceBus.Listeners.ServiceBusTriggerMetrics> GetMetricsAsync() { throw null; }
-        public Microsoft.Azure.WebJobs.Host.Scale.ScaleStatus GetScaleStatus(Microsoft.Azure.WebJobs.Host.Scale.ScaleStatusContext<Microsoft.Azure.WebJobs.ServiceBus.Listeners.ServiceBusTriggerMetrics> context) { throw null; }
+        Microsoft.Azure.WebJobs.Host.Scale.ScaleMonitorDescriptor Microsoft.Azure.WebJobs.Host.Scale.IScaleMonitor.Descriptor { get { throw null; } }
         System.Threading.Tasks.Task<Microsoft.Azure.WebJobs.Host.Scale.ScaleMetrics> Microsoft.Azure.WebJobs.Host.Scale.IScaleMonitor.GetMetricsAsync() { throw null; }
         Microsoft.Azure.WebJobs.Host.Scale.ScaleStatus Microsoft.Azure.WebJobs.Host.Scale.IScaleMonitor.GetScaleStatus(Microsoft.Azure.WebJobs.Host.Scale.ScaleStatusContext context) { throw null; }
+        System.Threading.Tasks.Task<Microsoft.Azure.WebJobs.ServiceBus.Listeners.ServiceBusTriggerMetrics> Microsoft.Azure.WebJobs.Host.Scale.IScaleMonitor<Microsoft.Azure.WebJobs.ServiceBus.Listeners.ServiceBusTriggerMetrics>.GetMetricsAsync() { throw null; }
+        Microsoft.Azure.WebJobs.Host.Scale.ScaleStatus Microsoft.Azure.WebJobs.Host.Scale.IScaleMonitor<Microsoft.Azure.WebJobs.ServiceBus.Listeners.ServiceBusTriggerMetrics>.GetScaleStatus(Microsoft.Azure.WebJobs.Host.Scale.ScaleStatusContext<Microsoft.Azure.WebJobs.ServiceBus.Listeners.ServiceBusTriggerMetrics> context) { throw null; }
     }
     public partial class ServiceBusTriggerMetrics : Microsoft.Azure.WebJobs.Host.Scale.ScaleMetrics
     {
         public ServiceBusTriggerMetrics() { }
-        public long MessageCount { get { throw null; } set { } }
-        public int PartitionCount { get { throw null; } set { } }
-        public System.TimeSpan QueueTime { get { throw null; } set { } }
+        public long MessageCount { get { throw null; } }
+        public int PartitionCount { get { throw null; } }
+        public System.TimeSpan QueueTime { get { throw null; } }
     }
 }
 namespace Microsoft.Extensions.Hosting
--- a/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/api/Microsoft.Azure.WebJobs.Extensions.ServiceBus.netstandard2.0.cs
+++ b/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/api/Microsoft.Azure.WebJobs.Extensions.ServiceBus.netstandard2.0.cs
@@ -111,6 +111,25 @@ namespace Microsoft.Azure.WebJobs.ServiceBus
         protected internal virtual System.Threading.Tasks.Task CompleteProcessingMessageAsync(Microsoft.Azure.WebJobs.ServiceBus.ServiceBusSessionMessageActions actions, Azure.Messaging.ServiceBus.ServiceBusReceivedMessage message, Microsoft.Azure.WebJobs.Host.Executors.FunctionResult result, System.Threading.CancellationToken cancellationToken) { throw null; }
     }
 }
+namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
+{
+    public partial class ServiceBusScaleMonitor : Microsoft.Azure.WebJobs.Host.Scale.IScaleMonitor, Microsoft.Azure.WebJobs.Host.Scale.IScaleMonitor<Microsoft.Azure.WebJobs.ServiceBus.Listeners.ServiceBusTriggerMetrics>
+    {
+        public ServiceBusScaleMonitor(string functionId, Microsoft.Azure.WebJobs.ServiceBus.ServiceBusEntityType serviceBusEntityType, string entityPath, string connection, Microsoft.Extensions.Logging.ILoggerFactory loggerFactory, Microsoft.Extensions.Configuration.IConfiguration configuration, Microsoft.Extensions.Azure.AzureComponentFactory componentFactory, Microsoft.Azure.WebJobs.ServiceBus.MessagingProvider messagingProvider, Microsoft.Extensions.Azure.AzureEventSourceLogForwarder logForwarder, Microsoft.Extensions.Options.IOptions<Microsoft.Azure.WebJobs.ServiceBus.ServiceBusOptions> options) { }
+        public Microsoft.Azure.WebJobs.Host.Scale.ScaleMonitorDescriptor Descriptor { get { throw null; } }
+        public System.Threading.Tasks.Task<Microsoft.Azure.WebJobs.ServiceBus.Listeners.ServiceBusTriggerMetrics> GetMetricsAsync() { throw null; }
+        public Microsoft.Azure.WebJobs.Host.Scale.ScaleStatus GetScaleStatus(Microsoft.Azure.WebJobs.Host.Scale.ScaleStatusContext<Microsoft.Azure.WebJobs.ServiceBus.Listeners.ServiceBusTriggerMetrics> context) { throw null; }
+        System.Threading.Tasks.Task<Microsoft.Azure.WebJobs.Host.Scale.ScaleMetrics> Microsoft.Azure.WebJobs.Host.Scale.IScaleMonitor.GetMetricsAsync() { throw null; }
+        Microsoft.Azure.WebJobs.Host.Scale.ScaleStatus Microsoft.Azure.WebJobs.Host.Scale.IScaleMonitor.GetScaleStatus(Microsoft.Azure.WebJobs.Host.Scale.ScaleStatusContext context) { throw null; }
+    }
+    public partial class ServiceBusTriggerMetrics : Microsoft.Azure.WebJobs.Host.Scale.ScaleMetrics
+    {
+        public ServiceBusTriggerMetrics() { }
+        public long MessageCount { get { throw null; } set { } }
+        public int PartitionCount { get { throw null; } set { } }
+        public System.TimeSpan QueueTime { get { throw null; } set { } }
+    }
+}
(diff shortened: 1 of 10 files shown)


sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusClientFactory.cs
```diff
@@ -35,7 +35,7 @@ internal class ServiceBusClientFactory
             logForwarder.Start();
         }
 
-        internal ServiceBusClient CreateClientFromSetting(string connection)
+        public ServiceBusClient CreateClientFromSetting(string connection)
```
mathewc, line 38: Are you going to be calling this from ScaleController? Because it seems that ServiceBusScaleMonitor only calls CreateAdministrativeClient internally - I would have thought that ScaleController would only need the APIs provided by the monitor.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusClientFactory.cs
```diff
@@ -14,7 +14,7 @@
 
 namespace Microsoft.Azure.WebJobs.Extensions.ServiceBus.Config
 {
-    internal class ServiceBusClientFactory
+    public class ServiceBusClientFactory
```
mathewc, line 17: For any new public types/methods, we need to have good xml doc.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusClientFactory.cs
JoshLove-msft: +1 on wondering why this needs to be public.

chiangvincent: This was needed because the instantiation of a ServiceBusScaleMonitor requires a ServiceBusReceiver, which we created through a ServiceBusClient. See code pointer [here](https://github.com/chiangvincent/ScaleController/blob/user/vchiang/msiIntegration/ScaleController/Triggers/ServiceBusTrigger.cs#:~:text=public%20override%20IScaleMonitor%20GetScaleMonitor()). Not sure if this is the best way to do this -- open to suggestions here

JoshLove-msft: I'd rather reduce the number of dependencies that the ServiceBusScaleMonitor requires. It looks like it basically needs the IConfiguration instance and then it could handle everything else internally, i.e. creating the ClientFactory, Receiver, AdminClient. We may still need to support the old flow where the ClientFactory is injected in the extensions, but that could be left as an internal constructor. For the public constructor, we should try to reduce dependencies.

JoshLove-msft:
We should make sure to address https://github.com/Azure/azure-sdk-for-net/issues/29536 before moving ahead with the switch in the scale controller.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusClientFactory.cs
chiangvincent: Created new public constructor for ServiceBusScaleMonitor, and removed changes for ServiceBusClientFactory (based on offline conversation). Let me know if this looks good!

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusClientFactory.cs
chiangvincent: Keeping the ServiceBusClientFactory internal, as discussed with @JoshLove-msft offline

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusTriggerMetrics.cs
```diff
@@ -8,7 +8,7 @@
 
 namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
 {
-    internal class ServiceBusTriggerMetrics : ScaleMetrics
+    public class ServiceBusTriggerMetrics : ScaleMetrics
```
JoshLove-msft, line 11: Need class level docs.

chiangvincent: Done

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusScaleMonitor.cs
```diff
@@ -42,6 +46,21 @@ public ServiceBusScaleMonitor(string functionId, ServiceBusEntityType serviceBus
             _nextWarningTime = DateTime.UtcNow;
         }
 
+        public ServiceBusScaleMonitor(string functionId, ServiceBusEntityType serviceBusEntityType, string entityPath, string connection, ILoggerFactory loggerFactory, IConfiguration configuration, AzureComponentFactory componentFactory, MessagingProvider messagingProvider, AzureEventSourceLogForwarder logForwarder, IOptions<ServiceBusOptions> options)
```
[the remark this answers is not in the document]
chiangvincent: Done

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusScaleMonitor.cs
```diff
@@ -14,7 +17,7 @@
 
 namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
 {
-    internal class ServiceBusScaleMonitor : IScaleMonitor<ServiceBusTriggerMetrics>
+    public class ServiceBusScaleMonitor : IScaleMonitor<ServiceBusTriggerMetrics>
```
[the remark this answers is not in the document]
chiangvincent: Done

JoshLove-msft:
You'll need to regenerate and commit the API surface area. See https://github.com/Azure/azure-sdk-for-net/blob/main/CONTRIBUTING.md#public-api-additions.

JoshLove-msft (approved):

chiangvincent:
Waiting to merge in Alexey's PR at https://github.com/Azure/azure-sdk-for-net/pull/30822/files#diff-e2d09bf3bc95b6b03e0ddc0f67ea411f770344d9f1f598cd2dd9fad0bd0aa847 before we merge this change

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusScaleMonitor.cs
```diff
@@ -42,20 +49,52 @@ public ServiceBusScaleMonitor(string functionId, ServiceBusEntityType serviceBus
...
+            _serviceBusEntityType = serviceBusEntityType;
+            _entityPath = entityPath;
+            _scaleMonitorDescriptor = new ScaleMonitorDescriptor($"{_functionId}-ServiceBusTrigger-{_entityPath}".ToLower(CultureInfo.InvariantCulture));
+            _isListeningOnDeadLetterQueue = entityPath.EndsWith(DeadLetterQueuePath, StringComparison.OrdinalIgnoreCase);
+            _logger = loggerFactory.CreateLogger<ServiceBusScaleMonitor>();
+            _nextWarningTime = DateTime.UtcNow;
+            _clientFactory = new Lazy<ServiceBusClientFactory>(() => new ServiceBusClientFactory(configuration, componentFactory, messagingProvider, logForwarder, options));
+            _administrationClient = new Lazy<ServiceBusAdministrationClient>(() => _clientFactory.Value.CreateAdministrationClient(connection));
+            _client = new Lazy<ServiceBusClient>(() => _clientFactory.Value.CreateClientFromSetting(connection));
+            _receiver = new Lazy<ServiceBusReceiver>(() => _client.Value.CreateReceiver(entityPath));
```
mathewc, line 77: Looks like this receiver is being created differently than the previous code [here](https://github.com/Azure/azure-sdk-for-net/blob/8947cc941137289363569dbeef191cb0e7b911e4/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusListener.cs#L78) which was going through the MessagingProvider to create a batch receiver, and was also ensuring ServiceBusOptions are applied. Why the difference?

JoshLove-msft: Good catch, we should make sure the receivers are created in the same way in both places. The point of this second constructor is that we can avoid making ServiceBusClientFactory public.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusTriggerMetrics.cs
```diff
@@ -8,22 +8,25 @@
...
+    /// <summary>
+    /// Trigger metrics class for Service Bus. Used to make scaling decisions in <see cref="ServiceBusScaleMonitor"/>.
+    /// </summary>
+    public class ServiceBusTriggerMetrics : ScaleMetrics
     {
         /// <summary>
         /// The number of messages currently in the queue/topic.
         /// </summary>
-        public long MessageCount { get; set; }
+        public long MessageCount { get; internal set; }
```
mathewc, line 19: I think the setters should be public. Kind of strange (and not testable) to have a public data type like this that you can instantiate but not provide values for. Also note that the base class property has a public setter.

JoshLove-msft: Normally what we do for output types like this is make the constructor internal and provide a model factory for testing. I missed that the base class already has a setter, so I would agree that we can just revert this change.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusScaleMonitor.cs
```diff
@@ -42,20 +49,52 @@ public ServiceBusScaleMonitor(string functionId, ServiceBusEntityType serviceBus
...
+        }
+
+        ScaleMonitorDescriptor IScaleMonitor.Descriptor
```
mathewc, line 80: Why the change here from public to explicit interface implementation? Can't it remain public?

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusScaleMonitor.cs
```diff
@@ -42,20 +49,52 @@ public ServiceBusScaleMonitor(string functionId, ServiceBusEntityType serviceBus
...
             }
         }
-
         async Task<ScaleMetrics> IScaleMonitor.GetMetricsAsync()
         {
             return await GetMetricsAsync().ConfigureAwait(false);
         }
 
-        public async Task<ServiceBusTriggerMetrics> GetMetricsAsync()
+        async Task<ServiceBusTriggerMetrics> IScaleMonitor<ServiceBusTriggerMetrics>.GetMetricsAsync()
```
mathewc, line 92: Same question here on explicit interface impl. Can't it stay public?

JoshLove-msft:
> Waiting to merge in Alexey's PR at https://github.com/Azure/azure-sdk-for-net/pull/30822/files#diff-e2d09bf3bc95b6b03e0ddc0f67ea411f770344d9f1f598cd2dd9fad0bd0aa847 before we merge this change

How does Alexey's PR fit with these changes? It looks like we are changing the interface that we implement there. Are we going to reflect that in this PR?

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/api/Microsoft.Azure.WebJobs.Extensions.ServiceBus.netstandard2.0.cs
```diff
@@ -111,6 +111,25 @@ public partial class SessionMessageProcessor
...
+        Microsoft.Azure.WebJobs.Host.Scale.ScaleStatus Microsoft.Azure.WebJobs.Host.Scale.IScaleMonitor.GetScaleStatus(Microsoft.Azure.WebJobs.Host.Scale.ScaleStatusContext context) { throw null; }
+        System.Threading.Tasks.Task<Microsoft.Azure.WebJobs.ServiceBus.Listeners.ServiceBusTriggerMetrics> Microsoft.Azure.WebJobs.Host.Scale.IScaleMonitor<Microsoft.Azure.WebJobs.ServiceBus.Listeners.ServiceBusTriggerMetrics>.GetMetricsAsync() { throw null; }
+        Microsoft.Azure.WebJobs.Host.Scale.ScaleStatus Microsoft.Azure.WebJobs.Host.Scale.IScaleMonitor<Microsoft.Azure.WebJobs.ServiceBus.Listeners.ServiceBusTriggerMetrics>.GetScaleStatus(Microsoft.Azure.WebJobs.Host.Scale.ScaleStatusContext<Microsoft.Azure.WebJobs.ServiceBus.Listeners.ServiceBusTriggerMetrics> context) { throw null; }
+    }
+    public partial class ServiceBusTriggerMetrics : Microsoft.Azure.WebJobs.Host.Scale.ScaleMetrics
+    {
+        public ServiceBusTriggerMetrics() { }
+        public long MessageCount { get { throw null; } }
+        public int PartitionCount { get { throw null; } }
+        public System.TimeSpan QueueTime { get { throw null; } }
```
JoshLove-msft, line 130: We may want to think about a better name for this property. What do you think of TimeOnQueue?

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusScaleMonitor.cs
```diff
@@ -42,20 +49,52 @@ public ServiceBusScaleMonitor(string functionId, ServiceBusEntityType serviceBus
             _nextWarningTime = DateTime.UtcNow;
         }
 
-        public ScaleMonitorDescriptor Descriptor
+        /// <summary>
+        /// Public constructor used to retrieve metrics and make scale decisions for Service Bus entities.
```
JoshLove-msft, line 53:
```suggestion
        /// Initializes an instance of ServiceBusScaleMonitor, which is used to retrieve metrics and make scale decisions for Service Bus entities.
```

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusTriggerMetrics.cs
```diff
@@ -8,22 +8,25 @@
...
 
         /// <summary>
         /// The number of partitions.
         /// </summary>
-        public int PartitionCount { get; set; }
+        public int PartitionCount { get; internal set; }
 
         /// <summary>
         /// The length of time the next message has been
```
JoshLove-msft, line 27:
```suggestion
        /// The length of time the top message on the queue has been enqueued.
```

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/api/Microsoft.Azure.WebJobs.Extensions.ServiceBus.netstandard2.0.cs
mathewc: There would be some complexity in changing these names at this point, since this is a serialization contract. These values are persisted and serialized/deserialized by hosts (e.g. Functions Host) and passed back into the monitor via GetScaleStatus.

JoshLove-msft: Ok, let's see if we can get this approved with the current name.

PR #30642, opened 2022-08-19
Refactor statsbeat

vishweshbankwar:
Refactoring to move initialization from static constructor to a static method.


rajkumar-rangaraj (approved):

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/Statsbeat/Statsbeat.cs
```diff
@@ -106,11 +121,24 @@ private static VmMetadataResponse GetVmMetadataResponse()
             }
             catch (Exception ex)
             {
-                AzureMonitorExporterEventSource.Log.WriteWarning("Failed to get VM metadata details", ex);
+                AzureMonitorExporterEventSource.Log.WriteInformational("Failed to get VM metadata details", ex.ToInvariantString());
```
TimothyMothra, line 124: I wouldn't do the `ToInvariantString` here.
Better to add a new `NonEvent` method to the EventSource class to handle this.
The main benefit is to potentially skip the work if noone is listening to the eventsource.

vishweshbankwar: Will fix this.

TimothyMothra: On second thought, maybe we leave this as Warning. :)

1. The original idea was that Informational logs wouldn't contain exceptions.
2. The AzureMetaData service won't be available to non-Azure environments. 
    If we get an Exception in Azure, the warning would be valid.
    Outside of Azure, I think the warning is noise which is why I originally suggested using Informational.
    
@rajkumar-rangaraj what do you think?

cijothomas: > The main benefit is to potentially skip the work if noone is listening to the eventsource.

^ This code is executed only once, right?

PR #30646, opened 2022-08-20
[Experiment] Prototype of multi-version protocol client

annelo-msft:
Initial prototype of simple multi-version client described in https://github.com/Azure/autorest.csharp/issues/2609

This includes:
- ServiceVersion enum is generated from versions specified in Cadl
- ServiceVersion provided in ClientOptions is used to create requests
- NotSupportedException is thrown from service methods that weren't defined in the service API corresponding to the ServiceVersion provided in ClientOptions
- Documentation comments include "Added in \<service-version\>" for operations and parameters
- Generator fails if operations don't support api-version parameter

The Cadl file describing this client is here: [main-v2client.cadl](https://github.com/annelo-msft/cadl-ranch/blob/cadl-test-multiversion-addop/packages/cadl-ranch-specs/http/versioning/operations/main-v2client.cadl), and is explained in [this README](https://github.com/Azure/cadl-ranch/blob/f0f23451ef801573a51f3b6dd019e9b590efdcc8/packages/cadl-ranch-specs/http/versioning/operations/README.md).


sdk/template/Azure.Template/src/Generated/MultiVersionClient.cs
```diff
@@ -0,0 +1,237 @@
...
+                return _pipeline.ProcessMessage(message, context);
+            }
+            catch (Exception e)
+            {
+                scope.Failed(e);
+                throw;
+            }
+        }
+
+        /// <summary>Servicev2Operation method. Added in service version 2022-02-02. </summary>
```
annelo-msft, line 130: Doc comment includes version added.

sdk/template/Azure.Template/src/Generated/MultiVersionClient.cs
```diff
@@ -0,0 +1,237 @@
...
+        /// Response response = await client.ServiceV2OperationAsync();
+        /// 
+        /// JsonElement result = JsonDocument.Parse(response.ContentStream).RootElement;
+        /// Console.WriteLine(result.ToString());
+        /// ]]></code>
+        /// </example>
+        public virtual async Task<Response> ServiceV2OperationAsync(RequestContext context = null)
+        {
+            // Generator knows from Cadl @added decorator which version an operation was added in.
+            if (_serviceVersion < MultiVersionClientOptions.ServiceVersion.V2022_02_02)
```
annelo-msft, line 150: Throw NotSupportedException if calling from client created for service version that does not support the operation.

sdk/template/Azure.Template/src/Generated/MultiVersionClient.cs
```diff
@@ -0,0 +1,237 @@
...
+
+        internal HttpMessage CreateServiceV1OperationRequest(RequestContext context)
+        {
+            var message = _pipeline.CreateMessage(context, ResponseClassifier200);
+            var request = message.Request;
+            request.Method = RequestMethod.Get;
+            var uri = new RawRequestUriBuilder();
+            uri.Reset(_endpoint);
+            uri.AppendPath("/ops/v1op", false);
+            uri.AppendQuery("api-version", _apiVersion, true);
```
annelo-msft, line 214: Validate that api-version is added to request (fail generator if operation doesn't support api-version parameter)

sdk/template/Azure.Template/src/Generated/MultiVersionClientOptions.cs
```diff
@@ -0,0 +1,44 @@
...
+
+namespace Azure.Template
+{
+    /// <summary> Client options for MultiVersionClient. </summary>
+    public partial class MultiVersionClientOptions : ClientOptions
+    {
+        private const ServiceVersion LatestVersion = ServiceVersion.V2022_02_02;
+
+        /// <summary> The version of the service to use. </summary>
+        public enum ServiceVersion
```
annelo-msft, line 19: ServiceVersion enum contains all service versions

sdk/template/Azure.Template/src/Generated/MultiVersionClientOptions.cs
```diff
@@ -0,0 +1,44 @@
...
+        {
+            /// <summary> Service version "2022-01-01". </summary>
+            V2022_01_01 = 1,
+            /// <summary> Service version "2022-02-02". </summary>
+            V2022_02_02 = 2,
+        }
+
+        internal ServiceVersion TargetVersion { get; }
```
annelo-msft, line 27: internal Version stored as enum to enable using comparison operators to test whether an API is supported by target service version.

sdk/template/Azure.Template/src/Generated/MultiVersionClient.cs
```diff
@@ -0,0 +1,237 @@
...
+namespace Azure.Template
+{
+    /// <summary> The MultiVersion service client. </summary>
+    public partial class MultiVersionClient
+    {
+        private static readonly string[] AuthorizationScopes = new string[] { "https://vault.azure.net/.default" };
+        private readonly TokenCredential _tokenCredential;
+        private readonly HttpPipeline _pipeline;
+        private readonly Uri _endpoint;
+        private readonly string _apiVersion;
```
KrzysztofCwalina, line 23: why do we store both version enum and string?

sdk/template/Azure.Template/src/Generated/MultiVersionClient.cs
```diff
@@ -0,0 +1,237 @@
...
+        /// var endpoint = new Uri("<https://my-service.azure.com>");
+        /// var client = new MultiVersionClient(endpoint, credential);
+        /// 
+        /// Response response = await client.ServiceV2OperationAsync();
+        /// 
+        /// JsonElement result = JsonDocument.Parse(response.ContentStream).RootElement;
+        /// Console.WriteLine(result.ToString());
+        /// ]]></code>
+        /// </example>
+        public virtual async Task<Response> ServiceV2OperationAsync(RequestContext context = null)
```
AlexanderSher, line 147: I'd prefer names of operations to be `Foo` and `Bar` rather than `ServiceV1` and `ServiceV2`, cause right now it looks like these methods are versions of the same operation rather than a completely separate operation added in V2.

sdk/template/Azure.Template/src/Generated/MultiVersionClient.cs
annelo-msft: The string is needed to add api-version to requests, and the enum is needed for comparison of enum values for version checks, e.g. here: https://github.com/Azure/azure-sdk-for-net/pull/30646/files#diff-54256e001b630987030f146f058babafde8e370adcfae19d8c84be16a788e5baR150

My thinking in storing both was that we could keep just the enum and get the string from it (but not vice versa), but that puts a string allocation on a hot path, i.e. request creation.  If we don't consider request creation to be on a hot path, we could store the enum instead, but this would remove an internal field on service-specific client options, which would require significant refactoring across libraries.

That's how I got to this decision, but I'm open to feedback - I'm not wild about this duplication.

sdk/template/Azure.Template/src/Generated/MultiVersionClient.cs
annelo-msft: That makes sense - I can update the example so we don't introduce that confusion.  Thanks, @AlexanderSher!

sdk/template/Azure.Template/src/Generated/MultiVersionClient.cs
```diff
@@ -0,0 +1,237 @@
...
+            _serviceVersion = options.TargetVersion;
+        }
+
+        /// <param name="context"> The request context, which can override default behaviors of the client pipeline on a per-call basis. </param>
+        /// <exception cref="RequestFailedException"> Service returned a non-success status code. </exception>
+        /// <returns> The response returned from the service. </returns>
+        /// <example>
+        /// This sample shows how to call ServiceV1OperationAsync and parse the result.
+        /// <code><![CDATA[
+        /// var credential = new DefaultAzureCredential();
```
heaths, line 72: Nit/question: to date, we've been avoiding `var` except when obvious e.g., constructors. Given the previous enforcer has left, we might want to revisit. I've always liked var, though, personally, I've started liking the `Class local = new(...)` syntax, but that's probably too new to use for examples and would confuse people.

sdk/template/Azure.Template/src/Generated/MultiVersionClient.cs
heaths: What about *also* adding an attribute so that we could, for example, create an analyzer customers could use to help watch for these at dev time? The attribute would only be a marker, though. I'm not advocating that we use them at runtime - in fact, we shouldn't: reflection is slow.

sdk/template/Azure.Template/src/Generated/MultiVersionClient.cs
heaths: Given a recent DPG meeting, I did want to point out that - even if we relegate errors to the service for model properties - I don't think we can/should for endpoints. Some services, like Key Vault, return a 404 along with the default ASP.NET error *HTML* page for unmapped endpoints, which is less than ideal and can be confusing. And if we're validating service versions for some things, why not consider it for everything? Why incur network IO if we know something won't work or, worse, work incorrectly if a service doesn't handle PATCH-like semantics correctly.

/cc @KrzysztofCwalina @tg-msft

PR #30657, opened 2022-08-22
Feature/call automation unit tests

yyw-msft:
- added unit tests for CallAutomationClient class
- added unit tests for CallConnection class

### DIFF
--- a/sdk/communication/Azure.Communication.CallingServer/tests/CallRecording/CallRecordingTests.cs
+++ b/sdk/communication/Azure.Communication.CallingServer/tests/CallRecording/CallRecordingTests.cs
@@ -12,7 +12,7 @@ namespace Azure.Communication.CallingServer
     public class CallRecordingTests : CallAutomationTestBase
     {
         private const string RecordingId = "sampleRecordingId";
-        private const string ServerCallId = "sampleServerCallId";
+        private new const string ServerCallId = "sampleServerCallId";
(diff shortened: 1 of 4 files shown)


sdk/communication/Azure.Communication.CallingServer/tests/CallRecording/CallRecordingTests.cs
```diff
@@ -12,7 +12,7 @@ namespace Azure.Communication.CallingServer
     public class CallRecordingTests : CallAutomationTestBase
     {
         private const string RecordingId = "sampleRecordingId";
-        private const string ServerCallId = "sampleServerCallId";
+        private new const string ServerCallId = "sampleServerCallId";
```
cochi2, line 15: question: what does this `new` keyword do here?

yyw-msft: i have a ServerCallId defined in the CallAutomationTestBase class used for CallConnectionTests and CallAutomationClientTests classes. this is to override the field in the CallRecordingTests. Should i remove this field in CallRecordingTests and just use the one in the base class instead?

cochi2: I think that it should work with your own values. Just get rid of it :)

cochi2 (approved):

PR #30674, opened 2022-08-23
[ApiManagement] Fix 30572

### LINKED ISSUE #30572: [ApiManagement] Track 2 code issues
We have hit couple of the issues during the testing of ApiManagement track 2 mgmt plane code, below are the details:

1. Id of content type is not a resource identifier

    According to the [swagger definition](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/apimanagement/resource-manager/Microsoft.ApiManagement/stable/2021-08-01/definitions.json#L6467-L6481), content type inherits from Resource, which means its Id should be a resource identifier, but the actual backend service doesn't return a resource identifier.
image

2. The Uri of PortalSettingsContractData may return empty string

    In the deserialization, we only have the check about null, it means the empty string will pass the check. And when we use the value to instantiate Uri, we will get exception

Yao725:
Fix #30572


sdk/apimanagement/Azure.ResourceManager.ApiManagement/src/Customized/ApiManagementUserResource.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Globalization;
+using Azure.Core;
+
+namespace Azure.ResourceManager.ApiManagement
+{
+    public partial class ApiManagementUserResource
+    {
+        private static readonly ResourceType ResourceTypeWithGroup = "Microsoft.ApiManagement/service/groups/users";
+
+        internal static void ValidateResourceId(ResourceIdentifier id)
```
Yao725, line 16: Use this kind of customized method to make sure the verification succeeds when calling the method [CreateGroupUserAsync](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/apimanagement/Azure.ResourceManager.ApiManagement/src/Generated/ApiManagementGroupResource.cs#L412)

sdk/apimanagement/Azure.ResourceManager.ApiManagement/src/autorest.md
```diff
@@ -304,6 +313,8 @@ directive:
       $.PortalSettingsContractProperties.properties.subscriptions['x-ms-client-name'] = 'IsSubscriptions';
       $.PortalSettingsContractProperties.properties.userRegistration['x-ms-client-name'] = 'IsUserRegistration';
       $.PrivateEndpointConnectionRequest.properties.id['x-ms-format'] = 'arm-id';
+      delete $.ContentTypeContract.allOf;
```
Yao725, line 316: Because the id of content type and content item is not resource identifier, therefore here we delete the `allof` to prevent the model from inheriting `ResourceData`.

sdk/apimanagement/Azure.ResourceManager.ApiManagement/api/Azure.ResourceManager.ApiManagement.netstandard2.0.cs
```diff
@@ -1473,9 +1394,17 @@ public partial class ApiManagementServiceResource : Azure.ResourceManager.ArmRes
...
+        public virtual System.Threading.Tasks.Task<Azure.Response<Azure.ResourceManager.ApiManagement.Models.ApiManagementContentItem>> CreateOrUpdateContentItemAsync(string contentTypeId, string contentItemId, string ifMatch = null, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
+        public virtual Azure.Response<Azure.ResourceManager.ApiManagement.Models.ApiManagementContentType> CreateOrUpdateContentType(string contentTypeId, string ifMatch = null, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
+        public virtual System.Threading.Tasks.Task<Azure.Response<Azure.ResourceManager.ApiManagement.Models.ApiManagementContentType>> CreateOrUpdateContentTypeAsync(string contentTypeId, string ifMatch = null, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
```
ArcturusZhang, line 1400: Should the ifMatch here be of type `ETag`?

sdk/apimanagement/Azure.ResourceManager.ApiManagement/api/Azure.ResourceManager.ApiManagement.netstandard2.0.cs
```diff
@@ -1473,9 +1394,17 @@ public partial class ApiManagementServiceResource : Azure.ResourceManager.ArmRes
...
         public virtual System.Threading.Tasks.Task<Azure.ResourceManager.ArmOperation> DeleteAsync(Azure.WaitUntil waitUntil, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
+        public virtual Azure.Response DeleteContentItem(string contentTypeId, string contentItemId, string ifMatch, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
+        public virtual System.Threading.Tasks.Task<Azure.Response> DeleteContentItemAsync(string contentTypeId, string contentItemId, string ifMatch, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
+        public virtual Azure.Response DeleteContentType(string contentTypeId, string ifMatch, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
+        public virtual System.Threading.Tasks.Task<Azure.Response> DeleteContentTypeAsync(string contentTypeId, string ifMatch, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
```
ArcturusZhang, line 1407: Same, should the ifMatch be `ETag`?

sdk/apimanagement/Azure.ResourceManager.ApiManagement/src/Customized/ApiResource.cs
```diff
@@ -0,0 +1,31 @@
...
+using Azure.ResourceManager;
+using Azure.ResourceManager.ApiManagement.Models;
+
+namespace Azure.ResourceManager.ApiManagement
+{
+    public partial class ApiResource
+    {
+        private static readonly ResourceType ResourceTypeWithGateway = "Microsoft.ApiManagement/service/gateways/apis";
+
+        private static readonly ResourceType ResourceTypeWithProduct = "Microsoft.ApiManagement/service/products/apis";
```
ArcturusZhang, line 23: I am kind of confused about this resource.
The static property of this resource will always return `Microsoft.ApiManagement/service/apis`, does this mean that when I call this API
`/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.ApiManagement/service/{serviceName}/apis/{apiId}`, sometimes it returns an ID of `Microsoft.ApiManagement/service/apis`, sometimes it does not and returns `Microsoft.ApiManagement/gateways/apis`?

ArcturusZhang: And our `CreateResourceIdentifier` method in this resource will also always return an ID of type `Microsoft.ApiManagement/service/apis`

sdk/apimanagement/Azure.ResourceManager.ApiManagement/src/autorest.md
ArcturusZhang: Therefore as a result, we lose the `type` property on this model. Is this expected or a workaround?

sdk/apimanagement/Azure.ResourceManager.ApiManagement/api/Azure.ResourceManager.ApiManagement.netstandard2.0.cs
Yao725: Based on the description, I changed the type to `ETag`.

sdk/apimanagement/Azure.ResourceManager.ApiManagement/src/Customized/ApiResource.cs
Yao725: If you call `/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.ApiManagement/service/{serviceName}/apis/{apiId}`, it returns an ID of `Microsoft.ApiManagement/service/apis`. But we can also add the created api to a gateway or a product, in this case, the response still represents an `ApiResource`, but the id will be [/subscriptions/bab08e11-7b12-4354-9fd1-4b5d64d40b68/resourceGroups/Api-Default-CentralUS/providers/Microsoft.ApiManagement/service/sdktestservice/gateways/gatewayid8779/apis/echo-api](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/apimanagement/Microsoft.Azure.Management.ApiManagement/tests/SessionRecords/GatewayTests/CreateListUpdateDelete.json#L667). That's the reason that I extended the scope of this validation.

sdk/apimanagement/Azure.ResourceManager.ApiManagement/src/autorest.md
Yao725: I believe this is just a workaround. I didn't find a way to prevent this model from inheriting `ResourceData`, therefore I decided to delete it. Once we have a better way to support this case, we can add it back as there is no breaking change.

ArcturusZhang: Could you please open an issue in atuorest.csharp? We might need a new configuration, or extend some existing configuration to support this.

Yao725: Opened here: https://github.com/Azure/autorest.csharp/issues/2639

sdk/apimanagement/Azure.ResourceManager.ApiManagement/src/Customized/ApiResource.cs
Yao725: We don't have this kind of resource, and I didn't test the scenario that you mentioned but I think it might work. Actually I think the better choice here is to return the data instead of introducing this kind of customization.

sdk/apimanagement/Azure.ResourceManager.ApiManagement/src/Customized/Models/ApiManagementContentType.cs
```diff
@@ -29,15 +30,27 @@ public ApiManagementContentTypeData()
...
         }
 
+        /// <summary> Content Type ID. </summary>
+        public string ContentTypeId { get; }
+        /// <summary> The name of the resource. </summary>
+        public string Name { get; }
+        /// <summary> The type of the resource. E.g. &quot;Microsoft.Compute/virtualMachines&quot; or &quot;Microsoft.Storage/storageAccounts&quot;. </summary>
+        public ResourceType ResourceType { get; }
+        /// <summary> Azure Resource Manager metadata containing createdBy and modifiedBy information. </summary>
+        public ResourceManager.Models.SystemData SystemData { get; }
```
ArcturusZhang, line 53: Just emphasizing this for you to double check - does it really have `systemData`?
The systemData might come from our ResourceData

sdk/apimanagement/Azure.ResourceManager.ApiManagement/src/autorest.md
```diff
@@ -321,6 +330,162 @@ directive:
...
+              "value": {
+                  "type": "array",
+                  "items": {
+                      "x-ms-client-name": "ApiManagementGroupUserData",
+                      "type": "object",
+                      "allOf": [
+                          {
+                              "$ref": "./definitions.json#/definitions/UserContract"
+                          }
+                      ]
```
ArcturusZhang, line 351: Why not just using a $ref here?

sdk/apimanagement/Azure.ResourceManager.ApiManagement/src/autorest.md
```diff
@@ -321,6 +330,162 @@ directive:
     where: $.parameters
     transform: >
       $.OpenIdConnectIdParameter['x-ms-client-name'] = 'OpenId';
+      $.IfMatchOptionalParameter['x-ms-format'] = 'etag';
+      $.IfMatchRequiredParameter['x-ms-format'] = 'etag';
+  - from: apimgroups.json
```
ArcturusZhang, line 335: These directives are confusing, could you please add some comments here to describe what are these things for?

ArcturusZhang (approved):

sdk/apimanagement/Azure.ResourceManager.ApiManagement/src/Customized/Models/ApiManagementContentType.cs
Yao725: Thanks for the reminder, double checked and removed this property.

sdk/apimanagement/Azure.ResourceManager.ApiManagement/src/autorest.md
Yao725: The original swagger is just using a $ref here, this is going to cause the return type of this method to become the corresponding resource, but we found out that the returned id will not match the type of the resource, therefore I use an allOf to force a new model to be generated and thus bypass the act of wrapping the data model as a resource.

sdk/apimanagement/Azure.ResourceManager.ApiManagement/src/autorest.md
Yao725: Added.

PR #30676, opened 2022-08-23
Add the azurite support

ArthurMa1978:


sima-zhu (approved):

benbp:
@sima-zhu @ArthurMa1978  Why do the resource manager tests require the storage emulator?

weshaggard:
> @sima-zhu @ArthurMa1978 Why do the resource manager tests require the storage emulator?

It is because the resourcemanager is going to start running any tests depend on it. The storage tests currently have a dependency on it so they are included. This is similar to how the core test pipeline also needs the storage emulator. However with that said this will essentially install the emulator for all the dependent tests and not just the storage tests which we should think about. This particular test dependency is an interesting one we might need to rethink if there become other tests with similar dependencies. Cosmos comes to mind but Cosmos is not in the mono .NET repo currently but will likely be an issue when we do this for Java.

benbp:
Yeah I think this could get hard to manage pretty quick if there are a lot of dependent tests with custom setup steps that are configured separately. I think we could structure service specific pre and post steps to be in their own step template file so we can more easily include them. Otherwise service owners could change their custom test steps without realizing they need to change resource manager tests, for example.

PR #30677, opened 2022-08-23
Address review comments for MySql, MediaServices

### LINKED ISSUE #30563: [BUG] Naming of Media Services objects needs to be adjusted
### Library name and version

Azure.ResourceManager.Media

### Describe the bug

The naming of the objects in Azure.ResourceManager.Media does not match the service naming. 

One annoyance is that the library mixes different cases for Media Services (likely due to some legacy issues with the Swagger file), for example:

The “GetMediaserviceResource” uses a lower case “s” in “services” - it should be GetMediaServicesResource
The “Azure.ResourceManager.Media.MediaServiceData” class uses an upper case “S” in “Services” - which is correct. 
 

Could all of these be updated to use an upper-case “S” in “Services” to be consistent across all uses? That would also match to the product name which is the plural form - Azure Media Services 

### Expected behavior

Naming of objects should be consistent with the service naming of Azure Media Services. 
Use MediaServices and not Mediaservices 
Use plural form and not MediaService - unless referring to a single account in Media Services.

### Actual behavior

See above 

### Reproduction Steps

See above 

### Environment

_No response_

ArthurMa1978:


sdk/mysql/Azure.ResourceManager.MySql/api/Azure.ResourceManager.MySql.netstandard2.0.cs
```diff
@@ -493,8 +493,8 @@ public partial class MySqlServerResource : Azure.ResourceManager.ArmResource
...
-        public virtual Azure.ResourceManager.ArmOperation<Azure.ResourceManager.MySql.Models.MySqlConfigurationList> UpdateConfigurations(Azure.WaitUntil waitUntil, Azure.ResourceManager.MySql.Models.MySqlConfigurationList value, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
-        public virtual System.Threading.Tasks.Task<Azure.ResourceManager.ArmOperation<Azure.ResourceManager.MySql.Models.MySqlConfigurationList>> UpdateConfigurationsAsync(Azure.WaitUntil waitUntil, Azure.ResourceManager.MySql.Models.MySqlConfigurationList value, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
+        public virtual Azure.ResourceManager.ArmOperation<Azure.ResourceManager.MySql.Models.MySqlConfigurationListResult> UpdateConfigurations(Azure.WaitUntil waitUntil, Azure.ResourceManager.MySql.Models.MySqlConfigurationListResult value, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
```
fengzhou-msft, line 496: This model is used as input as well, the `Result` suffix may be weird for a input type.

ArthurMa1978: fixed

fengzhou-msft (approved):

PR #30680, opened 2022-08-23
Longer wait for resource DNS propagation

christothes:

### DIFF
--- a/sdk/confidentialledger/test-resources-post.ps1
+++ b/sdk/confidentialledger/test-resources-post.ps1
@@ -1,3 +1,3 @@
 # Confidential Ledger resources can take a few minutes to resolve in DNS after being newly provisioned.
 Write-Host 'Waiting to let the new resource propagate to DNS'
-Start-Sleep -Seconds 240
+Start-Sleep -Seconds 360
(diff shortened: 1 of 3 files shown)


sdk/confidentialledger/test-resources-post.ps1
```diff
@@ -1,3 +1,3 @@
-# Confidential Ledger resources can take up to 60 seconds to resolve in DNS after being newly provisioned.
-Write-Host 'Waiting 60 seconds to let the new resource propagate to DNS'
-Start-Sleep -Seconds 60
+# Confidential Ledger resources can take a few minutes to resolve in DNS after being newly provisioned.
```
pallavit, line 1: Should we consider using shared resource instead of provisioning it every time? Assuming CL resources are not that costly to keep alive.

pallavit: Cognitive services (TA, FR) use shared resources by adding the secrets directly to the pipeline and test-resources is basically empty. Not saying that is the best solution here since we are only waiting a couple of mins but it may be worth considering if this does not fix things either.

JoshLove-msft: We shouldn't use static resources unless we absolutely need to. It makes it harder for contributors and it introduces additional risks.

https://dev.azure.com/azure-sdk/internal/_wiki/wikis/internal.wiki/365/Static-Test-Resources

JoshLove-msft: If you want to avoid hardcoding a wait time, you can look into overriding this method - https://github.com/Azure/azure-sdk-for-net/blob/0d44555ba2e14692ba7722864690c69e548c666d/sdk/core/Azure.Core.TestFramework/src/TestEnvironment.cs#L235

JoshLove-msft (approved):

PR #30700, opened 2022-08-24
Fix receive and peek scope instrumentation

### LINKED ISSUE #30656: [FEATURE REQ] ServiceBus receiver spans should be correlated to producer
### Library name

Azure.Messaging.ServiceBus

ServiceBus Receiver client does not correlate receive/settle activities to message context. 
We should do it in the following way:
- Batch receive: add links to contexts of received messages. Links could only be added at the start time, so we should create receive span after messages are received, but set the correct start time
- Async iterator: create receive span per message and add single link to the received context
- Settle:  create span per message  and add a single link to the context in the message

Related issue: https://github.com/microsoft/ApplicationInsights-dotnet/issues/2656

JoshLove-msft:
Fixes https://github.com/Azure/azure-sdk-for-net/issues/30656


lmolkova (approved):
Looks great, thank you!

PR #30712, opened 2022-08-24
Remove double-slash from default scope for non-public clouds

### LINKED ISSUE #29891: [QUERY] ARM scopes in non-public clouds
### Library name and version

Azure.Identity

### Query/Question

In Azure.Identity , AzureAuthorityHosts.cs lists the default scopes to use for ARM in various clouds:
https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/identity/Azure.Identity/src/AzureAuthorityHosts.cs#L47

Note that these scopes all have a double slash.

However, in Azure.ResourceManager, we have a list of ARM endpoints and their audiences (derived from ARM metadata).
https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/resourcemanager/Azure.ResourceManager/src/ArmEnvironment.cs

Note that only the public cloud audience has a trailing slash, which will result in the double slash for DefaultScope.

One of these probably needs to be changed. I'm inclined to go with ArmEnvironment.cs since it's derived from ARM metadata: i.e. only AzurePublicCloud should have the double-slash.

I can also believe that both forms work in the known clouds, but I recommend we be consistent.

### Environment

_No response_

pharring:
Fixes #29891

### DIFF
--- a/sdk/identity/Azure.Identity/src/AzureAuthorityHosts.cs
+++ b/sdk/identity/Azure.Identity/src/AzureAuthorityHosts.cs
@@ -50,13 +50,13 @@ namespace Azure.Identity
             {
                 case AzurePublicCloudHostUrl:
                     // The double slash is intentional for public cloud.
-                    return "https://management.core.windows.net//.default";
+                    return "https://management.azure.com//.default";
                 case AzureChinaHostUrl:
-                    return "https://management.core.chinacloudapi.cn/.default";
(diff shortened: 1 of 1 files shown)


pharring:
With this change, the values in Azure.Identity are the same as in Azure.ResourceManager

schaabs (approved):
Thanks for the fix @pharring! LGTM

PR #30730, opened 2022-08-25
Onboard the Alpha Ids SDK

guilhermeluizsp:
**NOTE: This is a private SDK intended for internal use only. This SDK will not be publicly supported**

This is the onboarding of a new internal SDK used by the Azure Communication Services team. This is an important piece of software that will be used for managing Alpha IDs. Alpha ID is an alphanumeric phone number (like MICROSOFT, instead of a short code like 23451), most commonly used in the EU market, that's generally used in enterprise/high throughput scenarios when sending SMS messages.

_This is a package that is only going to be used internally._


sdk/communication/Azure.Communication.AlphaIds/tests.yml
```diff
@@ -0,0 +1,18 @@
+trigger: none
+
+extends:
+  template: /eng/pipelines/templates/stages/archetype-sdk-tests.yml
+  parameters:
+    ServiceDirectory: communication
+    Project: Azure.Communication.ShortCodes
```
benbp, line 7: I believe this is the wrong project? Should be AlphaIds.

```suggestion
    Project: Azure.Communication.AlphaIds
```

guilhermeluizsp: Good catch, my mistake. Just fixed it.

sdk/communication/Azure.Communication.AlphaIds/CHANGELOG.md
```diff
@@ -0,0 +1,11 @@
+# Release History
+
+## 1.0.0-alpha.1 (Unreleased)
+
+### Features Added
```
danielortega-msft, line 5: I would suggest not leaving the changelog empty.

danielortega-msft (commented):
Other than the empty changelog, it looks good to me.

danielortega-msft (approved):

PR #30738, opened 2022-08-26
Polish FluidRelay & fix the SecurityInsights to support the new polymorphism

ArthurMa1978:


sdk/securityinsights/Azure.ResourceManager.SecurityInsights/src/autorest.md
```diff
@@ -16,6 +16,9 @@ skip-csproj: true
 modelerfour:
   flatten-payloads: false
 
+suppress-abstract-base-class:
+  - EntityQueryItem
```
fengzhou-msft, line 20: Why we need this for EntityQueryItem?

fengzhou-msft (approved):

sdk/securityinsights/Azure.ResourceManager.SecurityInsights/src/autorest.md
fengzhou-msft: From Arthur: This model is inherit from ResourceData, so should net be converted to abstract.

Yao725: This model inherits from ResourceData because the class that we want to change it to be abstract inherits from ResourceData. During the generating process, we incorrectly identified base class of `UnknownEntityQueryItem` to be `ResourceData` instread of `EntityQueryItem` because of some information missed. I have opened this [PR](https://github.com/Azure/autorest.csharp/pull/2649) to fix the issue.

ArthurMa1978: Can be removed this config when abstract issue fixed

PR #30747, opened 2022-08-26
IsEnvironmentReady is better than static waits :)

christothes:
Even with fairly long static wait times, the live tests still fail sometimes. This is more robust (Josh was right 😄)


JoshLove-msft (approved):

JoshLove-msft:
> Even with fairly long static wait times, the live tests still fail sometimes. This is more robust (Josh was right 😄)

Thank @kasobol-msft for this!

PR #30659, opened 2022-08-22
[Experiment] Prototype of multi-version models

annelo-msft:
Initial prototype of simple multi-version client described in https://github.com/Azure/azure-sdk-for-net/issues/47658

At present, it covers only the case where an optional property is added to a round-trip model across versions.

TODO
- [x] Add convenience API
- [x] Generate model
- [ ] Address serialization
- [ ] Add test cases for back-compat
- [ ] Test cases for forward compat?

The Cadl file describing this client is here: [main-v2client.cadl](https://github.com/annelo-msft/cadl-ranch/blob/cadl-test-multiversion-optprop/packages/cadl-ranch-specs/http/versioning/models/main-v2client.cadl), and is discussed in [this README](https://github.com/annelo-msft/cadl-ranch/blob/cadl-test-multiversion-optprop/packages/cadl-ranch-specs/http/versioning/models/README.md).

PR #30753, opened 2022-08-26
Validate endpoints, models for API version

### LINKED ISSUE #28060: [Text Analytics] Design Error Experience for Unsupported Fields and Operations in Legacy Versions

heaths:
Resolves #28060


sdk/textanalytics/Azure.AI.TextAnalytics/CHANGELOG.md
```diff
@@ -4,12 +4,24 @@
 
 ### Features Added
 
+- Added `DisplayName` property to the following classes to set the optional display name:
```
kristapratico, line 7: Should we say optional display name for the job/operation so it's not confused with action name?

sdk/textanalytics/Azure.AI.TextAnalytics/src/AnalyzeHealthcareEntitiesAction.cs
```diff
@@ -49,7 +49,7 @@ public AnalyzeHealthcareEntitiesAction(AnalyzeHealthcareEntitiesOptions options)
         /// </para>
         /// </summary>
         /// <remarks>
-        /// This property only applies for <see cref="TextAnalyticsClientOptions.ServiceVersion.V3_1"/>, <see cref="TextAnalyticsClientOptions.ServiceVersion.V2022_05_01"/>, and up.
+        /// This property only applies for <see cref="TextAnalyticsClientOptions.ServiceVersion.V3_1"/>, <see cref="TextAnalyticsClientOptions.ServiceVersion.V2022_05_01"/>, and newer.
```
kristapratico, line 52: AnalyzeHealthcareEntitiesAction wasn't supported until Language API, should the disclaimer should apply for the entire action type?

sdk/textanalytics/Azure.AI.TextAnalytics/src/SingleLabelClassifyOptions.cs
```diff
@@ -35,8 +35,13 @@ public SingleLabelClassifyOptions()
         /// </para>
         /// </summary>
         /// <remarks>
-        /// This property only applies for <see cref="TextAnalyticsClientOptions.ServiceVersion.V2022_05_01"/> and up.
+        /// This property only applies for <see cref="TextAnalyticsClientOptions.ServiceVersion.V2022_05_01"/> and newer.
```
kristapratico, line 38: was this meant to be added to DisplayName?

sdk/textanalytics/Azure.AI.TextAnalytics/src/AnalyzeHealthcareEntitiesOptions.cs
```diff
@@ -16,5 +16,17 @@ public class AnalyzeHealthcareEntitiesOptions : TextAnalyticsRequestOptions
         public AnalyzeHealthcareEntitiesOptions()
         {
         }
+
+        /// <summary>
+        /// Optional display name for the operation.
+        /// </summary>
+        public string DisplayName { get; set; }
+
+        internal new AnalyzeHealthcareEntitiesOptions CheckSupported(TextAnalyticsClientOptions.ServiceVersion version)
```
tg-msft, line 25: Why `new`?  Why not make this `private protected virtual` if we're going to use it like this?

sdk/textanalytics/Azure.AI.TextAnalytics/src/AnalyzeHealthcareEntitiesOptions.cs
```diff
@@ -16,5 +16,17 @@ public class AnalyzeHealthcareEntitiesOptions : TextAnalyticsRequestOptions
...
+        {
+            base.CheckSupported(version);
+            Validation.SupportsProperty(DisplayName, nameof(DisplayName), TextAnalyticsClientOptions.ServiceVersion.V2022_05_01, version);
```
tg-msft, line 28: Should we use an `Optional<string>` backing field to tell if the property was ever set instead of just a `null` check?

sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/LegacyServiceClient.cs
```diff
@@ -430,6 +434,8 @@ public override async Task<Response<PiiEntityCollection>> RecognizePiiEntitiesAs
         public override Response<PiiEntityCollection> RecognizePiiEntities(string document, string language = default, RecognizePiiEntitiesOptions options = default, CancellationToken cancellationToken = default)
         {
             Argument.AssertNotNullOrEmpty(document, nameof(document));
+            Validation.SupportsOperation(nameof(TextAnalyticsClient.RecognizePiiEntities), TextAnalyticsClientOptions.ServiceVersion.V3_1, ServiceVersion);
```
tg-msft, line 437: Do you need the method name here?  Or could you use https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.callermembernameattribute?view=net-6.0 instead?

sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/LegacyServiceClient.cs
```diff
@@ -1526,6 +1544,24 @@ private static Legacy.JobManifestTasks CreateTasks(TextAnalyticsActions actions)
             {
                 tasks.SentimentAnalysisTasks = Transforms.ConvertFromAnalyzeSentimentActionsToLegacyTasks(actions.AnalyzeSentimentActions);
             }
+
+            // Not supported in legacy service.
+            if (actions.AnalyzeHealthcareEntitiesActions != null && actions.AnalyzeHealthcareEntitiesActions.Count > 0)
+            {
+                throw Validation.NotSupported(nameof(AnalyzeHealthcareEntitiesAction), TextAnalyticsClientOptions.ServiceVersion.V2022_05_01);
```
tg-msft, line 1551: Should this include the version the customer tried using?

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
```diff
@@ -1058,7 +1064,7 @@ public TextAnalyticsClient(Uri endpoint, AzureKeyCredential credential, TextAnal
         /// <exception cref="RequestFailedException">Service returned a non-success
         /// status code.</exception>
         public virtual async Task<Response<AnalyzeSentimentResultCollection>> AnalyzeSentimentBatchAsync(IEnumerable<TextDocumentInput> documents, AnalyzeSentimentOptions options = default, CancellationToken cancellationToken = default) =>
-            await _serviceClient.AnalyzeSentimentBatchAsync(documents, options, cancellationToken).ConfigureAwait(false);
+            await _serviceClient.AnalyzeSentimentBatchAsync(documents, options?.CheckSupported(ServiceVersion), cancellationToken).ConfigureAwait(false);
```
tg-msft, line 1067: What if we create an internal shared source `IVersioningValidation` interface and then file a bug to create an analyzer making sure it's Validate method is called from any public API taking it as a param?

sdk/textanalytics/Azure.AI.TextAnalytics/src/Validation.cs
```diff
@@ -0,0 +1,44 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Runtime.CompilerServices;
+using static Azure.AI.TextAnalytics.TextAnalyticsClientOptions;
+
+namespace Azure.AI.TextAnalytics
+{
+    internal static class Validation
```
tg-msft, line 10: Can we move this to shared source so the versioning exceptions look the same across libraries?  Or even move this into the `Argument` class?

sdk/textanalytics/Azure.AI.TextAnalytics/src/Validation.cs
```diff
@@ -0,0 +1,44 @@
...
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static void SupportsOperation(string name, ServiceVersion added, ServiceVersion current)
```
tg-msft, line 13: Should we include the client name as well?  It might make it quicker to diagnose the failure.

sdk/textanalytics/Azure.AI.TextAnalytics/src/Validation.cs
```diff
@@ -0,0 +1,44 @@
...
+            where T : class
+        {
+            if (value is not null && current < added)
+            {
+                throw NotSupported(name, added);
+            }
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static NotSupportedException NotSupported(string name, ServiceVersion added) => new($"{name} is only available for API version {GetVersionString(added)} and newer.");
```
tg-msft, line 42:
```suggestion
        public static NotSupportedException NotSupported(string name, ServiceVersion added) => new($"{name} is only available for use with service API version {GetVersionString(added)} and newer.");
```

sdk/textanalytics/Azure.AI.TextAnalytics/src/Validation.cs
```diff
@@ -0,0 +1,44 @@
...
+using static Azure.AI.TextAnalytics.TextAnalyticsClientOptions;
+
+namespace Azure.AI.TextAnalytics
+{
+    internal static class Validation
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static void SupportsOperation(string name, ServiceVersion added, ServiceVersion current)
+        {
+            if (current < added)
```
tg-msft, line 15: I wonder if there's a way we can move this closer to the ServiceVersion enum definition so people updating it remember to patch up this logic?  I'm thinking about particularly screwy cases where we do things like rename `xyz_preview` to `Xyz_Preview`. I think the enum math should generally hold, but I'd love to give devs a chance to update that if they insist on doing something clever with their ServiceVersion.

tg-msft (commented):
I like the direction this is going.  I only skimmed the validation code and will let other folks do the actual TA review.

sdk/textanalytics/Azure.AI.TextAnalytics/src/AnalyzeHealthcareEntitiesOptions.cs
heaths: Because the return type changes. I considered making this generic, but not sure that offers any benefit over this. `call` is cheaper than `callvirt`, and with this approach you know that whatever type you call this method (convention by name) is what you get back. Thoughts?

sdk/textanalytics/Azure.AI.TextAnalytics/src/AnalyzeHealthcareEntitiesOptions.cs
heaths: Good idea. I'll look into that. Seems plausible.

sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/LegacyServiceClient.cs
heaths: I actually considered that, but the exception message should include the method people called - not the implementation, which *might* change. Probably not, but I was being paranoid about it.

sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/LegacyServiceClient.cs
heaths: Does it add value? They should know what they passed in. Even as-is, this is an actionable message. Making it more wordy doesn't seem to add value to me.

sdk/textanalytics/Azure.AI.TextAnalytics/src/Validation.cs
heaths: After talking with Anne, it sounded like we didn't even want to consider client-side validation yet. I didn't want to move it to shared source prematurely, but figured we could - just like we did with the challenge auth policy - when needed.

sdk/textanalytics/Azure.AI.TextAnalytics/src/Validation.cs
heaths: That'd be more to pass down from the caller which, on its own isn't a problem, but as asked above: does it add value? The exception would clearly show what was called where. And we only have one public client anyway.

sdk/textanalytics/Azure.AI.TextAnalytics/src/Validation.cs
heaths: But what version added what features is invariant, hence why I didn't add it to the `ServiceVersion` originally - something along those lines did cross my mind. As for validation, that's why I added tests for all covered cases. If the tests break, the dev regressed something.

sdk/textanalytics/Azure.AI.TextAnalytics/src/AnalyzeHealthcareEntitiesAction.cs
heaths: Good catch. I copied and pasted everywhere, but this would be misleading.

sdk/textanalytics/Azure.AI.TextAnalytics/src/AnalyzeHealthcareEntitiesOptions.cs
tg-msft: Why do we need to return it?

sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/LegacyServiceClient.cs
tg-msft: Yes - because many customers silently change versions on a package upgrade when the default of ClientOptions changes.

sdk/textanalytics/Azure.AI.TextAnalytics/src/Validation.cs
tg-msft: I'd like a general pattern and many libraries will have multiple clients - so I think it adds value.

sdk/textanalytics/Azure.AI.TextAnalytics/src/SingleLabelClassifyOptions.cs
heaths: No, it's on the right property, but like you're comment above, since SingleLabelClassify wasn't even added until 2022-05-01 this is redundant and probably confusing.

heaths: I've removed this remark from single-, multi-label classify, and recognize custom entities since those were added in 2022-05-01 anyway. The comment is redundant.

sdk/textanalytics/Azure.AI.TextAnalytics/src/Validation.cs
tg-msft: It might not always be - especially if someone does something against guidance and different clouds support different features in the same api-version.

sdk/textanalytics/Azure.AI.TextAnalytics/src/AnalyzeHealthcareEntitiesOptions.cs
heaths: To be able to call it inline, but honestly that's just a nice-to-have. I just have to go to block bodies now. :)

sdk/textanalytics/Azure.AI.TextAnalytics/src/Validation.cs
heaths: Per offline discussion, we'll move this when needed, but I'll make some of these other changes and add some comments with some of our future thoughts we don't need to implement right now.

sdk/textanalytics/Azure.AI.TextAnalytics/src/Validation.cs
jsquire: > different clouds support different features in the same api-version

sdk/textanalytics/Azure.AI.TextAnalytics/src/AnalyzeSentimentOptions.cs
```diff
@@ -32,8 +32,15 @@ internal AnalyzeSentimentOptions(TextAnalyticsRequestOptions options)
...
         /// <remarks>
-        /// This property only applies for <see cref="TextAnalyticsClientOptions.ServiceVersion.V3_1"/>, <see cref="TextAnalyticsClientOptions.ServiceVersion.V2022_05_01"/>, and up.
+        /// This property only applies for <see cref="TextAnalyticsClientOptions.ServiceVersion.V3_1"/>, <see cref="TextAnalyticsClientOptions.ServiceVersion.V2022_05_01"/>, and newer.
         /// </remarks>
         public bool? IncludeOpinionMining { get; set; }
+
+        internal new AnalyzeSentimentOptions CheckSupported(TextAnalyticsClientOptions.ServiceVersion version)
+        {
+            base.CheckSupported(version);
+            Validation.SupportsProperty(IncludeOpinionMining, nameof(IncludeOpinionMining), TextAnalyticsClientOptions.ServiceVersion.V3_1, version);
```
jsquire, line 42: Too bad we don't have a way to declare a min version as metadata and not take a perf hit.   I wonder if we should consider somehow organizing the version requirements in one place, but I don't have an idea for how to structure that in a way that wouldn't be a huge readability mess.

sdk/textanalytics/Azure.AI.TextAnalytics/tests/MockTests/Legacy/AnalyzeOperationV30MockTests.cs
```diff
@@ -0,0 +1,102 @@
...
+// Licensed under the MIT License.
+
+using System;
+using Azure.Core.Pipeline;
+using Azure.Core.TestFramework;
+using NUnit.Framework;
+
+namespace Azure.AI.TextAnalytics.Legacy.Tests
+{
+    public class AnalyzeOperationV30MockTests : ClientTestBase
```
jsquire, line 11: If we did find some way to centralize min versions, it'd be great to data-drive these so that we always have coverage when making changes.

jsquire (approved):

sdk/textanalytics/Azure.AI.TextAnalytics/src/AnalyzeSentimentOptions.cs
annelo-msft: In Cadl, we'll be able to generate that from the API description in a straightforward way.  The outstanding question I think is how we would want to use it.

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
```diff
@@ -218,7 +224,7 @@ public TextAnalyticsClient(Uri endpoint, AzureKeyCredential credential, TextAnal
         /// <exception cref="RequestFailedException">Service returned a non-success
         /// status code.</exception>
         public virtual async Task<Response<DetectLanguageResultCollection>> DetectLanguageBatchAsync(IEnumerable<string> documents, string countryHint = default, TextAnalyticsRequestOptions options = default, CancellationToken cancellationToken = default) =>
-            await _serviceClient.DetectLanguageBatchAsync(documents, countryHint, options, cancellationToken).ConfigureAwait(false);
+            await _serviceClient.DetectLanguageBatchAsync(documents, countryHint, options?.CheckSupported(ServiceVersion), cancellationToken).ConfigureAwait(false);
```
annelo-msft, line 227: What if we did this in post-facto validation and only if a service returned a 400 error?  That way, we only take the perf hit for validation if we're already invoking the exception framework.

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
annelo-msft: Do we want the perf overhead of that on all service calls?  Would all models with properties added in later versions need to implement this?

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
```diff
@@ -637,7 +643,7 @@ public TextAnalyticsClient(Uri endpoint, AzureKeyCredential credential, TextAnal
         /// <see href="https://aka.ms/azsdk/textanalytics/data-limits"/>.</para>
         /// </summary>
         /// <remarks>
-        /// This method is only available for <see cref="TextAnalyticsClientOptions.ServiceVersion.V3_1"/>, <see cref="TextAnalyticsClientOptions.ServiceVersion.V2022_05_01"/>, and up.
+        /// This method is only available for <see cref="TextAnalyticsClientOptions.ServiceVersion.V3_1"/>, <see cref="TextAnalyticsClientOptions.ServiceVersion.V2022_05_01"/>, and newer.
```
annelo-msft, line 646: nit: 
```suggestion
        /// This method is only available for <see cref="TextAnalyticsClientOptions.ServiceVersion.V3_1"/>, <see cref="TextAnalyticsClientOptions.ServiceVersion.V2022_05_01"/>, and later.
```

?

sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/LegacyServiceClient.cs
```diff
@@ -384,6 +386,8 @@ private Response<RecognizeEntitiesResultCollection> RecognizeEntitiesBatch(Multi
         public override async Task<Response<PiiEntityCollection>> RecognizePiiEntitiesAsync(string document, string language = default, RecognizePiiEntitiesOptions options = default, CancellationToken cancellationToken = default)
         {
             Argument.AssertNotNullOrEmpty(document, nameof(document));
+            Validation.SupportsOperation(nameof(TextAnalyticsClient.RecognizePiiEntitiesAsync), TextAnalyticsClientOptions.ServiceVersion.V3_1, ServiceVersion);
```
annelo-msft, line 389: I would also consider post-facto validation in this case.  It could go in the `catch` block below.

sdk/textanalytics/Azure.AI.TextAnalytics/src/Validation.cs
annelo-msft, line 42: [](url)

annelo-msft:
```suggestion
        public static NotSupportedException NotSupported(string name, ServiceVersion added) => new($"{name} is only available for use with service api-version={GetVersionString(added)} and newer.");
```
?

sdk/textanalytics/Azure.AI.TextAnalytics/src/Validation.cs
annelo-msft: @tg-msft, why shared source and not a public type?

sdk/textanalytics/Azure.AI.TextAnalytics/src/AnalyzeSentimentOptions.cs
heaths: @tg-msft and I have chatted about that briefly, and one way might be to use Cadl `@added` or even `@removed` decorators and combine them with source generators (that way we can do the same thing in HLCs) to "transform" a internal? method into a public one that runs checks, then calls that internal? (maybe private) method - effectively a compile-time equivalent of decorators in TypeScript/newer JavaScript and Python.

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
heaths: I like the idea. The only downside that came to mind is the customer is still incurring network IO. I would hope a bad request (or any non-2xx) wouldn't cost them either, but something to consider. @tg-msft, thoughts?

I wouldn't do it in this PR, but if we did it, in the subsequent change I talked about that, if necessary, could happen post-GA and service as a pattern (or even sharable, once moved) for other SDKs - maybe even DPGs.

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
```diff
@@ -246,7 +252,7 @@ public TextAnalyticsClient(Uri endpoint, AzureKeyCredential credential, TextAnal
         /// <exception cref="RequestFailedException">Service returned a non-success
         /// status code.</exception>
         public virtual Response<DetectLanguageResultCollection> DetectLanguageBatch(IEnumerable<string> documents, string countryHint = default, TextAnalyticsRequestOptions options = default, CancellationToken cancellationToken = default) =>
-            _serviceClient.DetectLanguageBatch(documents, countryHint, options, cancellationToken);
+            _serviceClient.DetectLanguageBatch(documents, countryHint, options?.CheckSupported(ServiceVersion), cancellationToken);
```
annelo-msft, line 255: Question about behavior.  It looks like this will throw a NotSupportedException if someone sets DisableServiceLogs to any value.  Is this behavior we think customers would want?  One of true/false must match the service default, so if a user set the value to the same value as the service default, they would change nothing about the operation but we would prevent this by throwing an exception.  This feels incorrect to me.  I would expect us to only prevent the request unless sending the property could somehow cause data corruption.  

As a general rule, I think we should only throw our own exception, and not the standard RequestFailedException with the service error, if we know the error message coming back from the service doesn't help the client user understand what they did wrong and how to fix it.  In a manually-written customization layer, I would use that to decide, and if we need to throw a post-facto exception to indicate a versioning issue caused the problem, I would add the service-side RequestFailedException as an inner exception.

heaths: It's a good point but knowing how - or even if - the service will return a specific enough error, or whether the error is even an HTTP 400, will be difficult. Right now, our guidance for swaggers is to only document your 2xx response model and assume any error is `default`, without going into specifics. Conceptually, I love the idea but I think doing it in practice will be much harder.

Without hardcoding what the default is - and, of course, that could change (KV has done it) - we can't really know. So only doing this for error responses would make even more sense; but again, I worry about practical feasibility. And @tg-msft's [suggestion above](https://github.com/Azure/azure-sdk-for-net/pull/30753#discussion_r957717131) wouldn't change that. We'd just know if it is changed to any value. So, in practice, it makes me wonder if I should leave which, yes, is assuming the default remains the same; however, that does at least lend to the oft possibility that they are setting the value to its non-default value.

annelo-msft: Yeah, I just picked being redundant with the default of one case of sending a request that might still be useful if accepted by the service.  It's worth noting here, however, that if the service is following Azure Guidelines, they would reject any request that has unknown JSON values.

Overall, I think this is further argument for using post-facto validation rather than request validation.

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
```diff
@@ -1423,7 +1429,7 @@ public TextAnalyticsClient(Uri endpoint, AzureKeyCredential credential, TextAnal
         /// <exception cref="RequestFailedException">Service returned a non-success
         /// status code.</exception>
         public virtual async Task<AnalyzeHealthcareEntitiesOperation> StartAnalyzeHealthcareEntitiesAsync(IEnumerable<string> documents, string language = default, AnalyzeHealthcareEntitiesOptions options = default, CancellationToken cancellationToken = default) =>
-            await _serviceClient.StartAnalyzeHealthcareEntitiesAsync(documents, language, options, cancellationToken).ConfigureAwait(false);
+            await _serviceClient.StartAnalyzeHealthcareEntitiesAsync(documents, language, options?.CheckSupported(ServiceVersion), cancellationToken).ConfigureAwait(false);
```
annelo-msft, line 1432: This feels a little strange to me that the service method is what was newly added, but we are validating the model properties rather than the version the service was added.  I would expect us to establish some sort of precedence hierarchy if we decide to do validation, and make sure we're doing the least-expensive check first.

heaths: Actually, the method is as well and the check would be redundant in a few other cases, but not this one: `AnalyzeHealthcareEntitiesOptions` also works for the legacy service. But your point is spot-on for single- and multi-label classify, as well as custom entity recognition. If I hadn't already, I'll be sure to remove the checks since the methods themselves are validated. If they are needed in the future, they should be added then.

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
jsquire: > What if we did this in post-facto validation and only if the service returned a 400 error? That way, we only take the perf hit for validation if we're already invoking the exception framework (and therefore in a known-bad performance case).

I don't think we can depend on a 400 in some cases.  For example, someone is doing a PUT request and has a model populated with properties that don't exist in the requested service version.  The service will silently ignore these properties - but they influence behavior in newer service versions.  As a result, the data is considered valid by the service, but the behavior does not match user expectations.

sdk/textanalytics/Azure.AI.TextAnalytics/src/AnalyzeHealthcareEntitiesOptions.cs
heaths: Per offline conversation, I'm going to punt some of this to a lower-pri issue and will try to get to it before GA, but I will make a few of these changes - especially around messaging - as part of this PR. Implementation details beyond messaging can wait, since there's little time before GA.

heaths: Actually, checking `Nullable<bool>.HasValue` (or just a null check), which `SupportsProperty` does, is better because that's what the generated `Create{Whatever}Request` does:
```C#
            if (loggingOptOut != null)
            {
                uri.AppendQuery("loggingOptOut", loggingOptOut.Value, true);
            }
```
I feel it's better to be consistent.

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
heaths: Most of the checks so far are happening in the LegacyServiceClient, so impact is minimal. Even still, this check is no more work that any argument checks we do.

Whatever and whenever we decide how we want to do this long-term, some maintainer can come back and make changes. I've isolated the brunt of the validation code (all the callees) to a separate "Validation" folder, though the namespace does not reflect that so no additional `using`s are needed in case we remove it.

sdk/textanalytics/Azure.AI.TextAnalytics/src/Validation.cs
heaths: I did this and isolated all the callees to a "Validation" folder, though the namespace doesn't reflect that so no addiitonal `using`s are needed and might have to be removed later.

sdk/textanalytics/Azure.AI.TextAnalytics/src/Validation.cs
heaths: I updated #30852 mentioning this and probably passing in some `IComparer<T>` for the service version if more complex rules are needed; however, since they are not here, I don't want to overload this implementation too much considering we may not even want to do it long term.

sdk/textanalytics/Azure.AI.TextAnalytics/src/AnalyzeSentimentOptions.cs
heaths: I've captured this in #30852 but is beyond the scope for this PR.

sdk/textanalytics/Azure.AI.TextAnalytics/tests/MockTests/Legacy/AnalyzeOperationV30MockTests.cs
heaths: Agreed, but seems beyond the scope of this PR, especially given timing. I'm pretty diligent in my testing, so lets hope at the very least future maintainers are as well.

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
heaths: This is being discussed in more detail by the REST API Stewardship board, but is beyond the scope of this PR. The version check is minimal and no more impactful than argument checks.

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
heaths: Oh, and as a long term setup-type person, "or newer" is more typical with version numbers, though "or later" is still better than "and up", which is what these all said before.

sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/LegacyServiceClient.cs
heaths: As noted above, the service error is not specific enough and the version check itself is minimal - no different than all our argument checks. If the perf of this is concerning, we should do more to elide argument checks in internal/private code and do that all in public code so it's not repeated.

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
heaths: After testing, in TA's case they don't err for unsupported query parameters at all. After discussing in the v-team, we've decided that we will do client-side checks because a customer expecting any information based on setting this or other properties that are only supported in newer versions will be unpleasantly surprised.

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
heaths: Turns out I already don't call `CheckSupported()` on these options classes.

annelo-msft (commented):
What do you think about this as an analogous implementation, but with types that could go in Core shared source? https://github.com/Azure/azure-sdk-for-net/pull/30659/commits/ee5cadfd2ae9ef2177a4c58f55f7b281ca328e32

PR #30754, opened 2022-08-26
Remove ServiceToTest, CodeCheck run on service only

### LINKED ISSUE #3999: Improve the performance of CodeCheck.ps1 for .NET
The .NET pipeline currently slow at `Analyze` job.
VerifyGeneratedCode step takes more than an hour.

Here is the analysis of the script:
2022-08-23T01:50:08.0985600Z Re-generating clients (1 hour)
2022-08-23T02:52:21.1489417Z Re-generating tests (1 min)
2022-08-23T02:53:06.5620905Z Re-generating snippets (40 sec)
2022-08-23T02:53:41.5062998Z Re-generating listings (6 min)

Will look into the most time-consuming part.

sima-zhu:
Resolve https://github.com/Azure/azure-sdk-tools/issues/3999
The PR is to speed up `Analyze` job for .NET core pipeline. 
Before it takes more than an hour to get it done. 
* https://dev.azure.com/azure-sdk/public/_build/results?buildId=1845151&view=logs&s=6884a131-87da-5381-61f3-d7acc3b91d76

After the changes
* (core):
https://dev.azure.com/azure-sdk/public/_build/results?buildId=1845393&view=logs&s=6884a131-87da-5381-61f3-d7acc3b91d76
* (Non-core)
https://dev.azure.com/azure-sdk/public/_build/results?buildId=1845431&view=results

----------------------------------------------
Our scope changes to run target service only in core pipeline for CodeCheck.ps1. Add code check in aggregate-report pipeline for awareness.
The following 2 scenarios can be covered from autogen pipeline. (For any possible flaws are supposed to address in autogen pipeine)
(Quote from Wes)
1. The code generator changed. If it updates code then it will trigger the individual pipelines for the service directory which will catch any issues where the service team might have manually changed something.
1. If the service team manually changed the generated code it will trigger the code check in their pipeline and will catch that scenario
 
Note: Also make some changes to test splitting to make it robust.
Testing pipeline: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1856319&view=logs&j=d41e9505-d81b-5b3c-5b50-be71bc9b0ccb&t=7061cab9-8edd-561e-5078-022c316ef413


eng/scripts/CodeChecks.ps1
```diff
@@ -5,6 +5,9 @@ param (
     [Parameter(Position=0)]
     [string] $ServiceDirectory,
 
+    [Parameter()]
```
weshaggard, line 8: I'm not sure I understand why we need to add this parameter. I was hoping we would actually be able to eliminate it everywhere after this work is done.

eng/pipelines/templates/jobs/ci.yml
```diff
@@ -218,6 +219,43 @@ jobs:
...
+        JobTemplatePath: /eng/pipelines/templates/jobs/code.gen.yml
+        MatrixConfigs:
+        - Name: NET_ci_codegen
+          Path: eng/pipelines/templates/stages/code-gen-matrix.json
+          Selection: sparse
+          GenerateVMJobs: true
+        MatrixFilters: ${{ parameters.MatrixFilters }}
+        MatrixReplace: ${{ parameters.MatrixReplace }}
+        CloudConfig:
+          Cloud: public
```
weshaggard, line 235: I know you are trying to reuse as much as possible for generating the matrix bot some parameters like this don't really apply but is it still needed?

eng/pipelines/templates/steps/dependency.tests.yml
```diff
@@ -23,12 +23,21 @@ parameters:
 - name: ServiceDirectory
```
weshaggard, line 23: We might want to rename this template to something like split.dependencies.yml

weshaggard: Nevermind given we are plugging this in as a set of tests I think we can keep the name dependency.tests.yml.

eng/pipelines/templates/jobs/ci.yml
sima-zhu: Removed the unneccessary parts.

eng/scripts/CodeChecks.ps1
sima-zhu: I don't want to put "remove $ServiceToTest" task into this PR.
Here is the reason:
CodeCheck.ps1 includes the following 4 tasks: 
(Note: $ServiceToTask is * for core/releasemanagerment, otherwise $ServiceDirectory)
```
1. Re-generating clients (dotnet msbuild). We use $ServiceDirectory for the PR
2. Re-generating tests (dotnet msbuild). We use $ServiceToTask for the PR
3. Re-generating snippets (script). We use $ServiceToTask for the PR
4. Re-generating listings (script). We use $ServiceToTask for the PR
```
To remove $ServiceToTask, we have to take care of the last two which either run everything in the splitting job or let them run on the Prop file. Both need some efforts to make them run properly.

I prefer to start another PR for $ServiceToTask removal. The PR is to decouple the two parameter `$ServiceDirectory` and `$ServiceToTest`, which is convenient for the next step.

eng/pipelines/templates/stages/code-gen-matrix.json
```diff
@@ -0,0 +1,10 @@
+{
+    "matrix": {
+      "Agent": {
+        "Ubuntu-20.04": {
+          "OSVmImage": "MMSUbuntu20.04",
+          "Pool": "azsdk-pool-mms-ubuntu-2004-general"
+        }
+      }
+    }
```
benbp, line 9:
```suggestion
  "matrix": {
    "Agent": {
      "Ubuntu-20.04": {
        "OSVmImage": "MMSUbuntu20.04",
        "Pool": "azsdk-pool-mms-ubuntu-2004-general"
      }
    }
  }
```

eng/scripts/splittestdependencies/generate-dependency-functions.ps1
```diff
@@ -26,15 +26,16 @@ function Split-Project-File-To-Groups($ProjectFile, $NumberOfTestsPerJob, $Exclu
...
+function Write-Test-Dependency-Group-To-Files($ProjectFileConfigName, $ProjectGroups, $MatrixOutputFolder) {
   if (!$ProjectGroups -or !$ProjectGroups[0]) {
     Throw "There should be some project files in the group. Please check the given project list."
   }
   $null = New-Item -Path $MatrixOutputFolder -ItemType "directory" -Force 
   $numOfGroups = $ProjectGroups.Count
   $allOutputFiles = @()
   for ($i = 0; $i -lt $numOfGroups; $i++) {
-    $projectFilePath = "Project_$i.props"
+    $projectFilePath = "$($ProjectFileConfigName)_Project_$i.props"
```
benbp, line 38:
```suggestion
    $projectFilePath = "${ProjectFileConfigName}_Project_$i.props"
```

benbp (approved):

eng/scripts/CodeChecks.ps1
sima-zhu: Talked offline. As this is the only place to use $ServiceToTest, we'd like to remove in the same PR.
@heaths @JoshLove-msft 
We have an assumption that #3, #4 have no necessary to run against everything (*) in core pipeline. Do you have any concerns for the assumption?

eng/pipelines/aggregate-reports.yml
```diff
@@ -49,6 +49,19 @@ stages:
...
+            displayName: "Verify generated code"
+            env:
+              EnableSourceLink: false
+            inputs:
+              filePath: "eng/scripts/CodeChecks.ps1"
+              arguments: >
+                -ServiceDirectory '*'
+                -SDKType 'all'
+                -SpellCheckPublicApiSurface
+                pwsh: true
```
heaths, line 62: Does this have the right whitespace? It's at the same level with arguments.

eng/pipelines/templates/jobs/ci.yml
```diff
@@ -198,7 +195,7 @@ jobs:
         inputs:
           filePath: "eng/scripts/CodeChecks.ps1"
           arguments: >
-           -ServiceDirectory ${{ parameters.ServiceToTest }}
+           -ServiceDirectory ${{ parameters.ServiceDirectory }}
```
heaths, line 198: Won't this become "*"? I don't think CodeChecks or its subordinate scripts support that.

eng/scripts/splittestdependencies/generate-dependency-functions.ps1
```diff
@@ -62,20 +63,25 @@ function Write-Project-Files-To-Matrix($ProjectFiles, $MatrixJsonPath, $MatrixOu
...
+    foreach ($include in $platformJson.include) {
+      if (!$overrideFiles) {
+        Throw "Need at least one project file for include property. Please adjust the NumberOfTestsPerJob for better split."
+      }
+      $lastElem = $overrideFiles.PSObject.Properties | Select-Object -Last 1
+      $overrideFiles.PSObject.properties.remove($lastElem.Name)
+      
+      $dependencyGroup = New-Object PSObject
+      $dependencyGroup | Add-Member -Name "DependencyGroup$n" -value $lastElem.Value -MemberType NoteProperty
+      $include | Add-Member -Name "OverrideFiles" -value $dependencyGroup -MemberType NoteProperty
```
heaths, line 83: Tip: it's often easier and faster to create new objects like so:
```powershell
$dependencyGroup = [pscustomobject]@{
  "DependencyGraph$n" = $lastElem.Value
}
```

eng/pipelines/templates/jobs/ci.yml
sima-zhu: For our new proposal, we don't support * anymore using $ServiceToTest.

heaths: Okay. It does appear `join-path <path> '*' | resolve-path` resolves correctly.

heaths (approved):

eng/pipelines/aggregate-reports.yml
sima-zhu: The [pipeline](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1855952&view=logs&s=f4f4e1a1-fb20-57e5-6b6d-1b695e9938d2) failed because of the alignment. I am kind of blind to check what was wrong.

eng/pipelines/aggregate-reports.yml
```diff
@@ -49,6 +49,19 @@ stages:
               Directory: ""
               CheckLinkGuidance: $true
 
+          - task: PowerShell@2
```
weshaggard, line 52: Let's remove this as we talked in the group chat I don't think it is necessary. We should only add it once find there is a need for it.

weshaggard (approved):
Please remove the aggregate-reports change but otherwise looks good.

Issue #26659, opened 2022-01-28
Add Deadletter overload that allows passing reason and props


JoshLove-msft:
Ah, I see. I'd recommend just hardcoding the values for these two keys - https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConstants.cs#L32-L39

I do think we should make it easier to support this scenario in the SDK though. I will file a feature request for this.

_Originally posted by @JoshLove-msft in https://github.com/Azure/azure-sdk-for-net/issues/26622#issuecomment-1024538588_

PR #30758, opened 2022-08-26
[Service Bus] Add new DeadLetterMessageAsync overloads

m-redding:
# Summary

The goal of these changes is to add new overloads to `ServiceBusReceiver.DeadletterMessageAsync` and `ServiceBusMessageActions.DeadLetterMessageAsync` that take in a properties dictionary and a dead letter reason and error description. Previously, there was the only the option to pass in a properties dictionary or a reason and error description.

If a user decides to pass a dead letter reason in through the method arguments AND using the AmqpMessageConstants.DeadLetterReason dictionary key, the method throws an exception if they are not the same value. This is true for the dead letter exception description as well.


m-redding:
resolves #26659

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpReceiver.cs
```diff
@@ -735,7 +735,20 @@ private void ThrowLockLostException()
...
                         {
-                            rejected.Error.Info.Add(pair.Key, amqpObject);
+                            // Attempting to set the dead letter reason or description header through the properties
+                            var isReasonHeaderKey = (pair.Key == AmqpMessageConstants.DeadLetterReasonHeader);
+                            var isDescriptionHeaderKey = (pair.Key == AmqpMessageConstants.DeadLetterErrorDescriptionHeader);
+
+                            // General case
+                            var isOtherProperty = !(isReasonHeaderKey || isDescriptionHeaderKey);
+
+                            // Only allow setting of a header property if the header was not passed in through the parameters
```
JoshLove-msft, line 745: We probably need to detect this and throw an exception if the user tries to set the value in both places. We could consider not throwing if the values in both places are identical. It might be easier to add this logic directly to the new method that adds the ability to specify in both places.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Receiver/ServiceBusReceiver.cs
```diff
@@ -683,6 +683,56 @@ public virtual async Task CloseAsync(CancellationToken cancellationToken = defau
                 cancellationToken: cancellationToken).ConfigureAwait(false);
         }
 
+        /// <summary>
```
jsquire, line 686: Nice comment block!

jsquire (commented):
LGTM.   Deferring to Josh for approval.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Receiver/ServiceBusReceiver.cs
```diff
@@ -683,6 +686,82 @@ public virtual async Task CloseAsync(CancellationToken cancellationToken = defau
...
+            string deadLetterErrorDescription = default,
+            CancellationToken cancellationToken = default)
+        {
+            Argument.AssertNotNull(message, nameof(message));
+            Argument.AssertNotNull(propertiesToModify, nameof(propertiesToModify));
+
+            // Prevent properties and arguments from setting distinct deadletter reasons or error descriptions
+            foreach (KeyValuePair<string, object> pair in propertiesToModify)
+            {
+                if (AmqpMessageConverter.TryGetAmqpObjectFromNetObject(pair.Value, MappingType.ApplicationProperty, out var amqpObject))
```
JoshLove-msft, line 736: Hmm.. do we need to convert to the amqpObject here? I think we can just directly check the passed in dictionary.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Receiver/ServiceBusReceiver.cs
```diff
@@ -683,6 +686,91 @@ public virtual async Task CloseAsync(CancellationToken cancellationToken = defau
...
+            // Prevent properties and arguments from setting distinct deadletter reasons or error descriptions
+            foreach (KeyValuePair<string, object> pair in propertiesToModify)
+            {
+                // Attempting to set the dead letter reason or description header through the properties
+                var isReasonHeaderKey = (pair.Key == AmqpMessageConstants.DeadLetterReasonHeader);
+                var isDescriptionHeaderKey = (pair.Key == AmqpMessageConstants.DeadLetterErrorDescriptionHeader);
+
+                if ((isReasonHeaderKey && deadLetterReason != null) && (pair.Value.ToString() != deadLetterReason))
+                {
+                    throw new ArgumentException("Differing deadletter reasons cannot be passed through both the properties and the parameter.");
```
JoshLove-msft, line 758: nit: I think InvalidOperationException is more intuitive here.

JoshLove-msft: Consider adding another sentence that says something like "pass it in one or the other place".

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Receiver/ReceiverTests.cs
```diff
@@ -348,5 +351,31 @@ public async Task CreatingReceiverWithIdentifierSetsIdentifier()
             var identifier = receiver.Identifier;
             Assert.AreEqual(setIdentifier, identifier);
         }
+
+        [Test]
+        public async Task DeadLetterMessageChecksArguments()
```
JoshLove-msft, line 356: We should also add a test case where the values are specified in properties and parameter, but they are the same so we are not expecting an exception.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Receiver/ServiceBusReceiver.cs
```diff
@@ -683,6 +686,86 @@ public virtual async Task CloseAsync(CancellationToken cancellationToken = defau
...
+            IDictionary<string, object> propertiesToModify,
+            string deadLetterReason,
+            string deadLetterErrorDescription = default,
+            CancellationToken cancellationToken = default)
+        {
+            Argument.AssertNotNull(message, nameof(message));
+            Argument.AssertNotNull(propertiesToModify, nameof(propertiesToModify));
+
+            // Prevent properties and arguments from setting distinct deadletter reasons or error descriptions
+            foreach (KeyValuePair<string, object> pair in propertiesToModify)
```
JoshLove-msft, line 745: Why do we loop through the properties rather than just attempting to look up the keys we are interested in?

m-redding: Ah I guess I just moved it over directly. Updated!

sdk/servicebus/Azure.Messaging.ServiceBus/src/Receiver/ServiceBusReceiver.cs
```diff
@@ -683,6 +686,83 @@ public virtual async Task CloseAsync(CancellationToken cancellationToken = defau
...
+            CancellationToken cancellationToken = default)
+        {
+            Argument.AssertNotNull(message, nameof(message));
+            Argument.AssertNotNull(propertiesToModify, nameof(propertiesToModify));
+
+            // Prevent properties and arguments from setting distinct deadletter reasons or error descriptions
+            var containsReasonHeader = propertiesToModify.TryGetValue(AmqpMessageConstants.DeadLetterReasonHeader, out var reasonHeaderProperty);
+            var containsDescriptionHeader = propertiesToModify.TryGetValue(AmqpMessageConstants.DeadLetterErrorDescriptionHeader, out var descriptionHeaderProperty);
+
+            if ((containsReasonHeader && deadLetterReason != null) && (reasonHeaderProperty.ToString() != deadLetterReason))
```
JoshLove-msft, line 748: I don't think we want to call ToString but rather check that the different values are strings, e.g.
```c#
bool containsReasonHeader = propertiesToModify.TryGetValue(AmqpMessageConstants.DeadLetterReasonHeader, out object reasonHeaderProperty) && reasonHeaderProperty is string;
```

JoshLove-msft, line 748:
```suggestion
            if ((containsReasonHeader && deadLetterReason != null) && (reasonHeaderProperty != deadLetterReason))
```

sdk/servicebus/Azure.Messaging.ServiceBus/src/Receiver/ServiceBusReceiver.cs
```diff
@@ -683,6 +686,83 @@ public virtual async Task CloseAsync(CancellationToken cancellationToken = defau
...
+            {
+                throw new InvalidOperationException("Differing deadletter reasons cannot be passed through both the properties and the parameter. Pass this value in one or the other place.");
+            }
+
+            if ((containsDescriptionHeader && deadLetterErrorDescription != null) && (descriptionHeaderProperty.ToString() != deadLetterErrorDescription))
```
JoshLove-msft, line 753:
```suggestion
            if ((containsDescriptionHeader && deadLetterErrorDescription != null) && (descriptionHeaderProperty != deadLetterErrorDescription))
```

sdk/servicebus/Azure.Messaging.ServiceBus/src/Receiver/ServiceBusReceiver.cs
```diff
@@ -683,6 +686,83 @@ public virtual async Task CloseAsync(CancellationToken cancellationToken = defau
...
+            IDictionary<string, object> propertiesToModify,
+            string deadLetterReason,
+            string deadLetterErrorDescription = default,
+            CancellationToken cancellationToken = default)
+        {
+            Argument.AssertNotNull(message, nameof(message));
+            Argument.AssertNotNull(propertiesToModify, nameof(propertiesToModify));
+
+            // Prevent properties and arguments from setting distinct deadletter reasons or error descriptions
+            var containsReasonHeader = propertiesToModify.TryGetValue(AmqpMessageConstants.DeadLetterReasonHeader, out var reasonHeaderProperty);
```
JoshLove-msft, line 745: nit:
```suggestion
            bool containsReasonHeader = propertiesToModify.TryGetValue(AmqpMessageConstants.DeadLetterReasonHeader, out object reasonHeaderProperty);
```

sdk/servicebus/Azure.Messaging.ServiceBus/src/Receiver/ServiceBusReceiver.cs
```diff
@@ -683,6 +686,83 @@ public virtual async Task CloseAsync(CancellationToken cancellationToken = defau
...
+            var containsDescriptionHeader = propertiesToModify.TryGetValue(AmqpMessageConstants.DeadLetterErrorDescriptionHeader, out var descriptionHeaderProperty);
```
JoshLove-msft, line 746: nit:
```suggestion
            bool containsDescriptionHeader = propertiesToModify.TryGetValue(AmqpMessageConstants.DeadLetterErrorDescriptionHeader, out object descriptionHeaderProperty);
```

sdk/servicebus/Azure.Messaging.ServiceBus/src/Receiver/ServiceBusReceiver.cs
JoshLove-msft: Calling ToString would either be a no-op in the case that the variable was already a string, or it would end up calling ToString on some reference type like an array. I think by just comparing the instances directly we get the behavior that we want - if they are both strings we get a value comparison, if one is not a string then we return true.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Receiver/ServiceBusReceiver.cs
```diff
@@ -683,6 +686,83 @@ public virtual async Task CloseAsync(CancellationToken cancellationToken = defau
...
+
+            if ((containsReasonHeader && deadLetterReason != null) && (reasonHeaderProperty.ToString() != deadLetterReason))
+            {
+                throw new InvalidOperationException("Differing deadletter reasons cannot be passed through both the properties and the parameter. Pass this value in one or the other place.");
```
JoshLove-msft, line 750:
```suggestion
                throw new InvalidOperationException("Differing deadletter reasons cannot be specified for both the 'propertiesToModify' and 'deadLetterReason' parameters. The values should either be identical or only be specified in one of the parameters.");
```

sdk/servicebus/Azure.Messaging.ServiceBus/src/Receiver/ServiceBusReceiver.cs
```diff
@@ -683,6 +686,83 @@ public virtual async Task CloseAsync(CancellationToken cancellationToken = defau
...
+            }
+
+            if ((containsDescriptionHeader && deadLetterErrorDescription != null) && (descriptionHeaderProperty.ToString() != deadLetterErrorDescription))
+            {
+                throw new InvalidOperationException("Differing deadletter error descriptions cannot be passed through both the properties and the parameter. Pass this value in one or the other place.");
```
JoshLove-msft, line 755:
```suggestion
                throw new InvalidOperationException(""Differing deadletter error descriptions cannot be specified for both the 'propertiesToModify' and 'deadLetterErrorDescription' parameters. The values should either be identical or only be specified in one of the parameters."");
```

JoshLove-msft (approved):
Looks great! Just a few remaining minor comments.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Receiver/ServiceBusReceiver.cs
```diff
@@ -683,6 +686,86 @@ public virtual async Task CloseAsync(CancellationToken cancellationToken = defau
...
+            Argument.AssertNotNull(propertiesToModify, nameof(propertiesToModify));
+
+            // Prevent properties and arguments from setting distinct deadletter reasons or error descriptions
+            bool containsReasonHeader = propertiesToModify.TryGetValue(AmqpMessageConstants.DeadLetterReasonHeader, out object reasonHeaderProperty);
+            bool containsDescriptionHeader = propertiesToModify.TryGetValue(AmqpMessageConstants.DeadLetterErrorDescriptionHeader, out object descriptionHeaderProperty);
+
+            bool setsReasonHeaderTwice = containsReasonHeader && deadLetterReason != null;
+            bool setsDescriptionHeaderTwice = containsDescriptionHeader && deadLetterErrorDescription != null;
+
+            if (setsReasonHeaderTwice && (reasonHeaderProperty is not string || reasonHeaderProperty.ToString() != deadLetterReason))
```
m-redding, line 751: I was getting build errors here without the `.ToString()` but with this change it shouldn't even get evaluated unless it is a string already

JoshLove-msft: Ah yes, I guess we can't use `==` but would instead have to do `.Equals`. I think what you have is more clear anyway.

PR #30765, opened 2022-08-28
Stop defaulting region as it breaks UsGov/China resource deployments

JoshLove-msft:
We hardcoded the region to westus3 when the default was changed because we were seeing latency and flaky tests. This broke the weekly test pipelines since the default region doesn't get picked up for other clouds. Testing out removing the hardcoded westus3 to see if tests are stable.


jsquire (approved):

benbp:
@JoshLove-msft we support per cloud locations as well if it's needed in the future, e.g. https://dev.azure.com/azure-sdk/internal/_wiki/wikis/internal.wiki/389/Configuring-Tests?anchor=test-environment-architecture

JoshLove-msft:
> @JoshLove-msft we support per cloud locations as well if it's needed in the future, e.g. https://dev.azure.com/azure-sdk/internal/_wiki/wikis/internal.wiki/389/Configuring-Tests?anchor=test-environment-architecture

Good to know!

PR #30775, opened 2022-08-29
[Logic] Add test cases

dvbb:


dvbb:
Linked issue: https://github.com/Azure/azure-rest-api-specs/issues/20428

sdk/logic/Azure.ResourceManager.Logic/src/autorest.md
```diff
@@ -168,8 +168,6 @@ format-by-name-rules:
   'trackingId': 'uuid'
   'actionTrackingId': 'uuid'
   'PublicCertificate': 'any'
-  'content': 'any'
```
ArthurMa1978, line 171: `BinaryData` support string format, don't change this, if you got serialization issue please add double quotation marks

sdk/logic/Azure.ResourceManager.Logic/src/autorest.md
```diff
@@ -221,7 +219,6 @@ directive:
       $.OpenAuthenticationAccessPolicies.properties.policies['x-ms-client-name'] = 'AccessPolicies';
       $.ResourceReference.properties.id['x-ms-format'] = 'arm-id';
       $.ResourceReference.properties.type['x-ms-format'] = 'resource-type';
-      $.IpAddress.properties.address['x-ms-format'] = 'ip-address';
```
ArthurMa1978, line 224: Not valid ip address? or null, or empty string?

dvbb: Service will return `address=20.98.144.224/27`, therefore will throw  Ip parse exception

sdk/logic/Azure.ResourceManager.Logic/src/autorest.md
dvbb: these properties will throw JSON serialization exceptions because their content is an XML format string, like that [SWAGGER](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/logic/resource-manager/Microsoft.Logic/stable/2019-05-01/examples/IntegrationAccountMaps_CreateOrUpdate.json#L11)

``` 
 System.Text.Json.JsonReaderException : '<' is an invalid start of a value. LineNumber: 0 | BytePositionInLine: 0.
```

dvbb: The SDK converts the read string in this logic below, but always throws an exception when converting an XML string
```
JsonSerializer.Serialize(writer, JsonDocument.Parse(Content.ToString()).RootElement);
``` 
#### example

```
var str = "value"
BinaryData Content = BinaryData.FromString($"\"{str}\"")
```
> Payload:  "content": "value"

```
var str = "<?xml version=\"2.0\" encoding=\"utf-8\" ?> \r\n"
BinaryData Content = BinaryData.FromString($"\"{str}\"")
```
> System.Text.Json.JsonReaderException : '2' is invalid after a single JSON value. Expected end of data. 

```
var str = "<?xml version='2.0' encoding='utf-8' ?>\r\n";
BinaryData Content = BinaryData.FromString($"\"{str}\"")
```
> System.Text.Json.JsonReaderException : '0x0D' is invalid within a JSON string. The string should be correctly escaped.

Yao725: Could you please try this sample and see if it can work?
```csharp
string val = "<?xml version=\"2.0\" encoding=\"utf-8\" ?> \r\n";
BinaryData data = BinaryData.FromObjectAsJson<string>(val);
```

dvbb: @Yao725 Thank you! That works

sdk/logic/Azure.ResourceManager.Logic/src/autorest.md
ArthurMa1978: Got it, please rename the name to `CidrAddress`, thank you.

ArthurMa1978 (approved):

PR #30820, opened 2022-08-29
Set Statsbeat connection string based on customer's endpoint

vishweshbankwar:


sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/Statsbeat/Statsbeat.cs
```diff
@@ -41,6 +43,22 @@ internal static class Statsbeat
 
         internal static MeterProvider s_attachStatsBeatMeterProvider;
 
+        internal static Regex s_endpoint_pattern = new("^https?://(?:www\\.)?([^/.-]+)");
+
+        internal static readonly HashSet<string> EU_Endpoints = new()
```
reyang, line 48: How would this be maintained in the future? E.g. if there a new Azure data center added in EU, do we need to release a new version of the exporter? What about customers who still use the old version of the exporter? Do we have EUDB violation in this case?

reyang (dismissed):
https://github.com/Azure/azure-sdk-for-net/pull/30820/files#r957777722

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/Statsbeat/Statsbeat.cs
vishweshbankwar: @heyams - Could you please provide input here?

heyams: if there is new azure data center added in EU, you will need to add it to EU_Endpoints. So yes, in that case exporter needs to be updated.
Unfortunately, Breeze ingestion service doesn't support this for us today. But in the near future, we might need to talk to Zaki's team if the new ingestion service can start redirecting ingestion endpoint for us for Statsbeat ikeys in any regions. There was a verbal promise but nothing substantial yet.

reyang: Thanks @heyams! Do you know who is making the decision regarding:

1. Whether to hard code the current EU endpoints.
2. If there are any new EU endpoints added, would existing exporters be considered as violation or not.

heyams: > Thanks @heyams! Do you know who is making the decision regarding:
> 
> 1. Whether to hard code the current EU endpoints.
**[Helen] sorry, just me after consulting with Ingestion Team. Hardcoding it seems to be the easiest and simplest.**
> 2. If there are any new EU endpoints added, would existing exporters be considered as violation or not.
**[Helen] Statsbeat ikey for EU is currently in north europe region. It would work for any new EU endpoints. When a new EU endpoint exists, our exporters will try to send statsbeat to the north america regioned Statsbeat ikey, which will not be allowed. In that case, we won't have statsbeat data for the new EU region. I don't think it will be considered as violation. Until we add the new EU endpoint to our list of EU_endpoints, it will forward data to north europe Statsbeat ikey.**

reyang: > [Helen] sorry, just me after consulting with Ingestion Team. Hardcoding it seems to be the easiest and simplest.

It might be "easy" for now, with the cost in the long run.

> [Helen] Statsbeat ikey for EU is currently in north europe region. It would work for any new EU endpoints. When a new EU endpoint exists, our exporters will try to send statsbeat to the north america regioned Statsbeat ikey, which will not be allowed. In that case, we won't have statsbeat data for the new EU region. I don't think it will be considered as violation. Until we add the new EU endpoint to our list of EU_endpoints, it will forward data to north europe Statsbeat ikey.

Do we know if EUDB is about the action that "sends the data" or the fact that "the data is received"? If the SDK attempted to send the data, and the data got rejected due to EUDB violation, is the SDK considered as EUDB compliant or not?

heyams: @reyang to your second comment, i don't think it's a violation. Java received IcM tickets regarding this issue, that's when we started implementing EU regions support. Statsbeat telemetry was failing, and customers observed the error in the log and then reported it. we didn't get flagged as EUDB violation/compliant AFAIK.

to your 1st comment, when the new ingestion service becomes available, we will reengage with Zaki's team to propose a long-term solution, which would support endpoint redirect for EU region and private endpoints (not supported today).

reyang: Got it, unblocking the PR. I suggest putting these design considerations in the spec.

heyams: good point, i can create a separate spec for it and propose it to Ingestion Team when it's a good timing. They don't have capacity to consider this yet.

reyang: > @reyang to your second comment, i don't think it's a violation.

Why do you think so? Is there any document or reference?

@trask in case you know more about this.

trask: > Do we know if EUDB is about the action that "sends the data" or the fact that "the data is received"? If the SDK attempted to send the data, and the data got rejected due to EUDB violation, is the SDK considered as EUDB compliant or not?

My understanding (which should not be considered in any way authoritative on this topic) is that the ingestion service does not reject data based on EUDB.

And if I had to guess, I'd say that the data crossing the network would violate EUDB, whether it's accepted or not.

vishweshbankwar: @heyams - Could you please clarify this part in the specification and get it reviewed by ingestion team?

heyams: sure, i will do that.

heyams: https://github.com/microsoft/Telemetry-Collection-Spec/pull/113

cijothomas (changes requested):
https://github.com/Azure/azure-sdk-for-net/pull/30820/files#r957777722 Need a resolution to this before merging.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/Statsbeat/Statsbeat.cs
```diff
@@ -15,6 +17,8 @@ namespace Azure.Monitor.OpenTelemetry.Exporter.Internals
 {
     internal static class Statsbeat
     {
+        private static bool s_isEnabled = true;
```
rajkumar-rangaraj, line 20: nit: Consider rearranging the variable in order of const, readonly, members.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/Statsbeat/Statsbeat.cs
```diff
@@ -41,6 +45,61 @@ internal static class Statsbeat
 
         internal static MeterProvider s_attachStatsBeatMeterProvider;
 
+        internal static Regex s_endpoint_pattern = new("^https?://(?:www\\.)?([^/.-]+)");
+
+        internal static readonly HashSet<string> EU_Endpoints = new()
+        {
+            "westeurope",
```
rajkumar-rangaraj, line 52: nit: arrange in alphabetical order, later it would be easy to check and update the list.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/Statsbeat/Statsbeat.cs
```diff
@@ -61,11 +120,20 @@ private static string GetOS()
 
         internal static void InitializeAttachStatsbeat(string connectionString)
         {
-            if (s_attachStatsBeatMeterProvider == null)
+            // check whether it is disabled or already initialized.
+            if (s_isEnabled && s_attachStatsBeatMeterProvider == null)
```
rajkumar-rangaraj, line 124: Why are we using two parameter check here. Do we need `s_isEnabled`?

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/StatsbeatTests.cs
```diff
@@ -0,0 +1,56 @@
...
+using System.Threading.Tasks;
+using Azure.Monitor.OpenTelemetry.Exporter.Internals;
+using Azure.Monitor.OpenTelemetry.Exporter.Internals.ConnectionString;
+using Xunit;
+
+namespace Azure.Monitor.OpenTelemetry.Exporter.Tests
+{
+    public class StatsbeatTests
+    {
+        [Fact]
```
rajkumar-rangaraj, line 17: nit: Consider theory instead of Fact.

rajkumar-rangaraj (approved):
LGTM, left few suggestions.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/Statsbeat/Statsbeat.cs
```diff
@@ -43,6 +47,48 @@ internal static class Statsbeat
...
+            "southafricanorth",
+            "southcentralus",
+            "southeastasia",
+            "southindia",
+            "uaecentral",
+            "uaenorth",
+            "westus",
+            "westus2",
+            "westus3",
+        };
```
TimothyMothra, line 90: nit: The EU list is one line, and the non-EU list is multiline.
I would recommend formatting these the same to make them easier to read and maintain.

Second, consider moving these to a new file: Statsbeat/Constants.cs

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/Statsbeat/Statsbeat.cs
```diff
@@ -63,11 +109,20 @@ private static string GetOS()
 
         internal static void InitializeAttachStatsbeat(string connectionString)
         {
-            if (s_attachStatsBeatMeterProvider == null)
+            // check whether it is disabled or already initialized.
+            if (s_isEnabled && s_attachStatsBeatMeterProvider == null)
```
TimothyMothra, line 113: The fact that this _could_ be initialized twice seems wrong. 
Will each exporter have a unique StatsBeat instance, or will the entire SDK have a single StatsBeat instance?

Consider that some customers will send telemetry to different IKeys.

What happens if a root customer has their own customers. The root customer may have an ikey, and the nested customers may have a unique ikey. I don't think we can assume that each customer will be in the same region.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/StatsbeatTests.cs
```diff
@@ -0,0 +1,101 @@
...
+        [InlineData("uksouth")]
+        [InlineData("ukwest")]
+        [InlineData("westeurope")]
+        public void StatsbeatConnectionStringIsSetBasedOnCustomersConnectionStringEndpointInEU(string euEndpoint)
+        {
+            var customer_ConnectionString = $"InstrumentationKey=1aa11111-bbbb-1ccc-8ddd-eeeeffff3333;IngestionEndpoint=https://{euEndpoint}.in.applicationinsights.azure.com/";
+            var parsedConectionString = ConnectionStringParser.GetValues(customer_ConnectionString);
+
+            // Set ConnectionString to null as it is static and may be impacted by other tests.
+            Statsbeat.s_statsBeat_ConnectionString = null;
```
TimothyMothra, line 36: This is a red flag for me. 
We should be able to initialize and test the Statsbeat class without affecting other tests.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/StatsbeatTests.cs
```diff
@@ -0,0 +1,101 @@
...
+        [InlineData("qatarcentral")]
+        [InlineData("southafricanorth")]
+        [InlineData("southcentralus")]
+        [InlineData("southeastasia")]
+        [InlineData("southindia")]
+        [InlineData("uaecentral")]
+        [InlineData("uaenorth")]
+        [InlineData("westus")]
+        [InlineData("westus2")]
+        [InlineData("westus3")]
```
TimothyMothra, line 75: I think InlineData can accept an array of values. May be able to provide the constant from the other class.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/StatsbeatTests.cs
```diff
@@ -0,0 +1,101 @@
...
+        public void StatsbeatIsNotInitializedForUnknownRegions()
+        {
+            var customer_ConnectionString = "InstrumentationKey=1aa11111-bbbb-1ccc-8ddd-eeeeffff3333;IngestionEndpoint=https://foo.in.applicationinsights.azure.com/";
+            var parsedConectionString = ConnectionStringParser.GetValues(customer_ConnectionString);
+
+            // Set ConnectionString to null as it is static and may be impacted by other tests.
+            Statsbeat.s_statsBeat_ConnectionString = null;
+            Statsbeat.SetStatsbeatConnectionString(parsedConectionString.IngestionEndpoint);
+
+            Assert.Null(Statsbeat.s_statsBeat_ConnectionString);
```
TimothyMothra, line 98: can this also assert that isEnabled is false?

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/StatsbeatTests.cs
rajkumar-rangaraj: Best practice is to use constructor and dispose to set `Statsbeat.s_statsBeat_ConnectionString` to null.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/Statsbeat/Statsbeat.cs
vishweshbankwar: There will be only one instance of StatsBeat.

```Consider that some customers will send telemetry to different IKeys.

What happens if a root customer has their own customers. The root customer may have an ikey, and the nested customers may have a unique ikey. I don't think we can assume that each customer will be in the same region.```

This ^ should not be an issue as far as all those ikeys are in EU or NonEU regions.

We will not add Statsbeat if customer tries to to use connection string from multiple regions i.e. one from EU and other from NonEU. I will fix this part in follow up PR.

File: sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/StatsbeatTests.cs (hunk 9 above)
Comment (vishweshbankwar) [reply sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/StatsbeatTests.cs:75]: We don't want the array here right? It needs to be tested one by one. I could have a loop inside the test but that will look odd. Let me know if I am missing something. I will follow up on that with different PR.

File: sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/StatsbeatTests.cs (hunk 8 above)
Comment (vishweshbankwar) [reply sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/StatsbeatTests.cs:36]: by other tests it is referring to tests related to Statsbeat class itself. I have updated it to make it more clear. I will explore the dispose option in a separate follow up as that may touch the meterprovider part as well.

File: sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/StatsbeatTests.cs (hunk 10 above)
Comment (vishweshbankwar) [reply sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/StatsbeatTests.cs:98]: s_isEnabled is private.

File: sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/StatsbeatTests.cs (hunk 9 above)
Comment (rajkumar-rangaraj) [reply sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/StatsbeatTests.cs:76]: Something like this one - https://github.com/Azure/azure-sdk-for-net/blob/045f589c736a6c76af682c3ac01f684b8da93a5e/sdk/core/Azure.Core/tests/RedirectPolicyTests.cs#L21

Comment (rajkumar-rangaraj) [reply sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/StatsbeatTests.cs:76]: e.g. https://andrewlock.net/creating-parameterised-tests-in-xunit-with-inlinedata-classdata-and-memberdata/#loading-data-from-a-property-on-the-test-class

**Comment by heyams** [comment]
```
   internal const string StatsBeat_ConnectionString_NonEU = "<Non-EU-ConnectionString>";
   internal const string StatsBeat_ConnectionString_EU = "EU-ConnectionString";
```

one with angle bracket and one doesn't.. i wonder how you initialize its value and assume this is just a placeholder.

**Comment by vishweshbankwar** [comment]
> ```
>    internal const string StatsBeat_ConnectionString_NonEU = "<Non-EU-ConnectionString>";
>    internal const string StatsBeat_ConnectionString_EU = "EU-ConnectionString";
> ```
> 
> one with angle bracket and one doesn't.. i wonder how you initialize its value and assume this is just a placeholder.

just a placeholder. Will replace the value with actual ones before enabling.

File: sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/StatsbeatTests.cs [hunk 11]
```diff
@@ -0,0 +1,104 @@
...
+        [InlineData("switzerlandwest")]
+        [InlineData("uksouth")]
+        [InlineData("ukwest")]
+        [InlineData("westeurope")]
+        public void StatsbeatConnectionStringIsSetBasedOnCustomersConnectionStringEndpointInEU(string euEndpoint)
+        {
+            var customer_ConnectionString = $"InstrumentationKey=1aa11111-bbbb-1ccc-8ddd-eeeeffff3333;IngestionEndpoint=https://{euEndpoint}.in.applicationinsights.azure.com/";
+            var parsedConectionString = ConnectionStringParser.GetValues(customer_ConnectionString);
+
+            Statsbeat.SetStatsbeatConnectionString(parsedConectionString.IngestionEndpoint);
```
cijothomas, line 35: would recommend doing an more e2e test for this to simulate what a end user would do. They simply create tracer/loger/meter builder with azmon exporter added. Lets just do that and validate statsbeat got initialized with right endpoint.

cijothomas (approved):
left a suggestion about better test coverage of the feature.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/Statsbeat/Statsbeat.cs
vishweshbankwar: will do as a follow up.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/StatsbeatTests.cs
vishweshbankwar: Ran in to some issue, will do this as a follow up.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/StatsbeatTests.cs
vishweshbankwar: We are not enabling the statsbeat feature for next release. I will add this e2e test once we enable it.

PR #30836, opened 2022-08-30
Address review comments for HDInsight, ResourceMover, NetApp and ConfidentialLedger

ArthurMa1978:


sdk/hdinsight/Azure.ResourceManager.HDInsight/src/Generated/Models/HDInsightIPConfiguration.cs
```diff
@@ -54,7 +54,7 @@ internal HDInsightIPConfiguration(ResourceIdentifier id, string name, ResourceId
         /// <summary> The name of private link IP configuration. </summary>
         public string Name { get; set; }
         /// <summary> The type of the private link IP configuration. </summary>
-        public ResourceIdentifier ResourceType { get; }
+        public ResourceType? ResourceType { get; }
```
Yao725, line 57: It seems that `HDInsightIPConfiguration` is a data model but it doesn't inherit `ResourceData` because of the setter of the property `Name`. Do you think if we need to generate its own `HDInsightResourceData` model within this RP and let `HDInsightIPConfiguration` inherits `HDInsightResourceData`?

Yao725 (approved):

sdk/hdinsight/Azure.ResourceManager.HDInsight/src/Generated/Models/HDInsightIPConfiguration.cs
ArthurMa1978: This is not a resource, but part of privatelink, so not inherit from `ResourceData` is right

PR #30844, opened 2022-08-30
Perf fixes for session processor

JoshLove-msft:
- ConcurrentExpiringSet was not being disposed when accepting a session timed out - this had a big impact when using session processor with high concurrency, as we would continually scan the set for expired messages every 30 seconds until being garbage collected. 
- When creating a SessionReceiverManager, we were also creating an underlying Receiver which was not necessary - this exacerbated the ConcurrentExpiringSet issue, because we'd have an additional N ConcurrentExpiringSet instances where N = MaxConcurrentSessions.
- Respect the acceptSession semaphore _before_ dispatching tasks to SessionReceiverManagers - this avoids creating a bunch of queued tasks that are just going to be waiting on the acceptSession semaphore anyway

### DIFF
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpReceiver.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpReceiver.cs
@@ -768,6 +768,8 @@ namespace Azure.Messaging.ServiceBus.Amqp
             string deadLetterReason = null,
             string deadLetterDescription = null)
         {
+            ThrowIfSessionLockLost();
+
             // Create an AmqpRequest Message to update disposition
             var amqpRequestMessage = AmqpRequestMessage.CreateRequest(ManagementConstants.Operations.UpdateDispositionOperation, timeout, null);
 
@@ -1348,6 +1350,13 @@ namespace Azure.Messaging.ServiceBus.Amqp
                         innerException: exception);
                 }
                 LinkException = exception;
+            }
+
+            if (IsSessionLinkClosed)
+            {
+                // Clean up and dispose the underlying resources. The receive link should already be closed, but management link may need to be
+                // closed as well.
+                _ = CloseAsync(CancellationToken.None);
             }
 
             ServiceBusEventSource.Log.ReceiveLinkClosed(
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpReceiver.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpReceiver.cs
@@ -111,7 +111,7 @@ namespace Azure.Messaging.ServiceBus.Amqp
         /// <summary>
         /// A map of locked messages received using the management client.
         /// </summary>
-        private readonly ConcurrentExpiringSet<Guid> _requestResponseLockedMessages;
+        internal readonly ConcurrentExpiringSet<Guid> RequestResponseLockedMessages;
 
         private static IReadOnlyList<ServiceBusReceivedMessage> s_backingEmptyList;
         private readonly bool _isProcessor;
@@ -173,7 +173,7 @@ namespace Azure.Messaging.ServiceBus.Amqp
             _isProcessor = isProcessor;
             _receiveMode = receiveMode;
             Identifier = identifier;
-            _requestResponseLockedMessages = new ConcurrentExpiringSet<Guid>();
+            RequestResponseLockedMessages = new ConcurrentExpiringSet<Guid>();
             SessionId = sessionId;
 
             _receiveLink = new FaultTolerantAmqpObject<ReceivingAmqpLink>(
@@ -416,7 +416,7 @@ namespace Azure.Messaging.ServiceBus.Amqp
             Guid lockToken,
             TimeSpan timeout)
         {
-            if (_requestResponseLockedMessages.Contains(lockToken))
+            if (RequestResponseLockedMessages.Contains(lockToken))
             {
                 await DisposeMessageRequestResponseAsync(
                     lockToken,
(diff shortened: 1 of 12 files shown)


sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpReceiver.cs
```diff
@@ -1348,13 +1350,6 @@ private void OnReceiverLinkClosed(object receiver, EventArgs e)
                 LinkException = exception;
             }
 
-            if (IsSessionLinkClosed)
```
JoshLove-msft, line 1351: Even if the underlying session link closed, the user still needs to dispose the receiver.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessor.cs
```diff
@@ -828,21 +829,28 @@ private async Task RunReceiveTaskAsync(CancellationToken cancellationToken)
                             break;
                         }
 
-                        // Do a quick synchronous check before we resort to async/await with the state-machine overhead.
-                        if (!_messageHandlerSemaphore.Wait(0, CancellationToken.None))
+                        bool messageHandlerLockAcquired = false;
```
jsquire, line 832: Why remove the quick sync check?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessor.cs
```diff
@@ -828,21 +829,28 @@ private async Task RunReceiveTaskAsync(CancellationToken cancellationToken)
...
                             {
-                                linkedHandlerTcs.Dispose();
-                                // reset the linkedHandlerTcs if it was already cancelled due to user updating the concurrency
-                                linkedHandlerTcs = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, _handlerCts.Token);
-                                // allow the loop to wake up when tcs is signaled
-                                break;
+                                // make sure to release semaphore if we are breaking out of the loop
+                                _messageHandlerSemaphore.Release();
                             }
+                            linkedHandlerTcs.Dispose();
```
jsquire, line 849: Should we consider calling `TrySetCancelled` here or do we know that nobody is awaiting?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/SessionReceiverManager.cs
```diff
@@ -279,15 +269,18 @@ public override async Task ReceiveAndProcessMessagesAsync(CancellationToken proc
...
-                when (ex.Reason == ServiceBusFailureReason.ServiceTimeout)
+                    when (ex.Reason == ServiceBusFailureReason.ServiceTimeout)
                 {
                     // these exceptions are expected when no messages are available
                     // so simply return and allow this to be tried again on next thread
                     return;
                 }
+                finally
+                {
+                    _concurrentAcceptSessionsSemaphore.Release();
```
jsquire, line 280: The class state mutation here makes me a bit uncomfortable.   Is there any way that we can have whomever owns semaphore acquisition also have responsibility for release?

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Processor/SessionProcessorLiveTests.cs
```diff
@@ -2306,6 +2306,86 @@ async Task ProcessMessage(ProcessSessionMessageEventArgs args)
...
+                            {
+                                messages.Add(new ServiceBusMessage("Hello World") { SessionId = sessionId });
+                            }
+                        }
+
+                        await sender.ScheduleMessagesAsync(messages, DateTimeOffset.Now.AddSeconds(10));
+                    }
+                }
+
+                Task ProcessMessageAsync(ProcessSessionMessageEventArgs args)
```
jsquire, line 2375: Should we assert that we've been called at least once and, ideally, the expected number of times?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessor.cs
JoshLove-msft: It was mostly just for code readability - now that we have two semaphores that we are locking it becomes messy to handle the sync check. I will look into adding this back for both semaphores.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/SessionReceiverManager.cs
JoshLove-msft: Unfortunately no, because we need to release the lock as soon as a session has been accepted or we determine that we don't need to accept a new session.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpReceiver.cs
JoshLove-msft: Reverted this change as it will be too confusing for users.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessor.cs
JoshLove-msft: Nobody is awaiting, this loop is single-threaded.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/SessionReceiverManager.cs
JoshLove-msft: Open to other ideas though.

jsquire: The best that I have is to comment around it.

jsquire (approved):

Issue #25434, opened 2021-11-19
Azure.Data.Tables - CreateTableIfNotExists throws System.NullReferenceException: 'Object reference not set to an instance of an object.'


hgjura:
Hello,
I am just writing some sample code to test the Azure.Data.Tables functionality and right away I ran into an exception.

Code is rather simple:
```
var serviceClient = new TableServiceClient(new Uri("<some uri>"), new TableSharedKeyCredential("<some account name>", "<some key>"));

            string tableName = "TESTDATES";
            TableItem table = serviceClient.CreateTableIfNotExists(tableName);

            var tc = serviceClient.GetTableClient(tableName);
```
The first iteration, when table does not exists, it runs fine. Subsequent iterations, when table already exists, it throws an excpetion in the CreateTableIfNotExists() line.

Exceptions details:
System.NullReferenceException
  HResult=0x80004003
  Message=Object reference not set to an instance of an object.
  Source=Azure.Core
  StackTrace:
   at Azure.Response`1.op_Implicit(Response`1 response)
   at ConsoleApp_TestingDatesInAzure.Program.Main(String[] args) in 

There is no inner exception.

Thanks
Herald

jsquire:
Thank you for your feedback.  Tagging and routing to the team member best able to assist.

christothes:
Hi @hgjura - Could you confirm that the URI being passed to the TableClient constructor does not contain a table name? It should be just the account URI.

hgjura:
Here is a screenshot of the code. It is pretty straightforward. And as you mentioned only the acct uri and the key.

![Capture](https://user-images.githubusercontent.com/9681487/143144013-32dcdaa8-d4c4-482c-acb9-b89497012ae4.PNG)

Thanks
Herald

christothes:
Are you using the latest released version? I am unable to reproduce this.

christothes:
I looked at this again and I realized that the issue is related to the implicit cast for Response<T>. The workaround for this is to either use `var` for the response type, or explicitly `Response<TableItem>`

I'll look into a viable fix for this as well. 

related to #25528

hgjura:
Good catch. I didn't think that response will be null if table exists, but it is indeed null. Thanks for the workaround!

pakrym:
Can we synthesize an instance of TableItem and return it instead of null? It has only the Name property that we have.

PR #30849, opened 2022-08-30
CreateIfNotExistsAsync does not throw or log error when table already exists

### LINKED ISSUE #25561: [BUG] TableClient.Delete/Async documentation incorrectly describes return value
The ref docs at https://docs.microsoft.com/en-us/dotnet/api/azure.data.tables.tableclient.deleteasync?view=azure-dotnet#returns say `null` will be returned when trying to delete a table that doesn't exist, but per https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/tables/Azure.Data.Tables/src/TableClient.cs#L530 we're returning the 404 response.

### LINKED ISSUE #27821: TableClient.CreateIfNotExists should not return null
### Library name

Azure.Data.Tables

Currently CreateIfNotExists returns null when the table already exists. This causes a NullReferenceException when an implict cast is attempted by Response<T>.

related to #25434

### LINKED ISSUE #28084: [QUERY] Tables CreateIfNotExists Error Logs
### Library name and version

Azure.Data.Tables 12.5.0

### Query/Question

Why does the following code, throw a 409 (conflict) if the table exists. The purpose of this call is to create a table if it dosn't exists. This is not an exception case. It leads to a ton of error messages in application insight (dependency Exceptions). It seems that it is handled by the library, so it does not throw it to the client but it sillt fills the logs. Is this fixed in a further release?

      
`tableClient = new TableClient(new Uri("tableuri"), "mytablename", new DefaultAzureCredential());
        await tableClient.CreateIfNotExistsAsync()
            .ConfigureAwait(false);`

### Environment

Azure Functions dotnet 6

christothes:
fixes #27821
fixes #28084
fixes #25561

### DIFF
--- a/sdk/tables/Azure.Data.Tables/src/TableClient.cs
+++ b/sdk/tables/Azure.Data.Tables/src/TableClient.cs
@@ -30,9 +30,7 @@ namespace Azure.Data.Tables
         private readonly string _version;
         private readonly bool _isCosmosEndpoint;
         private readonly ResponseFormat _returnNoContent = ResponseFormat.ReturnNoContent;
-        private const string ReturnNoContent = "return-no-content";
         private readonly QueryOptions _defaultQueryOptions = new() { Format = OdataMetadataFormat.ApplicationJsonOdataMinimalmetadata };
-        private const string MinimalMetadata = "application/json;odata=minimalmetadata";
         private string _accountName;
         private readonly Uri _endpoint;
         private Guid? _batchGuid;
@@ -482,8 +480,8 @@ namespace Azure.Data.Tables
                 context.AddClassifier((int)HttpStatusCode.Conflict, false);
                 var response = _tableOperations.Create(
                     RequestContent.Create(new { TableName = Name }),
-                    MinimalMetadata,
-                    ReturnNoContent,
+                    TableConstants.Odata.MinimalMetadata,
+                    TableConstants.ReturnNoContent,
                     context);
 
                 if (response.IsError || response.Status == (int)HttpStatusCode.Conflict)
@@ -518,8 +516,8 @@ namespace Azure.Data.Tables
                 var context = new RequestContext { CancellationToken = cancellationToken, ErrorOptions = ErrorOptions.NoThrow };
                 var response = await _tableOperations.CreateAsync(
                     RequestContent.Create(new { TableName = Name }),
-                    MinimalMetadata,
-                    ReturnNoContent,
+                    TableConstants.Odata.MinimalMetadata,
+                    TableConstants.ReturnNoContent,
                     context).ConfigureAwait(false);
 
                 if (response.IsError)
@@ -550,7 +548,7 @@ namespace Azure.Data.Tables
             scope.Start();
             try
             {
-                return _tableOperations.Delete(Name, CreateContextForDelete(cancellationToken));
+                return _tableOperations.Delete(Name, TableServiceClient.CreateContextForDelete(cancellationToken));
             }
             catch (Exception ex)
             {
@@ -570,20 +568,13 @@ namespace Azure.Data.Tables
             scope.Start();
             try
             {
-                return await _tableOperations.DeleteAsync(Name, CreateContextForDelete(cancellationToken)).ConfigureAwait(false);
+                return await _tableOperations.DeleteAsync(Name, TableServiceClient.CreateContextForDelete(cancellationToken)).ConfigureAwait(false);
             }
             catch (Exception ex)
             {
                 scope.Failed(ex);
                 throw;
             }
-        }
-
-        private static RequestContext CreateContextForDelete(CancellationToken cancellationToken)
-        {
-            var context = new RequestContext() { CancellationToken = cancellationToken };
-            context.AddClassifier(404, false);
-            return context;
         }
 
         /// <summary>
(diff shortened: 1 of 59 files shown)


sdk/tables/Azure.Data.Tables/src/TableClient.cs
```diff
@@ -471,16 +476,23 @@ public virtual Response<TableItem> CreateIfNotExists(CancellationToken cancellat
             scope.Start();
             try
             {
+                var context = new RequestContext { CancellationToken = cancellationToken, ErrorOptions = ErrorOptions.NoThrow };
```
annelo-msft, line 479: If CancellationToken is unset, you can potentially use a static and avoid an allocation here.

sdk/tables/Azure.Data.Tables/src/TableClient.cs
```diff
@@ -471,16 +476,23 @@ public virtual Response<TableItem> CreateIfNotExists(CancellationToken cancellat
...
-                    new TableProperties { TableName = Name },
-                    null,
-                    _defaultQueryOptions,
-                    cancellationToken);
-                return Response.FromValue(response.Value as TableItem, response.GetRawResponse());
-            }
-            catch (RequestFailedException ex) when (ex.Status == (int)HttpStatusCode.Conflict && ex.ErrorCode != TableErrorCode.TableBeingDeleted)
-            {
-                return default;
+                    RequestContent.Create(new { TableName = Name }),
```
annelo-msft, line 482: I think this makes a few allocations by creating the anonymous type and then the RequestContent and invokes the ObjectSerializer, which is a perf hit - if perf is a concern here, it might be worth looking into writing this to a desired RequestContent subclass directly.

sdk/tables/Azure.Data.Tables/src/TableClient.cs
```diff
@@ -524,24 +541,14 @@ public virtual async Task<Response<TableItem>> CreateIfNotExistsAsync(Cancellati
         /// Deletes the table specified by the tableName parameter used to construct this client instance.
         /// </summary>
         /// <param name="cancellationToken">A <see cref="CancellationToken"/> controlling the request lifetime.</param>
-        /// <returns>If the table exists, a <see cref="Response"/>. If the table already does not exist, <c>null</c>.</returns>
+        /// <returns>The <see cref="Response"/> indicating the result of the operation.</returns>
```
annelo-msft, line 544: Is a behavioral change in the return value?

annelo-msft:
Would it be possible to add to the description what the PR is doing?

sdk/tables/Azure.Data.Tables/tests/TableClientTests.cs
```diff
@@ -426,12 +426,7 @@ public async Task ValidateUri()
         [Test]
         public void CreateIfNotExistsThrowsWhenTableBeingDeleted()
         {
-            _transport = new MockTransport(
-                request => throw new RequestFailedException(
-                    (int)HttpStatusCode.Conflict,
-                    null,
-                    TableErrorCode.TableBeingDeleted.ToString(),
-                    null));
+            _transport = TableAlreadyExistsTransport(TableErrorCode.TableBeingDeleted);
```
annelo-msft, line 429: Where do you validate that errors are not logged/error traces are not sent?

sdk/tables/Azure.Data.Tables/src/TableServiceClient.cs
```diff
@@ -980,6 +973,13 @@ internal static bool IsPremiumEndpoint(Uri endpoint)
         /// <returns>A valid OData filter expression.</returns>
         public static string CreateQueryFilter(FormattableString filter) => TableOdataFilter.Create(filter);
 
+        internal static RequestContext CreateContextForDelete(CancellationToken cancellationToken)
+        {
+            var context = new RequestContext() { CancellationToken = cancellationToken };
+            context.AddClassifier(404, false);
```
annelo-msft, line 979: Nit: why use 404 here and HttpStatusCode.Xx elsewhere?

sdk/tables/Azure.Data.Tables/src/TableClient.cs
christothes: > it might be worth looking into writing this to a desired RequestContent subclass directly.

Is there an existing example of this pattern somewhere I could refer to?

sdk/tables/Azure.Data.Tables/src/TableClient.cs
christothes: Yes, but intentional. The way we return null today can cause a very strange NRE. https://github.com/Azure/azure-sdk-for-net/issues/25434

sdk/tables/Azure.Data.Tables/tests/TableClientTests.cs
christothes: My assumption here is that this is guaranteed by the `RequestContext.AddClassifier` behavior. I think I'd have to test the Core event source to validate either way.

sdk/tables/Azure.Data.Tables/src/TableClient.cs
christothes: realized that there is no reason to use an anonymous type here since we still have models :) 

Since this is typically only ever called once per client lifetime, I'm good with not micro-optimizing further.

sdk/tables/Azure.Data.Tables/src/TableClient.cs
christothes: This is typically only called once per client lifetime, so I think I'll leave it as is for now.

sdk/tables/Azure.Data.Tables/tests/TableClientTests.cs
jsquire: Validating the event source may be more trouble than it's worth, given that you'd have to set this up in a non-parallelized test, use the helper from the test framework, and hope that no unrelated error was logged, since there doesn't seem to be anything specific to Tables that you could key on.

Personally, I'm good with trusting the classifier and assuming that it has robust test coverage.

jsquire (approved):

Issue #30185, opened 2022-07-28
[BUG] Boolean filter creates invalid OData string


cheng93:
### Library name and version

Azure.Data.Tables 12.6.1

### Describe the bug

```csharp
table.QueryAsync<Entity>(x => x.Enabled);
```

creates an invalid OData url
`[28/Jul/2022:10:27:11 +0000] "GET /devstoreaccount1/privatedefinitions()?$format=application%2Fjson%3Bodata%3Dminimalmetadata&$filter=Enabled HTTP/1.1" 400`

```csharp
table.QueryAsync<Entity>(x => x.Enabled == true);
```

is valid
`[28/Jul/2022:10:31:01 +0000] "GET /devstoreaccount1/privatedefinitions()?$format=application%2Fjson%3Bodata%3Dminimalmetadata&$filter=Enabled%20eq%20true HTTP/1.1" 200`

### Expected behavior

should produce a valid OData url

### Actual behavior

see above

### Reproduction Steps

see above

jsquire:
Thank you for your feedback.  Tagging and routing to the team member best able to assist.

cheng93:
Closed in #30863

PR #30863, opened 2022-08-31
Add the capability of rewrite implicit boolean comparisons to explicit comparisons

alfonsohdez08:
This PR intention is to address the issue [#30185](https://github.com/Azure/azure-sdk-for-net/issues/30185). The idea is rewrite expressions such as `ent => ent.BooleanProperty`  to `ent => ent.BooleanProperty == true` so the OData expression writer produces a valid filter. This rewrite not only work when accessing an object’s boolean property, but also when calling the `TableEntity.GetBoolean` method: `ent => ent.GetBoolean("BoolProperty").Value` would get rewritten to `ent => ent.GetBoolean("BoolProperty").Value == true`. I call that this proposal cover cases where type casting is part of an expression: `ent => (bool)ent.GetBoolean("BoolProperty")`.


christothes:
Thanks for the contribution, @alfonsohdez08! I'll take a look in a bit.

sdk/tables/Azure.Data.Tables/src/Queryable/ExpressionNormalizer.cs
```diff
@@ -144,6 +188,11 @@ internal Expression VisitMethodCallNoRewrite(MethodCallExpression call)
 
             if (ReflectionUtil.s_dictionaryMethodInfosHash.Contains(visited.Method) && visited.Arguments.Count == 1 && visited.Arguments[0] is ConstantExpression)
             {
+                if (visited.Method.ReturnType == typeof(bool?) && IsImplicitBooleanComparison(visited))
+                {
+                    return CreateExplicitBooleanComparison(visited);
+                }
+
```
christothes, line 195: I don't this this code is ever reached

christothes (commented):
Looks great! Just a minor comment. 

Would you mind also adding a changelog entry for this?

sdk/tables/Azure.Data.Tables/src/Queryable/ExpressionNormalizer.cs
alfonsohdez08: Oh yeah, that's right! I missed revisiting this piece of code after I performed the changes for rewrite expressions that at least has a type casting expression on it. That said, the type casting rewrites will cover the scenario when having expressions such as `ent => (bool)ent.GetBoolean("BoolProp")`.

alfonsohdez08:
Thank you, @christothes! And yeah, absolutely! I just updated the PR with the changelog entry as well.

christothes (approved):

christothes:
@alfonsohdez08  One more thing - Unfortunately, I missed this as part of the release that went out today, so would you mind updating the changelog to move this fix into the latest upcoming version section? Thanks!

alfonsohdez08:
@christothes yeah, sure! I just updated the changelog.

PR #30866, opened 2022-09-01
Modify type casting in instrumentors to support client hierarchy

### LINKED ISSUE #30875: [Polymorphic Resources] Resolve the test framework issues for polymorphic resources
Related work with https://github.com/Azure/autorest.csharp/issues/2562

m-nash:
This is to adjust the test framework to support this feature https://github.com/Azure/autorest.csharp/issues/2562

Fixes https://github.com/Azure/azure-sdk-for-net/issues/30875


sdk/core/Azure.Core.TestFramework/src/Instrumentation/DiagnosticScopeValidatingInterceptor.cs
```diff
@@ -88,6 +88,33 @@ internal static void WrapAsyncResultCore<T>(IInvocation invocation, Type generic
                 async ValueTask<T> Await()
                 {
                     invocation.Proceed();
+                    var tType = typeof(T);
+                    if (tType.IsGenericType && tType.GetGenericTypeDefinition().Equals(typeof(Response<>)))
+                    {
+                        //guaranteed only one generic arg with Response<T>
+                        var tGenericType = tType.GetGenericArguments()[0];
+                        var returnType = invocation.ReturnValue.GetType();
```
christothes, line 96: Can ReturnValue be null here?

m-nash: 99.9% sure it can't but I can add the extra check.  If it is the original cast wouldn't work

JoshLove-msft, line 96: nit: it might be worth calling this something like `runtimeReturnType`.

sdk/core/Azure.Core.TestFramework/src/Instrumentation/DiagnosticScopeValidatingInterceptor.cs
```diff
@@ -88,6 +88,33 @@ internal static void WrapAsyncResultCore<T>(IInvocation invocation, Type generic
...
+                        if (returnGenericType is not null && returnGenericType.IsSubclassOf(tGenericType))
+                        {
+                            //get the object
+                            var responseObject = TaskExtensions.GetResultFromTask(invocation.ReturnValue);
+                            var rawResponse = responseObject.GetType().GetMethod("GetRawResponse", BindingFlags.Instance | BindingFlags.Public).Invoke(responseObject, null);
+                            var obj = responseObject.GetType().GetProperty("Value", BindingFlags.Public | BindingFlags.Instance).GetValue(responseObject);
+                            //reconstruct
+                            var genericFromMethod = typeof(Response).GetMethod("FromValue", BindingFlags.Static | BindingFlags.Public).MakeGenericMethod(tGenericType);
+                            var newResponseObject = genericFromMethod.Invoke(null, new object[] { obj, rawResponse });
+                            return (T)newResponseObject;
```
JoshLove-msft, line 115: Doesn't this need to be wrapped in a Task?

JoshLove-msft: Actually I guess this isn't needed as we directly return the `T` which as `ValueTask<T>`

sdk/core/Azure.Core.TestFramework/src/Instrumentation/DiagnosticScopeValidatingInterceptor.cs
```diff
@@ -88,6 +88,33 @@ internal static void WrapAsyncResultCore<T>(IInvocation invocation, Type generic
...
+                            var returnInnerType = returnType.GetGenericArguments()[0];
+                            if (returnInnerType.IsGenericType && returnInnerType.GetGenericTypeDefinition().Equals(typeof(Response<>)))
+                            {
+                                returnGenericType = returnInnerType.GetGenericArguments()[0];
+                            }
+                        }
+                        if (returnGenericType is not null && returnGenericType.IsSubclassOf(tGenericType))
+                        {
+                            //get the object
+                            var responseObject = TaskExtensions.GetResultFromTask(invocation.ReturnValue);
```
JoshLove-msft, line 109: Why do we not await this?

JoshLove-msft (approved):

Issue #16251, opened 2020-10-25
[TABLES] GetEntity/Async should not throw exception on 404


kieronlanning:
`Azure.Data.Tables`

Is there an option to prevent `GetEntity`/ `GetEntityAsync` from throwing an exception in the case of an entity not existing.

jsquire:
Thank you for your feedback.  Tagging and routing to the team member best able to assist.

christothes:
@kieronlanning If such an option existed, could you describe how you'd see it working?

kieronlanning:
@christothes I think the best solution would be to add it to the `TableOptions` class in the constructor for the `TableServiceClient` (sorry if these aren't quite correct, I'm remembering the names from memory!).

That way the developer in question can choose the default on creation. But I do think you'd want to set it per-client, rather than per-call.

As for the `GetEntity` method itself, I'd like to see that return a `null` value with the response. So you can check the response for a 404 and return `null` accordingly. More choice. Right now I'm catching the exception and using the when clause to return null on a 404.

christothes:
@kieronlanning Thanks. Could you also describe what you'd typically do after this call? For example, is the intent to just check for existence of an entity or would you actually want something like CreateIfNotExists or DeleteIfExists?

kieronlanning:
@christothes For `GetEntity` it's always to check if it exists, but not always will we want to create or delete based on the result.

Sometimes it can form part of a `HEAD` request from the client, so it's a quick and easy way of returning some metadata to the client about the state of a resource.

Other times I might want to create it, but that's a rare case for me.

---------------------------------------------------------------------------------------------------

My context is this:

I use Table Storage as the primary store of an event store - where the events are stored in table storage as individual entities. This gives my clients transactions (multiple events, plus additional entities for management of the store) and idempotency, as well as etag based updating where appropriate. Couple that with blob storage for snapshotting and large events and Storage is a really solid platform.

That's why I'm so keen to get this Table API right. I'm close at the moment too - about 50 integration tests failing.

christothes:
@kieronlanning Generally, the guideline we follow in the newer "track 2" SDKs is to throw in scenarios like this. I realize this is a change in behavior from the older client, but it would be consistent with the behavior found in other track 2 clients.

Is there any blocking reason your code couldn't handle the exception in the scenario where `GetEntity` throws due to a 404?

kieronlanning:
[...]

PR #30888, opened 2022-09-01
Add GetEntityIfExists method to TableClient

christothes:
related to #16251
related to #25626


sdk/tables/Azure.Data.Tables/src/TableClient.cs
```diff
@@ -513,14 +515,17 @@ public virtual async Task<Response<TableItem>> CreateIfNotExistsAsync(Cancellati
...
+                    new TablesRequestContext(TablesRequestContext.Conflict) { CancellationToken = cancellationToken, ErrorOptions = ErrorOptions.NoThrow };
+
                 var response = await _tableOperations.CreateAsync(
                     RequestContent.Create(new TableProperties { TableName = Name }),
                     TableConstants.Odata.MinimalMetadata,
                     TableConstants.ReturnNoContent,
                     context).ConfigureAwait(false);
 
-                if (response.IsError)
+                if (response.IsError || response.Status == (int)HttpStatusCode.Conflict)
```
jsquire, line 528: Should we consider using the `public static` member from the request context here and avoid the cast?

sdk/tables/Azure.Data.Tables/src/TableClient.cs
```diff
@@ -676,12 +681,49 @@ public virtual Response<T> GetEntity<T>(string partitionKey, string rowKey, IEnu
...
+        /// <param name="partitionKey">The partitionKey that identifies the table entity.</param>
+        /// <param name="rowKey">The rowKey that identifies the table entity.</param>
+        /// <param name="select">Selects which set of entity properties to return in the result set. Pass <c>null</c> to retreive all properties.</param>
+        /// <param name="cancellationToken">A <see cref="CancellationToken"/> controlling the request lifetime.</param>
+        /// <returns>The <see cref="Response"/> indicating the result of the operation.</returns>
+        /// <exception cref="RequestFailedException">Exception thrown if the entity doesn't exist.</exception>
+        /// <exception cref="ArgumentNullException"><paramref name="partitionKey"/> or <paramref name="rowKey"/> is null.</exception>
+        // Supression is required to avoid an ambiguous call site
+#pragma warning disable AZC0002 // DO ensure all service methods, both asynchronous and synchronous, take an optional CancellationToken parameter called cancellationToken.
```
jsquire, line 698: This is a weird disable since you have the parameter in here...

jsquire (approved):

sdk/tables/Azure.Data.Tables/src/TableClient.cs
christothes: oh this is left over from earlier experiments - thanks for catching!

sdk/tables/Azure.Data.Tables/src/TableClient.cs
christothes: If you are referring to the static on TablesRequestContext, that one is an `int[] { (int)HttoStatusCode.Conflict }` - I suppose we could index into that array, but I don't think the cast is very expensive.

jsquire: ahh... yeah.  My bad, that's what I was thinking of but forgot it was an array.   No concerns on the cast, just thought we had an easily available alias for it already.

PR #30891, opened 2022-09-01
Add github actions based check enforcer

benbp:
Adding check enforcer v2 action to repository. See https://github.com/Azure/azure-sdk-tools/pull/3287 for details.

### DIFF
--- a/.github/workflows/event.yml
+++ b/.github/workflows/event.yml
@@ -1,0 +1,16 @@
+name: GitHub Event Handler
+
+on:
+  check_suite:
+    types: [completed]
+  issue_comment:
+    types: [created]
+
+jobs:
+  event-handler:
+    name: Handle ${{ github.event_name }} ${{ github.event.action }} event
+    runs-on: ubuntu-latest
+    steps:
+      - uses: azure/azure-sdk-actions@main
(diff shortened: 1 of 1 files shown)


jsquire (approved):

.github/workflows/event.yml
```diff
@@ -0,0 +1,16 @@
+name: GitHub Event Handler
```
weshaggard, line 1: If I didn't know the implementation details I would never realize this has anything to do with check-enforcer. We should consider adding some comment in these files that might indicate what events are being handled.

weshaggard (approved):

.github/workflows/event.yml
benbp: That's because the original intention was for this to be extended for any github actions that need to handle github events, so that we can avoid multiple runners getting spun up for each event.

benbp: I'd rather the file be event type agnostic, with `azure-sdk-actions` being the place to look to know what's handled. Otherwise if we add new event handlers we'd have to update all these files in the repos.

IRIEBRAIN (approved):

PR #30892, opened 2022-09-01
Introduce new batch delete api

anwayshahane:

### DIFF
--- a/sdk/servicebus/Microsoft.Azure.ServiceBus/src/Core/MessageReceiver.cs
+++ b/sdk/servicebus/Microsoft.Azure.ServiceBus/src/Core/MessageReceiver.cs
@@ -914,7 +914,10 @@ namespace Microsoft.Azure.ServiceBus.Core
             }
             finally
             {
-                this.diagnosticSource.BatchDeleteStop(activity, timeout, messageCount, deleteTask?.Status, messagesDeleted);
+                if (isDiagnosticSourceEnabled)
+                {
+                    this.diagnosticSource.BatchDeleteStop(activity, timeout, messageCount, deleteTask.Status, messagesDeleted);
+                }
             }
 
             MessagingEventSource.Log.MessageBatchDeleteStop(this.ClientId, timeout, messageCount);
--- a/sdk/servicebus/Microsoft.Azure.ServiceBus/src/Core/MessageReceiver.cs
+++ b/sdk/servicebus/Microsoft.Azure.ServiceBus/src/Core/MessageReceiver.cs
@@ -879,6 +879,81 @@ namespace Microsoft.Azure.ServiceBus.Core
             return messages;
         }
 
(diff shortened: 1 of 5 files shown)


sdk/servicebus/Microsoft.Azure.ServiceBus/src/Core/MessageReceiver.cs
```diff
@@ -879,6 +879,81 @@ public async Task<IList<Message>> PeekBySequenceNumberAsync(long fromSequenceNum
...
+                {
+                    this.diagnosticSource.ReportException(exception);
+                }
+
+                MessagingEventSource.Log.MessagePeekException(this.ClientId, exception);
+                throw;
+            }
+            finally
+            {
+                this.diagnosticSource.BatchDeleteStop(activity, timeout, messageCount, deleteTask?.Status, messagesDeleted);
```
lchew7, line 917: 1. Should we check the isDaignosticSourceEnabled flag here before?
2. Why do we log the start and end with different logging providers?
3. There is the possibility that 'activity' could be null - is that ok?

sdk/servicebus/Microsoft.Azure.ServiceBus/tests/QueueClientTests.cs
```diff
@@ -253,5 +253,27 @@ public async Task UpdatingPrefetchCountOnQueueClientUpdatesTheReceiverPrefetchCo
                 }
             });
         }
+
+        [Fact]
+        [DisplayTestMethodName]
```
lchew7, line 258: Other tests have the LiveTest attribute.  Do we need it here too?

sdk/servicebus/Microsoft.Azure.ServiceBus/src/ServiceBusDiagnosticsSource.cs
```diff
@@ -337,6 +337,38 @@ internal void CompleteStop(Activity activity, IList<string> lockTokens, TaskStat
...
+        {
+            if (activity != null)
+            {
+                DiagnosticListener.StopActivity(activity, new
+                {
+                    Timeout = timeout,
+                    RequestedMessageCount = messageCount,
+                    Entity = this.entityPath,
+                    Endpoint = this.endpoint,
+                    Status = status ?? TaskStatus.Faulted,
```
lchew7, line 364: TaskStatus is null if the deleteTask is null.  deleteTask is returned by RetryPolicy, so what does it mean when it returns null?  Does it mean the task failed or does it mean it couldn't even create the task?  If the latter, then I don't think "Faulted" is the right value.

Also, in my opinion, I don't think this logging function should be deciding how to interpret a null status.  The caller should always pass in a concrete value (i.e., it should take in a TaskStatus, not a TaskStatus?).

lchew7 (approved):

PR #30909, opened 2022-09-02
[FormRecognizer] Renames and architect feedback

### LINKED ISSUE #17712: How to present errors when de serializing response from service
When the SDK can't parse the response value to the specified type, the SDK throws without giving information to the user about what field was the culprit.
i.e. for a Receipt. If the service returned:
```
"TransactionDate": {
  "type": "date",
  "valueDate": "2019-00-10",
  "text": "0/10/2019",
  ...
}
```
The SDK will throw when [trying to parse](https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/formrecognizer/Azure.AI.FormRecognizer/src/Generated/Models/FieldValue_internal.Serialization.cs#L46) the value to a `DateTimeOffset`. And the user will not be able to get the information from the service to understand what happened, or what value was wrong.

Need to improve this experience when something fails while the SDK deserialize so the user can know what went wrong. Also, should they have access to the RawResponse so they can take a look too?

### LINKED ISSUE #24471: What to do with models that have same name but different namespace?
@maririos commented on [Fri Sep 17 2021](https://github.com/Azure/azure-sdk-for-net-pr/issues/1492)

Examples:
- AccountProperties
- LenghtUnit

Although each type lives in a specific namespace, the user could make the mistake of not paying attention to the namespace and end up using a type that doesn't correspond with their version.

Should we rename? is there a trick we could use here?

---

@maririos commented on [Tue Oct 05 2021](https://github.com/Azure/azure-sdk-for-net-pr/issues/1492#issuecomment-934769823)

The guidelines say:
> Namespaces are also indispensable in resolving type name ambiguities that might arise when importing multiple namespaces. However, that facto should not be used as an excuse to introduce known ambiguities between types in different namespaces that are commonly used together. Developers should not be required to qualify type names in common scenarios

So maybe we shouldn't as a user can be migrating from v2 to v3 and have both namespaces and end up using the wrong type

kinelski:
These are the final changes in the public API before our next release. This PR includes:
- For cross-language consistency:
  - Renamed methods in `DocumentModelAnalysisClient` to use the term `DocumentModel` instead of `Model` only. For example, `BuildModel` became `BuildDocumentModel`. This also implies changes in our options bag (`BuildModelOptions` => `BuildDocumentModelOptions`) and in our LRO pollers (`BuildModelOperation` => `BuildDocumentModelOperation`).
- From customer feedback:
  - Added helper method `DocumentLine.GetWords` to help users to filter out the list of `DocumentWord` instances that compose a `DocumentLine`. This information is not directly returned by the service.
- From architects feedback:
  - Updated the way `DocumentField` is handled. When the field value is not returned by the service (which happens when they can't parse the extracted value into the expected type), we set `DocumentField.FieldType` to `DocumentFieldType.Unknown` in the SDK. Customers could then use the `DocumentField.ExpectedFieldType` in addition to the raw string `DocumentField.Content` to manually parse the field value in this case. This change permits handling this scenario without try/catch blocks (current approach).
  - `BoundingRegion` and `DocumentSpan` now implement the `IEquatable<T>` interface and had their `ToString` implementations overriden to help debugging.
- Based on our guidelines:
  - To avoid models with the same name across the old and new Form Recognizer namespaces, `CopyAuthorization`, `LengthUnit`, and `SelectionMarkState` have been renamed to `DocumentModelCopyAuthorization`, `DocumentPageLengthUnit`, and `DocumentSelectionMarkState`, respectivelly.

Also, added new tests for the new features.

Fixes #24471.
Fixes #17712.


sdk/formrecognizer/Azure.AI.FormRecognizer/src/DocumentLine.cs
```diff
@@ -39,5 +44,35 @@ private IReadOnlyList<float> Polygon
                 BoundingPolygon = ClientCommon.CovertToListOfPointF(value);
             }
         }
+
+        /// <summary>
+        /// Returns the list of <see cref="DocumentWord"/> that compose this line.
+        /// </summary>
+        /// <returns>The list of <see cref="DocumentWord"/> that compose this line.</returns>
+        public IReadOnlyList<DocumentWord> GetWords()
```
kinelski, line 52: Spans are sorted, so we could consider making a binary search to speed up the process. Lines tend to be pretty short, though, so I'm not sure if it's worth the trouble.

sdk/formrecognizer/Azure.AI.FormRecognizer/api/Azure.AI.FormRecognizer.netstandard2.0.cs
```diff
@@ -301,17 +301,23 @@ public partial class AnalyzeResult
         public System.Collections.Generic.IReadOnlyList<Azure.AI.FormRecognizer.DocumentAnalysis.DocumentTable> Tables { get { throw null; } }
     }
     [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
-    public readonly partial struct BoundingRegion
+    public readonly partial struct BoundingRegion : System.IEquatable<Azure.AI.FormRecognizer.DocumentAnalysis.BoundingRegion>
```
pallavit, line 304: Curious why we did this? Is this a customer ask where they want to check equality of the region or is this coming from our design guidelines?

sdk/formrecognizer/Azure.AI.FormRecognizer/api/Azure.AI.FormRecognizer.netstandard2.0.cs
```diff
@@ -501,19 +499,20 @@ public partial class DocumentFieldSchema
...
-        Double = 4,
-        Int64 = 5,
-        List = 6,
-        Dictionary = 7,
-        SelectionMark = 8,
-        CountryRegion = 9,
-        Signature = 10,
-        Currency = 11,
-        Address = 12,
+        Unknown = 0,
```
pallavit, line 502: Isn't this a breaking change? Should we not mention it explicitly in changelog?

sdk/formrecognizer/Azure.AI.FormRecognizer/src/BoundingRegion.cs
```diff
@@ -37,8 +40,84 @@ internal BoundingRegion(int pageNumber, IReadOnlyList<PointF> boundingPolygon)
...
+            }
+
+            if (BoundingPolygon.Count != other.BoundingPolygon.Count)
+            {
+                return false;
+            }
+
+            // Since points in a polygon are cyclical, two polygons could still be considered the same
+            // if points were shifted. For example, [(1,2),(3,4),(5,6),(7,8)] and [(7,8),(1,2),(3,4),(5,6)]
+            // represent the same polygon. This, however, is not expected to happen since points returned
```
pallavit, line 83: Does service guarantee this behavior?
Should we have a test to validate that the service is not breaking this since we are relying on this behavior.

sdk/formrecognizer/Azure.AI.FormRecognizer/src/DocumentField.cs
```diff
@@ -77,10 +77,19 @@ internal DocumentField(DocumentFieldType fieldType, DocumentFieldValue value, st
...
 
         /// <summary>
-        /// The data type of the field value.
+        /// The data type of the field value. If <see cref="DocumentFieldType.Unknown"/>,
+        /// it means the value of the field could not be parsed by the service. The expected
+        /// field type can be checked at <see cref="ExpectedFieldType"/>. Consider using
+        /// <see cref="Content"/> to get a textual representation of the field and parsing it
+        /// manually in this case.
+        /// </summary>
+        public DocumentFieldType FieldType => Value.FieldType;
```
pallavit, line 86: Should this be ActualFieldType perhaps? One looks at ExpectedFieldType and FieldType and it is confusing what FieldType represents if ExpectedFieldType is also available in the public API surface area? Just a thought - if this has been arch board reviewed already - I wouldn't block on it.

sdk/formrecognizer/Azure.AI.FormRecognizer/src/DocumentLine.cs
pallavit: Should we add a perf test for this scenario especially memory perf testing? You do not have to block GA release for this - but having a follow-up bug in the backlog so we continue to read the trend.

pallavit: Also since spans are sorted and we rely on that service behavior - add a test for it.

kinelski: Created an issue, as suggested: https://github.com/Azure/azure-sdk-for-net/issues/30914

The current is not the smartest, so it doesn't depend on the fact that `Words` is sorted. I do have a test, though.

sdk/formrecognizer/Azure.AI.FormRecognizer/api/Azure.AI.FormRecognizer.netstandard2.0.cs
kinelski: It's mentioned in the CHANGELOG:
> `DocumentField.ValueType` (now called `FieldType`) can now be `Unknown` when the field value couldn't be parsed by the service.

sdk/formrecognizer/Azure.AI.FormRecognizer/src/BoundingRegion.cs
kinelski: This service guarantees this behavior, according to their [documentation](https://github.com/Azure/azure-rest-api-specs/blob/2fd7dcd89afa70ff5ba7be88bee987da62099a28/specification/cognitiveservices/data-plane/FormRecognizer/stable/2022-08-31/FormRecognizer.json#L2063).
> Bounding polygon, with coordinates specified relative to the top-left of the page. The numbers represent the x, y values of the polygon vertices, clockwise from the left (-180 degrees inclusive) relative to the element orientation.

I think it would be easier to just update this method to not rely on this behavior anymore. Just need to check if points are shifted. I'll also add a test to ensure we're doing this correctly.

sdk/formrecognizer/Azure.AI.FormRecognizer/src/DocumentField.cs
kinelski: We added the `ExpectedFieldType` property to handle a very particular scenario, but most users will never need to use it. One of the things we had in mind when coming up with this design was keeping the usage simple for the average user. I think that having `Actual`/`Expected` could make it harder for users to know which one to use (on the other hand, it would force them to think better about it).

That said, these names were suggested by architects, so I also wanted to avoid additional changes.

sdk/formrecognizer/Azure.AI.FormRecognizer/api/Azure.AI.FormRecognizer.netstandard2.0.cs
kinelski: It's not explicitly written in our guidelines, but we usually override `IEquatable<T>` for small structs (this one in particular was requested by Krzysztof). I don't think customers will directly check for equality here, but implementing this interface is relevant if these are added to collections such as `Dictionary` or `Set`, which shouldn't accept two copies of the same key.

sdk/formrecognizer/Azure.AI.FormRecognizer/src/DocumentLine.cs
kinelski: Added the check to our `ValidatePage` helper test method (checked that spans don't overlap and are sorted). It's called by pretty much every document analysis test, so we now have a lot of data showing that the service is following the expected behavior.

sdk/formrecognizer/Azure.AI.FormRecognizer/src/BoundingRegion.cs
kinelski: Now that I thought more about it, it's not as simple since we would need to implement `GetHashCode` in a way that ensures shifted polygons generate the same hash code. I'll just add validation for the service then.

kinelski: Done. Had to update `GetHashCode` too, since shifted points should still give the same hash code. We're basically ordering the points before generating the code.

pallavit (approved):

PR #30948, opened 2022-09-06
[Azure.AI.TextAnalytics] Authoring Client

LouaiZahran:
Initial Version of TA authoring client, developed using 2022-05-01 GA specs.


sdk/textanalytics/Azure.AI.TextAnalytics/src/Authoring/TextAuthoringClientOptions.cs
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
heaths, line 1: Generally we use the same *ClientOptions class as the inference client. See https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/cognitivelanguage/Azure.AI.Language.Conversations/src/autorest.md and client and clientoptions files under https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/cognitivelanguage/Azure.AI.Language.Conversations/src (except in Generated) for how I pulled this off.

heaths (changes requested):

PR #30963, opened 2022-09-06
Pass client timeout to server when accepting next session

### LINKED ISSUE #30962: [Service Bus] Respect TryTimeout when opening next available session link
The service hardcodes a max of 1 minute and 5 seconds when accepting a session. This means the TryTimeout that the user specifies in the client is ignored if they specify a value higher than this. We should specify the timeout property in the link settings properties when opening a link for next available session to support long polling. In the .NET library, we've found that using long polling results in better resource usage and avoids CPU spikes.

JoshLove-msft:
Fixes https://github.com/Azure/azure-sdk-for-net/issues/30962

### DIFF
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpConnectionScope.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpConnectionScope.cs
@@ -72,6 +72,13 @@ namespace Azure.Messaging.ServiceBus.Amqp
         /// </summary>
         private static int AuthorizationBaseJitterSeconds { get; } = 30;
 
+        ///// <summary>
+        /////   The number of milliseconds to use as the basis for calculating a random jitter amount
+        /////   when opening receiver links. This is intended to ensure that multiple
+        /////   accept session open operations don't timeout at the same exact moment.
+        /////// </summary>
+        private static int OpenReceiveLinkBaseJitterMilliseconds { get; } = 100;
+
         /// <summary>
         ///   The minimum amount of time for authorization to be refreshed; any calculations that
         ///   call for refreshing more frequently will be substituted with this value.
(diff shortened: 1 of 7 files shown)


sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpConnectionScope.cs
```diff
@@ -672,6 +679,29 @@ public override void Dispose()
                     OperationTimeout = _operationTimeout
                 };
 
+                if (isSessionReceiver && sessionId == null)
+                {
+                    // Subtract a random amount up to 100ms from the operation timeout as the jitter when attempting to open next available session link.
```
jsquire, line 684: The commenting in this PR is exceptionally helpful!

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpConnectionScope.cs
```diff
@@ -672,6 +679,29 @@ public override void Dispose()
...
+                    var jitterBase = Math.Min(_operationTimeout.TotalMilliseconds / 100, OpenReceiveLinkBaseJitterMilliseconds);
+
+                    // We set the operation timeout on the properties not only to include the jitter, but also because the server will otherwise
+                    // restrict the maximum timeout to 1 minute and 5 seconds, regardless of the client timeout. We only do this for accepting next available
+                    // session as this is the only long-polling scenario.
+                    linkSettings.Properties = new Fields
+                    {
+                        {
+                            AmqpClientConstants.TimeoutName,
+                            (uint)_operationTimeout
```
jsquire, line 698: So the service timeout is always shorter than the client?

jsquire (approved):

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpConnectionScope.cs
JoshLove-msft: Yes, the service already subtracts 20ms off of their hardcoded 1 minute and 5 second timeout. This adds an additional buffer.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpConnectionScope.cs
```diff
@@ -72,6 +72,13 @@ internal class AmqpConnectionScope : TransportConnectionScope
         /// </summary>
         private static int AuthorizationBaseJitterSeconds { get; } = 30;
 
+        ///// <summary>
```
conniey, line 75: I think there may have accidentally been more forward-slashes than you wanted

JoshLove-msft: Ugh.. This is from when I was testing something and had commented out the new property.

JoshLove-msft: https://github.com/Azure/azure-sdk-for-net/pull/30988

PR #30965, opened 2022-09-06
Fix Search Live Tests

khmic5:

### DIFF
--- a/sdk/maps/Azure.Maps.Search/tests/SearchClientTestEnvironment.cs
+++ b/sdk/maps/Azure.Maps.Search/tests/SearchClientTestEnvironment.cs
@@ -15,6 +15,6 @@ namespace Azure.Maps.Search.Tests
                 return endpoint != null ? new Uri(endpoint) : null;
             }
         }
-        public string MapAccountClientId => GetRecordedVariable("MAPS_CLIENT_ID");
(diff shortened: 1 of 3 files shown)


jsquire (approved):

sdk/maps/Azure.Maps.Search/tests/SearchClientTestEnvironment.cs
```diff
@@ -15,6 +15,6 @@ public class SearchClientTestEnvironment : TestEnvironment
                 return endpoint != null ? new Uri(endpoint) : null;
             }
         }
-        public string MapAccountClientId => GetRecordedVariable("AZMAPS_CLIENT_ID");
+        public string MapAccountClientId => GetRecordedVariable("MAPS_CLIENT_ID");
```
weshaggard, line 18: Should this match the name of the output comming from the test-resources.json file? https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/maps/test-resources.json#L44 Similar question for the ENDPOINT_URL above.

benbp: The new value is correct, as it follows the `<service directory>_CLIENT_ID` convention for env vars that gets set by the deploy script, not the ARM template.

benbp (approved):

PR #30969, opened 2022-09-06
Track 2 Migration for Guest Configuration - Final Step

anineela:

Our latest spec is already merged to main: https://github.com/Azure/azure-rest-api-specs/tree/main/specification/guestconfiguration/resource-manager/Microsoft.GuestConfiguration/stable/2022-01-25

 Enabled and tested "GetAllAsync" list tests that were previously disabled due to a bug fix being worked on by Yao. 

image


Yao725 (approved):
LGTM

PR #30990, opened 2022-09-07
Support explicitly implemented interfaces in TypeBinder

### LINKED ISSUE #26514: Enable implementors of TypeBinder<TExchange> to serialize properties from explicitly implemented interfaces

christothes:
Adds support for serializing explicitly implemented interface properties and leverages this support in Tables for `ITableEntity`.

fixes https://github.com/Azure/azure-sdk-for-net/issues/26514


sdk/tables/Azure.Data.Tables/src/Extensions/DictionaryTableExtensions.cs
```diff
@@ -148,7 +148,7 @@ internal static T ToTableEntity<T>(this IDictionary<string, object> entity, Tabl
                 return result;
             }
 
-            typeInfo ??= TablesTypeBinder.Shared.GetBinderInfo(typeof(T));
+            typeInfo ??= TablesTypeBinder.Shared.GetBinderInfo(typeof(T), typeof(ITableEntity));
```
JoshLove-msft, line 151: Do we need to add this support to WebJobs extension?
https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/tables/Microsoft.Azure.WebJobs.Extensions.Tables/src/PocoToTableEntityConverter.cs#L20

JoshLove-msft (approved):

sdk/tables/Azure.Data.Tables/src/Extensions/DictionaryTableExtensions.cs
christothes: Good callout!

sdk/tables/Microsoft.Azure.WebJobs.Extensions.Tables/src/PocoToTableEntityConverter.cs
```diff
@@ -17,7 +17,7 @@ internal class PocoToTableEntityConverter<TInput>: IConverter<TInput, TableEntit
 
         public PocoToTableEntityConverter()
         {
-            _info = PocoTypeBinder.Shared.GetBinderInfo(typeof(TInput));
+            _info = PocoTypeBinder.Shared.GetBinderInfo(typeof(TInput), typeof(ITableEntity));
```
JoshLove-msft, line 20: Don't we need to support both interface and non-interface? Is this a breaking change?

JoshLove-msft: Or I suppose we are doing that here? If so, can we remove the other GetBinderInfo method and ctor?

christothes: What's the scenario you have in mind for non-interface?

JoshLove-msft: In the extension bindings I don't think we require that POCOs implement ITableEntity. We should have tests for this.

JoshLove-msft: It looks like we have a compile-time constraint for the Tables use case, so it is not a breaking change there.

christothes: It's definitely a behavior change, but I'm not sure if I would call it a breaking change since serialization just wouldn't work properly before when explicitly implemented ITableEntity properties. Do you think it would make more sense to throw behind an option?

JoshLove-msft: The part that is breaking is if the type you are binding to did not implement ITableEntity at all.

christothes: Looking at this again, after seeing the failed test, I don't think this converter needs the explicit interface support, since it is specifically to handle a scenario where the POCO does not implement `ITableEntity`

christothes: > The part that is breaking is if the type you are binding to did not implement ITableEntity at all.

That is currently not supported, at least in Tables proper. I'll do some more investigation for the extension scenarios.

JoshLove-msft (approved):

jsquire (approved):

sdk/core/Azure.Core/src/Shared/TypeBinder.cs
```diff
@@ -154,6 +174,23 @@ public BoundMemberInfo(MemberInfo memberInfo)
...
             public abstract void Serialize(object o, TExchange destination, TypeBinder<TExchange> binderImplementation);
             public abstract void Deserialize(TExchange source, object o, TypeBinder<TExchange> binderImplementation);
+
+            public bool Equals(BoundMemberInfo x, BoundMemberInfo y)
+            {
+                if (ReferenceEquals(x, y))
+                    return true;
+                if (x is null || y is null)
+                    return false;
+                return x.Name == y.Name;
```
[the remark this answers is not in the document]
christothes: I don't believe so - if they are cased differently, they are different properties

sdk/tables/Microsoft.Azure.WebJobs.Extensions.Tables/tests/TableCreationTests.cs
```diff
@@ -110,6 +112,7 @@ public static void Call([Table(TableNameExpression)] out T entity)
         [GenericTestCase(typeof(JObject))]
         [GenericTestCase(typeof(StructTableEntity))]
         [GenericTestCase(typeof(PocoTableEntity))]
+        [GenericTestCase(typeof(ExplicitlyImplementedITableEntity))]
```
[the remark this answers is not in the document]
christothes: I don't think that is a valid test since this operates on an explicit type rather than a type inferred from an instance. There is a test that does feed an explicit `ITableEntity` to the binding elsewhere.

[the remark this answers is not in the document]
JoshLove-msft: The test I'm thinking of binds to an ITableEntity, but the underlying instance is not TableEntity.

sdk/tables/Microsoft.Azure.WebJobs.Extensions.Tables/tests/TablesLiveTestBase.cs
```diff
@@ -37,7 +37,7 @@ public class TablesLiveTestBase : RecordedTestBase<TablesTestEnvironment>
 
         protected bool UseCosmos { get; }
 
-        protected TablesLiveTestBase(bool isAsync, bool useCosmos, bool createTable = true): base(isAsync: isAsync)
+        protected TablesLiveTestBase(bool isAsync, bool useCosmos, bool createTable = true) : base(isAsync: isAsync, RecordedTestMode.Playback)
```
JoshLove-msft, line 40:
```suggestion
        protected TablesLiveTestBase(bool isAsync, bool useCosmos, bool createTable = true) : base(isAsync: isAsync)
```

PR #31015, opened 2022-09-08
update common and fix race test condition

khmic5:

### DIFF
--- a/sdk/maps/Azure.Maps.Search/tests/FuzzySearchTests.cs
+++ b/sdk/maps/Azure.Maps.Search/tests/FuzzySearchTests.cs
@@ -72,6 +72,8 @@ namespace Azure.Maps.Search.Tests
                 new FuzzySearchQuery("pizza", new FuzzySearchOptions { Coordinates = new GeoPosition(121.56, 25.04) }),
             });
 
+            // delay 400 ms for the task to complete
+            await Task.Delay(400);
             var fuzzySearchBatchResp = operation.WaitForCompletion();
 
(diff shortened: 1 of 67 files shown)


sdk/maps/Azure.Maps.Search/tests/FuzzySearchTests.cs
```diff
@@ -72,6 +72,8 @@ public async Task CanPollFuzzySearchBatch()
                 new FuzzySearchQuery("pizza", new FuzzySearchOptions { Coordinates = new GeoPosition(121.56, 25.04) }),
             });
 
+            // delay 400 ms for the task to complete
+            await Task.Delay(400);
```
weshaggard, line 76: Is this delay required? What do we expect customers to do here? It seems like our wait should actually wait for completion as the name implies.

khmic5: This delay in the test case prevents the race condition error from happening so there is enough time to retrieve the batch result

khmic5: https://github.com/Azure/azure-sdk-for-net/issues/31019

khmic5: I have opened an issue to track what appears to be a server side bug

jsquire: I had the same concern.  We chatted offline; it appears this is a service bug where the operation request returns a 202 but the subsequent status poll returns a 404.   The Maps team is going to investigate.   For now, we'll keep the delay using #31019 to track the debt of revisiting once the service bug is fixed.

jsquire (approved):

PR #31021, opened 2022-09-08
update code checks to only run mgmt yml generator if there is a mgmt …

m-nash:
…library in current service directory


weshaggard:
We should also consider updating https://github.com/Azure/azure-sdk-for-net/blob/main/eng/checkenforcer.md with a note about this script and when to run it.

eng/scripts/CodeChecks.ps1
```diff
@@ -89,6 +89,15 @@ try {
         Invoke-Block {
             & dotnet msbuild $PSScriptRoot/../service.proj /restore /t:GenerateTests /p:SDKType=$SDKType /p:ServiceDirectory=$ServiceDirectory
         }
+
+        $mgmtPackagesInServiceDirectory = Get-ChildItem -Path "$PSScriptRoot/../../sdk/$ServiceDirectory" -Directory | Where-Object { $_.Name -match "(Azure.ResourceManager)" -and $(Test-Path("$($_.FullName)/src")) }
```
weshaggard, line 93: I don't think this check is going to really help as it is still going to run for pretty much every pipeline because I believe all the data plane directories also have a mgmt library in them. Perhaps for now we just blindly run this check and then we can see if it causes any headaches.

weshaggard: I will also add that the comments below with the git diff that will cause this to actually fail don't mention anything about this check we should probably mention how to fix this there as well.

m-nash: there are quite a few that don't have mgmt plane that will now be excluded.  I do think its appropriate that if we are going to regen mgmt plane code we should regen all of it to make sure no auto-generated portions have been manually touched

weshaggard:
Closing in favor of https://github.com/Azure/azure-sdk-for-net/pull/31361

PR #31026, opened 2022-09-09
Verify challenge resource matches request domain

heaths:


sdk/keyvault/Azure.Security.KeyVault.Administration/src/KeyVaultAccessControlClient.cs
```diff
@@ -36,7 +36,7 @@ protected KeyVaultAccessControlClient()
         /// <summary>
         /// Initializes a new instance of the <see cref="KeyVaultAccessControlClient"/> class for the specified vault.
         /// </summary>
-        /// <param name="vaultUri">A <see cref="Uri"/> to the vault on which the client operates. Appears as "DNS Name" in the Azure portal.</param>
+        /// <param name="vaultUri">A <see cref="Uri"/> to the vault on which the client operates. Appears as "DNS Name" in the Azure portal. You should validate that this URI references a valid Key Vault or Managed HSM resource. See https://aka.ms/azsdk/blog/vault-uri for details.</param>
```
pallavit, line 39: >https://aka.ms/azsdk/blog/vault-uri [](http://example.com/codeflow?start=233&length=35)

This link does not exist.

heaths: It will, which is why I allocated the link early. It won't be live till next Tuesday.

sdk/keyvault/Azure.Security.KeyVault.Administration/tests/ChallengeBasedAuthenticationPolicyTests.cs
```diff
@@ -10,39 +12,85 @@
...
             var keyvaultChallengeResponse = new MockResponse(401);
             keyvaultChallengeResponse.AddHeader(new HttpHeader("WWW-Authenticate", KeyVaultChallenge));
             MockTransport transport = CreateMockTransport(keyvaultChallengeResponse, new MockResponse(200));
 
-            var response = await SendGetRequest(transport, _policy, uri: new Uri("https://example.com"));
+            Response response = await SendGetRequest(transport, _policy, uri: new Uri("https://myvault.vault.azure.net"));
 
             Assert.That(response.Status, Is.EqualTo(200));
 
             // Construct a new policy so that we can get the Scopes from cache.
```
pallavit, line 43: I am a little confused by this test case. How are we validating here that the scope is getting picked from the  cache in the second call to sendgetRequest. Challengebasedauthpolicy may have it cached the first time but the second sendGetRequest will re add it if it were not present. Did you mean to force the check even if scope was not present (via cache) and accidently used the same request the second time?

heaths: The cache is static, and because I only added one challenge mock response, if it weren't cache the client (via the new policy) would've expected a WWW-Authenticate header and failed since one wasn't queued by the `MockTransport` (responses are popped as they are used unless you use the delegate form constructor).

pallavit: I may be mistaken here but line 46 has the following  transport = CreateMockTransport(keyvaultChallengeResponse, new MockResponse(200)); --> which is 2 responses. Did you intent to only add one response to the transport instead of 2?

heaths: Sorry, you're right! Must've gotten a little overzealous on copy/paste. Fix incoming.

Issue #30692, opened 2022-08-23
[Service Bus] Create Troubleshooting Guide


JoshLove-msft:
Here is the [template]( https://gist.github.com/JoshLove-msft/4b3ffbc041067602612a2d81fe514a97) that each language can follow to ensure all topics are covered in a consistent way.

As part of this work, we should update relevant exceptions to link out to aka.ms links that point to the corresponding section of the Troubleshooting guide, similar to what we did for [Identity](https://azure.grep.app/search?q=aka.ms&filter%5Brepo%5D%5B0%5D=Azure/azure-sdk-for-net&filter%5Blang%5D%5B0%5D=C%23&filter%5Bpath%5D%5B0%5D=sdk/identity/Azure.Identity/).

JoshLove-msft:
https://github.com/Azure/azure-sdk-for-java/issues/30617

https://github.com/Azure/azure-sdk-for-js/issues/22987

https://github.com/Azure/azure-sdk-for-python/issues/25826

https://github.com/Azure/azure-sdk-for-go/issues/18937

JoshLove-msft:
As part of this work, we should update relevant exceptions to link out to aka.ms links that point to the corresponding section of the Troubleshooting guide.

https://microsoft.sharepoint.com/:o:/r/teams/AzureDeveloperExperience/Shared%20Documents/Service%20-%20Azure.Identity/Azure%20Identity?d=w238779ca6b5f4c078796a52ef0296997&csf=1&web=1

JoshLove-msft:
Outline - https://gist.github.com/JoshLove-msft/4b3ffbc041067602612a2d81fe514a97

Issue #27637, opened 2022-03-18
[Service Bus] Retry Guidance for Transactions Sample


jsquire:
Due to the way that transactions work with Service Bus, retries for service operations cannot be performed when there is an active ambient transaction. We should add a section to the [transactions sample](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/servicebus/Azure.Messaging.ServiceBus/samples/Sample06_Transactions.md) that highlights this.

jsquire:
Prerequisite:  #27931

JoshLove-msft:
Closing as this has been added to the troubleshooting guide - https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/servicebus/Azure.Messaging.ServiceBus/TROUBLESHOOTING.md#operations-in-a-transaction-are-not-retried

PR #31030, opened 2022-09-09
Create SB troubleshooting guide

JoshLove-msft:
Contributes to https://github.com/Azure/azure-sdk-for-net/issues/30692

Still need to update exception messages to link to the relevant troubleshooting guide sections once this is merged.


sdk/servicebus/Azure.Messaging.ServiceBus/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,305 @@
...
+
+This is often caused by thread starvation, particularly when using the session processor and using a very high value for [MaxConcurrentSessions][MaxConcurrentSessions] (e.g. > 50). The first thing to check would be to make sure you are not doing sync-over-async in any of your event handlers. Sync-over-async is an easy way to cause deadlocks and thread starvation. Even if you are not doing sync over async, any pure sync code in your handlers could contribute to thread starvation. If you've determined that this is not the issue, e.g. because you have pure async code, you can try increasing your [TryTimeout][TryTimeout]. This will relieve pressure on the threadpool by reducing the number of context switches and timeouts that may occur when using the session processor in particular. The default value for [TryTimeout][TryTimeout] is 60 seconds, but it can be set all the way up to 1 hour. Try setting it to something like 5 minutes as a starting point, and iterate from there. If none of this works, you may simply need to scale out to multiple instances. Reduce the concurrency in your application, but run the application on multiple instances to achieve the desired overall concurrency.
```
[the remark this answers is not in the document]
jsquire: I'd also consider mentioning that even with pure asynchronous code, there's no fairness in the scheduling algorithm.  If there's a high number of continuations waiting, some work/sessions/entitites/whatever may appear to stall while others do not.

jsquire, line 226:
```suggestion
This is often caused by thread starvation, particularly when using the session processor and using a very high value for [MaxConcurrentSessions][MaxConcurrentSessions] (e.g. > 50). The first thing to check would be to make sure you are not doing sync-over-async in any of your event handlers. Sync-over-async is an easy way to cause deadlocks and thread starvation. Even if you are not doing sync over async, any pure sync code in your handlers could contribute to thread starvation. If you've determined that this is not the issue, e.g. because you have pure async code, you can try increasing your [TryTimeout][TryTimeout]. This will relieve pressure on the threadpool by reducing the number of context switches and timeouts that may occur when using the session processor in particular. The default value for [TryTimeout][TryTimeout] is 60 seconds, but it can be set all the way up to 1 hour.  We recommend testing with the `TryTimeout` set to 5 minutes as a starting point and iterate from there. If none of these suggestiosn work, you may simply need to scale out to multiple hosts, reducing the concurrency in your application, but running the application on multiple hosts to achieve the desired overall concurrency.
```

sdk/servicebus/Azure.Messaging.ServiceBus/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,305 @@
...
+
+This can be configured using the [SessionIdleTimeout][SessionIdleTimeout], which tells the processor how long to wait to receive a message from a session, before giving up and moving to another one. This is useful if you have many sparsely populated sessions, where each session may only have a few messages. If you expect that each session will have many messages that trickle in, setting this too low can be counter productive, as it will result in unnecessary closing of the session.
+
+### Processor stops immediately
+
+This typically comes up for demo or testing scenarios. Calling StartProcessingAsync returns immediately after the processor has started. Calling this method will not keep your app alive for as long as the processor is running, so you'll need some other mechanism to keep the app alive until you wish to shut it down. For demos or testing, it may be sufficient to just add a Console.ReadKey() call after you start the processor. For production scenarios, you will likely want to use some sort of framework integration like [BackgroundService][BackgroundService] to provide convenient application lifecycle hooks that can be used for starting and disposing the processor.
```
jsquire, line 240:
```suggestion
This often is observed for demo or testing scenarios.  `StartProcessingAsync` returns immediately after the processor has started. Calling this method will not block and keep your application alive while the processor is running, so you'll need some other mechanism to do so.  For demos or testing, it may be sufficient to just add a `Console.ReadKey()` call after you start the processor. For production scenarios, you will likely want to use some sort of framework integration like [BackgroundService][BackgroundService] to provide convenient application lifecycle hooks that can be used for starting and disposing the processor.
```

sdk/servicebus/Azure.Messaging.ServiceBus/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,305 @@
...
+
+For general information about transactions in Service Bus, see the [Overview of Service Bus transaction processing][Transactions].
+
+### Supported operations
+
+Not all operations are supported when using transactions. To see the list of supported transactions, see [Operations within a transaction scope][TransactionOperations].
+
+### Timeout
+
+A transaction times out after 2 minutes. The transaction timer starts when the first operation in the transaction starts.
```
jsquire, line 252: Is there service documentation we can point to for this rather than keeping the value here?

sdk/servicebus/Azure.Messaging.ServiceBus/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,305 @@
...
+
+### Operations in a transaction are not retried
+
+This is by design. Consider the following scenario - you are attempting to complete a message within a transaction, but there is some transient error that occurs, e.g. ServiceBusException with Reason of ServiceCommunicationProblem. Suppose the request does actually make it to the service. If the client were to retry, the service would see two complete requests. The first complete will not be finalized until the transaction is committed. The second complete is not able to even be evaluated before the first complete finishes. The transaction on the client is waiting for the complete to finish. This creates a deadlock where the service is waiting for the client to complete the transaction, but the client is waiting for the service to acknowledge the second complete operation. The transaction will eventually timeout after 2 minutes, but this is a bad user experience. For this reason, we do not retry operations within a transaction.
```
jsquire, line 256:
```suggestion
This is by design. Consider the following scenario - you are attempting to complete a message within a transaction, but there is some transient error that occurs, e.g. `ServiceBusException` with a `Reason` of `ServiceCommunicationProblem`. Suppose the request does actually make it to the service. If the client were to retry, the service would see two complete requests. The first complete will not be finalized until the transaction is committed. The second complete is not able to even be evaluated before the first complete finishes. The transaction on the client is waiting for the complete to finish. This creates a deadlock where the service is waiting for the client to complete the transaction, but the client is waiting for the service to acknowledge the second complete operation. The transaction will eventually timeout after 2 minutes, but this is a bad user experience. For this reason, we do not retry operations within a transaction.
```

jsquire: Seems like we now have a definitive answer and should move forward with #27637.

sdk/servicebus/Azure.Messaging.ServiceBus/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,305 @@
...
+
+### Transactions across entities are not working
+
+In order to perform transactions that involve multiple entities, you'll need to set the ServiceBusClientOptions.EnableCrossEntityTransactions property to `true`. For details, see the [Transactions across entities][CrossEntityTransactions] sample. 
+
+## Quotas
```
jsquire, line 262: Interesting choice to include this as a stand-alone section since it is referenced inline for the exception scenarios.   What's the thinking?
_(context: Not objecting to it, but I am trying to decide whether to steal the idea for Event Hubs...)_

sdk/servicebus/Azure.Messaging.ServiceBus/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,305 @@
...
+[ApplicationProperties]: https://docs.microsoft.com/dotnet/api/azure.messaging.servicebus.servicebusmessage.applicationproperties?view=azure-dotnet
+[AppInsights]: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/samples/Diagnostics.md#applicationinsights-with-azure-monitor
+[TransactionOperations]: https://docs.microsoft.com/azure/service-bus-messaging/service-bus-transactions#operations-within-a-transaction-scope
+[Transactions]: https://docs.microsoft.com/azure/service-bus-messaging/service-bus-transactions
+[CrossEntityTransactions]: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/servicebus/Azure.Messaging.ServiceBus/samples/Sample06_Transactions.md#transactions-across-entities
+[LargeMessageSupport]: https://docs.microsoft.com/azure/service-bus-messaging/service-bus-premium-messaging#large-messages-support
+[GitHubDiscussionOnBatching]: https://github.com/Azure/azure-sdk-for-net/issues/25381#issuecomment-1227917960
+[BackgroundService]: https://docs.microsoft.com/dotnet/api/microsoft.extensions.hosting.backgroundservice?view=dotnet-plat-ext-6.0
```
jsquire, line 305:
```suggestion
[BackgroundService]: https://docs.microsoft.com/dotnet/api/microsoft.extensions.hosting.backgroundservice?view=dotnet
```

sdk/servicebus/Azure.Messaging.ServiceBus/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,305 @@
...
+[ReceiveMessages]: https://docs.microsoft.com/dotnet/api/azure.messaging.servicebus.servicebusreceiver.receivemessagesasync?view=azure-dotnet#azure-messaging-servicebus-servicebusreceiver-receivemessagesasync(system-int32-system-nullable((system-timespan))-system-threading-cancellationtoken)
+[ReceiveDeferredMessages]: https://docs.microsoft.com/dotnet/api/azure.messaging.servicebus.servicebusreceiver.receivedeferredmessagesasync?view=azure-dotnet
+[MessageState]: https://docs.microsoft.com/dotnet/api/azure.messaging.servicebus.servicebusreceivedmessage.state?view=azure-dotnet#azure-messaging-servicebus-servicebusreceivedmessage-state
+[SequenceNumber]: https://docs.microsoft.com/dotnet/api/azure.messaging.servicebus.servicebusreceivedmessage.sequencenumber?view=azure-dotnet
+[Logging]: https://docs.microsoft.com/dotnet/azure/sdk/logging
+[ActivitySourceSupport]: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/samples/Diagnostics.md#activitysource-support
+[ActivitySource]: https://docs.microsoft.com/dotnet/api/system.diagnostics.activitysource?view=net-6.0
```
jsquire, line 297:
```suggestion
[ActivitySource]: https://docs.microsoft.com/dotnet/api/system.diagnostics.activitysource?view=dotnet
```

sdk/servicebus/Azure.Messaging.ServiceBus/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,305 @@
...
+
+### Transactions across entities are not working
+
+In order to perform transactions that involve multiple entities, you'll need to set the ServiceBusClientOptions.EnableCrossEntityTransactions property to `true`. For details, see the [Transactions across entities][CrossEntityTransactions] sample.
```
jsquire, line 260:
```suggestion
In order to perform transactions that involve multiple entities, you'll need to set the `ServiceBusClientOptions.EnableCrossEntityTransactions` property to `true`. For details, see the [Transactions across entities][CrossEntityTransactions] sample. 
```

sdk/servicebus/Azure.Messaging.ServiceBus/TROUBLESHOOTING.md
JoshLove-msft: Was trying to make this a one stop shop for all FAQ/troubleshooting. Quota questions are pretty common.

sdk/servicebus/Azure.Messaging.ServiceBus/TROUBLESHOOTING.md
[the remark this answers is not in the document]
JoshLove-msft: > I wonder if we should consider representing the "high concurrency" scenario as a factor of the available cores on the host rather than an absolute number. For a 2-core machine, 50 is crazy. For a 32-core machine, 50 is quite reasonable.

That concurrency can work okay. We throttle when accepting sessions based on the number of cores already. I plan to revise this section based on the recent developments re: the service bug.

sdk/servicebus/Azure.Messaging.ServiceBus/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,305 @@
...
+`ServiceBusSessionReceiver.SetSessionState`  
+`ServiceBusProcessor.ProcessMessage`  
+`ServiceBusSessionProcessor.ProcessSessionMessage`  
+`ServiceBusRuleManager.CreateRule`  
+`ServiceBusRuleManager.DeleteRule`  
+`ServiceBusRuleManager.GetRules`  
+
+Most of the spans are fairly self-explanatory and are started and stopped during the operation that bears its name. The span that ties the others together is `Message`. The way that the message is traced is via the the `Diagnostic-Id` that is set in the [ServiceBusMessage.ApplicationProperties][ApplicationProperties] property by the library during send and schedule operations. In Application Insights, `Message` spans will be displayed as linking out to the various other spans that were used to interact with the message, e.g. the receive span, the send span, and the complete span would all be linked from the `Message` span. Here is an example of what this looks like in Application Insights.
+
+![image](assets/Tracing.png)
```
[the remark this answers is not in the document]
JoshLove-msft: Will need to update once this is merged.

sdk/servicebus/Azure.Messaging.ServiceBus/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,309 @@
...
+  
+  - **SessionLockLost**: This indicates that the lock on the session has expired. Callers should attempt to accept the session again. This only applies to session-enabled entities. This error occurs if processing takes longer than the lock duration and the session lock is not renewed. Note that this error can also occur when the link is detached due to a transient network issue or when the link is idle for 10 minutes.
+
+  - **SessionCannotBeLocked**: This indicates that the requested session cannot be locked because the lock is already held elsewhere. Once the lock expires, the session can be accepted.
+
+### Other common exceptions
+
+- **ArgumentException** : An exception deriving from `ArgumentException` is thrown by clients when a parameter provided when interacting with the client is invalid. Information about the specific parameter and the nature of the problem can be found in the `Message`.
```
richardpark-msft, line 77: These exceptions don't seem like they're worth documenting, mostly because the _text_ of the error message for all of them should tell you what your next step is, unlike the others which might not be as intuitive.

sdk/servicebus/Azure.Messaging.ServiceBus/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,309 @@
...
+
+- `Reason` : Provides a set of well-known reasons for the failure that help to categorize and clarify the root cause. These are intended to allow for applying exception filtering and other logic where inspecting the text of an exception message wouldn't be ideal. Some key failure reasons are:
+
+  - **Service Timeout** : This indicates that the Service Bus service did not respond to an operation within the expected amount of time. This may have been caused by a transient network issue or service problem. The Service Bus service may or may not have successfully completed the request; the status is not known. In the case of accepting the next available session, this exception indicates that there were no unlocked sessions available in the entity.
+
+  - **Quota Exceeded** : This typically indicates that there are too many active receiver operations for a single entity. In order to avoid this error, reduce the number of potential concurrent receives. You can use batch receives to attempt to receive multiple messages per receive request. Please see [Service Bus quotas][ServiceBusQuotas] for more information.
```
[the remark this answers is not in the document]
richardpark-msft: btw, you mention "you can use batch receives" but is that terminology used in your library somewhere? I looked at ReceiveMessages(), for instance, and it doesn't use the term batch (it does say 'Receives a list', however). 

Are you really just saying you should ReceiveMessages( n > 1)?

sdk/servicebus/Azure.Messaging.ServiceBus/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,309 @@
...
+`ServiceBusSessionReceiver.RenewSessionLock`  
+`ServiceBusSessionReceiver.GetSessionState`  
+`ServiceBusSessionReceiver.SetSessionState`  
+`ServiceBusProcessor.ProcessMessage`  
+`ServiceBusSessionProcessor.ProcessSessionMessage`  
+`ServiceBusRuleManager.CreateRule`  
+`ServiceBusRuleManager.DeleteRule`  
+`ServiceBusRuleManager.GetRules`  
+
+Most of the spans are self-explanatory and are started and stopped during the operation that bears its name. The span that ties the others together is `Message`. The way that the message is traced is via the the `Diagnostic-Id` that is set in the [ServiceBusMessage.ApplicationProperties][ApplicationProperties] property by the library during send and schedule operations. In Application Insights, `Message` spans will be displayed as linking out to the various other spans that were used to interact with the message, e.g. the receive span, the send span, and the complete span would all be linked from the `Message` span. Here is an example of what this looks like in Application Insights:
```
richardpark-msft, line 182:
```suggestion
Most of the spans are self-explanatory and are started and stopped during the operation that bears its name. The span that ties the others together is `Message`. The way that the message is traced is via the the `Diagnostic-Id` that is set in the [ServiceBusMessage.ApplicationProperties][ApplicationProperties] property by the library during send and schedule operations. In Application Insights, `Message` spans will be displayed as linking out to the various other spans that were used to interact with the message, e.g. the `ServiceBusReceiver.Receive` span, the `ServiceBusSender.Send` span, and the `ServiceBusReceiver.Complete` span would all be linked from the `Message` span. Here is an example of what this looks like in Application Insights:
```

Just some formatting to make it more obvious these are terms, not generic words.

sdk/servicebus/Azure.Messaging.ServiceBus/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,309 @@
...
+
+  - **Message Size Exceeded** : This indicates that the max message size has been exceeded. For standard tier namespaces, a single message may not exceed 256KB. For premium tier namespaces, the max message size is 100MB. This includes the body of the message, as well as any associated metadata and system overhead. The best approach for resolving this error is to reduce the number of messages being sent in a batch or the size of the body included in the message. Because size limits are subject to change, please refer to [Service Bus quotas][ServiceBusQuotas] for specifics.  
+  
+  - **MessageLockLost** : This indicates that the lock on the message is lost. Callers should call attempt to receive and process the message again. This only applies to non-session entities. This error occurs if processing takes longer than the lock duration and the message lock is not renewed. Note that this error can also occur when the link is detached due to a transient network issue or when the link is idle for 10 minutes.
```
richardpark-msft, line 69:
```suggestion
  - **MessageLockLost** : This indicates that the lock on the message is lost. Callers should attempt to receive and process the message again. This only applies to non-session entities. This error occurs if processing takes longer than the lock duration and the message lock is not renewed. Note that this error can also occur when the link is detached due to a transient network issue or when the link is idle for 10 minutes.
```

richardpark-msft: Or maybe "call receive and attempt to process the message again"

sdk/servicebus/Azure.Messaging.ServiceBus/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,309 @@
...
+
+  - **Quota Exceeded** : This typically indicates that there are too many active receiver operations for a single entity. In order to avoid this error, reduce the number of potential concurrent receives. You can use batch receives to attempt to receive multiple messages per receive request. Please see [Service Bus quotas][ServiceBusQuotas] for more information.
+
+  - **Message Size Exceeded** : This indicates that the max message size has been exceeded. For standard tier namespaces, a single message may not exceed 256KB. For premium tier namespaces, the max message size is 100MB. This includes the body of the message, as well as any associated metadata and system overhead. The best approach for resolving this error is to reduce the number of messages being sent in a batch or the size of the body included in the message. Because size limits are subject to change, please refer to [Service Bus quotas][ServiceBusQuotas] for specifics.
```
richardpark-msft, line 67: I'm a bit averse to putting server-side limits in _our_ documentation if we can avoid it, mostly because if the doc is referred to by commit ID or copied then it can become misleadingly stale. 

Do you know if either of those will be the case here? Do we link to the troubleshooting guide in a way that it's always going to be the absolute latest in git, and not just some copy, or even a fixed commit ID?

sdk/servicebus/Azure.Messaging.ServiceBus/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,309 @@
...
+
+### Enable logging
+
+The Service Bus client logs are available to any `EventListener` by opting into the sources starting with "Azure-Messaging-ServiceBus" or by opting into all sources that have the trait "AzureEventSource". To make capturing logs from the Azure client libraries easier, the `Azure.Core` library used by Service Bus offers an `AzureEventSourceListener`.
+
+For more information, see: [Logging with the Azure SDK for .NET][Logging].
+
+### Distributed tracing
+
+The Service Bus client library supports distributed tracing though integration with the Application Insights SDK. It also supports **experimental** support for the the OpenTelemetry specification via the .NET [ActivitySource][ActivitySource] type introduced in .NET 5. In order to enable `ActivitySource` support for use with OpenTelemetry, see [ActivitySource support][ActivitySourceSupport].
```
richardpark-msft, line 155:
```suggestion
The Service Bus client library supports distributed tracing though integration with the Application Insights SDK. It also has **experimental** support for the the OpenTelemetry specification via the .NET [ActivitySource][ActivitySource] type introduced in .NET 5. In order to enable `ActivitySource` support for use with OpenTelemetry, see [ActivitySource support][ActivitySourceSupport].
```

sdk/servicebus/Azure.Messaging.ServiceBus/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,309 @@
...
+
+### Batch fails to send
+
+We define a message batch as either [ServiceBusMessageBatch][ServiceBusMessageBatch] containing 2 or more messages, or a call to [SendMessagesAsync][SendMessages] where 2 or more messages are passed in. The service does not allow a message batch to exceed 1MB. This is true whether or not the [Premium large message support][LargeMessageSupport] feature is enabled. If you intend to send a message greater than 1MB, it must be sent individually rather than grouped with other messages. Unfortunately, the [ServiceBusMessageBatch][ServiceBusMessageBatch] type does not currently support validating that a batch does not contain any messages greater than 1MB as we try to avoid hardcoding things in the client. So if you intend to use the premium large message support feature, you will need to ensure you send the messages that over 1MB individually. See this [GitHub discussion][GitHubDiscussionOnBatching] for more info.
```
richardpark-msft, line 196:
```suggestion
We define a message batch as either [ServiceBusMessageBatch][ServiceBusMessageBatch] containing 2 or more messages, or a call to [SendMessagesAsync][SendMessages] where 2 or more messages are passed in. The service does not allow a message batch to exceed 1MB. This is true whether or not the [Premium large message support][LargeMessageSupport] feature is enabled. If you intend to send a message greater than 1MB, it must be sent individually rather than grouped with other messages. Unfortunately, the [ServiceBusMessageBatch][ServiceBusMessageBatch] type does not currently support validating that a batch does not contain any messages greater than 1MB as we try to avoid hardcoding things in the client. So if you intend to use the premium large message support feature, you will need to ensure you send messages over 1MB individually. See this [GitHub discussion][GitHubDiscussionOnBatching] for more info.
```

sdk/servicebus/Azure.Messaging.ServiceBus/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,309 @@
...
+
+## Troubleshoot receiver issues
+
+### Number of messages returned does not match number requested in batch receive
+
+When attempting to do a batch receive, i.e. passing a `maxMessages` value of 2 or greater to the [ReceiveMessagesAsync][ReceiveMessages] method, you are not guaranteed to receive the number of messages requested, even if the queue or subscription has that many messages available at that time, and even if the entire configured `maxWaitTime` has not yet elapsed. To maximize throughput and avoid lock expiration, once the first message comes over the wire, the receiver will wait an additional 20ms for any additional messages before dispatching the messages for processing.  The `maxWaitTime` applies to how long the receiver will wait to receive the *first* message - subsequent messages will be waited for 20ms. Therefore, your application should not assume that all messages available will be received in one call.
```
richardpark-msft, line 202:
```suggestion
When attempting to do a batch receive, i.e. passing a `maxMessages` value of 2 or greater to the [ReceiveMessagesAsync][ReceiveMessages] method, you are not guaranteed to receive the number of messages requested, even if the queue or subscription has that many messages available at that time, and even if the entire configured `maxWaitTime` has not yet elapsed. To maximize throughput and avoid lock expiration, once the first message comes over the wire, the receiver will wait an additional 20ms for any additional messages before dispatching the messages for processing.  The `maxWaitTime` controls how long the receiver will wait to receive the *first* message - subsequent messages will be waited for 20ms. Therefore, your application should not assume that all messages available will be received in one call.
```

sdk/servicebus/Azure.Messaging.ServiceBus/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,309 @@
...
+
+### How to browse scheduled or deferred messages
+
+Scheduled and deferred messages are included when peeking messages. They can be identified by the [ServiceBusReceivedMessage.State][MessageState] property. Once you have the [SequenceNumber][SequenceNumber] of a deferred message, you can receive it with a lock via the [ReceiveDeferredMessagesAsync][ReceiveDeferredMessages] method.
+
+When working with topics, you cannot peek scheduled messages on the subscription, as the messages remain in the topic until the scheduled enqueue time. As a workaround, you can construct a [ServiceBusReceiver][ServiceBusReceiver] passing in the topic name in order to peek such messages. Note that no other operations with the receiver will work when using a topic name.
+
+### How to browse session messages across all sessions
+
+You can use a regular [ServiceBusReceiver][ServiceBusReceiver] to peek across all sessions. To peek for a specific session you can use the [ServiceBusSessionReceiver][ServiceBusSessionReceiver], but you will need to obtain a session lock.
```
richardpark-msft, line 216: Ha, I did not know you could do that!

richardpark-msft (commented):
Some comments, small typos.

Overall really nice. I particularly enjoyed the parts about configuring the processor and thread starvation.

sdk/servicebus/Azure.Messaging.ServiceBus/TROUBLESHOOTING.md
jsquire: Josh is a wizard.   He knows the things!

sdk/servicebus/Azure.Messaging.ServiceBus/TROUBLESHOOTING.md
jsquire: I share these concerns and don't believe that we can guarantee how the guide is linked/used.   I'm also uncomfortable with having to remember to tweak this for any service changes.

sdk/servicebus/Azure.Messaging.ServiceBus/TROUBLESHOOTING.md
jsquire: You'd be surprised at the feedback/questions.   There's value in having these represented for the support teams.

sdk/servicebus/Azure.Messaging.ServiceBus/TROUBLESHOOTING.md
jsquire, line 196:
```suggestion
We define a message batch as either [ServiceBusMessageBatch][ServiceBusMessageBatch] containing 2 or more messages, or a call to [SendMessagesAsync][SendMessages] where 2 or more messages are passed in. The service does not allow a message batch to exceed 1MB. This is true whether or not the [Premium large message support][LargeMessageSupport] feature is enabled. If you intend to send a message greater than 1MB, it must be sent individually rather than grouped with other messages. Unfortunately, the [ServiceBusMessageBatch][ServiceBusMessageBatch] type does not currently support validating that a batch does not contain any messages greater than 1MB as the size is constrained by the service and may change. So if you intend to use the premium large message support feature, you will need to ensure you send the messages that over 1MB individually. See this [GitHub discussion][GitHubDiscussionOnBatching] for more info.
```

jsquire (approved):
Other than the concern about including specific service limitations around size and such, looks good to me!

sdk/servicebus/Azure.Messaging.ServiceBus/TROUBLESHOOTING.md
richardpark-msft: Next time it happens it'd be interesting to see if the error message is somehow unclear. 

On the other hand, I do recall us having some issues with people noticing the actual error because it can sometimes become buried under the callstack. So I see your point. :)

sdk/servicebus/Azure.Messaging.ServiceBus/TROUBLESHOOTING.md
JoshLove-msft: Thank you!

PR #31037, opened 2022-09-09
[Identity] Adding AdditionallyAllowedTenants to constrain multi-tenant auth

schaabs:


sdk/identity/Azure.Identity/src/Credentials/DefaultAzureCredentialOptions.cs
```diff
@@ -11,11 +14,33 @@ namespace Azure.Identity
...
+                return _tenantId;
+            }
+            set
+            {
+                _tenantId = value;
+                InteractiveBrowserTenantId = value;
+                SharedTokenCacheTenantId = value;
+                VisualStudioCodeTenantId = value;
+                VisualStudioTenantId = value;
+            }
```
christothes, line 35: I think this behavior might be unexpected for developers. Should we consider throwing here if one of the individual TenantIds is set to a different value than the incoming `value`?

schaabs: It's a good point, throwing would probably be preferred to overwriting a previously set value. Do you think the error message should suggest changing the ordering, or just blankly state they can't be different values?

sdk/identity/Azure.Identity/tests/AuthorizationCodeCredentialTests.cs
```diff
@@ -71,6 +74,26 @@ public async Task AuthenticateWithAuthCodeHonorsTenantId([Values(null, TenantIdH
             Assert.AreEqual(token2.Token, expectedToken, "Should be the expected token value");
         }
 
+        [Test]
```
heaths, line 77: Tip: when attributing with `[TestCaseSource]` you don't need `[Test]` as well.

heaths: ...same for `[TestCase]` if doing inline values.

sdk/identity/Azure.Identity/tests/samples/BreakingChangesSnippets.cs
```diff
@@ -14,5 +14,25 @@ public void SetExcludeSharedTokenCacheCredentialToFalse()
             });
             #endregion
         }
+
+        public void AddExplicitAdditionallyAllowedTenants()
+        {
+            #region Snippet:Identity_BreakingChanges_AddExplicitAdditionallyAllowedTenants
+            var credential = new DefaultAzureCredential(new DefaultAzureCredentialOptions
+            {
+                AdditionallyAllowedTenants = { "0000-0000-0000-0000", "1111-1111-1111-1111" }
```
heaths, line 23: Noted above, but just in case: seems these should be valid GUIDs. If too long, just put them on separate lines.

heaths (approved):

sdk/identity/Azure.Identity/src/ManagedIdentityClient.cs
```diff
@@ -38,13 +38,16 @@ public ManagedIdentityClient(ManagedIdentityClientOptions options)
...
             Pipeline = options.Pipeline;
             _identitySource = new Lazy<ManagedIdentitySource>(() => SelectManagedIdentitySource(options));
         }
 
         internal CredentialPipeline Pipeline { get; }
 
-        protected string ClientId { get; }
+        internal protected string ClientId { get; }
+
+        internal ResourceIdentifier ResourceIdentifier { get; }
```
christothes, line 50: Where is this used?

sdk/identity/Azure.Identity/tests/AuthorizationCodeCredentialTests.cs
```diff
@@ -71,6 +74,26 @@ public async Task AuthenticateWithAuthCodeHonorsTenantId([Values(null, TenantIdH
...
+        [TestCaseSource(nameof(GetAllowedTenantsTestCasesWithRequiredTenantId))]
+        public async Task VerifyAllowedTenantEnforcement(AllowedTenantsTestParameters parameters)
+        {
+            Console.WriteLine(parameters.ToDebugString());
```
christothes, line 81: Did you intend to leave this in all the tests?

sdk/identity/Azure.Identity/tests/DefaultAzureCredentialOptionsTests.cs
```diff
@@ -0,0 +1,89 @@
...
+
+            using (new TestEnvVar("AZURE_CLIENT_ID", envVarValue))
+            {
+                var options = new DefaultAzureCredentialOptions();
+
+                Assert.AreEqual(options.ManagedIdentityClientId, expValue);
+            }
+        }
+
+        public static IEnumerable<string> IdListTestValues()
```
christothes, line 67: This appears to be unused

sdk/identity/Azure.Identity/tests/DefaultAzureCredentialTests.cs
```diff
@@ -66,302 +66,235 @@ public void ValidateCtorIncludedInteractiveParam([Values(true, false)] bool incl
...
-                Assert.AreEqual(!excludeManagedIdentityCredential, managedIdentityCredentialIncluded);
-                Assert.AreEqual(!excludeSharedTokenCacheCredential, sharedTokenCacheCredentialIncluded);
-                Assert.AreEqual(!excludeCliCredential, cliCredentialIncluded);
-                Assert.AreEqual(!excludeAzurePowerShellCredential, powerShellCredentialsIncluded);
-                Assert.AreEqual(!excludeInteractiveBrowserCredential, interactiveBrowserCredentialIncluded);
-                Assert.AreEqual(!excludeVisualStudioCredential, visualStudioCredentialIncluded);
-                Assert.AreEqual(!excludeVisualStudioCodeCredential, visualStudioCodeCredentialIncluded);
-            }
-        }
+        //public enum ManagedIdentityIdType
```
christothes, line 69: remove all the commented lines?

sdk/identity/Azure.Identity/tests/TenantIdResolverTests.cs
```diff
@@ -61,7 +61,7 @@ public static IEnumerable<object[]> ResolveInputs()
                         disableMultiTenantAuth.Value.ToString());
                 }
 
-                result = TenantIdResolver.Resolve(tenantId, context);
+                result = TenantIdResolver.Resolve(tenantId, context, TenantIdResolver.AllTenants);
```
christothes, line 64: Do we need a test for the new `AuthenticationFailedException` scenario?

sdk/identity/Azure.Identity/src/Credentials/EnvironmentCredential.cs
```diff
@@ -85,6 +92,7 @@ internal EnvironmentCredential(CredentialPipeline pipeline, TokenCredentialOptio
                         AuthorityHost = _options.AuthorityHost,
                         IsLoggingPIIEnabled = _options.IsLoggingPIIEnabled,
                         Transport = _options.Transport,
+                        AdditionallyAllowedTenantsCore = new List<string>(_options.AdditionallyAllowedTenantsCore),
```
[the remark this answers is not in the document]
schaabs: The `EnvironmentCredential` doesn't do any authentication when `GetToken(Async)` is called. It farms that call out to the underlying credential which are what ultimately authenticates, and perform the tenant id validations. In the case the underlying credential is `ClientSecretCredential` or `UsernamePassword` credential we can pass `_options` directly through, but in the case of `ClientCertificateCredential` we need to possibly send an option which is specific to `ClientCertificateCredentialOptions`, so since we can't clone the options class to the more specific subclass, we're manually copying the relevent options to a new instance of `ClientCertificateCredentialOptions`. We could add some cloning mechanisms which would allow us to create a more specific subclass of options copying all the properties, but this would require some new APIs in Azure.Core.

sdk/identity/Azure.Identity/src/Credentials/EnvironmentCredential.cs
```diff
@@ -65,6 +67,11 @@ internal EnvironmentCredential(CredentialPipeline pipeline, TokenCredentialOptio
             string username = EnvironmentVariables.Username;
             string password = EnvironmentVariables.Password;
 
+            if (_options.AdditionallyAllowedTenantsCore.Count == 0)
+            {
+                _options.AdditionallyAllowedTenantsCore = EnvironmentVariables.AdditionallyAllowedTenants;
```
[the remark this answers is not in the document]
schaabs: As implemented the only caller of this constructor which sets `AdditionallyAllowedTenantsCore` is the `DefaultAzureCredential`. Here the values have either been read from the environment (`DefaultAzureCredentialOptoins.AdditionallyAllowedTenants` defaults to the value in the environment), or have been explicitly edited in code by the application. In either case we don't want to alter the list.  

However, this is far from obvious and at least should be documented in some comments. I'd also be open to alternate suggestions. I agree it's a pretty hacky.

sdk/identity/Azure.Identity/src/ManagedIdentityClient.cs
schaabs: It's used for validation in `DefaultAzureCredentialFactoryTests.ValidateManagedIdentityCtorOptionsHonored`.

sdk/identity/Azure.Identity/src/Credentials/ClientCertificateCredentialOptions.cs
```diff
@@ -17,5 +19,10 @@ public class ClientCertificateCredentialOptions : TokenCredentialOptions, IToken
         /// Will include x5c header in client claims when acquiring a token to enable subject name / issuer based authentication for the <see cref="ClientCertificateCredential"/>.
         /// </summary>
         public bool SendCertificateChain { get; set; }
+
+        /// <summary>
+        /// For multi-tenant applications, specifies additional tenants for which the credential may acquire tokens. Add the wildcard value "*" to allow the credential to acquire tokens for any tenant the application is installed.
```
scottaddie, line 24:
```suggestion
        /// For multi-tenant applications, specifies additional tenants for which the credential may acquire tokens. Add the wildcard value "*" to allow the credential to acquire tokens for any tenant in which the application is installed.
```

sdk/identity/Azure.Identity/src/Credentials/ClientAssertionCredentialOptions.cs
```diff
@@ -11,5 +13,10 @@ public class ClientAssertionCredentialOptions : TokenCredentialOptions
         internal CredentialPipeline Pipeline { get; set; }
 
         internal MsalConfidentialClient MsalClient { get; set; }
+
+        /// <summary>
+        /// For multi-tenant applications, specifies additional tenants for which the credential may acquire tokens. Add the wildcard value "*" to allow the credential to acquire tokens for any tenant the application is installed.
```
scottaddie, line 18:
```suggestion
        /// For multi-tenant applications, specifies additional tenants for which the credential may acquire tokens. Add the wildcard value "*" to allow the credential to acquire tokens for any tenant in which the application is installed.
```

sdk/identity/Azure.Identity/src/Credentials/AzurePowerShellCredentialOptions.cs
```diff
@@ -9,8 +11,15 @@ namespace Azure.Identity
...
         /// <summary>
-        /// The Azure Active Directory tenant (directory) Id of the service principal
+        /// The tenant ID the credential will authenticate to by default. If not specified the credential will authenticate to any requested tenant, and will default to the tenant specified to the 'Connect-AzAccount' cmdlet.
         /// </summary>
         public string TenantId { get; set; }
+
+        /// <summary>
+        /// Specifies tenants in addition to the specified <see cref="TenantId"/> for which the credential may acquire tokens.
+        /// Add the wildcard value "*" to allow the credential to acquire tokens for any tenant the logged in account can access.
+        /// If no value is specified for <see cref="TenantId"/> this option will have no effect, and the credential will acquire tokens for any requested tenant.
```
scottaddie, line 21:
```suggestion
        /// If no value is specified for <see cref="TenantId"/>, this option will have no effect, and the credential will acquire tokens for any requested tenant.
```

sdk/identity/Azure.Identity/src/Credentials/AzurePowerShellCredentialOptions.cs
```diff
@@ -9,8 +11,15 @@ namespace Azure.Identity
     public class AzurePowerShellCredentialOptions : TokenCredentialOptions
     {
         /// <summary>
-        /// The Azure Active Directory tenant (directory) Id of the service principal
+        /// The tenant ID the credential will authenticate to by default. If not specified the credential will authenticate to any requested tenant, and will default to the tenant specified to the 'Connect-AzAccount' cmdlet.
```
scottaddie, line 14:
```suggestion
        /// The ID of the tenant to which the credential will authenticate by default. If not specified, the credential will authenticate to any requested tenant, and will default to the tenant provided to the 'Connect-AzAccount' cmdlet.
```

sdk/identity/Azure.Identity/src/Credentials/AzureCliCredentialOptions.cs
```diff
@@ -9,8 +11,15 @@ namespace Azure.Identity
     public class AzureCliCredentialOptions : TokenCredentialOptions
     {
         /// <summary>
-        /// The Azure Active Directory tenant (directory) Id of the service principal
+        /// The tenant ID the credential will authenticate to by default. If not specified the credential will authenticate to any requested tenant, and will default to the tenant specified to the 'az login' command.
```
scottaddie, line 14:
```suggestion
        /// The ID of the tenant to which the credential will authenticate by default. If not specified, the credential will authenticate to any requested tenant, and will default to the tenant provided to the 'az login' command.
```

sdk/identity/Azure.Identity/CHANGELOG.md
```diff
@@ -1,5 +1,30 @@
...
+    - `ClientSecretCredentialOptions`
+    - `DefaultAzureCredentialOptions`
+    - `OnBehalfOfCredentialOptions`
+    - `UsernamePasswordCredentialOptions`
+    - `VisualStudioCodeCredentialOptions`
+    - `VisualStudioCredentialOptions`
+- Added `TenantId` to `DefaultAzureCredentialOptions` to avoid having to set `InteractiveBrowserTenantId`, `SharedTokenCacheTenantId`, `VisualStudioCodeTenantId`, and `VisualStudioTenantId` individually.
+
+### Breaking Changes
+- Credential types supporting multi-tenant authentication will now throw `AuthenticationFailedException` if the requested tenant ID doesn't match the tenant id of the credential, and is not included in the `AdditionallyAllowedTenants` option. Applications must now explicitly add additional tenents to the `AdditionallyAllowedTenants` list, or add '*' to list, to enable acquiring tokens from tenants other than the originally specified tenant ID. See [BREAKING_CHANGES.md](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/identity/Azure.Identity/BREAKING_CHANGES.md#170)
```
scottaddie, line 21:
```suggestion
- Credential types supporting multi-tenant authentication will now throw `AuthenticationFailedException` if the requested tenant ID doesn't match the credential's tenant ID, and is not included in the `AdditionallyAllowedTenants` option. Applications must now explicitly add additional tenants to the `AdditionallyAllowedTenants` list, or add '*' to list, to enable acquiring tokens from tenants other than the originally specified tenant ID. See [BREAKING_CHANGES.md](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/identity/Azure.Identity/BREAKING_CHANGES.md#170).
```

sdk/identity/Azure.Identity/src/TenantIdResolver.cs
```diff
@@ -29,12 +35,34 @@ public static string Resolve(string explicitTenantId, TokenRequestContext contex
...
+        public static string[] ResolveAddionallyAllowedTenantIds(IList<string> additionallyAllowedTenants)
+        {
+            if (additionallyAllowedTenants == null)
+            {
+                return Array.Empty<string>();
+            }
+
+            if (additionallyAllowedTenants.Contains("*"))
+            {
+                return AllTenants;
```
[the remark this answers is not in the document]
schaabs: I discussed this with @KrzysztofCwalina and his opinion was we shouldn't do extra validation. I could see both sides, validation could prevent bugs, but also could be inconvenient for other cases where users want to extend this list to all tenants without having to check if other tenants were previously added. I'll look to @tg-msft as a tie breaker.

sdk/identity/Azure.Identity/tests/AuthorizationCodeCredentialTests.cs
schaabs: Yes. Given the input here is unique on each execution it's quite helpful in diagnosing test failures without having to re-run and debug.

sdk/identity/Azure.Identity/src/TenantIdResolver.cs
```diff
@@ -29,12 +35,34 @@ public static string Resolve(string explicitTenantId, TokenRequestContext contex
...
+            }
+
+            return additionallyAllowedTenants.ToArray();
```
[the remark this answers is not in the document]
schaabs: I was debating whether the complication was justified, but given it doesn't add to much complexity, I went ahead and did it.

sdk/identity/Azure.Identity/src/Credentials/DefaultAzureCredentialOptions.cs
```diff
@@ -117,5 +153,37 @@ private static string GetNonEmptyStringOrNull(string str)
         /// Specifies whether the <see cref="AzurePowerShellCredential"/> will be excluded from the <see cref="DefaultAzureCredential"/> authentication flow.
         /// </summary>
         public bool ExcludeAzurePowerShellCredential { get; set; }
+
+        internal DefaultAzureCredentialOptions ShallowClone()
```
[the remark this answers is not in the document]
schaabs: Added.

sdk/identity/Azure.Identity/src/Credentials/EnvironmentCredential.cs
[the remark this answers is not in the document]
schaabs: Added a comment to this effect.

sdk/identity/Azure.Identity/tests/DefaultAzureCredentialOptionsTests.cs
schaabs: I was trying to use it in the test below but accidently used the test source from another test. Good catch 👍

sdk/identity/Azure.Identity/tests/TenantIdResolverTests.cs
schaabs: Added

sdk/identity/Azure.Identity/src/Credentials/DefaultAzureCredentialOptions.cs
schaabs: Updated the behavior to throw if setting `TenantId` after setting one of the more specific properties, and vice-versa.

sdk/identity/Azure.Identity/src/Credentials/InteractiveBrowserCredential.cs
```diff
@@ -215,7 +218,7 @@ private async Task<AccessToken> GetTokenViaBrowserLoginAsync(TokenRequestContext
                 _ => Prompt.NoPrompt
             };
 
-            var tenantId = TenantIdResolver.Resolve(_tenantId ?? Record?.TenantId, context);
+            var tenantId = TenantIdResolver.Resolve(TenantId ?? Record?.TenantId, context, TenantIdResolver.AllTenants);
```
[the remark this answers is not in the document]
schaabs: This should have been passing `AdditionallyAllowedTenants`. I updated it to `AllTenants` before we were planning to support `AdditionallyAllowedTenants` in this credential and missed updating this instance. Good catch.

sdk/identity/Azure.Identity/src/TenantIdResolver.cs
```diff
@@ -29,12 +35,34 @@ public static string Resolve(string explicitTenantId, TokenRequestContext contex
...
+            string resolvedTenantId = disableMultiTenantAuth switch
             {
                 true => explicitTenantId,
                 false when explicitTenantId == Constants.AdfsTenantId => explicitTenantId,
                 _ => context.TenantId ?? explicitTenantId
             };
+
+            if (explicitTenantId != null && resolvedTenantId != explicitTenantId && additionallyAllowedTenantIds != AllTenants && !additionallyAllowedTenantIds.Contains(resolvedTenantId))
+            {
+                throw new AuthenticationFailedException($"The current credential is not configured to acquire tokens for tenant {resolvedTenantId}. To enable acquiring tokens for this tenant add it to the AdditionallyAllowedTenants on the credential options, or add \"*\" to AdditionallyAllowedTenants to allow acquiring tokens for any tenant.");
```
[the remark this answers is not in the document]
schaabs: Added this error to TROUBLESHOOTING.md, and put a link in the error message.

sdk/identity/Azure.Identity/src/TenantIdResolver.cs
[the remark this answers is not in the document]
tg-msft: I'd also lean toward not validating and just document that any list with `*` will accept everything regardless of other explicitly specified tenants.  It's kind of like having an `if (foo || bar || true) { ... }`.  I don't really know why you'd do it, but the C# compiler doesn't stop you.

sdk/identity/Azure.Identity/src/Credentials/DefaultAzureCredentialOptions.cs
```diff
@@ -11,12 +14,90 @@ namespace Azure.Identity
...
+        /// The ID of the tenant to which the credential will authenticate by default. If not specified, the credential will authenticate to any requested tenant, and will default to the tenant to which the chosen authentication method was originally authenticated.
+        /// </summary>
+        public string TenantId
+        {
+            get => _tenantId.Value;
+            set
+            {
+                if (_interactiveBrowserTenantId.Updated && value != _interactiveBrowserTenantId.Value)
+                {
+                    throw new InvalidOperationException("Applications should not set both TenantId and InteractiveBrowserTenantId. TenantId is preferred, and is functionally equivalent. InteractiveBrowserTenantId exists only to provide backwards compatibility.");
```
christothes, line 57: possible simplification: It might be nice to implement a method on `UpdateTracker` like `ThrowIfChangedFrom(UpdateTracker other)` and centralize all this throw logic. You'd just need an extra property to store the friendly name of the property so that this exception message could be interpolated.

christothes (approved):

sdk/identity/Azure.Identity/tests/samples/CustomCredentialSnippets.cs
```diff
@@ -68,7 +68,7 @@ public void ConfidentialClientCredentialUsage()
             #endregion
         }
 
-        #region Snippet:OnBehalfOfCredential
+        //#region Snippet:OnBehalfOfCredential
```
heaths, line 71: Nit: given it's in source control, any reason to not just remove this region declaration?

schaabs: This is actually the state that this snippet file is in right now in main: https://github.com/Azure/azure-sdk-for-net/blob/5a29f2e99c52728ea8aaa646f1101761509bc57e/sdk/identity/Azure.Identity/tests/samples/CustomCredentialSnippets.cs#L71. I left it originally since it's a code snippet many have picked up and we might have to support them. It possibly could be removed but I'm not going to do it as part of this PR.

sdk/identity/Azure.Identity/src/Credentials/DefaultAzureCredentialOptions.cs
schaabs: We could possibly consolidate this logic, but my first crack at this was a bit hacky. I'm tempted to leave the straight forward implementation here and if you feel it should be simplified create a follow up work-item. Thoughts?

christothes: I think leaving it as-is is fine.

PR #31067, opened 2022-09-12
[AzureMonitorExporter] mark exporter as internal

TimothyMothra:
### Changes
- changed `AzureMonitorTraceExporter` to `internal`.
  LogExporter and MetricExporter were already internal.


rajkumar-rangaraj (approved):

cijothomas:
Would be good to have this in changelog as this is breaking change.

PR #31073, opened 2022-09-12
[AzureMonitorExporter] storage updates

TimothyMothra:
### Changes
- refactor `StorageHelper`. 
  - Removed multiple nested if statements
  - Changed `CreateTelemetryDirectory` method to `TryCreateTelemetryDirectory`
  - Simplified `IsWindowsOS` method
- changed directory name to "Microsoft\AzureMonitor` according to spec.
- updated Changelog since this is technically a breaking change

### DIFF
--- a/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/CHANGELOG.md
+++ b/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/CHANGELOG.md
@@ -20,7 +20,8 @@
 * [Request and Dependency Success criteria will now be decided based on
   `Activity.Status`](https://github.com/Azure/azure-sdk-for-net/pull/31024)
 * [Changed `AzureMonitorTraceExporter` to internal](https://github.com/Azure/azure-sdk-for-net/pull/31067)
-* [Changed offline storage directory from "Microsoft\ApplicationInsights" to "Microsoft\AzureMonitor"](https://github.com/Azure/azure-sdk-for-net/pull/31073)
+* [Changed default offline storage directory from "Microsoft\ApplicationInsights" to "Microsoft\AzureMonitor"](https://github.com/Azure/azure-sdk-for-net/pull/31073).
+  Users may override the default location by setting `AzureMonitorExporterOptions.StorageDirectory`.
 
 ### Bugs Fixed
 
--- a/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/CHANGELOG.md
+++ b/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/CHANGELOG.md
@@ -20,7 +20,7 @@
 * [Request and Dependency Success criteria will now be decided based on
   `Activity.Status`](https://github.com/Azure/azure-sdk-for-net/pull/31024)
 * [Changed `AzureMonitorTraceExporter` to internal](https://github.com/Azure/azure-sdk-for-net/pull/31067)
-* Changed offline storage directory from "Microsoft\ApplicationInsights" to "Microsoft\AzureMonitor".
+* [Changed offline storage directory from "Microsoft\ApplicationInsights" to "Microsoft\AzureMonitor"](https://github.com/Azure/azure-sdk-for-net/pull/31073)
 
 ### Bugs Fixed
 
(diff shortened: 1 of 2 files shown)


vishweshbankwar (approved):

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/PersistentStorage/StorageHelper.cs
```diff
@@ -29,50 +25,18 @@ internal static string GetDefaultStorageDirectory()
...
-                        if (dirPath != null)
-                        {
-                            s_defaultStorageDirectory = dirPath;
-                            return s_defaultStorageDirectory;
-                        }
-                    }
-
-                    string temp = environmentVars["TEMP"]?.ToString();
-                    if (temp != null)
+                    if (TryCreateTelemetryDirectory(path: environmentVars["LOCALAPPDATA"]?.ToString(), createdDirectoryPath: out dirPath)
```
rajkumar-rangaraj, line 28: This or statement makes it very difficult read the code. Considering the removal of nested statement we are fine.

rajkumar-rangaraj (approved):

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/CHANGELOG.md
```diff
@@ -20,6 +20,7 @@
 * [Request and Dependency Success criteria will now be decided based on
   `Activity.Status`](https://github.com/Azure/azure-sdk-for-net/pull/31024)
 * [Changed `AzureMonitorTraceExporter` to internal](https://github.com/Azure/azure-sdk-for-net/pull/31067)
+* Changed offline storage directory from "Microsoft\ApplicationInsights" to "Microsoft\AzureMonitor".
```
reyang, line 23: 1. Add link to the actual PR.
2. What does this mean to the reader/user? (what does the user need to do with it?)

TimothyMothra: 1. Thanks for the catch! Updating this now.
2. Meant to be informative, if a user was expecting files in one directory that will have changed. I don't think it's actionable.

TimothyMothra: I felt this was justified for the changelog. The local storage is something we have to disclose because data here is unencrypted. [link](https://docs.microsoft.com/en-us/azure/azure-monitor/app/data-retention-privacy#does-the-sdk-create-temporary-local-storage)

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/PersistentStorage/StorageHelper.cs
TimothyMothra: discussed offline and decided to leave as is.

Changing this to a loop would be the cleanest implementation but would have some allocations that we're deciding to avoid here.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/CHANGELOG.md
reyang: > I felt this was justified for the changelog. The local storage is something we have to disclose because data here is unencrypted. [link](https://docs.microsoft.com/en-us/azure/azure-monitor/app/data-retention-privacy#does-the-sdk-create-temporary-local-storage)

+1

reyang: > 2. Meant to be informative, if a user was expecting files in one directory that will have changed. I don't think it's actionable.

Does the user have an option to explicitly specify a path/folder?

TimothyMothra: Yes! `AzureMonitorExporterOptions.StorageDirectory`
Should I call attention to this in the changelog?

reyang: Consider writing changelog in a way that makes sense to the reader. For example, instead of "updated the ABS system logic", use "adjusted the brake system to provide better handling when the road is wet/slippery, this feature will be automatically turned off when the vehicle is at 25mph or less speed".

Here are the key points:
1. What is the actual change.
2. Why should the user care about it, what is the side-effect for the user.
3. What should the user do if they don't want to be affected, or there is nothing they can do besides taking the impact.

Issue #30913, opened 2022-09-02
[Azure.Monitor.Ingestion] Add concurrency


nisha-bhatia:
Add concurrency for parallel uploads to the Upload/UploadAsync methods
Use queue

PR #31074, opened 2022-09-12
[Azure.Monitor.Ingestion] Add Concurrency Upload method

nisha-bhatia:


nisha-bhatia:
https://github.com/Azure/azure-sdk-for-net/issues/30913

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -196,6 +212,142 @@ public virtual Response<UploadLogsResult> Upload<T>(string ruleId, string stream
...
+            Response response = null;
+
+            using HttpMessage message = CreateUploadRequest(ruleId, streamName, batch.LogsData, "gzip", requestContext);
+            response = await _pipeline.ProcessMessageAsync(message, requestContext, cancellationToken).ConfigureAwait(false);
+            if (response.Status != 204) // if any error is thrown log it
+            {
+                RequestFailedException requestFailedException = new RequestFailedException(response);
+                ResponseError responseError = new ResponseError(requestFailedException.ErrorCode, requestFailedException.Message);
+                List<Object> objectLogs = new List<Object>((IEnumerable<object>)batch.LogsList);
+                errors.Add(new UploadLogsError(responseError, objectLogs));
```
JoshLove-msft, line 346: Why is errors a list? There seems to be at most one error.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -196,6 +212,142 @@ public virtual Response<UploadLogsResult> Upload<T>(string ruleId, string stream
...
+                        {
+                            Task runningTask = runningTasks[i];
+                            if (!runningTask.IsCompleted)
+                            {
+                                continue;
+                            }
+
+                            await runningTask.ConfigureAwait(false);
+                            runningTasks.RemoveAt(i);
+                            i--;
```
JoshLove-msft, line 293: This is confusing - it might be better to avoid mutating the list that we are iterating through. Here is an example where do something very similar in Service Bus - https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessor.cs#L871-L887

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
nisha-bhatia: Errors here represents the list of logs that have failed. It's always going to be one list of logs so I can update the name to represent that so it's less confusing

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -196,6 +212,142 @@ public virtual Response<UploadLogsResult> Upload<T>(string ruleId, string stream
...
+                        // Clear any completed blocks from the task list
+                        for (int i = 0; i < runningTasks.Count; i++)
+                        {
+                            Task runningTask = runningTasks[i];
+                            if (!runningTask.IsCompleted)
+                            {
+                                continue;
+                            }
+
+                            await runningTask.ConfigureAwait(false);
```
JoshLove-msft, line 291: We shouldn't need to await this task as we already know that it is completed.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
JoshLove-msft: Also, don't we need to hold onto all tasks somewhere so we can do the processing on L304?

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -196,6 +212,142 @@ public virtual Response<UploadLogsResult> Upload<T>(string ruleId, string stream
...
+                }
+            }
+            catch (Exception ex)
+            {
+                scope.Failed(ex);
+                throw;
+            }
+            finally
+            {
+                scope.Dispose();
```
JoshLove-msft, line 323: This isn't necessary as the scope is created with `using`

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -196,6 +212,142 @@ public virtual Response<UploadLogsResult> Upload<T>(string ruleId, string stream
...
+            }
+            return null;
```
JoshLove-msft, line 325: This shouldn't be necessary.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
JoshLove-msft: but aren't we only adding one value here? I'm curious why `errors` is defined as a `List<UploadLogsError>`.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -37,6 +41,18 @@ public BatchedLogs(List<T> logsList, BinaryData logsData)
             public BinaryData LogsData { get; }
         }
 
+        internal readonly struct RunningTask<T>
```
JoshLove-msft, line 44: This is pretty cool - a lot cleaner than what I did with using Tuples in the SB snippet I linked.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
JoshLove-msft: > 

This is really just a style comment so feel free to ignore.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -158,48 +175,63 @@ private static void WriteMemory(MemoryStream stream, ReadOnlyMemory<byte> memory
...
+                        batch,
+                        ruleId,
+                        streamName,
+                        false,
+                        cancellationToken).EnsureCompleted();
+                }
+
+                // Process all errors after tasks are done to determine status
+                // Will run on a single thread
+                foreach (Task<RunningTask<T>> task in runningTasks)
```
pallavit, line 210: We did not add anything to the runningTasks list? Did you intend to insert to it in the foreach loop?

JoshLove-msft: It looks like it didn't get copied over from the async implementation - https://github.com/Azure/azure-sdk-for-net/pull/31074/files#diff-6767d7871ae91cf5eb10fa3e4f24a9d5bbd9f9dbcc22bcbd72be0eabe315c103R290

It would be great to factor this out into a separate private method that can be called from both the sync and async public methods.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -158,48 +175,63 @@ private static void WriteMemory(MemoryStream stream, ReadOnlyMemory<byte> memory
...
+                // always be smaller than MaxWorkerCount
+                List<Task<RunningTask<T>>> runningTasks = new();
+                // Partition the stream into individual blocks
                 foreach (BatchedLogs<T> batch in Batch(logEntries))
                 {
-                    //TODO: catch errors and correlate with Batch.start
-                    using HttpMessage message = CreateUploadRequest(ruleId, streamName, batch.LogsData, "gzip", requestContext);
-                    response = _pipeline.ProcessMessage(message, requestContext, cancellationToken);
-                    if (response.Status != 204) // if any error is thrown log it
+                    // Because we are uploading in sequence, wait for each batch to upload before starting the next batch
```
pallavit, line 199: Why are we uploading in sequence? Does the service expects the order to be always maintained?

sdk/monitor/Azure.Monitor.Ingestion/api/Azure.Monitor.Ingestion.net461.cs
```diff
@@ -25,6 +26,12 @@ public partial class UploadLogsError
         public Azure.ResponseError Error { get { throw null; } }
         public System.Collections.Generic.IEnumerable<object> FailedLogs { get { throw null; } }
     }
+    public partial class UploadLogsOptions
+    {
+        public UploadLogsOptions() { }
+        public int MaxConcurrency { get { throw null; } set { } }
+        public Azure.Core.Serialization.ObjectSerializer ObjectSerializer { get { throw null; } set { } }
```
JoshLove-msft, line 33: nit: call the property Serializer

We used that naming in Search - https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/search/Azure.Search.Documents/api/Azure.Search.Documents.netstandard2.0.cs#L81

sdk/monitor/Azure.Monitor.Ingestion/src/Models/UploadLogsOptions.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        /// <summary>
+        /// The serializer to use to convert the log objects to JSON.
+        /// </summary>
+        public ObjectSerializer ObjectSerializer { get; set; }
+
+        /// <summary>
+        /// The max concurrent requests to send to the Azure Monitor service when uploading logs.
+        /// </summary>
+        public int MaxConcurrency { get; set; }
```
JoshLove-msft, line 21: Do we need to have any validation here to ensure users don't enter invalid values, e.g. negative numbers.

sdk/monitor/Azure.Monitor.Ingestion/src/Models/UploadLogsOptions.cs
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Azure.Monitor.Ingestion
+{
+    /// <summary>
+    /// The options model to configure the request to upload logs to Azure Monitor.
+    /// </summary>
+    public class UploadLogsOptions
+    {
+        /// <summary>
+        /// The serializer to use to convert the log objects to JSON.
```
JoshLove-msft, line 14: Should we mention what the default serializer will be? I think it would be JsonObjectSerializer which uses System.Text.Json.

sdk/monitor/Azure.Monitor.Ingestion/tests/ConcurrencyCounterPolicy.cs
```diff
@@ -0,0 +1,29 @@
...
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Azure.Core;
+using Azure.Core.Pipeline;
+
+namespace Azure.Monitor.Ingestion.Tests
+{
+    internal class ConcurrencyCounterPolicy : HttpPipelinePolicy
```
JoshLove-msft, line 12: This is cool.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
nisha-bhatia: the non-async method will run in sequence - ensure one batch has been uploaded before moving to the next. Sync methods will have just the one thread

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
nisha-bhatia: oops! forgot to add to the list.

sdk/monitor/Azure.Monitor.Ingestion/src/Models/UploadLogsOptions.cs
nisha-bhatia: will add <= 0 checking to method

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
JoshLove-msft: Chatted with @nisha-bhatia - parallel uploads are not supported when using the sync method, hence the different implementations.

JoshLove-msft: Thinking about this more, I think we should still try to factor out the implementation into a separate private method. We would just need to branch in a few places on `isAsync` to avoid awaiting tasks in the sync path. In the sync path, the tasks would all be completed whenever we get them back.

sdk/monitor/Azure.Monitor.Ingestion/tests/SessionRecords/MonitorIngestionLiveTest/ValidInputFromArrayAsJsonWithSingleBatchWithGzip.json
```diff
@@ -9,24 +9,23 @@
...
+        "x-ms-client-request-id": "18efc8d807330dab26e468ec942d4044",
         "x-ms-return-client-request-id": "true"
       },
       "RequestBody": [
         "\u001F\uFFFD\b\u0000\u0000\u0000\u0000\u0000\u0000\n",
         "\uFFFD\uFFFD;\n",
-        "\uFFFD0\u0010E\u047DLm\uFFFD\uFFFDH\uFFFD\u003E]\uFFFD\u0016\uFFFD\u0019\u0015\u0001\uFFFD\u0010\uFFFDV\b\n",
-        "\u0004\uFFFD\uFFFDn\uFFFDH7\uFFFDt\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\u000F\uFFFD\uFFFDf\uFFFD@H\u0523\uFD5A\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\u0060\uFFFD\u000B\uFFFDA\u0007c]_\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD?\uFFFD\uFFFD\uFFFDe)\uFFFD\uFFFD\uFFFD\uFFFD\u001C\uFFFD\uFFFD\uFFFDA\uFFFD=u\u0002U\uD969\uDCA9\u012B$S5\uFFFDj\uFFFDjx\uFFFD\uFFFDT\u02EBV\uFFFD\u000E\uFFFD:\uFFFD\uFFFD\u002B3\uFFFD\uFFFD\uFFFD\uFFFDW\uFFFDL\r",
-        "\uFFFD\u001A\uFFFD\uFFFD\u000E\u0000\u0000\u0000\uFFFD\uFFFD"
+        "\uFFFD0\u0010E\u047DLm\u0007\uFFFDH\uFFFD\u003E]\uFFFD\u0016\uFFFD\u0019\u0015\u0001\uFFFD\u0010\uFFFDV\b\n",
```
JoshLove-msft, line 20: FYI it looks like we will need to re-record these once https://github.com/Azure/azure-sdk-tools/pull/4165 is merged, so probably not worth spending a lot of time on the test recordings before that.

nisha-bhatia: I will change the concurrency tests to be LiveOnly for now and add the parameter to change the Upload size as well for testing.

nisha-bhatia: Now that I can reduce the size of the batch - we can have recordings for concurrency

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
nisha-bhatia: I have a helper method now that handles both sync/async

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -221,44 +255,112 @@ public virtual Response<UploadLogsResult> Upload<T>(string ruleId, string stream
...
+            }
+
+            if (response.Status != 204) // if any error is thrown log it
+            {
+                RequestFailedException requestFailedException = new RequestFailedException(response);
+                ResponseError responseError = new ResponseError(requestFailedException.ErrorCode, requestFailedException.Message);
+                List<Object> objectLogs = new List<Object>((IEnumerable<object>)batch.LogsList);
+                error = new UploadLogsError(responseError, objectLogs);
+            }
+            return new RunningTask<T>(response, null);
```
JoshLove-msft, line 361: Do we need to pass the error in here?

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -221,44 +255,112 @@ public virtual Response<UploadLogsResult> Upload<T>(string ruleId, string stream
...
+            else
+            {
+                response = _pipeline.ProcessMessage(message, requestContext, cancellationToken);
+            }
+
+            if (response.Status != 204) // if any error is thrown log it
+            {
+                RequestFailedException requestFailedException = new RequestFailedException(response);
+                ResponseError responseError = new ResponseError(requestFailedException.ErrorCode, requestFailedException.Message);
+                List<Object> objectLogs = new List<Object>((IEnumerable<object>)batch.LogsList);
```
JoshLove-msft, line 358: Why is this a list of `object`? Shouldn't it be BinaryData?

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
JoshLove-msft: We chatted about this - the desired behavior is that we don't want to throw if the server returned a response error. In this case we are not throwing anywhere so the task will not actually throw. If we get an unexpected exception, we do want to throw, which should happen as part of the `await Task.WhenAll` call. This line isn't necessary as the task is completed, and we will already handle propagating exceptions on L315.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -221,44 +255,112 @@ public virtual Response<UploadLogsResult> Upload<T>(string ruleId, string stream
...
+                List<Task<RunningTask<T>>> runningTasks = new();
+                // Partition the stream into individual blocks
+                foreach (BatchedLogs<T> batch in Batch(logs, options))
                 {
-                    //TODO: catch errors and correlate with Batch.start
-                    using HttpMessage message = CreateUploadRequest(ruleId, streamName, batch.LogsData, "gzip", requestContext);
-                    response = await _pipeline.ProcessMessageAsync(message, requestContext, cancellationToken).ConfigureAwait(false);
-                    if (response.Status != 204) // if any error is thrown log it
+                    // Start staging the next batch (but don't await the Task!)
+                    Task<RunningTask<T>> task = CommitBatchListSyncOrAsync(
```
JoshLove-msft, line 283: nit: it might be easier to reason about this if we renamed the `RunningTask` type to something like `BatchUploadResult` or `SingleUploadResult`.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -37,6 +40,20 @@ public BatchedLogs(List<T> logsList, BinaryData logsData)
             public BinaryData LogsData { get; }
         }
 
+        internal readonly struct RunningTask<T>
```
JoshLove-msft, line 43: Is the `T` needed here? It doesn't appear to be used.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -273,17 +375,27 @@ private static RequestContext GenerateRequestContext(CancellationToken cancellat
         private static UploadLogsStatus Status<T>(IEnumerable<T> logEntries, List<UploadLogsError> errors)
         {
             UploadLogsStatus status;
-            if (errors.Count == 0)
+            // Errors holds the lists of all failed logs per batch so summing up these gives us the total number of failed logs
+            int totalLogsFailed = 0;
+            foreach (var x in errors)
```
JoshLove-msft, line 380: nit: use a more descriptive variable name - maybe `error`?

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
JoshLove-msft: Related to this - should the [FailedLogs](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/monitor/Azure.Monitor.Ingestion/src/Models/UploadLogsError.cs#L25) property be an `IReadOnlyList<BinaryData>`?

sdk/monitor/Azure.Monitor.Ingestion/tests/ConcurrencyCounterPolicy.cs
```diff
@@ -0,0 +1,29 @@
...
+    {
+        public volatile int Count;
+        public override void Process(HttpMessage message, ReadOnlyMemory<HttpPipelinePolicy> pipeline)
+        {
+            Interlocked.Increment(ref Count);
```
JoshLove-msft, line 17: Should we try to add assertions directly in the Process/ProcessAsync methods to ensure the Count doesn't go above the max concurrency?

sdk/monitor/Azure.Monitor.Ingestion/tests/MonitorIngestionLiveTest.cs
```diff
@@ -159,5 +141,42 @@ public async Task InvalidInputFromObjectAsJsonNoBatchingNoGzipAsync()
...
+            UploadLogsOptions options = new UploadLogsOptions();
+            options.MaxConcurrency = 10;
+            var tasks = client.UploadAsync(TestEnvironment.DCRImmutableId, TestEnvironment.StreamName, GenerateEntries(500, Recording.Now.DateTime), options).ConfigureAwait(false);
+            Assert.Greater(policy.Count, 1);
+
+            var response = await tasks;
+
+            // Check the response
+            Assert.AreEqual(UploadLogsStatus.Success, response.Value.Status);
+            Assert.AreEqual(0, response.Value.Errors.Count());
```
JoshLove-msft, line 163: nit: can use Assert.IsEmpty

JoshLove-msft (commented):
This is looking really good. Just a few remaining comments/questions.

sdk/monitor/Azure.Monitor.Ingestion/tests/ConcurrencyCounterPolicy.cs
nisha-bhatia: Depending on if we are running in sequence or parallel the Asserts would have to be different. I thought it was easier to just do these checks in the different tests themselves

JoshLove-msft: Yes, you would have to take in the maxConcurrency as an argument to the policy's constructor. Checking in the tests will just do a one time snapshot whereas adding to the policy itself will ensure that the condition is always true.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
nisha-bhatia: yes - I think I had it as more generic before but I will update

nisha-bhatia: specifying also reduces the number of casts

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -37,6 +41,20 @@ public BatchedLogs(List<T> logsList, BinaryData logsData)
             public BinaryData LogsData { get; }
         }
 
+        internal readonly struct BatchUploadTask
```
JoshLove-msft, line 44: The part that I found confusing about the name is that it has "Task" in it so you end up having `Task<BatchUploadTask>`.

sdk/monitor/Azure.Monitor.Ingestion/tests/MonitorIngestionLiveTest.cs
```diff
@@ -159,5 +141,47 @@ public async Task InvalidInputFromObjectAsJsonNoBatchingNoGzipAsync()
...
+        {
+            var policy = new ConcurrencyCounterPolicy();
+            LogsIngestionClient client = CreateClient(policy);
+            LogsIngestionClient.SingleUploadThreshold = 1000; // make batch size smaller for Uploads for test recording size
+
+            // Make the request
+            UploadLogsOptions options = new UploadLogsOptions();
+            options.MaxConcurrency = 10;
+            System.Runtime.CompilerServices.ConfiguredTaskAwaitable<Response<UploadLogsResult>> tasks;
+            using (Recording.DisableRecording())
```
JoshLove-msft, line 158: Hmm, why do we need this? The test is marked as LiveOnly so it should not be doing any recording.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
nisha-bhatia: oh I was trying to indicate it was used for the concurrency - maybe this just makes it more confusing

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -112,7 +140,7 @@ internal static IEnumerable<BatchedLogs<T>> Batch<T>(IEnumerable<T> logEntries)
                     WriteMemory(stream, memory);
                     if ((entryCount + 1) == logEntries.Count())
                     {
-                        // reached end of logEntries and we haven't returned yet
+                        // reached end of logs and we haven't returned yet
```
JoshLove-msft, line 143: I'm curious why we are writing the JSON ourselves rather than using the System.Text.Json primitives like [Utf8JsonWriter](https://docs.microsoft.com/en-us/dotnet/api/system.text.json.utf8jsonwriter?view=net-6.0). Doesn't need to be addressed as part of this PR, but I'm interested in whether we can change the implementation so that we don't have to write the array ourselves.

nisha-bhatia: I will make a note to address this before GA.

JoshLove-msft (approved):
Looks great!

PR #31088, opened 2022-09-13
Introduce abstract base class breaking change into container registry

Yao725:


sdk/containerregistry/Azure.ResourceManager.ContainerRegistry/src/Generated/Models/UnknownContainerRegistryTaskStepProperties.cs
```diff
@@ -0,0 +1,25 @@
...
+// <auto-generated/>
+
+#nullable disable
+
+using System.Collections.Generic;
+
+namespace Azure.ResourceManager.ContainerRegistry.Models
+{
+    /// <summary> The UnknownContainerRegistryTaskStepProperties. </summary>
+    internal partial class UnknownContainerRegistryTaskStepProperties : ContainerRegistryTaskStepProperties
```
fengzhou-msft, line 13: Can we rename it to `UnknownContainerRegistryTaskStep` to align with previous naming pattern where only the base class keep the `Properties` suffix?

sdk/containerregistry/Azure.ResourceManager.ContainerRegistry/src/Azure.ResourceManager.ContainerRegistry.csproj
```diff
@@ -1,8 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <Version>1.1.0-beta.2</Version>
```
fengzhou-msft, line 3: Change the version to a stable one.

sdk/containerregistry/Azure.ResourceManager.ContainerRegistry/src/Generated/Models/UnknownContainerRegistryTaskStepProperties.cs
Yao725: Currently we can't rename this model as it's auto generated and I'd not like to use customization to do that. Considering it's an internal model, I think we can merge it first and I documented this issue [here](https://github.com/Azure/autorest.csharp/issues/2678#issuecomment-1245045891) to facilitate future support for this renaming in generator.

sdk/containerregistry/Azure.ResourceManager.ContainerRegistry/src/Azure.ResourceManager.ContainerRegistry.csproj
Yao725: The version will be updated in this prepare release [PR](https://github.com/Azure/azure-sdk-for-net/pull/31090)

fengzhou-msft (approved):

sdk/containerregistry/Azure.ResourceManager.ContainerRegistry/src/Generated/Models/UnknownContainerRegistryTaskStepProperties.cs
fengzhou-msft: I see. I didn't notice it's internal. This is added by us working as a fallback when type is not set, not swaggers?

Yao725: Yes, correct.

PR #31090, opened 2022-09-13
Prepare release for the stable rps with abstract base class change

Yao725:
Prepare release for #31088  and #31009


ArcturusZhang (approved):
Should we also include the information about the class has been changed to abstract along with the accessibility of the constructor?

Yao725:
> Should we also include the information about the class has been changed to abstract along with the accessibility of the constructor?

These breaking changes can be classified into two types, one is we change the class to be abstract and the ctor to be protected, such as Cdn, ContainerRegistey and Network. Another is we only change the ctor to be protected, such as CosmosDB and Servicelinker. That's because we already used customization to support the abstract class, just missed the ctor change. Therefore you will only see the ctor change description in the changelog of these two RPs.

sdk/containerregistry/Azure.ResourceManager.ContainerRegistry/CHANGELOG.md
```diff
@@ -1,15 +1,19 @@
 # Release History
 
-## 1.1.0-beta.2 (Unreleased)
-
-### Features Added
+## 1.1.0 (2022-09-13)
```
ArthurMa1978, line 3: release 1.0.1

Yao725: Do you mean that all these 5 RPs' version should be 1.0.1？

PR #31115, opened 2022-09-13
Initial Batch 7 (1)

ArthurMa1978:


sdk/advisor/Azure.ResourceManager.Advisor/ci.mgmt.yml
```diff
@@ -0,0 +1,13 @@
+# NOTE: Please refer to https://aka.ms/azsdk/engsys/ci-yaml before editing this file.
```
Yao725, line 1: Why do we have two ci.mgmt.yml files in this RP?

Yao725 (commented):
I think we also need to run this [command](https://github.com/Azure/azure-sdk-for-net/blame/main/eng/templates/README.md#L30) to update the yml in resource manager.

sdk/advisor/Azure.ResourceManager.Advisor/ci.mgmt.yml
ArthurMa1978: fixed

Yao725 (approved):

PR #31122, opened 2022-09-14
Fix resource string usage

### LINKED ISSUE #31053: [BUG] Service Bus - Renew Message Lock Exception
### Library name and version

Microsoft.Azure.WebJobs.Extensions.ServiceBus 5.7.0
Azure.Messaging.ServiceBus 7.10.0

### Describe the bug

An unexpected error occurs terminating the process whilst attempting to renew a message lock within a V4 Azure Function (.NET 6 in process) using the Service Bus Queue Trigger.

Using the following libraries and their versions:
Microsoft.Azure.WebJobs.Extensions.ServiceBus 5.7.0
Azure.Messaging.ServiceBus 7.10.0

### Expected behavior

The message lock is renewed for the duration set for the service bus queue.

### Actual behavior

An error occurs whilst attempting to renew the message lock throwing an exception with the following message: 

`System.Private.CoreLib: Exception while executing function: ImportServiceBus. System.Private.CoreLib: Could not find the resource "Azure.Messaging.ServiceBus.Resources.resources" among the resources "" embedded in the assembly "Microsoft.Azure.WebJobs.Extensions.ServiceBus", nor among the resources in any satellite assemblies for the specified culture. Perhaps the resources were embedded with an incorrect name.`

The actual exception is:
`System.Resources.MissingManifestResourceException`

### Reproduction Steps

To reproduce, it requires the following:

- Azure Service Bus Queue (in the example code below, this queue is called 'import queue') with sessions enabled
- V4 Azure Function (in process .NET 6) using the service bus queue trigger
- Add aforementioned packages (in bug description) to project using Nuget
- Use the following code below to replace the pre-existing run method for the function.

`public async Task Run([ServiceBusTrigger("import-queue", Connection = "ConnectionString", IsSessionsEnabled = true)] ServiceBusReceivedMessage message, ServiceBusSessionMessageActions sessionActions, ServiceBusReceiveActions receiveActions)  
{  
await sessionActions.RenewMessageLockAsync(message);
return;}`

- Run the function
- Send a message to queue and an error occurs when attempting to call the renew message lock

### Environment

Windows 10 .NET 6 (6.0.302)
Microsoft Visual Studio Community 2022 (64-bit) 17.2.6

JoshLove-msft:
Fixes https://github.com/Azure/azure-sdk-for-net/issues/31053

### DIFF
--- a/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Primitives/ServiceBusMessageActions.cs
+++ b/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Primitives/ServiceBusMessageActions.cs
@@ -245,7 +245,8 @@ namespace Microsoft.Azure.WebJobs.ServiceBus
         {
             if (_receiver is ServiceBusSessionReceiver || _sessionEventArgs != null)
             {
-                throw new InvalidOperationException(Resources.CannotLockMessageOnSessionEntity);
+                throw new InvalidOperationException("Messages cannot be locked when working with session-enabled entities. Locks are handled at the session level." +
+                                                    "In order to manually renew the session lock, bind to the 'ServiceBusSessionMessageActions' type and call 'RenewSessionLockAsync'.");
             }
             if (_receiver != null)
             {
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Core/Argument.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Core/Argument.cs
@@ -9,7 +9,7 @@ namespace Azure.Core
 {
     /// <summary>
     ///   Provides a consistent means for verifying arguments and other invariants for a given
-    ///   member.
+    ///   member. Because this can be shared across multiple projects, avoid using Resource files.
     /// </summary>
     ///
     /// <remarks>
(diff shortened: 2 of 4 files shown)


JoshLove-msft:
Also filed https://github.com/Azure/azure-sdk-for-net/issues/31124

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Primitives/ServiceBusMessageActions.cs
```diff
@@ -245,7 +245,8 @@ protected ServiceBusMessageActions()
         {
             if (_receiver is ServiceBusSessionReceiver || _sessionEventArgs != null)
             {
-                throw new InvalidOperationException(Resources.CannotLockMessageOnSessionEntity);
+                throw new InvalidOperationException("Messages cannot be locked when working with session-enabled entities. Locks are handled at the session level." +
+                                                    "In order to manually renew the session lock, bind to the 'ServiceBusSessionMessageActions' type and call 'RenewSessionLockAsync'.");
```
jsquire, line 249: Should we consider `nameof` for the method and type names?

jsquire (approved):

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Primitives/ServiceBusMessageActions.cs
JoshLove-msft: Hmm, I think we are safe here as these are public types.

PR #31149, opened 2022-09-14
Add truncation for activity exception event

vishweshbankwar:


TimothyMothra (approved):

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TraceHelper.cs
```diff
@@ -232,12 +232,12 @@ internal static MonitorBase GetExceptionDataDetailsOnTelemetryItem(IEnumerable<K
                 return null;
             }
 
-            TelemetryExceptionDetails exceptionDetails = new(exceptionMessage)
+            TelemetryExceptionDetails exceptionDetails = new(exceptionMessage.Truncate(SchemaConstants.ExceptionDetails_Message_MaxLength))
             {
-                Stack = exceptionStackTrace,
+                Stack = exceptionStackTrace.Truncate(SchemaConstants.ExceptionDetails_Stack_MaxLength),
```
rajkumar-rangaraj, line 237: Did we test this manually and does the ingestion accept the truncated stack?

vishweshbankwar: it does - the only thing that I am noticing is, it is not able to parse the stack like below
image

rajkumar-rangaraj: This is different from the experience from exception captured from Logs. Should we truncate to the stack to last carriage return on the trimmed length?

TimothyMothra: @rajkumar-rangaraj would this algorithm help here?
https://github.com/Azure/azure-sdk-for-net/blob/708c95c25b43192eed35fa657b662595a0dbd765/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Customizations/Models/TelemetryExceptionDetails.cs#L59-L61

I had forgotten about this until your last comment.

vishweshbankwar: cannot use that as we don't have access to exception object in this case.

rajkumar-rangaraj (approved):

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TraceHelper.cs
vishweshbankwar: Keeping the truncation as is for now. Will bring this up in UI testing.

PR #31154, opened 2022-09-14
[Identity] Updating scope validation

### LINKED ISSUE #30647: [BUG] Accept underline "_" for scope parameter
### Library name and version

Azure.Identity, 1.6.1

### Describe the bug

When we use Azure.Identity to acquire a token, if the `scopes` parameter contains a scope that has some not allowed characters, we would get the following error:

```
The specified scope is not in expected format. 
Only alphanumeric characters, '.', '-', ':', and '/' are allowed (Parameter 'scope')
```

For some services (like Dataverse,) they may use `_` for scope names (for example, the legacy `user_impersonation` scope), having this restrict blocks our code to acquire a token to connect to Dataverse.

I understand we can use MSAL to workaround this as it accept `_` in the scope parameter, but could we just relax this at least for `_`? The scope validation is too strict. 

A sample Dataverse scope: 
```
https://mktcxptip05070515sg891crm04.crm10.dynamics.com/user_impersonation
```

### Expected behavior

Accept `_` as part of the scope name and successfully acquire the token

### Actual behavior

Cannot pass the scope parameter validation, as error message shown above. 

### Reproduction Steps

Write code like this:
```csharp
var credential = new DefaultAzureCredential();
var scopes = new string[] { "https://mktcxptip05070515sg891crm04.crm10.dynamics.com/user_impersonation" };
var token = await credential.GetTokenAsync(new TokenRequestContext(scopes), default)
```

### Environment

Windows 11 x64
.NET 7.0 RC2

schaabs:
Relaxes scope validation to allow '_' character, for common scopes such as `user_impersonation`

Fixes #30647

### DIFF
--- a/sdk/identity/Azure.Identity/src/ScopeUtilities.cs
+++ b/sdk/identity/Azure.Identity/src/ScopeUtilities.cs
@@ -13,7 +13,7 @@ namespace Azure.Identity
     internal static class ScopeUtilities
     {
         private const string DefaultSuffix = "/.default";
-        private const string ScopePattern = "^[0-9a-zA-Z-.:/]+$";
+        private const string ScopePattern = "^[0-9a-zA-Z-_.:/]+$";
 
         private const string InvalidScopeMessage = "The specified scope is not in expected format. Only alphanumeric characters, '.', '-', ':', and '/' are allowed";
         private static readonly Regex scopeRegex = new Regex(ScopePattern);
--- a/sdk/identity/Azure.Identity/tests/ScopeUtilitiesTests.cs
+++ b/sdk/identity/Azure.Identity/tests/ScopeUtilitiesTests.cs
@@ -1,0 +1,28 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using NUnit.Framework;
+
+namespace Azure.Identity.Tests
+{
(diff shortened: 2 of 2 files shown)


sdk/identity/Azure.Identity/src/ScopeUtilities.cs
```diff
@@ -13,7 +13,7 @@ namespace Azure.Identity
     internal static class ScopeUtilities
     {
         private const string DefaultSuffix = "/.default";
-        private const string ScopePattern = "^[0-9a-zA-Z-.:/]+$";
+        private const string ScopePattern = "^[0-9a-zA-Z-_.:/]+$";
```
christothes, line 16: Should this validation be used for all sub-process based credentials? It looks like only Cli and Powershell use it today.

schaabs: Yes, it should, although it might not be at the moment. Also, when I made this update, I did a brief investigation I found that there are other possible scope characters that this regex is still not supporting. We need to do a deeper investigation of both, but given that we have 2 customers blocked by this, and `user_impersonation` is such a common scope, I think we should push out this minor fix although it might be incomplete. I created a new issue https://github.com/Azure/azure-sdk-for-net/issues/31170, to track what work remains here.

christothes (approved):

PR #31157, opened 2022-09-15
Integrate Asset-Sync Features

scbedd:
@JoshLove-msft for FYI, feedback, and 🗡️ callouts. 🗡️ 

I have no doubt that the efficiency of doing path traversal for _every test_ is pretty garbo, but I'm not even certain these are the correct places to inject what I have. Let me know what you think.

Summary of how the assets.json works (and kinda sorta the changes to the framework):

- `assets.json` sit in the language repo. These point at external recordings.
- When a recording or playback is _started_, if the additional key of `x-recording-assets-file` is provided, the test-proxy itself will shell out to `git` and perform the restore operations as necessary. The locally cloned files will be present at `<repo-root>/.assets/<10 character key>/`
  - Before you ask, we shell out so we can get your logged in creds if you have them. Gotta set a `GIT_TOKEN` env variable otherwise. 
- After the restore operation completes, the root of that restored directory is used as "repo root" and the relative path to the test file provided from `x-recording-file` should combine/align perfectly.
- If one does not provide the additional assets json, we just try to find the targeted file from the current context directory that the server is set to.

Extensive documentation for this entire process (including what a "push" looks like) _is_ coming, but I am prioritizing integration and catching gotchas before people hit em in MQ.

I should probably go after `storage` as another stress test right? I'll get on that.


sdk/core/Azure.Core.TestFramework/src/RecordedTestBase.cs
```diff
@@ -241,9 +245,79 @@ protected internal string GetSessionFilePath()
...
+                var evaluation = EvaluateDirectory(path);
+
+                if (evaluation.IsGitRoot)
+                {
+                    return path;
+                }
+                else if (evaluation.IsRoot)
+                {
+                    // talk to JoshLove about what exception should actually be thrown here when this isn't a proto
+                    throw new Exception("Search for assets.json failed. Reached system root without finding one.");
```
JoshLove-msft, line 293: The failure here is that we reached the root without finding a git root, not that we didn't find assets.json I thought.

sdk/core/Azure.Core.TestFramework/src/RecordedTestBase.cs
```diff
@@ -241,9 +245,79 @@ protected internal string GetSessionFilePath()
...
+
+                if (evaluation.AssetsJsonPresent)
+                {
+                    return Path.Combine(path, AssetsJson);
+                }
+                else if (evaluation.IsGitRoot || evaluation.IsRoot)
+                {
+                    // talk to JoshLove about what exception should actually be thrown here when this isn't a proto
+                    // and how to make this optional
+                    return null;
```
JoshLove-msft, line 316: I think this is fine for now - unless you were planning to update the entire repo to include assets.json in this PR.

sdk/core/Azure.Core.TestFramework/src/RecordedTestBase.cs
```diff
@@ -241,9 +245,79 @@ protected internal string GetSessionFilePath()
...
+                {
+                    // talk to JoshLove about what exception should actually be thrown here when this isn't a proto
+                    throw new Exception("Search for assets.json failed. Reached system root without finding one.");
+                }
+
+                path = Path.GetDirectoryName(path);
+            }
+        }
+
+        public string GetAssetsJson(string testFile)
```
JoshLove-msft, line 300:
```suggestion
        private string GetAssetsJson(string testFile)
```

sdk/core/Azure.Core.TestFramework/src/RecordedTestBase.cs
```diff
@@ -241,9 +245,79 @@ protected internal string GetSessionFilePath()
...
+
+            return new DirectoryEvaluation()
+            {
+                AssetsJsonPresent = File.Exists(assetsJsonLocation),
+                IsGitRoot = Directory.Exists(gitLocation),
+                IsRoot = new DirectoryInfo(directoryPath).Parent == null
+            };
+        }
+
+        public string GetRepoRoot()
```
JoshLove-msft, line 278:
```suggestion
        private string GetRepoRoot()
```

sdk/core/Azure.Core.TestFramework/src/RecordedTestBase.cs
```diff
@@ -241,9 +245,79 @@ protected internal string GetSessionFilePath()
...
+
+            // this needs to be updated to purely relative to repo root
+            var result = Path.Combine(
                 GetSessionFileDirectory(),
                 fileName);
+
+            return Regex.Replace(result.Replace(repoRoot, String.Empty), @"^[\\/]*", string.Empty);
+        }
+
+        private class DirectoryEvaluation
```
JoshLove-msft, line 258:
```suggestion
        private readonly struct DirectoryEvaluation
```

sdk/core/Azure.Core.TestFramework/src/RecordedTestBase.cs
```diff
@@ -241,9 +245,79 @@ protected internal string GetSessionFilePath()
...
+        {
+            public bool IsRoot;
```
JoshLove-msft, line 260:
```suggestion
            public bool IsRoot { init; get; }
```

JoshLove-msft: Similar suggestion for other fields.

sdk/core/Azure.Core.TestFramework/src/RecordedTestBase.cs
```diff
@@ -241,9 +245,79 @@ protected internal string GetSessionFilePath()
...
 
-            return Path.Combine(
+            var repoRoot = GetRepoRoot();
+
+            // this needs to be updated to purely relative to repo root
+            var result = Path.Combine(
                 GetSessionFileDirectory(),
                 fileName);
+
+            return Regex.Replace(result.Replace(repoRoot, String.Empty), @"^[\\/]*", string.Empty);
```
JoshLove-msft, line 255: Might be able to use this instead - https://learn.microsoft.com/en-us/dotnet/api/system.io.path.getrelativepath?view=net-6.0

sdk/core/Azure.Core.TestFramework/src/RecordedTestBase.cs
JoshLove-msft, line 278: We can use TestEnvironment.RepoRoot instead of this method.

sdk/core/Azure.Core.TestFramework/src/RecordedTestBase.cs
```diff
@@ -241,9 +245,79 @@ protected internal string GetSessionFilePath()
...
+
+        public string GetRepoRoot()
+        {
+            var path = GetSessionFileDirectory();
+
+            while (true)
+            {
+                var evaluation = EvaluateDirectory(path);
+
+                if (evaluation.IsGitRoot)
```
JoshLove-msft, line 286: Since we know the repo root, can we stop once we reach that path?

sdk/core/Azure.Core.TestFramework/src/RecordedTestBase.cs
```diff
@@ -241,9 +245,79 @@ protected internal string GetSessionFilePath()
...
+
+            while (true)
+            {
+                var evaluation = EvaluateDirectory(path);
+
+                if (evaluation.AssetsJsonPresent)
+                {
+                    return Path.Combine(path, AssetsJson);
+                }
+                else if (evaluation.IsGitRoot || evaluation.IsRoot)
```
JoshLove-msft, line 312: Since we know the repo root, can we check if the current path is equal to the repo root path instead?

sdk/core/Azure.Core.TestFramework/src/RecordedTestBase.cs
scbedd: We are not. I want it to be backwards compatible for certain.

sdk/core/Azure.Core.TestFramework/src/RecordedTestBase.cs
scbedd: Very possible, I'll give it a shot. The test-proxy [is a lot better](https://github.com/Azure/azure-sdk-tools/pull/4578) about handling `\` vs `/` now 👍.

sdk/core/Azure.Core.TestFramework/src/RecordedTestBase.cs
scbedd: 🤦 

will do 👍

sdk/core/Azure.Core.TestFramework/src/RecordedTestBase.cs
scbedd: Yep. With the fact that we'll just exit if we don't discover an assets.json at repo root, I have eliminated the `Directory Evaluation` type. I'll do it if we are checking for three variables, but once we go back to two I don't think it's worth the extra type def.

sdk/core/Azure.Core.TestFramework/src/RecordedTestBase.cs
scbedd: image

Hmmm. Do we need the test framework to be cross-compatible? If so, We'll probably keep the current. I'm pretty certain this is actually why I didn't use GetRelativePath to begin with actually.

sdk/core/Azure.Core.TestFramework/src/RecordedTestBase.cs
scbedd: Removed the construct entirely!

sdk/core/Azure.Core.TestFramework/src/RecordedTestBase.cs
scbedd: Not even present anymore after my latest commit. Just ripped the entire thing out 👍

sdk/core/Azure.Core.TestFramework/src/RecordedTestBase.cs
JoshLove-msft: Ah ok. Yes it would need to be able to compile against all target frameworks.

sdk/core/Azure.Core.TestFramework/src/RecordedTestBase.cs
scbedd: Ah actually. I use this in `TestRecording.InitializeProxySettingsAsync`. I wanted to base the location of the assets.json on the test file.

sdk/core/Azure.Core.TestFramework/src/TestProxy.cs
```diff
@@ -67,6 +67,7 @@ private TestProxy(string proxyPath, bool debugMode = false)
                 UseShellExecute = false,
                 RedirectStandardOutput = true,
                 RedirectStandardError = true,
+                WorkingDirectory = TestEnvironment.RepositoryRoot,
```
JoshLove-msft, line 70: It doesn't seem like this approach is working and it would also break our current debug workflow. Could we pass the context directory to the test proxy instead?

scbedd: Really? Yes, we can. I'll adjust the `arguments` to pass `--storage-location`.

scbedd: Done.

sdk/confidentialledger/Azure.Security.ConfidentialLedger/assets.json
```diff
@@ -0,0 +1,6 @@
+{
+  "AssetsRepo": "Azure/azure-sdk-assets",
+  "AssetsRepoPrefixPath": "net",
```
christothes, line 3: Is this a convention, or arbitrary? If convention, could this be derived automatically?

sdk/confidentialledger/Azure.Security.ConfidentialLedger/assets.json
```diff
@@ -0,0 +1,6 @@
...
+  "TagPrefix": "net/confidentialledger/Azure.Security.ConfidentialLedger",
+  "Tag": "net/confidentialledger/Azure.Security.ConfidentialLedger_081caadb5d"
```
christothes, line 5: If the prefix here is consistent with the `AssetsRepoPrefixPath`, should we de-dupe it from here?

christothes, line 5: Where does `081caadb5d` come from? Would it typically be auto-generated?

sdk/core/Azure.Core.TestFramework/src/Generated/Models/StartInformation.cs
```diff
@@ -25,5 +25,7 @@ public StartInformation(string xRecordingFile)
 
         /// <summary> Gets the x recording file. </summary>
         public string XRecordingFile { get; }
+        /// <summary> Gets or sets the x recording assets file. </summary>
+        public string XRecordingAssetsFile { get; set; }
```
christothes, line 29: Should this model and serialize all the properties of the file rather than the path?

JoshLove-msft: I guess the proxy looks up the full data from the path? But agreed it might be simpler to just pass all the information.

sdk/confidentialledger/Azure.Security.ConfidentialLedger/assets.json
scbedd: We could, but it's also just keeping it simple to be able to use raw.

scbedd: It is actually the SHA of assets repo when creating the tag :)

So you have a tag that you point to, but that tag also points at the actual data reference itself.

scbedd: I should be clear, I am intending on backing out my `assets.json` change. I will not be moving _anyone_ that doesn't explicitly do themselves. This is messing with your day-to-day dev and I totally realize that👍

christothes: No worries - my comments are purely about the approach.

sdk/core/Azure.Core.TestFramework/src/RecordedTestBase.cs
```diff
@@ -204,6 +209,7 @@ protected bool UseLocalDebugProxy
         protected RecordedTestBase(bool isAsync, RecordedTestMode? mode = null) : base(isAsync)
         {
             Mode = mode ?? TestEnvironment.GlobalTestMode;
+            AssetsJsonPath = GetAssetsJson();
```
scbedd, line 212: @JoshLove-msft I _believe_ this should be safe given that each test class should inherit from this guy.

It also means that we aren't running through a path traversal for every single recording start, which will actually save ourselves some disk flailing.

sdk/core/Azure.Core.TestFramework/src/RecordedTestBase.cs
```diff
@@ -241,9 +247,43 @@ protected internal string GetSessionFilePath()
...
+            var repoRoot = TestEnvironment.RepositoryRoot;
+
+            // this needs to be updated to purely relative to repo root
+            var result = Path.Combine(
                 GetSessionFileDirectory(),
                 fileName);
+
+            if (!string.IsNullOrWhiteSpace(AssetsJsonPath))
+            {
+                return Regex.Replace(result.Replace(repoRoot, String.Empty), @"^[\\/]*", string.Empty);
```
scbedd, line 259: Relative when assets.json. Absolute otherwise.

JoshLove-msft (approved):

sdk/confidentialledger/Azure.Security.ConfidentialLedger/assets.json
scbedd: We haven't built assumptions of structure into the test-proxy. This path is used as part of the path when storing your recordings.

It's derived automatically, and will have real use. For stuff going into the assets repo, it has real utility. Otherwise 🤷

sdk/core/Azure.Core.TestFramework/src/Generated/Models/StartInformation.cs
scbedd: I mean, you could pass it but the test-proxy wouldn't understand it.

The reason we're passing an assets file is that you _won't_ know where the absolute path to the recordings is when they are externalized. You'll need to get them by relative + the root path described by the assets.json cloned directory.

JoshLove-msft:
FYI @kinelski

PR #31160, opened 2022-09-15
[DeviceProvisioningServices] Polish the API

Yao725:


sdk/deviceprovisioningservices/Azure.ResourceManager.DeviceProvisioningServices/src/Generated/Models/DeviceProvisioningServicesCertificateProperties.cs
```diff
@@ -10,45 +10,45 @@
...
-            Updated = updated;
+            CreatedOn = createdOn;
+            UpdatedOn = updatedOn;
         }
 
         /// <summary> The certificate&apos;s subject name. </summary>
         public string Subject { get; }
         /// <summary> The certificate&apos;s expiration date and time. </summary>
-        public DateTimeOffset? Expiry { get; }
+        public DateTimeOffset? ExpireOn { get; }
```
Yao725, line 42: The three DateTimeOffset properties of this class are marked as "format": "date-time-rfc1123" in swagger, therefore I'm afraid that our deserialization can't work here. Maybe we need to change their type to `string`

ArthurMa1978 (approved):

sdk/deviceprovisioningservices/Azure.ResourceManager.DeviceProvisioningServices/src/Generated/Models/DeviceProvisioningServicesCertificateProperties.cs
ArthurMa1978: I think `DataTimeOffset` support rfc1123

Issue #28649, opened 2022-03-22
Provide an input binding to indicate checkpoints


andrewjw1995:
The `host.json` file allows us to specify `batchCheckpointFrequency`. When any value other than `1` is specified, there is no way to tell if the current function execution will cause a checkpoint to be saved or not.

I have some state information that I cache in-memory, and would like to flush to the persistence layer before saving the EventHub checkpoint. At the moment, the only way to synchronize the flush and the checkpoint is to run them both on every execution of the function. However, this is causing unnecessary load on our systems. What I would like to see is an input binding for the batch context, maybe like the following:

```
[FunctionName("EventHubTriggerCSharp")]
public static void Run([EventHubTrigger("foo", Connection = "Bar")] EventData[] eventHubMessages, EventDataBatchContext context)
{
    foreach (var message in eventHubMessages)
    {
        // Process the message
    }

    if (context.IsCheckpoint)
    {
        // Here we have a chance to flush our cache to a database
    }
}
```

jsquire:
@JoshLove-msft:  Would you please triage and advise?

jsquire:
Addressed by #31165

PR #31165, opened 2022-09-15
Event Hubs Trigger Checkpoint Context

creyke:
# Overview

When statefully stream processing via an Event Hubs triggered Azure Functions instance it is sometimes necessary to periodically checkpoint in-memory state (which has been materialized from multiple event batches) to durable storage.

For consistency and efficiency this checkpointing should ideally not be constrained to happening every batch of messages per partition and should arguably occur at the end of the batch of messages which will trigger a checkpoint of the event hub by the underlying `EventHubListener`, immediately prior to that checkpoint.

To do this, the application needs be aware of the Event Hubs trigger runtime's intent to checkpoint a given batch on a given partition.

# Considerations
A `PartitionContext` is already exposed by as a bindable parameter but it is immutable. It is also not publicly extendable with additional properties within the context of the Event Hubs binding library.

Currently the `EventHubListener` is unaware of whether it needs to checkpoint until after a function is invoked. This will now need to be calculated prior.

# Proposal
~Expose a new `CheckpointContext` which, initially, exposes to the application whether it `IsCheckpointingAfterInvocation`. This context in future could also provide more detail around how close the runtime is to checkpointing, for example the number of remaining invocations, so application checkpointing can be distributed uniformly over the Event Hub batch checkpointing frequency if desired.~

As per @jsquire's propsal, extend `PartitionContext` with a new publicly available class (proposing `TriggerPartitionContext`) which, initially, exposes to the application whether it `IsCheckpointingAfterInvocation`. This context in future could also provide more detail around how close the runtime is to checkpointing, for example the number of remaining invocations, so application checkpointing can be distributed uniformly over the Event Hub batch checkpointing frequency if desired.

# Usage
```C#
[FunctionName(nameof(OnEventHubTrigger))]
public async Task Run(
    [EventHubTrigger("%EventHubName%", Connection = "EventHubConnectionString")] EventData[] events,
    TriggerPartitionContext partitionContext)
    {
        await ProcessBatchAsync(events);

        if (partitionContext.IsCheckpointingAfterInvocation)
        {
            await CheckpointInMemoryStateToStorageAsync();
        }
    }
```


sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubListener.cs
```diff
@@ -212,26 +214,39 @@ public async Task ProcessEventsAsync(EventProcessorHostPartition context, IEnume
                 }
             }
 
-            private async Task CheckpointAsync(EventData checkpointEvent, EventProcessorHostPartition context)
+            private void UpdateCheckpointContext(IEnumerable<EventData> events, EventProcessorHostPartition context)
```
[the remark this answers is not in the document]
creyke: Implemented as proposed.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/CheckpointContext.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.EventHubs
+{
+    public class CheckpointContext
```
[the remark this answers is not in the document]
creyke: Implemented as proposed.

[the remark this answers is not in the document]
creyke: Happy to implement it this way, although as per `Considerations` in the PR description, would this not be a breaking change due to the public API currently binding on `PartitionContext` rather than `EventProcessorHostPartitionContext`.

I assume either binding would have to be supported to ensure that isn't the case?

Thanks for the quick feedback so far.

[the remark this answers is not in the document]
creyke: Based on this proposal I've implemented a `TriggerPartitionContext` (in case there are more properties later which are not checkpoint specific) and pushed it.

If this is agreeable, we may need to make a tweak to the unit test mocks to support it.

There's an issue now the context is referenced during tests where the `ProcessorHost` is null now the `PartitionContext` property is required during the test (when the checkpoint property is updated).

In short this is currently being worked around in the constructor of `TriggerPartitionContext` with defaulting of values but I didn't want to go further than that without API approval.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/TriggerPartitionContext.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="TriggerPartitionContext"/> class.
+        /// </summary>
+        /// <param name="fullyQualifiedNamespace">The fully qualified Event Hubs namespace this context is associated with.</param>
+        /// <param name="eventHubName">The name of the Event Hub partition this context is associated with.</param>
+        /// <param name="consumerGroup">The name of the consumer group this context is associated with.</param>
+        /// <param name="partitionId">The identifier of the Event Hub partition this context is associated with.</param>
+        public TriggerPartitionContext(string fullyQualifiedNamespace, string eventHubName, string consumerGroup, string partitionId)
+            : base(fullyQualifiedNamespace ?? "default", eventHubName ?? "default", consumerGroup ?? "default", partitionId)
```
jsquire, line 26: I don't know that we want to default missing values - they should be populated.   Were there scenarios you ran into that these were `null`?

jsquire (commented):
Thank you, @creyke!   The structure and implementation look good to me.   The one thing that we'll need to figure out is what name the new context type should have.   

Because the new type is part of the public API, it is subject to review and approval by the Azure SDK architecture board.   Once @JoshLove-msft has a chance to offer his thoughts, we'll open discussion with the right folks and see if we can get this merged in.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/api/Microsoft.Azure.WebJobs.Extensions.EventHubs.netstandard2.0.cs
```diff
@@ -49,6 +49,11 @@ public enum OffsetType
         FromEnd = 1,
         FromEnqueuedTime = 2,
     }
+    public partial class TriggerPartitionContext : Azure.Messaging.EventHubs.Consumer.PartitionContext
```
jsquire, line 52: Additional thoughts on naming:

- `EventHubTriggerPartitionContext` :  This is a mouthful, but explicitly identifies what it is and where it appears.

- `TriggerPartitionContext` : This doesn't offer a hint that it is Event Hubs-specific, but it may be enough.  Even were a developer using a non-Event Hubs trigger and referencing this package for the output binding, the association with the `PartitionContext` from the Event Hubs library would differentiate.

- `EventHubPartitionContext` : This doesn't offer a hint of the association with a Functions trigger.  My concern is that it may cause confusion and developers may assume that it is part of the Event Hubs library and not the extension.

- `PartitionContext` : This isn't really an option; even when differentiated by namespace, this would be confusing and difficult to implement in the binding logic.

jsquire:
closes #28649

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/api/Microsoft.Azure.WebJobs.Extensions.EventHubs.netstandard2.0.cs
creyke: - `HubTriggerPartitionContext` : Potential compromise with a nod to Event Hubs but maybe still too verbose.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/TriggerPartitionContext.cs
creyke: The issue is a combination of existing test coverage and mocking strategy.

With the changes proposed, the following tests were then failing:
- [ProcessEvents_SingleDispatch_CheckpointsCorrectly](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/tests/EventHubListenerTests.cs#L26-L58)
- [ProcessEvents_MultipleDispatch_CheckpointsCorrectly](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/tests/EventHubListenerTests.cs#L60-L91)
- [ProcessEvents_Failure_Checkpoints](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/tests/EventHubListenerTests.cs#L98-L134)
- [ProcessEvents_CancellationToken_CancelsExecution](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/tests/EventHubListenerTests.cs#L273-L302)

These test were then failing because they cover [ProcessEventsAsync()](https://github.com/Azure/azure-sdk-for-net/blob/d0885934d8a34a7cbccd24f9fade404737f4bf65/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubListener.cs#L151-L215) which now calls the new [UpdateCheckpointContext()](https://github.com/Azure/azure-sdk-for-net/blob/d0885934d8a34a7cbccd24f9fade404737f4bf65/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubListener.cs#L217-L238) which in turn references (thereby instantiates) the [PartitionContext](https://github.com/Azure/azure-sdk-for-net/blob/d0885934d8a34a7cbccd24f9fade404737f4bf65/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Processor/EventProcessorHostPartition.cs#L25) (which wasn't test covered before) thereby now triggering it's instantiating during the unit test.

As each of the above tests mock the [EventProcessorHost](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/tests/EventHubListenerTests.cs#L39-L44), but then Moq is unable to setup the properties passed to the partition context in the constructor of the [EventProcessorHostPartition](https://github.com/Azure/azure-sdk-for-net/blob/d0885934d8a34a7cbccd24f9fade404737f4bf65/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Processor/EventProcessorHostPartition.cs#L63-L67) (not new code) as they are not interceptable by the dyanmic proxy, null references are passed (in the test only) and are not accepted.

There are many solutions to this, but most are non-trivial changes to the mocking strategy / extracting interfaces etc., and guidance on the preferred way to balance the need to keep the existing test coverage but implement the new feature without sweeping changes to the mocks would be appreciated.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Triggers/EventHubTriggerBindingStrategy.cs
```diff
@@ -49,6 +49,7 @@ public EventData[] BindMultiple(EventHubTriggerInput value, ValueBindingContext
         public Dictionary<string, Type> GetBindingContract(bool isSingleDispatch = true)
         {
             var contract = new Dictionary<string, Type>(StringComparer.OrdinalIgnoreCase);
+            contract.Add("TriggerPartitionContext", typeof(TriggerPartitionContext));
```
JoshLove-msft, line 52: Since TriggerPartitionContext derives from PartitionContext, could we just update the existing binding of "PartitionContext" to return a TriggerPartitionContext? Is there any downside to doing so?

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/tests/EventHubTests.cs
```diff
@@ -27,7 +27,8 @@ public void GetStaticBindingContract_ReturnsExpectedValue()
             var strategy = new EventHubTriggerBindingStrategy();
             var contract = strategy.GetBindingContract();
 
-            Assert.AreEqual(7, contract.Count);
+            Assert.AreEqual(8, contract.Count);
```
JoshLove-msft, line 30: We should also add a live test in [EventHubEndToEndTests](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/tests/EventHubEndToEndTests.cs) that uses the binding and the new property.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Triggers/EventHubTriggerBindingStrategy.cs
jsquire: Because the parameter type in the function is `PartitionContext`, I believe we'd be downcast and the new member wouldn't be visible without an explicit cast in the Function body.

JoshLove-msft: Nevermind, I guess this won't work as it would break anyone binding to the original type.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/api/Microsoft.Azure.WebJobs.Extensions.EventHubs.netstandard2.0.cs
jsquire: @KrzysztofCwalina, @tg-msft:  Would you be so kind as to provide the architect perspective around naming for this?

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/tests/EventHubTests.cs
creyke: Other than this end-to-end test tweak (which I'm not confident I could validate myself due to unfamiliarity with the process), I think this is good to merge now? Is it possible for one of the MS team tweak and validate this end-to-end test?

JoshLove-msft: Sure, I can take ownership for adding a test and merging this PR. Thanks @creyke!

JoshLove-msft (approved):

PR #31169, opened 2022-09-15
Solve review comments for HDInsight

ArthurMa1978:


ArcturusZhang (approved):
rename-mapping might have issues when assigned `datetime`. This might need double confirmation

PR #31187, opened 2022-09-16
Make RetryPolicy public

JoshLove-msft:
Also fixes delay calculation to correctly use fixed delay when Mode is fixed in response to exceptions. This was only an issue for the exception path, not the response error path.


sdk/core/Azure.Core/src/Pipeline/Internal/RetryPolicy.cs
```diff
@@ -20,46 +23,59 @@ internal class RetryPolicy : HttpPipelinePolicy
...
                 {
                     if (async)
                     {
+                        await OnTryRequestAsync(message).ConfigureAwait(false);
                         await ProcessNextAsync(message, pipeline).ConfigureAwait(false);
+                        await OnResponseAsync(message).ConfigureAwait(false);
                     }
                     else
                     {
+                        OnTryRequest(message);
```
JoshLove-msft, line 72: We probably don't want to catch user exceptions in OnTryRequest/OnResponse. Instead we can move these outside of the try.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -984,6 +993,20 @@ public partial class HttpPipelineTransportOptions
         public System.Collections.Generic.IList<System.Security.Cryptography.X509Certificates.X509Certificate2> ClientCertificates { get { throw null; } }
         public System.Func<Azure.Core.Pipeline.ServerCertificateCustomValidationArgs, bool>? ServerCertificateCustomValidationCallback { get { throw null; } set { } }
     }
+    public abstract partial class RetryPolicy : Azure.Core.Pipeline.HttpPipelinePolicy
```
JoshLove-msft, line 996: I'm thinking we can probably scope down the number of methods here, but this is all of the interesting extension points I thought of.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -645,6 +647,13 @@ public partial class ResponseClassifier
         public bool TryGetValue(string name, out string? value) { throw null; }
         public bool TryGetValues(string name, out System.Collections.Generic.IEnumerable<string>? values) { throw null; }
     }
+    public partial class RetryContext
```
JoshLove-msft, line 650: Probably could be a struct.

sdk/core/Azure.Core/src/HttpMessage.cs
```diff
@@ -33,6 +33,7 @@ public HttpMessage(Request request, ResponseClassifier responseClassifier)
             Request = request;
             ResponseClassifier = responseClassifier;
             BufferResponse = true;
+            RetryContext = new RetryContext(DateTimeOffset.UtcNow);
```
christothes, line 36: Since this is on the hot path, we should consider if there is a way to only allocate this on the first retry attempt while guaranteeing that it always gets done even in with a user provided custom policy.

sdk/core/Azure.Core/src/Pipeline/Internal/RetryPolicy.cs
```diff
@@ -89,54 +95,65 @@ private async ValueTask ProcessAsync(HttpMessage message, ReadOnlyMemory<HttpPip
                     message.RetryContext.LastException = ex;
                 }
 
+                // If we got a response for this request, trigger OnResponse. We don't rely on exception because it's possible
+                // a policy later in the pipeline could throw after receiving a response, but we still want to allow OnResponse to be called
+                // in this case.
+                if (message.Request.ClientRequestId == message.Response.ClientRequestId)
```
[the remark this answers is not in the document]
JoshLove-msft: no I don't think so because the response would still be set to the previous response in case of an exception.

sdk/core/Azure.Core/src/HttpMessage.cs
christothes: Another thing to consider - should we leverage the [internal typed propertybag ](https://github.com/Azure/azure-sdk-for-net/blob/75e00583fbaf2839d9bb6773a28db81f2fe4bb34/sdk/core/Azure.Core/src/HttpMessage.cs#L161-L177)rather than growing the size of the message object?

JoshLove-msft: Yes, we can do this. There is a slight issue of this not being thread-safe, but if we are making that trade-off for perf it should be fine.

JoshLove-msft: Based on feedback from @KrzysztofCwalina I've renamed this to ProcessingContext. This will be available as a property on HttpMessage even before retrying. I'm not sure the property bag approach will work here because we can't populate this lazily.

JoshLove-msft:
Still need to add some tests/docs and prototype the GeoRD policy.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -356,6 +356,7 @@ public abstract partial class ClientOptions
         public static Azure.Core.ClientOptions Default { get { throw null; } }
         public Azure.Core.DiagnosticsOptions Diagnostics { get { throw null; } }
         public Azure.Core.RetryOptions Retry { get { throw null; } }
+        public Azure.Core.Pipeline.HttpPipelinePolicy? RetryPolicy { get { throw null; } set { } }
```
annelo-msft, line 359: Curious: Why is this on ClientOptions rather than RetryOptions?

Have we given consideration to how we might replace other "fixed" policies in the pipeline?  For example, the redirect or buffering policies?  Would we add them at this same level on ClientOptions?

sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -470,6 +472,13 @@ public static partial class MultipartResponse
         public static Azure.Response[] Parse(Azure.Response response, bool expectCrLf, System.Threading.CancellationToken cancellationToken) { throw null; }
         public static System.Threading.Tasks.Task<Azure.Response[]> ParseAsync(Azure.Response response, bool expectCrLf, System.Threading.CancellationToken cancellationToken) { throw null; }
     }
+    public partial class ProcessingContext
+    {
+        internal ProcessingContext() { }
+        public int AttemptNumber { get { throw null; } }
```
annelo-msft, line 478: Might make sense to call this "RetryAttempt" so it's clear it's related to retries, with the type being renamed s.t. it doesn't name retries specifically.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -470,6 +472,13 @@ public static partial class MultipartResponse
...
+        public System.Exception? LastException { get { throw null; } }
```
annelo-msft, line 479: Is this the last exception from the service from a retry?  Or anywhere in the library?  Would it make sense for it to be RequestFailedException?

sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -470,6 +472,13 @@ public static partial class MultipartResponse
...
+        public System.DateTimeOffset OperationStartTime { get { throw null; } }
```
annelo-msft, line 480: Is this the time the caller invoked the service method?  The time the message was created?  How would this be used by a library author/library user?

sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -984,6 +993,20 @@ public partial class HttpPipelineTransportOptions
         public System.Collections.Generic.IList<System.Security.Cryptography.X509Certificates.X509Certificate2> ClientCertificates { get { throw null; } }
         public System.Func<Azure.Core.Pipeline.ServerCertificateCustomValidationArgs, bool>? ServerCertificateCustomValidationCallback { get { throw null; } set { } }
     }
+    public abstract partial class RetryPolicy : Azure.Core.Pipeline.HttpPipelinePolicy
+    {
+        protected RetryPolicy(Azure.Core.RetryOptions? options = null) { }
+        protected virtual System.TimeSpan CalculateNextDelay(Azure.Core.HttpMessage message) { throw null; }
+        protected virtual System.Threading.Tasks.ValueTask<System.TimeSpan> CalculateNextDelayAsync(Azure.Core.HttpMessage message) { throw null; }
+        protected virtual void OnResponse(Azure.Core.HttpMessage message) { }
```
annelo-msft, line 1001: In HttpPipelineSynchronousPolicy, we have a method called `OnReceivedResponse` - would it make sense to use the same names for these?

sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -984,6 +993,20 @@ public partial class HttpPipelineTransportOptions
...
+        protected virtual System.Threading.Tasks.ValueTask OnResponseAsync(Azure.Core.HttpMessage message) { throw null; }
+        protected virtual void OnTryRequest(Azure.Core.HttpMessage message) { }
```
annelo-msft, line 1003: Similarly, in HttpPipelineSynchronousPolicy, the method is named `OnSendingRequest` - is there a reason to do something different for the retry policy?

sdk/core/Azure.Core/src/Pipeline/Internal/RetryPolicy.cs
```diff
@@ -197,11 +260,39 @@ private void GetDelay(HttpMessage message, int attempted, out TimeSpan delay)
             {
                 delay = serverDelay;
             }
+
+            return delay;
         }
 
-        private void GetDelay(int attempted, out TimeSpan delay)
+        internal virtual TimeSpan GetServerDelay(HttpMessage message)
```
annelo-msft, line 267: This makes me think it's code running on the server - do we mean GetRetryAfterValue?

JoshLove-msft: Yeah that is what this is doing. This is an existing method - the diff is a little wonky here.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
JoshLove-msft: Yeah it would be the time the message was created which should be pretty close to when the method was called. The use case for this is if someone wants to create a policy that enforces an overall time limit on the operation (including retries).

sdk/core/Azure.Core/api/Azure.Core.net461.cs
JoshLove-msft: This would be an exception that occurred as part of the pipeline - so most likely something like an IOException, SocketException, or an exception thrown from another policy. RFEs are created higher in the stack. We could wrap it in an RFE here, but I don't see the benefit.

sdk/core/Azure.Core/src/ProcessingContext.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        /// <summary>
+        /// The attempt number for the current attempt.
+        /// </summary>
+        public int AttemptNumber { get; internal set; }
+
+        /// <summary>
+        /// The exception that occurred on the previous attempt, if any.
+        /// </summary>
+        public Exception? LastException { get; internal set; }
```
annelo-msft, line 26: If someone implements their own retry policy, should they be able to modify any of these values?

sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -984,6 +993,20 @@ public partial class HttpPipelineTransportOptions
         public System.Collections.Generic.IList<System.Security.Cryptography.X509Certificates.X509Certificate2> ClientCertificates { get { throw null; } }
         public System.Func<Azure.Core.Pipeline.ServerCertificateCustomValidationArgs, bool>? ServerCertificateCustomValidationCallback { get { throw null; } set { } }
     }
+    public abstract partial class RetryPolicy : Azure.Core.Pipeline.HttpPipelinePolicy
+    {
+        protected RetryPolicy(Azure.Core.RetryOptions? options = null) { }
+        protected virtual System.TimeSpan CalculateNextDelay(Azure.Core.HttpMessage message) { throw null; }
```
annelo-msft, line 999: Would this override any server-specified retry-after values?  Might be nice to word the name to clarify.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
JoshLove-msft: > Curious: Why is this on ClientOptions rather than RetryOptions?

RetryOptions is currently protocol-agnostic and I'd like to keep it that way if possible.


> Have we given consideration to how we might replace other "fixed" policies in the pipeline? For example, the redirect or buffering policies? Would we add them at this same level on ClientOptions?

Good question - I think we would use the same approach for replacing other policies that we choose to expose.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
annelo-msft, line 1001: It took me a bit to realize that OnResponse can be called multiple times in a retry policy.  It'd be good to add this to the method's XML comments to make sure people like me who haven't had enough coffee yet get this so they can implement it accordingly.

sdk/core/Azure.Core/src/ProcessingContext.cs
```diff
@@ -0,0 +1,34 @@
...
+// Licensed under the MIT License.
+
+using System;
+
+namespace Azure.Core
+{
+    /// <summary>
+    /// Contains information related to the processing of the <see cref="HttpMessage"/>.
+    /// </summary>
+    public class ProcessingContext
```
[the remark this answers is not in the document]
JoshLove-msft: Yes, it is the full round-trip. Any policy can read from it, but currently only Core owned policies would write to it. The difference from the Properties bag is that customers can't write to it. The difference from InternalProperties is that customers can read from it.

sdk/core/Azure.Core/src/ProcessingContext.cs
JoshLove-msft: Hmm, good question. Right now it is read-only, but we could consider making these settable so that someone that is fully overriding the RetryPolicy can choose to set these properties.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
annelo-msft, line 999: Any reason not to call this `ComputeDelay`?  It sounds a bit cleaner.

JoshLove-msft: Yes, do you have a suggestion of how to incorporate that? I think in general, retry-after, is a fairly niche thing.

JoshLove-msft: I like having `Next` in there because it helps to clarify that this will be called for each retry attempt. I don't have a preference for Compute vs Calculate.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
annelo-msft, line 996: This may be a weird Anne thing, but I'm not used to seeing an abstract class without an abstract method - i.e. where it is clear from the contract what the caller is expected to implement in the base/concrete class.  Do we have expectations regarding what customers would want to override, or are we just saying "here are some extensibility points, you can use any of them"?

JoshLove-msft: > or are we just saying "here are some extensibility points, you can use any of them"?

Yep!

JoshLove-msft: Here is an example of another abstract policy with no abstract methods - https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/Pipeline/HttpPipelineSynchronousPolicy.cs

sdk/core/Azure.Core/api/Azure.Core.net461.cs
JoshLove-msft: Changed to RetryNumber.

sdk/core/Azure.Core/src/Diagnostics/AzureCoreEventSource.cs
```diff
@@ -121,7 +121,7 @@ public void ErrorResponseContentTextBlock(string requestId, int blockNumber, str
             WriteEvent(ErrorResponseContentTextBlockEvent, requestId, blockNumber, content);
         }
 
-        [Event(RequestRetryingEvent, Level = EventLevel.Informational, Message = "Request [{0}] retry number {1} took {2:00.0}s")]
+        [Event(RequestRetryingEvent, Level = EventLevel.Informational, Message = "Request [{0}] attempt number {1} took {2:00.0}s")]
```
JoshLove-msft, line 124: This was incorrect because it would say "retry 1 took X" for the first attempt.

sdk/core/Azure.Core/src/HttpMessage.cs
```diff
@@ -85,6 +86,23 @@ public Response Response
...
+        internal Exception? LastException { get; set; }
+
+        internal DateTimeOffset OperationStartTime { get; set; }
+
+        /// <summary>
+        /// The processing context for the message.
+        /// </summary>
+        public ProcessingContext ProcessingContext
+        {
+            get => _processingContext ??= new ProcessingContext(this);
```
JoshLove-msft, line 100: The point here is to lazily allocate the ProcessingContext.

KrzysztofCwalina: doesn't it need to be thread safe? BTW, if we make it a struct, then we don't have to lazy allocate it and no need for thread safety.

sdk/core/Azure.Core/src/Pipeline/DefaultRetryPolicy.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Azure.Core.Pipeline
+{
+    internal class DefaultRetryPolicy : RetryPolicy
```
KrzysztofCwalina: Why do we need this type? Why not use the RetryPolicy?

sdk/core/Azure.Core/src/ProcessingContext.cs
```diff
@@ -0,0 +1,114 @@
...
+        public DateTimeOffset OperationStartTime
+        {
+            get
+            {
+                if (_message != null)
+                {
+                    return _message.OperationStartTime;
+                }
+
+                return _operationStartTime;
```
KrzysztofCwalina, line 26: Why do we have these two modes: one when message is null and one when it's not? i.e. why would we allo this type to not be associated with a message?

KrzysztofCwalina (changes requested):

sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -984,6 +993,20 @@ public partial class HttpPipelineTransportOptions
...
+        protected virtual System.Threading.Tasks.ValueTask<System.TimeSpan> CalculateNextDelayAsync(Azure.Core.HttpMessage message) { throw null; }
+        protected virtual void OnReceivedResponse(Azure.Core.HttpMessage message) { }
+        protected virtual System.Threading.Tasks.ValueTask OnReceivedResponseAsync(Azure.Core.HttpMessage message) { throw null; }
+        protected virtual void OnSendingRequest(Azure.Core.HttpMessage message) { }
+        protected virtual System.Threading.Tasks.ValueTask OnSendingRequestAsync(Azure.Core.HttpMessage message) { throw null; }
+        public override void Process(Azure.Core.HttpMessage message, System.ReadOnlyMemory<Azure.Core.Pipeline.HttpPipelinePolicy> pipeline) { }
```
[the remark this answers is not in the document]
JoshLove-msft: I don't think so because user's may want to inject logic into Process but still call the base method, as in the GlobalTimeoutPolicy scenario:

```c#
internal class GlobalTimeoutPolicy : RetryPolicy
{
    private readonly TimeSpan _timeout;
    public GlobalTimeoutPolicy(TimeSpan timeout)
    {
        _timeout = timeout;
    }

    public override void Process(HttpMessage message, ReadOnlyMemory<HttpPipelinePolicy> pipeline)
    {
        ProcessAsyncInternal(message, pipeline, false).EnsureCompleted();
    }

    public override async ValueTask ProcessAsync(HttpMessage message, ReadOnlyMemory<HttpPipelinePolicy> pipeline)
    {
        await ProcessAsyncInternal(message, pipeline, true).ConfigureAwait(false);
    }

    private async ValueTask ProcessAsyncInternal(HttpMessage message, ReadOnlyMemory<HttpPipelinePolicy> pipeline, bool async)
    {
        // create a linked token source that we will use to cancel pipeline activity if the timeout is reached
        using var cts = CancellationTokenSource.CreateLinkedTokenSource(message.CancellationToken);
        cts.CancelAfter(_timeout);

        // replace the original cancellation token with the linked one, but store off the original
        var originalToken = message.CancellationToken;
        message.CancellationToken = cts.Token;

        try
        {
            if (async)
            {
                await base.ProcessAsync(message, pipeline).ConfigureAwait(false);
            }
            else
            {
                base.Process(message, pipeline);
            }
        }
        finally
        {
            // reset the original cancellation token
            message.CancellationToken = originalToken;
        }
    }
}
```

sdk/core/Azure.Core/src/HttpMessage.cs
JoshLove-msft: No longer lazily allocating.

sdk/core/Azure.Core/src/Pipeline/DefaultRetryPolicy.cs
JoshLove-msft: This is an internal concrete implementation of the abstract policy.

sdk/core/Azure.Core/src/Diagnostics/AzureCoreEventSource.cs
JoshLove-msft: I think this is a fancy way of saying use the {2} parameter but apply a format string.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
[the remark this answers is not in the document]
KrzysztofCwalina: OK. Makes sense.

sdk/core/Azure.Core/src/HttpMessage.cs
```diff
@@ -33,6 +33,8 @@ public HttpMessage(Request request, ResponseClassifier responseClassifier)
             Request = request;
             ResponseClassifier = responseClassifier;
             BufferResponse = true;
+            OperationStartTime = DateTimeOffset.UtcNow;
+            ProcessingContext = new ProcessingContext(this);
```
KrzysztofCwalina, line 37: We should not have a field that stores the context. It just makes the class larger. Instead, the property that returns the context, should just return "new ProcessingContext(this)"

sdk/core/Azure.Core/src/Pipeline/HttpPipelineBuilder.cs
```diff
@@ -152,8 +152,7 @@ void AddNonNullPolicies(HttpPipelinePolicy[] policiesToAdd)
                 policies.Add(CreateTelemetryPolicy(buildOptions.ClientOptions));
             }
 
-            RetryOptions retryOptions = buildOptions.ClientOptions.Retry;
-            policies.Add(new RetryPolicy(retryOptions.Mode, retryOptions.Delay, retryOptions.MaxDelay, retryOptions.MaxRetries));
+            policies.Add(buildOptions.ClientOptions.RetryPolicy ?? new DefaultRetryPolicy(buildOptions.ClientOptions.Retry));
```
KrzysztofCwalina, line 155: Does this honor the static client options?

JoshLove-msft: Good catch.

JoshLove-msft: Fixed, will need to add tests.

sdk/core/Azure.Core/src/Pipeline/DefaultRetryPolicy.cs
JoshLove-msft: Will consider just making RetryPolicy concrete instead.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
JoshLove-msft: We explored this question here - https://github.com/Azure/azure-sdk-for-net/pull/31892#discussion_r999940604

The consensus was that using a single property is preferable as we don't have a lot of reason to expect the set of replaceable policies to grow beyond this one.

sdk/core/Azure.Core/src/ProcessingContext.cs
```diff
@@ -0,0 +1,51 @@
...
+// Licensed under the MIT License.
+
+using System;
+
+namespace Azure.Core
+{
+    /// <summary>
+    /// Contains information related to the processing of the <see cref="HttpMessage"/>.
+    /// </summary>
+    public readonly struct ProcessingContext
```
christothes, line 11: Josh and I measured this just to validate it does perform better from an allocation perspective over modeling this as a class with local properties.

```
|     Method |     Mean |   Error |  StdDev | Ratio | RatioSD |  Gen 0 | Allocated |
|----------- |---------:|--------:|--------:|------:|--------:|-------:|----------:|
| StructType | 177.1 ns | 3.40 ns | 3.78 ns |  0.97 |    0.04 | 0.0057 |      48 B |
|  ClaseType | 182.4 ns | 3.67 ns | 3.93 ns |  1.00 |    0.00 | 0.0086 |      72 B |
```

sdk/core/Azure.Core/src/ClientOptions.cs
```diff
@@ -90,6 +91,14 @@ public HttpPipelineTransport Transport
         /// </summary>
         public RetryOptions Retry { get; }
 
+        /// <summary>
+        /// The policy to use for retries. If a policy is specified, it will be used in place of the <see cref="Retry"/> property.
```
christothes, line 95:
```suggestion
        /// Gets or sets the policy to use for retries. If a policy is specified, it will be used in place of the <see cref="Retry"/> property.
```

sdk/core/Azure.Core/src/ClientOptions.cs
```diff
@@ -90,6 +91,14 @@ public HttpPipelineTransport Transport
...
+        /// If no policy is specified, a default policy respecting the <see cref="Retry"/> property will be used. The <see cref="RetryPolicy"/> type can be
```
christothes, line 96:
```suggestion
        /// If no policy is set, a default policy respecting the <see cref="Retry"/> property will be used. The <see cref="RetryPolicy"/> type can be
```

sdk/core/Azure.Core/src/ClientOptions.cs
```diff
@@ -90,6 +91,14 @@ public HttpPipelineTransport Transport
...
+        /// derived from to introduce a retry policy that does not require implementing the full retry logic. If <see cref="RetryPolicy.Process"/> is overriden or a
```
christothes, line 97:
```suggestion
        /// derived from to modify the default behavior without needing to fully implement the retry logic. If <see cref="RetryPolicy.Process"/> is overriden or a
```

sdk/core/Azure.Core/src/ClientOptions.cs
```diff
@@ -90,6 +91,14 @@ public HttpPipelineTransport Transport
...
+        /// custom <see cref="HttpPipelinePolicy"/> is specified, it is the implementer's responsibility to populate the <see cref="HttpMessage.ProcessingContext"/> property.
```
christothes, line 98: We're not going to do this for them in a helper policy?

sdk/core/Azure.Core/src/HttpMessage.cs
```diff
@@ -33,6 +33,7 @@ public HttpMessage(Request request, ResponseClassifier responseClassifier)
             Request = request;
             ResponseClassifier = responseClassifier;
             BufferResponse = true;
+            ProcessingStartTime = DateTimeOffset.UtcNow;
```
christothes, line 36: Might be an edge case, but should we set this in the first policy in the pipeline instead? In theory the message could be created and not immediately sent/processed.

sdk/core/Azure.Core/src/Pipeline/RetryPolicy.cs
```diff
@@ -0,0 +1,331 @@
...
+
+        /// <summary>
+        /// This method can be overriden to control how long to delay before retrying. This method will only be called for async methods.
+        /// </summary>
+        /// <param name="message">The message containing the request and response.</param>
+        /// <returns>The amount of time to delay before retrying.</returns>
+        protected virtual ValueTask<TimeSpan> CalculateNextDelayAsync(HttpMessage message) => new(CalculateNextDelayInternal(message));
+
+        /// <summary>
+        /// This method can be overriden to introduce logic before the request is sent. This will run even for the first attempt.
```
christothes, line 240:
```suggestion
        /// This method can be overridden to introduce logic before each request attempt is sent. This will run even for the first attempt.
```

sdk/core/Azure.Core/src/Pipeline/RetryPolicy.cs
```diff
@@ -0,0 +1,331 @@
...
+
+        /// <summary>
+        /// This method can be overriden to introduce logic that runs before the request is sent. This will run even for the first attempt.
+        /// This method will only be called for async methods.
+        /// </summary>
+        /// <param name="message">The message containing the request and response.</param>
+        protected virtual ValueTask OnSendingRequestAsync(HttpMessage message) => default;
+
+        /// <summary>
+        /// This method can be overriden to introduce logic that runs after a response is received. This method will only be called for sync methods.
```
christothes, line 256:
```suggestion
        /// This method can be overridden to introduce logic that runs after a response is received on each attempt. This method will only be called for sync methods.
```

sdk/core/Azure.Core/src/ProcessingContext.cs
KrzysztofCwalina: 24 bytes savings! :-)

KrzysztofCwalina (approved):

sdk/core/Azure.Core/src/Pipeline/RetryPolicy.cs
```diff
@@ -0,0 +1,331 @@
...
+using Azure.Core.Diagnostics;
+
+namespace Azure.Core.Pipeline
+{
+    /// <summary>
+    /// Represents a policy that can be overriden to customize whether or not a request will be retried and how long to wait before retrying.
+    /// </summary>
+    public abstract class RetryPolicy : HttpPipelinePolicy
+    {
+        private readonly RetryMode _mode;
```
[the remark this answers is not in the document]
JoshLove-msft: I don't think this would work since the user may not override this method, so we would still need the logic to live in the base public class and have it respect the options. But perhaps I'm misunderstanding.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -470,6 +472,15 @@ public static partial class MultipartResponse
...
         public static System.Threading.Tasks.Task<Azure.Response[]> ParseAsync(Azure.Response response, bool expectCrLf, System.Threading.CancellationToken cancellationToken) { throw null; }
     }
+    [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
+    public readonly partial struct ProcessingContext
+    {
+        private readonly object _dummy;
+        private readonly int _dummyPrimitive;
+        public System.Exception? LastException { get { throw null; } set { } }
+        public System.DateTimeOffset ProcessingStartTime { get { throw null; } set { } }
+        public int RetryNumber { get { throw null; } set { } }
```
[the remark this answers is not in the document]
JoshLove-msft: This is what I had initially, but @KrzysztofCwalina pointed out that it wasn't clear what you are attempting. I guess we could go with RequestAttemptCount, but I think RetryNumber clearly implies that it will be 1 for the first retry.

sdk/core/Azure.Core/src/ClientOptions.cs
JoshLove-msft: Nope - we discussed this and agreed that it is not worth adding a new policy.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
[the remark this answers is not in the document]
JoshLove-msft: Asked and answered 😆 - https://github.com/Azure/azure-sdk-for-net/pull/31187#discussion_r989291774

sdk/core/Azure.Core/src/HttpMessage.cs
JoshLove-msft: Fair point - though I don't really like the idea of coupling a particular policy to the fact that it is the first in the pipeline. Users can always reset this if they want in a custom policy.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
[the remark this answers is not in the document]
annelo-msft: Why not `RetryCount` like @KrzysztofCwalina suggested?

sdk/core/Azure.Core/src/HttpMessage.cs
annelo-msft: We could set it in Pipeline.Send before starting the message down the pipeline if we wanted to ensure it got called at the very beginning without relying on policy ordering.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -470,6 +472,15 @@ public static partial class MultipartResponse
         public static Azure.Response[] Parse(Azure.Response response, bool expectCrLf, System.Threading.CancellationToken cancellationToken) { throw null; }
         public static System.Threading.Tasks.Task<Azure.Response[]> ParseAsync(Azure.Response response, bool expectCrLf, System.Threading.CancellationToken cancellationToken) { throw null; }
     }
+    [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
+    public readonly partial struct ProcessingContext
+    {
+        private readonly object _dummy;
+        private readonly int _dummyPrimitive;
+        public System.Exception? LastException { get { throw null; } set { } }
+        public System.DateTimeOffset ProcessingStartTime { get { throw null; } set { } }
```
annelo-msft, line 481: Do we need the `Processing` at the beginning of `StartTime` if it is on a type called `ProcessingContext`?

sdk/core/Azure.Core/api/Azure.Core.net461.cs
annelo-msft: I'm still a little nervous about letting libraries override a service-specified retry-after header.  It would let people DOS services with our libraries, possibly accidentally.

annelo-msft: It could be implemented by calling the virtual method from inside a non-virtual method that uses the service's retry-after value if it's provided.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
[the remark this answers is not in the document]
JoshLove-msft: I chatted with @KrzysztofCwalina and he was fine with either RetryNumber or RetryCount. I don't have a strong preference.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
JoshLove-msft: My opinion is that this is an advanced feature to allow users to have more control. As we've discussed previously, users can get into bad situations when modifying/adding policies if they are not careful. The user would have access to the retry-after header so they can use it in their implementation.

sdk/core/Azure.Core/src/HttpMessage.cs
JoshLove-msft: Makes sense.

sdk/core/Azure.Core/src/Pipeline/RetryPolicy.cs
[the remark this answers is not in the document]
annelo-msft: I like the idea of making internal subclasses do something, and I'd be interested to see how much you could simplify the implementation by doing this.

If we don't do that, do we even need to make this abstract with an internal implementation?  People could just subclass the type and optionally override virtual methods.

sdk/core/Azure.Core/src/ProcessingContext.cs
```diff
@@ -0,0 +1,51 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+
+namespace Azure.Core
+{
+    /// <summary>
+    /// Contains information related to the processing of the <see cref="HttpMessage"/>.
```
annelo-msft, line 9: It might be worth mentioning that processing means what happens when we send the message down and back the pipeline.

sdk/core/Azure.Core/src/Pipeline/RetryPolicy.cs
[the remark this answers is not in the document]
JoshLove-msft: Fine with making this non-abstract if that is what the consensus is. I preferred abstract because IMO it communicates to users the intent of the type - it is meant to be overriden. 

As for using the two subclasses for exponential and fixed, as I mentioned above, I don't think this will work given that customers can subclass the policy without implementing the CalculateNextDelay method.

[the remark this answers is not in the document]
annelo-msft: > I don't think this will work given that customers can subclass the policy without implementing the CalculateNextDelay method.

Hm, that is kind of interesting.  Have we clarified and documented the expected semantics regarding how a replaced RetryPolicy would work with any RetryOptions set on ClientOptions?

annelo-msft (commented):
Was there interest in addressing how library authors would implement changes to retry logic?  Or is that an additive feature based on customer request?

JoshLove-msft:
> Was there interest in addressing how library authors would implement changes to retry logic? Or is that an additive feature based on customer request?

The retry policy piece is not intended for library authors. Library authors can use the ProcessingContext in their policies. I will document this.

sdk/core/Azure.Core/src/Pipeline/RetryPolicy.cs
```diff
@@ -0,0 +1,331 @@
...
+        private readonly TimeSpan _delay;
+        private readonly TimeSpan _maxDelay;
+        private readonly int _maxRetries;
+
+        private readonly Random _random = new ThreadSafeRandom();
```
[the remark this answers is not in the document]
JoshLove-msft: I don't think we would want that as it could cause more contention between different client/pipeline instances. I do think this type can be improved though to avoid using locks similar to what we do in SB/EH - https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Azure.Messaging.EventHubs/src/Amqp/AmqpConnectionScope.cs#L53

annelo-msft:
> The retry policy piece is not intended for library authors. Library authors can use the ProcessingContext in their policies. I will document this.

So if I'm a library author and I wanted to use a custom retry policy for my library, that's not allowed?

I'm largely interested in this due to the discussion about whether we should allow turning off redirects in any library, or only in the ACR library by letting us put a different redirect policy in the pipeline.  If we decided to do the latter, how would we explain that library authors can specify a custom redirect policy, but can't modify the retry policy?

JoshLove-msft:
> > The retry policy piece is not intended for library authors. Library authors can use the ProcessingContext in their policies. I will document this.
> 
> So if I'm a library author and I wanted to use a custom retry policy for my library, that's not allowed?
> 
> I'm largely interested in this due to the discussion about whether we should allow turning off redirects in any library, or only in the ACR library by letting us put a different redirect policy in the pipeline. If we decided to do the latter, how would we explain that library authors can specify a custom redirect policy, but can't modify the retry policy?

We aren't stopping library authors from setting a custom retry policy, but being that the property is on ClientOptions, we are communicating that this is intended to be an end user feature. My understanding is that the redirect policy is meant to be replaced by library authors rather than customers. As long as it is one or the other, it should be okay.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
JoshLove-msft: Will change

sdk/core/Azure.Core/src/Pipeline/RetryPolicy.cs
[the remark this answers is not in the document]
JoshLove-msft: I will document this. The way it works is that the options can be passed into the policy so the user can choose to respect the options for any overridden methods. For methods that are not overridden, the base policy will respect the options.

sdk/core/Azure.Core/tests/FixedRetryPolicyTests.cs
```diff
@@ -54,12 +55,34 @@ public async Task WaitsSameAmountEveryTime()
             Assert.AreEqual(500, response.Status);
         }
 
+        [Test]
+        public async Task WaitsSameAmountBetweenTries_Exceptions()
```
JoshLove-msft, line 59: This test exercises a fix in this PR. Previously, the delay would always be exponential (even when the mode was Fixed) when computing the delay after an exception - https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/Pipeline/Internal/RetryPolicy.cs#L204.

sdk/core/Azure.Core/src/ProcessingContext.cs
JoshLove-msft: Let me know if you are okay with the update.

sdk/core/Azure.Core/src/Pipeline/RetryPolicy.cs
```diff
@@ -0,0 +1,331 @@
...
+
+                    exceptions.Add(ex);
+
+                    message.LastException = ex;
+                }
+
+                // If we got a response for this request, trigger OnResponse. We can't assume that an exception being thrown means there was no response
+                // since it's possible a policy later in the pipeline could throw after receiving a response. We need to compare the request/response count
+                // to make sure that the response is for this attempt.
+                if (message.HasResponse && message.ResponseNumber == message.RetryNumber + 1)
```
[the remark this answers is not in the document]
JoshLove-msft: Good idea!

sdk/core/Azure.Core/src/Pipeline/RetryPolicy.cs
```diff
@@ -106,10 +107,8 @@ private async ValueTask ProcessAsync(HttpMessage message, ReadOnlyMemory<HttpPip
                     message.LastException = ex;
                 }
 
-                // If we got a response for this request, trigger OnResponse. We can't assume that an exception being thrown means there was no response
-                // since it's possible a policy later in the pipeline could throw after receiving a response. We need to compare the request/response count
-                // to make sure that the response is for this attempt.
-                if (message.HasResponse && message.ResponseNumber == message.RetryNumber + 1)
+                // If we got a response for this request, trigger OnResponse.
+                if (message.HasResponse && !ReferenceEquals(message.Response, lastResponse))
```
KrzysztofCwalina, line 111: I consider this a hack, but maybe we can take it if there are no other options

sdk/core/Azure.Core/src/Pipeline/RetryPolicy.cs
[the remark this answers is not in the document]
JoshLove-msft: Added a sample - https://github.com/Azure/azure-sdk-for-net/pull/31187/files#diff-72428161e5066dbbc5e348e6031475d84f40c2b1241d6d4d8503e945bf217469R25

sdk/core/Azure.Core/src/Pipeline/DefaultRetryPolicy.cs
JoshLove-msft: For now I will leave it as abstract. We can discuss further during API review.

sdk/core/Azure.Core/tests/RetryPolicyTestBase.cs
```diff
@@ -62,6 +64,71 @@ public async Task OnlyRetriesRetriableCodes()
             Assert.AreEqual(501, response.Status);
         }
 
+        [Test]
+        public async Task ShouldRetryIsCalledOnlyForErrors()
+        {
+            (HttpPipelinePolicy policy, AsyncGate<TimeSpan, object> gate) = CreateRetryPolicy(maxRetries: 3);
```
christothes, line 70: consider factoring duplicate setup code into a helper method just to clarify what is different about each test scenario.

christothes (approved):

sdk/core/Azure.Core/api/Azure.Core.net461.cs
JoshLove-msft: I'm not sure if we would want it to be virtual as this might make users think they need to override it. What if we go with protected static?

sdk/core/Azure.Core/src/ClientOptions.cs
```diff
@@ -90,6 +91,14 @@ public HttpPipelineTransport Transport
         /// </summary>
         public RetryOptions Retry { get; }
 
+        /// <summary>
+        /// Gets or sets the policy to use for retries. If a policy is specified, it will be used in place of the <see cref="Retry"/> property.
```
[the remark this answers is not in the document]
JoshLove-msft: This isn't quite accurate. The options passed to the retry policy are passed separately, so I think the current docs are correct as-is.

sdk/core/Azure.Core/src/Pipeline/RetryPolicy.cs
```diff
@@ -0,0 +1,328 @@
...
+                catch (Exception ex)
+                {
+                    if (exceptions == null)
+                    {
+                        exceptions = new List<Exception>();
+                    }
+
+                    exceptions.Add(ex);
+
+                    message.LastException = ex;
```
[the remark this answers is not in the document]
JoshLove-msft: Not sure it is worth any potential perf impact.

sdk/core/Azure.Core/src/Pipeline/RetryPolicy.cs
```diff
@@ -286,7 +287,14 @@ private TimeSpan CalculateNextDelayInternal(HttpMessage message)
...
 
-        internal virtual TimeSpan GetServerDelay(HttpMessage message)
+        /// <summary>
+        /// Gets the server specified delay. If the message has no response, <see cref="TimeSpan.Zero"/> is returned.
+        /// This method can be used to help calculate the next delay when overriding <see cref="CalculateNextDelay(HttpMessage)"/>, i.e.
+        /// implementors may want to add the server delay to their own custom delay.
+        /// </summary>
+        /// <param name="message">The message to inspect for the server specified delay.</param>
+        /// <returns>The server specified delay.</returns>
+        protected static TimeSpan GetServerDelay(HttpMessage message)
```
JoshLove-msft, line 297: FYI @annelo-msft

Issue #30717, opened 2022-08-24
[Core] Introduce JsonData type in Azure.Core


annelo-msft:
### API Change Needed

Yes

### Background and motivation

The JsonData type provides a dynamic layer for accessing JSON and is intended to improve the developer experience, over JsonDocument which we recommend to users today, when working with protocol methods.  This type will improve usability for raw JSON that comes back from protocol methods as BinaryData.

#### Goals
- Make samples showing clients with protocol methods look like .NET codecode calling protocol methods easier to read.
- Make code reading semi-structured data, e.g. code reading storage table data, easier to read.
- Make it easy easier for customers to update code when Azure SDK-supported models are added to library in a later version

#### Scenarios
- Use content from a protocol method’s response through a dynamic layer, as though it were a model type
- Use content from a pageable service method, treating each element in a returned collection as an individual dynamic, e.g. `Pageable<dynamic>`
- Use content from an LRO, treating the final returned value as the dynamic content, e.g. Operation<dynamic>
- Round-tripping is possible, i.e. get payload from response, update a value in payload, send in request.
- Read storage table data using readable syntax. See [CloudMachine/DatabaseTests.cs at main · KrzysztofCwalina/CloudMachine (github.com)](https://github.com/KrzysztofCwalina/CloudMachine/blob/main/Tests/DatabaseTests.cs#L24)

##### Samples
- [Example motivating scenario](https://gist.github.com/tg-msft/9831ac8e943fa041d2ee952b24cbe015) per @tg-msft 
- [Same scenario using JsonNode](https://gist.github.com/JoshLove-msft/34897c9f5d3250e8b23167ce5c01c2b6) per @JoshLove-msft
- https://github.com/Azure/azure-sdk-for-net/pull/31769
    - Description includes initial requirements
    - Provides illustration of use in CLU and Confidential Ledger samples
- https://github.com/Azure/azure-sdk-for-net/pull/32501
    - Explores making properties settable for round-tripping

#### Notes
In the past, we could not finalize the prototype of JsonData in Azure.Core.Experimental because we tried to make it read-write.  If we scope the problem to just parsing responses, we think we can provide it as a useful type for customers of DPG clients.

#### Creation

- [x] Can get dynamic from Response, e.g. response.Content.ToDynamic();
- [x] Internal API allows JsonData to be constructed from BinaryData
- [ ] Internal API allows JsonData to be constructed from object (e.g. from anonymous type)

#### Serialization/Deserialization

[...]

PR #31203, opened 2022-09-16
Make JsonData Immutable

annelo-msft:
Modifications to JsonData by @jsquire.

Addresses https://github.com/Azure/azure-sdk-for-net/issues/30717


sdk/core/Azure.Core.Experimental/src/JsonData.cs
```diff
@@ -36,26 +36,69 @@ public class JsonData : IDynamicMetaObjectProvider, IEquatable<JsonData>
         private static readonly JsonSerializerOptions DefaultJsonSerializerOptions = new JsonSerializerOptions();
 
         /// <summary>
-        ///  Creates a new JsonData object which represents an JSON object with no properties.
+        ///  Creates a new JsonData object which represents the value of the given JsonDocument.
         /// </summary>
-        public JsonData() : this(Array.Empty<KeyValuePair<string, JsonData>>())
```
annelo-msft, line 41: @jsquire, what was the thinking behind removing the parameterless constructor?

sdk/core/Azure.Core.Experimental/tests/JsonDataTests.cs
```diff
@@ -23,14 +23,15 @@ public void DefaultConstructorMakesEmptyObject()
...
         public void CanCreateFromJson()
         {
-            var jsonData = JsonData.FromString("\"string\"");
+            var jsonData = new JsonData("\"string\"");
 
             Assert.AreEqual("\"string\"", jsonData.ToJsonString());
         }
 
         [Test]
         public void CanCreateFromNull()
```
annelo-msft, line 32: @ellismg, what scenario was driving this requirement?

sdk/core/Azure.Core.Experimental/src/JsonData.cs
jsquire: The type is intended to be immutable and enable consumption scenarios only.   There's no value in having a completely empty instance that you're unable to update... or is there?

sdk/core/Azure.Core.Experimental/tests/JsonDataTests.cs
jsquire: Can you clarify what scenario?   I'm not sure if we're discussing creating an instance form a primitive string or from JSON in string form.

annelo-msft: The scenario of creating a JsonData instance from null -- it creates an ambiguous call site.

sdk/core/Azure.Core.Experimental/src/JsonData.cs
annelo-msft: I see.  There is a test case for it -- `DefaultConstructorMakesEmptyObject` -- but this was likely miscategorized and should have been included with the mutable tests.

sdk/core/Azure.Core.Experimental/tests/JsonDataTests.cs
jsquire: I believe that was related to the mutable starting point.  There's no benefit to doing so that I can see if our goal is to enable consuming a blob of JSON data.

annelo-msft: Would it be possible to get a null response.Content in a service response?

jsquire: Maybe?   I'm not sure what we return for content in the case of something like a 204.

sdk/core/Azure.Core.Experimental/src/JsonData.cs
jsquire: There were a few that I think I needed to kill in the main tests file because they relied on mutability.

sdk/core/Azure.Core.Experimental/tests/JsonDataTests.cs
annelo-msft: Yeah, and I guess so long as the null is "typed", it wouldn't result in an ambiguous call site.

ellismg: > @ellismg, what scenario was driving this requirement?

I do not recall a specific scenario driving this requirement (e.g. some service that had some response of this shape.

I think the goal here was we were trying to provide a mapping between CLR and JSON types and "null" is a type in both domains.

(from node)

```
> JSON.parse('null');
null
> JSON.stringify(null)
'null'
>
```

In the same way we have `new JsonData(5)` and `new JsonData("foo")`, I think we wanted to make sure that `new JsonData(null)` (which was going to be valid code since we already had a constructor that took a reference type) did what you "might expect" which is get a JSON null.

I suspect you could change the behavior and the world wouldn't fall down (I feel like if there was a really subtle reason we had to do this I would have written a load bearing comment somewhere or would have at least remembered the rationale better than I do now)

sdk/core/Azure.Core.Experimental/src/JsonData.cs
```diff
@@ -260,315 +269,17 @@ public string ToJsonString()
...
-        {
-            JsonData value = EmptyArray();
-            EnsureArray().Add(value);
-            return value;
+            using var memoryStream = new MemoryStream();
+            using (var writer = new Utf8JsonWriter(memoryStream))
+            {
+                WriteTo(writer);
+            }
+            return new BinaryData(memoryStream.ToArray());
```
danielmarbach, line 282: How about?

```suggestion
            return new BinaryData(memoryStream.GetBuffer().AsMemory(0, (int)stream.Position));
```

JoshLove-msft: This is how the BinaryData FromStream method constructs the instance after copying into a separate MemoryStream - https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/System.Memory.Data/src/BinaryData.cs#L179

I think the reason we didn't type sniff for MemoryStream and attempt to extract the ArraySegment is that we wanted the semantics to be consistent across different streams.

annelo-msft (commented):
@jsquire, do you know why we implement IDynamicMetaObjectProvider instead of inheriting from DynamicObject?

jsquire:
> @jsquire, do you know why we implement IDynamicMetaObjectProvider instead of inheriting from DynamicObject?

I do not.  @ellismg would probably have insight.

ellismg:
> I do not. @ellismg would probably have insight.

I can't recall the exact reason, but if I had to hazard a guess, it would be that we didn't want to inherit all of the public methods from `DynamicObject` on to our public surface area and into intellisense.  Suspect that's also why we decided to explicitly implement the interface.

annelo-msft:
> I can't recall the exact reason, but if I had to hazard a guess, it would be that we didn't want to inherit all of the public methods from `DynamicObject` on to our public surface area and into intellisense. Suspect that's also why we decided to explicitly implement the interface.

We're investigating making the JsonData API fully internal, so I revisited this question.  

It looks like all the `TryXx` methods on DynamicObject only get called after an attempt to access the dynamic property/conversion/method etc on the underlying typed instance fails and throws an exception, which would degrade performance.  So, I think we will still want to implement IDynamicMetaObjectProvider even if the API is internal.  

There's also the following comment in the DynamicObject reference docs, supporting this hypothesis:
> if you need to manage DLR fast dynamic dispatch caching, create your own implementation of the [IDynamicMetaObjectProvider](https://learn.microsoft.com/en-us/dotnet/api/system.dynamic.idynamicmetaobjectprovider?view=net-7.0) interface.
>
> See: https://learn.microsoft.com/en-us/dotnet/api/system.dynamic.dynamicobject?view=net-7.0#remarks

annelo-msft:
Closing in favor of https://github.com/Azure/azure-sdk-for-net/pull/31769

Issue #25275, opened 2021-11-11
[BUG] ServiceBus MessageLock Exception on CompleteAsync


IvanG96:
**Describe the bug**
I'm using the Azure.Messaging.ServiceBus.dll version 7.4.0, ServiceBusReceiver.
It is used to get message out of the queue using peek\lock mode and then process it (sometimes processing takes up to 1 hour).
Lock time is 5 min, and I have a timer that calls token renew every 4 minutes.
The issue is that for the whole time token renew works without any exceptions from library side, but failed on CompleteAsync:
_The lock supplied is invalid. Either the lock expired, or the message has already been removed from the queue, or was received by a different receiver instance. (MessageLockLost)_
 
According to my logging. Application renewed token for another 5 min, and then failed to complete after 1 min,
_11/11/2021, 9:47:38.650 AM | Message lock on '18a3d6f3-0173-4e02-a457-9b118874473b' is renewed. Locked until: '11/11/2021 09:52:38 +00:00'_

_11/11/2021, 9:48:38.681 AM | Azure.Messaging.ServiceBus.ServiceBusException at Azure.Messaging.ServiceBus.Amqp.AmqpReceiver.ThrowLockLostException_

Points of interest:
1) it happens randomly
2) after exception message is still in the queue for the remaining 4 minutes and can not be received by any other receiver
3) I have one receiver per pode/application
4) I have 8 parallel running pods
5) each reveiver consumes only 1 message at a time

**Stak trace (from app insights)**
{"method":"Azure.Messaging.ServiceBus.Amqp.AmqpReceiver.ThrowLockLostException","assembly":"Azure.Messaging.ServiceBus, Version=7.4.0.0, Culture=neutral, PublicKeyToken=92742159e12e44c8","level":0,"line":0}
{"method":"Azure.Messaging.ServiceBus.Amqp.AmqpReceiver+<DisposeMessageAsync>d__42.MoveNext","assembly":"Azure.Messaging.ServiceBus, Version=7.4.0.0, Culture=neutral, PublicKeyToken=92742159e12e44c8","level":1,"line":0}
{"method":"Azure.Messaging.ServiceBus.Amqp.AmqpReceiver+<CompleteInternalAsync>d__41.MoveNext","assembly":"Azure.Messaging.ServiceBus, Version=7.4.0.0, Culture=neutral, PublicKeyToken=92742159e12e44c8","level":5,"line":0}
[...]

PR #31206, opened 2022-09-16
[ServiceBus] ServicebusRetryPolicy remove locking

danielmarbach:
While trying to figure out what is going on with https://github.com/Azure/azure-sdk-for-net/issues/25275 I observed that the retry policy uses `volatile` and locking to achieve some sort of "consistent" state in the server busy case. I was wondering whether we are running into lock reentrancy problems and potential deadlocks there in edge cases. I tried to get rid of the lock as an attempt and figured I send in the PR to see what you think. 

In the best case, we have something valuable to proceed. In the worst case, I make a complete fool out of myself :smirk: but at least I tried :)

### DIFF
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/ServiceBusRetryPolicy.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/ServiceBusRetryPolicy.cs
@@ -30,7 +30,7 @@ namespace Azure.Messaging.ServiceBus
         /// Represents a state flag that is used to make sure the server busy value can be observed with
         /// reasonable fresh values without having to acquire a lock.
         /// </summary>
-        private volatile int serverBusyState;
+        private volatile int _serverBusyState;
 
         private const int ServerNotBusyState = 0; // default value of serverBusy
         private const int ServerBusyState = 1;
@@ -38,7 +38,7 @@ namespace Azure.Messaging.ServiceBus
         /// <summary>
         /// Determines whether or not the server returned a busy error.
         /// </summary>
-        internal bool IsServerBusy => serverBusyState == ServerBusyState;
+        internal bool IsServerBusy => _serverBusyState == ServerBusyState;
 
         /// <summary>
         /// Gets the exception message when a server busy error is returned.
@@ -195,26 +195,26 @@ namespace Azure.Messaging.ServiceBus
         private void SetServerBusy(string exceptionMessage, CancellationToken cancellationToken)
         {
             // multiple call to this method will not prolong the timer.
-            if (serverBusyState == ServerBusyState)
+            if (_serverBusyState == ServerBusyState)
             {
                 return;
             }
 
             ServerBusyExceptionMessage = string.IsNullOrWhiteSpace(exceptionMessage) ?
                 Resources.DefaultServerBusyException : exceptionMessage;
-            Interlocked.Exchange(ref serverBusyState, ServerBusyState);
+            Interlocked.Exchange(ref _serverBusyState, ServerBusyState);
             _ = ScheduleResetServerBusy(cancellationToken);
         }
 
         private void ResetServerBusy()
         {
-            if (serverBusyState == ServerNotBusyState)
+            if (_serverBusyState == ServerNotBusyState)
             {
                 return;
             }
 
             ServerBusyExceptionMessage = Resources.DefaultServerBusyException;
-            Interlocked.Exchange(ref serverBusyState, ServerNotBusyState);
+            Interlocked.Exchange(ref _serverBusyState, ServerNotBusyState);
         }
 
         private async Task ScheduleResetServerBusy(CancellationToken cancellationToken)
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/ServiceBusRetryPolicy.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/ServiceBusRetryPolicy.cs
@@ -148,7 +148,6 @@ namespace Azure.Messaging.ServiceBus
                 {
                     return await operation(t1, tryTimeout, cancellationToken).ConfigureAwait(false);
                 }
-
                 catch (Exception ex)
                 {
                     Exception activeEx = AmqpExceptionHelper.TranslateException(ex);
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/ServiceBusRetryPolicy.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/ServiceBusRetryPolicy.cs
@@ -26,8 +26,6 @@ namespace Azure.Messaging.ServiceBus
     ///
     public abstract class ServiceBusRetryPolicy
     {
-        private static readonly TimeSpan ServerBusyBaseSleepTime = TimeSpan.FromSeconds(10);
-
         /// <summary>
         /// Represents a state flag that is used to make sure the server busy value can be observed with
         /// reasonable fresh values without having to acquire a lock.
@@ -40,12 +38,18 @@ namespace Azure.Messaging.ServiceBus
         /// <summary>
         /// Determines whether or not the server returned a busy error.
         /// </summary>
-        private bool IsServerBusy => serverBusyState == ServerBusyState;
+        internal bool IsServerBusy => serverBusyState == ServerBusyState;
 
         /// <summary>
         /// Gets the exception message when a server busy error is returned.
         /// </summary>
-        private string ServerBusyExceptionMessage { get; set; }
+        internal string ServerBusyExceptionMessage { get; set; }
+
+        /// <summary>
+        /// Gets the server busy base sleep time
+        /// </summary>
+        /// <remarks>Defaults to TimeSpan.FromSeconds(10)</remarks>
+        internal TimeSpan ServerBusyBaseSleepTime { get; set; } = TimeSpan.FromSeconds(10);
 
         /// <summary>
         ///   The instance of <see cref="ServiceBusEventSource" /> which can be mocked for testing.
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/ServiceBusRetryPolicy.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/ServiceBusRetryPolicy.cs
@@ -186,7 +186,7 @@ namespace Azure.Messaging.ServiceBus
             throw new TaskCanceledException();
         }
 
-        internal void SetServerBusy(string exceptionMessage, CancellationToken cancellationToken)
+        private void SetServerBusy(string exceptionMessage, CancellationToken cancellationToken)
         {
             // multiple call to this method will not prolong the timer.
             if (Interlocked.CompareExchange(ref serverBusyLock, 1, 0) != 0)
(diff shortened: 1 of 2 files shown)


JoshLove-msft:
> In the best case, we have something valuable to proceed. In the worst case, I make a complete fool out of myself 😏 but at least I tried :)

You certainly wouldn't make a fool out of yourself 😄 We are grateful for all of your contributions! I'm fairly sure that the SetServerBusy stuff is not the cause of the issue because I actually added this in an attempt to fix https://github.com/Azure/azure-sdk-for-net/issues/25275)](https://github.com/Azure/azure-sdk-for-net/issues/25275. So the issue was happening even without that code. Instead, this appears to be a server bug where the server is releasing the lock on all messages that have been delivered when throttling occurs. We were able to pinpoint the issue in the server https://github.com/Azure/azure-sdk-for-net/issues/25275#issuecomment-988325752, so at this point we are waiting for someone from the service team to confirm the issue and move forward with a fix. I will take a look at this PR as the changes here may still be worth making, though I don't expect them to fix the issue.

jsquire:
>> In the best case, we have something valuable to proceed. In the worst case, I make a complete fool out of myself 😏 but at least I tried :)

> You certainly wouldn't make a fool out of yourself 😄 We are grateful for all of your contributions!

+1 to this!

sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/ServiceBusRetryPolicy.cs
```diff
@@ -185,49 +186,44 @@ public abstract class ServiceBusRetryPolicy
...
+            IsServerBusy = true;
+            Volatile.Write(ref serverBusyLock, 0);  // release
+            _ = ScheduleResetServerBusy(cancellationToken);
         }
 
-        internal void ResetServerBusy()
+        private void ResetServerBusy()
         {
-            if (!encounteredServerBusy)
+            if (Interlocked.CompareExchange(ref serverBusyLock, 1, 0) != 0)
```
jsquire, line 206: I wonder if `Volatile.Read` would suffice here, as we don't want to actually set the value and it still ensures a memory barrier.   Functionally, I believe they're equivalent, but the `Read` more clearly signals the intent.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/ServiceBusRetryPolicy.cs
```diff
@@ -28,10 +28,11 @@ public abstract class ServiceBusRetryPolicy
...
-        // This is a volatile copy of IsServerBusy. IsServerBusy is synchronized with a lock, whereas encounteredServerBusy is kept volatile for performance reasons.
-        private volatile bool encounteredServerBusy;
+        /// <summary>
+        /// Represents a state flag that is used to make sure the server busy value can be observed with
+        /// reasonable fresh values without having to acquire a lock.
+        /// </summary>
+        private int serverBusyLock;
 
         /// <summary>
         /// Determines whether or not the server returned a busy error.
```
jsquire, line 38: Should `IsServerBusy` just do a `VolatileRead` of the flag variable rather than being managed as a separate state flag?   If I'm not mistaken, the read on L139 in the loop may potentially see a stale value as written.   _(I don't know that it is really impactful, but...)_

sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/ServiceBusRetryPolicy.cs
```diff
@@ -28,10 +28,11 @@ public abstract class ServiceBusRetryPolicy
     {
```
jsquire, line 28: I don't know that we have test coverage for this functionality at all.   Should we consider adding some in the [BasicRetryPolicy tests](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/servicebus/Azure.Messaging.ServiceBus/tests/Core/BasicRetryPolicyTests.cs) or roll a test class dedicated for the base functionality?

jsquire (commented):
Even if it doesn't fix the deadlock issue... this is still a nice upgrade.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/ServiceBusRetryPolicy.cs
danielmarbach: The idea was here to set the value to make sure we are not setting it to non-busy if simultaneously another thread sets it to busy.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/ServiceBusRetryPolicy.cs
danielmarbach: Right now this only represents the state whether we entered the "critical" region and not the state of server busy. I can see if I can tweak things a bit

danielmarbach: Did you mean something like the following?

```csharp
        private bool IsServerBusy => Volatile.Read(ref serverBusy) == 1;

        private void SetServerBusy(string exceptionMessage, CancellationToken cancellationToken)
        {
            // multiple call to this method will not prolong the timer.
            if (Volatile.Read(ref serverBusy) == 1)
            {
                return;
            }

            ServerBusyExceptionMessage = string.IsNullOrWhiteSpace(exceptionMessage) ?
                Resources.DefaultServerBusyException : exceptionMessage;
            Volatile.Write(ref serverBusy, 1);
            _ = ScheduleResetServerBusy(cancellationToken);
        }

        private void ResetServerBusy()
        {
            if (Volatile.Read(ref serverBusy) == 0)
            {
                return;
            }

            ServerBusyExceptionMessage = Resources.DefaultServerBusyException;
            Volatile.Write(ref serverBusy, 0);
        }
``

File: sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/ServiceBusRetryPolicy.cs (hunk 3 above)
Comment (danielmarbach) [reply sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/ServiceBusRetryPolicy.cs:28]: This would probably require to make the `ServerBusyBaseSleepTime`, `IsServerBusy` and `ServerBusyExceptionMessage` and internal property. I can push a proposal.

Comment (danielmarbach) [reply sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/ServiceBusRetryPolicy.cs:28]: Pushed a few tests to get started so that you can see whether this goes into the right direction

File: sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/ServiceBusRetryPolicy.cs (hunk 2 above)
Comment (jsquire) [reply sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/ServiceBusRetryPolicy.cs:39]: That's kind of what I was thinking, yup.

Comment (danielmarbach) [reply sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/ServiceBusRetryPolicy.cs:39]: Pushed a simplified version on the weekend that was inspired by how the `CancellationTokenSource` handles it states in a thread safe and lock free manner with the [underlying state field](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Threading/CancellationTokenSource.cs#L38).

File: sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/ServiceBusRetryPolicy.cs [hunk 4]
```diff
@@ -26,22 +26,30 @@ namespace Azure.Messaging.ServiceBus
...
     public abstract class ServiceBusRetryPolicy
     {
-        private static readonly TimeSpan ServerBusyBaseSleepTime = TimeSpan.FromSeconds(10);
-
-        private readonly object serverBusyLock = new object();
+        /// <summary>
+        /// Represents a state flag that is used to make sure the server busy value can be observed with
+        /// reasonable fresh values without having to acquire a lock.
+        /// </summary>
+        private volatile int serverBusyState;
```
jsquire, line 33:
```suggestion
        private volatile int _serverBusyState;
```

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Core/ServiceBusRetryPolicyTests.cs
```diff
@@ -0,0 +1,90 @@
...
+    [TestFixture]
+    public class ServiceBusRetryPolicyTests
+    {
+        [Test]
+        public void HasDefaultValues()
+        {
+            var policy = new MockServiceBusRetryPolicy();
+
+            Assert.That(policy.IsServerBusy, Is.False);
+            Assert.That(policy.ServerBusyBaseSleepTime, Is.EqualTo(TimeSpan.FromSeconds(10)));
```
jsquire, line 25: I wonder if this may be too specific.   Should we consider just a check for > 0?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/ServiceBusRetryPolicy.cs
jsquire: Approach looks good to me.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/ServiceBusRetryPolicy.cs
danielmarbach: Yeah the underscores. Sorry for that. Will fix

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Core/ServiceBusRetryPolicyTests.cs
danielmarbach: Will change

sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/ServiceBusRetryPolicy.cs
```diff
@@ -185,49 +192,42 @@ public abstract class ServiceBusRetryPolicy
...
-            lock (serverBusyLock)
-            {
-                if (encounteredServerBusy)
-                {
-                    encounteredServerBusy = false;
-                    ServerBusyExceptionMessage = Resources.DefaultServerBusyException;
-                    IsServerBusy = false;
-                }
-            }
+            ServerBusyExceptionMessage = Resources.DefaultServerBusyException;
```
JoshLove-msft, line 216: Aside from doing away with the lock, I think the code is a lot clearer now. Thanks @danielmarbach!

JoshLove-msft (approved):
LGTM pending remaining nits!

sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/ServiceBusRetryPolicy.cs
danielmarbach: Addressed the nits. I shied away from adding more tests that cover additional already previously untested branches of the policy for now, since I feel I have done enough "leave the camping ground cleaner than it was before :wastebasket: " for now. I hope you don't mind me being a bit lazy for this time ;)

PR #31214, opened 2022-09-18
[ServiceBus] Optimize body eager copying

danielmarbach:
I was following up on a new idea to save allocations when receiving messages and had a fresh look at the eager copying code for the `ServiceBusReceivedMessage` body management and this is the result of it

image

This benchmark used is

https://github.com/danielmarbach/MicroBenchmarks/commit/93d216e6d1489bfa5fde82fde8be6d317eedf3f4

The same tricks can be applied for EventHubs :tada: which I'm happy to do once the discussions and nits have settled here ;)

### DIFF
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/MessageBody.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/MessageBody.cs
@@ -80,6 +80,11 @@ namespace Azure.Messaging.ServiceBus.Amqp
             private IList<ReadOnlyMemory<byte>> _segments;
             private IEnumerable<ReadOnlyMemory<byte>> _lazySegments;
 
+            internal CopyingOnConversionMessageBody(IEnumerable<ReadOnlyMemory<byte>> dataSegments)
+            {
+                _lazySegments = dataSegments;
+            }
+
             protected override ReadOnlyMemory<byte> WrittenMemory
             {
                 get
@@ -98,9 +103,9 @@ namespace Azure.Messaging.ServiceBus.Amqp
                 }
             }
 
-            internal CopyingOnConversionMessageBody(IEnumerable<ReadOnlyMemory<byte>> dataSegments)
+            public override IEnumerator<ReadOnlyMemory<byte>> GetEnumerator()
             {
-                _lazySegments = dataSegments;
+                return _segments?.GetEnumerator() ?? _lazySegments.GetEnumerator();
             }
 
             private void Append(ReadOnlyMemory<byte> segment)
@@ -112,11 +117,6 @@ namespace Azure.Messaging.ServiceBus.Amqp
                 segment.CopyTo(memory);
                 _writer.Advance(segment.Length);
                 _segments.Add(memory.Slice(0, segment.Length));
-            }
-
-            public override IEnumerator<ReadOnlyMemory<byte>> GetEnumerator()
-            {
-                return _segments?.GetEnumerator() ?? _lazySegments.GetEnumerator();
             }
         }
 
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/MessageBody.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/MessageBody.cs
@@ -141,7 +141,7 @@ namespace Azure.Messaging.ServiceBus.Amqp
                     segments ??= dataSegments is IReadOnlyCollection<Data> readOnlyList
                         ? new List<ReadOnlyMemory<byte>>(readOnlyList.Count)
                         : new List<ReadOnlyMemory<byte>>();
-                    ReadOnlyMemory<byte> dataToAppend =  segment switch
+                    ReadOnlyMemory<byte> dataToAppend = segment switch
                     {
                         ReadOnlyMemory<byte> romSegment => romSegment,
                         Data data => data.Value switch
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/MessageBody.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/MessageBody.cs
@@ -2,7 +2,6 @@
 // Licensed under the MIT License.
 
 using System;
-using System.Buffers;
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/MessageBody.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/MessageBody.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Buffers;
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
@@ -130,10 +131,7 @@ namespace Azure.Messaging.ServiceBus.Amqp
 
             internal EagerCopyingMessageBody(IEnumerable<Data> dataSegments)
             {
-                foreach (var segment in dataSegments)
-                {
-                    Append(segment);
-                }
+                Append(dataSegments);
             }
 
             protected override ReadOnlyMemory<byte> WrittenMemory => _writer?.WrittenMemory ?? ReadOnlyMemory<byte>.Empty;
@@ -143,23 +141,44 @@ namespace Azure.Messaging.ServiceBus.Amqp
                 return _segments.GetEnumerator();
             }
 
-            private void Append(Data segment)
+            private void Append(IEnumerable<Data> dataSegments)
             {
-                // fields are lazy initialized to not occupy unnecessary memory when there are no data segments
-                _writer ??= new ArrayBufferWriter<byte>();
-                _segments ??= new List<ReadOnlyMemory<byte>>();
-
-                ReadOnlyMemory<byte> dataToAppend = segment.Value switch
+                int length = 0;
+                int numberOfSegments = 0;
+                List<ReadOnlyMemory<byte>> segments = null;
+                foreach (var segment in dataSegments)
                 {
-                    byte[] byteArray => byteArray,
-                    ArraySegment<byte> arraySegment => arraySegment,
-                    _ => ReadOnlyMemory<byte>.Empty
-                };
+                    segments ??= dataSegments is IReadOnlyCollection<Data> readOnlyList
+                        ? new List<ReadOnlyMemory<byte>>(readOnlyList.Count)
+                        : new List<ReadOnlyMemory<byte>>();
+                    ReadOnlyMemory<byte> dataToAppend = segment.Value switch
+                    {
+                        byte[] byteArray => byteArray,
+                        ArraySegment<byte> arraySegment => arraySegment,
+                        _ => ReadOnlyMemory<byte>.Empty
+                    };
+                    length += dataToAppend.Length;
+                    numberOfSegments++;
+                    segments.Add(dataToAppend);
+                }
 
-                var memory = _writer.GetMemory(dataToAppend.Length);
-                dataToAppend.CopyTo(memory);
-                _writer.Advance(dataToAppend.Length);
-                _segments.Add(memory.Slice(0, dataToAppend.Length));
+                if (segments == null)
+                {
+                    return;
+                }
+
+                // fields are lazy initialized to not occupy unnecessary memory when there are no data segments
+                _writer = length > 0 ? new ArrayBufferWriter<byte>(length) : new ArrayBufferWriter<byte>();
+                _segments = segments;
+
+                for (var i = 0; i < numberOfSegments; i++)
+                {
+                    var dataToAppend = segments[i];
+                    var memory = _writer.GetMemory(dataToAppend.Length);
+                    dataToAppend.CopyTo(memory);
+                    _writer.Advance(dataToAppend.Length);
+                    segments[i] = memory.Slice(0, dataToAppend.Length);
+                }
             }
         }
     }
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/MessageBody.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/MessageBody.cs
@@ -90,7 +90,8 @@ namespace Azure.Messaging.ServiceBus.Amqp
                 {
                     if (_lazySegments != null)
                     {
-                        Append(_lazySegments);
+                        // fields are lazy initialized to not occupy unnecessary memory when there are no data segments
+                        Segments.Append(_lazySegments,  ref _writer, ref _segments);
                         _lazySegments = null;
                     }
 
@@ -102,40 +103,6 @@ namespace Azure.Messaging.ServiceBus.Amqp
             {
                 return _segments?.GetEnumerator() ?? _lazySegments.GetEnumerator();
             }
-
-            private void Append(IEnumerable<ReadOnlyMemory<byte>> dataSegments)
-            {
-                int length = 0;
-                int numberOfSegments = 0;
-                List<ReadOnlyMemory<byte>> segments = null;
-                foreach (var segment in dataSegments)
-                {
-                    segments ??= dataSegments is IReadOnlyCollection<ReadOnlyMemory<byte>> readOnlyList
-                        ? new List<ReadOnlyMemory<byte>>(readOnlyList.Count)
-                        : new List<ReadOnlyMemory<byte>>();
-                    length += segment.Length;
-                    numberOfSegments++;
-                    segments.Add(segment);
-                }
-
-                if (segments == null)
-                {
-                    return;
-                }
-
-                // fields are lazy initialized to not occupy unnecessary memory when there are no data segments
-                _writer = length > 0 ? new ArrayBufferWriter<byte>(length) : new ArrayBufferWriter<byte>();
-                _segments = segments;
-
-                for (var i = 0; i < numberOfSegments; i++)
-                {
-                    var dataToAppend = segments[i];
-                    var memory = _writer.GetMemory(dataToAppend.Length);
-                    dataToAppend.CopyTo(memory);
-                    _writer.Advance(dataToAppend.Length);
-                    segments[i] = memory.Slice(0, dataToAppend.Length);
-                }
-            }
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/MessageBody.cs
```diff
@@ -143,23 +159,44 @@ public override IEnumerator<ReadOnlyMemory<byte>> GetEnumerator()
                 return _segments.GetEnumerator();
             }
 
-            private void Append(Data segment)
+            private void Append(IEnumerable<Data> dataSegments)
             {
+                int length = 0;
+                int numberOfSegments = 0;
+                List<ReadOnlyMemory<byte>> segments = null;
```
jsquire, line 166: Just to make sure that I'm understanding - the main changes are here, calling once rather than multiple times to reduce allocations and right-sizing the set of segments?

danielmarbach: The main changes are:

- Allocating the segments list with the right number of items because we mostly know the data segments upfront
- Precalculating the required length upfront
- Instantiating the `ArrayBufferWriter<byte>`  with the known required length to avoid it growing during the calls to `GetMemory` and saving additional allocations because the `CheckAndResizeBuffer` potentially allocates more than we ever need due to [doubling the array size](https://github.com/dotnet/runtime/blob/main/src/libraries/Common/src/System/Buffers/ArrayBufferWriter.cs#L176-L177)

jsquire (approved):
Thanks, @danielmarbach!   This looks good to me.   I'm going to steal this for Event Hubs and I'll defer to @JoshLove-msft for final approval and merge here.

danielmarbach:
@jsquire 

https://github.com/Azure/azure-sdk-for-net/pull/31322

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/MessageBody.cs
```diff
@@ -143,23 +159,44 @@ public override IEnumerator<ReadOnlyMemory<byte>> GetEnumerator()
                 return _segments.GetEnumerator();
             }
 
-            private void Append(Data segment)
+            private void Append(IEnumerable<Data> dataSegments)
```
JoshLove-msft, line 162: The Append method is almost identical between CopyingOnConversionMessageBody and EagerCopyingMessageBody. Can we extract it somehow?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/MessageBody.cs
```diff
@@ -143,23 +159,44 @@ public override IEnumerator<ReadOnlyMemory<byte>> GetEnumerator()
...
+                int numberOfSegments = 0;
+                List<ReadOnlyMemory<byte>> segments = null;
+                foreach (var segment in dataSegments)
+                {
+                    segments ??= dataSegments is IReadOnlyCollection<Data> readOnlyList
+                        ? new List<ReadOnlyMemory<byte>>(readOnlyList.Count)
+                        : new List<ReadOnlyMemory<byte>>();
+                    ReadOnlyMemory<byte> dataToAppend = segment.Value switch
+                    {
+                        byte[] byteArray => byteArray,
```
JoshLove-msft, line 174: Is this only necessary for the EagerCopying because this will already be done by the time CopyingOnConversion is used?

danielmarbach: Yes these are the data segment value cases we have previously optimized for when interacting with the amqp lib for the cases we know what we add to the value property (or at least that is how I remember things)

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/MessageBody.cs
JoshLove-msft: Couldn't the code be exactly the same? The switch case would still work for the CopyOnConversion path, even though it is unnecessary. You could also just pass a flag.

JoshLove-msft: Regarding the coupling, I think that is where having tests for both paths helps, which I think we already have.

danielmarbach: I generally would not abstract things on such a path but even though I have different opinions I don't think I need to argue since I'm happy to adjust to what you as owners prefer.

How about something like this?

```csharp
static class MessageBodyHelper
    {
        public static void AppendSegments<T>(IEnumerable<T> dataSegments, Func<T, ReadOnlyMemory<byte>> converter, ref ArrayBufferWriter<byte> refWriter,
            ref IList<ReadOnlyMemory<byte>> refSegments)
        {
            int length = 0;
            int numberOfSegments = 0;
            List<ReadOnlyMemory<byte>> segments = null;
            foreach (var segment in dataSegments)
            {
                segments ??= dataSegments is IReadOnlyCollection<Data> readOnlyList
                    ? new List<ReadOnlyMemory<byte>>(readOnlyList.Count)
                    : new List<ReadOnlyMemory<byte>>();
                ReadOnlyMemory<byte> dataToAppend = converter(segment);
                length += dataToAppend.Length;
                numberOfSegments++;
                segments.Add(dataToAppend);
            }

            if (segments == null)
            {
                return;
            }

            // fields are lazy initialized to not occupy unnecessary memory when there are no data segments
            refWriter = length > 0 ? new ArrayBufferWriter<byte>(length) : new ArrayBufferWriter<byte>();
            refSegments = segments;

            for (var i = 0; i < numberOfSegments; i++)
            {
                var dataToAppend = segments[i];
                var memory = refWriter.GetMemory(dataToAppend.Length);
                dataToAppend.CopyTo(memory);
                refWriter.Advance(dataToAppend.Length);
                segments[i] = memory.Slice(0, dataToAppend.Length);
            }
        }
```

JoshLove-msft: That works for me!

danielmarbach: The price of the abstraction is non trivial. 

image

JoshLove-msft: Hmm what do you mean by abstraction? Do you mean calling the static helper method?

JoshLove-msft: Could it be due to passing the func? Could we instead just pass a flag or use the same logic for both cases?

danielmarbach: This is how it would look like

```csharp
       private sealed class CopyingOnConversionMessageBody : MessageBody
        {
            private ArrayBufferWriter<byte> _writer;
            private IList<ReadOnlyMemory<byte>> _segments;
            private IEnumerable<ReadOnlyMemory<byte>> _lazySegments;
            private static readonly Func<ReadOnlyMemory<byte>, ReadOnlyMemory<byte>> _mapToSelf = segment => segment;

            internal CopyingOnConversionMessageBody(IEnumerable<ReadOnlyMemory<byte>> dataSegments)
            {
                _lazySegments = dataSegments;
            }

            protected override ReadOnlyMemory<byte> WrittenMemory
            {
                get
                {
                    if (_lazySegments != null)
                    {
                        // fields are lazy initialized to not occupy unnecessary memory when there are no data segments
                        Segments.Append(_lazySegments,  _mapToSelf, ref _writer, ref _segments);
                        _lazySegments = null;
                    }

                    return _writer?.WrittenMemory ?? ReadOnlyMemory<byte>.Empty;
                }
            }

            public override IEnumerator<ReadOnlyMemory<byte>> GetEnumerator()
            {
                return _segments?.GetEnumerator() ?? _lazySegments.GetEnumerator();
            }
        }

        private sealed class EagerCopyingMessageBody : MessageBody
        {
            private readonly ArrayBufferWriter<byte> _writer;
            private readonly IList<ReadOnlyMemory<byte>> _segments;

            internal EagerCopyingMessageBody(IEnumerable<Data> dataSegments)
            {
                // fields are lazy initialized to not occupy unnecessary memory when there are no data segments
                Segments.Append(dataSegments,  static segment =>
                {
                    return segment.Value switch
                    {
                        byte[] byteArray => byteArray,
                        ArraySegment<byte> arraySegment => arraySegment,
                        _ => ReadOnlyMemory<byte>.Empty
                    };
                }, ref _writer, ref _segments);
            }

            protected override ReadOnlyMemory<byte> WrittenMemory => _writer?.WrittenMemory ?? ReadOnlyMemory<byte>.Empty;

            public override IEnumerator<ReadOnlyMemory<byte>> GetEnumerator()
            {
                return _segments.GetEnumerator();
            }
        }

        private static class Segments
        {
            public static void Append<T>(IEnumerable<T> dataSegments, Func<T, ReadOnlyMemory<byte>> converter, ref ArrayBufferWriter<byte> refWriter,
                ref IList<ReadOnlyMemory<byte>> refSegments)
            {
                int length = 0;
                int numberOfSegments = 0;
                List<ReadOnlyMemory<byte>> segments = null;
                foreach (var segment in dataSegments)
                {
                    segments ??= dataSegments is IReadOnlyCollection<Data> readOnlyList
                        ? new List<ReadOnlyMemory<byte>>(readOnlyList.Count)
                        : new List<ReadOnlyMemory<byte>>();
                    ReadOnlyMemory<byte> dataToAppend = converter(segment);
                    length += dataToAppend.Length;
                    numberOfSegments++;
                    segments.Add(dataToAppend);
                }

                if (segments == null)
                {
                    return;
                }

                refWriter = length > 0 ? new ArrayBufferWriter<byte>(length) : new ArrayBufferWriter<byte>();
                refSegments = segments;

                for (var i = 0; i < numberOfSegments; i++)
                {
                    var dataToAppend = segments[i];
                    var memory = refWriter.GetMemory(dataToAppend.Length);
                    dataToAppend.CopyTo(memory);
                    refWriter.Advance(dataToAppend.Length);
                    segments[i] = memory.Slice(0, dataToAppend.Length);
                }
            }
        }
```

danielmarbach: Another run of the current version

image

another run with the abstracted version

image

I could try to run a very long run with a higher iteration count to see if I get less statistical deviation but these runs take ages :( Anyway I need to finish for today.

danielmarbach: @JoshLove-msft yes the helper method including the infrastructure it needs. That is what I mean as an abstraction.

> Could it be due to passing the func? Could we instead just pass a flag or use the same logic for both cases?

How would that work? One accepts `Data` segments and one `ReadOnlyMemory<byte>` there is no common ground to switch case over. I would need to do nested switches then 

```csharp
        private static class Segments
        {
            public static void Append<T>(IEnumerable<T> dataSegments, ref ArrayBufferWriter<byte> refWriter,
                ref IList<ReadOnlyMemory<byte>> refSegments)
            {
                int length = 0;
                int numberOfSegments = 0;
                List<ReadOnlyMemory<byte>> segments = null;
                foreach (var segment in dataSegments)
                {
                    segments ??= dataSegments is IReadOnlyCollection<Data> readOnlyList
                        ? new List<ReadOnlyMemory<byte>>(readOnlyList.Count)
                        : new List<ReadOnlyMemory<byte>>();
                    ReadOnlyMemory<byte> dataToAppend =  segment switch
                    {
                        ReadOnlyMemory<byte> romSegment => romSegment,
                        Data data => data.Value switch
                        {
                            byte[] byteArray => byteArray,
                            ArraySegment<byte> arraySegment => arraySegment,
                            _ => ReadOnlyMemory<byte>.Empty
                        },
                        _ => throw new ArgumentOutOfRangeException()
                    };
                    length += dataToAppend.Length;
                    numberOfSegments++;
                    segments.Add(dataToAppend);
                }

                if (segments == null)
                {
                    return;
                }

                refWriter = length > 0 ? new ArrayBufferWriter<byte>(length) : new ArrayBufferWriter<byte>();
                refSegments = segments;

                for (var i = 0; i < numberOfSegments; i++)
                {
                    var dataToAppend = segments[i];
                    var memory = refWriter.GetMemory(dataToAppend.Length);
                    dataToAppend.CopyTo(memory);
                    refWriter.Advance(dataToAppend.Length);
                    segments[i] = memory.Slice(0, dataToAppend.Length);
                }
            }
        }
```

danielmarbach: Pushed the new version. In case you decide you don't like it we can revert that commit

JoshLove-msft: I'm fine with the nested switch. @jsquire, what do you think?

JoshLove-msft (approved):

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/MessageBody.cs
jsquire: I'm good with either approach and defer to your judgement, @JoshLove-msft.

For my two cents, I don't mind a bit of code duplication when it is focused and serves a purpose.  In this case I prefer it; I found the original version to be simpler, clearer, and have better locality of information.  In particular, the reliance on mutation and side-effects in `Segments::Append` is a pattern that I try hard to avoid as I feel that it makes it more difficult to reason about the code when reading.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/MessageBody.cs
```diff
@@ -142,24 +126,58 @@ public override IEnumerator<ReadOnlyMemory<byte>> GetEnumerator()
...
+                    var memory = refWriter.GetMemory(dataToAppend.Length);
+                    dataToAppend.CopyTo(memory);
+                    refWriter.Advance(dataToAppend.Length);
+                    segments[i] = memory.Slice(0, dataToAppend.Length);
+                }
+            }
+
+            private static ReadOnlyMemory<byte> ThrowArgumentOutOfRange(string paramName)
+            {
+                throw new ArgumentOutOfRangeException(paramName, "Provided data segment is not compatible.");
```
jsquire, line 180: Should this text be in the `Constants` that have started to replace embedded resources for consistency?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/MessageBody.cs
danielmarbach: These were precisely my concerns too, that I tried to (poorly) share in https://github.com/Azure/azure-sdk-for-net/pull/31214#discussion_r977000920

JoshLove-msft: > For my two cents, I don't mind a bit of code duplication when it is focused and serves a purpose. In this case I prefer it; I found the original version to be simpler, clearer, and have better locality of information. In particular, the reliance on mutation and side-effects in `Segments::Append` is a pattern that I try hard to avoid as I feel that it makes it more difficult to reason about the code when reading.

I think the complexity of the new approach is partly due to using a static method here. I understand that we are doing this for performance reasons, but if this were implemented as an instance method on the base class, it would be just about as simple as the original duplicated code.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/MessageBody.cs
JoshLove-msft: I don't think it's necessary since it is only used here. This method could be inlined as well, but also not important.

danielmarbach: This is essentially a Throw Helper. Originally I wanted to also add the caller and the doesnotreturn attributes but I realized I would need to add them to the sources here which would surface more changes.

Under the assumption this is a throw helper wouldn't the hole point of it be to not online it but use it to increase the chance of the using method to be inlineable due to reduced IL code size? That's why I did not add aggressive inlining here

JoshLove-msft: Sorry, I meant that this didn't need to be a separate method, not that we should add the inlining attribute to the method. But will defer to your judgement!

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/MessageBody.cs
danielmarbach: I actually thought about adding it to the base class but then realized the method would have no bearing on Message Body base because what we do there is not a general concern for all message bodies and as such it would become and even more leaky abstraction in my view. Then I thought about adding another base class for the segment based message bodies and putting it there but dismissed it due to the complexity in the inheritance chain. That's why I made it static.

JoshLove-msft: You are right - we'd probably end up needing a new class between the MessageBody and the other segment based message classes. I guess whichever approach we landed on was going to have its tradeoffs 😆 
Thanks for thinking through this with me @danielmarbach!

danielmarbach: Well I could have done a better job outlining my thoughts from the beginning 😂

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/MessageBody.cs
danielmarbach: Would you prefer if I made this a full blown throw helper including all the bells and whistles? Example https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/ArgumentNullException.cs#L67-L69

JoshLove-msft: No, I think it is fine as is or you can just throw directly from the calling method. We already have some helpers in our shared source that you could look at if you like - https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/Shared/Argument.cs#L190

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/MessageBody.cs
JoshLove-msft: Hah no you're good!

PR #31218, opened 2022-09-19
Polish Azure.ResourceManager.Healthbot

ArthurMa1978:


sdk/healthbot/Azure.ResourceManager.Healthbot/api/Azure.ResourceManager.Healthbot.netstandard2.0.cs
```diff
@@ -54,6 +54,14 @@ public partial class HealthBotResource : Azure.ResourceManager.ArmResource
 }
 namespace Azure.ResourceManager.Healthbot.Models
 {
```
Yao725, line 56: We are both using the `HealthBot` and `Healthbot` at the same time in this RP. For example, the namespace is Healthbot but we also have a model named `HealthBotKeyVaultProperties`. This may cause inconsistency and isn't friendly for the user. Therefore I'm wondering if we can unify and just use one of them.

Yao725 (approved):

PR #31219, opened 2022-09-19
Polish the Azure.ResourceManager.MarketplaceOrdering

ArthurMa1978:


sdk/marketplaceordering/Azure.ResourceManager.MarketplaceOrdering/src/autorest.md
```diff
@@ -16,6 +16,19 @@ skip-csproj: true
...
+  AgreementTerms.properties.accepted: IsAccepted
+  AgreementTerms.properties.licenseTextLink: -|uri
+  AgreementTerms.properties.privacyPolicyLink: -|uri
+  AgreementTerms.properties.marketplaceTermsLink: -|uri
+  AgreementTerms.properties.retrieveDatetime: RetrievedOn
+  OfferType: AgreementOfferType
+
+request-path-to-resource-name:
+  /subscriptions/{subscriptionId}/providers/Microsoft.MarketplaceOrdering/offerTypes/{offerType}/publishers/{publisherId}/offers/{offerId}/plans/{planId}/agreements/current: MarketplaceAgreementTerm
+  /subscriptions/{subscriptionId}/providers/Microsoft.MarketplaceOrdering/agreements/{publisherId}/offers/{offerId}/plans/{planId}: MarketplaceAgreement
```
Yao725, line 30: Why did we call this resource `MarketplaceAgreement`? It has three segments after the namespace and it seems we only pick the first one `agreement`, therefore I'd like to know why.

Yao725 (approved):

sdk/marketplaceordering/Azure.ResourceManager.MarketplaceOrdering/src/autorest.md
ArthurMa1978: This is from the operationId and description:
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/marketplaceordering/resource-manager/Microsoft.MarketplaceOrdering/stable/2021-01-01/Agreements.json#L249

PR #31222, opened 2022-09-19
[ResourceMover] Fix the tag operations issue

Yao725:


sdk/resourcemover/Azure.ResourceManager.ResourceMover/src/Customized/MoverResourceSetResource.cs
```diff
@@ -0,0 +1,241 @@
...
+{
+    /// <summary>
+    /// A Class representing a MoverResourceSet along with the instance operations that can be performed on it.
+    /// If you have a <see cref="ResourceIdentifier" /> you can construct a <see cref="MoverResourceSetResource" />
+    /// from an instance of <see cref="ArmClient" /> using the GetMoverResourceSetResource method.
+    /// Otherwise you can get one from its parent resource <see cref="ResourceGroupResource" /> using the GetMoverResourceSet method.
+    /// </summary>
+    public partial class MoverResourceSetResource : ArmResource
+    {
+        /// <summary>
```
Yao725, line 29: This file is used to customize the tag related operations in MoverResourceSetResource. We need this because if we use the original generated methods to update or remove the tag, we will get a service error that tells us `Identity cannot be removed`. This behavior like the patch method of this resource very much, therefore I guess that the backend service implemetation is based on the patch method.

ArthurMa1978: please file an issue to trace this tag operation issue

ArthurMa1978 (approved):

sdk/resourcemover/Azure.ResourceManager.ResourceMover/src/Customized/MoverResourceSetResource.cs
Yao725: Created: https://github.com/Azure/azure-sdk-for-net/issues/31228

PR #31226, opened 2022-09-19
update compute api-version

ArcturusZhang:


sdk/compute/Azure.ResourceManager.Compute/src/Customize/VirtualMachineExtensionData.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// </item>
+        /// </list>
+        /// </para>
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public BinaryData ProtectedSettingsFromKeyVault
+        {
+            get
+            {
+                string raw;
```
m-nash, line 53: This doesn't look like the pattern we talked about last night?  Did we find out that it didn't work?

sdk/compute/Azure.ResourceManager.Compute/src/Customize/Models/CapacityReservationInstanceView.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Azure.ResourceManager.Compute.Models
+{
+    public partial class CapacityReservationInstanceView
+    {
+        /// <summary> A list of all virtual machines resource ids allocated against the capacity reservation. </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public IReadOnlyList<SubResource> UtilizationInfoVirtualMachinesAllocated
+        {
+            get => UtilizationInfo?.VirtualMachinesAllocated;
```
m-nash, line 18: This seems problematic since before the list would have never been null and now if UtilizationInfo is null the list will be null.  Since its IReadOnlyList its probably safe to return an empty list here.

sdk/compute/Azure.ResourceManager.Compute/src/Customize/Models/GalleryDiskImage.cs
```diff
@@ -0,0 +1,22 @@
...
+namespace Azure.ResourceManager.Compute.Models
+{
+    public partial class GalleryDiskImage
+    {
+        /// <summary>
+        /// The source for the disk image.
+        /// This is no longer supported. Please use <see cref="GallerySource"/> instead.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public GalleryArtifactVersionSource Source { get; set; } // the old property to keep backward compatiblility, working as a backing property of GallerySource
```
m-nash, line 17: Can we add unit tests around this. It appears that if someone does this it will fail
```c#
var galleryDiskImage = new GalleryDiskImage(...);
galleryDiskImage.Source = new GalleryArtifactVersionSource(...);
var gallerySource = galleryDiskImage.GallerySource; // <-- invalid cast exception?
```

sdk/compute/Azure.ResourceManager.Compute/src/Customize/Models/GalleryImageVersionStorageProfile.cs
```diff
@@ -0,0 +1,22 @@
...
+namespace Azure.ResourceManager.Compute.Models
+{
+    public partial class GalleryImageVersionStorageProfile
+    {
+        /// <summary>
+        /// The gallery artifact version source.
+        /// This is no longer supported. Please use <see cref="GallerySource"/> instead.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public GalleryArtifactVersionSource Source { get; set; } // the old property to keep backward compatiblility, working as a backing property of GallerySource
```
m-nash, line 17: Same comment as `GalleryDiskImage`

sdk/compute/Azure.ResourceManager.Compute/src/Customize/Models/VirtualMachineExtensionPatch.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// </item>
+        /// </list>
+        /// </para>
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public BinaryData ProtectedSettingsFromKeyVault
+        {
+            get
+            {
+                string raw;
```
m-nash, line 52: Same comment as `VirtualMachineExtensionData`

sdk/compute/Azure.ResourceManager.Compute/src/Customize/Models/VirtualMachineScaleSetExtensionPatch.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// <description>Creates a payload of { "key": "value" }.</description>
+        /// </item>
+        /// <item>
+        /// <term>BinaryData.FromString("{\"key\": \"value\"}")</term>
+        /// <description>Creates a payload of { "key": "value" }.</description>
+        /// </item>
+        /// </list>
+        /// </para>
+        /// </summary>
+        public BinaryData ProtectedSettingsFromKeyVault { get; set; }
```
m-nash, line 47: Don't we need to wrapping here to share the memory with the new property?

sdk/compute/Azure.ResourceManager.Compute/src/Customize/Models/VirtualMachineScaleSetVmExtensionPatch.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// </item>
+        /// <item>
+        /// <term>BinaryData.FromString("{\"key\": \"value\"}")</term>
+        /// <description>Creates a payload of { "key": "value" }.</description>
+        /// </item>
+        /// </list>
+        /// </para>
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public BinaryData ProtectedSettingsFromKeyVault { get; set; }
```
m-nash, line 44: Don't we need to wrapping here to share the memory with the new property?

sdk/compute/Azure.ResourceManager.Compute/src/Customize/VirtualMachineScaleSetVmExtensionData.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// </item>
+        /// </list>
+        /// </para>
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public BinaryData ProtectedSettingsFromKeyVault
+        {
+            get
+            {
+                string raw;
```
m-nash, line 53: Same comment as `VirtualMachineExtensionData`

sdk/compute/Azure.ResourceManager.Compute/tests/Unit/VirtualMachineExtensionDataTests.cs
```diff
@@ -0,0 +1,86 @@
...
+
+namespace Azure.ResourceManager.Compute.Tests.Unit
+{
+    public class VirtualMachineExtensionDataTests
+    {
+        private const string secretUri = "https://kvname.vault.azure.net/secrets/secretname/79b88b3a6f5440ffb2e73e44a0db712e";
+        private const string vaultId = "/subscriptions/a53f7094-a16c-47af-abe4-b05c05d0d79a/resourceGroups/myResourceGroup/providers/Microsoft.KeyVault/vaults/kvName";
+
+        [TestCase]
+        public void ValidateSetBinaryDataGetFromModel()
```
m-nash, line 20: We should have these for all BinaryData to Concrete conversions since it looks like a few of them have no shared memory backer.

sdk/compute/Azure.ResourceManager.Compute/src/Customize/VirtualMachineExtensionData.cs
ArcturusZhang: Sorry I forgot we have multiple occurrences of this case. I will fix them

sdk/compute/Azure.ResourceManager.Compute/src/Customize/Models/CapacityReservationInstanceView.cs
ArcturusZhang: Well I believe before this list also could be null
This is the code for this before this update:
```
        /// <summary> Unutilized capacity of the capacity reservation. </summary>
        internal CapacityReservationUtilization UtilizationInfo { get; }
        /// <summary> A list of all virtual machines resource ids allocated against the capacity reservation. </summary>
        public IReadOnlyList<SubResource> UtilizationInfoVirtualMachinesAllocated
        {
            get => UtilizationInfo?.VirtualMachinesAllocated;
        }
```
I just moved it here. `UtilizationInfo` could be null if the response does not contain it, and hence this list will be null if any one calls it

sdk/compute/Azure.ResourceManager.Compute/src/Customize/Models/GalleryDiskImage.cs
ArcturusZhang: I added a test - but we will not get a cast exception, just get back null because we are using `as`

sdk/compute/Azure.ResourceManager.Compute/src/Customize/Models/CapacityReservationInstanceView.cs
ArcturusZhang: This might be something that could updated on the safe flatten feature in the generator

sdk/compute/Azure.ResourceManager.Compute/CHANGELOG.md
```diff
@@ -4,6 +4,12 @@
 
 ### Features Added
 
+Bumps the api-version
```
Yao725, line 7: Are we going to release this update? If so, maybe we should also specify the version and release date.

sdk/compute/Azure.ResourceManager.Compute/api/Azure.ResourceManager.Compute.netstandard2.0.cs
```diff
@@ -448,8 +449,10 @@ public partial class CommunityGalleryImageVersionData : Azure.ResourceManager.Co
     {
         internal CommunityGalleryImageVersionData() { }
         public System.DateTimeOffset? EndOfLifeOn { get { throw null; } }
+        [System.ComponentModel.EditorBrowsableAttribute(System.ComponentModel.EditorBrowsableState.Never)]
         public bool? ExcludeFromLatest { get { throw null; } }
         public Azure.Core.ResourceIdentifier Id { get { throw null; } }
+        public bool? IsExcludedFromLatest { get { throw null; } }
```
Yao725, line 455: Maybe we don't need to rename this property, in which case we wouldn't need to add the customized code？

sdk/compute/Azure.ResourceManager.Compute/api/Azure.ResourceManager.Compute.netstandard2.0.cs
```diff
@@ -1502,8 +1512,10 @@ public partial class SharedGalleryImageVersionData : Azure.ResourceManager.Compu
     {
         internal SharedGalleryImageVersionData() { }
         public System.DateTimeOffset? EndOfLifeOn { get { throw null; } }
+        [System.ComponentModel.EditorBrowsableAttribute(System.ComponentModel.EditorBrowsableState.Never)]
         public bool? ExcludeFromLatest { get { throw null; } }
         public Azure.Core.ResourceIdentifier Id { get { throw null; } }
+        public bool? IsExcludedFromLatest { get { throw null; } }
```
Yao725, line 1518: Same as above.

sdk/compute/Azure.ResourceManager.Compute/api/Azure.ResourceManager.Compute.netstandard2.0.cs
```diff
@@ -3437,19 +3511,33 @@ public partial class GalleryArtifactPublishingProfileBase
     {
         public GalleryArtifactPublishingProfileBase() { }
         public System.DateTimeOffset? EndOfLifeOn { get { throw null; } set { } }
+        [System.ComponentModel.EditorBrowsableAttribute(System.ComponentModel.EditorBrowsableState.Never)]
         public bool? ExcludeFromLatest { get { throw null; } set { } }
+        public bool? IsExcludedFromLatest { get { throw null; } set { } }
```
Yao725, line 3516: Same as above.

sdk/compute/Azure.ResourceManager.Compute/api/Azure.ResourceManager.Compute.netstandard2.0.cs
```diff
@@ -3091,6 +3120,24 @@ public partial class DiskAccessPatch
         public System.Collections.Generic.IDictionary<string, string> Tags { get { throw null; } }
     }
     [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
+    public readonly partial struct DiskControllerType : System.IEquatable<Azure.ResourceManager.Compute.Models.DiskControllerType>
+    {
+        private readonly object _dummy;
+        private readonly int _dummyPrimitive;
+        public DiskControllerType(string value) { throw null; }
+        public static Azure.ResourceManager.Compute.Models.DiskControllerType NVMe { get { throw null; } }
```
Yao725, line 3128: Should we rename this to `Nvme`?

sdk/compute/Azure.ResourceManager.Compute/api/Azure.ResourceManager.Compute.netstandard2.0.cs
```diff
@@ -3405,9 +3453,33 @@ public partial class EncryptionSettingsGroup
...
     }
+    public partial class GalleryApplicationCustomAction
+    {
+        public GalleryApplicationCustomAction(string name, string script) { }
+        public string Description { get { throw null; } set { } }
+        public string Name { get { throw null; } set { } }
+        public System.Collections.Generic.IList<Azure.ResourceManager.Compute.Models.GalleryApplicationCustomActionParameter> Parameters { get { throw null; } }
+        public string Script { get { throw null; } set { } }
+    }
+    public partial class GalleryApplicationCustomActionParameter
```
Yao725, line 3464: I think we'd better not have this model name end with `Parameter`.

Yao725 (approved):

sdk/compute/Azure.ResourceManager.Compute/api/Azure.ResourceManager.Compute.netstandard2.0.cs
ArcturusZhang: well i changed a few of this case to align our current standard of polishing the SDK

sdk/compute/Azure.ResourceManager.Compute/api/Azure.ResourceManager.Compute.netstandard2.0.cs
ArcturusZhang: it is kind of a special word, see here: https://nvmexpress.org/
therefore I prefer we leave it as is

sdk/compute/Azure.ResourceManager.Compute/api/Azure.ResourceManager.Compute.netstandard2.0.cs
ArcturusZhang: This ends with `Parameter` because it stands for a configurable "parameter" - not like the case of request body's parameter.
For instance, there is another model with a property `Parameters`, and the type of the property is a list of this "GalleryApplicationCustomActionParameter". Therefore I believe it is reasonable that this ends with Parameter

sdk/compute/Azure.ResourceManager.Compute/CHANGELOG.md
ArcturusZhang: No release plan yet. I will open another PR to prepare release

PR #31227, opened 2022-09-19
[Identity] Expose MsaPassthrough as an option

isra-fel:
This PR exposes the [`WindowsBrokerOptions.MsaPassthrough`](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/4.47.0/src/client/Microsoft.Identity.Client/ApiConfig/WindowsBrokerOptions.cs#L42) option of MSAL to two credential option types of `Azure.Identity.BrokeredAuthentication` libraries as it's essential to supporting MSA account.


sdk/identity/Azure.Identity.BrokeredAuthentication/src/InteractiveBrowserCredentialBrokerOptions.cs
```diff
@@ -13,21 +13,29 @@ namespace Azure.Identity.BrokeredAuthentication
...
         private IntPtr _parentWindowHandle;
+        private bool _msaPassthrough;
 
         /// <summary>
         /// Creates a new instance of <see cref="InteractiveBrowserCredentialBrokerOptions"/> to configure a <see cref="InteractiveBrowserCredential"/>.
         /// </summary>
         /// <param name="parentWindowHandle">Handle of the parent window the system authentication broker should be docked to.</param>
-        public InteractiveBrowserCredentialBrokerOptions(IntPtr parentWindowHandle) : base()
+        /// <param name="msaPassthrough">A legacy option available only to old Microsoft applications. Should be avoided where possible. Support is experimental.</param>
+        public InteractiveBrowserCredentialBrokerOptions(IntPtr parentWindowHandle, bool msaPassthrough = false) : base()
```
bgavrilMS, line 23: New default param is a subtle breaking change (a binary breaking change). Prefer to use a new ctor.

isra-fel: Thanks for reviewing and good to know that :) `Azure.Identity.BrokeredAuthentication` is a beta so I thought we needn't worry too much about breaking change. Anyway, I'll let @schaabs 's team decide the final look of the API.

bgavrilMS, line 23: Another way of doing this without exposing public API is to turn on this option based on client id and authority. AzurePS client id is configured as MSA-PT. So basically, hardcode some client_ids.

Afaik, it is not possible to configure an app to move away from being MSA-PT. Identity recommends folks to create a new client_id.

sdk/identity/Azure.Identity.BrokeredAuthentication/api/Azure.Identity.BrokeredAuthentication.net462.cs
```diff
@@ -2,11 +2,11 @@ namespace Azure.Identity.BrokeredAuthentication
 {
     public partial class InteractiveBrowserCredentialBrokerOptions : Azure.Identity.InteractiveBrowserCredentialOptions
     {
-        public InteractiveBrowserCredentialBrokerOptions(System.IntPtr parentWindowHandle) { }
+        public InteractiveBrowserCredentialBrokerOptions(System.IntPtr parentWindowHandle, bool msaPassthrough = false) { }
```
scottaddie, line 5: Nit: Let's change this parameter name's casing all throughout for improved readability.
```suggestion
        public InteractiveBrowserCredentialBrokerOptions(System.IntPtr parentWindowHandle, bool msaPassThrough = false) { }
```

Issue #30961, opened 2022-09-06
[BUG] Service Bus trigger using identity-based connection uses incorrect user identity when run locally


ashtmMSFT:
### Library name and version

Microsoft.Azure.WebJobs.Extensions.ServiceBus 5.7.0

### Describe the bug

I had a lot of trouble getting a managed identity to work correctly locally for a C# Azure Function with a Service Bus Queue-triggered binding on my devbox with VS Code. The problem seems to stem from the fact that the user identity being picked up at runtime is incorrect.

**Context -- Local Tooling**
- Visual Studio Code -- `code --version` == 1.70.2
- Visual Studio Code extensions Azure Account and Azure Functions are installed.
- `dotnet --version` == 6.0.400 
- .csproj Packages: Microsoft.NET.Sdk.Functions 4.1.1, Microsoft.Azure.WebJobs.Extensions.ServiceBus 5.7.0

**Context -- Cloud**
- I have two users which belong to separate tenants: user A who is homed to tenant A, and user B who is homed to tenant B.
- Tenant B contains a Service Bus namespace with a Queue.
- User B is assigned appropriate RBAC roles to read and write from the Queue.

**Context -- Local Dev Workflow**
- I am logged into Windows as user A.
- From Terminal, I perform `az account show`, which shows me I am logged in as user A (associated with tenant A).
- From Terminal, I do an `az login` and login as user B.
- From Terminal, I perform `az account show`, which shows me I am now logged in as user B (associated with tenant B).
- From this Terminal, I launch VS Code (`code .`) in the folder with my Azure Functions project containing a Service Bus Queue-triggered function.
- Inside VS code, I login to the Azure Account extension with user B. I confirm this login succeeded: I see user B's email address in the bottom status bar AND see user B's Azure subscriptions/resources in the Azure extension blade.
- In my local.settings.json file, I specify the ServiceBusConnection__fullyQualifiedNamespace setting as the service bus namespace's hostname.
- In my local.settings.json file, I specify the ServiceBusConnection__tenantId setting as the id for tenant B.
- I start debugging the project from VS Code.

**The Error / Actual Behavior**

When I run the code, I see this error spammed in the VS Code terminal:

[...]

PR #31295, opened 2022-09-20
Support passing `additionallyAllowedTenants` via config

JoshLove-msft:


sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientFactory.cs
```diff
@@ -141,13 +161,37 @@ internal static TokenCredential CreateCredential(IConfiguration configuration, T
...
+                }
+
                 store.Close();
 
                 return credential;
             }
 
             // TODO: More logging
+
+            if (additionallyAllowedTenantsList != null)
```
JoshLove-msft, line 186: I'm not sure about this part - it relies on the fact that the default CredentialFactory uses DefaultAzureCredential, but this can actually be changed by the user.
https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/extensions/Microsoft.Extensions.Azure/src/Internal/AzureComponentFactoryImpl.cs#L25

sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientFactory.cs
```diff
@@ -93,6 +94,16 @@ internal static TokenCredential CreateCredential(IConfiguration configuration, T
...
             var certificateStoreLocation = configuration["clientCertificateStoreLocation"];
+            var additionallyAllowedTenants = configuration["additionallyAllowedTenants"];
+            List<string> additionallyAllowedTenantsList = null;
+            if (!string.IsNullOrWhiteSpace(additionallyAllowedTenants))
+            {
+                // not relying on StringSplitOptions.RemoveEmptyEntries as we want to remove leading/trailing whitespace between entries
+                additionallyAllowedTenantsList = additionallyAllowedTenants.Split(TenantDelimiter)
+                    .Select(t => t.Trim())
+                    .Where(t => !string.IsNullOrWhiteSpace(t))
+                    .ToList();
```
christothes, line 105: nit: `ToList()` isn't needed here

sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientFactory.cs
JoshLove-msft: Ideally, we would flow through any config to the credential factory fallback, but we already do not do this for all of the other settings, which has led to some confusion - https://github.com/Azure/azure-sdk-for-net/issues/30961

christothes: How was DefaultAzureCredential utilized before?

JoshLove-msft: If CreateCredential returned null, AzureComponentFactory will fall back to a user configurable factory. By default the factory will construct a DefaultAzureCredential.
https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/extensions/Microsoft.Extensions.Azure/src/Internal/AzureComponentFactoryImpl.cs#L25

christothes: Is it valid to provide both a IConfiguration options and a custom `CredentialFactory`?

JoshLove-msft: There is nothing preventing someone from doing that. It is exposed through the [UseCredential](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.azure.azureclientfactorybuilder.usecredential?view=azure-dotnet) overloads, which you'd have access to after call [AddAzureClients](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.azure.azureclientservicecollectionextensions.addazureclients?view=azure-dotnet). I don't think it is obvious that this is actually just the fallback and it will be ignored if the supplied configuration matches one of the supported credentials from AzureComponentFactory.

christothes: It would be nice if we could log when this occurs (along with the other config decisions above), but barring that, perhaps we should just add some additional doc comments to explain some of the flow for these cases. Otherwise, it seems reasonable to default to DefaultAzureCredential here.

JoshLove-msft: Yes, we have some work to do as part of resolving https://github.com/Azure/azure-sdk-for-net/issues/30961

christothes (approved):

eng/Packages.Data.props
```diff
@@ -93,7 +93,7 @@
     <PackageReference Update="Azure.Messaging.ServiceBus" Version="7.10.0" />
     <PackageReference Update="Azure.Messaging.WebPubSub" Version="1.0.0" />
     <PackageReference Update="Azure.Monitor.Query" Version="1.1.0" />
-    <PackageReference Update="Azure.Identity" Version="1.6.0" />
+    <PackageReference Update="Azure.Identity" Version="1.7.0" />
```
JoshLove-msft, line 96: Note, I am bumping the central dependency to the version containing the breaking change @schaabs

sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientFactory.cs
```diff
@@ -93,6 +94,15 @@ internal static TokenCredential CreateCredential(IConfiguration configuration, T
             var certificate = configuration["clientCertificate"];
             var certificateStoreName = configuration["clientCertificateStoreName"];
             var certificateStoreLocation = configuration["clientCertificateStoreLocation"];
+            var additionallyAllowedTenants = configuration["additionallyAllowedTenants"];
```
JoshLove-msft, line 97: One other thought - should we make this a JSON list configuration rather than delimited string?

christothes: I'm not sure if that makes it any easier for the user. Plus, then we'd also have to handle invalid json.

eng/Packages.Data.props
schaabs: I had a PR out to do this, https://github.com/Azure/azure-sdk-for-net/pull/31284, but it seems the PR failed because the restore failed on one leg. I restarted the leg so it might merge in that PR.

sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientFactory.cs
schaabs: I agree I think the semi-colon delimited string is less error prone. This is the format expected in the environment variable `AZURE_ADDITIONALLY_ALLOWED_TENANTS`, and I think semi-colon delimited is pretty standard for environment variables. But if JSON is more regularly used in functions config, we could consider using it here.

schaabs (approved):

sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientFactory.cs
JoshLove-msft: woops, it is currently comma delimited - should it be semi-colon?

sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientFactory.cs
```diff
@@ -141,13 +160,37 @@ internal static TokenCredential CreateCredential(IConfiguration configuration, T
...
             // TODO: More logging
+
+            if (additionallyAllowedTenantsList != null)
+            {
+                var options = new DefaultAzureCredentialOptions();
+                foreach (string tenant in additionallyAllowedTenantsList)
+                {
+                    options.AdditionallyAllowedTenants.Add(tenant);
+                }
+                return new DefaultAzureCredential(options);
```
JoshLove-msft, line 192: Since we are creating the DefaultAzureCredential, should we also pass the other relevant options that can be configured:
- tenantId
- clientId
- managedIdentityResourceId
Any others?

christothes: If this is a more convenient alternative to providing a full CredentialFactory, I think it's a good change. Although it customers were previously supplying a factory for this scenario, it may break them due to the current behavior of ignoring the factory when an IConfig is present.

sdk/extensions/Microsoft.Extensions.Azure/CHANGELOG.md
```diff
@@ -1,14 +1,10 @@
...
+## 1.6.0 (2022-09-22)
 
 ### Features Added
 
-### Breaking Changes
-
-### Bugs Fixed
-
-### Other Changes
+- Support passing `additionallyAllowedTenants` via config.
```
jsquire, line 7: Should we mention that this could potentially break some stuff and describe how/what?   Maybe link back to the Identity write-up for things?

sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientFactory.cs
```diff
@@ -93,6 +94,15 @@ internal static TokenCredential CreateCredential(IConfiguration configuration, T
...
             var certificateStoreName = configuration["clientCertificateStoreName"];
             var certificateStoreLocation = configuration["clientCertificateStoreLocation"];
+            var additionallyAllowedTenants = configuration["additionallyAllowedTenants"];
+            IEnumerable<string> additionallyAllowedTenantsList = null;
+            if (!string.IsNullOrWhiteSpace(additionallyAllowedTenants))
+            {
+                // not relying on StringSplitOptions.RemoveEmptyEntries as we want to remove leading/trailing whitespace between entries
+                additionallyAllowedTenantsList = additionallyAllowedTenants.Split(TenantDelimiter)
+                    .Select(t => t.Trim())
+                    .Where(t => !string.IsNullOrWhiteSpace(t));
```
jsquire, line 104: Since you've already trimmed, you could just go with `IsNotNullOrEmpty` here.

sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientFactory.cs
```diff
@@ -93,6 +94,15 @@ internal static TokenCredential CreateCredential(IConfiguration configuration, T
             var certificate = configuration["clientCertificate"];
             var certificateStoreName = configuration["clientCertificateStoreName"];
             var certificateStoreLocation = configuration["clientCertificateStoreLocation"];
+            var additionallyAllowedTenants = configuration["additionallyAllowedTenants"];
+            IEnumerable<string> additionallyAllowedTenantsList = null;
+            if (!string.IsNullOrWhiteSpace(additionallyAllowedTenants))
+            {
+                // not relying on StringSplitOptions.RemoveEmptyEntries as we want to remove leading/trailing whitespace between entries
+                additionallyAllowedTenantsList = additionallyAllowedTenants.Split(TenantDelimiter)
+                    .Select(t => t.Trim())
```
jsquire, line 103: I believe this will throw if `t` is `null`.   Should we use `t?.Trim()` instead?

jsquire (approved):

sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientFactory.cs
JoshLove-msft: I don't see how `t` can be `null` here. Do you have a scenario?

sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientFactory.cs
JoshLove-msft: I think we can reasonably treat the existing behavior where we ignore the config if we don't have all the matching config needed for a credential as a bug.

sdk/extensions/Microsoft.Extensions.Azure/CHANGELOG.md
schaabs: Technically it's already broken, but we could just link to the Identity breaking change doc here. https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/identity/Azure.Identity/BREAKING_CHANGES.md#170

sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientFactory.cs
```diff
@@ -141,13 +160,60 @@ internal static TokenCredential CreateCredential(IConfiguration configuration, T
...
+                    var options = new ClientCertificateCredentialOptions();
+                    foreach (string tenant in additionallyAllowedTenantsList)
+                    {
+                        options.AdditionallyAllowedTenants.Add(tenant);
+                    }
+                    credential = new ClientCertificateCredential(tenantId, clientId, certs[0], options);
+                }
+                else
+                {
+                    credential = new ClientCertificateCredential(tenantId, clientId, certs[0]);
```
schaabs, line 175: Just FYI it's totally acceptable to pass `null` for the options parameter. It could clean up a lot of the if..else statements here. This applies to the other credential types as well.

schaabs: You could also pass an options bag that has no options set.

JoshLove-msft: Thanks, I think passing an empty options bag would be best.

sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientFactory.cs
jsquire: No, I don't.   I was thinking of the case where there was a double-delimiter, but that comes back as empty.   In that case, the `where` can be simplified to `Where(t => t != string.Empty)`, I think.

PR #31299, opened 2022-09-20
[ACR] Split Data Plane and Management Plane Pipeline

ShivangiReja:

### DIFF
--- a/sdk/containerregistry/ci.mgmt.yml
+++ b/sdk/containerregistry/ci.mgmt.yml
@@ -7,7 +7,7 @@ pr: none
 extends:
   template: /eng/pipelines/templates/stages/archetype-sdk-client.yml
   parameters:
-    SdkType: mgmtclient
+    SDKType: mgmtclient
     ServiceDirectory: containerregistry
     BuildSnippets: true
     Artifacts:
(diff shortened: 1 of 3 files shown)


ArthurMa1978 (approved):

annelo-msft (approved):

jsquire (approved):

pallavit:
I am curious which other SDKs need this change? Based on the thread with Wes --> can we assume a grep for SDKType: client in ci.mgmt.yml is the right way to go about it?

jsquire:
Event Hubs and Service Bus split out the management libraries a while back in order to reduce consistent test failures unrelated to the client libraries.

weshaggard:
> I am curious which other SDKs need this change? Based on the thread with Wes --> can we assume a grep for SDKType: client in ci.mgmt.yml is the right way to go about it?

That might find some but I suspect not many. Ultimately if we want to take this fully to the extreme we would need to make sure every service directory that has both a ci.yml and ci.mgmt.yml also has a service.project and the correct SDKType's set.

pallavit:
Thank you. I have created the following issue - https://github.com/Azure/azure-sdk-for-net/issues/31323 to track. We do not have to proactively fix all - however any instances where this poorly reflects on pipeline health for our team owned SDKs should be a priority.

PR #31304, opened 2022-09-21
Rename check enforcer github action

benbp:


.github/workflows/event.yml
```diff
@@ -1,4 +1,7 @@
-name: GitHub Event Handler
+# NOTE: currently azure-sdk-actions only hosts check enforcer code.
+# If further functionality is added, this name should be updated to reflect
+# the more generic behavior
+name: Check Enforcer
```
jsquire, line 4: Should we also rename the file to match?

jsquire (approved):

weshaggard:
@benbp I know we talked about this a little but what made you pull the trigger of updating this?

benbp:
@weshaggard i was already updating the files so figured I would change the title at the same time. I think we can change it back if/when it gets generalized but otherwise it's just "premature generalization" :)

weshaggard (approved):

weshaggard:
@benbp I guess I missed you updated that in the other PRs :)

.github/workflows/event.yml
benbp: I think it's a fair suggestion, but I want people be directed into this file if they try to add new event handlers. Otherwise we may end up having lots of duplicate runners per event (hence why I named it "GitHub Event Handler" originally).

PR #31312, opened 2022-09-21
Seprate the mgmt. with data plane pipeline to avoid live test timeout

ArthurMa1978:


ArthurMa1978:
same as https://github.com/Azure/azure-sdk-for-net/pull/31299

PR #31327, opened 2022-09-21
[Schema Registry] Support for fetching schema using group name, schema name, and version

m-redding:
These changes add support for getting schemas from Schema Registry using their group name, schema name, and version.


m-redding:
Resolves #30968

sdk/schemaregistry/Azure.Data.SchemaRegistry/tests/SchemaRegistryClientLiveTests.cs
```diff
@@ -11,7 +11,7 @@ namespace Azure.Data.SchemaRegistry.Tests
 {
     public class SchemaRegistryClientLiveTests : RecordedTestBase<SchemaRegistryClientTestEnvironment>
     {
-        public SchemaRegistryClientLiveTests(bool isAsync) : base(isAsync)
+        public SchemaRegistryClientLiveTests(bool isAsync) : base(isAsync, RecordedTestMode.Record)
```
JoshLove-msft, line 14: nit: you'll need to remove the mode

sdk/schemaregistry/Azure.Data.SchemaRegistry/tests/SchemaRegistryClientLiveTests.cs
```diff
@@ -92,10 +92,29 @@ public async Task CanGetSchema()
             AssertSchemaProperties(registerProperties, schemaName);
 
             SchemaRegistrySchema schema = await client.GetSchemaAsync(registerProperties.Value.Id);
-            AssertSchema(schema, schemaName);
+            AssertSchema(schema, schemaName, SchemaContent);
             AssertPropertiesAreEqual(registerProperties, schema.Properties);
         }
 
+        [RecordedTest]
+        public async Task CanGetSchemaVersion()
```
JoshLove-msft, line 100:
```suggestion
        public async Task CanGetSchemaByVersion()
```

sdk/schemaregistry/Azure.Data.SchemaRegistry/tests/SchemaRegistryClientLiveTests.cs
```diff
@@ -92,10 +92,29 @@ public async Task CanGetSchema()
...
+            var groupName = TestEnvironment.SchemaRegistryGroup;
+            var format = SchemaFormat.Avro;
+
+            var registerPropertiesv1 = await client.RegisterSchemaAsync(groupName, schemaName, SchemaContent, format);
+            AssertSchemaProperties(registerPropertiesv1, schemaName);
+
+            var registerPropertiesv2 = await client.RegisterSchemaAsync(groupName, schemaName, SchemaContent_V2, format);
+            AssertSchemaProperties(registerPropertiesv2, schemaName);
+
+            SchemaRegistrySchema schema = await client.GetSchemaAsync(groupName, schemaName, 2);
```
JoshLove-msft, line 113: nit: maybe also add a look up of schema version 1?

sdk/schemaregistry/Azure.Data.SchemaRegistry/src/SchemaRegistryClient.cs
```diff
@@ -234,6 +248,48 @@ await GetSchemaPropertiesInternalAsync(groupName, schemaName, schemaDefinition,
...
+        /// <param name="cancellationToken">The cancellation token for the operation.</param>
+        /// <returns>The properties of the schema, including the schema content provided by the service.</returns>
+#pragma warning disable AZC0015 // Unexpected client method return type.
+        public virtual Response<SchemaRegistrySchema> GetSchema(string groupName, string schemaName, int schemaVersion, CancellationToken cancellationToken = default) =>
+#pragma warning restore AZC0015 // Unexpected client method return type.
+            GetSchemaInternalAsync(groupName, schemaName, schemaVersion, false, cancellationToken).EnsureCompleted();
+
+        private async Task<Response<SchemaRegistrySchema>> GetSchemaInternalAsync(string groupName, string schemaName, int version, bool async, CancellationToken cancellationToken)
+        {
+            using DiagnosticScope scope = _clientDiagnostics.CreateScope(GetSchemaScopeName);
```
JoshLove-msft, line 266: We should align on the scope name used here across languages. I suspect other languages may be using "version" in the name because they have defined a separate method. Would you mind bringing this up in the Teams chat?

JoshLove-msft (commented):
Looks great! I think we just need some samples for the new methods and to figure out the correct scope name to use.

sdk/schemaregistry/Azure.Data.SchemaRegistry/tests/SchemaRegistryClientLiveTests.cs
```diff
@@ -144,11 +163,11 @@ public void GetSchemaPropertiesForNonexistentSchemaReturnsItemNotFoundErrorCode(
                     .And.Property(nameof(RequestFailedException.ErrorCode)).EqualTo("ItemNotFound"));
         }
 
-        private void AssertSchema(SchemaRegistrySchema schema, string schemaName)
+        private void AssertSchema(SchemaRegistrySchema schema, string schemaName, string schemaContent)
```
JoshLove-msft, line 166:
```suggestion
        private void AssertSchema(SchemaRegistrySchema schema, string expectedSchemaName, string expectedSchemaContent)
```

JoshLove-msft:
One thing I forgot to mention - we want to add a Version property to SchemaProperties now that we have a method that takes the Version.

sdk/schemaregistry/Azure.Data.SchemaRegistry/src/SchemaRegistryClient.cs
m-redding: Consensus as far as I could tell was to go with having version in the name

JoshLove-msft: Hmm I'm not sure this will work for us as there is tooling that checks that the name matches the method name.

JoshLove-msft: Though perhaps this validation is skipped when using ForwardClientCalls.

jsquire (commented):
LGTM.  I'll defer to Josh for approval.

sdk/schemaregistry/Azure.Data.SchemaRegistry/src/SchemaRegistryClient.cs
m-redding: If all the checks are passing does that mean that the validation is skipped? Or should we consider switching the method name to be `GetSchemaByVersion`?

m-redding: Updated the scope to reflect the method name based on the teams conversation

JoshLove-msft: Thanks for chasing this down!

JoshLove-msft (approved):

sdk/schemaregistry/Azure.Data.SchemaRegistry/src/SchemaProperties.cs
```diff
@@ -35,5 +36,10 @@ internal SchemaProperties(SchemaFormat format, string schemaId, string groupName
         /// The name of the schema.
         /// </summary>
         public string Name { get; }
+
+        /// <summary>
+        /// The version of the schema.
+        /// </summary>
+        public int? Version { get; }
```
JoshLove-msft, line 43: Hmm.. Why is this nullable? I guess it is based on the swagger not marking it as required..

m-redding: Yeah I was having trouble passing it through from the response since it's nullable there, it didn't seem to make sense to give it a default value like 0 or 1 or something, so I left it as nullable

JoshLove-msft: We should be able to customize this, but I wonder if we should ask Nick from the service team if it is correct that these properties are not listed as required.

JoshLove-msft: Turns out you can't mark response headers as required in swagger. We can just make this non-nullable though.

sdk/schemaregistry/Azure.Data.SchemaRegistry/tests/Samples/Sample01_ReadmeSnippets.cs
```diff
@@ -101,5 +101,21 @@ public void RetrieveSchema()
 
             Assert.AreEqual(Regex.Replace(_definition, @"\s+", string.Empty), definition);
         }
+
```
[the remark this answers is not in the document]
m-redding: I can put this change in its own PR

PR #31328, opened 2022-09-21
Azure Monitor Exporter - Add SampleRate to Activity.Event Exceptions

rajkumar-rangaraj:
- Added `SampleRate` to `Activity.Event` Exceptions with tests.

### DIFF
--- a/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Customizations/Models/TelemetryItem.cs
+++ b/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Customizations/Models/TelemetryItem.cs
@@ -74,6 +74,7 @@ namespace Azure.Monitor.OpenTelemetry.Exporter.Models
             Tags[ContextTagKeys.AiCloudRoleInstance.ToString()] = telemetryItem.Tags[ContextTagKeys.AiCloudRoleInstance.ToString()];
             Tags[ContextTagKeys.AiInternalSdkVersion.ToString()] = SdkVersionUtils.s_sdkVersion;
             InstrumentationKey = telemetryItem.InstrumentationKey;
+            SampleRate = telemetryItem.SampleRate;
         }
 
         public TelemetryItem (LogRecord logRecord, string roleName, string roleInstance, string instrumentationKey) :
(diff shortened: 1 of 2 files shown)


sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Customizations/Models/TelemetryItem.cs
```diff
@@ -74,6 +74,7 @@ internal partial class TelemetryItem
             Tags[ContextTagKeys.AiCloudRoleInstance.ToString()] = telemetryItem.Tags[ContextTagKeys.AiCloudRoleInstance.ToString()];
             Tags[ContextTagKeys.AiInternalSdkVersion.ToString()] = SdkVersionUtils.s_sdkVersion;
             InstrumentationKey = telemetryItem.InstrumentationKey;
+            SampleRate = telemetryItem.SampleRate;
```
reyang, line 77: Do we want to call this "sample rate" or "sample ratio"?
My understanding is that "sample rate" is an integer number for frequency based sampling (e.g. an audio stream has a sample rate of 44kHz), "sample ratio" is a floating point between [0, 1] which represents the actual samples taken divided by the total number of observable samples.

TimothyMothra: Excellent point, but I think our hands are tied on this one.

`SampleRate` comes from Breeze's schema.
https://github.com/Azure/azure-rest-api-specs/blob/e88242fc7da7860d36ab4a1700e3404a51d53eeb/specification/applicationinsights/data-plane/Monitor.Exporters/preview/2020-09-15_Preview/swagger.json#L855-L859

rajkumar-rangaraj: Customer will set sample ratio in the `TracerProvider`. Our backend schema accepts only `sampleRate`, we have to convert ratio to percentage and send it.

```
using var tracerProvider = Sdk.CreateTracerProviderBuilder()
                            .SetSampler(new ApplicationInsightsSampler(0.5F))
                            .AddAzureMonitorTraceExporter(o =>
                            {
                                o.ConnectionString = ConnectionString;
                            })
                            .Build();
```

TimothyMothra (approved):

vishweshbankwar (approved):

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Customizations/Models/TelemetryItem.cs
reyang: Anything we can do to make it clear? e.g. adding a comment explaining that the naming/concept is misleading, e.g. raising this to the ingestion team

reyang (approved):

PR #31344, opened 2022-09-22
Make sequence field in AmqpMessageBody readonly like the other ctor assigned fields

danielmarbach:

### DIFF
--- a/sdk/core/Azure.Core.Amqp/src/AmqpMessageBody.cs
+++ b/sdk/core/Azure.Core.Amqp/src/AmqpMessageBody.cs
@@ -27,7 +27,7 @@ namespace Azure.Core.Amqp
         /// <summary>
         /// The sequence sections for the AMQP message body.
         /// </summary>
-        private IEnumerable<IList<object>>? _sequence;
+        private readonly IEnumerable<IList<object>>? _sequence;
 
         /// <summary>
         /// Gets the type of the message body.


jsquire (approved):

jsquire:
Build failure was unrelated to these changes; I restarted the leg that timed out.

christothes (commented):
If there are any perf benefits here, it would be great to see a micro-benchmark.

jsquire:
> If there are any perf benefits here, it would be great to see a micro-benchmark.

@christothes:  Just to clarify - does that indicate concern about taking the change if there's no impact?   From a consistency of pattern standpoint, I think there's merit to this.

danielmarbach:
@christothes I did not look at this from a perf perspective. I noticed that `_data` and `_value` are `readonly` while `_segment` is not while it could be. Since there was no explicit reason mentioned why it shouldn't be read-only I aligned it with the other fields. That's it. 

https://github.com/Azure/azure-sdk-for-net/blob/969cb74b10e3137026e94d2764514cc429cc2345/sdk/core/Azure.Core.Amqp/src/AmqpMessageBody.cs#L20-L26

I'm OK if you plan not to take this. It took my only a few min to contribute, including this comment. Yet, I'm not planning to spend more time on this, like writing and running a benchmark, than I already have by sending in this change. Thanks for your patience with me.

PR #31361, opened 2022-09-22
Check that mgmt libraries have a trigger path in resourcemanager ci.mgmt.yml file

weshaggard:
A better way to add the check from https://github.com/Azure/azure-sdk-for-net/pull/31021.

@JoshLove-msft FYI we should try to move more of the checks from CodeChecks into msbuild targets similar to this so that we can reduce some of the complexity inside there and prevent jumping in and out of PS and msbuild.

### DIFF
--- a/eng/Directory.Build.Common.targets
+++ b/eng/Directory.Build.Common.targets
@@ -273,9 +273,9 @@
   </Target>
 
   <!-- Validates that all the mgmt libraries have a trigger path in the core resourcemanager pipeline -->
-  <Target Name="ValidateResourceManagerPipelineTriggers" AfterTargets="Build" Condition="'$(IsMgmtLibrary)' == 'true' and '$(IsShippingLibrary)' == 'true'" >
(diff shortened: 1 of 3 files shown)


benbp (approved):

JoshLove-msft:
> A better way to add the check from #31021.
> 
> @JoshLove-msft FYI we should try to move more of the checks from CodeChecks into msbuild targets similar to this so that we can reduce some of the complexity inside there and prevent jumping in and out of PS and msbuild.

That's fair, and I suppose this would let devs catch these instances locally rather than waiting for the Analyze step in CI.

weshaggard:
>Though I must say that I believe it was a conscious choice not to perform all of these checks on build as it would significantly add to local dev loop.

We can still pick and choose which ones we like to run by default in a build and capture them behind some flags that we can run in CI or devs can still run locally if they choose by using msbuild commands though instead of a wrapper script.  Don't get me wrong there will still be some purpose for this codechecks.ps1 script but I think a number of these checks can just happen in msbuild.

PR #31366, opened 2022-09-22
Update NUnit3TestAdapter to support net7.0

### LINKED ISSUE #31365: Visual Studio 2022 Preview no longer runs .NET Framework tests
After the most recent update to Visual Studio 2022 Version 17.4.0 Preview 2.1, Test Explorer no longer discovers/runs .NET Framework tests. This appears to be because of https://github.com/nunit/nunit3-vs-adapter/issues/987. See https://github.com/dotnet/test-templates/pull/252 for more information as well.

heaths:
Fixes #31365

### DIFF
--- a/eng/Packages.Data.props
+++ b/eng/Packages.Data.props
@@ -245,7 +245,7 @@
     <PackageReference Update="Newtonsoft.Json" Version="10.0.3" />
     <PackageReference Update="NSubstitute" Version="3.1.0" />
     <PackageReference Update="NUnit" Version="3.13.2" />
-    <PackageReference Update="NUnit3TestAdapter" Version="4.2.1" />
+    <PackageReference Update="NUnit3TestAdapter" Version="4.3.0-alpha-net7.4" />
     <PackageReference Update="OpenTelemetry.Extensions.Hosting" Version="1.0.0-rc9.3" />
     <PackageReference Update="OpenTelemetry.Instrumentation.AspNetCore" Version="1.0.0-rc9.3" />
     <PackageReference Update="OpenTelemetry.Instrumentation.Http" Version="1.0.0-rc9.3" />


eng/Packages.Data.props
```diff
@@ -245,7 +245,7 @@
     <PackageReference Update="Newtonsoft.Json" Version="10.0.3" />
     <PackageReference Update="NSubstitute" Version="3.1.0" />
     <PackageReference Update="NUnit" Version="3.13.2" />
-    <PackageReference Update="NUnit3TestAdapter" Version="4.2.1" />
+    <PackageReference Update="NUnit3TestAdapter" Version="4.3.0-alpha-net7.4" />
```
JoshLove-msft, line 248: Are we starting to test against .NET 7?

JoshLove-msft (approved):

eng/Packages.Data.props
JoshLove-msft: Ah I read the linked issue.

weshaggard, line 248: In general we try to avoid prerelease packages because of stability issues. We should make a team decision before taking this update. /cc @pallavit @jsquire

heaths: Certainly, for production, but this is test-only. Without it, VS Tests against .NET Framework don't run. Seems like a pretty serious gap.

weshaggard: They don't run on a system with a preview version of .NET 7? That seems like a pretty big back compat bug on the VS side. Do you know if this is temporary?

JoshLove-msft: This alpha version was shipped 3 months ago.. Perhaps we should ask the NUnit folks when they plan on releasing a stable version.

JoshLove-msft: Perhaps they are waiting for the stable .NET 7..

jsquire (approved):

eng/Packages.Data.props
jsquire: I'm good with this based on the need and scope of test-only.

PR #31367, opened 2022-09-23
Target base scale support

alrod:
To support "Target Base Scale" for "Runtime Driven Scale" the ServiceBus extension needs to implement [ITargetScalerProvider](https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Scale/ITargetScalerProvider.cs) and [ITargetScaler](https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Scale/ITargetScaler.cs) which was introduced in "Microsoft.Azure.WebJobs.3.0.36".  The approach is similar to incremental scale - `IScaleMonitorProvider/IScaleMonitor.`

The PR is similar to the EventHub PR:
https://github.com/Azure/azure-sdk-for-net/pull/33675/

### DIFF
--- a/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusListener.cs
+++ b/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusListener.cs
@@ -9,8 +9,10 @@ using System.Threading;
 using System.Threading.Tasks;
 using Azure.Core.Pipeline;
 using Azure.Messaging.ServiceBus;
+using Azure.Messaging.ServiceBus.Administration;
 using Azure.Messaging.ServiceBus.Diagnostics;
 using Microsoft.Azure.WebJobs.Extensions.ServiceBus.Config;
+using Microsoft.Azure.WebJobs.Extensions.ServiceBus.Listeners;
 using Microsoft.Azure.WebJobs.Host.Executors;
 using Microsoft.Azure.WebJobs.Host.Listeners;
 using Microsoft.Azure.WebJobs.Host.Scale;
@@ -18,7 +20,7 @@ using Microsoft.Extensions.Logging;
 
 namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
 {
-    internal sealed class ServiceBusListener : IListener, IScaleMonitorProvider
+    internal sealed class ServiceBusListener : IListener, IScaleMonitorProvider, ITargetScalerProvider
     {
         private readonly ITriggeredFunctionExecutor _triggerExecutor;
         private readonly string _entityPath;
@@ -34,6 +36,8 @@ namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
         private readonly Lazy<ServiceBusClient> _client;
         private readonly Lazy<SessionMessageProcessor> _sessionMessageProcessor;
         private readonly Lazy<ServiceBusScaleMonitor> _scaleMonitor;
+        private readonly Lazy<ServiceBusTargetScaler> _targetScaler;
+        private readonly Lazy<ServiceBusAdministrationClient> _administrationClient;
         private readonly ConcurrencyUpdateManager _concurrencyUpdateManager;
 
         // internal for testing
@@ -72,8 +76,7 @@ namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
             _functionId = functionId;
 
             _client = new Lazy<ServiceBusClient>(
-                () =>
-                    clientFactory.CreateClientFromSetting(connection));
+                () => clientFactory.CreateClientFromSetting(connection));
 
             _batchReceiver = new Lazy<ServiceBusReceiver>(
                 () => messagingProvider.CreateBatchMessageReceiver(
@@ -95,15 +98,31 @@ namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
                     return messagingProvider.CreateSessionMessageProcessor(_client.Value,_entityPath, sessionProcessorOptions);
                 });
 
+            _administrationClient = new Lazy<ServiceBusAdministrationClient>(
+                () => clientFactory.CreateAdministrationClient(connection));
+
             _scaleMonitor = new Lazy<ServiceBusScaleMonitor>(
                 () => new ServiceBusScaleMonitor(
                     functionId,
-                    entityType,
                     _entityPath,
-                    connection,
+                    entityType,
                     _batchReceiver,
-                    loggerFactory,
-                    clientFactory));
+                    _administrationClient,
+                    loggerFactory
+                    ));
+
+            _targetScaler = new Lazy<ServiceBusTargetScaler>(
+                () => new ServiceBusTargetScaler(
+                    functionId,
+                    _entityPath,
+                    entityType,
+                    _batchReceiver,
+                    _administrationClient,
+                    options,
+                    _isSessionsEnabled,
+                    _singleDispatch,
+                    loggerFactory
+                    ));
 
             _scopeFactory = new Lazy<EntityScopeFactory>(
                 () => new EntityScopeFactory(_batchReceiver.Value.EntityPath, _batchReceiver.Value.FullyQualifiedNamespace));
@@ -534,6 +553,11 @@ namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
         public IScaleMonitor GetMonitor()
         {
             return _scaleMonitor.Value;
+        }
+
+        public ITargetScaler GetTargetScaler()
+        {
+            return _targetScaler.Value;
         }
 
         /// <summary>
--- a/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusMetricsProvider.cs
+++ b/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusMetricsProvider.cs
@@ -70,7 +70,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.ServiceBus.Listeners
             }
             catch (UnauthorizedAccessException ex)
             {
-                _logger.LogWarning($"Connection string does not have 'Manage Claim' for {entityName} '{_entityPath}'.Unable to determine active message count.", ex);
+                _logger.LogWarning($"Connection string does not have 'Manage Claim' for {entityName} '{_entityPath}'. Unable to determine active message count.", ex);
                 throw ex;
             }
             catch (Exception e)
--- a/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusMetricsProvider.cs
+++ b/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusMetricsProvider.cs
@@ -70,7 +70,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.ServiceBus.Listeners
             }
             catch (UnauthorizedAccessException ex)
             {
-                _logger.LogWarning($"Connection string does not have 'Manage Claim' for {entityName} '{_entityPath}'. Unable to determine active message count.", ex);
+                if (TimeToLogWarning())
+                {
+                    _logger.LogWarning($"Connection string does not have 'Manage Claim' for {entityName} '{_entityPath}'. Unable to determine active message count.", ex);
+                }
                 throw ex;
             }
             catch (Exception e)
--- a/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusMetricsProvider.cs
+++ b/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusMetricsProvider.cs
@@ -1,0 +1,236 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Azure.WebJobs.ServiceBus.Listeners;
+using Microsoft.Azure.WebJobs.ServiceBus;
+using System;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Azure.Messaging.ServiceBus;
+using Azure.Messaging.ServiceBus.Administration;
+using Microsoft.Azure.WebJobs.Extensions.ServiceBus.Config;
+
+namespace Microsoft.Azure.WebJobs.Extensions.ServiceBus.Listeners
+{
+    internal class ServiceBusMetricsProvider
+    {
+        private const string DeadLetterQueuePath = @"/$DeadLetterQueue";
+
+        private readonly ILogger _logger;
+        private readonly string _entityPath;
+        private readonly ServiceBusEntityType _serviceBusEntityType;
+        private readonly Lazy<ServiceBusReceiver> _receiver;
+        private readonly bool _isListeningOnDeadLetterQueue;
+        private readonly Lazy<ServiceBusAdministrationClient> _administrationClient;
+
+        private DateTime _nextWarningTime;
+
+        public ServiceBusMetricsProvider(
+            string entityPath,
+            ServiceBusEntityType serviceBusEntityType,
+            Lazy<ServiceBusReceiver> receiver,
+            Lazy<ServiceBusAdministrationClient> administrationClient,
+            ILoggerFactory loggerFactory)
+        {
+            _serviceBusEntityType = serviceBusEntityType;
+            _receiver = receiver;
+            _entityPath = entityPath;
+            _isListeningOnDeadLetterQueue = entityPath.EndsWith(DeadLetterQueuePath, StringComparison.OrdinalIgnoreCase);
+            _administrationClient = administrationClient;
+            _logger = loggerFactory.CreateLogger<ServiceBusMetricsProvider>();
+            _nextWarningTime = DateTime.UtcNow;
+        }
+
+        public async Task<long> GetMessageCountAsync()
+        {
+            long activeMessageCount = 0;
+            long deadLetterCount = 0;
+            string entityName = _serviceBusEntityType == ServiceBusEntityType.Queue ? "queue" : "topic";
+            try
+            {
+                if (_serviceBusEntityType == ServiceBusEntityType.Queue)
+                {
+                    QueueRuntimeProperties queueRuntimeProperties = await _administrationClient.Value.GetQueueRuntimePropertiesAsync(_entityPath).ConfigureAwait(false);
+                    activeMessageCount = queueRuntimeProperties.ActiveMessageCount;
+                    deadLetterCount = queueRuntimeProperties.DeadLetterMessageCount;
+                }
+                else
+                {
+                    ServiceBusEntityPathHelper.ParseTopicAndSubscription(_entityPath, out string topicPath, out string subscriptionPath);
+
+                    SubscriptionRuntimeProperties subscriptionProperties = await _administrationClient.Value.GetSubscriptionRuntimePropertiesAsync(topicPath, subscriptionPath).ConfigureAwait(false);
+                    activeMessageCount = subscriptionProperties.ActiveMessageCount;
+                    deadLetterCount = subscriptionProperties.DeadLetterMessageCount;
+                }
+            }
+            catch (ServiceBusException ex)
+            when (ex.Reason == ServiceBusFailureReason.MessagingEntityNotFound)
+            {
+                _logger.LogWarning($"ServiceBus {entityName} '{_entityPath}' was not found.");
+            }
+            catch (UnauthorizedAccessException ex)
+            {
+                _logger.LogWarning($"Connection string does not have 'Manage Claim' for {entityName} '{_entityPath}'.Unable to determine active message count.", ex);
+                throw ex;
+            }
+            catch (Exception e)
+            {
+                _logger.LogWarning($"Error querying for Service Bus {entityName} scale status: {e.Message}");
+            }
+
+            long totalNewMessageCount = 0;
+            if ((!_isListeningOnDeadLetterQueue && activeMessageCount > 0) || (_isListeningOnDeadLetterQueue && deadLetterCount > 0))
+            {
+                totalNewMessageCount = _isListeningOnDeadLetterQueue ? deadLetterCount : activeMessageCount;
+            }
+
+            return totalNewMessageCount;
+        }
+
+        public async Task<ServiceBusTriggerMetrics> GetMetricsAsync()
+        {
+            ServiceBusReceivedMessage activeMessage = null;
+            string entityName = _serviceBusEntityType == ServiceBusEntityType.Queue ? "queue" : "topic";
+
+            try
+            {
+                // Do a first attempt to peek one message from the head of the queue
+                var peekedMessage = await _receiver.Value.PeekMessageAsync(fromSequenceNumber: 0).ConfigureAwait(false);
+                if (peekedMessage == null)
+                {
+                    // ignore it. The Get[Queue|Topic]MetricsAsync methods deal with activeMessage being null
+                }
+                else if (peekedMessage.State == ServiceBusMessageState.Active)
+                {
+                    activeMessage = peekedMessage;
+                }
+                else
+                {
+                    // Do another attempt to peek ten message from last peek sequence number
+                    var peekedMessages = await _receiver.Value.PeekMessagesAsync(10, fromSequenceNumber: peekedMessage.SequenceNumber).ConfigureAwait(false);
+                    foreach (var receivedMessage in peekedMessages)
+                    {
+                        if (receivedMessage.State == ServiceBusMessageState.Active)
+                        {
+                            activeMessage = receivedMessage;
+                            break;
+                        }
+                    }
+
+                    // Batch contains messages but none are active in the peeked batch
+                    if (peekedMessages.Count > 0 && activeMessage == null)
+                    {
+                        _logger.LogDebug("{_serviceBusEntityType} {_entityPath} contains multiple messages but none are active in the peeked batch.");
+                    }
+                }
+
+                if (_serviceBusEntityType == ServiceBusEntityType.Queue)
+                {
+                    return await GetQueueMetricsAsync(activeMessage).ConfigureAwait(false);
+                }
+                else
+                {
+                    return await GetTopicMetricsAsync(activeMessage).ConfigureAwait(false);
+                }
+            }
+            catch (ServiceBusException ex)
+            when (ex.Reason == ServiceBusFailureReason.MessagingEntityNotFound)
+            {
+                _logger.LogWarning($"ServiceBus {entityName} '{_entityPath}' was not found.");
+            }
+            catch (UnauthorizedAccessException) // When manage claim is not used on Service Bus connection string
+            {
+                if (TimeToLogWarning())
+                {
+                    _logger.LogWarning($"Connection string does not have Manage claim for {entityName} '{_entityPath}'. Failed to get {entityName} description to " +
+                        $"derive {entityName} length metrics. Falling back to using first message enqueued time.");
+                }
+            }
+            catch (Exception e)
+            {
+                _logger.LogWarning($"Error querying for Service Bus {entityName} scale status: {e.Message}");
+            }
+
+            // Path for connection strings with no manage claim
+            return CreateTriggerMetrics(activeMessage, 0, 0, 0, _isListeningOnDeadLetterQueue);
+        }
+
+        private async Task<ServiceBusTriggerMetrics> GetQueueMetricsAsync(ServiceBusReceivedMessage message)
+        {
+            QueueRuntimeProperties queueRuntimeProperties;
+            QueueProperties queueProperties;
+            long activeMessageCount = 0;
+            long deadLetterCount = 0;
+            int partitionCount = 0;
+
+            queueRuntimeProperties = await _administrationClient.Value.GetQueueRuntimePropertiesAsync(_entityPath).ConfigureAwait(false);
+            activeMessageCount = queueRuntimeProperties.ActiveMessageCount;
+            deadLetterCount = queueRuntimeProperties.DeadLetterMessageCount;
+
+            // If partitioning is turned on, then Service Bus automatically partitions queues into 16 partitions
+            // See more information here: https://docs.microsoft.com/azure/service-bus-messaging/service-bus-partitioning#standard
+            queueProperties = await _administrationClient.Value.GetQueueAsync(_entityPath).ConfigureAwait(false);
+            partitionCount = queueProperties.EnablePartitioning ? 16 : 0;
+
+            return CreateTriggerMetrics(message, activeMessageCount, deadLetterCount, partitionCount, _isListeningOnDeadLetterQueue);
+        }
+
+        private async Task<ServiceBusTriggerMetrics> GetTopicMetricsAsync(ServiceBusReceivedMessage message)
+        {
+            TopicProperties topicProperties;
+            SubscriptionRuntimeProperties subscriptionProperties;
+            string topicPath, subscriptionPath;
+            long activeMessageCount = 0;
+            long deadLetterCount = 0;
+            int partitionCount = 0;
+
+            ServiceBusEntityPathHelper.ParseTopicAndSubscription(_entityPath, out topicPath, out subscriptionPath);
+
+            subscriptionProperties = await _administrationClient.Value.GetSubscriptionRuntimePropertiesAsync(topicPath, subscriptionPath).ConfigureAwait(false);
+            activeMessageCount = subscriptionProperties.ActiveMessageCount;
+            deadLetterCount = subscriptionProperties.DeadLetterMessageCount;
+
+            // If partitioning is turned on, then Service Bus automatically partitions queues into 16 partitions
+            // See more information here: https://docs.microsoft.com/azure/service-bus-messaging/service-bus-partitioning#standard
+            topicProperties = await _administrationClient.Value.GetTopicAsync(topicPath).ConfigureAwait(false);
+            partitionCount = topicProperties.EnablePartitioning ? 16 : 0;
+
+            return CreateTriggerMetrics(message, activeMessageCount, deadLetterCount, partitionCount, _isListeningOnDeadLetterQueue);
+        }
+
+        private bool TimeToLogWarning()
+        {
+            DateTime currentTime = DateTime.UtcNow;
+            bool timeToLog = currentTime >= _nextWarningTime;
+            if (timeToLog)
+            {
+                _nextWarningTime = currentTime.AddHours(1);
+            }
+            return timeToLog;
+        }
+
+        internal static ServiceBusTriggerMetrics CreateTriggerMetrics(ServiceBusReceivedMessage message, long activeMessageCount, long deadLetterCount, int partitionCount, bool isListeningOnDeadLetterQueue)
+        {
+            long totalNewMessageCount = 0;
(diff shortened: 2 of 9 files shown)


sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusTargetScaler.cs
```diff
@@ -0,0 +1,65 @@
...
+                }
+            }
+            else
+            {
+                concurrency = context.InstanceConcurrency.Value;
+            }
+
+            return new TargetScalerResult
+            {
+                WorkerCountDifference = (int)Math.Ceiling((metrics.MessageCount - context.WorkerCount * concurrency) / (decimal)concurrency)
```
chiangvincent, line 61: Nit: I think this formula could be clearer, `(int)Math.Ceiling((metrics.MessageCount / (decimal)concurrency))` is the `targetWorkerCount`, and the `WorkerCountDifference = targetWorkerCount - context.WorkerCount`

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusTargetScaler.cs
```diff
@@ -0,0 +1,65 @@
...
+            _options = options;
+            _isSessionsEnabled = isSessionsEnabled;
+            _targetScalerDescriptor = new TargetScalerDescriptor($"{_functionId}-ServiceBusTrigger-{_entityPath}".ToLower(CultureInfo.InvariantCulture), functionId);
+        }
+
+        public TargetScalerDescriptor TargetScalerDescriptor => _targetScalerDescriptor;
+
+        public async Task<TargetScalerResult> GetScaleResultAsync(TargetScaleStatusContext context)
+        {
+            ServiceBusTriggerMetrics metrics = await _serviceBusMetricsReceiver.GetMetricsAsync().ConfigureAwait(false);
```
mathewc, line 36: GetMetricsAsync does additional work that a target scaler doesn't need, for example peeking messages to get QueueTime. That's an unnecessary performance hit for target scale scenarios. To avoid this, I think we can just have the "metrics receiver" expose a new GetMessageCountAsync() method that does the isListeningOnDeadLetterQueue logic and returns the count, and the GetQueueMetrics/GetTopicMetrics method can be refactored to call that, and you can then just call that from this scaler as well?

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusTargetScaler.cs
mathewc: I don't think a target scaler should be computing a worker count difference. It should just compute the target worker count and return that. It should be the responsibility of scale controller to us that information to compute any deltas as needed.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusMetricsReceiver.cs
```diff
@@ -0,0 +1,195 @@
...
+using System;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Azure.Messaging.ServiceBus;
+using Azure.Messaging.ServiceBus.Administration;
+using Microsoft.Azure.WebJobs.Extensions.ServiceBus.Config;
+
+namespace Microsoft.Azure.WebJobs.Extensions.ServiceBus.Listeners
+{
+    internal class ServiceBusMetricsReceiver
```
mathewc, line 15: I think we should rename this to ServiceBusMetricsProvider. "Receiver" has special meaning in SB SDK - it generally refers to an event based client model.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusListener.cs
```diff
@@ -95,15 +98,27 @@ internal sealed class ServiceBusListener : IListener, IScaleMonitorProvider
...
                     _entityPath,
-                    connection,
-                    _batchReceiver,
-                    loggerFactory,
-                    clientFactory));
+                    _serviceBusMetricsReceiver,
+                    loggerFactory
+                    ));
+
+            if (singleDispatch)
```
mathewc, line 111: Why do we only support this for single dispatch? For batch, wouldn't concurrency be MaxMessageBatchSize?

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/tests/Listeners/ServiceBusScaleMonitorTests.cs
```diff
@@ -122,31 +123,31 @@ public void GetMetrics_ReturnsExpectedResult()
             var message = ServiceBusModelFactory.ServiceBusReceivedMessage(enqueuedTime: utcNow);
 
             // Test base case
-            var metrics = ServiceBusScaleMonitor.CreateTriggerMetrics(null, 0, 0, 0, false);
+            var metrics = ServiceBusMetricsReceiver.CreateTriggerMetrics(null, 0, 0, 0, false);
```
mathewc, line 126: After our refactoring these methods have moved to a new type. We should move the corresponding tests into a new ServiceBusMetricsProvider test suite

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusTargetScaler.cs
```diff
@@ -0,0 +1,65 @@
...
+using Microsoft.Azure.WebJobs.Host.Scale;
+using Microsoft.Azure.WebJobs.ServiceBus;
+using Microsoft.Azure.WebJobs.ServiceBus.Listeners;
+using System;
+using System.Globalization;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs.Extensions.ServiceBus.Listeners
+{
+    internal class ServiceBusTargetScaler : ITargetScaler
```
chiangvincent, line 13: This class will be instantiated and directly called by the ScaleControllerRedesign should be `public`

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusScaleMonitor.cs
```diff
@@ -11,35 +11,25 @@
 using Azure.Messaging.ServiceBus;
 using Azure.Messaging.ServiceBus.Administration;
 using Microsoft.Azure.WebJobs.Extensions.ServiceBus.Config;
+using Microsoft.Azure.WebJobs.Extensions.ServiceBus.Listeners;
 
 namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
 {
     internal class ServiceBusScaleMonitor : IScaleMonitor<ServiceBusTriggerMetrics>
```
chiangvincent, line 18: This class will be called directly by the ScaleControllerRedesign

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusMetricsReceiver.cs
chiangvincent, line 15: This class will be called directly by the ScaleControllerRedesign, please make `public`

mathewc: Because it was internal, I wasn't too concerned with its public interface. For any surface area we decide we need to promote to public, we'll need to do a much more thorough review/design to ensure we're only exposing what is needed, and that the APIs are clean and future proof.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusMetricsReceiver.cs
```diff
@@ -0,0 +1,195 @@
...
+        public ServiceBusMetricsReceiver(
+            string connection,
+            string entityPath,
+            ServiceBusEntityType serviceBusEntityType,
+            Lazy<ServiceBusReceiver> receiver,
+            ServiceBusClientFactory clientFactory,
+            ILoggerFactory loggerFactory)
+        {
+            _serviceBusEntityType = serviceBusEntityType;
+            _receiver = receiver;
```
chiangvincent, line 37: duplicated

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusTargetScaler.cs
alrod: The additional work is getting partition information - we can add a parameter to existing GetMetricsAsync to skip this. Introducing the new method will require a lot of the code duplication.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusListener.cs
alrod: MaxMessageBatchSize limits the batch size, actual batch size can by anything between [1..MaxMessageBatchSize].
A function app can have the triggers with or without TBS so SC will handle this case.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusMetricsReceiver.cs
alrod: I changed the `ServiceBusMetricsReceiver` ctor so it can stay internal

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusTargetScaler.cs
```diff
@@ -0,0 +1,83 @@
...
+                    concurrency = _options.MaxConcurrentCalls;
+                }
+            }
+            else
+            {
+                concurrency = context.InstanceConcurrency.Value;
+            }
+
+            int targetWorkerCount = (int)Math.Ceiling(messageCount / (decimal)concurrency);
+            _logger.LogInformation($"'Target worker count for function '{_functionId}' is '{targetWorkerCount}' (EntityPath='{_entityPath}', MessageCount ='{messageCount}', Concurrecny='{concurrency}').");
```
chiangvincent, line 75: Nit: typo Concurrency

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusTargetScaler.cs
```diff
@@ -0,0 +1,83 @@
...
+            ServiceBusEntityType entityType,
+            Lazy<ServiceBusReceiver> receiver,
+            Lazy<ServiceBusAdministrationClient> administrationClient,
+            ServiceBusOptions options,
+            bool isSessionsEnabled,
+            ILoggerFactory loggerFactory
+            )
+        {
+            _functionId = functionId;
+            _serviceBusMetricsProvider = new ServiceBusMetricsProvider(entityPath, entityType, receiver, administrationClient, loggerFactory);
```
seilorjunior, line 39: @alrod  do you think that will be better to have the ServiceBusMetricsProvider as interface? with that we can have an abstraction level in this constructor and will be easy to create unit tests.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Constants.cs
```diff
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
+using System.Data.SqlTypes;
```
mathewc, line 4: Why was this change needed?

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusMetricsProvider.cs
```diff
@@ -0,0 +1,198 @@
...
+            _serviceBusEntityType = serviceBusEntityType;
+            _receiver = receiver;
+            _entityPath = entityPath;
+            _isListeningOnDeadLetterQueue = entityPath.EndsWith(DeadLetterQueuePath, StringComparison.OrdinalIgnoreCase);
+            _administrationClient = administrationClient;
+            _logger = loggerFactory.CreateLogger<ServiceBusMetricsProvider>();
+            _nextWarningTime = DateTime.UtcNow;
+        }
+
+        public async Task<ServiceBusTriggerMetrics> GetMetricsAsync(bool skipPartitionCount = false)
```
mathewc, line 44: Why do we want to skip partition counts? E.g. for TBS just like the existing code for scale monitor, we don't want to scale beyond the number of partitions, right?

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusTargetScaler.cs
```diff
@@ -0,0 +1,83 @@
...
+            _options = options;
+            _isSessionsEnabled = isSessionsEnabled;
+        }
+
+        public TargetScalerDescriptor TargetScalerDescriptor => _targetScalerDescriptor;
+
+        public async Task<TargetScalerResult> GetScaleResultAsync(TargetScalerContext context)
+        {
+            ServiceBusTriggerMetrics metrics = await _serviceBusMetricsProvider.GetMetricsAsync(true).ConfigureAwait(false);
+            return GetScaleResultInternal(context, metrics.MessageCount);
```
mathewc, line 52: Target scaler only needs this MessageCount, doesn't need QueueTime or anything else. So calling GetMetricsAsync will result in unnecessary Peek operations to determine QueueTime when it's not even used. So assuming we also need PartitionCount, I think you'd have a separate method ServiceBusMetricsProvider.GetRuntimeProperties that returns a type with MessageCount/PartitionCount. And that method would be used here, as well as from GetQueueMetricsAsync/GetTopicMetricsAsync.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusTargetScaler.cs
```diff
@@ -0,0 +1,83 @@
...
+                {
+                    concurrency = _options.MaxConcurrentCalls;
+                }
+            }
+            else
+            {
+                concurrency = context.InstanceConcurrency.Value;
+            }
+
+            int targetWorkerCount = (int)Math.Ceiling(messageCount / (decimal)concurrency);
```
mathewc, line 74: As commented above, why would we ever want to scale beyond the partition count?

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusTriggerMetrics.cs
```diff
@@ -8,7 +8,7 @@
 
 namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
 {
-    internal class ServiceBusTriggerMetrics : ScaleMetrics
+    public class ServiceBusTriggerMetrics : ScaleMetrics
```
mathewc, line 11: why does this need to be public. shouldn't have to be

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusTargetScaler.cs
mathewc: We're trying to minimize public surface area and don't want a new public type for this abstraction, that's why it's internal.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusTriggerMetrics.cs
alrod: @chiangvincent asked `ServiceBusScaleMonitor` to be paublic for SC redesign. ServiceBusScaleMonitor inherits from  
`public class ServiceBusScaleMonitor : IScaleMonitor<ServiceBusTriggerMetrics>`

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusMetricsProvider.cs
alrod: As discussed offline we can scale beyond the partition count

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusTargetScaler.cs
alrod: As discussed offline we can scale beyond the partition count

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusTargetScaler.cs
alrod: there are 3 await operations in the `ServiceBusMetricsProvider`:
1. `PeekMessageAsync`
2. `GetQueueRuntimePropertiesAsync`or `GetSubscriptionRuntimePropertiesAsync`
3. `GetQueueAsync` or `GetTopicAsync`

We need 1 and 2 to get active message count. PeekMessageAsync is needed for ['no manage claim case'](https://github.com/Azure/azure-sdk-for-net/pull/31367/files/8110f7a257d6506a7906bc07b6284998633f3d76#diff-a3f037667869ab7bb60b69537d01e7999b070bc2318b0478ead0d987bc4000c9R182)

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/tests/Listeners/ServiceBusTargetScalerTests.cs
```diff
@@ -0,0 +1,64 @@
...
+            _loggerFactory = new LoggerFactory();
+            _loggerProvider = new TestLoggerProvider();
+            _loggerFactory.AddProvider(_loggerProvider);
+        }
+
+        [TestCase(100, false, true, null, 6)]
+        [TestCase(100, true, true, null, 4)]
+        [TestCase(100, false, true, 19, 6)]
+        [TestCase(100, false, false, null, 3)]
+        public void ServiceBusTargetScaler_Returns_Expected(int messageCount, bool isSessionEnabled, bool signleDispatch, int? concurrecny,int expected)
```
chiangvincent, line 30: nit: typo singleDispatch

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/tests/Microsoft.Azure.WebJobs.Extensions.ServiceBus.Tests.csproj
```diff
@@ -29,4 +29,9 @@
   <ItemGroup Condition="'$(IsTargetingNetFx)' == 'true'">
     <Reference Include="System.Transactions" />
   </ItemGroup>
-</Project>  
\ No newline at end of file
+
+  <ItemGroup>
+    <PackageReference  Include="Microsoft.Azure.WebJobs" VersionOverride="3.0.34-11969" />
```
chiangvincent, line 34: update to 3.0.35?

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Microsoft.Azure.WebJobs.Extensions.ServiceBus.csproj
```diff
@@ -36,4 +36,8 @@
     <Compile Include="$(AzureCoreSharedSources)AppContextSwitchHelper.cs" LinkBase="Shared" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference  Include="Microsoft.Azure.WebJobs" VersionOverride="3.0.34-11969" />
```
chiangvincent, line 40: update to 3.0.35?

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusTargetScaler.cs
```diff
@@ -0,0 +1,94 @@
...
+                    {
+                        concurrency = _options.MaxConcurrentCalls;
+                    }
+                }
+            }
+            else
+            {
+                concurrency = context.InstanceConcurrency.Value;
+            }
+
```
chiangvincent, line 84: We should validate that whatever concurrency value we end up using is > 0 (to ensure we don't do division by 0). Suggestion: throw ArgumentOutofRangeException?

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusScaleMonitor.cs
```diff
@@ -57,140 +54,7 @@ async Task<ScaleMetrics> IScaleMonitor.GetMetricsAsync()
...
-                MessageCount = totalNewMessageCount,
-                PartitionCount = partitionCount,
-                QueueTime = queueTime
-            };
-        }
-
-        private static bool MessageIsActive(ServiceBusReceivedMessage message)
-        {
-            return message.State == ServiceBusMessageState.Active;
+            return await _serviceBusMetricsProvider.GetMetricsAsync().ConfigureAwait(false);
```
chiangvincent, line 57: Just curious what is purpose of ConfigureAwait(false) in this context?

NuGet.Config
```diff
@@ -10,6 +10,8 @@
       Used for azure-sdk-tools repo until issue https://github.com/Azure/azure-sdk-tools/issues/1329 is addressed
     -->
     <add key="azure-sdk-tools" value="https://azuresdkartifacts.blob.core.windows.net/azure-sdk-tools/index.json" />
+    <add key="azure-appservice" value="https://www.myget.org/F/azure-appservice/api/v3/index.json" /> <!-- Draft PR,  remove after WebJobs SDK is released-->
```
chiangvincent, line 13: Can be removed

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusScaleMonitor.cs
alrod: This is for performance, "Azure SDK for NET" has a rule that all awaits should be executed with `ConfigureAwait(false)` so we do not care about synchronization context (`SynchronizationContext.Current = null`)

NuGet.Config
alrod: we still need this to build the PR -  Microsoft.Azure.WebJobs 3.0.35 haven't been released to official nuget yet. We can release 3.0.35 only after function host release with reference to 3.0.35.
This will be removed in future.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusMetricsProvider.cs
```diff
@@ -0,0 +1,231 @@
...
+                else
+                {
+                    ServiceBusEntityPathHelper.ParseTopicAndSubscription(_entityPath, out string topicPath, out string subscriptionPath);
+
+                    SubscriptionRuntimeProperties subscriptionProperties = await _administrationClient.Value.GetSubscriptionRuntimePropertiesAsync(topicPath, subscriptionPath).ConfigureAwait(false);
+                    activeMessageCount = subscriptionProperties.ActiveMessageCount;
+                    deadLetterCount = subscriptionProperties.DeadLetterMessageCount;
+                }
+            }
+            catch (UnauthorizedAccessException ex) // When manage claim is not used on Service Bus connection string
```
mathewc, line 65: I think from this method you should just log a warning like we did before then rethrow the UAE, and have the target scaler impl. catch that and throw NSE. That keeps this method general - it could be used by other code as well, with exceptions being handled as needed.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusMetricsProvider.cs
```diff
@@ -0,0 +1,231 @@
...
+            {
+                string message = $"Connection string does not have 'Manage Claim' for {entityName} '{_entityPath}'. Target base scale won't work due impossibility to get active message count.";
```
mathewc, line 67: Remove TBS specific wording here since this is a general method. E.g. "Connection string does not have 'Manage Claim' for {entityName} '{_entityPath}'. Unable to determine active message count."

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusMetricsProvider.cs
```diff
@@ -0,0 +1,238 @@
...
+            catch (Exception e)
+            {
+                _logger.LogWarning($"Error querying for Service Bus {entityName} scale status: {e.Message}");
+            }
+
+            // Path for connection strings with no manage claim
+            return CreateTriggerMetrics(activeMessage, 0, 0, 0, _isListeningOnDeadLetterQueue);
+        }
+
+        private static bool MessageIsActive(ServiceBusReceivedMessage message)
```
jsquire, line 157: I'm not sure what this indirection provides us since the comparison is just a single equality check.   Are we thinking there will be additional logic for this in the future?

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusMetricsProvider.cs
```diff
@@ -0,0 +1,238 @@
...
+            QueueProperties queueProperties;
+            long activeMessageCount = 0, deadLetterCount = 0;
+            int partitionCount = 0;
+
+            queueRuntimeProperties = await _administrationClient.Value.GetQueueRuntimePropertiesAsync(_entityPath).ConfigureAwait(false);
+            activeMessageCount = queueRuntimeProperties.ActiveMessageCount;
+            deadLetterCount = queueRuntimeProperties.DeadLetterMessageCount;
+
+            // If partitioning is turned on, then Service Bus automatically partitions queues into 16 partitions
+            // See more information here: https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-partitioning#standard
```
jsquire, line 174:
```suggestion
            // See more information here: https://docs.microsoft.com/azure/service-bus-messaging/service-bus-partitioning#standard
```

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusMetricsProvider.cs
```diff
@@ -0,0 +1,238 @@
...
+            int partitionCount = 0;
+
+            ServiceBusEntityPathHelper.ParseTopicAndSubscription(_entityPath, out topicPath, out subscriptionPath);
+
+            subscriptionProperties = await _administrationClient.Value.GetSubscriptionRuntimePropertiesAsync(topicPath, subscriptionPath).ConfigureAwait(false);
+            activeMessageCount = subscriptionProperties.ActiveMessageCount;
+            deadLetterCount = subscriptionProperties.DeadLetterMessageCount;
+
+            // If partitioning is turned on, then Service Bus automatically partitions queues into 16 partitions
+            // See more information here: https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-partitioning#standard
```
jsquire, line 196:
```suggestion
            // See more information here: https://docs.microsoft.com/azure/service-bus-messaging/service-bus-partitioning#standard
```

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusMetricsProvider.cs
```diff
@@ -0,0 +1,238 @@
...
+            long activeMessageCount = 0, deadLetterCount = 0;
+            int partitionCount = 0;
+
+            queueRuntimeProperties = await _administrationClient.Value.GetQueueRuntimePropertiesAsync(_entityPath).ConfigureAwait(false);
+            activeMessageCount = queueRuntimeProperties.ActiveMessageCount;
+            deadLetterCount = queueRuntimeProperties.DeadLetterMessageCount;
+
+            // If partitioning is turned on, then Service Bus automatically partitions queues into 16 partitions
+            // See more information here: https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-partitioning#standard
+            queueProperties = await _administrationClient.Value.GetQueueAsync(_entityPath).ConfigureAwait(false);
```
jsquire, line 175: @JoshLove-msft :  Do you know of a way where we can remove the hardcoded 16 for this?

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusMetricsProvider.cs
```diff
@@ -0,0 +1,238 @@
...
+using Microsoft.Extensions.Logging;
+using Azure.Messaging.ServiceBus;
+using Azure.Messaging.ServiceBus.Administration;
+using Microsoft.Azure.WebJobs.Extensions.ServiceBus.Config;
+
+namespace Microsoft.Azure.WebJobs.Extensions.ServiceBus.Listeners
+{
+    internal class ServiceBusMetricsProvider
+    {
+        private const string DeadLetterQueuePath = @"/$DeadLetterQueue";
```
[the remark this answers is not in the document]
JoshLove-msft: Filed https://github.com/Azure/azure-sdk-for-net/issues/34034

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusMetricsProvider.cs
alrod: removed the method `MessageIsActive`

jsquire (approved):

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusMetricsProvider.cs
JoshLove-msft: It doesn't look like this is exposed in the Admin API, but 16 only applies to basic and standard. For premium, it is configurable on namespace creation.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusMetricsProvider.cs
```diff
@@ -0,0 +1,236 @@
...
+                }
+            }
+            catch (ServiceBusException ex)
+            when (ex.Reason == ServiceBusFailureReason.MessagingEntityNotFound)
+            {
+                _logger.LogWarning($"ServiceBus {entityName} '{_entityPath}' was not found.");
+            }
+            catch (UnauthorizedAccessException ex)
+            {
+                _logger.LogWarning($"Connection string does not have 'Manage Claim' for {entityName} '{_entityPath}'.Unable to determine active message count.", ex);
```
JoshLove-msft, line 73:
```suggestion
                _logger.LogWarning($"Connection string does not have 'Manage Claim' for {entityName} '{_entityPath}'. Unable to determine active message count.", ex);
```

JoshLove-msft: Don't we need the TimeToLog guard here?

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusTargetScaler.cs
```diff
@@ -0,0 +1,106 @@
...
+            {
+                if (!_singleDispatch)
+                {
+                    concurrency = _options.MaxMessageBatchSize;
+                }
+                else
+                {
+                    if (_isSessionsEnabled)
+                    {
+                        concurrency = _options.MaxConcurrentSessions;
```
JoshLove-msft, line 79: Do we care about MaxCallsPerSession?

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusTargetScaler.cs
```diff
@@ -0,0 +1,106 @@
...
+                concurrency = context.InstanceConcurrency.Value;
+            }
+
+            if (concurrency < 1)
+            {
+                throw new ArgumentOutOfRangeException($"Unexpected concurrency='{concurrency}' - the value must be > 0.");
+            }
+
+            int targetWorkerCount = (int)Math.Ceiling(messageCount / (decimal)concurrency);
+            _logger.LogInformation($"'Target worker count for function '{_functionId}' is '{targetWorkerCount}' (EntityPath='{_entityPath}', MessageCount ='{messageCount}', Concurrency='{concurrency}').");
```
JoshLove-msft, line 98:
```suggestion
            _logger.LogInformation($"Target worker count for function '{_functionId}' is '{targetWorkerCount}' (EntityPath='{_entityPath}', MessageCount ='{messageCount}', Concurrency='{concurrency}').");
```

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusTargetScaler.cs
JoshLove-msft: Nevermind - not configurable in the extension.

PR #31398, opened 2022-09-23
Rename `ExpiresOn` to `ExpireOn`

ArthurMa1978:


sdk/storage/Azure.ResourceManager.Storage/src/Customize/Models/StorageTableAccessPolicy.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        /// <summary> The deleted date. </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public DateTimeOffset? ExpiresOn {
+            get {
+                return ExpireOn;
+            }
+            set {
+                ExpireOn = value;
+            }
```
ArcturusZhang, line 22: This piece of code need formatting

ArcturusZhang (approved):

sdk/storage/Azure.ResourceManager.Storage/src/Customize/Models/StorageTableAccessPolicy.cs
ArthurMa1978: fixed

PR #31456, opened 2022-09-27
Support binding to POCO when message body uses AMQP value

### LINKED ISSUE #31440: [Service Bus Extension] Binding to POCO does not handle non-standard message bodies
Similar to how the string binding logic will [inspect the message body](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Triggers/MessageToStringConverter.cs#L53) to determine whether it is a value body, we need to update the [POCO binding logic](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Triggers/MessageToPocoConverter.cs) to have the same handling.

JoshLove-msft:
Fixes https://github.com/Azure/azure-sdk-for-net/issues/31440

### DIFF
--- a/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Triggers/MessageToPocoConverter.cs
+++ b/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Triggers/MessageToPocoConverter.cs
@@ -5,6 +5,7 @@ using System;
 using System.Runtime.Serialization;
 using Newtonsoft.Json;
 using System.Globalization;
+using Azure.Core.Amqp;
 using Azure.Messaging.ServiceBus;
 using Microsoft.Azure.WebJobs.Extensions.ServiceBus.Triggers;
 using Newtonsoft.Json.Linq;
@@ -24,7 +25,7 @@ namespace Microsoft.Azure.WebJobs.ServiceBus.Triggers
         public TElement Convert(ServiceBusReceivedMessage message)
         {
             // 1. If ContentType is "application/json" or binding to a JObject deserialize as JSON
-            // 2. If ContentType is not "application/json" attempt to deserialize using Message.GetBody, which will handle cases like XML object serialization
+            // 2. If ContentType is not "application/json" and it is a Data body message, attempt to deserialize using DataContractSerializer, which will handle cases like XML object serialization
             // 3. If this deserialization fails, do a final attempt at JSON deserialization to catch cases where the content type might be incorrect
 
             if (message.ContentType == ContentTypes.ApplicationJson ||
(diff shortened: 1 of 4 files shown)


sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Triggers/MessageToPocoConverter.cs
```diff
@@ -24,31 +25,56 @@ public MessageToPocoConverter(JsonSerializerSettings jsonSerializerSettings)
         public TElement Convert(ServiceBusReceivedMessage message)
         {
             // 1. If ContentType is "application/json" or binding to a JObject deserialize as JSON
-            // 2. If ContentType is not "application/json" attempt to deserialize using Message.GetBody, which will handle cases like XML object serialization
+            // 2. If ContentType is not "application/json" and it is a Data body message, attempt to deserialize using DataContractSerializer, which will handle cases like XML object serialization
             // 3. If this deserialization fails, do a final attempt at JSON deserialization to catch cases where the content type might be incorrect
 
             if (message.ContentType == ContentTypes.ApplicationJson ||
```
jsquire, line 31: What happens if the POCO is `byte[]` or another form that implies "just give me the message body bytes"?

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Triggers/MessageToPocoConverter.cs
```diff
@@ -61,7 +87,7 @@ private TElement DeserializeJsonObject(ServiceBusReceivedMessage message)
                     CultureInfo.InvariantCulture,
 @"Binding parameters to complex objects (such as '{0}') uses Json.NET serialization or XML object serialization. 
  1. If ContentType is 'application/json' deserialize as JSON
- 2. If ContentType is not 'application/json' attempt to deserialize using Message.GetBody, which will handle cases like XML object serialization
+ 2. If ContentType is not 'application/json' and it is a Data body message, attempt to deserialize using DataContractSerializer, which will handle cases like XML object serialization
```
jsquire, line 90: Should we consider linking out to something AMQP here?  I suspect "data body message" isn't going to mean something to a lot of people.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Triggers/MessageToPocoConverter.cs
JoshLove-msft: Binding to a byte array would not result in the Poco converter being executed. Instead, it is handled here - https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Triggers/MessageToByteArrayConverter.cs

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Triggers/MessageToPocoConverter.cs
JoshLove-msft: I think we should add this scenario to our troubleshooting guide and link to that.

jsquire (approved):

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Triggers/MessageToPocoConverter.cs
JoshLove-msft: Will need to do a separate PR once the updated TSG is merged.

PR #31474, opened 2022-09-27
Enrich additional exception messages

JoshLove-msft:


m-redding (approved):
These changes look good from my perspective

PR #31485, opened 2022-09-28
Improvement for Automated testing

minwoolee-msft:
1. Adding more tests
2. Added all related recorded test
3. Updated assert method to be more readable 
4. Updated namespaces for all tests


sdk/communication/Azure.Communication.CallAutomation/tests/CallAutomationClients/CallAutomationClientAutomatedLiveTests.cs
```diff
@@ -0,0 +1,149 @@
...
+
+namespace Azure.Communication.CallAutomation.Tests.CallAutomationClients
+{
+    internal class CallAutomationClientAutomatedLiveTests : CallAutomationClientAutomatedLiveTestsBase
+    {
+        public CallAutomationClientAutomatedLiveTests(bool isAsync) : base(isAsync)
+        {
+        }
+
+        [RecordedTest]
```
yyw-msft, line 25: curious, what does this [RecordedTest] do?

yyw-msft (approved):

sdk/communication/Azure.Communication.CallAutomation/tests/CallAutomationClients/CallAutomationClientAutomatedLiveTests.cs
minwoolee-msft: Similar to test - but this allows it to be re-recorded when recording is not present. Description says:

This attribute replaces the [Test] attribute and will dynamically re-record recorded tests on failure when run in Playback mode.
Tests that are re-recorded will complete with a error status and indicate that copying the updated recording to SessionRecords is needed.

PR #31503, opened 2022-09-28
Add a DataFactoryExpression object

### LINKED ISSUE #31507: [FEATURE REQ] Add sdk support for DataFactoryExpression
### Library name

This applies to management plane and dataplane DataFactory and Synapse libraries.

## Summary
In the datafactory API many of the properties have the ability to either be a constant value or an expression which will be evaluated at runtime.  For more information on expressions go [here](https://learn.microsoft.com/en-us/azure/data-factory/control-flow-expression-language-functions#expressions).  The structure of an expression is different that a constant value for example the [FolderPath](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/datafactory/resource-manager/Microsoft.DataFactory/stable/2018-06-01/entityTypes/Dataset.json#L1353) property of an AzureBlobDataset can either be a "string (or Expression with resultType string)". 

#### Constant
```json
"folderPath": "foo/bar"
```
#### Expression
```json
"folderpath": {
  "type": "Expression",
  "value": "foo/bar-@{ pipeline().TriggerTime}"
}
```
In this example when the pipeline is run in the first case the folder is always `foo/bar`, but in the second case the service will append the time the pipeline kicked off to the folder name.

Because the property can have multiple types the only real way to express this in openapi is to set it to an object which gets generated as `object` for dotnet dataplane and `BinaryData` for mgmtplane.  This makes it difficult for customers to know how they are supposed to construct these.

#### Constant
```c#
  azureBlobDataset.FolderPath = "foo/bar";
```
#### Expression
```c#
  Dictionary<string, object> properties = new Dictionary<string, object>();
  properties.Add("type", "Expression");
  properties.Add("value", "foo/bar-@{ pipeline().TriggerTime}");
  azureBlobDataset.FolderPath = properties;
```

#### Constant
```c#
  azureBlobDataset.FolderPath = BinaryData.FromObjectAsJson("foo/bar");
```
#### Expression
```c#
  azureBlobDataset.FolderPath = BinaryData.FromObjectAsJson(new { type = "Expression", value = "foo/bar-@{ pipeline().TriggerTime}"});
```

## Solution
The main challenge from a client side is that we need an object that can dynamically serialize and deserialize a primitive or an object based on its current value.  We also should restrict the primitive type in strongly typed languages that can be assigned to the property since because its an object I can easily assign an integer to a field that expects a string.

To solve for this we should introduce a new type `DataFactoryExpression<T>` which defines which type can be assigned as well as which type the expression must evaluate to aligning with the documentation of the API but moving it from documentation to actual API shape.

This would allow us to use the exact same code whether we are assigning a constant or an expression
#### Constant
```c#
  azureBlobDataset.FolderPath = new DataFactoryExpression<string>("foo/bar");
```
#### Expresion
```c#
  azureBlobDataset.FolderPath = new DataFactoryExpression<string>("foo/bar-@{ pipeline().TriggerTime}");
```

There a lot of considerations on the API shape of `DataFactoryExpression<T>` that need to be worked out, but conceptually this would allow us to provide a much more user friendly SDK for libraries like DataFactory and Synapse who use these expression objects.

m-nash:
Fixes https://github.com/Azure/azure-sdk-for-net/issues/31507


sdk/datafactory/Azure.ResourceManager.DataFactory/src/Azure.ResourceManager.DataFactory.csproj
```diff
@@ -5,4 +5,7 @@
     <Description>Azure Resource Manager client SDK for Azure resource provider DataFactory</Description>
     <PackageTags>azure;management;arm;resource manager;datafactory</PackageTags>
   </PropertyGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\..\..\core\Azure.Core.Expressions.DataFactory\src\Azure.Core.Expressions.DataFactory.csproj" />
```
m-nash, line 9: should be able to make this a package ref, and probably need to add the version to packages.data.props

sdk/datafactory/Azure.ResourceManager.DataFactory/src/autorest.md
```diff
@@ -225,7 +225,7 @@ rename-mapping:
   PurviewConfiguration: FactoryPurviewConfiguration
   RunFilterParameters: RunFilterContent
   SecretBase: FactorySecretBaseDefinition
-  SecureString: FactorySecretString
+  SecureString: FactorySecureString
```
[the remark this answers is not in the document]
JoshLove-msft: Would it be just FactorySecretString in that case? All of the other types are prefixed with Factory.

sdk/datafactory/Azure.ResourceManager.DataFactory/src/Azure.ResourceManager.DataFactory.csproj
JoshLove-msft: Generally we use project refs rather than put alpha versions in the packages.data.props file.

sdk/datafactory/Azure.ResourceManager.DataFactory/src/autorest.md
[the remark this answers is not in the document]
m-nash: I see that we are using Factory in other places.  Lets track what prefix we want to use in another issue, this PR is fine to use `Factory` since that seems to be the "current prefix" choice.

sdk/datafactory/Azure.ResourceManager.DataFactory/src/Azure.ResourceManager.DataFactory.csproj
m-nash: this won't get checked in until we at least beta the new library, so I always add it in the way it would be in the end just to validate all the structure / ci stuff works appropriately.

JoshLove-msft: > this won't get checked in until we at least beta the new library

Well that wouldn't be the case if we used a project reference 😄

PR #31511, opened 2022-09-28
[Azure.Monitor.Ingestion] Restore Protocol methods to public

nisha-bhatia:


nisha-bhatia:
Chatted with LLC folks - we will GA with public Protocol (DPG) methods

sdk/monitor/Azure.Monitor.Ingestion/tests/MonitorIngestionLiveTest.cs
```diff
@@ -175,5 +175,33 @@ public void ConcurrencySingleThread()
             Assert.AreEqual(UploadLogsStatus.Success, response.Value.Status);
             Assert.IsEmpty(response.Value.Errors);
         }
+
+        [Test]
+        public async Task ValidInputFromObjectAsJsonNoBatchingAsync()
```
JoshLove-msft, line 180: Do we also need a test that passes in some other content-encoding now that we are exposing that parameter?

sdk/monitor/Azure.Monitor.Ingestion/src/Generated/LogsIngestionClient.cs
```diff
@@ -153,7 +153,7 @@ internal virtual async Task<Response> UploadAsync(string ruleId, string streamNa
         /// ]]></code>
         /// </example>
         /// <remarks> See error response code and error response message for more detail. </remarks>
-        internal virtual Response Upload(string ruleId, string streamName, RequestContent content, string contentEncoding = "gzip", RequestContext context = null)
+        public virtual Response Upload(string ruleId, string streamName, RequestContent content, string contentEncoding = "gzip", RequestContext context = null)
```
JoshLove-msft, line 156: Do we Gzip the user content here? Or is this meant to be a method that will just take exactly what the user provides and send it to the service. If that is the case, I'm not sure that it makes sense to default the contentEncoding parameter as "gzip".

sdk/monitor/Azure.Monitor.Ingestion/CHANGELOG.md
```diff
@@ -1,14 +1,8 @@
...
 ## 1.0.0-beta.4 (Unreleased)
-
-### Features Added
-
-### Breaking Changes
-
-### Bugs Fixed
-
-### Other Changes
+## Features Added
```
jsquire, line 4:
```suggestion

## Features Added
```

sdk/monitor/ci.yml
```diff
@@ -13,7 +13,7 @@ trigger:
     - sdk/monitor/service.projects
     - sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter
     - sdk/monitor/Azure.Monitor.Query
-    - sdk/monitor/AzureMonitorIngestion
+    - sdk/monitor/Azure.Monitor.Ingestion
```
jsquire, line 16: This is also needed on L31

sdk/monitor/Azure.Monitor.Ingestion/src/Generated/LogsIngestionClient.cs
nisha-bhatia: We are trying to gzip by default, so any input provided by the user will be gzipped

nisha-bhatia: request gets constructed on 109

JoshLove-msft: Got it! So this method will still GZip for users based on the content-encoding.

sdk/monitor/Azure.Monitor.Ingestion/tests/MonitorIngestionLiveTest.cs
```diff
@@ -39,7 +39,7 @@ private LogsIngestionClient CreateClient(HttpPipelinePolicy policy = null)
         public void NullInput()
         {
             LogsIngestionClient client = CreateClient();
-            Assert.ThrowsAsync<ArgumentNullException>(async () => await client.UploadAsync(TestEnvironment.DCRImmutableId, TestEnvironment.StreamName, null).ConfigureAwait(false));
+            Assert.ThrowsAsync<NullReferenceException>(async () => await client.UploadAsync(TestEnvironment.DCRImmutableId, TestEnvironment.StreamName, null).ConfigureAwait(false));
```
JoshLove-msft, line 42: We shouldn't need to change this test - this suggests that we are missing some null validation now that we are not calling the generated method.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -403,5 +403,115 @@ private static UploadLogsStatus GetStatus<T>(IEnumerable<T> logEntries, List<Upl
...
+        /// var data = new[] {
+        ///     new {}
+        /// };
+        ///
+        /// Response response = await client.UploadAsync("<ruleId>", "<streamName>", RequestContent.Create(data), <gzip>);
+        /// Console.WriteLine(response.Status);
+        /// ]]></code>
+        /// </example>
+        /// <remarks> See error response code and error response message for more detail. </remarks>
+        public virtual async Task<Response> UploadAsync(string ruleId, string streamName, RequestContent content, RequestContext context = null, CancellationToken cancellationToken = default)
```
JoshLove-msft, line 446: Do we need to take in the CancellationToken as a parameter? I think it can be set on the RequestContext. We should try to match what the DPG generator does.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -15,6 +15,9 @@ namespace Azure.Monitor.Ingestion
 {
     /// <summary> The IngestionUsingDataCollectionRules service client. </summary>
     [CodeGenClient("IngestionUsingDataCollectionRulesClient")]
+    [CodeGenSuppress("Upload", typeof(string), typeof(string), typeof(RequestContent), typeof(string), typeof(RequestContext))]
+    [CodeGenSuppress("UploadAsync", typeof(string), typeof(string), typeof(RequestContent), typeof(string), typeof(RequestContext))]
+    [CodeGenSuppress("CreateUploadRequest", typeof(string), typeof(string), typeof(RequestContent), typeof(string), typeof(RequestContext))]
```
JoshLove-msft, line 20: nit: I think this one is probably unnecessary as it shouldn't be included if the first two are excluded.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
nisha-bhatia: yes - it can be set in the RequestContext, I'll remove it from the DPG generated methods

jsquire (approved):

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -403,5 +405,121 @@ private static UploadLogsStatus GetStatus<T>(IEnumerable<T> logEntries, List<Upl
...
+        /// var data = new[] {
+        ///     new {}
+        /// };
+        ///
+        /// Response response = await client.UploadAsync("<ruleId>", "<streamName>", RequestContent.Create(data), <gzip>);
+        /// Console.WriteLine(response.Status);
+        /// ]]></code>
+        /// </example>
+        /// <remarks> See error response code and error response message for more detail. </remarks>
+        public virtual async Task<Response> UploadAsync(string ruleId, string streamName, RequestContent content, RequestContext context = null)
```
JoshLove-msft, line 447: nit: it might be worth creating a separate private method that has all the logic and having the sync/async public methods call into it.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -403,5 +403,121 @@ private static UploadLogsStatus GetStatus<T>(IEnumerable<T> logEntries, List<Upl
...
+        /// var data = new[] {
+        ///     new {}
+        /// };
+        ///
+        /// Response response = client.Upload("<ruleId>", "<streamName>", RequestContent.Create(data), <gzip>);
+        /// Console.WriteLine(response.Status);
+        /// ]]></code>
+        /// </example>
+        /// <remarks> See error response code and error response message for more detail. </remarks>
+        internal virtual Response Upload(string ruleId, string streamName, RequestContent content, RequestContext context = null)
```
JoshLove-msft, line 503: Is this method used anywhere? Maybe this can be the internal method that the public methods call into.

nisha-bhatia: this method is supposed to be public

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -403,5 +403,118 @@ private static UploadLogsStatus GetStatus<T>(IEnumerable<T> logEntries, List<Upl
...
+        /// };
+        ///
+        /// Response response = client.Upload("<ruleId>", "<streamName>", RequestContent.Create(data), <gzip>);
+        /// Console.WriteLine(response.Status);
+        /// ]]></code>
+        /// </example>
+        /// <remarks> See error response code and error response message for more detail. </remarks>
+        public virtual Response Upload(string ruleId, string streamName, RequestContent content, RequestContext context = null)
+        {
+            return UploadRequestContentAsync(ruleId, streamName, content, false, context).Result;
```
JoshLove-msft, line 490:
```suggestion
            return UploadRequestContentAsync(ruleId, streamName, content, false, context).EnsureCompleted();
```

JoshLove-msft (approved):

PR #31512, opened 2022-09-28
Add Uri property to TableClient and TableServiceClient

### LINKED ISSUE #24415: [FEATURE REQ] Add `Uri` property on `TableClient` and `TableServiceClient`
Add `TableServiceClient.Uri` property
   - This is helpful for diagnostics, for example if you have the wrong cloud endpoint suffix provided
   - `BlobServiceClient` has this
   - `QueueServiceClient` has this

(extracted from #19881)

christothes:
closes #24415

### DIFF
--- a/sdk/tables/Azure.Data.Tables/src/TableClient.cs
+++ b/sdk/tables/Azure.Data.Tables/src/TableClient.cs
@@ -71,6 +71,11 @@ namespace Azure.Data.Tables
         }
 
         /// <summary>
+        /// The Uri of the Table.
+        /// </summary>
+        public virtual Uri Uri { get; }
+
+        /// <summary>
         /// Initializes a new instance of the <see cref="TableClient"/> using the specified <see cref="Uri" /> which contains a SAS token.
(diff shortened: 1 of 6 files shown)


sdk/tables/Azure.Data.Tables/src/TableClient.cs
```diff
@@ -70,6 +70,11 @@ public virtual string AccountName
             }
         }
 
+        /// <summary>
+        /// The Uri of the Table.
```
jsquire, line 74: Is Table intended to be capitalized?   Seems the rest of the comments use it with lower case.

sdk/tables/Azure.Data.Tables/src/TableServiceClient.cs
```diff
@@ -59,6 +59,11 @@ public virtual string AccountName
             }
         }
 
+        /// <summary>
+        /// The Uri for the Table account.
+        /// </summary>
+        public virtual Uri Uri { get; }
```
jsquire, line 65: Same question about capitalization.

jsquire (approved):

PR #31534, opened 2022-09-30
Use correct default ActivityKind

### LINKED ISSUE #30816: Logical spans have client kind that results in broken experience in AzureMonitor
### Library name and version

Azure.Core

### Describe the bug

https://github.com/Azure/azure-sdk-for-net/blob/13a52c41cd4c085019a96122a00c5016cbc0166b/sdk/core/Azure.Core/src/Shared/DiagnosticScopeFactory.cs#L44

By default, all logical spans are created with `Client`, kind and (at least most) client do not change the default.
So what happens is:
- KeyVault creates  `SecretClient.GetSecret` with `client` kind
  - `RequestActivityPolicy` creates another `client` HTTP span

Azure Monitor (via ApplicationInsights SDK or OTel) use kind to distinguish logical and physical operation, that's essential part of the Application Map. Not having a distinction results in the following: `GetSecret` is a node on ApplicationMap, while HTTP `*.vault.azure.net` is another node.

image

Apparently, this issue was there from the beginning, and changing it now would be breaking.

**I suggest the following:  we planned to do a number of such small breaking changes at some point and we batch this one along with others (I'll start listing them).**

### Expected behavior

Logical operations should have `Internal` kind: it affects all REST-based SDKs. 

### Actual behavior

Logical operations have `Client` kind (checked Storage, KeyVault, and EventGrid)

### Reproduction Steps

- create simple KV app
- enable Application Insights SDK or OTel
- generate some telemetry
- check the appmap

### Environment

_No response_

JoshLove-msft:
Fixes https://github.com/Azure/azure-sdk-for-net/issues/30816

Before:
before

After:
after

### DIFF
--- a/sdk/core/Azure.Core/src/Shared/DiagnosticScopeFactory.cs
+++ b/sdk/core/Azure.Core/src/Shared/DiagnosticScopeFactory.cs
@@ -41,7 +41,7 @@ namespace Azure.Core.Pipeline
 
         public bool IsActivityEnabled { get; }
 
-        public DiagnosticScope CreateScope(string name, DiagnosticScope.ActivityKind kind = DiagnosticScope.ActivityKind.Client)
+        public DiagnosticScope CreateScope(string name, DiagnosticScope.ActivityKind kind = DiagnosticScope.ActivityKind.Internal)
         {
             if (_source == null)
             {
--- a/sdk/core/Azure.Core/tests/ClientDiagnosticsTests.cs
+++ b/sdk/core/Azure.Core/tests/ClientDiagnosticsTests.cs
@@ -35,6 +35,12 @@ namespace Azure.Core.Tests
 
             Activity activity = Activity.Current;
 
+#if NET5_0_OR_GREATER
+            // ActivityKind is only present in .NET 5.0+
+            // Validate that the default activity kind is used
+            Assert.AreEqual(ActivityKind.Internal, activity.Kind);
+#endif
+
             scope.Dispose();
 
             (string Key, object Value, DiagnosticListener) stopEvent = testListener.Events.Dequeue();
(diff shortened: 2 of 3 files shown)


sdk/core/Azure.Core/src/Shared/DiagnosticScopeFactory.cs
```diff
@@ -41,7 +41,7 @@ public DiagnosticScopeFactory(string clientNamespace, string? resourceProviderNa
 
         public bool IsActivityEnabled { get; }
 
-        public DiagnosticScope CreateScope(string name, DiagnosticScope.ActivityKind kind = DiagnosticScope.ActivityKind.Client)
+        public DiagnosticScope CreateScope(string name, DiagnosticScope.ActivityKind kind = DiagnosticScope.ActivityKind.Internal)
```
JoshLove-msft, line 44: We probably need to validate that all non-HTTP services are specifying a value for Kind already.

JoshLove-msft: I did a quick scan of SB and EH and that appears to be the case.
@jsquire can you double check my work?

sdk/core/Azure.Core/tests/ClientDiagnosticsTests.cs
```diff
@@ -35,6 +35,12 @@ public void CreatesActivityWithNameAndTags()
 
             Activity activity = Activity.Current;
 
+#if NET5_0_OR_GREATER
```
lmolkova, line 38: we have 5.0-specific tests, would you mind moving it there?

lmolkova: could you also check the kind attriubute value for non-5.0 path?

lmolkova (approved):
LGTM, thank you @JoshLove-msft !

sdk/core/Azure.Core/src/Shared/DiagnosticScopeFactory.cs
jsquire: I double-checked and I believe SB and EH are good.

jsquire (approved):

christothes (approved):

PR #31566, opened 2022-09-30
Update mgmt pipelines to add PR triggers 

weshaggard:
This adds back the PR triggers for the mgmt pipelines but also limits the amount of work a PR triggered pipeline will cause to just build and analyze jobs. The expectation is that testing will be covered by another pipeline, in the mgmt case it is covered in the resourcemanager pipeline.


weshaggard:
/azp net - websites - mgmt - ci

eng/scripts/Update-Mgmt-CI.ps1
```diff
@@ -9,38 +9,41 @@ function Update-CIFile() {
...
-        if($shouldRemove) {
-            if($line.StartsWith(" ")) {
-                Continue
-            }
-            $shouldRemove = $false
-        }
+    $content = Get-Content $mgmtCiFile -Raw
 
-        $newLines.Add($line) | Out-Null
+    if ($content -match "(?s)ServiceDirectory: (?<sd>[^\r\n]+).*- name: (?<p>[^\r\n]+)")
```
benbp, line 14: I assume the templating makes this consistent, but is it worth handling whitespace more flexibly (e.g. if someone adds two spaces)?

eng/scripts/Update-Mgmt-CI.ps1
```diff
@@ -9,38 +9,41 @@ function Update-CIFile() {
...
+    {
+        $serviceDirectory = $matches["sd"]
+        $packageName = $matches["p"]
+    }
+    else {
+        Write-Error "Could not parse out the service directory and package from $mgmtCiFile, so skipping."
```
benbp, line 20: Do we have continueOnError or should this be warning?

eng/scripts/Update-Mgmt-CI.ps1
```diff
@@ -9,38 +9,41 @@ function Update-CIFile() {
...
+    - hotfix/*
+    - release/*
+  paths:
+    include:
+    - sdk/$serviceDirectory/$packageName
+
+"@
+
+    $content = $content -replace "(?s)pr:[^\n]*(\n([ ]+[^\n]*|))*", $prtriggers
+    $content = $content -replace "(?s)trigger:[^\n]*(\n([ ]+[^\n]*|))*", "trigger: none`r`n"
```
benbp, line 39: Can we not inject a carriage return on the output?

benbp (approved):

eng/scripts/Update-Mgmt-CI.ps1
```diff
@@ -9,38 +9,41 @@ function Update-CIFile() {
...
+pr:
+  branches:
+    include:
+    - main
+    - feature/*
+    - hotfix/*
+    - release/*
+  paths:
+    include:
+    - sdk/$serviceDirectory/$packageName
```
m-nash, line 34: should we add another trigger here for `sdk/$serviceDirectory/ci.mgmt.yml` so that the ci runs if someone changes its definition?

eng/scripts/Update-Mgmt-CI.ps1
weshaggard: These regex's are definitely hacky but right now there isn't any good yml parser we can use to maintain formatting. For these consistency should be kept by the fact that these are mostly generated. That said I probably should add some \s+ instead of hardcoding exacts. The stupid multiline matching is kind of a pain though because \s also matches newlines.

eng/scripts/Update-Mgmt-CI.ps1
weshaggard: Fair I can include that.

eng/scripts/Update-Mgmt-CI.ps1
weshaggard: this is a raw string with newlines included so we need to directly inject it here.

eng/scripts/Update-Mgmt-CI.ps1
weshaggard: If we hit this then I want to fail fast and let folks fix it.

weshaggard:
Closing PR now and pushing the change directly as most of the changes cannot be validated in PR because they update path triggers.

Issue #31579, opened 2022-10-03
[FEATURE REQ] Add ability to register ArmClient to DI via AddAzureClients


Stolpe:
### Library name

Azure.ResourceManager

### Please describe the feature.

I would like to be able to register an `ArmClient` via `AddAzureClients` from Microsoft.Extensions.Azure.

The reason I want to do this is to be able to create my `TokenCredential` from configuration, an approach I use in order to be able to use different credential types locally and when deployed to Azure, without any code changes. Also, it helps a lot to be able to mock my client and insert it to the constructor of what ever service I'm currently unit testing.

jsquire:
Thank you for your feedback.  Tagging and routing to the team member best able to assist.

xboxeer:
@Stolpe , you can use the below code to register ArmClient in DI
```csharp
using IHost host = Host.CreateDefaultBuilder(args)
   .ConfigureServices((context,services) =>
   {
       services.AddAzureClients(builder =>
       {
           
           builder.AddClient<ArmClient, ArmClientOptions>(options =>
           {
               options.Environment = ArmEnvironment.AzurePublicCloud;
               options.Diagnostics.IsLoggingContentEnabled=true;
               options.Diagnostics.IsLoggingEnabled=true;
               var client = new ArmClient(new DefaultAzureCredential(
                   new DefaultAzureCredentialOptions() 
                   { VisualStudioTenantId=context.Configuration.GetValue<string>("VisualStudioTenantId") }),
                   context.Configuration.GetValue<string>("DefaultSubscriptionId"),
                   options);
               
               return client;                
           });

       });
       services.AddScoped<VirtualMachineCollection>((sp) =>
       {
           var armClient = sp.GetService<ArmClient>();
           return armClient?.GetDefaultSubscription().GetResourceGroup(RESOURCE_GROUP_NAME).Value.GetVirtualMachines();
       });
}
```
I have some more comments regarding the scenario you are sharing in this issue
`an approach I use in order to be able to use different credential types locally and when deployed to Azure, without any code changes` -> You can actually use DefaultAzureCredential in your code 
The order and locations in which DefaultAzureCredential looks for credentials is shown in the diagram and table below.
![image](https://user-images.githubusercontent.com/5089150/194686058-2074c51e-a056-471e-a667-71dbfab13dcf.png)

` it helps a lot to be able to mock my client and insert it to the constructor of what ever service I'm currently unit testing.` ->
[...]

PR #31580, opened 2022-10-03
Add extension methods and constructor overload to ArmClient

Stolpe:

Solves #31579 by adding the necessary extension methods and constructor overloads.


Stolpe:
Failed test seems unrelated to my change to me. Noisy neighbor perhaps?
https://dev.azure.com/azure-sdk/public/_build/results?buildId=1891733&view=logs&j=5448e786-65d1-5f78-4134-a73e0ac1f58f&t=f734e3bd-0b62-5d2c-efbc-0aabc79136bc&l=3859

sdk/resourcemanager/Azure.ResourceManager/src/ArmClient.cs
```diff
@@ -59,6 +59,16 @@ public ArmClient(TokenCredential credential, string defaultSubscriptionId) : thi
         {
         }
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ArmClient"/> class.
+        /// </summary>
+        /// <param name="credential"> A credential used to authenticate to an Azure Service. </param>
+        /// <param name="options"> The client parameters to use in these operations. </param>
+        /// <exception cref="ArgumentNullException"> If <see cref="TokenCredential"/> is null. </exception>
+        public ArmClient(TokenCredential credential, ArmClientOptions options) : this(credential, default, options)
```
m-nash, line 68: Lets remove this ctor and call one of the others that are provided.  I assume `ArmClient(TokenCredential credential, string defaultSubscriptionId, ArmClientOptions options)` will get used.  The `defaultSubscriptionId` can be left null, but if passed in it saves 1 API call when later using `armClient.GetDefaultSubscription()` as this value is cached for the lifetime of the client instance.

m-nash (changes requested):

sdk/resourcemanager/Azure.ResourceManager/src/ArmClient.cs
Stolpe: Unfortunately, this constructor overload is required. Note that we are not calling the constructor ourselves, that is done by code inside Microsoft.Extensions.Azure. Here is that code: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientFactory.cs#L22

Take this simplest example. We expect our `ArmClient` to be created using a `DefaultAzureCredential` and default options.
```csharp
var configuration = new ConfigurationBuilder()
    .AddInMemoryCollection()
    .Build();
var services = new ServiceCollection();
services
    .AddAzureClients(builder =>
        builder.AddArmClient(configuration));
var client = services
    .BuildServiceProvider()
    .GetRequiredService<ArmClient>();
```
With this overload existing, the client is created just fine. Without it, we get the following exception (details from Visual Studio from a local scratch project):
```
System.InvalidOperationException
  HResult=0x80131509
  Message=Unable to find matching constructor while trying to create an instance of ArmClient.
Expected one of the follow sets of configuration parameters:
1., defaultSubscriptionId

Found the following configuration keys:
  Source=Microsoft.Extensions.Azure
  StackTrace:
   at Microsoft.Extensions.Azure.ClientFactory.CreateClient(Type clientType, Type optionsType, Object options, IConfiguration configuration, TokenCredential credential)
   at Microsoft.Extensions.Azure.AzureClientFactoryBuilder.<>c__DisplayClass4_0`2.<Azure.Core.Extensions.IAzureClientFactoryBuilderWithConfiguration<Microsoft.Extensions.Configuration.IConfiguration>.RegisterClientFactory>b__0(TOptions options, TokenCredential credentials)
   at Microsoft.Extensions.Azure.ClientRegistration`1.GetClient(IServiceProvider serviceProvider, Object options, TokenCredential tokenCredential)
   at Microsoft.Extensions.Azure.AzureClientFactory`2.CreateClient(String name)
   at Microsoft.Extensions.Azure.AzureClientFactoryBuilder.<>c__15`2.<RegisterClientFactory>b__15_1(IServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitRootCache(ServiceCallSite callSite, RuntimeResolverContext context)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteVisitor`2.VisitCallSite(ServiceCallSite callSite, TArgument argument)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.Resolve(ServiceCallSite callSite, ServiceProviderEngineScope scope)
   at Microsoft.Extensions.DependencyInjection.ServiceProvider.CreateServiceAccessor(Type serviceType)
   at System.Collections.Concurrent.ConcurrentDictionary`2.GetOrAdd(TKey key, Func`2 valueFactory)
   at Microsoft.Extensions.DependencyInjection.ServiceProvider.GetService(Type serviceType, ServiceProviderEngineScope serviceProviderEngineScope)
   at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService(IServiceProvider provider, Type serviceType)
   at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService[T](IServiceProvider provider)
   at Program.<<<Main>$>g__ResourceManagerTestaren|0_12>d.MoveNext() in C:\Users\Per.Stolpe\source\repos\ConsoleApp1\ConsoleApp1\Program.cs:line 687
   at Program.<<Main>$>d__0.MoveNext() in C:\Users\Per.Stolpe\source\repos\ConsoleApp1\ConsoleApp1\Program.cs:line 44
```

So, what is happening is that the internal `ClientFactory` from Microsoft.Extensions.Azure is looking for any constructor that has a parameter that takes an `ArmClientOptions`. Without this change, the only possible alternative is
```csharp
public ArmClient(TokenCredential credential, string defaultSubscriptionId, ArmClientOptions options)
```
The `credential` parameter is constructed using magic, but basically defaults into a `DefaultAzureCredential`, since it cannot find any relevant information about how it should construct it. This is just fine, and just as expected.

Now, it sees that we have no value for the parameter `defaultSubscriptionId`, and an exception is thrown. My overload is just this overload, but without the `defaultSubscriptionId` parameter. From eg.
```csharp
public ArmClient(TokenCredential credential) : this(credential, default, default)
```
we know that this value is not required.

Stolpe:
/reopen

I was away on Christmas holidays. I'm still awaiting a response to my reply on the requested changes.

Stolpe:
@microsoft-github-policy-service agree [company="Collector Bank AB"]

Stolpe:
@microsoft-github-policy-service agree company="Collector Bank AB"

ArcturusZhang:
@m-nash could you please take a look at the latest changes in this PR?

m-nash:
We might need a change to the underlying extensions library @joshfree is helping to take a look.

JoshLove-msft:
> We might need a change to the underlying extensions library @JoshLove-msft is helping to take a look.

I agree that it is not intuitive 

> We might need a change to the underlying extensions library @JoshLove-msft is helping to take a look.

So one way around needing the new ctor, would be to specify an empty defaultSubscriptionId in config, e.g.
```c#
  var configuration = new ConfigurationBuilder().AddInMemoryCollection(new KeyValuePair<string, string>[]
  {
      new KeyValuePair<string, string>("defaultSubscriptionId", "")
  }).Build();

  var services = new ServiceCollection();
  services.AddAzureClients(builder =>
  {
      builder.AddArmClient(configuration);
  });
  var client = services.BuildServiceProvider().GetRequiredService<ArmClient>();
```
This would cause the extension to use the ctor with the defaultSubscriptionId. However, we could go one step further and add this logic into Microsoft.Extensions.Azure so that we check for this setting and if not present, we pass null.

JoshLove-msft:
> > We might need a change to the underlying extensions library @JoshLove-msft is helping to take a look.
> 
> So one way around needing the new ctor, would be to specify an empty defaultSubscriptionId in config, e.g.
> 
> ```cs
>   var configuration = new ConfigurationBuilder().AddInMemoryCollection(new KeyValuePair<string, string>[]
>   {
>       new KeyValuePair<string, string>("defaultSubscriptionId", "")
>   }).Build();
> 
>   var services = new ServiceCollection();
>   services.AddAzureClients(builder =>
>   {
>       builder.AddArmClient(configuration);
>   });
>   var client = services.BuildServiceProvider().GetRequiredService<ArmClient>();
> ```
> 
> This would cause the extension to use the ctor with the defaultSubscriptionId. However, we could go one step further and add this logic into Microsoft.Extensions.Azure so that we check for this setting and if not present, we pass null.

Alternatively, we can add this into the extension method signature:
```c#
public static IAzureClientBuilder<ArmClient, ArmClientOptions> AddArmClient<TBuilder>(this TBuilder builder, string defaultSubscriptionId)    where TBuilder : IAzureClientFactoryBuilderWithCredential{    return builder.RegisterClientFactory<ArmClient, ArmClientOptions>((options, token) => new ArmClient(token, defaultSubscriptionId, options));}
```

m-nash:
@Stolpe created a PR into this branch to show the changes I think are needed. https://github.com/Stolpe/azure-sdk-for-net/pull/1

Issue #29342, opened 2022-06-16
Archboard feedback on DPG generation of pageables


heaths:
Based on feedback highlighted in #29331, the archboard gave feedback on pageables. Currently, DPG is generating methodes like:

```C#
public virtual Pageable<BinaryData> GetProjects(int? top = null, int? skip = null, int? maxpagesize = null, RequestContext context = null);
```

Based on the [archboard conversation](https://apiview.dev/Assemblies/Review/20a58d7611e845939ebfd25707b16b88#06f51472b6894935a0d0c153094ad53e):

- [ ] `top` should be `size` in .NET (or "Size" as a property; more on that...).
- [ ] `Size` (nee `top`) and `Skip` should be moved to properties, but @tg-msft may need to fill in the desire here. He thought we already had these on pages or something, but I can find no evidence of that in Azure.Core or Azure.Storage.*.
- [ ] `maxpagesize` should not be exposed directly, and instead should be populated from `Pageable<T>.AsPage()`'s `pageSizeHint` as shown [here](https://github.com/Azure/azure-sdk-for-net/blob/6a1c4098a17b99df1992e0a9120b61a1bdd8c12a/sdk/core/Azure.Core/src/Pageable.cs#L61).

lirenhe:
We will first take a look at the first item.

lirenhe:
@heaths, I don't quite understand the second item, so could you help to provide more details?  thanks

ArthurMa1978:
I really doubt what's the value of this naming change?
`top` + `skip` is a well-known parameter pair for pageable REST API, and others like `limit` + `offset`, I think all these are acceptable because it can represent the meaning clearly, no matter which language.

archerzz:
My concern is that there are multiple combinations of pagination parameters. Shall we convert all of them? Then we could miss some of them.
If the requirement is not restricted to .Net SDK, then why not we apply it to swagger/cadl so that every language SDK can benefit.

heaths:
> I really doubt what's the value of this naming change? `top` + `skip` is a well-known parameter pair for pageable REST API, and others like `limit` + `offset`, I think all these are acceptable because it can represent the meaning clearly, no matter which language.

This was feedback from the architects @KrzysztofCwalina and @tg-msft. This is something they wanted for all LROs in .NET.

PR #31603, opened 2022-10-04
Remove pageable parameters for QA

heaths:
Resolves architect feedback. We can add an overload if needed when #29342 is resolved.


kristapratico (approved):
Might be worth calling out on the changelog

heaths:
> Might be worth calling out on the changelog

We never shipped a version with those parameters...or even the client. :smile:

Issue #25462, opened 2021-11-22
Replace Microsoft.Azure.ApplicationInsights.Query dependency


scottaddie:
The [Azure.Monitor.Query](https://www.nuget.org/packages/Azure.Monitor.Query/) package, which shipped last month, is intended to replace [Microsoft.Azure.ApplicationInsights.Query](https://www.nuget.org/packages/Microsoft.Azure.ApplicationInsights.Query). Migrate away from the Track 1 dependency that currently exists in the [OTel Exporter integration tests project](https://github.com/Azure/azure-sdk-for-net/blob/9c9ec82c36580deff41b8940af420519a28cd4e1/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Integration.Tests/Azure.Monitor.OpenTelemetry.Exporter.Integration.Tests.csproj#L8). We're actively working toward deprecating the *Microsoft.Azure.ApplicationInsights.Query* package.

scottaddie:
The Microsoft.Azure.ApplicationInsights.Query package is now deprecated.

PR #31608, opened 2022-10-05
Integration/Live tests using `Azure.Monitor.Query`

vishweshbankwar:
Addresses [#25462](https://github.com/Azure/azure-sdk-for-net/issues/25462)

This PR is a proof of concept and can be used as a base for adding additional test scenarios for live testing.

Steps to add/re-run live tests

1) Run Connect-AzAccount -Subscription subId
2) Run .\New-TestResources.ps1 from https://github.com/Azure/azure-sdk-for-net/tree/main/eng/common/TestResources
3) Update tests and re-run
4) For live tests json files will be updated under SessionRecords


sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/E2E.Tests/AzureMonitorTraceExporterLiveTests.cs
```diff
@@ -0,0 +1,110 @@
...
+            {
+                { "service.name", roleName },
+            };
+
+            var resourceBuilder = ResourceBuilder.CreateDefault().AddAttributes(resourceAttributes);
+
+            using var tracerProvider = Sdk.CreateTracerProviderBuilder()
+                .SetResourceBuilder(resourceBuilder)
+                .AddSource("TestActivitySource")
+                .SetSampler(new AlwaysOnSampler())
```
TimothyMothra, line 48: I think the sampler can be removed?
Correct me if I'm wrong, but no sampler would be equivalent to AlwaysOn.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/E2E.Tests/AzureMonitorTraceExporterLiveTests.cs
```diff
@@ -0,0 +1,110 @@
...
+
+            string query = $"AppDependencies | where AppRoleName == '{roleName}'";
+
+            LogsTable table = await CheckForRecord(client, query);
+
+            // Assert
+            Assert.True(table.Rows.Count() > 0);
+        }
+
+        private async Task<LogsTable> CheckForRecord(LogsQueryClient client, string query)
```
TimothyMothra, line 71: Consider including a Guid in the telemetry item. Maybe `activity.SetTag` ?
The point is that if you run this test multiple times, the verify would fail.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/E2E.Tests/Azure.Monitor.OpenTelemetry.Exporter.E2E.Tests.csproj
```diff
@@ -6,6 +6,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Azure.ApplicationInsights.Query" />
```
TimothyMothra, line 8: I think this can be removed.

```suggestion
```

vishweshbankwar: Not right now. will do delete separately (I have not removed the test based on this package).

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/E2E.Tests/AzureMonitorTraceExporterLiveTests.cs
vishweshbankwar: I am keeping the roleName unique. I know that can fail too if the tests is run in live mode multiple times as there will be more than one record matching the roleName. Right now, the assumption is this will be run only once (as per my understanding live tests are run only during nightly build)

I was trying to use guid but then it needs special handling when the test is run in playback mode as the query will fail (we will have a new guid every time the test runs) and there will be no recorded data that will match this newly created guid. I also tried doing special handling in case of playback versus recorded mode (e.g. using `like` versus `==` in `where` clause) it did not work as expected. This is one of the things I need to look in to :)

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/E2E.Tests/TestFramework/MonitorExporterTestEnvironment.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Azure.Monitor.OpenTelemetry.Exporter.E2E.Tests.TestFramework
+{
+    using System;
+
+    using Azure.Core.TestFramework;
+
+    public class MonitorExporterTestEnvironment : TestEnvironment
```
TimothyMothra, line 10: Why the new TestEnvironment class? 
I'm guessing you want an isolated class to test with while getting the new test working.

When you're ready to merge, I expect we only need one class.

vishweshbankwar: Yes - I am planning to remove the old one as there are many things in there we don't need.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/E2E.Tests/AzureMonitorTraceExporterLiveTests.cs
vishweshbankwar: I have used roleName for uniqueness. To resolve case of duplicates getting top(1) by timegenerated.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/E2E.Tests/AzureMonitorTraceExporterLiveTests.cs
```diff
@@ -0,0 +1,125 @@
...
+            int count = 0;
+
+            // Try every 30 secs for total of 5 minutes.
+            int maxTries = 10;
+            while (count == 0 && maxTries > 0)
+            {
+                Response<LogsQueryResult> response = await client.QueryWorkspaceAsync(
+                TestEnvironment.WorkspaceId,
+                query,
+                new QueryTimeRange(TimeSpan.FromMinutes(30)));
```
TimothyMothra, line 93: nit: fix indentation here

TimothyMothra (approved):

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/E2E.Tests/Azure.Monitor.OpenTelemetry.Exporter.E2E.Tests.csproj
scottaddie: @vishweshbankwar The linked issue isn't truly fixed until this package reference has been removed. I suggest changing your "fixes" comment to "addresses".

vishweshbankwar: Done - Thanks!

PR #31609, opened 2022-10-05
Call StopPlayback on TestRecording Dispose in Playback mode

christothes:
Previously, `_proxy.Client.StopPlaybackAsync` was never called when ending a playback test session. This prevents TestProxy from cleaning up resources created when playback starts.


christothes:
> What is with the test recordings?

They were never actually recorded in the first place. This only became obvious when we tried to call stop on a non-existent recording session :)

JoshLove-msft (approved):

PR #31627, opened 2022-10-06
add kind property to deserializer logic

ostoliarova-ee:
# Description

Made `CommunicationIdentifierModelKind` to have the sdame name and namespace for different modalities and introduced the logic to process `CommunicationIdentifierModelKind` when deserializing.

[User Story 3013988](https://skype.visualstudio.com/SPOOL/_workitems/edit/3013988): [SDK][Common][AllLangs] Prefer the "kind" property when deserializing the CommunicationIdentifier - follow up


sdk/communication/Azure.Communication.MediaComposition/src/Generated/Models/CommunicationIdentifierModelKind.cs
```diff
@@ -8,10 +8,10 @@
 using System;
 using System.ComponentModel;
 
-namespace Azure.Communication.MediaComposition
+namespace Azure.Communication
 {
     /// <summary> The identifier kind, for example &apos;communicationUser&apos; or &apos;phoneNumber&apos;. </summary>
-    public readonly partial struct CommunicationIdentifierModelKind : IEquatable<CommunicationIdentifierModelKind>
```
petrsvihlik, line 14: we can't do this, it'd be a breaking change. let's keep it public
- https://apiview.dev/Assemblies/Review/5b1c0e6abcad4356839133822c404c91/c7381770117e4541937ad9c500e339d0?diffRevisionId=73c97198062443d5b8a60114d36d4ac1&doc=False&diffOnly=True

sdk/communication/Azure.Communication.Identity/README.md
```diff
@@ -125,12 +125,20 @@ Console.WriteLine($"Token: {token.Token}");
...
 ```C# Snippet:RevokeCommunicationUserTokenAsync
 Response revokeResponse = await client.RevokeTokensAsync(user);
 ```
 
 ### Deleting a user
 
+```C# Snippet:DeleteACommunicationUser
+Response deleteResponse = client.DeleteUser(user);
+```
+
```
petrsvihlik, line 141: why are these changes necessary?

sdk/communication/Azure.Communication.Chat/src/Models/CommunicationIdentifierModelKind.cs
```diff
@@ -0,0 +1,58 @@
...
+
+#nullable disable
+
+using System;
+using System.ComponentModel;
+
+namespace Azure.Communication
+{
+    /// <summary> Type of CommunicationIdentifierModel. </summary>
+    public readonly partial struct CommunicationIdentifierModelKind : IEquatable<CommunicationIdentifierModelKind>
```
petrsvihlik, line 15: would it be possible to keep this internal for chat for now? not to generate the API change...?

https://apiview.dev/Assemblies/Review/83a5714c2c6a4d03a9b671fe5148e9a1?diffRevisionId=17ce6284d9c844ab81b74d5261342ee4&diffOnly=False&revisionId=565e7b4109db46eda72443f540140776&doc=False

otherwise, we'd need an ARB review

sdk/communication/Azure.Communication.Chat/src/Models/CommunicationIdentifierModel.cs
```diff
@@ -1,12 +1,21 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+// This is a temporary solution. Please remove this code when swagger is updated to support `kind` property.
```
petrsvihlik, line 4: please mark the target Common swagger version to make it 100% clear

sdk/communication/Shared/src/CommunicationIdentifierSerializer.cs
```diff
@@ -11,12 +11,45 @@ internal class CommunicationIdentifierSerializer
...
+            string rawId = AssertNotNull(identifier.RawId, nameof(identifier.RawId), nameof(CommunicationIdentifierModel));
 
             AssertMaximumOneNestedModel(identifier);
 
-            if (identifier.CommunicationUser is CommunicationUserIdentifierModel user)
-                return new CommunicationUserIdentifier(AssertNotNull(user.Id, nameof(user.Id), nameof(CommunicationUserIdentifierModel)));
+            if (identifier.Kind is not null)
+            {
+                if (identifier.Kind == CommunicationIdentifierModelKind.CommunicationUser
+                    && identifier.CommunicationUser is not null)
```
petrsvihlik, line 21: I wonder if we should support kind being set and the related object not being set...but I suppose if we don't break and fall back to the unknownidentifier and make no further assumptions about it, it's fine...
hmm, yeah, I suppose it's fine.

sdk/communication/Shared/src/CommunicationIdentifierSerializerTest.cs
```diff
@@ -172,5 +172,45 @@ public void DeserializeMicrosoftTeamsUser(bool isAnonymous)
             Assert.AreEqual(expectedIdentifier.RawId, identifier.RawId);
             Assert.AreEqual(expectedIdentifier, identifier);
         }
+
+        [Test]
+        public void DeserializeMicrosoftTeamsUser_WithKind_DeserializesSuccessfully()
```
petrsvihlik, line 177: can you pls add tests for all kinds? you never know what can go wrong :)

petrsvihlik (changes requested):

sdk/communication/Shared/src/CommunicationIdentifierSerializerTest.cs
ostoliarova-msft: Done!

sdk/communication/Azure.Communication.MediaComposition/src/Generated/Models/CommunicationIdentifierModelKind.cs
ostoliarova-msft: I've rolled back this change, thanks!

sdk/communication/Azure.Communication.Identity/README.md
ostoliarova-msft: Rolled back

sdk/communication/Azure.Communication.MediaComposition/api/Azure.Communication.MediaComposition.netstandard2.0.cs
```diff
@@ -1,3 +1,26 @@
+namespace Azure.Communication
```
petrsvihlik, line 1: The namespace change is a breaking change.
And I suppose we need it to be `Azure.Communication` to be able to work with it in the serializer, right?

Can you pls check with the media composition team, if they've already GA'ed their API or if we can make this change.

adding @DominikMe for visibility. pls consult with him while I'm away

petrsvihlik (changes requested):

sdk/communication/Azure.Communication.MediaComposition/api/Azure.Communication.MediaComposition.netstandard2.0.cs
petrsvihlik: looking at their [readme](https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/communication/Azure.Communication.MediaComposition), I don't think media composition is even in private preview so I think it should be safe to make this change

DominikMe: Yes, it's okay to change this. The MediaComposition SDK hasn't been released yet to anybody.

DominikMe (approved):

sdk/communication/Azure.Communication.MediaComposition/api/Azure.Communication.MediaComposition.netstandard2.0.cs
ostoliarova-msft: Thank you @DominikMe and @petrsvihlik !

sdk/communication/Azure.Communication.Chat/src/Models/CommunicationIdentifierModel.cs
```diff
@@ -1,12 +1,22 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+// This is a temporary solution.
+// Please remove this code when swagger is updated to support `kind` property: https://github.com/Azure/azure-rest-api-specs/pull/19675
```
AikoBB, line 5: perhaps we should create an ADO for it, otherwise it will be easily forgotten

ostoliarova-msft: @AikoBB I think that when code is regenerated for a newer swagger version, there will be a conflict with this file and it won't compile, so once updating thew swagger and recompiling, a team will already know that something is wrong here and the comment explains why the file was created and why it could be separately removed. ADO for this one could be "update swagger and regenerate your modality code" but I think it's something that is planned by teams anyway, isn't it?

ostoliarova-msft: @AikoBB  done

AikoBB (approved):

petrsvihlik (approved):

sdk/communication/Shared/src/CommunicationIdentifierSerializer.cs
```diff
@@ -11,12 +11,45 @@ internal class CommunicationIdentifierSerializer
...
+                {
+                    var user = identifier.MicrosoftTeamsUser;
+                    return new MicrosoftTeamsUserIdentifier(
+                          AssertNotNull(user.UserId, nameof(user.UserId), nameof(MicrosoftTeamsUserIdentifierModel)),
+                          AssertNotNull(user.IsAnonymous, nameof(user.IsAnonymous), nameof(MicrosoftTeamsUserIdentifierModel)),
+                          Deserialize(AssertNotNull(user.Cloud, nameof(user.Cloud), nameof(MicrosoftTeamsUserIdentifierModel))),
+                          rawId);
+                }
+
+                return new UnknownIdentifier(rawId);
```
petrsvihlik, line 45: We should probably replace all occurrences of `return new UnknownIdentifier(rawId);` with `return CommunicationIdentifier.FromRawId(rawId)` here and adjust the test accordingly.

petrsvihlik: Reasoning: It's really weird that we're expecting an `UnknownIdentifier` when we've specified the `kind`.
image

AikoBB: totally agree, good point. Even if the other variables are not set we can identify the identifier type from the rawID, btw can we apply also here my comments from java(https://github.com/Azure/azure-sdk-for-java/pull/31522#discussion_r999458918, https://github.com/Azure/azure-sdk-for-java/pull/31522#discussion_r998399746)

petrsvihlik: Ok, after the internal discussion I gave it some more thought and here's how I'd design the behaviors:

-	RawId missing – fail as it’s a required property in responses (as specified in the swagger). We do fail now -> OK.
-	IdentifierModel missing entirely OR not matching the kind property (if specified) - The polymorphic type is broken, we don’t know what’s going on (probably a service returns something unexpected). Let’s not try to be smart and use UnknownIdentifier
-	RawId not matching IdentifierModel  - We haven’t done these checks before, we probably shouldn’t add them now either. If there is a problem with a service, the error will bubble up either way. We could try to fail early but if we add a new MRI format for, for instance, PhoneNumber in the future, this validation would prevent using an older version of an SDK. It’s not even clear how we should do it – deserialize through both MRI and IdentifierModel and then compare?...
-	Kind is not present, IdentifierModel is missing, RawID is present – this probably means that the IdentifierModel has not been recognized (we might have added a new model in the service that’s not yet understood by an older version of the SDK). It’s unlikely that FromRawId will be able to recognize the model either.

So, for the most part, the PR is fine. But to distil some changes from all this:
- I'd get rid of the ` return new UnknownIdentifier(rawId);` -> it's redundant
- I'd also refactor the logic on lines 48-68 to extract the "kind" and use it as a secondary input for the `kind` variable (if kind is not defined in the response). this will hugely reduce the code duplication
```c#
var kind = identifier.Kind ?? GetKind(identifier);
if (kind is not null)
```

ostoliarova-msft: @petrsvihlik please see PR https://github.com/Azure/azure-sdk-for-net/pull/31913

PR #31631, opened 2022-10-06
Add succeeded() to condition for "Install DevOps Logger"

mikeharder:
- Step should only run if previous steps have succeeded


weshaggard (approved):
Thanks for fixing this as I saw it the other day as well.

PR #31641, opened 2022-10-07
[PostgreSQL] SDK upgrade for PostgreSQL flexible server to version 2022-03-08-privatepreview

alanenriqueo:
- Use tag [flexibleserver-2022-03-privatepreview](https://github.com/Azure/azure-rest-api-specs/blob/05a9cdab363b8ec824094ee73950c04594325172/specification/postgresql/resource-manager/readme.md) for `PostgreSqlFlexibleServers`.
- Added `FastCreate` custom function for fast provisioning feature.

---


sdk/postgresql/Azure.ResourceManager.PostgreSql/src/PostgreSqlFlexibleServers/Custom/FastProvisioning.cs
```diff
@@ -0,0 +1,128 @@
...
+    public partial class PostgreSqlFlexibleServerCollection
+    {
+        private GetCachedServerNameRestOperations _getCachedServerNameRestClient;
+
+        private GetCachedServerNameRestOperations GetCachedServerNameRestClient => _getCachedServerNameRestClient ??= new GetCachedServerNameRestOperations(Pipeline, Diagnostics.ApplicationId, Endpoint);
+
+        /// <summary>
+        /// Creates a new server using fast provisioning method.
+        /// </summary>
+        /// <param name="waitUntil"> <see cref="WaitUntil.Completed"/> if the method should wait to return until the long-running operation has completed on the service; <see cref="WaitUntil.Started"/> if it should return after starting the operation. For more information on long-running operations, please see <see href="https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/samples/LongRunningOperations.md"> Azure.Core Long-Running Operation samples</see>. </param>
```
DaeunYim, line 25: Why don't we use FastCreate instead of fast provisioning if we want to align with the existing regular provisioning command?

sdk/postgresql/Azure.ResourceManager.PostgreSql/tests/Scenario/PostgreSqlFlexibleServerTests.cs
```diff
@@ -95,5 +95,68 @@ public async Task CreateUpdateGetDelete()
             // Delete
             await serverFromGet.DeleteAsync(WaitUntil.Completed);
         }
+
+        [TestCase]
+        [RecordedTest]
+        public async Task FastProvision()
+        {
+            var version = PostgreSqlFlexibleServerVersion.Ver12;
```
DaeunYim, line 103: Good that you added test here.

sdk/postgresql/Azure.ResourceManager.PostgreSql/src/PostgreSqlFlexibleServers/Custom/FastProvisioning.cs
```diff
@@ -0,0 +1,128 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Creates a new server using fast provisioning method.
+        /// </summary>
+        /// <param name="waitUntil"> <see cref="WaitUntil.Completed"/> if the method should wait to return until the long-running operation has completed on the service; <see cref="WaitUntil.Started"/> if it should return after starting the operation. For more information on long-running operations, please see <see href="https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/samples/LongRunningOperations.md"> Azure.Core Long-Running Operation samples</see>. </param>
+        /// <param name="data"> The required parameters for creating a server. </param>
+        /// <param name="cancellationToken"> The cancellation token to use. </param>
+        public virtual ArmOperation<PostgreSqlFlexibleServerResource> FastProvision(WaitUntil waitUntil, PostgreSqlFlexibleServerData data, CancellationToken cancellationToken = default)
```
DaeunYim, line 81: What will be the example of using this method like? Can we possibly add a sample of this in the sample repo if possible?

sdk/postgresql/Azure.ResourceManager.PostgreSql/src/PostgreSqlFlexibleServers/Custom/FastProvisioning.cs
alanenriqueo: Done.

sdk/postgresql/Azure.ResourceManager.PostgreSql/src/PostgreSqlFlexibleServers/Custom/FastProvisioning.cs
alanenriqueo: Sure, after this gets merged, I'll add a sample.

sdk/postgresql/Azure.ResourceManager.PostgreSql/src/PostgreSqlFlexibleServers/Custom/FastCreate.cs
```diff
@@ -0,0 +1,128 @@
...
+
+        private GetCachedServerNameRestOperations GetCachedServerNameRestClient => _getCachedServerNameRestClient ??= new GetCachedServerNameRestOperations(Pipeline, Diagnostics.ApplicationId, Endpoint);
+
+        /// <summary>
+        /// Creates a new server using fast provisioning method.
+        /// </summary>
+        /// <param name="waitUntil"> <see cref="WaitUntil.Completed"/> if the method should wait to return until the long-running operation has completed on the service; <see cref="WaitUntil.Started"/> if it should return after starting the operation. For more information on long-running operations, please see <see href="https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/samples/LongRunningOperations.md"> Azure.Core Long-Running Operation samples</see>. </param>
+        /// <param name="data"> The required parameters for creating a server. </param>
+        /// <param name="cancellationToken"> The cancellation token to use. </param>
+        public virtual async Task<ArmOperation<PostgreSqlFlexibleServerResource>> FastCreateAsync(WaitUntil waitUntil, PostgreSqlFlexibleServerData data, CancellationToken cancellationToken = default)
```
ArcturusZhang, line 28: Could you please explain which part in the original generated code we are trying to fix using this piece of customized code?
If there is anything we could fix it using configuration, we should use configuration instead of using customized code because customized code will be hard to maintain without proper comments to explain why they are here

sdk/postgresql/Azure.ResourceManager.PostgreSql/api/Azure.ResourceManager.PostgreSql.netstandard2.0.cs
```diff
@@ -426,6 +426,8 @@ public static partial class FlexibleServersExtensions
...
         public static System.Threading.Tasks.Task<Azure.Response<Azure.ResourceManager.PostgreSql.FlexibleServers.Models.PostgreSqlFlexibleServerNameAvailabilityResult>> CheckPostgreSqlFlexibleServerNameAvailabilityAsync(this Azure.ResourceManager.Resources.SubscriptionResource subscriptionResource, Azure.ResourceManager.PostgreSql.FlexibleServers.Models.PostgreSqlFlexibleServerNameAvailabilityContent content, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
+        public static Azure.Response<Azure.ResourceManager.PostgreSql.FlexibleServers.Models.PostgreSqlFlexibleServerCachedServerName> ExecuteGetCachedServerName(this Azure.ResourceManager.Resources.ResourceGroupResource resourceGroupResource, Azure.Core.AzureLocation locationName, Azure.ResourceManager.PostgreSql.FlexibleServers.Models.PostgreSqlFlexibleServerCachedServerNameContent content, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
```
ArcturusZhang, line 429: The word `Execute` looks a bit redundant, can we rename this operation to simply `GetCachedServerName`?

sdk/postgresql/Azure.ResourceManager.PostgreSql/src/PostgreSqlFlexibleServers/Generated/Models/PostgreSqlFlexibleServerCapabilityProperties.cs
```diff
@@ -16,34 +16,36 @@ public partial class PostgreSqlFlexibleServerCapabilityProperties
...
         /// <param name="isGeoBackupSupported"> A value indicating whether a new server in this region can have geo-backups to paired region. </param>
         /// <param name="isZoneRedundantHASupported"> A value indicating whether a new server in this region can support multi zone HA. </param>
         /// <param name="isZoneRedundantHAAndGeoBackupSupported"> A value indicating whether a new server in this region can have geo-backups to paired region. </param>
         /// <param name="supportedFlexibleServerEditions"></param>
         /// <param name="supportedHyperscaleNodeEditions"></param>
-        /// <param name="supportedHAModes"> Supported high availability mode. </param>
         /// <param name="status"> The status. </param>
-        internal PostgreSqlFlexibleServerCapabilityProperties(string zone, bool? isGeoBackupSupported, bool? isZoneRedundantHASupported, bool? isZoneRedundantHAAndGeoBackupSupported, IReadOnlyList<PostgreSqlFlexibleServerEditionCapability> supportedFlexibleServerEditions, IReadOnlyList<PostgreSqlFlexibleServerHyperscaleNodeEditionCapability> supportedHyperscaleNodeEditions, IReadOnlyList<string> supportedHAModes, string status)
```
archerzz, line 33: This is a breaking change. I saw the new version is `1.1.0-beta1`, so it is OK. But you only change the sequence of the parameter, which should be avoided.

sdk/postgresql/Azure.ResourceManager.PostgreSql/src/PostgreSqlFlexibleServers/Generated/Models/PostgreSqlFlexibleServerVirtualNetworkSubnetUsageResult.Serialization.cs
```diff
@@ -15,43 +15,43 @@ public partial class PostgreSqlFlexibleServerVirtualNetworkSubnetUsageResult
     {
         internal static PostgreSqlFlexibleServerVirtualNetworkSubnetUsageResult DeserializePostgreSqlFlexibleServerVirtualNetworkSubnetUsageResult(JsonElement element)
         {
+            Optional<IReadOnlyList<PostgreSqlFlexibleServerDelegatedSubnetUsage>> delegatedSubnetsUsage = default;
             Optional<AzureLocation> location = default;
             Optional<string> subscriptionId = default;
-            Optional<IReadOnlyList<PostgreSqlFlexibleServerDelegatedSubnetUsage>> delegatedSubnetsUsage = default;
```
archerzz, line 20: Same breaking change. Not an issue but it's better to avoid.

sdk/postgresql/Azure.ResourceManager.PostgreSql/api/Azure.ResourceManager.PostgreSql.netstandard2.0.cs
```diff
@@ -624,6 +632,36 @@ public partial class PostgreSqlFlexibleServerResource : Azure.ResourceManager.Ar
...
         public virtual System.Threading.Tasks.Task<Azure.ResourceManager.ArmOperation<Azure.ResourceManager.PostgreSql.FlexibleServers.PostgreSqlFlexibleServerResource>> UpdateAsync(Azure.WaitUntil waitUntil, Azure.ResourceManager.PostgreSql.FlexibleServers.Models.PostgreSqlFlexibleServerPatch patch, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
     }
+    public partial class PostgreSqlFlexibleServerServerBackupCollection : Azure.ResourceManager.ArmCollection, System.Collections.Generic.IAsyncEnumerable<Azure.ResourceManager.PostgreSql.FlexibleServers.PostgreSqlFlexibleServerServerBackupResource>, System.Collections.Generic.IEnumerable<Azure.ResourceManager.PostgreSql.FlexibleServers.PostgreSqlFlexibleServerServerBackupResource>, System.Collections.IEnumerable
```
Yao725, line 635: From the SDK API perspective, this isn't a good resource name. Why do we have two consecutive word `Server`? Can we omit one of them?

sdk/postgresql/Azure.ResourceManager.PostgreSql/src/PostgreSqlFlexibleServers/Custom/FastCreate.cs
alanenriqueo: The purpose of `FastCreate` custom function is to get a server name using `GetCachedServerName` and then use it for server `Create`, this server name allows for a fast creation. I'm not sure if this can be customized through configuration, any thoughts?

cc @rajsell @DaeunYim

sdk/postgresql/Azure.ResourceManager.PostgreSql/api/Azure.ResourceManager.PostgreSql.netstandard2.0.cs
alanenriqueo: Done.

sdk/postgresql/Azure.ResourceManager.PostgreSql/api/Azure.ResourceManager.PostgreSql.netstandard2.0.cs
alanenriqueo: Done.

sdk/postgresql/Azure.ResourceManager.PostgreSql/src/PostgreSqlFlexibleServers/Generated/Models/PostgreSqlFlexibleServerCapabilityProperties.cs
archerzz: @alanenriqueo I just realized that there is already an official release `1.0.0`. So this is a breaking change which needs to be resolved.

Here are 2 options:
  - Add manual codes to provide backward compatibility, see compute for reference: https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/compute/Microsoft.Azure.Management.Compute/src/Customizations
- Or you need to bump the major version, e.g. `2.0.0`

sdk/postgresql/Azure.ResourceManager.PostgreSql/src/PostgreSqlFlexibleServers/Generated/Models/PostgreSqlFlexibleServerVirtualNetworkSubnetUsageResult.Serialization.cs
archerzz: Need to resolve since there is already an official version.

sdk/postgresql/Azure.ResourceManager.PostgreSql/src/PostgreSqlFlexibleServers/Custom/FastCreate.cs
ArcturusZhang: Let me ask in this way - do we have this method in the generated code if we do not have this piece of customized code?
And if we do, what is the difference between the generated method and this method?
I assume if we have a RestOperation class for your operation `GetCachedServerNameRestOperations` the generator should be able to generate that method, and your customization should be trying to fix something in that generated method

sdk/postgresql/Azure.ResourceManager.PostgreSql/src/PostgreSqlFlexibleServers/Generated/Models/PostgreSqlFlexibleServerCapabilityProperties.cs
ArcturusZhang: Changing order of properties is no longer a brekaing change in track 2 SDKs

sdk/postgresql/Azure.ResourceManager.PostgreSql/src/PostgreSqlFlexibleServers/Generated/Models/PostgreSqlFlexibleServerVirtualNetworkSubnetUsageResult.Serialization.cs
ArcturusZhang: I only see order changes here, this is not a breaking change. The API file in this RP shows this PR only introduces additions

sdk/postgresql/Azure.ResourceManager.PostgreSql/src/PostgreSqlFlexibleServers/Custom/FastCreate.cs
alanenriqueo: No, the method `FastCreate` doesn't exist in the generated code. This new method contains a new logic that is a combination of both `GetCachedServerName` and `Create`.

sdk/postgresql/Azure.ResourceManager.PostgreSql/src/autorest.md
```diff
@@ -14,13 +14,13 @@ modelerfour:
 
 batch:
   - tag: package-2020-01-01
-  - tag: package-flexibleserver-2021-06
+  - tag: flexibleserver-2022-03-privatepreview
```
ArthurMa1978, line 17: The https://github.com/Azure/azure-rest-api-specs/pull/20212  was merged 20 mins ago, do you need to update the sdk accordingly？

sdk/postgresql/Azure.ResourceManager.PostgreSql/CHANGELOG.md
```diff
@@ -1,14 +1,11 @@
 # Release History
 
-## 1.1.0-beta.1 (Unreleased)
+## 1.1.0-beta.1 (2022-10-11)
```
ArthurMa1978, line 3: Please update this to a new release day as the swagger has just been merged 20 mins ago

ArthurMa1978: And please also flip the pr from `draft` to `Ready for review`

PR #31655, opened 2022-10-07
[Event Hubs] Processor README Updates

jsquire:
# Summary

The focus of these changes is to include the processor-specific mocking sample in the `Additional concepts` section.


JoshLove-msft:
The rendering doesn't look right:

image

jsquire:
> The rendering doesn't look right:
> 
> image

Yeah.... that was an odd typo.   Fixed!

JoshLove-msft (approved):

PR #31657, opened 2022-10-07
[Text Analytics] Re-gen based on version 2022-10-01-preview

### LINKED ISSUE #31266: [Text Analytics] Regenerate implementation based on REST API version 2022-10-01-preview

# Tasks
- [ ] Implementation
- [ ] Tests
- [ ] Samples
- [ ] Docs

joseharriaga:
Resolves #31266.

- Re-generated the library based on the draft swagger spec for version 2022-10-01-preview.
- Fixed things that broke as a consequence.
- Added a new value to the `ServiceVersion` enum.
- Added a new value to the SDK version table in README.md.
- Re-ran Export-API.ps1.


sdk/textanalytics/Azure.AI.TextAnalytics/api/Azure.AI.TextAnalytics.netstandard2.0.cs
```diff
@@ -1287,6 +1289,444 @@ public enum TextSentiment
         Mixed = 3,
     }
 }
+namespace Azure.AI.TextAnalytics.Models
+{
+    [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
+    public readonly partial struct AgeUnit : System.IEquatable<Azure.AI.TextAnalytics.Models.AgeUnit>
```
kinelski, line 1295: It seems structs are not being generated with the `internal` modifier by default. You may need to add those to `InternalTypes.cs` if you want to keep them internal for now.

heaths:
> Fixed things that broke as a consequence.

Like what? Generally, nothing should break or this indicates a problem in the swagger.

sdk/textanalytics/Azure.AI.TextAnalytics/README.md
```diff
@@ -27,10 +27,11 @@ dotnet add package Azure.AI.TextAnalytics
 
 This table shows the relationship between SDK versions and supported API versions of the service:
 
-> Note that `5.2.0-beta.3` is the first version of the client library that targets the Azure Cognitive Service for Language APIs which includes the existing text analysis and natural language processing features found in the Text Analytics client library. In addition, the service API has changed from semantic to date-based versioning. This version of the client library defaults to the latest supported API version, which currently is `2022-04-01-preview`.
+> Note that `5.2.0` is the first stable version of the client library that targets the Azure Cognitive Service for Language APIs, which include the existing text analysis and natural language processing features found in the Text Analytics client library. In addition, the service API has changed from semantic to date-based versioning.
```
heaths, line 30:
```suggestion
> Note that `5.2.0` is the first stable version of the client library that targets the Azure Cognitive Service for Language APIs which includes the existing text analysis and natural language processing features found in the Text Analytics client library. In addition, the service API has changed from semantic to date-based versioning.
```

The comma is extraneous, and the subject of this sentence is singular so "includes" is correct.

sdk/textanalytics/Azure.AI.TextAnalytics/api/Azure.AI.TextAnalytics.netstandard2.0.cs
```diff
@@ -1029,6 +1029,7 @@ public enum ServiceVersion
             V3_0 = 1,
             V3_1 = 2,
             V2022_05_01 = 3,
+            V2022_10_01_Preview = 4,
```
heaths, line 1032: So none of the new features have been exposed yet, right?

sdk/textanalytics/Azure.AI.TextAnalytics/src/GeneratorStubs/InternalTypes.cs
```diff
@@ -7,29 +7,89 @@ namespace Azure.AI.TextAnalytics.Models
 {
 #pragma warning disable SA1402 // File may only contain a single type
 
+    [CodeGenModel("AbstractiveSummarizationResultBaseDocumentsItem")]
```
heaths, line 10: Something is very wrong here. They should all be `internal` by default. This becomes an expensive endeavor to update this every time. Did you try using a transform to set `x-accessibility: internal`?

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsModelFactory.cs
```diff
@@ -219,7 +219,7 @@ public static AnalyzeSentimentResultCollection AnalyzeSentimentResultCollection(
         public static DetectedLanguage DetectedLanguage(string name, string iso6391Name, double confidenceScore, IList<TextAnalyticsWarning> warnings = default)
         {
             warnings ??= new List<TextAnalyticsWarning>();
-            return new DetectedLanguage(new DetectedLanguageInternal(name, iso6391Name, confidenceScore), warnings);
+            return new DetectedLanguage(new DetectedLanguageInternal(name, iso6391Name, confidenceScore, default), warnings);
```
heaths, line 222: If this now takes an additional parameter, you need to expose that as a new overload and make all parameters here have no default values + mark it as hidden. See https://azure.github.io/azure-sdk/dotnet_introduction.html#dotnet-mocking-factory-builder-methods

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextDocumentBatchStatistics.cs
```diff
@@ -51,5 +52,11 @@ internal TextDocumentBatchStatistics(int documentCount, int validDocumentCount,
         /// </summary>
         [CodeGenMember("TransactionsCount")]
         public long TransactionCount { get; }
+
+        /// <summary>
+        /// Gets the additional properties
+        /// </summary>
+        [CodeGenMember("AdditionalProperties")]
+        internal IDictionary<string, object> AdditionalProperties { get; }
```
heaths, line 60: How do you plan to expose this? Is there a tracking issue?

sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.cs
```diff
@@ -85,7 +85,7 @@ internal static DetectLanguageResultCollection ConvertToDetectLanguageResultColl
             var detectedLanguages = new List<DetectLanguageResult>(results.Documents.Count);
 
             //Read errors
-            foreach (DocumentError error in results.Errors)
+            foreach (InputError error in results.Errors)
```
heaths, line 88: None of these renamed types are public, right?

heaths (changes requested):

sdk/textanalytics/Azure.AI.TextAnalytics/src/GeneratorStubs/InternalTypes.cs
joseharriaga: There are two categories of models that ended up here:

Category 1. Extensible enums. I was able to add a transform to the autorest.md to take care of these.

Category 2. Models that are defined in the swagger spec as the "items" in an array while also using the "allOf" property. For example, the swagger definition that generated the "ExtractiveSummarizationResultDocumentsItem" class: [[link]](https://github.com/Azure/azure-rest-api-specs/blob/6bb39b483e16fe68c1fda9aa55801d5f814aa327/dev/cognitiveservices/data-plane/Language/analyzetext.json#L2122). I found this issue on GitHub describing a similar problem: https://github.com/Azure/autorest.csharp/issues/1163. Then I noticed that other models in the same circumstances (such as the "CustomEntitiesResultDocumentsItem" class) were already here, so assumed this was the acceptable workaround for those.

sdk/textanalytics/Azure.AI.TextAnalytics/api/Azure.AI.TextAnalytics.netstandard2.0.cs
joseharriaga: Correct. I'll be exposing them one by one in subsequent PRs (unless there's a better way of doing it?)

sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.cs
joseharriaga: Correct. Also, to answer your other question, this is an example of something that broke after re-generating (fortunately it's just an internal type so it's not an actual breaking change for consumers of the library).

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsModelFactory.cs
joseharriaga: Yes, I will do that. Although my plan was to do so as a separate PR to resolve https://github.com/Azure/azure-sdk-for-net/issues/31272. What is the team's preference for partitioning work?

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextDocumentBatchStatistics.cs
joseharriaga: I do not know how yet, which is why I'm keeping it internal for now. Here's an issue to formally track it: https://github.com/Azure/azure-sdk-for-net/issues/31694

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsModelFactory.cs
heaths: Up to you, but at least making sure there's tracking items. Personally, something like this I would've included here because you're already changing the model factory, so doing it in one shot makes it easier not to forget later. A change like this won't show up in APIView, so it'd be hard to draw the conclusion that another method should've been added.

heaths (approved):

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsModelFactory.cs
joseharriaga: Got it. I will keep this in mind going forward. Thanks!

PR #31664, opened 2022-10-09
[Maps] Add tests

HarveyLink:

### DIFF
--- a/sdk/maps/Azure.ResourceManager.Maps/tests/MapsManagementTestBase.cs
+++ b/sdk/maps/Azure.ResourceManager.Maps/tests/MapsManagementTestBase.cs
@@ -83,7 +83,7 @@ namespace Azure.ResourceManager.Maps.Tests
             }
         }
 
-        public async Task<MapsAccountResource> CreateDefaultMapsAccount(MapsAccountCollection mapCollection, string rgname)
+        public async Task<MapsAccountResource> CreateDefaultMapsAccount(MapsAccountCollection mapCollection)
         {
             var accountName = Recording.GenerateAssetName("maps");
             var parameters = GetDefaultMapsAccountData();
--- a/sdk/maps/Azure.ResourceManager.Maps/tests/MapsManagementTestBase.cs
+++ b/sdk/maps/Azure.ResourceManager.Maps/tests/MapsManagementTestBase.cs
@@ -56,7 +56,7 @@ namespace Azure.ResourceManager.Maps.Tests
             return account;
         }
 
-        public void VerifyAccountProperties(MapsAccountData account, bool useDefaults, Name skuName)
+        public void VerifyAccountProperties(MapsAccountData account, bool useDefaults, Name skuName, string location = "East US")
         {
             Assert.NotNull(account.Id);
             Assert.NotNull(account.Location);
@@ -67,7 +67,7 @@ namespace Azure.ResourceManager.Maps.Tests
 
             if (useDefaults)
             {
-                Assert.AreEqual(DefaultLocation, account.Location);
+                Assert.AreEqual("East US", account.Location.DisplayName);
                 Assert.AreEqual(Name.S0, account.Sku.Name);
 
                 Assert.NotNull(account.Tags);
@@ -79,18 +79,17 @@ namespace Azure.ResourceManager.Maps.Tests
             else
             {
                 Assert.AreEqual(skuName, account.Sku.Name);
-                Assert.AreEqual(DefaultLocation, account.Location);
+                Assert.AreEqual(location, account.Location.DisplayName);
             }
         }
 
-        public async Task<string> CreateDefaultMapsAccount(MapsAccountCollection mapCollection, string rgname)
+        public async Task<MapsAccountResource> CreateDefaultMapsAccount(MapsAccountCollection mapCollection, string rgname)
         {
             var accountName = Recording.GenerateAssetName("maps");
             var parameters = GetDefaultMapsAccountData();
             var newAccount = (await mapCollection.CreateOrUpdateAsync(WaitUntil.Completed, accountName, parameters)).Value;
             VerifyAccountProperties(newAccount.Data, true, Name.S0);
-
-            return accountName;
+            return newAccount;
         }
     }
 }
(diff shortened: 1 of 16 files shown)


sdk/maps/Azure.ResourceManager.Maps/tests/MapsManagementTestBase.cs
```diff
@@ -25,17 +32,65 @@ protected MapsManagementTestBase(bool isAsync)
...
+                Assert.AreEqual("value2", account.Tags["key2"]);
+            }
+            else
+            {
+                Assert.AreEqual(skuName, account.Sku.Name);
+                Assert.AreEqual(DefaultLocation, account.Location);
+            }
+        }
+
+        public async Task<string> CreateDefaultMapsAccount(MapsAccountCollection mapCollection, string rgname)
```
ArcturusZhang, line 86: parameter `rgname` is not used. We could remove it

ArcturusZhang: BTW it should be better if we could return the MapsAccountResource instead of only returning the name

sdk/maps/Azure.ResourceManager.Maps/tests/ScenarioTests/MapsAccountTests.cs
```diff
@@ -0,0 +1,208 @@
...
+        }
+
+        [Test]
+        public async Task MapsAccountDeleteTest()
+        {
+            await SetCollection();
+
+            // Delete an account which does not exist
+            var id = MapsAccountResource.CreateResourceIdentifier(DefaultSubscription.Data.SubscriptionId, ResourceGroup.Data.Name, "missingaccount");
+            var falseAccount = new MapsAccountResource(Client, id);
```
ArcturusZhang, line 92: We should not use internal methods in test cases.
```suggestion
            var falseAccount = Client.GetMapsAccountResource(id);
```

sdk/maps/Azure.ResourceManager.Maps/tests/ScenarioTests/MapsAccountTests.cs
```diff
@@ -0,0 +1,208 @@
...
+
+        private ResourceGroupResource ResourceGroup { get; set; }
+
+        private MapsAccountCollection MapCollection { get; set; }
+
+        private async Task SetCollection()
+        {
+            ResourceGroup = await CreateResourceGroupAsync();
+            MapCollection = ResourceGroup.GetMapsAccounts();
+        }
```
ArcturusZhang, line 34: If these properties are not meant to be shared between different test cases, we do not need to make them properties.
We could just let this method to return both the results.

sdk/maps/Azure.ResourceManager.Maps/tests/ScenarioTests/MapsAccountTests.cs
```diff
@@ -0,0 +1,208 @@
...
+
+        [Test]
+        public async Task MapsAccountCreateTest()
+        {
+            await SetCollection();
```
ArcturusZhang, line 39: actually we could simply do this here
```suggestion
            var ResourceGroup = await CreateResourceGroupAsync();
            var MapCollection = ResourceGroup.GetMapsAccounts();
```
therefore we no longer need these variables as properties

sdk/maps/Azure.ResourceManager.Maps/tests/ScenarioTests/MapsAccountTests.cs
```diff
@@ -0,0 +1,208 @@
...
+        public async Task MapsAccountRegenerateKeyTest()
+        {
+            await SetCollection();
+
+            // prepare account properties
+            string accountName = Recording.GenerateAssetName("maps");
+            var parameters = GetDefaultMapsAccountData();
+
+            // Create account
+            var newAccount = (await MapCollection.CreateOrUpdateAsync(WaitUntil.Completed, accountName, parameters)).Value;
```
ArcturusZhang, line 187: any reason we cannot use the `CreateDefaultMapsAccount` method here? We are using the same parameters as `CreateDefaultMapsAccount`

ArcturusZhang (approved):

sdk/maps/Azure.ResourceManager.Maps/tests/MapsManagementTestBase.cs
```diff
@@ -25,17 +32,65 @@ protected MapsManagementTestBase(bool isAsync)
...
+                Assert.NotNull(account.Tags);
+                Assert.NotNull(account.Properties.UniqueId);
+                Assert.AreEqual(2, account.Tags.Count);
+                Assert.AreEqual("value1", account.Tags["key1"]);
+                Assert.AreEqual("value2", account.Tags["key2"]);
+            }
+            else
+            {
+                Assert.AreEqual(skuName, account.Sku.Name);
+                Assert.AreEqual(DefaultLocation, account.Location);
```
fengzhou-msft, line 82: This check is the same as the other branch when `useDefaults` is true. We should probably pass in an optional `location` parameter for the check in this branch?

HarveyLink: Fixed

sdk/maps/Azure.ResourceManager.Maps/tests/ScenarioTests/MapsAccountTests.cs
HarveyLink: Done

sdk/maps/Azure.ResourceManager.Maps/tests/ScenarioTests/MapsAccountTests.cs
HarveyLink: Fixed

sdk/maps/Azure.ResourceManager.Maps/tests/ScenarioTests/MapsAccountTests.cs
HarveyLink: Fixed

sdk/maps/Azure.ResourceManager.Maps/tests/ScenarioTests/MapsAccountTests.cs
HarveyLink: Fixed

sdk/maps/Azure.ResourceManager.Maps/tests/MapsManagementTestBase.cs
HarveyLink: Fixed

PR #31685, opened 2022-10-10
[Event Hubs] Modernize IoT Hub Sample

jsquire:
# Summary

The focus of these changes is to update the IoT Hub connection string translation sample to reference an up to date version of the Event Hubs package and to shift the target platform to .NET 6, given the upcoming end-of-life for .NET Core 3.1.


samples/iothub-connect-to-eventhubs/Program.cs
```diff
@@ -29,7 +29,7 @@ public static class Program
         ///
         public static async Task<int> Main(string[] args)
         {
-            var connectionString = default(string);
+            var connectionString = "HostName=test-this-thing.azure-devices.net;SharedAccessKeyName=iothubowner;SharedAccessKey=2u+Yzq9wmHlTVsuUyCy8eTzK0gNpcbYbz/UUbXYSjZc="; //default(string);
```
JoshLove-msft, line 32: Where does this come from? Should we replace with something like "<access key">?

jsquire: Accidential push.   Branch was deleted and the associated IoT Hub has been deleted.

PR #31686, opened 2022-10-10
[Event Hubs] Modernize IoT Hub Sample

jsquire:
# Summary

The focus of these changes is to update the IoT Hub connection string translation sample to reference an up to date version of the Event Hubs package and to shift the target platform to .NET 6, given the upcoming end-of-life for .NET Core 3.1.


samples/iothub-connect-to-eventhubs/IotHubToEventHubsSample.csproj
```diff
@@ -10,8 +10,8 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Azure.Messaging.EventHubs" Version="5.6.0" />
-    <PackageReference Include="Microsoft.Azure.Amqp" Version="2.5.3" />
+    <PackageReference Include="Azure.Messaging.EventHubs" Version="5.7.2" />
+    <PackageReference Include="Microsoft.Azure.Amqp" Version="2.5.10" />
```
JoshLove-msft, line 14: Should we just remove this dependency and allow it to be pulled in via Event Hubs?

JoshLove-msft (approved):

samples/iothub-connect-to-eventhubs/IotHubToEventHubsSample.csproj
jsquire: We use it directly in the sample; I didn't want to pull it in just via transitive dependency - it felt confusing.   Thoughts?

JoshLove-msft: Ah that makes sense - didn't realize it was used in the sample.

jsquire:
Merging; the Core timeout is unrelated to this PR.

PR #31690, opened 2022-10-10
[Azure.Monitor.Ingestion] Update parallel default thread count

nisha-bhatia:

### DIFF
--- a/sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
+++ b/sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
@@ -27,8 +27,8 @@ namespace Azure.Monitor.Ingestion
         // request or stage as multiple blocks.
         internal static int SingleUploadThreshold = 1000000; // 1 Mb in byte format
 
-        // If no concurrency count is provided, default to serial upload (one block at a time).
-        private int DefaultWorkerCount = 1;
+        // If no concurrency count is provided for a parallel upload, default to 5 threads.
+        private int DefaultParallelWorkerCount = 5;
 
         internal readonly struct BatchedLogs <T>
         {
@@ -262,8 +262,8 @@ namespace Azure.Monitor.Ingestion
             Argument.AssertNotNullOrEmpty(logs, nameof(logs));
 
             // Calculate the number of threads to use.
-            // If there are 0 workers or an UploadLogsOptions object was not provided, method will run serially. Otherwise will run in parallel with number of workers given.
-            int _maxWorkerCount = (options == null || options.MaxConcurrency <= 0) ? DefaultWorkerCount : options.MaxConcurrency;
+            // If there are 0 workers or an UploadLogsOptions object was not provided, method will run will 5 threads. Otherwise will run in parallel with number of workers given.
+            int _maxWorkerCount = (options == null || options.MaxConcurrency <= 0) ? DefaultParallelWorkerCount : options.MaxConcurrency;
             using var scope = ClientDiagnostics.CreateScope("LogsIngestionClient.Upload");
 
             RequestContext requestContext = GenerateRequestContext(cancellationToken);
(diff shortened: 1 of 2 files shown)


sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -27,8 +27,8 @@ protected LogsIngestionClient()
         // request or stage as multiple blocks.
         internal static int SingleUploadThreshold = 1000000; // 1 Mb in byte format
 
-        // If no concurrency count is provided, default to serial upload (one block at a time).
-        private int DefaultWorkerCount = 1;
+        // If no concurrency count is provided for a parallel upload, default to 5 threads.
```
JoshLove-msft, line 30:
```suggestion
        // If no concurrency count is provided for a parallel upload, default to 5 workers.
```

sdk/monitor/Azure.Monitor.Ingestion/src/Models/UploadLogsOptions.cs
```diff
@@ -18,7 +18,7 @@ public class UploadLogsOptions
 
         /// <summary>
         /// The max concurrent requests to send to the Azure Monitor service when uploading logs.
-        /// <remarks> In the Upload method, this parameter is not used as the batches are uploaded in sequence. </remarks>
+        /// <remarks> In the Upload method, this parameter is not used as the batches are uploaded in sequence. For parallel uploads, if this value is not set the default thread count will be 5. </remarks>
```
JoshLove-msft, line 21:
```suggestion
        /// <remarks> In the Upload method, this parameter is not used as the batches are uploaded in sequence. For parallel uploads, if this value is not set the default concurrency will be 5. </remarks>
```

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -27,8 +27,8 @@ protected LogsIngestionClient()
...
+        private int DefaultParallelWorkerCount = 5;
```
JoshLove-msft, line 31:
```suggestion
        private const int DefaultParallelWorkerCount = 5;
```

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -262,8 +262,8 @@ public virtual async Task<Response<UploadLogsResult>> UploadAsync<T>(string rule
             Argument.AssertNotNullOrEmpty(logs, nameof(logs));
 
             // Calculate the number of threads to use.
-            // If there are 0 workers or an UploadLogsOptions object was not provided, method will run serially. Otherwise will run in parallel with number of workers given.
-            int _maxWorkerCount = (options == null || options.MaxConcurrency <= 0) ? DefaultWorkerCount : options.MaxConcurrency;
+            // If there are 0 workers or an UploadLogsOptions object was not provided, method will run will 5 threads. Otherwise will run in parallel with number of workers given.
```
JoshLove-msft, line 265:
```suggestion
            // If there are 0 workers or an UploadLogsOptions object was not provided, method will run with 5 workers. Otherwise will run in parallel with number of workers given.
```

JoshLove-msft (approved):

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
pallavit: Curious why we chose 5? Is this an ask from the service team or have we run some perf benchmark to decide this number?

nisha-bhatia: 5 is the standard number in .Net. Other libraries like Storage use the same value

pallavit: Got it, thanks.

PR #31706, opened 2022-10-11
Added Quickstart code for Language QuestionAnswering SDK

sehajduggal:
Added QuickStart code for Language QuestionAnswering SDK. Added following operations:

1. Create a project
2. Update a project
3. Deploy a project
4. Getting Project data (Sources, QA pairs and synonyms)
5. Get Answer
6. Delete Project

### DIFF
--- a/sdk/cognitivelanguage/Azure.AI.Language.QuestionAnswering/tests/Samples/QuickStart/Program.cs
+++ b/sdk/cognitivelanguage/Azure.AI.Language.QuestionAnswering/tests/Samples/QuickStart/Program.cs
@@ -1,0 +1,293 @@
+﻿/*
+* ==========================================
+  Install Azure.AI.Language.QuestionAnswering package with command
+* ==========================================
+*
+* dotnet add package Azure.AI.Language.QuestionAnswering
+*
+* ==========================================
+  Tasks Included
+* ==========================================
+* Create a project
+* Update a project
+* Deploy a project
+* Getting Project data (Sources, QA pairs and synonyms)
+* Get Answer
+* Delete Project
+* ==========================================
+  Further reading
+* General documentation: https://learn.microsoft.com/en-us/azure/cognitive-services/language-service/question-answering/overview
+* Reference documentation: https://learn.microsoft.com/en-us/dotnet/api/Azure.AI.Language.QuestionAnswering?view=azure-dotnet
+* ==========================================
+*/
+
+namespace Knowledgebase_Quickstart
+{
+    // Dependencies
+    using Azure;
+    using Azure.Core;
+    using Azure.AI.Language.QuestionAnswering;
+    using Azure.AI.Language.QuestionAnswering.Authoring;
+    using System;
+    using System.Threading.Tasks;
+
+    class Program
+    {
+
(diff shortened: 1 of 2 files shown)


sdk/cognitivelanguage/Azure.AI.Language.QuestionAnswering/tests/Samples/QuickStart/Program.cs
```diff
@@ -0,0 +1,293 @@
...
+            QuestionAnsweringAuthoringClient authoringClient = new QuestionAnsweringAuthoringClient(endpoint, credential);
+            QuestionAnsweringClient inferenceClient = new QuestionAnsweringClient(endpoint, credential);
+
+            // Operations
+            await CreateProject(authoringClient, projectName);
+            await UpdateProject(authoringClient, projectName);
+            await DeployProject(authoringClient, projectName);
+            await GetProjectData(authoringClient, projectName);
+            await GetAnswer(inferenceClient, projectName);
+            await DeleteProject(authoringClient, projectName);
```
heaths, line 63: This sample is not in the format we use, and we already have all these scenarios in our tests/Samples directory, which get inserted automatically into the samples/ directory we already have that shows up in Samples Explorer: https://learn.microsoft.com/samples/azure/azure-sdk-for-net/azureailanguagequestionanswering-samples/. So what do these samples add?

If you're looking to fill gaps we don't have, please see the format we use already.

1. We don't add a new project.
2. We add .cs files under tests/Samples with a naming convention you can see present.
3. We have separate sync and async samples, and use preproc conditions like `#if SNIPPET` to have separate code to run tests or extract snippets.
4. All our tests are buildable/runnable to make sure samples always work. Your code here might get built because the parent .csproj would include the .cs files, but the .csproj added here doesn't get built. But nothing will run this code since it includes a `Main` and isn't test code.

sdk/cognitivelanguage/Azure.AI.Language.QuestionAnswering/tests/Samples/QuickStart/Program.cs
```diff
@@ -0,0 +1,293 @@
...
+        /// <summary>
+        /// Update Project
+        /// </summary>
+        /// <param name="client">Question Answering Authoring Client.</param>
+        /// <param name="projectName">Project name.</param>
+        /// <returns>Task that completes on completion of operation</returns>
+        private static async Task UpdateProject(QuestionAnsweringAuthoringClient client, string projectName)
+        {
+            // Adding Knowledgebase Source
+            var sourceUri = "https://learn.microsoft.com/en-us/azure/cognitive-services/language-service/question-answering/how-to/troubleshooting";
```
heaths, line 75: Please use explicit types in samples.

heaths (changes requested):
This doesn't follow our standards for test samples. See https://github.com/Azure/azure-sdk-for-net/tree/main/samples for more information.

You also include a lot of samples we already had. Please look through our existing samples and either make changes as necessary, or add additional samples in the same style.

sehajduggal:
Abandoning the PR as Samples already exists

PR #31715, opened 2022-10-11
Revert "Exclude the failed package for .NET docs"

sima-zhu:
Reverts Azure/azure-sdk-for-net#31649 as the WebJob.Extension package problem has been resolved.

https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1910244&view=logs&s=96ac2280-8cb4-5df5-99de-dd2da759617d


danieljurek (approved):
Looks good. I filed https://github.com/Azure/azure-sdk-for-net/issues/31783 to track that we should skip such packages.

PR #31735, opened 2022-10-11
[For Discussion] Samples of different ACR APIs for transfer from ACR to storage

annelo-msft:
There are two rough proposed APIs here:

1. **Low-Level:** The [addition of DPG methods](https://apiview.dev/Assemblies/Review/45ef9778411745329b8d66c76cef1c35/b829808a968b49879ef687cc0ad4a794?diffRevisionId=a766972e03c94c1abcf0fc8cbfd5bab4&doc=False&diffOnly=True) to the ACR Blob Client API
2. **High-Level:** The [addition of high level convenience methods](https://apiview.dev/Assemblies/Review/45ef9778411745329b8d66c76cef1c35/ece02c9743d74bc0bd644f4b246c69ed?diffRevisionId=a766972e03c94c1abcf0fc8cbfd5bab4&doc=False&diffOnly=True) to the ACR Blob Client API

The low-level approach allows users to pull down chunks and interleave those downloads with other actions, like upload -- for our customer scenario, that would mean they could be downloading an artifact while simultaneously uploading it to blob storage.  See usage sample here: [TransferInChunks](https://github.com/Azure/azure-sdk-for-net/pull/31735/files#diff-abfb20a7a1d2782c1e8cdaf988a6602d207833cb9ee643039e270391b74d5f3dR18)

The high-level approach provides more convenience by abstracting the details of the download, but would require downloading the entire artifact prior to beginning upload to storage.  See usage sample here: [TransferArtifactAsync](https://github.com/Azure/azure-sdk-for-net/pull/31735/files#diff-cdcef23303fb947945f5856687aebd265f01accd6f871de08a26918e8854dc0dR16)


annelo-msft:
Some thoughts:
- [ ] We should move high-level APIs to the high-level clients
- [ ] Experiment with HLC convenience methods for "missing methods" instead of DPG methods
- [ ] Strawman proposal for models in blob APIs

annelo-msft:
Closing as we have decided on an API to target for Zinc.

PR #31741, opened 2022-10-12
Remove swagger version tag in autorest.md if it's the default

### LINKED ISSUE #30721: Remove tag in autorest.md if it is the same as the default tag in the required swagger readme.md
Having the tag in autorest.md will cause the SDK automation pipeline in the swagger PR of updating the default tag to still generate the .NET SDK with the old tag version.

As the swagger readme.md is forced to be referenced with a commit hash in autorest.md, removing the tag in autorest.md should have no impact in future regeneration if the specified tag is the same with the default tag in the swagger readme.md of the commit. We should only add a tag in autorest.md if we don't want to use the default tag from swagger readme.md.

fengzhou-msft:
Resolve #30721

All SDKs are regenerated locally, and no difference found.


Yao725 (approved):
LGTM

PR #31748, opened 2022-10-12
[EventHubs] Fix broken tests

Yao725:
Open this PR to fix the broken tests in eventhubs as it turns out the sanitizer that we use in eventhubs' test will accidently applies to the tag operation's test recording and make the assertion fail.


fengzhou-msft:
Can you update here as well? The test is ignored, no re-recording needed. But when it's added back, it will have the same issue.
https://github.com/Azure/azure-sdk-for-net/blob/100aedb7e84c9ffb1aaeb3770ad9e57c0d45daca/sdk/eventhub/Azure.ResourceManager.EventHubs/tests/Tests/EventHubClusterTests.cs#L68-L70

fengzhou-msft (approved):

PR #31769, opened 2022-10-12
[For Discussion] Move JsonData to Core

annelo-msft:
### Context

- Addresses https://github.com/Azure/azure-sdk-for-net/issues/30717
- Builds on discussions in https://github.com/Azure/azure-sdk-for-net/pull/31203

### Requirements

The main work needed, given that we have the prototype, is to:

- [x] Remove mutating APIs.
- [x] Decide how arrays work (see above)
  _There is a test showing this capability in the old PR: [CanAccessArrayValues](https://github.com/annelo-msft/azure-sdk-for-net/blob/immutable-jsondata/sdk/core/Azure.Core.Experimental/tests/JsonDataTests.cs#L124-L131)_
- [x] Decide how casts work, e.g. `var date = (DateTime)json;` or cast to custom model: `var model = (FooModel)json.Foo`
  _There is a proposal regarding an approach to enabling this in [this commit](https://github.com/Azure/azure-sdk-for-net/commit/ffc809b717bc294786913e57a492b435acfa30ae#diff-d1e1dffde30bd2fa0bbec8609516edcc0bfe965ed0235b51929d20c7b7422115R717-R728).  It suffers from the issue that it uses `JsonSerializer.Deserialize<T>` to do the conversion between JsonData and T, and we know that not all of our Azure SDK models are serializable via STJ._
- [x] Decide how roundtripping would work. It will not be very easy/usable, but it should be possible.

_Suggested approach (tracked separately by https://github.com/Azure/azure-sdk-for-net/issues/31943):_
```csharp
// ***
// Round-trip: JsonData and Anonymous types
// ***
Response current = await client.GetKeyValueAsync("FontColor");
RequestContent updated = RequestContent.Merge(current, patch:
    new
    {
        value = "updated value";
    }
);
dynamic keyValue = new JsonData(current.Content);
Response putResponse = await client.PutKeyValueAsync((string)keyValue.key, updated, ContentType.ApplicationJson);
```
- [x] Implement a good debugger visualizer

image
_If this isn't "good" yet, let's clarify the ask._
- [ ] Update our protocol method docs
_We will file issues to track this._

### Code sample

Per [this comment](https://github.com/Azure/azure-sdk-for-net/issues/30717#issuecomment-1228706388):

The output of:
```csharp
dynamic json = JsonData.Parse(new BinaryData(new
{
    Foo = new
    {
        Bar = "hi"
    }
}));
Console.WriteLine(json.Foo);
Console.WriteLine(json.Foo.Bar);
Console.WriteLine(json);

dynamic json = JsonData.Parse("5");
Console.WriteLine(((int)json) + 5);

dynamic json = JsonData.Parse("[1, 2, 3, 4, 5, 6]");
Console.WriteLine((int)json[5]);
```
is
```
{"Bar":"hi"}
hi
{"Foo":{"Bar":"hi"}}
10
6
```


KrzysztofCwalina:
> If this isn't "good" yet, let's clarify the ask.

I worry that this view does not scale to large JSON payloads.

sdk/cognitivelanguage/Azure.AI.Language.Conversations/tests/Samples/Sample1_AnalyzeConversation_ConversationPrediction.cs
```diff
@@ -91,7 +89,7 @@ public void AnalyzeConversation()
             #endregion
 
             Assert.That(response.Status, Is.EqualTo(200));
-            Assert.That(conversationPrediction.GetProperty("topIntent").GetString(), Is.EqualTo("Send"));
+            Assert.That((string)conversationPrediction.TopIntent, Is.EqualTo("Send"));
```
annelo-msft, line 92: A string cast is required here for this to succeed right now.

sdk/cognitivelanguage/Azure.AI.Language.Conversations/tests/Samples/Sample6_AnalyzeConversation_ConversationSummarization.cs
```diff
@@ -82,23 +79,22 @@ public void AnalyzeConversation_ConversationSummarization()
...
-                    foreach (JsonElement summary in conversation.GetProperty("summaries").EnumerateArray())
+                    foreach (var summary in conversation.Summaries)
                     {
-                        Console.WriteLine($"Text: {summary.GetProperty("text").GetString()}");
-                        Console.WriteLine($"Aspect: {summary.GetProperty("aspect").GetString()}");
+                        Console.WriteLine($"Text: {summary.Text}");
+                        Console.WriteLine($"Aspect: {summary.Aspect}");
 #if !SNIPPET
-                        aspects.Add(summary.GetProperty("aspect").GetString());
+                        aspects.Add((string)summary.Aspect);
```
annelo-msft, line 97: A string cast is required here to succeed.

sdk/core/Azure.Core/src/JsonData.cs
```diff
@@ -0,0 +1,806 @@
...
+        private List<JsonData>? _arrayRepresentation;
+
+        private static readonly JsonSerializerOptions DefaultJsonSerializerOptions = new JsonSerializerOptions();
+
+        /// <summary>
+        /// Parses a UTF8 encoded string representing a single JSON value into a <see cref="JsonData"/>.
+        /// </summary>
+        /// <param name="utf8Json">A UTF8 encoded string representing a JSON value.</param>
+        /// <returns>A <see cref="JsonData"/> representation of the value.</returns>
+        internal static JsonData Parse(BinaryData utf8Json) => new JsonData(JsonDocument.Parse(utf8Json));
```
annelo-msft, line 42: Handle IDisposable properly.

annelo-msft: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Text.Json/src/System/Text/Json/Document/JsonDocument.cs#L61

https://github.com/dotnet/runtime/blob/main/src/libraries/System.Text.Json/src/System/Text/Json/Document/JsonDocument.Parse.cs#L87

annelo-msft: Per @AlexanderSher:

> I've created a small benchmark. It looks like disposing JsonDocument is quite important:
> https://gist.github.com/AlexanderSher/251a035cfbbe539337e4a1c6504ff52b

sdk/cognitivelanguage/Azure.AI.Language.Conversations/README.md
```diff
@@ -162,39 +162,38 @@ var data = new
...
 Response response = client.AnalyzeConversation(RequestContent.Create(data));
 
-using JsonDocument result = JsonDocument.Parse(response.ContentStream);
-JsonElement conversationalTaskResult = result.RootElement;
-JsonElement conversationPrediction = conversationalTaskResult.GetProperty("result").GetProperty("prediction");
+var json = response.Content.ToDynamic();
+var conversationPrediction = json.Result.Prediction;
 
-Console.WriteLine($"Top intent: {conversationPrediction.GetProperty("topIntent").GetString()}");
+Console.WriteLine($"Top intent: {conversationPrediction.TopIntent}");
```
heaths, line 168: Discussed with @KrzysztofCwalina offline, but I don't think we should automatically convert camelCase JSON properties to PascalCase .NET properties on responses. Requests don't get the same treatment, so this may be confusing or even misleading. Say you get a response back, want to update it, and then send it back (which you also can't since JsonData is effectively immutable):

```C#
var json = response.Content.ToDynamic();
var data = new
{
  Id = json.Id,
  Foo = "new foo",
};
client.UpdateFoo(Request.Create(data)); // throws RequestFailedException: 400 BadParameter or something because the properties are supposed to be camelCase.
```

I appreciate we want to be idiomatic, but as soon as customers are having to deal with raw JSON - via System.Text.Json like my samples were before vs. what you're prototyping here with `JsonData` (which is a remarkable improvement) - I don't think we should "auto-correct" anything. Python nor JS would since customers are working directly with the JSON data. I don't think C# / .NET should either.

Model classes give us this ability, and once we generate them (for Cadl or otherwise) we can and should do that translation because we can do so consistently.

/cc @tg-msft

annelo-msft: I guess I don't understand the thinking there.  We do this in our public models, and the idea is that JsonData is a stand-in until we add those types.  Ideally, the customer would have to change very little to move to any types we added to a library.  It's not that they're dealing with raw Json (which they can do easily if they want) -- it's a layer over the Json (or Xml, or whatever) that makes it conform to a higher .NET standard for look-and-feel.  Is consistency with anonymous types your primary concern here, because I agree that is not good.  Are there any other considerations I'm not thinking of?

annelo-msft: Ah, I see that this is a question about round-tripping the type, my apologies for having missed that the first time.  We are planning to provide a separate API to handle-round tripping, that may be able to address this.  I am still curious to know if there are other considerations here beyond the inconsistency across input/output.

heaths: Assume with my example above that the JSON schema is indeed camelCase, as most services define (and should): the request would accept `id` and `foo`, not `Id` and `Foo`. However, by translating from camelCase to PascalCase we could give callers the impression that the properties are indeed `Id` and `Foo`.

This isn't just about roundtripping, though. It's about perception. If all the response properties are PascalCase, why wouldn't many think that request properties are too?

I appreciate this isn't "raw" JSON. That's not what I intended. But it's a *very thin* wrapper over raw JSON that is meant to provide a JSON-like interaction for the customer, yes? By morphing into something it's not just to seem idiomatic, we create this confusion of casing for customers. Why not just leave it with the exact shape - casing and all - as the service returns (response) and expects (request)?

annelo-msft: That's totally fair - all good points.  Although I feel like the question hinges on how thin the wrapper is and how well we document and illustrate via samples what the type is actually doing.   If, for example, it also wraps other formats like XML, it might be more clear that you can't make too many assumptions about what's underneath.

heaths: If this supports XML, I don't see my point changing: if an XML document uses camelCase or PascalCase (and in my experience it varies even more than with JSON, which is more often camelCase), we should leave casing as is to avoid confusion. If we start showing that customers can always use PascalCase for responses, why wouldn't they think they can use it for input as well? Or if they use the `obj["foo"]["bar"]` pattern it will also vary from `obj.Foo.Bar`. Changing cases like that just seems so jarring.

What benefit to the customer does translating the casing have? I can certainly see a lot of cons. The only pro is "it feels like C#", but they aren't working with generated types (C# or otherwise): they are working with thin wrappers over raw JSON, XML, etc.

tg-msft: Should `ToDynamic` take an optional `ObjectSerailizer` so folks can customize this?  I feel like that would easily allow this to work for people who wanted it, but then we could get away with less magic by default if we didn't change the case.

sdk/core/Azure.Core/src/JsonData.cs
annelo-msft: @AlexanderSher, does this address your concern on this? https://github.com/Azure/azure-sdk-for-net/pull/31769/commits/795e2ff92cc19240f588d6595594e5b661ea2c08

PR #31790, opened 2022-10-13
NullableResponse<T> and consumption by tables

### LINKED ISSUE #27852: Add `NullableResponse<T>` that `Response<T>` inherits from
`Response<T>` requires that `T Value` is not null.  When we want to work around that and return a `Response<T>` with no `Value` _(like conditional requests that return 3XX, batching requests that haven't been sent to the service yet, DPG errors that were classified as success, ...)_ we usually `throw` when accessed.  See [NoBodyResponse](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/Shared/NoBodyResponseOfT.cs) and [ErrorResponse](https://github.com/Azure/autorest.csharp/blob/feature/v3/src/assets/Generator.Shared/ErrorResponse.cs) as examples.

While that _technically_ meets the contract for `Response<T>`, it doesn't match the spirit of it.  We'd like to introduce a `NullableResponse<T>` that we can start using for this case.  To avoid a bifurcated response hierarchy, we'd like to make the existing `Response<T>` inherit from it.  Here's a rough example of what the API might look like:

```C#
public abstract class NullableResponse<T> {
    public abstract bool HasValue { get; }
    public abstract T Value { get;  }
    public abstract Response GetRawResponse() {}
}
 
public abstract class Response<T> : NullableResponse<T> {
     public override bool HasValue => true;
     public static implicit operator T(Response<T> response)
}
```

We also want to evaluate whether DPG protocol methods (and anything accepting `RequestContext`) should return this `NullableResponse<T>` instead of regular `Response<T>`.

christothes:
closes #27852


sdk/appconfiguration/Azure.Data.AppConfiguration/src/Azure.Data.AppConfiguration.csproj
```diff
@@ -9,11 +9,12 @@
...
     <TargetFrameworks>$(RequiredTargetFrameworks)</TargetFrameworks>
     <NoWarn>$(NoWarn);3021;AZC0007;AZC0006</NoWarn>
-    <IncludeOperationsSharedSource>true</IncludeOperationsSharedSource> 
+    <IncludeOperationsSharedSource>true</IncludeOperationsSharedSource>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Azure.Core" />
+    <!-- <PackageReference Include="Azure.Core" /> -->
+    <ProjectReference Include="..\..\..\core\Azure.Core\src\Azure.Core.csproj" />
```
jsquire, line 17: I'm not sure that I follow why this is needed.  Did something change that I'm overlooking?

sdk/core/Azure.Core/CHANGELOG.md
```diff
@@ -191,7 +192,7 @@ options.Diagnostics.LoggedQueryParameters.Remove("api-version");
 - `AzureSasCredential` and its respective policy.
 
 ### Key Bug Fixes
-- Avoid a causing and ignoring an exception when setting network stream timeout on .NET Core 
+- Avoid a causing and ignoring an exception when setting network stream timeout on .NET Core
```
jsquire, line 195:
```suggestion
- Avoid a causing and ignoring an exception when setting network stream timeout on .NET Core.
```

sdk/core/Azure.Core/src/NullableResponseOfT.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// <inheritdoc />
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public override bool Equals(object? obj) => base.Equals(obj);
+
+        /// <inheritdoc />
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public override int GetHashCode() => base.GetHashCode();
+
+        /// <inheritdoc />
+        public override string ToString()
```
jsquire, line 41: Out of curiosity, why expression-bodied methods above and the long form here?

sdk/storage/Azure.Storage.Common/src/Azure.Storage.Common.csproj
```diff
@@ -20,7 +20,8 @@
     <IncludeOperationsSharedSource>true</IncludeOperationsSharedSource>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Azure.Core" />
+    <!-- <PackageReference Include="Azure.Core" /> -->
+    <ProjectReference Include="..\..\..\core\Azure.Core\src\Azure.Core.csproj" />
```
jsquire, line 24: Not sure that I understand why this is needed.

sdk/tables/Azure.Data.Tables/src/TableClient.cs
```diff
@@ -700,10 +700,12 @@ public virtual Response<T> GetEntity<T>(string partitionKey, string rowKey, IEnu
         /// <param name="rowKey">The rowKey that identifies the table entity.</param>
         /// <param name="select">Selects which set of entity properties to return in the result set. Pass <c>null</c> to retreive all properties.</param>
         /// <param name="cancellationToken">A <see cref="CancellationToken"/> controlling the request lifetime.</param>
-        /// <returns>The <see cref="Response"/> indicating the result of the operation.</returns>
-        /// <exception cref="RequestFailedException">Exception thrown if the entity doesn't exist.</exception>
+        /// <returns> NullableResponse of T whose HasValue property will return <c>true</c> if the entity existed, otherwise <c>false</c>.</returns>
```
jsquire, line 703:
```suggestion
        /// <returns> The <see cref="NullableResponse{T}" whose <c>HasValue</c> property will return <c>true</c> if the entity existed, otherwise <c>false</c>.</returns>
```

sdk/tables/Azure.Data.Tables/src/TableClient.cs
```diff
@@ -715,10 +717,12 @@ public virtual Response<T> GetEntityIfExists<T>(string partitionKey, string rowK
         /// <param name="rowKey">The rowKey that identifies the table entity.</param>
         /// <param name="select">Selects which set of entity properties to return in the result set. Pass <c>null</c> to retreive all properties.</param>
         /// <param name="cancellationToken">A <see cref="CancellationToken"/> controlling the request lifetime.</param>
-        /// <returns>The <see cref="Response"/> indicating the result of the operation.</returns>
+        /// <returns> NullableResponse of T whose HasValue property will return <c>true</c> if the entity existed, otherwise <c>false</c>.</returns>
```
jsquire, line 720:
```suggestion
        /// <returns> The <see cref="NullableResponse{T}" whose <c>HasValue</c> property will return <c>true</c> if the entity existed, otherwise <c>false</c>.</returns>
```

sdk/tables/Azure.Data.Tables/src/TableClient.cs
```diff
@@ -715,10 +717,12 @@ public virtual Response<T> GetEntityIfExists<T>(string partitionKey, string rowK
...
         /// <exception cref="RequestFailedException">Exception thrown if the entity doesn't exist.</exception>
         /// <exception cref="ArgumentNullException"><paramref name="partitionKey"/> or <paramref name="rowKey"/> is null.</exception>
-        public virtual async Task<Response<T>> GetEntityIfExistsAsync<T>(string partitionKey, string rowKey, IEnumerable<string> select = null, CancellationToken cancellationToken = default)
+#pragma warning disable AZC0015 // Unexpected client method return type.
```
jsquire, line 723: Should we consider opening an issue to tweak this analyzer?   Seems like it may need awareness of the new type.

sdk/core/Azure.Core/CHANGELOG.md
```diff
@@ -229,7 +230,7 @@ options.Diagnostics.LoggedQueryParameters.Remove("api-version");
 ```xml
  <ItemGroup>
     <RuntimeHostConfigurationOption Include="Azure.Core.Pipeline.DisableHttpWebRequestTransport" Value="true" />
-  </ItemGroup> 
+  </ItemGroup>
```
annelo-msft, line 233: nit: have we made sure IDE settings across the team aren't going to cause whitespace edit wars across changes?

sdk/core/Azure.Core/src/NullableResponseOfT.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            return $"Status: {GetRawResponse().Status}, Value: {Value}";
```
annelo-msft, line 43: Should this check Value for null?

sdk/tables/Azure.Data.Tables/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - Added a `Uri` property to `TableClient` and `TableServiceClient`
 
 ### Breaking Changes
+- `TableClient.GetEntityIfNotExists` now returns `NullableResponse<T>` which has a `HasValue` property which returns false when the entity did not exist. Accessing the `Value` property in this case will throw an exception.
```
annelo-msft, line 9:
```suggestion
- `TableClient.GetEntityIfExists` now returns `NullableResponse<T>` which has a `HasValue` property which returns false when the entity did not exist. Accessing the `Value` property in this case will throw an exception.
```

sdk/storage/Azure.Storage.Common/src/Azure.Storage.Common.csproj
jsquire: But there's no explicit use of the new features/types here.   That's why I'm asking.  😄

sdk/tables/Azure.Data.Tables/api/Azure.Data.Tables.netstandard2.0.cs
```diff
@@ -37,8 +37,8 @@ public partial class TableClient
...
-        public virtual System.Threading.Tasks.Task<Azure.Response<T>> GetEntityIfExistsAsync<T>(string partitionKey, string rowKey, System.Collections.Generic.IEnumerable<string> select = null, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) where T : class, Azure.Data.Tables.ITableEntity, new() { throw null; }
-        public virtual Azure.Response<T> GetEntityIfExists<T>(string partitionKey, string rowKey, System.Collections.Generic.IEnumerable<string> select = null, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) where T : class, Azure.Data.Tables.ITableEntity, new() { throw null; }
+        public virtual System.Threading.Tasks.Task<Azure.NullableResponse<T>> GetEntityIfExistsAsync<T>(string partitionKey, string rowKey, System.Collections.Generic.IEnumerable<string> select = null, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) where T : class, Azure.Data.Tables.ITableEntity, new() { throw null; }
```
annelo-msft, line 40: Isn't this an unacceptable breaking change, per [D.11.6](https://learning.oreilly.com/library/view/framework-design-guidelines/9780135896457/app04.xhtml#secD_11)?

I think we may be stuck with the existing method the way it is, but we can follow the new guidance for new methods going forward.  Or maybe architects would approve this breaking change this once.

sdk/tables/Azure.Data.Tables/src/TableClient.cs
```diff
@@ -780,16 +784,15 @@ public virtual async Task<Response<T>> GetEntityIfExistsAsync<T>(string partitio
                     selectArg,
                     filter: null,
                     context);
-                if (!response.IsError)
+                if (response.Status == (int)HttpStatusCode.NotFound)
                 {
-                    var dictionary = SerializationHelpers.ResponseToDictionary(response);
-                    var result = dictionary.ToTableEntity<T>();
-                    return Response.FromValue(result, response);
+                    return new NoBodyResponse<T>(response);
```
annelo-msft, line 789: I would have expected to see us return an internal type that derived directly from NullableResponse<T> here?  It feels awkward that we return a type that hides and then reinstates "HasValue".

annelo-msft (commented):
We should add tests to Core for the new type.

sdk/core/Azure.Core/src/ResponseOfT.cs
```diff
@@ -14,20 +14,15 @@ namespace Azure
...
 #pragma warning restore AZC0012 // Avoid single word type names
 #pragma warning restore SA1649 // File name should match first type name
     {
-        /// <summary>
-        /// Returns the HTTP response returned by the service.
-        /// </summary>
-        /// <returns>The HTTP response returned by the service.</returns>
-        public abstract Response GetRawResponse();
-
         /// <summary>
```
JoshLove-msft, line 21: nit: Update refdocs to refer to HasValue (or use inheritdoc)

sdk/appconfiguration/Azure.Data.AppConfiguration/src/Azure.Data.AppConfiguration.csproj
christothes: Users of `NoResponseBody` need to take these changes from Core for it to compile.

sdk/storage/Azure.Storage.Common/src/Azure.Storage.Common.csproj
christothes: Same comment as above.

sdk/tables/Azure.Data.Tables/src/TableClient.cs
christothes: https://github.com/Azure/azure-sdk-tools/issues/4444

sdk/appconfiguration/Azure.Data.AppConfiguration/src/Azure.Data.AppConfiguration.csproj
jsquire: Ah, gotcha.   Indirect use due to the new hierarchy.

sdk/tables/Azure.Data.Tables/api/Azure.Data.Tables.netstandard2.0.cs
christothes: We haven't GA'd the current method yet.

sdk/tables/Azure.Data.Tables/src/TableClient.cs
christothes: I'm not sure it's worth adding a new type unless the API behavior would be different. Is there something that adding a dedicated type here would add that I'm missing?

sdk/tables/Azure.Data.Tables/src/TableClient.cs
christothes: Thanks - I couldn't remember what the generic syntax was.

sdk/tables/Azure.Data.Tables/src/TableClient.cs
annelo-msft: It's what it would subtract.  I could cast this to Response<T> or ask `response is Response<T>` and reasonably assume I'd gotten a value for T, since that's what the semantics of Response<T> imply, and it hides `HasValue`.  If we're going to return a Response<T> when T is null, it feels like we're undoing all the work we've done to add a type that indicates the T could be null.  But I'm fine if you wanted to get an architect perspective to see if I'm being too nit-picky.

sdk/core/Azure.Core/src/NullableResponseOfT.cs
annelo-msft: I was more worried about `Value` - can't this throw if `HasValue` is false?

sdk/tables/Azure.Data.Tables/src/TableClient.cs
christothes: While it seems like an edge case, it's a good point that someone could cast this to `Response<T>` and get bad hints. 

Would you name the internal type `NullResponse<T>`, `EmptyResponse<T>`, `NoValueResponse<T>`, (any other ideas)?

sdk/core/Azure.Core/src/NullableResponseOfT.cs
christothes: Agreed - it's covered in the latest changes.

sdk/tables/Azure.Data.Tables/src/TableClient.cs
jsquire: What about `NoContentResponse<T>` - that would seem to align with the official nomenclature for an HTTP 204

christothes: After introducing this new internal response type, I think `NoBodyResponse<T>` should effectively be deprecated moving forward. 

Public APIs that already return it are stuck (it's only a hand full of APIs that use it today), but going forward, it would have the same problem as called out here.

annelo-msft: Cool.  I was going to vote for `NullResponse<T>`, to match the corresponding `ValueResponse<T>` and parent, but I don't feel super strongly since it's an internal API.

sdk/core/Azure.Core/src/NullableResponseOfT.cs
annelo-msft: Not sure how I missed that.

sdk/core/Azure.Core/src/NullableResponseOfT.cs
```diff
@@ -0,0 +1,45 @@
...
+{
+    /// <summary>
+    /// Represents a result of Azure operation.
+    /// </summary>
+    /// <typeparam name="T">The type of returned value.</typeparam>
+#pragma warning disable SA1649 // File name should match first type name
+    public abstract class NullableResponse<T>
+#pragma warning restore SA1649 // File name should match first type name
+    {
+        private const string NoValue = "The Response has no value";
```
JoshLove-msft, line 16: Why is this preferable to just leaving it as empty string?

christothes: I think it is more explicit - similar to "this page intentionally blank".

annelo-msft: Another convention we have in the code base is to return `"<null>"` ... if this is sanctioned, it might read nicely in the NullableResponse type

JoshLove-msft: Yeah I would stick with either empty string or "<null>". If users need to know if the value is really null, they have to check HasValue anyway so it doesn't help much to try to make the string representation convey that.

JoshLove-msft: For that matter, should we include HasValue in the ToString here, and add back the override of ToString into ResponseOfT for back compat?

christothes: I think the value of HasValue is implied by this message. What's the back-compat scenario you had in mind? To restore the throwing behavior, we had before?

sdk/tables/Azure.Data.Tables/src/TableClient.cs
JoshLove-msft: Seems like we are just trying to represent a NullableResponse that is, in fact, null. So NullResponse seems to convey that. We should probably seal this type so nobody tries to change the semantics.

sdk/core/Azure.Core/src/NullableResponseOfT.cs
JoshLove-msft: No, I meant that if we include HasValue in the ToString for this type, we probably would want to NOT include it for ResponseOfT to retain the existing format.

sdk/tables/Azure.Data.Tables/src/TableClient.cs
christothes: My current vote is `NoValueResponse<T>` since "nullable" is a misnomer we only used due to the similarity with the Nullable<T> API. But in implementation, what we are needing to do is create counterpart to `ValueResponse<T>`

JoshLove-msft: That makes sense to me.

sdk/core/Azure.Core/src/NullableResponseOfT.cs
christothes: But the format is effectively un-changed. `Response<T>` must always have a value.

JoshLove-msft: Yes, agreed. I'm saying if we were to add HasValue into the ToString of NullableResponseOfT, we would not want to include it in the ToString of ResponseOfT.

sdk/core/Azure.Core/src/Shared/NoBodyResponseOfT.cs
```diff
@@ -16,6 +16,9 @@ public NoBodyResponse(Response response)
             _response = response;
         }
 
+        /// <inheritdoc />
```
JoshLove-msft, line 19: Please add a comment saying this should not be used for new code.

christothes: I was going to do that, but I think it would be more effective to tweak the analyzer rule instead.

JoshLove-msft: I didn't realize that we did this.

JoshLove-msft: Since we've filed an issue and haven't addressed it yet, should we at least add a comment as a safeguard?

JoshLove-msft (approved):

sdk/core/Azure.Core/src/Shared/NoBodyResponseOfT.cs
christothes: added

sdk/core/Azure.Core/src/Shared/NoValueResponseOfT.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        public override Response GetRawResponse() => _response;
+
+        public override string ToString()
+        {
+            return $"Status: {GetRawResponse().Status}, Service returned no content";
+        }
+
+#pragma warning disable CA1064 // Exceptions should be public
+        private class ResponseBodyNotFoundException : Exception
```
annelo-msft, line 42: This name seems to correspond to the `NoBodyResponse` above ... it's possible that a `NullableResponse<T>` could have a body, but no value (i.e. if the payload failed to deserialize, or if the body contained an error).  It seems like another name might be better, or would it make sense to return an InvalidOperationException, since that's what `Nullable<T>` does?

sdk/core/Azure.Core/tests/NullableResponseOfTTests.cs
```diff
@@ -0,0 +1,55 @@
...
+        [Test]
+        public void DoesNotThrowWhenValueIsAccessedWithValue()
+        {
+            var target = new TestValueResponse<string>(new MockResponse(200), "test");
+            Assert.AreEqual("test", target.Value);
+            Assert.That(target.ToString(), Does.Not.Contain("<null>"));
+            Assert.That(target.ToString(), Does.Contain("test"));
+        }
+
+        private class TestValueResponse<T> : NullableResponse<T>
```
annelo-msft, line 29: Would it make sense to also add tests for `NoValueResponse<T>`?

sdk/tables/Azure.Data.Tables/src/TableClient.cs
```diff
@@ -685,12 +688,12 @@ public virtual Response<T> GetEntity<T>(string partitionKey, string rowKey, IEnu
...
-            string partitionKey,
-            string rowKey,
-            IEnumerable<string> select = null,
-            CancellationToken cancellationToken = default) where T : class, ITableEntity, new()
-            => await GetEntityInternalAsync<T>(true, partitionKey, rowKey, false, select, cancellationToken).ConfigureAwait(false);
+        public virtual async Task<Response<T>> GetEntityAsync<T>(string partitionKey, string rowKey, IEnumerable<string> select = null, CancellationToken cancellationToken = default)
+                where T : class, ITableEntity, new()
+        {
+            NullableResponse<T> response = await GetEntityInternalAsync<T>(true, partitionKey, rowKey, false, select, cancellationToken).ConfigureAwait(false);
+            return Response.FromValue(response.Value, response.GetRawResponse());
```
annelo-msft, line 695: Could this throw if response.Value is null?
If we know it won't (e.g. it's a `Response<T>`, would it make sense to downcast to `Response<T>` instead so as not to create a new instance?

christothes: It will never be null. If the entity did not exist, this would have thrown in `GetEntityInternalAsync`

sdk/core/Azure.Core/src/Shared/NoValueResponseOfT.cs
christothes: > it's possible that a NullableResponse<T> could have a body, but no value (i.e. if the payload failed to deserialize, or if the body contained an error).

I don't think this type is intended to represent those scenarios. We should throw in those examples.

>  It seems like another name might be better, or would it make sense to return an InvalidOperationException, since that's what Nullable<T> does?

I don't feel strongly about the name since it's private, but I like using a known exception type as you suggest.

jsquire (approved):

heaths:
FWIW, this broke Key Vault's sample, "sharelink". @weshaggard, @sima-zhu, @JoshLove-msft I know you've been making changes with the pipelines, but seems in those changes we no longer build as much - like the Key Vault samples that used to build. Once this change was merged, the `net - keyvault` and `net - keyvault - tests` pipelines started breaking. Maybe autorest updates should still be running all pipelines before merging? Seems we're missing coverage.

heaths:
I opened #32064 to track instead of commenting on this issue.

PR #9782, opened 2020-02-03
Sanitize header values in exceptions

pakrym:
1. Split redaction logic into HttpMessageSanitizer
2. Moves RequestFailedException creation logic into client diagnostics and uses HttpMessageSanitizer there
3. Adds additional information and inner exception support to ClientDiagnostics
4. Switches storage to use ClientDiagnostics instead of custom implementation

Fixes: https://github.com/Azure/azure-sdk-for-net/issues/9739

PR #31793, opened 2022-10-13
Couple of small perf tweaks

AlexanderSher:
dotnet C:\GitHub\azure-sdk-for-net\artifacts\bin\Azure.Storage.Blobs.Perf\Release\net6.0\Azure.Storage.Blobs.Perf.dll DownloadBlob --size 16384 --parallel 1 --warmup 60 --duration 60

Before:
Completed 1,756 operations in a weighted-average of 59.97s (29.28 ops/s, 0.03415 s/op, 0.23% CPU)

After:
Completed 1,836 operations in a weighted-average of 59.99s (30.60 ops/s, 0.03268 s/op, 0.059% CPU)

### DIFF
--- a/sdk/core/Azure.Core/src/RequestUriBuilder.cs
+++ b/sdk/core/Azure.Core/src/RequestUriBuilder.cs
@@ -278,16 +278,7 @@ namespace Azure.Core
                 stringBuilder.Append(PathSeparator);
             }
 
-            // TODO: Escaping can be done in-place
-            if (!HasQuery)
-            {
-                stringBuilder.Append(_pathAndQuery);
-            }
-            else
-            {
-                stringBuilder.Append(_pathAndQuery.ToString(0, _queryIndex));
-                stringBuilder.Append(_pathAndQuery.ToString(_queryIndex, _pathAndQuery.Length - _queryIndex));
-            }
+            stringBuilder.Append(_pathAndQuery);
 
             return stringBuilder.ToString();
         }
--- a/sdk/storage/Azure.Storage.Common/src/Shared/StorageSharedKeyPipelinePolicy.cs
+++ b/sdk/storage/Azure.Storage.Common/src/Shared/StorageSharedKeyPipelinePolicy.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
 using System.Net.Http.Headers;
@@ -76,6 +77,7 @@ namespace Azure.Storage
                 contentLengthString = contentLength.ToString(CultureInfo.InvariantCulture);
             }
             var uri = message.Request.Uri.ToUri();
+
             var stringBuilder = new StringBuilder(uri.AbsolutePath.Length + 64);
             stringBuilder.Append(message.Request.Method.ToString().ToUpperInvariant()).Append('\n');
             stringBuilder.Append(contentEncoding ?? "").Append('\n');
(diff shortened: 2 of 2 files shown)


sdk/storage/Azure.Storage.Common/src/Shared/StorageSharedKeyPipelinePolicy.cs
```diff
@@ -99,18 +101,24 @@ private static void BuildCanonicalizedHeaders(StringBuilder stringBuilder, HttpM
...
+            var headers = new List<HttpHeader>();
+            foreach (var header in message.Request.Headers)
+            {
+                if (header.Name.StartsWith(Constants.HeaderNames.XMsPrefix, StringComparison.OrdinalIgnoreCase))
+                {
+                    headers.Add(header);
+                }
+            }
+
+            headers.Sort(static (x, y) => string.CompareOrdinal(x.Name, y.Name));
```
christothes, line 113: Is this refactor of the sorting and enumeration needed below in `BuildCanonicalizedResource` also?

christothes (approved):

sdk/core/Azure.Core/src/RequestUriBuilder.cs
```diff
@@ -278,16 +278,7 @@ public override string ToString()
                 stringBuilder.Append(PathSeparator);
             }
 
-            // TODO: Escaping can be done in-place
```
JoshLove-msft, line 281: What is the impact of this change?

sdk/storage/Azure.Storage.Common/src/Shared/StorageSharedKeyPipelinePolicy.cs
```diff
@@ -99,18 +101,24 @@ private static void BuildCanonicalizedHeaders(StringBuilder stringBuilder, HttpM
...
+
+            foreach (var header in headers)
             {
-                stringBuilder.Append(header.Item1);
-                stringBuilder.Append(':');
-                stringBuilder.Append(header.Value);
-                stringBuilder.Append('\n');
+                stringBuilder
+                    .Append(header.Name.ToLowerInvariant())
```
JoshLove-msft, line 118: Don't we need to call Trim to keep the same transformations?

sdk/storage/Azure.Storage.Common/src/Shared/StorageSharedKeyPipelinePolicy.cs
```diff
@@ -99,18 +101,24 @@ private static void BuildCanonicalizedHeaders(StringBuilder stringBuilder, HttpM
...
+                    .Append(':')
+                    .Append(header.Value)
```
JoshLove-msft, line 120: Don't we need to call Trim here to keep parity with the old behavior?

sdk/storage/Azure.Storage.Common/src/Shared/StorageSharedKeyPipelinePolicy.cs
JoshLove-msft, line 113: Should we call Trim here to maintain existing behavior?

AlexanderSher: No, we don't. All headers start with `Constants.HeaderNames.XMsPrefix`, so `Trim` is essentially `TrimEnd`, which doesn't affect sorting.

AlexanderSher: Maybe. `BuildCanonicalizedResource` is not visible in the profiler, probably because I don't have that many query parameters in test.

sdk/core/Azure.Core/src/RequestUriBuilder.cs
AlexanderSher: Two years ago, this code look like this: 
```cs
if (!HasQuery)
{
    stringBuilder.Append(_pathAndQuery);
}
else
{
    stringBuilder.Append(_pathAndQuery.ToString(0, _queryIndex));
    if (allowedQueryParameters == null)
    {
        stringBuilder.Append(_pathAndQuery.ToString(_queryIndex, _pathAndQuery.Length - _queryIndex));
    }
    else
    {
        AppendRedactedQuery(stringBuilder, allowedQueryParameters, redactedValue);
    }
}
```

It was changed in this PR: https://github.com/Azure/azure-sdk-for-net/pull/9782/files#diff-890ad460c6bc0ff005f2f636abc7fc7ddde1ef191a2f5407c4193e2ca31fcb28L284-L294

sdk/storage/Azure.Storage.Common/src/Shared/StorageSharedKeyPipelinePolicy.cs
JoshLove-msft: In that case, we should update the [comment](https://github.com/Azure/azure-sdk-for-net/pull/31793/files/cc2e010d0f1f60932eed969bd71467c0420c9722#diff-2ca4e8d05dd448866d56400a1db672172908ed1d538cb52e4d177690ba9df139R102) that explains we trim whitespace.

PR #31802, opened 2022-10-13
[Event Hubs] Tune processor ownership docs

jsquire:
# Summary

The focus of these changes is to add guidance for choosing the number of processor instances to the configuration sample and reduce the frequency that the "owns too many partitions" warning is logged.  It will now occur only when the owned partition count changes, rather than on each load balancing cycle.

### DIFF
--- a/sdk/eventhub/Azure.Messaging.EventHubs.Processor/CHANGELOG.md
+++ b/sdk/eventhub/Azure.Messaging.EventHubs.Processor/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 ### Other Changes
 
-- Adjusted the frequency that a warning logged when the processor owns more partitions than a basic heuristic believes is ideal.  Warnings will no longer log on each load balancing cycle, only when the number of partitions owned changes.
+- Adjusted the frequency that a warning is logged when the processor owns more partitions than a basic heuristic believes is ideal.  Warnings will no longer log on each load balancing cycle, only when the number of partitions owned changes.
 
 ## 5.7.3 (2022-10-11)
 
--- a/sdk/eventhub/Azure.Messaging.EventHubs.Processor/CHANGELOG.md
+++ b/sdk/eventhub/Azure.Messaging.EventHubs.Processor/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ### Other Changes
 
+- Adjusted the frequency that a warning logged when the processor owns more partitions than a basic heuristic believes is ideal.  Warnings will no longer log on each load balancing cycle, only when the number of partitions owned changes.
+
 ## 5.7.3 (2022-10-11)
 
 ### Other Changes
--- a/sdk/eventhub/Azure.Messaging.EventHubs.Processor/samples/Sample02_EventProcessorConfiguration.md
+++ b/sdk/eventhub/Azure.Messaging.EventHubs.Processor/samples/Sample02_EventProcessorConfiguration.md
@@ -6,7 +6,7 @@ To begin, please ensure that you're familiar with the items discussed in the [Ge
 
 ## Table of contents
 
-- [Choosing the number of processors for the consumer group](choosing-the-number-of-processors-for-the-consumer-group)
+- [Choosing the number of processors for the consumer group](#choosing-the-number-of-processors-for-the-consumer-group)
 - [Influencing load balancing behavior](#influencing-load-balancing-behavior)
     - [Load balancing strategy](#load-balancing-strategy)
     - [Load balancing intervals](#load-balancing-intervals)
--- a/sdk/eventhub/Azure.Messaging.EventHubs.Processor/samples/Sample02_EventProcessorConfiguration.md
+++ b/sdk/eventhub/Azure.Messaging.EventHubs.Processor/samples/Sample02_EventProcessorConfiguration.md
@@ -6,6 +6,7 @@ To begin, please ensure that you're familiar with the items discussed in the [Ge
 
 ## Table of contents
 
+- [Choosing the number of processors for the consumer group](choosing-the-number-of-processors-for-the-consumer-group)
 - [Influencing load balancing behavior](#influencing-load-balancing-behavior)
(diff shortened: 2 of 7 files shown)


sdk/eventhub/Azure.Messaging.EventHubs.Processor/CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
 
 ### Other Changes
 
+- Adjusted the frequency that a warning logged when the processor owns more partitions than a basic heuristic believes is ideal.  Warnings will no longer log on each load balancing cycle, only when the number of partitions owned changes.
```
christothes, line 13:
```suggestion
- Adjusted the frequency that a warning is logged when the processor owns more partitions than a basic heuristic believes is ideal.  Warnings will no longer log on each load balancing cycle, only when the number of partitions owned changes.
```

sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/EventProcessor.cs
```diff
@@ -2027,12 +2032,18 @@ private async Task ValidateStorageConnectionAsync(CancellationToken cancellation
...
         /// <returns>The maximum number of partitions that it is advised this processor own.</returns>
         ///
-        private static int CalculateMaximumAdvisedOwnedPartitions() => (int)Math.Floor(Environment.ProcessorCount * 1.5);
+        /// <remarks>
+        ///   The safe number of partitions owned will vary quite a bit by application logic, event structure, and the
+        ///   host environment.  This is a general approximation based on support issues observed since the initial
+        ///   processor release.
+        /// </remarks>
+        ///
+        private static int CalculateMaximumAdvisedOwnedPartitions() => (Environment.ProcessorCount * 2);
```
christothes, line 2046: Did I read incorrectly, or do the docs say this should be 3?
```suggestion
        private static int CalculateMaximumAdvisedOwnedPartitions() => (Environment.ProcessorCount * 3);
```

jsquire: It's intentionally 2 here.   The maximum that we recommend is 3, but I want to issue a warning at 2 that there may be problems.

sdk/eventhub/Azure.Messaging.EventHubs.Processor/samples/Sample02_EventProcessorConfiguration.md
```diff
@@ -6,6 +6,7 @@ To begin, please ensure that you're familiar with the items discussed in the [Ge
 
 ## Table of contents
 
+- [Choosing the number of processors for the consumer group](choosing-the-number-of-processors-for-the-consumer-group)
```
jsquire, line 9:
```suggestion
- [Choosing the number of processors for the consumer group](#choosing-the-number-of-processors-for-the-consumer-group)
```

sdk/eventhub/Azure.Messaging.EventHubs/TROUBLESHOOTING.md
```diff
@@ -227,7 +227,7 @@ The event processor works in a concurrent and highly asynchronous manner.  Each
 
 When a processor owns too many partitions, it will often experience contention in the thread pool leading to starvation.  During this time, continuations will start to queue while waiting to be scheduled causing stalls in the processor.  Because there is no fairness guarantee in scheduling, some partitions may appear to stop processing or load balancing may not be able to update ownership, causing partitions to "bounce" between owners.
 
-It is generally recommended that an event processor own no more than three partitions for every 1 CPU core of the host.  It is often helpful to start a ratio of 1.5 partitions for each CPU core and test increasing the number of owned partitions gradually to measure what works best for the specific application.
+Generally, it is recommended that an event processor own no more than 3 partitions for every 1 CPU core of the host.  Since the ratio will vary for each application, it is often helpful to start with using 1.5 partitions for each CPU core and test increasing the number of owned partitions gradually to measure what works best for your application.
```
m-redding, line 230:
```suggestion
Generally, it is recommended that an event processor own no more than 3 partitions for every 1 CPU core of the host.  Since the ratio will vary for each application, it is often helpful to start with 1.5 partitions for each CPU core and increase the number of owned partitions gradually to determine what works best for your application.
```

sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/EventProcessor.cs
```diff
@@ -1588,12 +1593,12 @@ private async Task RunProcessingAsync(CancellationToken cancellationToken)
                         _ = InvokeOnProcessingErrorAsync(slowException, null, Resources.OperationEventProcessingLoop, CancellationToken.None);
                     }
 
-                    // If the number of partitions owned is maximum advisable set, emit a warning.  This is an inaccurate heuristic and does
-                    // not apply to all workloads.  The error handler will not be pinged to avoid false positive notifications.
+                    // If the number of partitions owned has changed since the cycle started and is above maximum advisable set, emit a warning.  This is
+                    // an inaccurate heuristic and does not apply to all workloads.  The error handler will not be pinged to avoid false positive notifications.
```
m-redding, line 1597: Inaccurate seems like it could be a strong word, like it's always off

m-redding (approved):

sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/EventProcessor.cs
jsquire, line 1597:
```suggestion
                     // If the number of partitions owned has changed since the cycle started and is above maximum advisable set, emit a warning.  This is
                    // a loose heuristic and does not apply to all workloads.  The error handler will not be pinged to avoid false positive notifications.
```

jsquire: Fair enough; I softened it a bit.

Issue #31730, opened 2022-10-11
Add live test support to Azure.Monitor.Ingestion with bicep file


nisha-bhatia:
After this is completed we can separate out the monitor pipeline by SDK for live tests

vishweshbankwar:
@nisha-bhatia - will this resolve https://github.com/Azure/azure-sdk-for-net/issues/31727?

nisha-bhatia:
> 

Yes it should.

vishweshbankwar:
> > 
> 
> Yes it should.

Would you be able to disable the ingestion live tests in the meantime since it fails the entire pipeline?

nisha-bhatia:
The live test pipeline has been created

PR #31803, opened 2022-10-13
[Azure.Monitor.Ingestion] update Bicep file for live testing

nisha-bhatia:


nisha-bhatia:
https://github.com/Azure/azure-sdk-for-net/issues/31730

sdk/monitor/Azure.Monitor.Ingestion/tests/MonitorIngestionTestEnvironment.cs
```diff
@@ -1,22 +1,20 @@
...
     public class MonitorIngestionTestEnvironment : TestEnvironment
     {
-        public string DCRImmutableId => GetRecordedVariable("AZURE_MONITOR_INGESTION_LOGS_DCR_RULE_ID");
+        public string DCRImmutableId => GetRecordedVariable("INGESTION_DATA_COLLECTION_RULE_IMMUTABLE_ID");
 
-        public string DCREndpoint => GetRecordedVariable("AZURE_MONITOR_INGESTION_DATA_COLLECTION_ENDPOINT");
+        public string DCREndpoint => GetRecordedVariable("INGESTION_DATA_COLLECTION_ENDPOINT");
 
-        public string Ingestion_WorkspaceId => GetRecordedVariable("AZURE_MONITOR_INGESTION_INGESTION_WORKSPACE_ID");
+        public string Ingestion_WorkspaceId => GetRecordedVariable("WORKSPACE_ID");
```
JoshLove-msft, line 14: I think we probably want INGESTION in the name in case other services have a similar env var.

nisha-bhatia: both Monitor.Query and Monitor.Ingestion use the same value

JoshLove-msft: Ah ok

sdk/monitor/test-resources.bicep
```diff
@@ -69,6 +71,104 @@ resource metricReaderRole 'Microsoft.Authorization/roleAssignments@2018-01-01-pr
   }
 }
 
+resource table 'Microsoft.OperationalInsights/workspaces/tables@2022-10-01' = {
```
pallavit, line 74: Are we the only language that uses Bicep over ARM template? Also, do we have any tools that can say convert an ARM template that we use Internally?

sdk/monitor/test-resources.bicep
```diff
@@ -69,6 +71,104 @@ resource metricReaderRole 'Microsoft.Authorization/roleAssignments@2018-01-01-pr
...
+  name: tableName
+  parent: primaryWorkspace
+  properties:{
+    totalRetentionInDays: 30
+    plan: 'Analytics'
+    schema: {
+      name: tableName
+      displayName: tableName
+      description: 'Table for Ingestion testing'
```
pallavit, line 83: Also, now that we are separating the pipelines how would the shared test resources work? Do we need to do anything extra to make sure this continues to happen?

sdk/monitor/test-resources.bicep
nisha-bhatia: Go uses Bicep as well for Monitor Query. I believe this is the internal command but it didn't quite work for me. I used this tutorial page: https://learn.microsoft.com/en-us/training/modules/introduction-to-infrastructure-as-code-using-bicep/5-how-bicep-works. This link to the Deployment Resources page (https://learn.microsoft.com/en-us/azure/templates/microsoft.authorization/roleassignments?pivots=deployment-language-bicep) also helps for a quickly formatted bicep resource deployment.

sdk/monitor/test-resources.bicep
nisha-bhatia: The test-resources.bicep file is under the monitor/  directory and is called with "./eng/common/TestResources/New-TestResources.ps1 monitor" command. I think separating the pipelines should not have an affect on this but maybe we need  three test-resources.bicep files?

pallavit: I think your understanding is correct (a common test resources file will likely work). But it would be best to confirm with the engsys team.

sdk/monitor/test-resources.bicep
pallavit: Thanks! Can you point me to the internal command that we use which doesn't work anymore?

sdk/monitor/test-resources.bicep
JoshLove-msft: I think things should still work as we [pass the service directory](https://github.com/Azure/azure-sdk-for-net/blob/main/eng/pipelines/templates/jobs/live.tests.yml#L110) when deploying the resources, but I think the difference is that we will now be creating the same set of resources 3 times. Only 2 of the test pipelines require the resources, and of those 2 each one only uses a subset.

JoshLove-msft: We might need to update the New-TestResources.ps1 script to allow passing an optional library directory if we want to avoid the test resource duplication.

pallavit: I think ensuring that each pipeline runs on a staggered schedule is likely the easiest solution here and good enough to prevent conflict.

JoshLove-msft: Conflict isn't the issue - each run will use a random ID as part of the resource group name. The issue is unnecessary resource creation.

JoshLove-msft: That said, I would agree it is not something that is a blocking issue. We could probably address it in MQ.

JoshLove-msft: Created https://github.com/Azure/azure-sdk-tools/issues/4450

sdk/monitor/test-resources.bicep
JoshLove-msft: This is part of the AZ CLI - https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/decompile?tabs=azure-cli#decompile-from-json-to-bicep

JoshLove-msft (approved):

Issue #31259, opened 2022-09-19
Separate monitor pipelines for ingestion, query and open telemetry


pallavit:
Currently we have a single pipeline for ingestion, query and open telemetry in .NET and we should separate them to have better visibility into test pipeline quality of each as we do in other languages.

https://dev.azure.com/azure-sdk/internal/_dashboards/dashboard/d2224fd5-beee-47a0-90df-b24b73fdc8e7

PR #31804, opened 2022-10-13
[Azure.Monitor.Ingestion] Separate Monitor Live Test Pipelines

nisha-bhatia:


nisha-bhatia:
https://github.com/Azure/azure-sdk-for-net/issues/31259

sdk/monitor/Azure.Monitor.Ingestion/tests.yml
```diff
@@ -0,0 +1,10 @@
+trigger: none
+
+stages:
+  - template: /eng/pipelines/templates/stages/archetype-sdk-tests.yml
+    parameters:
+      ServiceDirectory: monitor
+      Artifacts:
+        - name: azure-monitor-ingestion
```
JoshLove-msft, line 8: name and safename should use the .NET naming style used here - https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/monitor/ci.yml#L44

sdk/monitor/Azure.Monitor.Ingestion/tests.yml
```diff
@@ -0,0 +1,10 @@
...
+          groupId: com.azure
```
JoshLove-msft, line 9: This is Java specific and can be removed.

JoshLove-msft:
I'm curious how you would run the pipelines manually now that we are adding library specific tests.yml files. I guess the question is how the pipeline names will be generated. 
@benbp do you happen to know?

sdk/monitor/tests.yml
```diff
@@ -4,3 +4,10 @@ extends:
   template: ../../eng/pipelines/templates/stages/archetype-sdk-tests.yml
```
JoshLove-msft, line 4: Is this file still needed?

benbp:
> I'm curious how you would run the pipelines manually now that we are adding library specific tests.yml files. I guess the question is how the pipeline names will be generated. @benbp do you happen to know?

I would suggest picking either the top level tests.yml (one pipeline), or the inner tests.yml (three pipelines) vs. having both, unless that's really desired.

The naming convention is based on the parent directory @JoshLove-msft, so for the following:

- `sdk/monitor/tests.yml` => `net - monitor - tests`
- `sdk/monitor/Azure.Monitor.Ingestion/tests.yml` => `net - Azure.Monitor.Ingestion - tests`

JoshLove-msft:
> > I'm curious how you would run the pipelines manually now that we are adding library specific tests.yml files. I guess the question is how the pipeline names will be generated. @benbp do you happen to know?
> 
> I would suggest picking either the top level tests.yml (one pipeline), or the inner tests.yml (three pipelines) vs. having both, unless that's really desired.
> 
> The naming convention is based on the parent directory @JoshLove-msft, so for the following:
> 
> * `sdk/monitor/tests.yml` => `net - monitor - tests`
> * `sdk/monitor/Azure.Monitor.Ingestion/tests.yml` => `net - Azure.Monitor.Ingestion - tests`

Thank you!

sdk/monitor/tests.yml
JoshLove-msft: Based on @benbp's comment, we can remove this one.

JoshLove-msft:
I guess we also need to update the [live test dashboard ](https://dev.azure.com/azure-sdk/internal/_dashboards/dashboard/d2224fd5-beee-47a0-90df-b24b73fdc8e7) to point to the new pipelines. It doesn't look like we have permissions to edit it. @benbp can you help with this?

sdk/monitor/Azure.Monitor.Query/tests.yml
```diff
@@ -0,0 +1,9 @@
+trigger: none
+
+stages:
+  - template: /eng/pipelines/templates/stages/archetype-sdk-tests.yml
```
JoshLove-msft, line 4: This doesn't seem to be the right relative path.

JoshLove-msft, line 4:
```suggestion
  - template: ../../../eng/pipelines/templates/stages/archetype-sdk-tests.yml
```

pallavit: There are two modes in the yml. The beginning as '/' means from the root folder while ../.. is the relative path. Both should work in this case.

pallavit:
> I guess we also need to update the [live test dashboard ](https://dev.azure.com/azure-sdk/internal/_dashboards/dashboard/d2224fd5-beee-47a0-90df-b24b73fdc8e7) to point to the new pipelines. It doesn't look like we have permissions to edit it. @benbp can you help with this?

@hallipr may also know I believe he is the owner as well.

pallavit:
We may also need to manually delete older pipelines

sdk/monitor/Azure.Monitor.Query/tests.yml
JoshLove-msft: Hmm, I was basing it on this error - https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1920999&view=results

pallavit: Ahh may be we do not need the '/' at all.

JoshLove-msft:
> We may also need to manually delete older pipelines

I ran the prepare pipelines command but I guess that doesn't handle deleting unused pipelines (which probably makes sense). Do you know if we would need Eng Sys help to delete https://dev.azure.com/azure-sdk/internal/_apps/hub/ms.vss-build-web.ci-designer-hub?pipelineId=2545?

sdk/monitor/Azure.Monitor.Query/tests.yml
JoshLove-msft: Actually, I think you were right about the path not being the issue (it should be fine as-is though in the .NET repo we seem to always use the relative path syntax). 

I think the issue is that in the .NET [archetype-sdk-tests.yml](https://github.com/Azure/azure-sdk-for-net/blob/main/eng/pipelines/templates/stages/archetype-sdk-tests.yml), we don't have an artifacts parameter. I think we can try removing the `artifacts` from the parameters list here.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests.yml
```diff
@@ -0,0 +1,9 @@
+trigger: none
+
+stages:
+  - template: /eng/pipelines/templates/stages/archetype-sdk-tests.yml
+    parameters:
+      ServiceDirectory: monitor
+      Artifacts:
+        - name: Azure.Monitor.OpenTelemetry.Exporter
+          safeName: AzureMonitorOpenetemetryExporter
```
JoshLove-msft, line 9:
```suggestion
```

sdk/monitor/Azure.Monitor.Query/tests.yml
```diff
@@ -0,0 +1,9 @@
...
+    parameters:
+      ServiceDirectory: monitor
+      Artifacts:
+        - name: Azure.Monitor.Query
+          safeName: AzureMonitorQuery
```
JoshLove-msft, line 9:
```suggestion
```

sdk/monitor/Azure.Monitor.Ingestion/tests.yml
```diff
@@ -0,0 +1,9 @@
+trigger: none
+
+stages:
+  - template: /eng/pipelines/templates/stages/archetype-sdk-tests.yml
+    parameters:
+      ServiceDirectory: monitor
+      Artifacts:
+        - name: Azure.Monitor.Ingestion
+          safeName: AzureMonitorIngestion
```
JoshLove-msft, line 9:
```suggestion
```

pallavit:
> > We may also need to manually delete older pipelines
> 
> I ran the prepare pipelines command but I guess that doesn't handle deleting unused pipelines (which probably makes sense). Do you know if we would need Eng Sys help to delete https://dev.azure.com/azure-sdk/internal/_apps/hub/ms.vss-build-web.ci-designer-hub?pipelineId=2545?

Yes, @hallipr or @benbp should be able to help us there. We may want to archive it for now until everything is running smoothly for the week or so. I will let engsys give us guidance here on best approach forward.

benbp:
> > > We may also need to manually delete older pipelines
> > 
> > 
> > I ran the prepare pipelines command but I guess that doesn't handle deleting unused pipelines (which probably makes sense). Do you know if we would need Eng Sys help to delete https://dev.azure.com/azure-sdk/internal/_apps/hub/ms.vss-build-web.ci-designer-hub?pipelineId=2545?
> 
> Yes, @hallipr or @benbp should be able to help us there. We may want to archive it for now until everything is running smoothly for the week or so. I will let engsys give us guidance here on best approach forward.

We can manually delete, which will soft delete for 30 days. Want me to do it now, or wait a week?

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests.yml
```diff
@@ -0,0 +1,6 @@
+trigger: none
+
+stages:
```
JoshLove-msft, line 3:
```suggestion
extends:
```

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests.yml
```diff
@@ -0,0 +1,6 @@
...
+  - template: /eng/pipelines/templates/stages/archetype-sdk-tests.yml
```
JoshLove-msft, line 4:
```suggestion
    template: /eng/pipelines/templates/stages/archetype-sdk-tests.yml
```

JoshLove-msft (approved):

JoshLove-msft:
Still not working, I think we need to pass the Project parameter in each of the test.yml files - https://github.com/Azure/azure-sdk-for-net/blob/main/eng/pipelines/templates/stages/archetype-sdk-tests.yml#L132

JoshLove-msft (approved):

JoshLove-msft:
@benbp looks like check enforcer commands are not working. Is this a known issue with new version?

JoshLove-msft:
> @JoshLove-msft the PR looks mergeable to me, so I think override worked?

Ah yes, was expecting to see the failed checks cleared, but you are right it is mergeable now.

benbp:
> Ah yes, was expecting to see the failed checks cleared, but you are right it is mergeable now.

Yep, it doesn't erase anything, just changes the status of the "required check" (the devops checks are not listed as required checks since github doesn't support dynamic/variable checks).

PR #31843, opened 2022-10-17
Add credential samples

JoshLove-msft:
Addresses https://github.com/Azure/azure-sdk-for-net/issues/28215

### DIFF
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/TROUBLESHOOTING.md
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/TROUBLESHOOTING.md
@@ -79,6 +79,20 @@ The exception includes some contextual information to assist in understanding th
 
   - **ServiceCommunicationProblem**: This indicates that there was an error communicating with the service. The issue may stem from a transient network problem, or a service problem. These are transient errors that will be automatically retried.
 
+As an example of how to handle a `ServiceBusException` and filter by the `Reason`, see the following:
+
+```C# Snippet:ServiceBusExceptionFailureReasonUsage
+try
+{
+    // Receive messages using the receiver client
+}
+catch (ServiceBusException ex) when
+    (ex.Reason == ServiceBusFailureReason.ServiceTimeout)
+{
+    // Take action based on a service timeout
+}
+```
+
 ### Other common exceptions
 
 - **ArgumentException** : An exception deriving from `ArgumentException` is thrown by clients when a parameter provided when interacting with the client is invalid. Information about the specific parameter and the nature of the problem can be found in the `Message`.
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/samples/Sample00_AuthenticateClient.md
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/samples/Sample00_AuthenticateClient.md
@@ -37,7 +37,14 @@ Shared access signatures (SAS) are recommended over shared access keys when [RBA
 
 ```C# Snippet:ServiceBusAuthSasKey
 using var hmac = new HMACSHA256(Encoding.UTF8.GetBytes(key));
-var url = WebUtility.UrlEncode(fullyQualifiedNamespace);
+var builder = new UriBuilder(fullyQualifiedNamespace)
+{
+    Scheme = "amqps",
+    // scope our SAS token to the queue that is being used to adhere to the principle of least privilege
+    Path = queueName
+};
+
+var url = WebUtility.UrlEncode(builder.Uri.AbsoluteUri);
 var exp = DateTimeOffset.Now.AddHours(1).ToUnixTimeSeconds();
 var sig = WebUtility.UrlEncode(Convert.ToBase64String(hmac.ComputeHash(Encoding.UTF8.GetBytes(url + "\n" + exp))));
 
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/samples/Sample00_AuthenticateClient.md
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/samples/Sample00_AuthenticateClient.md
@@ -1,0 +1,56 @@
+# Authenticate the client
+
+This sample demonstrates how to authenticate the `ServiceBusClient`, which is the starting point for all interactions with the Service Bus client library.
+
+## Authenticate with a connection string
+
+The simplest way to authenticate with the Service Bus is to use a connection string, which is created automatically when creating a Service Bus namespace. If you aren't familiar with shared access policies in Azure, you may wish to follow the step-by-step guide to [get a Service Bus connection string](https://docs.microsoft.com/azure/service-bus-messaging/service-bus-quickstart-topics-subscriptions-portal#get-the-connection-string). For production scenarios, we recommend using Azure.Identity authentication as it provides [role-based access control](https://docs.microsoft.com/azure/role-based-access-control/overview) without the need to manage connection strings or shared access keys.
+
+```C# Snippet:ServiceBusAuthConnString
+// Create a ServiceBusClient that will authenticate using a connection string
+string connectionString = "<connection_string>";
+await using var client = new ServiceBusClient(connectionString);
+```
+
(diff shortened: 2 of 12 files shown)


sdk/servicebus/Azure.Messaging.ServiceBus/TROUBLESHOOTING.md
```diff
@@ -79,6 +79,20 @@ The exception includes some contextual information to assist in understanding th
 
   - **ServiceCommunicationProblem**: This indicates that there was an error communicating with the service. The issue may stem from a transient network problem, or a service problem. These are transient errors that will be automatically retried.
 
+As an example of how to handle a `ServiceBusException` and filter by the `Reason`, see the following:
+
+```C# Snippet:ServiceBusExceptionFailureReasonUsage
```
jsquire, line 84: Nice...  I shall have to consider stealing this.

sdk/servicebus/Azure.Messaging.ServiceBus/samples/Sample00_AuthenticateClient.md
```diff
@@ -0,0 +1,56 @@
+# Authenticate the client
+
+This sample demonstrates how to authenticate the `ServiceBusClient`, which is the starting point for all interactions with the Service Bus client library.
+
+## Authenticate with a connection string
+
+The simplest way to authenticate with the Service Bus is to use a connection string, which is created automatically when creating a Service Bus namespace. If you aren't familiar with connection strings in Service Bus, you may wish to follow the step-by-step guide to [get a Service Bus connection string](https://docs.microsoft.com/azure/service-bus-messaging/service-bus-quickstart-topics-subscriptions-portal#get-the-connection-string). For production scenarios, we recommend using Azure.Identity authentication as it provides [role-based access control](https://docs.microsoft.com/azure/role-based-access-control/overview) without the need to manage connection strings or shared access keys.
```
m-redding, line 7:
```suggestion
The simplest way to authenticate with Service Bus is to use a connection string, which is created automatically when creating a Service Bus namespace. If you aren't familiar with connection strings in Service Bus, you may wish to follow the step-by-step guide to [get a Service Bus connection string](https://docs.microsoft.com/azure/service-bus-messaging/service-bus-quickstart-topics-subscriptions-portal#get-the-connection-string). For production scenarios, we recommend using Azure.Identity authentication as it provides [role-based access control](https://docs.microsoft.com/azure/role-based-access-control/overview) without the need to manage connection strings or shared access keys.
```
This might be totally fine, I just haven't read service bus referred to as the service bus way before

sdk/servicebus/Azure.Messaging.ServiceBus/samples/Sample00_AuthenticateClient.md
```diff
@@ -0,0 +1,56 @@
+# Authenticate the client
```
jsquire, line 1: Sample 00, eh?   That's cheating!    We may want to consider naming this "Hello World" to snap closer to the [samples guidelines](https://azure.github.io/azure-sdk/dotnet_introduction.html#dotnet-samples) - though I don't feel strongly about it, personally.

sdk/servicebus/Azure.Messaging.ServiceBus/samples/Sample00_AuthenticateClient.md
```diff
@@ -0,0 +1,56 @@
...
+
+Shared access keys for Service Bus authorization are generated when access policies are created for a Service Bus namespace. Since these keys are most often used in association with a connection string, the article providing instructions to [get a Service Bus connection string](https://docs.microsoft.com/azure/service-bus-messaging/service-bus-quickstart-topics-subscriptions-portal#get-the-connection-string) is the best source of information on generating and accessing them. One of the key benefits to using a shared access key credential over a connection string is that the credential allows support for updating the key without creating a new client. In step 3 of the article, the policy that you select will be the name of your shared access key when used for credential authorization. You'll want to copy the "Primary key" rather than connection string.
+
+
+```C# Snippet:ServiceBusAuthNamedKey
+var credential = new AzureNamedKeyCredential("<name>", "<key>");
+await using var client = new ServiceBusClient("yournamespace.servicebus.windows.net", credential);
+```
+
+## Authenticate with a Shared Access Signature Credential
```
jsquire, line 34: This is a good idea as well.    I don't think the `url` below is correct.   Shouldn't it be more like:

```csharp
private static string BuildAudience(
    string fullyQualifiedNamespace,
    string entityName)
{    
    var builder = new UriBuilder(fullyQualifiedNamespace)
    {
        Scheme = "amqps",
        Path = entityName,
        Port = -1,
        Fragment = string.Empty,
        Password = string.Empty,
        UserName = string.Empty,
    };

    builder.Path = builder.Path.TrimEnd('/');
    return builder.Uri.AbsoluteUri.ToLowerInvariant();
}
```

By the time we get a namespace name, I think we strip the protocol and such out.

sdk/servicebus/Azure.Messaging.ServiceBus/samples/Sample00_AuthenticateClient.md
```diff
@@ -0,0 +1,56 @@
...
+
+Shared access signatures (SAS) are recommended over shared access keys, when RBAC cannot be used. A shared access signature allows for granular and time-limited access to Service Bus resources. 
+
+```C# Snippet:ServiceBusAuthSasKey
+using var hmac = new HMACSHA256(Encoding.UTF8.GetBytes(key));
+var url = WebUtility.UrlEncode(fullyQualifiedNamespace);
+var exp = DateTimeOffset.Now.AddHours(1).ToUnixTimeSeconds();
+var sig = WebUtility.UrlEncode(Convert.ToBase64String(hmac.ComputeHash(Encoding.UTF8.GetBytes(url + "\n" + exp))));
+
+// See https://learn.microsoft.com/azure/service-bus-messaging/service-bus-sas#generate-a-shared-access-signature-token for the authoritative documentation
```
m-redding, line 44: It might be nice to have the link clickable in L36

m-redding (approved):
👍

sdk/servicebus/Azure.Messaging.ServiceBus/samples/Sample00_AuthenticateClient.md
JoshLove-msft: Ah I didn't realize that was an official guideline but I have to say I'm not a big fan of it.

jsquire: No objections if we choose to ignore it here.

sdk/servicebus/Azure.Messaging.ServiceBus/samples/Sample00_AuthenticateClient.md
JoshLove-msft: You are correct that the protocol is not there, but it works without the protocol, so I guess it is not necessary.

jsquire: Interesting.  My understanding was that it needed to match exactly.    That said, without the entity name, you're demonstrating a namespace-level grant, which goes against the principle of least privilege.

jsquire: I'm good if you decide that you want to keep it simple, but please throw a big disclaimer if you choose to do so and mention that they should refer to the SAS guide that you linked for demonstration of creating one scoped to the entity.

jsquire (approved):

sdk/servicebus/Azure.Messaging.ServiceBus/samples/Sample00_AuthenticateClient.md
JoshLove-msft: I tested it out and the scheme seems to not matter at all - you can specify any thing or leave it off entirely. Agreed on including entity path.

Issue #31395, opened 2022-09-23
EventDataBatch.TryAdd does not copy buffer to batch as documentation indicates


Stabzs:
### Library name and version

Azure.Messaging.EventHubs v5.7.2

### Describe the bug

The [documentation](https://learn.microsoft.com/en-us/dotnet/api/azure.messaging.eventhubs.producer.eventdatabatch.tryadd?view=azure-dotnet#azure-messaging-eventhubs-producer-eventdatabatch-tryadd(azure-messaging-eventhubs-eventdata)) for `Azure.Messaging.EventHubs.Producer.EventDataBatch.TryAdd` has the following remarks:

> Remarks
> When an event is accepted into the batch, its content and state are frozen; any changes made to the event will not be reflected in the batch nor will any state transitions be reflected to the original instance.

However, when trying to use a shared buffer as the backing source per EventData instance, the data is not persisted in the EventDataBatch

### Expected behavior

```
using var eventDataBatch = await _eventHubClient.CreateBatchAsync();
var buffer = new byte[1024];
while (true)
{
    buffer.Clear();
    buffer = someInput;
    if (!eventDataBatch.TryAdd(new EventData(buffer))
        break;
}
```

I would expect, based on the remark, that during TryAdd, the buffer's contents are serialized into the raw amqp message and added to the batch, allowing it to be reused in the add loop.

### Actual behavior

What I am seeing however is that clearing the buffer is being reflected back into the batch itself, making the use of a buffer completely impossible and causing the above remark to be inaccurate at best and (in my case), quite dangerous.

### Reproduction Steps

See the above.

### Environment

I am seeing the same behavior on Windows, Linux and MacOS, on .NET 6.

jsquire:
Hi @Stabzs.  Thank you for reaching out and we regret that you're experiencing difficulties.  The value that you're passing the `EventData` constructor is of type `ReadOnlyMemory<byte>` which, by contract, offers a guarantee that it will not change and is safe to use as a buffer.   Internally, the body of the serialized message references this read-only memory buffer until it is sent over the network.

That said, it wasn't our intent to allow a side-effect like this and the serialized form is intended to be immutable; we did not account for this specific scenario.  I'll take a look at getting this patched up so the side-effect is no longer possible.

The mitigation, in the meantime, would be to create a new buffer for each `EventData` to honor the contract of `ReadOnlyMemory<T>`

Stabzs:
Thanks for the quick response @jsquire!

[...]

PR #31845, opened 2022-10-17
[Event Hubs] Data body copy for EventDataBatch

jsquire:
# Summary

The focus of these changes is to avoid an unintentional loophole allowing an event batch to be mutated by mutating the data body of an `EventData` instance that has been added to it.  When an `EventData` is serialized into the interim `AmqpMessage` format used by the transport library, a projection over the existing `ReadOnlyMemory<byte>` buffer has been used to avoid allocations.

To combat this, the batch will now make a defensive copy of the data body of an `EventData` when accepted into the batch, using a shared buffer that has been rented.  When the batch is disposed, the buffer is returned.

**Additional context:**

Though the buffer is typed as `ReadOnlyMemory<byte>`, since the it was passed to the EventData constructor, and the scope of the "call" is the lifetime of the associated instance.  We explicitly intend that the `EventData` instance is no longer relevant to the batch after `TryAdd` completes. Because of this, it should not matter what a caller does with the body buffer after, the batch should be fully immutable.

# References and Related

- [EventDataBatch.TryAdd does not copy buffer to batch as documentation indicates (#31395)](https://github.com/Azure/azure-sdk-for-net/issues/31395)

### DIFF
--- a/sdk/eventhub/Azure.Messaging.EventHubs/src/Amqp/AmqpEventBatch.cs
+++ b/sdk/eventhub/Azure.Messaging.EventHubs/src/Amqp/AmqpEventBatch.cs
@@ -9,7 +9,6 @@ using System.Runtime.CompilerServices;
 using Azure.Core;
 using Azure.Core.Amqp;
 using Azure.Messaging.EventHubs.Core;
-using Azure.Messaging.EventHubs.Diagnostics;
 using Azure.Messaging.EventHubs.Producer;
 using Microsoft.Azure.Amqp;
 
@@ -42,9 +41,6 @@ namespace Azure.Messaging.EventHubs.Amqp
 
         /// <summary>The set of events that have been added to the batch, in their <see cref="AmqpMessage" /> serialized format.</summary>
         private readonly List<AmqpMessage> _batchMessages = new();
-
-        /// <summary>The set of buffers rented from the shared array pool which must be returned on disposal.</summary>
-        private readonly List<byte[]> _rentedBuffers = new();
 
         /// <summary>The first sequence number of the batch; if not sequenced, <c>null</c>.</summary>
         private int? _startingSequenceNumber;
@@ -178,9 +174,7 @@ namespace Azure.Messaging.EventHubs.Amqp
                        var bodyArraySegment = (ArraySegment<byte>)bodySegment.Value;
                        var buffer = ArrayPool<byte>.Shared.Rent(bodyArraySegment.Count);
 
-                       _rentedBuffers.Add(buffer);
                        CopyArraySegmentTo(bodyArraySegment, buffer);
-
                        bodySegment.Value = new ArraySegment<byte>(buffer, 0, bodyArraySegment.Count);
                    }
                 }
@@ -200,15 +194,10 @@ namespace Azure.Messaging.EventHubs.Amqp
         ///
         public override void Clear()
         {
-            foreach (var message in _batchMessages)
-            {
-                message.Dispose();
-            }
+            FreeBatchMessages(_batchMessages, true);
 
             _batchMessages.Clear();
             _sizeBytes = _reservedOverheadBytes;
-
-            ReturnRentedBuffers(_rentedBuffers);
         }
 
         /// <summary>
@@ -284,8 +273,8 @@ namespace Azure.Messaging.EventHubs.Amqp
         {
             _disposed = true;
 
-            // Rented buffers are returned when the batch is
-            // cleared.
+            // Clearing the batch will free its messages and perform any other
+            // necessary cleanup of their state.
 
             Clear();
             GC.SuppressFinalize(this);
@@ -297,43 +286,49 @@ namespace Azure.Messaging.EventHubs.Amqp
         ///
         ~AmqpEventBatch()
         {
-            ReturnRentedBuffers(_rentedBuffers);
+            FreeBatchMessages(_batchMessages, false);
         }
 
         /// <summary>
-        ///   Returns any rented buffers held by the batch back into the
-        ///   shared array pool.
+        ///   Returns any rented buffers held by the AMQP messages in the batch back into the
+        ///   shared array pool and optionally disposes the messages.
         /// </summary>
         ///
-        /// <param name="rentedBuffers">The set of rented buffers to return.</param>
-        ///
-        /// <remarks>
-        ///   If an attempt is made to return a buffer that is no longer rented, the
-        ///   failure will be ignored and the remaining buffers returned.
-        /// </remarks>
+        /// <param name="batchMessages">The set of messages to free and, optionally, dispose.</param>
+        /// <param name="disposeMessages"><c>true</c> to dispose messages after freeing; otherwise. <c>false</c>.</param>
         ///
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        private static void ReturnRentedBuffers(List<byte[]> rentedBuffers)
+        private static void FreeBatchMessages(List<AmqpMessage> batchMessages,
+                                              bool disposeMessages)
         {
-            foreach (var rentedBuffer in rentedBuffers)
+            foreach (var message in batchMessages)
             {
-                try
+                if (((message.BodyType & SectionFlag.Data) > 0) && (message.DataBody != null))
                 {
-                    ArrayPool<byte>.Shared.Return(rentedBuffer);
+                    foreach (var segment in message.DataBody)
+                    {
+                        try
+                        {
+                            ArrayPool<byte>.Shared.Return(((ArraySegment<byte>)segment.Value).Array);
+                        }
+                        catch (ArgumentException)
+                        {
+                            // Ignore.
+                            //
+                            // This occurs when the buffer passed is not rented. If this scenario
+                            // is encountered, it likely indicates an error happened during TryAdd
+                            // when renting the buffer or at Dispose.  The remainder of the rented
+                            // buffers should still be returned.
+                        }
+                    }
                 }
-                catch (ArgumentException)
+
+                if (disposeMessages)
                 {
-                    // Ignore.
-                    //
-                    // This occurs when the buffer passed is not rented. If this scenario
-                    // is encountered, it likely indicates an error happened during TryAdd
-                    // when renting the buffer or at Dispose.  The remainder of the rented
-                    // buffers should still be returned.
+                    message.Dispose();
                 }
             }
-
-            rentedBuffers.Clear();
-        }
+}
 
         /// <summary>
         ///   Calculates the next sequence number based on the current sequence number.
--- a/sdk/eventhub/Azure.Messaging.EventHubs/src/Amqp/AmqpEventBatch.cs
+++ b/sdk/eventhub/Azure.Messaging.EventHubs/src/Amqp/AmqpEventBatch.cs
@@ -2,11 +2,14 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Buffers;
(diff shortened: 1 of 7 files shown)


jsquire:
@danielmarbach:  If you've any thoughts on the implementation, they would be appreciated.

JoshLove-msft (approved):

danielmarbach:
Will do as soon as I have some wiggle room

sdk/eventhub/Azure.Messaging.EventHubs/src/Amqp/AmqpEventBatch.cs
```diff
@@ -157,6 +163,28 @@ public override bool TryAdd(EventData eventData)
...
+                {
+                   foreach (var bodySegment in message.DataBody)
+                   {
+                       var bodyArraySegment = (ArraySegment<byte>)bodySegment.Value;
+                       var buffer = ArrayPool<byte>.Shared.Rent(bodyArraySegment.Count);
+
+                       _rentedBuffers.Add(buffer);
+                       CopyArraySegmentTo(bodyArraySegment, buffer);
+
+                       bodySegment.Value = new ArraySegment<byte>(buffer, 0, bodyArraySegment.Count);
```
christothes, line 184: Do we keep a reliable reference around for the collection of bodySegments anywhere? If so, I wonder if we could eliminate the need for the `_rentedBuffers` List since we can get a reference to the array reference that was used to construct the ArraySegment via its [Array](https://learn.microsoft.com/en-us/dotnet/api/system.arraysegment-1.array?view=net-6.0) property

jsquire: It's possible to do from the set of AMQP messages in the batch, but it's ugly with nested loops and casts:

```csharp
private static void ReturnRentedBuffers(List<AmqpMessage> batchMessages)
{
    foreach (var message in batchMessages)
    {
        if (((message.BodyType & SectionFlag.Data) > 0) && (message.DataBody != null))
        {
            foreach (var segment in message.DataBody)
            {
                try
                {
                    ArrayPool<byte>.Shared.Return(((ArraySegment<byte>)segment.Value).Array);
                }
                catch (ArgumentException)
                {
                    // Ignore.
                    //
                    // This occurs when the buffer passed is not rented. If this scenario
                    // is encountered, it likely indicates an error happened during TryAdd
                    // when renting the buffer or at Dispose.  The remainder of the rented
                    // buffers should still be returned.
                }
            }
        }
    }
}
```

The current implementation trades an allocation and a bit of overhead to amortize the clean-up cost.   Thoughts?

christothes: I don't have much context as to whether or not the allocation savings are important enough in this case, but I was more curious if we had considered that as an option. I defer to your judgement on the tradeoff :)

christothes (approved):

sdk/eventhub/Azure.Messaging.EventHubs/src/Amqp/AmqpEventBatch.cs
jsquire: It's ugly, but after playing with it a bit, I decided to go that way and remove the need for the tracking list of rented buffers.

danielmarbach:
The remark on the documentation surprised me a lot because I would have expected it as a client of the library that it is not allowed to reuse the memory until the batch has been actually sent. That is for me the most logical and intuitive way to reason about memory ownership and boundaries between my code and the "handing it over to the lib" part. I was reading the service bus batch documentation and that falls more into how I would expect a batch to behave. But given the remark is there I guess the behavior has to be aligned with what is mentioned and the remark and the argument of "how service bus batches behave is more natural in terms of ownership of memory" is kind of moot.

One thing that bothers me about the approach here is that I feel we have tried to neatly encapsulate things in the message body abstractions and now there is body management that leaks out again. Wouldn't it be much cleaner to encapsulate this case here as well in a special message body type and does eager copying?

sdk/eventhub/Azure.Messaging.EventHubs/src/Amqp/AmqpEventBatch.cs
```diff
@@ -157,6 +159,26 @@ public override bool TryAdd(EventData eventData)
...
+                // it will need to own a buffer for the data body to guard against mutation of the
+                // source event.
+                //
+                // Because the batch is disposable and allows for deterministic clean-up, rent buffers
+                // from the shared pool rather than allocating a dedicated instance for each body segment.
+
+                if (eventData.GetRawAmqpMessage().Body.BodyType == AmqpMessageBodyType.Data)
+                {
+                   foreach (var bodySegment in message.DataBody)
+                   {
```
danielmarbach, line 173: Since the segments are known why not rent one buffer that is sliced for all of them?

sdk/eventhub/Azure.Messaging.EventHubs/src/Amqp/AmqpEventBatch.cs
```diff
@@ -274,5 +348,25 @@ private static int NextSequence(int currentSequence)
...
+        /// <param name="source">The array segment to read content from.</param>
+        /// <param name="destination">The array to copy the segment content to.</param>
+        ///
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static void CopyArraySegmentTo(ArraySegment<byte> source,
+                                               byte[] destination)
+        {
+            var index = -1;
+
+            foreach (var item in source)
```
danielmarbach, line 366: Could we operate on the span level here directly?

danielmarbach (commented):
Sorry I'm currently at a conference until tomorrow so I was only able to squeeze in a shallow review.

jsquire:
@danielmarbach, @amis92:

I appreciate your comments and perspectives.  Not going to lie - I really don't like this change and am still waffling on the need for it.  

To clarify, the intent has always been that items in a batch cannot mutate the batch.  Originally, a defensive copy of the entire event was made at the time it was added due to some issues that we were experiencing our AMQP transport library.   Eventually, we were able to improve things by just keeping the "serialized" form - which is just an interim object graph from the transport library.   Thanks to many of Daniel's contributions, that serialization is projecting over buffers to avoid copies.  

The challenge here is the guarantees around `ReadOnlyMemory<T>` - which, though not well documented, I confirmed internally they imply that when passed to a constructor the memory should be treated as immutable for the lifetime of the object.   In this case, there's friction between "the batch is now taking ownership of the buffer and you shouldn't change it while the batch is alive" and "the buffer is owned by the event, and we don't care about the event once it's in the batch."

Though, I will say - having written it out like that, I think documenting that adding to the batch transfers ownership frames things in a reasonable perspective that neither violates the guidance for `ReadOnlyMemory<T>` nor the intent of the batch.   It's definitely worth considering discarding this change and revising the documentation, if we can do so in a clear way that doesn't violate the memory guarantee.

@Stabzs - as the author of the original issue, does the above framing as an ownership transfer seem reasonable to you and would it have clarified when writing your app?
@JoshLove-msft, @m-redding, @christothes - I'd also appreciate your thoughts on the above.

> I was reading the service bus batch documentation and that falls more into how I would expect a batch to behave.

That's not accurate for Service Bus anymore - we fixed the original bug (and apparently forgot to update the docs).   However, it is the same implementation as Event Hubs and has the same gap.  😄 

> This change impacts performance (are there any tests how much?),

This is true, of course.   I've been running benchmarks (though not keeping results) and after a warm-up period, it's been surprisingly close, roughly 1-3% slower for scenarios focused on publishing throughput, so long as the overall number of batches in-flight remains relatively constant.  So long as a buffer is available, the cost is mostly just copying the data.  There's a small number of allocations for boxing the new array segment - but the majority case has a single segment per event.

That said, that this change degrades performance for the majority scenario to guard against an edge case is the thing that most bothers me and why this hasn't yet been committed.

christothes:
> Though, I will say - having written it out like that, I think documenting that adding to the batch transfers ownership frames things in a reasonable perspective that neither violates the guidance for `ReadOnlyMemory<T>` nor the intent of the batch. It's definitely worth considering discarding this change and revising the documentation, if we can do so in a clear way that doesn't violate the memory guarantee.

Based on the [usage guidelines](https://learn.microsoft.com/en-us/dotnet/standard/memory-and-spans/memory-t-usage-guidelines), ownership is an established through convention. My interpretation of the guidelines as applied to this scenario is that when using the `ReadonlyMemory<T>` ctor of EventData:
https://github.com/Azure/azure-sdk-for-net/blob/8db6d7fee9e82d8f5fabbe0d17c0b76437864698/sdk/eventhub/Azure.Messaging.EventHubs/src/EventData.cs#L506

EventData is declaring that 
- It is the consumer of this buffer
- It will not modify the buffer
- It is not the owner of the buffer (will not be responsible for lifetime management)

I don't think there are any explicit guarantees made about the future modification of the buffer by the owner, with the exception of lifetime management by this convention alone (perhaps this is a gap that should be addressed in the guidance). So perhaps the only way to make this explicit is via documentation, as you've suggested.

JoshLove-msft:
I would tend to agree that documentation seems to be the best path forward here. I think the simplest way to communicate this is to say that users should not modify the message/event until the batch that it has been added to has been sent. If we want to get into more detail, we can say that technically all properties other than the body could be modified without impacting what will be sent.

Stabzs:
@JoshLove-msft I would love a little context on this if you don't mind. What is the technical reason that the properties/metadata of the message are sealed but the body is not? Does it make sense to make that consistent? If so, that would solve my ask and that was my original expectation.

jsquire:
> @JoshLove-msft I would love a little context on this if you don't mind. What is the technical reason that the properties/metadata of the message are sealed but the body is not? Does it make sense to make that consistent? If so, that would solve my ask and that was my original expectation.

I can take that one.    

The body properties are primitive values that are eagerly translated into AMQP encoding as part of serialization, because their form must change.  The body doesn't require encoding or transform for AMQP and is represented by an `ArraySegment<byte>` in the serialized transport model.  Essentially, when the AMQP message is sent over the network, the body bytes are just slurped from the `ReadOnlyMemory<T>` into the network stream.

In order to avoid copying, when we serialize into the AMQP interim model, the `ArraySegment<byte>` that we use we use for the message body is just a view over the `ReadOnlyMemory<byte>` held by the `EventData` instance.   In case you're interested, the body projection can be found [here](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Azure.Messaging.EventHubs/src/Amqp/AmqpMessageConverter.cs#L848), which is called from [here](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Azure.Messaging.EventHubs/src/Amqp/AmqpMessageConverter.cs#L400).

Stabzs:
@jsquire  thanks for the links and the explanation! Very helpful.  It makes sense why you'd have to make a full copy internally. Coming full circle, I'm good with closing the issue if the performance is impacted.

t-test-t:
m performance22, 2022 7:09:01 AM GMT+08:00, Stabzs ***@***.***> wrote:
***@***.***  thanks for the links and the explanation! Very helpful.  It makes sense why you'd have to make a full copy internally. Coming full circle, I'm good with closing the issue if the performance is impacted.

jsquire:
Thank you all for the perspectives and discussion.   I believe that we're all in agreement that the performance cost of making a defensive copy outweighs the desire to guarantee the batch is fully immutable.   As discussed, I've updated documentation to clarify (see: #32044) and am going to close this out.

Issue #31796, opened 2022-10-13
[Azure.Monitor.Ingestion] Resolve error handling for Upload


nisha-bhatia:
Determine if we need UploadLogsResult and UploadLogsError type
update docs

nisha-bhatia:
Resolved: https://github.com/Azure/azure-sdk-for-net/pull/31846

Issue #31767, opened 2022-10-12
[Azure.Monitor.Ingestion] Update Batch code to use ArrayPool and Utf8RequestContent


nisha-bhatia:
- [ ] resolve open TODOs in Batch method code on LogsIngestionClient:  https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs#L99
- [ ] instead of handwritting JSON chars '{' replace with values from JSON library

nisha-bhatia:
updated batching to use ArrayBufferWriter

PR #31846, opened 2022-10-17
[Azure.Monitor.Ingestion] Proposal for Error Handling for Ingestion Upload methods

nisha-bhatia:


nisha-bhatia:
https://github.com/Azure/azure-sdk-for-net/issues/31796

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -188,45 +174,44 @@ private static void WriteMemory(MemoryStream stream, ReadOnlyMemory<byte> memory
         /// ]]></code>
         /// </example>
         /// <remarks> See error response code and error response message for more detail. </remarks>
-        public virtual Response<UploadLogsResult> Upload<T>(string ruleId, string streamName, IEnumerable<T> logs, UploadLogsOptions options = null, CancellationToken cancellationToken = default)
+        public virtual Response Upload<T>(string ruleId, string streamName, IEnumerable<T> logs, UploadLogsOptions options = null, CancellationToken cancellationToken = default)
```
JoshLove-msft, line 177: It looks like the Response will just be from whichever request happened to complete last. This seems unintuitive. I wonder if we should even return a Response at all.

/cc @tg-msft

nisha-bhatia:
Another concern is let's say we have 40 batches we're uploading and 2 batches fail. In this case we would throw an Aggregate Exception holding these two failed batch upload exceptions. In this case it doesn't seem like the Upload was a success as even though 38 batches were uploaded successfully the user still gets the exception. We don't want the user to see the failure and try the Upload method with the same content again. Is there a way we can indicate to the user that the majority of logs have uploaded but there were some that failed? Maybe a percentage succeeded?

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
tg-msft: We do this in a lot of our higher level conveniences.  LRO polling only returns the last response.  Parallel upload for Blobs only returns the response of calling Commit.  Customers who care about every request can use lower level methods (the protocol method in this case) or submit batches small enough to fit in a single request.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -188,45 +174,44 @@ private static void WriteMemory(MemoryStream stream, ReadOnlyMemory<byte> memory
...
-                foreach (BatchedLogs<T> batch in Batch(logs, options))
+                try
                 {
                     // Because we are uploading in sequence, wait for each batch to upload before starting the next batch
-                    BatchUpload task = CommitBatchListSyncOrAsync(
+                    response = CommitBatchListSyncOrAsync(
                         batch,
                         ruleId,
                         streamName,
                         false,
```
tg-msft, line 199: nit - for boolean constants like this it's very helpful to pass it via a named parameter like `async: false`

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -188,45 +174,44 @@ private static void WriteMemory(MemoryStream stream, ReadOnlyMemory<byte> memory
...
             Argument.AssertNotNullOrEmpty(ruleId, nameof(ruleId));
             Argument.AssertNotNullOrEmpty(streamName, nameof(streamName));
             Argument.AssertNotNullOrEmpty(logs, nameof(logs));
 
             using var scope = ClientDiagnostics.CreateScope("LogsIngestionClient.Upload");
 
-            RequestContext requestContext = GenerateRequestContext(cancellationToken);
             Response response = null;
-            List<UploadLogsError> errors = new List<UploadLogsError>();
+            var exceptions = new List<Exception>();
```
tg-msft, line 186: Nit - consider not allocating this until you have any exceptions.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -188,45 +174,44 @@ private static void WriteMemory(MemoryStream stream, ReadOnlyMemory<byte> memory
...
+
             scope.Start();
-            try
+            // Partition the stream into individual blocks
+            foreach (BatchedLogs<T> batch in Batch(logs, options)) //TODO: check if we should throw early for sync case? as soon as we find exception
```
tg-msft, line 190: Do we need these TODOs or have they been figured out?

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -188,45 +174,44 @@ private static void WriteMemory(MemoryStream stream, ReadOnlyMemory<byte> memory
...
-                scope.Failed(ex);
-                throw;
+                throw new AggregateException(exceptions);
             }
 
-            // Calculate the status using the helper method Status
-            UploadLogsResult finalResult = new UploadLogsResult(errors, GetStatus(logs, errors));
-            return Response.FromValue(finalResult, response);
+            // If no exceptions return response
+            return response; //204 - return response of last batch (would include header) or create a blank response with success?
```
tg-msft, line 214: Return the last one

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -297,110 +282,52 @@ public virtual async Task<Response<UploadLogsResult>> UploadAsync<T>(string rule
...
-                scope.Failed(ex);
-                throw;
-            }
 
-            // Calculate the status using the helper method Status
-            UploadLogsResult finalResult = new UploadLogsResult(errors, GetStatus(logs, errors));
-            return Response.FromValue(finalResult, response);
-        }
+            // Wait for all the remaining blocks to finish uploading
+            await Task.WhenAll(runningTasks).ConfigureAwait(false);
```
tg-msft, line 304: Does this need to happen inside your try/catch block for any exceptions that might get raised?

tg-msft (approved):
I like the direction and think it'll make the library easier to use.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
JoshLove-msft: I think for LROs that behavior probably makes more sense as you are doing a bunch of requests serially and the last response is meaningful. Here, the response you get back is arbitrary and just represents whichever response came back last from a bunch of parallel requests. I wonder if we need a pattern for this kind of higher level helper that does not directly correspond to a particular response.

JoshLove-msft: For Blobs, there is a natural response to return - the response from Commit. There is no equivalent response here.

tg-msft:
Maybe we could make the partial failure clearer in the `AggregateException`'s `Message`?  Imagine something like `$"{failedCount} of {totalCount} log entries failed to upload.  See the InnerExceptions for more details."`.  We could also add entries to the exception's `Data` in the future if we find customers have any scenarios to access that programmatically.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
tg-msft: I'm not really bothered about whether the last response of a batch is somehow less "natural."  If we ever added a parallel append helper, we wouldn't have a final commit and just return the last response that completed.  Our rule of thumb for conveniences is it must be _possible_ to see every response for customers who care a lot, but the happy path doesn't need to surface them all - especially in cases like this where there's nothing interesting about them.

JoshLove-msft: Agree that we don't need to surface all responses - but my thought was that it might be better to just not surface any responses here. Just throwing it out there as an observation.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -188,45 +174,47 @@ private static void WriteMemory(MemoryStream stream, ReadOnlyMemory<byte> memory
...
+                catch (Exception ex)
+                {
+                    // If we have an error, add error from response into exceptions list which represents the errors from all the batches
+                    exceptions.Add(ex);
                 }
             }
-            catch (Exception ex)
+            if (exceptions.Count > 0)
             {
-                scope.Failed(ex);
```
JoshLove-msft, line 223: Why do we no longer mark the scope as failed?

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -297,110 +287,52 @@ public virtual async Task<Response<UploadLogsResult>> UploadAsync<T>(string rule
...
                         {
-                            Task<BatchUpload> runningTask = runningTasks[i];
+                            Task<Response> runningTask = runningTasks[i];
                             if (!runningTask.IsCompleted)
                             {
                                 continue;
                             }
-
-                            // Before removing runningTask - check if it has any errors and add to error list
-                            if (runningTask.Result.Error != null)
```
JoshLove-msft, line 307: I think we still need to keep track of the removed tasks.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
JoshLove-msft: If we await all tasks using WhenAll, we can use this [workaround](https://github.com/dotnet/runtime/issues/31494#issuecomment-553895267) to make sure an AggregateException is thrown for multiple exceptions.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -297,110 +287,52 @@ public virtual async Task<Response<UploadLogsResult>> UploadAsync<T>(string rule
...
                     }
                 }
-
-                // Wait for all the remaining blocks to finish uploading
-                await Task.WhenAll(runningTasks).ConfigureAwait(false);
-
-                // Process all errors after tasks are done to determine status
-                // Will run on a single thread
-                foreach (Task<BatchUpload> task in runningTasks)
+                catch (Exception ex)
```
JoshLove-msft, line 301: For the async method, we should be able to await all the tasks with Task.WhenAll (and use the linked workaround) to propagate exceptions.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -188,45 +174,47 @@ private static void WriteMemory(MemoryStream stream, ReadOnlyMemory<byte> memory
...
+            foreach (BatchedLogs<T> batch in Batch(logs, options))
             {
-                // Partition the stream into individual blocks
-                foreach (BatchedLogs<T> batch in Batch(logs, options))
+                batchCount++;
+                try
                 {
                     // Because we are uploading in sequence, wait for each batch to upload before starting the next batch
-                    BatchUpload task = CommitBatchListSyncOrAsync(
+                    response = CommitBatchListSyncOrAsync(
```
JoshLove-msft, line 198: I've been curious about the naming of the method. Would it make more sense to rename it as UploadBatchSyncOrAsync, or just UploadBatchAsync (the async parameter should convey that this can actually be run synchronously)?

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
JoshLove-msft: It might be good to discuss this across languages to make sure the behavior is consistent.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
JoshLove-msft: > Maybe we could make the partial failure clearer in the `AggregateException`'s `Message`? Imagine something like `$"{failedCount} of {totalCount} log entries failed to upload. See the InnerExceptions for more details."`. We could also add entries to the exception's `Data` in the future if we find customers have any scenarios to access that programmatically.

It might be nice if we could somehow also include the indices of the failed log entries from the initial list, e.g. "Upload failed for indices 4,5,6,12,13,14,15"

sdk/monitor/Azure.Monitor.Ingestion/tests/SessionRecords/MonitorIngestionLiveTest/ConcurrencyMultiThreadAsync.json
```diff
@@ -1,180 +1,180 @@
...
-        "traceparent": "00-0feae899d9b5ac4f8bb76518014d8cfe-a1bf7b798824644e-00",
-        "User-Agent": "azsdk-net-Monitor.Ingestion/1.0.0-alpha.20221012.1 (.NET Core 3.1.30; Microsoft Windows 10.0.22621)",
-        "x-ms-client-request-id": "b37c02315c1b49105bd815dbc0da4c3b",
+        "traceparent": "00-84024e655bda424a8663ab11d00b1efd-ce07df2b0dc90d49-00",
+        "User-Agent": "azsdk-net-Monitor.Ingestion/1.0.0-alpha.20221024.1 (.NET Core 3.1.30; Microsoft Windows 10.0.22621)",
+        "x-ms-client-request-id": "abdc147dba7aeba7b74c902c74a99d4d",
         "x-ms-return-client-request-id": "true"
       },
-      "RequestBody": "[{\u0022Time\u0022:\u00222022-10-13T14:10:38.720123\u0022,\u0022Computer\u0022:\u0022Computer4\u0022,\u0022AdditionalContext\u0022:4}]]",
+      "RequestBody": "[[{\u0022Time\u0022:\u00222022-10-24T17:04:30.6669136\u0022,\u0022Computer\u0022:\u0022Computer0\u0022,\u0022AdditionalContext\u0022:0}],]",
```
JoshLove-msft, line 17: Hmm this is strange - it seems we now have nested arrays? Did something change in the test?

sdk/monitor/Azure.Monitor.Ingestion/tests/SessionRecords/MonitorIngestionLiveTest/ConcurrencyMultiThreadAsync.json
```diff
@@ -1,180 +1,180 @@
...
-        "traceparent": "00-0feae899d9b5ac4f8bb76518014d8cfe-f6328e069d83074f-00",
-        "User-Agent": "azsdk-net-Monitor.Ingestion/1.0.0-alpha.20221012.1 (.NET Core 3.1.30; Microsoft Windows 10.0.22621)",
-        "x-ms-client-request-id": "d31e865338010fb562c34b2dc4a14b5d",
+        "traceparent": "00-84024e655bda424a8663ab11d00b1efd-5d86e36799803c49-00",
+        "User-Agent": "azsdk-net-Monitor.Ingestion/1.0.0-alpha.20221024.1 (.NET Core 3.1.30; Microsoft Windows 10.0.22621)",
+        "x-ms-client-request-id": "163563cb06016ef2cb933bd0f582becb",
         "x-ms-return-client-request-id": "true"
       },
-      "RequestBody": "[[{\u0022Time\u0022:\u00222022-10-13T14:10:38.720123\u0022,\u0022Computer\u0022:\u0022Computer0\u0022,\u0022AdditionalContext\u0022:0}],]",
+      "RequestBody": "[{\u0022Time\u0022:\u00222022-10-24T17:04:30.6669136\u0022,\u0022Computer\u0022:\u0022Computer1\u0022,\u0022AdditionalContext\u0022:1}]]",
```
JoshLove-msft, line 161: And this one removed the nested array 🤔

sdk/monitor/Azure.Monitor.Ingestion/tests/SessionRecords/MonitorIngestionLiveTest/ConcurrencyMultiThreadAsync.json
```diff
@@ -1,180 +1,180 @@
...
-        "x-ms-client-request-id": "c198e9bf-851f-b6c8-0016-7fd3db0e9149",
-        "x-ms-request-id": "cca18c51-0094-44f4-9d89-9d1eff035660"
+        "x-ms-client-request-id": "644bc58d-4041-e71d-9437-9271b0646ad3",
+        "x-ms-request-id": "17d2fd51-1590-4053-8485-ef92f4d59bc1"
       },
       "ResponseBody": null
     },
     {
-      "RequestUri": "https://98c96978-a857-58f1-a299-7d216d4b043a-p2j8.westus-1.ingest.monitor.azure.com/dataCollectionRules/dcr-37a02bb9a8ba4f97a80716652c78e678/streams/Custom-MyTableRawData?api-version=2021-11-01-preview",
+      "RequestUri": "https://86923540-75e9-5143-84b6-7535316a68dc-lipn.westus-1.ingest.monitor.azure.com/dataCollectionRules/dcr-ba536bf3cee94b1db8d04b30ee3f2e80/streams/Custom-MyTableRawData?api-version=2021-11-01-preview",
```
JoshLove-msft, line 52: In general, we shouldn't need to re-record tests whenever a change is made. Was there a mismatch in playback that made it necessary to re-record?

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -266,141 +263,132 @@ public virtual async Task<Response<UploadLogsResult>> UploadAsync<T>(string rule
...
-            RequestContext requestContext = GenerateRequestContext(cancellationToken);
-            Response response = null;
-            List<UploadLogsError> errors = new List<UploadLogsError>();
+            List<Exception> exceptions = null;
             scope.Start();
-            try
+
+            // A list of tasks that are currently executing which will
+            // always be smaller than or equal to MaxWorkerCount
+            List<Tuple<Task<Response>, int>> runningTasks = new();
```
JoshLove-msft, line 271: I think you can define names for the fields of the tuple, e.g.:
```c#
Tuple<Task<Response> Task, int LogsCount>
```

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -266,141 +263,132 @@ public virtual async Task<Response<UploadLogsResult>> UploadAsync<T>(string rule
...
                     // Add the block to our task and commit lists
-                    runningTasks.Add(task);
+                    runningTasks.Add(Tuple.Create(task, batch.LogsList.Count));
 
                     // If we run out of workers
                     if (runningTasks.Count >= _maxWorkerCount)
                     {
                         // Wait for at least one of them to finish
-                        await Task.WhenAny(runningTasks).ConfigureAwait(false);
+                        await Task.WhenAny(runningTasks.Select(_ => _.Item1).ToList()).ConfigureAwait(false);
```
JoshLove-msft, line 294: nit: shouldn't need the ToList

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -266,141 +263,132 @@ public virtual async Task<Response<UploadLogsResult>> UploadAsync<T>(string rule
...
+                            if (runningTask.Exception != null)
+                            {
+                                exceptions ??= new List<Exception>();
+                                exceptions.Add(runningTask.Exception);
+                                logsFailed += runningTasks[i].Item2;
+                            }
+                            // If current task returned a response that was not a success, log the exception and add number of logs in this task to failed logs
+                            if (runningTask.Result.Status != 204)
+                            {
+                                exceptions ??= new List<Exception>();
```
JoshLove-msft, line 313: I think you can make this a single line:
```c#
exceptions ??= new List<Exception> { new Exception() };
```

JoshLove-msft: Nevermind.. that won't work for the case where the exceptions list is already instantiated.

But consider moving this into a separate method, e.g.:
```c#
private void AddToList(ref List<Exception> exceptions, Exception exception)
```

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -266,141 +264,161 @@ public virtual async Task<Response<UploadLogsResult>> UploadAsync<T>(string rule
...
+                            // If current task returned a response that was not a success, log the exception and add number of logs in this task to failed logs
+                            if (runningTask.Result.Status != 204)
+                            {
+                                exceptions = AddException(
+                                    exceptions,
+                                    runningTask.Exception,
+                                    isRequestFailed: true,
+                                    runningTask.Result);
+                                logsFailed += runningTasks[i].Item2;
+                            }
```
JoshLove-msft, line 323: nit: extract into method.

JoshLove-msft:
Looks good! Just a few minor comments.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -266,141 +264,161 @@ public virtual async Task<Response<UploadLogsResult>> UploadAsync<T>(string rule
...
+            try
+            {
                 // Wait for all the remaining blocks to finish uploading
-                await Task.WhenAll(runningTasks).ConfigureAwait(false);
+                await Task.WhenAll(runningTasks.Select(_ => _.Item1).ToList()).ConfigureAwait(false);
+            }
+            catch (Exception ex)
+            {
+                // If we have an error, add error from response into exceptions list which represents the errors from all the batches
+                exceptions = AddException(
```
JoshLove-msft, line 348: We don't want to add this one since we will end up looping over all tasks later and adding exceptions.

sdk/monitor/Azure.Monitor.Ingestion/tests/SessionRecords/MonitorIngestionLiveTest/ConcurrencyMultiThreadAsync.json
nisha-bhatia: Removed test recordings from this PR, will take a look in https://github.com/Azure/azure-sdk-for-net/issues/31767

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -188,45 +179,52 @@ private static void WriteMemory(MemoryStream stream, ReadOnlyMemory<byte> memory
...
+                    if (response.Status != 204)
+                    {
+                        throw new RequestFailedException(response);
+                    }
+                }
+                catch (Exception ex)
+                {
+                    logsFailed += batch.LogsList.Count;
+                    // If we have an error, add error from response into exceptions list which represents the errors from all the batches
+                    exceptions = AddException(
```
JoshLove-msft, line 219: nit: consider using ref keyword so that we don't have to return the exception list.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -266,141 +264,147 @@ public virtual async Task<Response<UploadLogsResult>> UploadAsync<T>(string rule
...
-            RequestContext requestContext = GenerateRequestContext(cancellationToken);
             Response response = null;
 
-            using HttpMessage message = CreateUploadRequest(ruleId, streamName, batch.LogsData, "gzip", requestContext);
+            using HttpMessage message = CreateUploadRequest(ruleId, streamName, batch.LogsData, Compression, null);
 
             if (async)
             {
-                response = await _pipeline.ProcessMessageAsync(message, requestContext, cancellationToken).ConfigureAwait(false);
+                response = await _pipeline.ProcessMessageAsync(message, null, cancellationToken).ConfigureAwait(false);
```
JoshLove-msft, line 383: We can use pipeline.Send

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -266,141 +264,147 @@ public virtual async Task<Response<UploadLogsResult>> UploadAsync<T>(string rule
...
-            requestContext.AddClassifier(429, false);
-            requestContext.AddClassifier(503, false);
-            return requestContext;
+            exceptions ??= new List<Exception>();
+            exceptions.Add(ex);
+            return exceptions;
         }
 
-        private static UploadLogsStatus GetStatus<T>(IEnumerable<T> logEntries, List<UploadLogsError> errors)
+        private static void ThrowException<T>(IEnumerable<T> logs, DiagnosticScope scope, List<Exception> exceptions, int logsFailed)
```
JoshLove-msft, line 400: Why is this generic?

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -188,45 +179,52 @@ private static void WriteMemory(MemoryStream stream, ReadOnlyMemory<byte> memory
...
+                    exceptions,
+                    ex);
                 }
             }
-            catch (Exception ex)
-            {
-                scope.Failed(ex);
-                throw;
-            }
+            ThrowException(logs, scope, exceptions, logsFailed);
```
JoshLove-msft, line 224: I think this method has enough parameters that it is probably more confusing than helpful. I would just repeat the code in the sync/async methods.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -266,141 +264,147 @@ public virtual async Task<Response<UploadLogsResult>> UploadAsync<T>(string rule
...
+                            // If current task has an exception, log the exception and add number of logs in this task to failed logs
+                            if (runningTask.Exception != null)
+                            {
+                                exceptions = AddException(
+                                    exceptions,
+                                    runningTask.Exception);
+                                logsFailed += runningTasks[i].LogsCount;
+                            }
+                            // If current task returned a response that was not a success, log the exception and add number of logs in this task to failed logs
+                            if (runningTask.Result.Status != 204)
```
JoshLove-msft, line 314: Should this be an else if, like it is here - https://github.com/Azure/azure-sdk-for-net/pull/31846/files#diff-6767d7871ae91cf5eb10fa3e4f24a9d5bbd9f9dbcc22bcbd72be0eabe315c103R358?

Can we extract this into a method, e.g. ProcessCompletedTasks?

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -266,141 +264,147 @@ public virtual async Task<Response<UploadLogsResult>> UploadAsync<T>(string rule
...
                             // Remove completed task from task list
                             runningTasks.RemoveAt(i);
                             i--;
                         }
                     }
                 }
+                catch (Exception ex)
+                {
+                    // If we have an error, add error from response into exceptions list which represents the errors from all the batches
+                    AddException(
```
JoshLove-msft, line 330: I don't think this is necessary as we will loop over all the tasks anyway. We can probably combine this try/catch with the one that starts on line 336.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
nisha-bhatia: I'm not sure of the difference between ProcessMessage and Send - the protocol methods generated ProcessMessage so I kept it the same. Maybe it should be updated?

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
nisha-bhatia: the method didn't have anything to be returned but I deleted it

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
JoshLove-msft: ProcessMessage is for use with RequestContext. Since our convenience methods don't expose RequestContext parameter, we can simplify to use Send.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -188,45 +179,57 @@ private static void WriteMemory(MemoryStream stream, ReadOnlyMemory<byte> memory
...
             Argument.AssertNotNullOrEmpty(ruleId, nameof(ruleId));
             Argument.AssertNotNullOrEmpty(streamName, nameof(streamName));
             Argument.AssertNotNullOrEmpty(logs, nameof(logs));
 
             using var scope = ClientDiagnostics.CreateScope("LogsIngestionClient.Upload");
 
-            RequestContext requestContext = GenerateRequestContext(cancellationToken);
             Response response = null;
-            List<UploadLogsError> errors = new List<UploadLogsError>();
+            var exceptions = new List<Exception>();
```
JoshLove-msft, line 191: nit: we are allocating the exceptions

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -188,45 +179,57 @@ private static void WriteMemory(MemoryStream stream, ReadOnlyMemory<byte> memory
...
                         ruleId,
                         streamName,
-                        false,
+                        async: false,
                         cancellationToken).EnsureCompleted();
 
-                    // If we have an error, add error from response into errors list which represents the errors from all the batches
-                    if (task.Error != null)
-                        errors.Add(task.Error);
+                    if (response.Status != 204)
```
JoshLove-msft, line 210: nit: add directly to exceptions list without throwing.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -188,45 +179,57 @@ private static void WriteMemory(MemoryStream stream, ReadOnlyMemory<byte> memory
...
+                    {
+                        throw new RequestFailedException(response);
+                    }
+                }
+                catch (Exception ex)
+                {
+                    logsFailed += batch.LogsList.Count;
```
JoshLove-msft, line 217: Need to mark scope as failed. Wouldn't this have failed the error test?

sdk/monitor/Azure.Monitor.Ingestion/tests/ErrorTest.cs
```diff
@@ -0,0 +1,112 @@
...
+            entries.Add(new object[] {
+                    new {
+                        Time = Recording.Now.DateTime,
+                        Computer = "Computer" + new string('*', Mb),
+                        AdditionalContext = 1
+                    }
+                });
+
+            // Make the request
+            // Make the request
```
JoshLove-msft, line 65:
```suggestion
```

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -266,141 +269,136 @@ public virtual async Task<Response<UploadLogsResult>> UploadAsync<T>(string rule
...
+                {
+                    // We do not want to log exceptions here as we will loop through all the tasks later
+                }
+            }
 
-                // Process all errors after tasks are done to determine status
-                // Will run on a single thread
-                foreach (Task<BatchUpload> task in runningTasks)
+            // At this point, all tasks have completed. Examine tasks to see if they have exceptions. If Status code != 204, add RequestFailedException to list of exceptions. Increment logsFailed accordingly
+            foreach (var task in runningTasks)
```
JoshLove-msft, line 328: Need to use ProcessCompletedTask

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -266,141 +269,136 @@ public virtual async Task<Response<UploadLogsResult>> UploadAsync<T>(string rule
...
             {
-                status = UploadLogsStatus.PartialFailure;
+                _pipeline.Send(message, cancellationToken);
             }
 
-            return status;
+            return message.Response;
+        }
+
+        private static List<Exception> AddException(ref List<Exception> exceptions, Exception ex)
```
JoshLove-msft, line 397:
```suggestion
        private static void AddException(ref List<Exception> exceptions, Exception ex)
```

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -266,141 +269,136 @@ public virtual async Task<Response<UploadLogsResult>> UploadAsync<T>(string rule
...
+        {
+            exceptions ??= new List<Exception>();
+            exceptions.Add(ex);
+            return exceptions;
```
JoshLove-msft, line 401:
```suggestion
```

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -188,45 +179,57 @@ private static void WriteMemory(MemoryStream stream, ReadOnlyMemory<byte> memory
...
+                    if (response.Status != 204)
+                    {
+                        throw new RequestFailedException(response);
+                    }
+                }
+                catch (Exception ex)
+                {
+                    logsFailed += batch.LogsList.Count;
+                    // If we have an error, add error from response into exceptions list which represents the errors from all the batches
+                    AddException(
```
JoshLove-msft, line 219: nit: indentation for the arguments

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
nisha-bhatia: I think this is okay because we mark the scope as failed on line 227. If we have any exception we would go into the if statemet

JoshLove-msft: Ah yes, my mistake.

JoshLove-msft (approved):

PR #31849, opened 2022-10-17
Update dependencies

JoshLove-msft:
See https://github.com/Azure/azure-amqp/issues/227#issuecomment-1281490365

Also updating extensions version.

### DIFF
--- a/eng/Packages.Data.props
+++ b/eng/Packages.Data.props
@@ -102,7 +102,7 @@
     <PackageReference Update="Azure.ResourceManager" Version="1.3.1" />
 
     <!-- Other approved packages -->
-    <PackageReference Update="Microsoft.Azure.Amqp" Version="2.5.10" />
+    <PackageReference Update="Microsoft.Azure.Amqp" Version="2.5.12" />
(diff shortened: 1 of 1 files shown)


eng/Packages.Data.props
```diff
@@ -102,7 +102,7 @@
     <PackageReference Update="Azure.ResourceManager" Version="1.3.1" />
 
     <!-- Other approved packages -->
-    <PackageReference Update="Microsoft.Azure.Amqp" Version="2.5.10" />
+    <PackageReference Update="Microsoft.Azure.Amqp" Version="2.5.12" />
```
jsquire, line 105: Beat me to it.    Should we consider updating the SB and EH change logs to include this?

jsquire (approved):

eng/Packages.Data.props
JoshLove-msft: Hmm I usually update SB when releasing as opposed to doing it for each change.

jsquire: All good.   I'll throw the EH updates in the other open PR that I've got.

PR #31859, opened 2022-10-18
Add ReadOnlySpan<char> signatures to RequestUriBuilder methods

AlexanderSher:
`RawRequestUriBuilder` creates enormous amount of unnecessary small strings that often get promoted to gen 1. First set of changes is added in https://github.com/Azure/autorest.csharp/pull/2782. These changes in `RequestUriBuilder` are required to reduce these allocations even further.

### DIFF
--- a/sdk/core/Azure.Core/src/RequestUriBuilder.cs
+++ b/sdk/core/Azure.Core/src/RequestUriBuilder.cs
@@ -117,15 +117,13 @@ namespace Azure.Core
             }
         }
 
-        /// <summary> Gets whether or not this instance of <see cref="RequestUriBuilder"/> has path. </summary>
-        public bool HasPath => _pathAndQuery.Length > 0;
+        private bool HasQuery => _queryIndex != -1;
 
-        /// <summary> Gets whether or not this instance of <see cref="RequestUriBuilder"/> has any query information. </summary>
-        public bool HasQuery => _queryIndex != -1;
+        /// <summary> Gets the length of the path in this instance of <see cref="RequestUriBuilder"/>. </summary>
+        protected int PathLength => HasQuery ? _queryIndex : _pathAndQuery.Length;
 
-        private int QueryLength => HasQuery ? _pathAndQuery.Length - _queryIndex : 0;
-
-        private int PathLength => HasQuery ? _queryIndex : _pathAndQuery.Length;
+        /// <summary> Gets the length of the query in this instance of <see cref="RequestUriBuilder"/>. </summary>
+        protected int QueryLength => HasQuery ? _pathAndQuery.Length - _queryIndex : 0;
 
         /// <summary>
         /// Gets the path and query string to the resource referenced by the URI.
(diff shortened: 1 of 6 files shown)


sdk/core/Azure.Core/src/RequestUriBuilder.cs
```diff
@@ -116,7 +117,11 @@ public string Path
             }
         }
 
-        private bool HasQuery => _queryIndex != -1;
+        /// <summary> Gets whether or not this instance of <see cref="RequestUriBuilder"/> has path. </summary>
+        public bool HasPath => _pathAndQuery.Length > 0;
```
annelo-msft, line 121: Curious: Why do `HasPath` and `HasQuery` need to be public?

sdk/core/Azure.Core/src/RequestUriBuilder.cs
```diff
@@ -185,15 +201,21 @@ public void AppendQuery(string name, string value, bool escapeValue)
...
             _pathAndQuery.Append(name);
+#else
+            _pathAndQuery.Append(name.ToString());
+#endif
             _pathAndQuery.Append('=');
-            if (escapeValue && !string.IsNullOrEmpty(value))
+            if (escapeValue && !value.IsEmpty)
             {
-                value = Uri.EscapeDataString(value);
+                value = Uri.EscapeDataString(value.ToString()).AsSpan();
```
annelo-msft, line 212: It looks like we're calling `value.ToString()` multiple times... would it make sense to cache the value if perf is the concern?

sdk/core/Azure.Core/src/RequestUriBuilder.cs
```diff
@@ -116,7 +117,11 @@ public string Path
             }
         }
 
-        private bool HasQuery => _queryIndex != -1;
+        /// <summary> Gets whether or not this instance of <see cref="RequestUriBuilder"/> has path. </summary>
+        public bool HasPath => _pathAndQuery.Length > 0;
+
+        /// <summary> Gets whether or not this instance of <see cref="RequestUriBuilder"/> has any query information. </summary>
+        public bool HasQuery => _queryIndex != -1;
```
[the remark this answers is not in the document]
AlexanderSher: We set `_queryIndex` ourselves and it can't be lower than -1

sdk/core/Azure.Core/src/RequestUriBuilder.cs
AlexanderSher: They can be protected.

christothes, line 121: Shouldn't this be something like:
```suggestion
        public bool HasPath => _pathAndQuery.Length > QueryLength;
```

christothes (commented):
Can we run a micro-benchmark for these changes to compare before and after?

AlexanderSher:
> Can we run a micro-benchmark for these changes to compare before and after?

image

sdk/core/Azure.Core/src/RequestUriBuilder.cs
AlexanderSher: I've missed that there are `PathLength` and `QueryLength` properties that are more useful, so I'll make those protected instead.

AlexanderSher: Replace with protected `PathLength` and `QueryLength`

christothes (approved):

PR #31864, opened 2022-10-18
[Smoke Tests] MSAL Native Interop Fix

jsquire:
# Summary

The focus of these changes is to correct an issue loading the MSAL native interop assembly introduced with the new Azure.Identity.BrokeredAuthentication library.  Going forward, assemblies named to indicate native interoperability will no longer be loaded/type sniffed.

### DIFF
--- a/common/SmokeTests/SmokeTest/Program.cs
+++ b/common/SmokeTests/SmokeTest/Program.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Reflection;
 using System.Runtime.InteropServices;
 using System.Text;
@@ -53,7 +54,7 @@ foreach (var assembly in LoadAssemblies(Assembly.GetEntryAssembly(), AssemblyFil
             ProcessType(type);
         }
     }
-    catch (ReflectionTypeLoadException ex) when (ex.Message.Contains("does not have an implementation"))
+    catch (ReflectionTypeLoadException ex) when (ex.LoaderExceptions.All(lex => lex.Message.Contains("does not have an implementation")))
     {
         // Expected for some assemblies that serve as runtime shims or type-forwarding
         // assemblies.
--- a/common/SmokeTests/SmokeTest/Program.cs
+++ b/common/SmokeTests/SmokeTest/Program.cs
@@ -247,7 +247,7 @@ IEnumerable<Assembly> LoadAssemblies(Assembly rootAssembly, string assemblyFileM
     {
         var assembly = assembliesToProcess.Pop();
 
-        if ((!assembly.FullName.StartsWith("System.")) && (!processedAssemblies.Contains(assembly.FullName)))
+        if ((ShouldLoadAssembly(assembly)) && (!processedAssemblies.Contains(assembly.FullName)))
         {
             processedAssemblies.Add(assembly.FullName);
             yield return assembly;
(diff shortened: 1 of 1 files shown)


pallavit:
What are we testing as part of the smoke tests? Do we intend to load only track2 data plane SDKs only? My understanding is this is done to make the smoke tests green however can in future we may need to exclude more SDKs based on name. And if so, is there a more wholistic condition we can put in.

jsquire:
> What are we testing as part of the smoke tests? Do we intend to load only track2 data plane SDKs only? 

Smoke Tests are intended to load as many Azure SDK libraries into the same process as possible and ensure that there are no conflicts between direct or transitive dependencies.  The libraries that are included come in based on packages published to our Dev Feed, covering T2 client and management libraries.   _(see: [GetPackage-Version.ps1](https://github.com/Azure/azure-sdk-for-net/blob/main/eng/scripts/smoketest/Get-Package-Version.ps1#L26-L28))_

> My understanding is this is done to make the smoke tests green however can in future we may need to exclude more SDKs based on name. 

The goal is to try and avoid false matches without creating a skip-list for individual assemblies.  There are some screened today via the package script, but that's a last resort as it tends to be difficult to maintain and to rot.

In general, failures for platform-specific assemblies are expected and there are a lot of conditions that aim to screen them based on error patterns.  In this particular case, the exception was triggered when loading metadata for the assembly and was unable to be caught.   The best that we can do is avoid the assembly all together.   

The name "Interop" is a common pattern for these platform-specific assembly types, but rather than narrow the criteria too much, we'll go with the less general "NativeInterop" for now until we have a failure that causes us to widen things.

> And if so, is there a more wholistic condition we can put in.

Not sure that I follow.   If you've got suggestions for a pattern, I'm happy to consider them.

pallavit:
Discussed it offline with @jsquire.

common/SmokeTests/SmokeTest/Program.cs
```diff
@@ -53,6 +53,11 @@
             ProcessType(type);
         }
     }
+    catch (ReflectionTypeLoadException ex) when (ex.Message.Contains("does not have an implementation"))
```
jsquire, line 56: I was not able to find an attribute to sniff on the assembly that would let us proactively exclude these.  As much as I hate this magic-string approach, it was the only consistent bit of information that I could find to key on.

common/SmokeTests/SmokeTest/Program.cs
```diff
@@ -247,7 +253,7 @@ IEnumerable<Assembly> LoadAssemblies(Assembly rootAssembly, string assemblyFileM
     {
         var assembly = assembliesToProcess.Pop();
 
-        if ((!assembly.FullName.StartsWith("System.")) && (!processedAssemblies.Contains(assembly.FullName)))
+        if ((ShouldLoadAssembly(assembly)) && (!processedAssemblies.Contains(assembly.FullName)))
```
heaths, line 256: For readability, I'm always happen to see extraneous parens...but around single operands seems excessive and actually hurts readability, IMO:
```suggestion
        if (ShouldLoadAssembly(assembly) && !processedAssemblies.Contains(assembly.FullName))
```

heaths (approved):

common/SmokeTests/SmokeTest/Program.cs
jsquire: I'm a fan of explicit parenthesis, personally.   Since the rest of the code is written in this style, I'm going to keep the original formatting for consistency with surrounding code.   If you'd prefer to reformat, I'm not strongly opposed but ask that you do a pass over the entire file in a separate PR.

heaths: Not passionate enough to reformat no. But as pointed out, I'm also a fan of explicit parens, but around a single operand? It adds nothing. It just makes reviewers have to parse even more parens that add no value. Around something like `((foo && bar) || baz)` sure - that's the behavior, but the parens make it more obvious for readers, especially if they don't know operator precedence as well. But taking that same example, how does parens around `baz` provide value and instead just not make it less readable: `((foo && bar) || (baz))`?

PR #31869, opened 2022-10-18
Improve AcceptSession docs

JoshLove-msft:
Addresses https://github.com/Azure/azure-sdk-for-net/issues/31868

### DIFF
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Client/ServiceBusClient.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Client/ServiceBusClient.cs
@@ -385,8 +385,8 @@ namespace Azure.Messaging.ServiceBus
         /// <summary>
         /// Creates a <see cref="ServiceBusSessionReceiver"/> instance that can be used for receiving
         /// and settling messages from a session-enabled queue by accepting the next unlocked session that contains Active messages. If there
-        /// are no unlocked sessions with Active messages, then the call will timeout after the configured <see cref="RetryOptions.TryTimeout"/> value and throw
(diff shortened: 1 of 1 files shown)


sdk/servicebus/Azure.Messaging.ServiceBus/src/Client/ServiceBusClient.cs
```diff
@@ -420,8 +423,11 @@ public virtual ServiceBusSender CreateSender(string queueOrTopicName, ServiceBus
 
         /// <summary>
         /// Creates a <see cref="ServiceBusSessionReceiver"/> instance that can be used for receiving
-        /// and settling messages from a specific session-enabled subscription. It uses <see cref="ServiceBusReceiveMode"/> to specify
-        /// how messages are received. Defaults to PeekLock mode. The <see cref="ServiceBusReceiveMode"/> is set in <see cref="ServiceBusReceiverOptions"/>.
+        /// and settling messages from a session-enabled subscription by accepting the next unlocked session that contains Active messages. If there
+        /// are no unlocked sessions with Active messages, then the call will timeout after the configured <see cref="RetryOptions.TryTimeout"/> value and throw
```
jsquire: What will it throw?   I think we're also missing an ending period.   Same for L389

jsquire (approved):

sdk/servicebus/Azure.Messaging.ServiceBus/src/Client/ServiceBusClient.cs
JoshLove-msft: Huh? It's on the next line.

jsquire: that was obscured by the validation errors that ran 1/2 my screen.   I see it now.  😄

PR #31886, opened 2022-10-19
Fix BlobBaseClient.SetNameFields - trim slashes according to ClientCo…

slab-msft:
Follow up fix for newly introduced support for leading and trailing '/' in v. 12.14.0 (2022-10-12).

If a new BlobBaseClient is created directly from a URI with BlobClientOptions, TrimBlobNameSlashes set to true
e.g.
https://accountName.blob.core.windows.net/containerName///blobName.

The URI would be indeed correctly preserved with the leading slash, but the .Name property would be trimmed, therefore not reflecting the actual blobName in the container.

### DIFF
--- a/sdk/storage/Azure.Storage.Blobs/tests/BlobBaseClientTests.cs
+++ b/sdk/storage/Azure.Storage.Blobs/tests/BlobBaseClientTests.cs
@@ -7382,31 +7382,6 @@ namespace Azure.Storage.Blobs.Test
         }
 
         [RecordedTest]
-        public void SetNameFields_TrimBlobNameSlashes([Values("blobname", "/blobname", "//blobname", "vdir//blobname", "/vdir//blobname")] string blobName, [Values(true, false)] bool trimBlobNameSlashes)
-        {
-            // arrange
-            var accountName = "account";
-            var containerName = "container";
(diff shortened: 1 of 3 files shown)


sdk/storage/Azure.Storage.Blobs/tests/BlobBaseClientTests.cs
```diff
@@ -7381,6 +7381,31 @@ public async Task CanGetParentContainerClient_WithContainerSAS()
             Assert.IsNotNull(blobItems);
         }
 
+        [RecordedTest]
```
jaschrep-msft, line 7384: For organizational purposes, could you move these tests to `BlobUriBuilderTests.cs` along with the other related tests from [this PR](https://github.com/Azure/azure-sdk-for-net/pull/30271)? Will be good to go once that is handled.

jaschrep-msft (changes requested):

sdk/storage/Azure.Storage.Blobs/tests/BlobBaseClientTests.cs
slab-msft: Sure, I have updated the PR

jaschrep-msft (approved):

PR #31892, opened 2022-10-19
[For Discussion] RetryPolicy/GeoRedundantPolicy exploration

annelo-msft:
### Description
Some thoughts in response to @JoshLove-msft's work on RetryPolicy:
- [RetryPolicy Gist](https://gist.github.com/JoshLove-msft/28993cd17754b2b23671092b7c45336d)
- [Prior work by @JoshLove-msft that this is building on](https://github.com/Azure/azure-sdk-for-net/pull/31187)

### Motivating Scenarios
- Implement [Polly](https://github.com/App-vNext/Polly#usage--fault-handling-reactive-policies) retry policy: [Sample](https://github.com/Azure/azure-sdk-for-net/pull/31892/files#diff-35c33a1cf3598d8668ff885a81242466d8dfe659fe245c32d628ebfe5f22af8bR12-R15)
- Enable GeoRedundant policy logic to be implemented with knowledge of retries. [Sample](https://gist.github.com/JoshLove-msft/28993cd17754b2b23671092b7c45336d#option-2---retrycontext-on-httpmessage)
- Let callers compose simple retry logic conditions: [Sample](https://github.com/Azure/azure-sdk-for-net/pull/31892/files#diff-77158d2866ddb3696077cbc90380dfca5694a3e2d134995c0b44429f6562df09R22)

### Design Goals
- Pipeline policy PerCall/PerRetry semantics are respected
- Single responsibility: retry logic lives in a single policy
- Design indicates RetryPolicy replacement is an advanced scenario
- Precedence of modifications to policy is clear and respected: Per invocation (end-user)/Per client (end-user)/Per client (library author)
- Avoid adding lots of policies to the pipeline
- Library author can do same modifications to retry logic as end-user, but end-user options take precedence
- Safeguard against policy replacement having negative side effects elsewhere in the pipeline
- Leverage patterns that appear elsewhere in Core (i.e. condition/classifier chaining, APIs for modifying pipeline).

### DIFF
--- a/sdk/core/Azure.Core/SampleApp/SampleClient.cs
+++ b/sdk/core/Azure.Core/SampleApp/SampleClient.cs
@@ -1,0 +1,58 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Azure;
+using Azure.Core;
+using Azure.Core.Pipeline;
+
+namespace SampleApp
+{
+    public class SampleClient
+    {
+        private readonly HttpPipeline _pipeline;
+
+        internal static int MyServiceSLAMinutes;
+
+        public SampleClient(Uri endpoint, TokenCredential credential, ClientOptions options)
+        {
+            // Pipeline author can add retry conditions or override the default retry policy
+            HttpPipelineOptions pipelineOptions = new HttpPipelineOptions(options);
+            pipelineOptions.RetryConditions.Add(new GlobalTimeoutRetryCondition(TimeSpan.FromMinutes(MyServiceSLAMinutes)));
+
+            _pipeline = HttpPipelineBuilder.Build(pipelineOptions);
+        }
+
+        public Response GetWithSpecialRetries(RequestContent content, RequestContext context = default)
+        {
+            HttpMessage message = _pipeline.CreateMessage(context);
+
+            // In reality, we would use the protocol method send that takes context, but it's not in Core yet.
+            _pipeline.Send(message, context.CancellationToken);
+            return message.Response;
+        }
+    }
+
+    public class GlobalTimeoutRetryCondition : RetryCondition
+    {
+        private TimeSpan _maxTime;
+
+        public GlobalTimeoutRetryCondition(TimeSpan maxTime)
+        {
+            _maxTime = maxTime;
+        }
+
+        public override bool TryGetShouldRetry(HttpMessage message, out bool shouldRetry)
+        {
+            if ((message.PipelineContext.CreatedOn + _maxTime) >= DateTimeOffset.UtcNow)
+            {
+                shouldRetry = false;
+                return true;
+            }
+
+            shouldRetry = true;
+            return false;
+        }
+    }
+}
--- a/sdk/core/Azure.Core/api/Azure.Core.net461.cs
+++ b/sdk/core/Azure.Core/api/Azure.Core.net461.cs
@@ -199,6 +199,7 @@ namespace Azure
         public void AddClassifier(int statusCode, bool isError) { }
         public void AddPolicy(Azure.Core.Pipeline.HttpPipelinePolicy policy, Azure.Core.HttpPipelinePosition position) { }
         public static implicit operator Azure.RequestContext (Azure.ErrorOptions options) { throw null; }
+        public void ReplacePolicy(Azure.Core.Pipeline.HttpPipelinePolicy policy, Azure.Core.PipelinePolicyReplacement replace) { }
     }
     public partial class RequestFailedException : System.Exception, System.Runtime.Serialization.ISerializable
     {
@@ -362,6 +363,7 @@ namespace Azure.Core
         public override bool Equals(object? obj) { throw null; }
         [System.ComponentModel.EditorBrowsableAttribute(System.ComponentModel.EditorBrowsableState.Never)]
         public override int GetHashCode() { throw null; }
+        public void ReplacePolicy(Azure.Core.Pipeline.HttpPipelinePolicy policy, Azure.Core.PipelinePolicyReplacement replace) { }
         [System.ComponentModel.EditorBrowsableAttribute(System.ComponentModel.EditorBrowsableState.Never)]
         public override string? ToString() { throw null; }
     }
@@ -451,6 +453,7 @@ namespace Azure.Core
         public System.Threading.CancellationToken CancellationToken { get { throw null; } }
         public bool HasResponse { get { throw null; } }
         public System.TimeSpan? NetworkTimeout { get { throw null; } set { } }
+        public Azure.Core.PipelineContext PipelineContext { get { throw null; } }
         public Azure.Core.Request Request { get { throw null; } }
         public Azure.Response Response { get { throw null; } set { } }
         public Azure.Core.ResponseClassifier ResponseClassifier { get { throw null; } set { } }
@@ -469,6 +472,16 @@ namespace Azure.Core
     {
         public static Azure.Response[] Parse(Azure.Response response, bool expectCrLf, System.Threading.CancellationToken cancellationToken) { throw null; }
         public static System.Threading.Tasks.Task<Azure.Response[]> ParseAsync(Azure.Response response, bool expectCrLf, System.Threading.CancellationToken cancellationToken) { throw null; }
+    }
+    public partial class PipelineContext
+    {
+        internal PipelineContext() { }
+        public System.DateTimeOffset CreatedOn { get { throw null; } }
+        public int RetryAttempt { get { throw null; } set { } }
+    }
+    public enum PipelinePolicyReplacement
+    {
+        RetryPolicy = 0,
     }
     public abstract partial class Request : System.IDisposable
     {
@@ -649,6 +662,11 @@ namespace Azure.Core
         public bool TryGetValue(string name, out string? value) { throw null; }
         public bool TryGetValues(string name, out System.Collections.Generic.IEnumerable<string>? values) { throw null; }
     }
+    public abstract partial class RetryCondition
+    {
+        protected RetryCondition() { }
+        public abstract bool TryGetShouldRetry(Azure.Core.HttpMessage message, out bool shouldRetry);
+    }
     public enum RetryMode
     {
         Fixed = 0,
@@ -662,6 +680,7 @@ namespace Azure.Core
         public int MaxRetries { get { throw null; } set { } }
         public Azure.Core.RetryMode Mode { get { throw null; } set { } }
         public System.TimeSpan NetworkTimeout { get { throw null; } set { } }
+        public void AddRetryCondition(Azure.Core.RetryCondition condition) { }
     }
     public partial class StatusCodeClassifier : Azure.Core.ResponseClassifier
     {
@@ -958,6 +977,8 @@ namespace Azure.Core.Pipeline
         public System.Collections.Generic.IList<Azure.Core.Pipeline.HttpPipelinePolicy> PerRetryPolicies { get { throw null; } }
         public Azure.Core.RequestFailedDetailsParser RequestFailedDetailsParser { get { throw null; } set { } }
         public Azure.Core.ResponseClassifier? ResponseClassifier { get { throw null; } set { } }
+        public System.Collections.Generic.IList<Azure.Core.RetryCondition> RetryConditions { get { throw null; } }
+        public Azure.Core.Pipeline.HttpPipelinePolicy? RetryPolicy { get { throw null; } set { } }
     }
     public abstract partial class HttpPipelinePolicy
     {
--- a/sdk/core/Azure.Core/src/Pipeline/Internal/RetryPolicy.cs
+++ b/sdk/core/Azure.Core/src/Pipeline/Internal/RetryPolicy.cs
@@ -17,15 +17,22 @@ namespace Azure.Core.Pipeline
         private readonly TimeSpan _delay;
         private readonly TimeSpan _maxDelay;
         private readonly int _maxRetries;
+        private readonly List<RetryCondition> _retryConditions;
 
         private readonly Random _random = new ThreadSafeRandom();
 
-        public RetryPolicy(RetryMode mode, TimeSpan delay, TimeSpan maxDelay, int maxRetries)
+        public RetryPolicy(RetryMode mode, TimeSpan delay, TimeSpan maxDelay, int maxRetries, IEnumerable<RetryCondition> retryConditions)
         {
             _mode = mode;
             _delay = delay;
             _maxDelay = maxDelay;
             _maxRetries = maxRetries;
+
+            _retryConditions = new List<RetryCondition>();
+            _retryConditions.AddRange(retryConditions);
+
+            // Add the Max Attempts retry condition to the end of the list to ensure we have a default condition.
+            _retryConditions.Add(new MaxAttemptsRetryCondition(maxRetries));
         }
 
         private const string RetryAfterHeaderName = "Retry-After";
@@ -44,7 +51,7 @@ namespace Azure.Core.Pipeline
 
         private async ValueTask ProcessAsync(HttpMessage message, ReadOnlyMemory<HttpPipelinePolicy> pipeline, bool async)
         {
-            int attempt = 0;
+            message.PipelineContext.RetryAttempt = 0;
             List<Exception>? exceptions = null;
             while (true)
             {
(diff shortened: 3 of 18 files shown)


sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -958,6 +977,8 @@ public partial class HttpPipelineOptions
         public System.Collections.Generic.IList<Azure.Core.Pipeline.HttpPipelinePolicy> PerRetryPolicies { get { throw null; } }
         public Azure.Core.RequestFailedDetailsParser RequestFailedDetailsParser { get { throw null; } set { } }
         public Azure.Core.ResponseClassifier? ResponseClassifier { get { throw null; } set { } }
+        public System.Collections.Generic.IList<Azure.Core.RetryCondition> RetryConditions { get { throw null; } }
```
annelo-msft, line 980: - Library author has same knobs for modifying/replacing retry policy as end-user
- Precedence rules are the same as for response classification
  - Per-invocation takes precedent over
  - Per-client specified by end-user over
  - Library author modifications
  - Default pipeline implementation

sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -470,6 +473,16 @@ public static partial class MultipartResponse
         public static Azure.Response[] Parse(Azure.Response response, bool expectCrLf, System.Threading.CancellationToken cancellationToken) { throw null; }
         public static System.Threading.Tasks.Task<Azure.Response[]> ParseAsync(Azure.Response response, bool expectCrLf, System.Threading.CancellationToken cancellationToken) { throw null; }
     }
+    public partial class PipelineContext
+    {
+        internal PipelineContext() { }
+        public System.DateTimeOffset CreatedOn { get { throw null; } }
+        public int RetryAttempt { get { throw null; } set { } }
```
annelo-msft, line 480: RetryAttempt value is available downstream from RetryPolicy in the pipeline, for use in GeoRedunant policy implementations.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -470,6 +473,16 @@ public static partial class MultipartResponse
...
+    }
+    public enum PipelinePolicyReplacement
+    {
+        RetryPolicy = 0,
```
annelo-msft, line 484: We can expand this list if needed.

sdk/core/Azure.Core/src/Pipeline/Internal/RetryPolicy.cs
```diff
@@ -212,4 +233,28 @@ private TimeSpan CalculateExponentialDelay(int attempted)
                     _maxDelay.TotalMilliseconds));
         }
     }
+
+#pragma warning disable SA1402 // File may only contain a single type
+    internal class MaxAttemptsRetryCondition : RetryCondition
```
annelo-msft, line 238: We could also make this public if we wanted to provide a library of possible retry conditions.

sdk/core/Azure.Core/src/PipelineContext.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// <summary>
+        /// The number indicating which retry attempt is currently in progress.
+        /// </summary>
+        public int RetryAttempt
+        {
+            get
+            {
+                if (_retryAttempt == null)
+                {
+                    throw new InvalidOperationException("Tried to read RetryAttempt when it did not have a value. " +
```
annelo-msft, line 37: - This ensures that someone won't try to depend on this value downstream from a custom retry policy that forgets to set the attempt number.
- The assumption is that any issues would be caught at development time.

sdk/core/Azure.Core/SampleApp/SampleClient.cs
```diff
@@ -0,0 +1,58 @@
...
+    public class GlobalTimeoutRetryCondition : RetryCondition
+    {
+        private TimeSpan _maxTime;
+
+        public GlobalTimeoutRetryCondition(TimeSpan maxTime)
+        {
+            _maxTime = maxTime;
+        }
+
+        public override bool TryGetShouldRetry(HttpMessage message, out bool shouldRetry)
```
JoshLove-msft, line 46: The one thing that is missing here from the mutable context approach is the ability for each condition to know what the current state of ShouldRetry is.

sdk/core/Azure.Core/src/Pipeline/Internal/RetryPolicy.cs
JoshLove-msft: What is the benefit of this as opposed to the RetryOptions setting?

sdk/core/Azure.Core/src/PipelineContext.cs
JoshLove-msft: I don't think we can rely on this. We'd need to extract the setting of the context into a policy that goes after RetryPolicy.

annelo-msft:
Reopening per request.

sdk/core/Azure.Core/SampleApp/SampleClient.cs
annelo-msft: Yeah, that's not the intention of this pattern.  This pattern basically implements a chain of discrete conditions where any one can decide whether or not to retry, and the first one that does ends the chain.  If different interaction logic between the components being composed is desired, you'd probably want to use a different pattern.  Although, the intention is to make it easy to reason about and debug any issues, so if there's a lot of interaction between logical components, it might be better to go back to the abstract RetryPolicy, and let someone write out the logic in an override of a ShouldRetry() method.

sdk/core/Azure.Core/src/PipelineContext.cs
annelo-msft: Why is the extra policy required?

sdk/core/Azure.Core/src/Pipeline/Internal/RetryPolicy.cs
annelo-msft: I don't understand the question ... which RetryOptions setting?

annelo-msft: Ah, maybe I understand ... this works with the RetryOptions list of RetryConditions ... it's a concrete implementation of a RetryCondition, which we could use by default to maintain the existing RetryPolicy behavior.

sdk/core/Azure.Core/src/PipelineContext.cs
JoshLove-msft: The reason for having the extra policy is that we would be able to ensure the context values are set correctly. Having a runtime check doesn't guarantee that the context is set correctly.

sdk/core/Azure.Core/src/Pipeline/Internal/RetryPolicy.cs
JoshLove-msft: So if retry conditions are specified, the other retry options are ignored?

sdk/core/Azure.Core/api/Azure.Core.net461.cs
JoshLove-msft: I think this is an interesting pattern for how to grow the list of replaceable policies. Pragmatically, I'm not sure we would grow it beyond RetryPolicy though.

sdk/core/Azure.Core/src/Pipeline/Internal/RetryPolicy.cs
annelo-msft: No, everything in RetryOptions is applied if you use the internal RetryPolicy.

JoshLove-msft: Ah, I see. This is how we implement the MaxAttempts logic.

sdk/core/Azure.Core/src/PipelineContext.cs
annelo-msft: Makes sense.  It might be worth exploring if there's any other way to approach that that avoids adding an extra policy.  Or maybe it is better to not rely on other policies to set Retry/Processing/PipelineContext, in which case it might make sense to consider what else could go on the context and where it would be best to put the other policy.  For example, we currently do [runtime validation that the Response is set](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/HttpMessage.cs#L54) -- would it be better to do that in the extra policy?  Would we always want to have that policy in the pipeline, or only when there are policies that rely on RetryContext being set?

JoshLove-msft: I think we'd always want this extra policy to ensure the context is set regardless of whether the user replaces the retry policy or not. If it turns out this is not feasible, then I think we could use some sort of runtime validation like you have suggested.

annelo-msft: I had understood from Krzysztof at some point that it was a goal to try not to add new required policies to the pipeline, but you may want to double check that with him and how concerned about that he still is.

sdk/core/Azure.Core/SampleApp/SampleClient.cs
JoshLove-msft: Couldn't we pass in the list of conditions and provide a helper method like GetNextShouldRetry?

sdk/core/Azure.Core/src/PipelineContext.cs
JoshLove-msft: Pinged @KrzysztofCwalina

KrzysztofCwalina: > This ensures that someone won't try to depend on this value downstream from a custom retry policy that forgets to set the attempt number.

I don't think we should be throwing from this property. It should start with  the first value (1 I think) and we should assume that the retry policy increments it properly.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
KrzysztofCwalina: Could we call it RetryCount? Also, I assume it starts with the value of zero, i.e. no retries attempted yet.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
KrzysztofCwalina: Why don't we use HttpPipelinePosition and just add RetryPolicy value

sdk/core/Azure.Core/src/PipelineContext.cs
JoshLove-msft: Discussed offline - @KrzysztofCwalina suggested just leaving the context fully mutable and not adding any kind of validation. It it is the user's responsibility to update the context if they override the policy.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -199,6 +199,7 @@ public partial class RequestContext
         public void AddClassifier(int statusCode, bool isError) { }
         public void AddPolicy(Azure.Core.Pipeline.HttpPipelinePolicy policy, Azure.Core.HttpPipelinePosition position) { }
         public static implicit operator Azure.RequestContext (Azure.ErrorOptions options) { throw null; }
+        public void ReplacePolicy(Azure.Core.Pipeline.HttpPipelinePolicy policy, Azure.Core.PipelinePolicyReplacement replace) { }
```
KrzysztofCwalina, line 202: I would call it SetPolicy. .NET naming conventions for collections are (Add/Set/Remove)

sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -649,6 +662,11 @@ public partial class ResponseClassifier
         public bool TryGetValue(string name, out string? value) { throw null; }
         public bool TryGetValues(string name, out System.Collections.Generic.IEnumerable<string>? values) { throw null; }
     }
+    public abstract partial class RetryCondition
```
KrzysztofCwalina, line 665: Why do we need this type? Shouldn't the RetryPolicy be doing reties and so it knows then to retry?

sdk/core/Azure.Core/src/PipelineContext.cs
annelo-msft: Thanks for following up on that Josh!

sdk/core/Azure.Core/src/Pipeline/Internal/RetryPolicy.cs
```diff
@@ -44,7 +51,7 @@ public override ValueTask ProcessAsync(HttpMessage message, ReadOnlyMemory<HttpP
 
         private async ValueTask ProcessAsync(HttpMessage message, ReadOnlyMemory<HttpPipelinePolicy> pipeline, bool async)
         {
-            int attempt = 0;
+            message.PipelineContext.RetryAttempt = 0;
```
KrzysztofCwalina, line 54: Why are we setting this value here to 0? I think the ctor for the context should set it to zero and then the retry policy should be only incrementing it.

KrzysztofCwalina (changes requested):
Thanks for implementing it so quickly! I added some comments, most small. The only one a bit more substantial is about the RetryCondition. I would love to understand why we need that.

sdk/core/Azure.Core/SampleApp/SampleClient.cs
annelo-msft: Yeah, I'm open to whatever logic we decide we need for the feature.  We designed the ResponseClassifier chaining pattern to meet the following criteria:
- Each classifier can say whether it does or doesn't "have an opinion" on whether the response is an error (i.e. one that classifies 409 responses doesn't have an opinion on a 404 response) -- that's the "TryGet" semantics
- Having no interactions across classifiers makes it simple to reason about/ no tie-breaking is needed if classifiers disagree
- Precedence is specified via how the chain of classifiers is ordered
- The last item in the chain provides the default classification logic, so you don't end up with no one deciding whether something is an error or not.

In the case you describe above, I'd be curious how you'd break ties if one condition wanted to retry and another didn't, and one didn't want to vote at all if it wasn't relevant to a given response.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
annelo-msft: The goal of making this suggestion in this Draft Discussion PR, was to suggest an approach to introducing the "composition" feature without needing multiple additional policies in the pipeline.  There is a discussion regarding this pattern here: https://github.com/Azure/azure-sdk-for-net/pull/31892#discussion_r999782115

sdk/core/Azure.Core/api/Azure.Core.net461.cs
JoshLove-msft: In the approach I proposed, I had this called AttemptNumber and it starts as 1. I think this is a little bit more intuitive, but am open to other opinions.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
JoshLove-msft: Do we really think this list will grow enough to warrant the separate method, and type?

sdk/core/Azure.Core/api/Azure.Core.net461.cs
JoshLove-msft: That is an interesting idea. We could probably even not add the new SetPolicy method as a user could just call AddPolicy and pass in RetryPolicy value. It does muddy the semantics somewhat though.

annelo-msft: Yeah, I would worry about anyone calling AddPolicy now being able to override the RetryPolicy easily if we added it to HttpPipelinePosition.  I liked that AddPolicy was more causal and Replace/Set policy was more advanced.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
annelo-msft: I really just threw it out there as a suggestion to consider with the others.  The tradeoff is potentially adding more policies that can be replaced to ClientOptions vs. having one thing that grows without increasing the ClientOptions API surface area.  The other piece to consider is consistency across different approaches to setting - would it make sense to use the same API for library authors and protocol method callers, would we be happy with this on RequestContext, etc?  My instinct is to make things more general to be able to respond to unforseen changes later, but if we really think we won't add more policies, I agree the overhead of extra types would be unnecessary.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
KrzysztofCwalina: I think we have too many places that can make retry decisions already.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
JoshLove-msft: I mean, they are sort of adding the policy so maybe it is okay. They are adding a retry policy and the constraints are that there is only 1 retry policy.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
annelo-msft: Sounds good - thanks for considering my ideas, and excited to see your implementation, @JoshLove-msft!

sdk/core/Azure.Core/api/Azure.Core.net461.cs
christothes: I think the other consideration is usability and discoverability for users. While I like the push towards generalizing the design, my gut says it won't add enough additional policies to justify the UX downsides.

PR #31900, opened 2022-10-20
Adding BlobUriBuilder overload which takes only accountName

### LINKED ISSUE #31897: [FEATURE REQ] BlobUriBuilder as a real URI builder
### Library name

Azure.Storage.Blobs

I'd like to see if we can make BlobUriBuilder an actual URI builder as I find it frustrating and odd that I have to give it a URI to begin with. I would rather prefer it to have overloads like:

```csharp
BlobUriBuilder(string accountName)
BlobUriBuilder(string accountName, string containerName)
BlobUriBuilder(string accountName, string containerName, string blobPath)
```

and have it actually build the URI for me. In this way, I don't, as a new user, have to go look up that whole URI format (`.blob.core.windows.net`) and hard-code it into my usage which feels brittle.

Consider System.UriBuilder:

I recognize we might need some nuance to support private/gov't clouds but I think we can figure it out :)

brandonh-msft:
Builds the default target URL based on Public cloud

Fixes #31897


sdk/storage/Azure.Storage.Blobs/src/BlobUriBuilder.cs
```diff
@@ -161,6 +161,20 @@ public string Query
...
+        /// Initializes a new instance of the <see cref="BlobUriBuilder"/>
+        /// class for the specified Azure Storage Account.
+        /// </summary>
+        /// <param name="accountName">
+        /// The account name of the target storage account.
+        /// </param>
+        /// <remarks>
+        /// This defaults to Azure Public Cloud - i.e. a blob.core.windows.net suffix on the URL to the storage account
+        /// </remarks>
+        public BlobUriBuilder(string accountName) : this(new Uri($"https://{accountName}.blob.core.windows.net"))
```
amnguye, line 174: Accepting only the storage account name as a default, and assuming the ending suffix (e.g. blob.core.windows.net) is not a good idea.

We should be taking in the ending suffix of the endpoint. This would not be a good experience for customers who are new to storage and have a different type of endpoint details.

brandonh-msft:
> Hi @brandonh-msft Brandon Hurlburt FTE. Thank you for your interest in helping to improve the Azure SDK experience and for your contribution. We've noticed that there hasn't been recent engagement on this pull request. If this is still an active work stream, please let us know by pushing some changes or leaving a comment. Otherwise, we'll close this out in 7 days.

it is still my belief this is a worthwhile change to improve the developer experience

sdk/storage/Azure.Storage.Blobs/src/BlobUriBuilder.cs
```diff
@@ -161,6 +164,20 @@ public string Query
...
+        /// Initializes a new instance of the <see cref="BlobUriBuilder"/>
+        /// class for the specified Azure Storage Account.
+        /// </summary>
+        /// <param name="accountName">
+        /// The account name of the target storage account.
+        /// </param>
+        /// <param name="storageSuffix">
+        /// The suffix of the storage cloud. Defaults to Azure public cloud (blob.core.windows.net).
+        /// </param>
+        public BlobUriBuilder(string accountName, string storageSuffix = "blob.core.windows.net") : this(new Uri($"https://{accountName}.{storageSuffix.TrimStart('.')}"))
```
amnguye, line 177: Seeing as my other comment got marked as resolved since the source changed.

Accepting only the storage account name as a default, and assuming the ending suffix (e.g. blob.core.windows.net) is not a good idea.

We should be taking in the ending suffix of the endpoint. This would not be a good experience for customers who are new to storage and have a different type of endpoint details.

amnguye:
This would not be a good experience for customers overall. Maybe for your exact scenario and other customers who ONLY have storage accounts that have ending suffixes in "core.windows.net" would benefit.

But as customers who start making new accounts, (e.g. some accounts can have "storage.azure.net") and will in the future have this ending suffix. Also as DNS space will run out in the future the endpoints will look like http://storageaccount.<Zone>.<Service>.storage.azure.net.

Please see the read through the reply I made on your github post 
https://github.com/Azure/azure-sdk-for-net/issues/31897#issuecomment-1284642275
https://github.com/Azure/azure-sdk-for-net/issues/31897#issuecomment-1285855582

Making this a default, provides no way out for the SDK in the future to change it as we cannot make breaking changes in the future if the "new" common ending suffix becomes something else. We do not have telemetry to support this default, and we should not make a decision now on the default for this.

Also a lot of customers in storage look to the SDK as a form of "documentation" for the service functionality. And defining a default here would be saying to our customers that their URLs should have this ending suffix.

We want our customers to write flexible code as well. So if they know ahead of time "hey the ending suffix could change / be different" then having them provide the full URL would save them a lot of heartache in the future and long run.

I've dealt with many customer issues where they were manually constructing URLs and then realizing, the ending suffix changed when they started making new storage accounts, thus breaking their applications code.

brandonh-msft:
I'm still failing to see an issue as my approach is flexible and [there are other places where we make this same assumption](https://learn.microsoft.com/EN-us/azure/azure-functions/functions-reference?tabs=blob#connecting-to-host-storage-with-an-identity-preview).

sdk/storage/Azure.Storage.Blobs/src/BlobUriBuilder.cs
brandonh-msft: this code _does_ take in the suffix of the endpoint...

amnguye:
> I'm still failing to see an issue as my approach is flexible and [there are other places where we make this same assumption](https://learn.microsoft.com/EN-us/azure/azure-functions/functions-reference?tabs=blob#connecting-to-host-storage-with-an-identity-preview).

Just because another Azure product making setting a precedent doesn't make it a correct precedent to follow.

I'm also failing to see why your approach is the correct approach. If you want you can set up a meeting with me and my team to discuss this feature. Because at the end of the day since this is an API change this something that my team would have to choose to support and push through getting officially approved.

I've explained it already that endpoints will change in the future. Setting a default here of any value will set the user to believe this is should be the endpoint. If they ever get a new storage account with a different endpoint this would break them cause they're using the default thus creating a customer issue which my team has to deal with.

Issue #29600, opened 2022-06-30
CLU Authoring APIs and AAD support not working on 2022-05-15-preview


heaths:
Using both the UK South and Western Europe regions, testing our CLU authoring APIs or AAD support returns a 404 Not Found. For now we will test these only on 2022-05-01 but any customers trying this with the new CLU 1.1 beta SDKs will get the failure unless they specifically use the 2022-05-01 version for those authoring calls and do not use AAD for conversation summarization and PII extraction.

heaths:
This was previously fixed.

heaths:
Reopened because the problem was fixed but I forgot I still have to unignore a few tests.

heaths:
Turns out that AAD is not working for the inference client - only authoring.

PR #31901, opened 2022-10-20
Regenerate on 2022-10-01-preview

### LINKED ISSUE #31895: Regenerate on 2022-10-01-preview for CLU
See Azure/azure-sdk-for-python#26194 and related PR for congruency.

heaths:
Resolves #31895 and fixes most of #29600.


sdk/cognitivelanguage/Azure.AI.Language.Conversations/src/autorest.md
```diff
@@ -146,19 +146,19 @@ directive:
...
     var apiVersionParam = $.parameters[apiVersionParamIndex];
+    $.parameters.splice(apiVersionParamIndex, 1);
 
-    $.parameters.splice(stringIndexTypeParamIndex, 1);
-    $.parameters.splice(apiVersionParamIndex - 1, 1);
-
-    $.parameters.push(stringIndexTypeParam);
+    var trainedModelLabelIndex = $.parameters.findIndex(param => param.name === "trainedModelLabel");
+    $.parameters.splice(trainedModelLabelIndex, 0, stringIndexTypeParam);
     $.parameters.push(apiVersionParam);
```
heaths, line 162: @tg-msft could you take a look at this? I had to do something kind of gnarly to keep DPG parameter org consistent with our guidelines, especially since all the optional parameters are strings. The generated code, since the swagger parameter was inserted second to last (before api-version), effectively injected a parameter in the "middle" because you had me move `stringIndexType` last (well, before api-version, which we don't expose in the methods anyway).

I think this points out that we shouldn't mess with parameter order. If there's concern, we should make sure ordering issues are done in swagger or Cadl. I wouldn't expect autorest transforms like this (or whatever we'd use for Cadl) to be obvious, and straight-up auto-generating a client might not've caught this without test coverage (and maybe not even then depending on how the service responds).

/cc @m-nash

heaths: One possibility at least with Cadl since we'd have some better versioning info is that if such a change is detected, perhaps we just generate a new method (keeping the old as-is) that uses an options class. Too bad there's no way to required named properties instead of having to support positional, too, but an options class would provide the same benefit.

heaths: I talked a bit with @m-nash about this issue this morning, and he said this happens constantly in management and he's been thinking about mitigations. 👍🏻

heaths:
I opened bug #31947 to track adding tests for new inference features once our sub is approved for the gated preview. @kristapratico is following up with the service team in Teams.

sdk/cognitivelanguage/Azure.AI.Language.Conversations/src/Authoring/ConversationAuthoringClient.cs
```diff
@@ -72,5 +74,213 @@ public ConversationAuthoringClient(Uri endpoint, TokenCredential credential, Con
...
+        ///   ], # Optional. The errors encountered while executing the job.
+        /// }
+        /// </code>
+        ///
+        /// </remarks>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+#pragma warning disable AZC0002 // DO ensure all service methods, both asynchronous and synchronous, take an optional CancellationToken parameter called cancellationToken.
+        public virtual Operation<BinaryData> ExportProject(WaitUntil waitUntil, string projectName, string exportedProjectFormat, string assetKind, string stringIndexType, RequestContext context) =>
+            ExportProject(waitUntil, projectName, exportedProjectFormat, assetKind, stringIndexType, default, context);
+#pragma warning restore AZC0002 // DO ensure all service methods, both asynchronous and synchronous, take an optional CancellationToken parameter called cancellationToken.
```
kristapratico, line 284: Maybe I'm just missing it, but in your changelog you say that an overload was added to ExportProject to include trainedModelLabel -- I can't find it.

sdk/cognitivelanguage/Azure.AI.Language.Conversations/tests/ConversationAnalysisClientLiveTests.cs
```diff
@@ -307,38 +305,20 @@ public async Task AnalyzeConversation_ConversationPII_TextInput()
...
         {
-            var transciprtConversationItemOne = new TranscriptConversationItem(id: "1", participantId: "speaker")
-            {
-                Itn = "hi",
-                MaskedItn = "hi",
-                Text = "Hi",
-                Lexical = "hi",
-            };
-            transciprtConversationItemOne.AudioTimings.Add(new WordLevelTiming(4500000, 2800000, "hi"));
+            var transciprtConversationItemOne = new TranscriptConversationItem("1", "speaker", "hi", "hi", "Hi", "hi");
```
kristapratico, line 308: nit spelling. 

```suggestion
            var transcriptConversationItemOne = new TranscriptConversationItem("1", "speaker", "hi", "hi", "Hi", "hi");
```

kristapratico (approved):

sdk/cognitivelanguage/Azure.AI.Language.Conversations/src/Authoring/ConversationAuthoringClient.cs
heaths: It's in the generated class. We don't currently handle this case in generated code, but something Michael Nash is looking at going forward for both swaggers and Cadl. Since the generator currently just ignores what was defined before and would otherwise cause a breaking change, I define these methods manually and call into the generated one per our guidelines on client overloads.

PR #31908, opened 2022-10-20
Add abstract base class for Manifest

annelo-msft:
In the current ACR beta API, the `DownloadBlobResult` type has a Manifest property of type `OciManifest`.  This means download will only accommodate OCI artifacts, but the library should be able to support download of any type of artifact.

This PR introduces a base `ArtifactManifest` class and changes the Manifest property on `DownloadBlobResult` to be of this type.  Callers can downcast to the relevant manifest type to use the Manifest value.


sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ArtifactManifest.cs
```diff
@@ -0,0 +1,17 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Azure.Containers.ContainerRegistry.Specialized
+{
+    /// <summary>
+    /// </summary>
+    public abstract class ArtifactManifest
```
ShivangiReja, line 12: Description for this class is missing.

annelo-msft: @ShivangiReja, `ArtifactManifest` now gets its doc comment from the generator, see [ArtifactManifest in APIView]( https://apiview.dev/Assemblies/Review/38848951c47540cbbd20dc2c969cca9e/25d11a3d978b4520bdd04a9860a5e6ad?doc=True&diffOnly=False#Azure.Containers.ContainerRegistry.Specialized.ArtifactManifest)

We can improve these doc comments in the swagger or in swagger transforms as needed.

ShivangiReja (approved):

PR #31910, opened 2022-10-20
[For Discussion] Possible implementation for ACR GetBlobLocation

annelo-msft:
This implementation uses RequestContext to modify the pipeline and turn off redirects.

The first commit shows a naive implementation that just adds a policy, but does not turn off redirects.


annelo-msft:
Closed in favor of https://github.com/Azure/azure-sdk-for-net/pull/32259

PR #31913, opened 2022-10-20
Refactor logic of deserialization of CommunicationIdentifierModel

ostoliarova-ee:
# Description

Refactored logic of deserialization of `CommunicationIdentifierModel` using `CommunicationIdentifierModelKind` and updated changelog files for respective modalities.

[User Story 3013988](https://skype.visualstudio.com/SPOOL/_workitems/edit/3013988): [SDK][Common][AllLangs] Prefer the "kind" property when deserializing the CommunicationIdentifier - follow up

### DIFF
--- a/sdk/communication/Azure.Communication.Chat/CHANGELOG.md
+++ b/sdk/communication/Azure.Communication.Chat/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 1.2.0-beta.1 (Unreleased)
 
 ### Features Added
+- Added support for an optional `kind` property for `CommunicationIdentifierSerializer`.
 
 ### Breaking Changes
 
--- a/sdk/communication/Azure.Communication.Chat/CHANGELOG.md
+++ b/sdk/communication/Azure.Communication.Chat/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 1.2.0-beta.1 (Unreleased)
 
 ### Features Added
-- Added support for an optional `kind` property for `CommunicationIdentifierSerializer`.
+- Optimized the logic for deserializing types derived from the `CommunicationIdentifier`.
 
 ### Breaking Changes
 
--- a/sdk/communication/Shared/src/CommunicationIdentifierSerializer.cs
+++ b/sdk/communication/Shared/src/CommunicationIdentifierSerializer.cs
@@ -59,7 +59,7 @@ namespace Azure.Communication
             }
         }
 
-        public static CommunicationIdentifierModelKind GetKind(CommunicationIdentifierModel identifier)
+        private static CommunicationIdentifierModelKind GetKind(CommunicationIdentifierModel identifier)
         {
             if (identifier.CommunicationUser is not null)
             {
--- a/sdk/communication/Shared/src/CommunicationIdentifierSerializer.cs
+++ b/sdk/communication/Shared/src/CommunicationIdentifierSerializer.cs
@@ -15,56 +15,31 @@ namespace Azure.Communication
 
             AssertMaximumOneNestedModel(identifier);
 
-            if (identifier.Kind is not null)
+            var kind = identifier.Kind ?? GetKind(identifier);
+
+            if (kind == CommunicationIdentifierModelKind.CommunicationUser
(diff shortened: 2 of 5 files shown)


sdk/communication/Azure.Communication.Chat/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## 1.2.0-beta.1 (Unreleased)
 
 ### Features Added
+- Added support for an optional `kind` property for `CommunicationIdentifierSerializer`.
```
petrsvihlik, line 6: `CommunicationIdentifierSerializer` is an `internal` class that's not part of the public API and the customers have no knwledge about it. 
if you want to add a changelog, you may choose to add something like "Optimized the logic for deserializing types derived from the `CommunicationIdentifier`".

petrsvihlik (changes requested):
I see that all the about-to-be-released packages are betas. figuring out the version works similarly to non-preview releases.
- check the latest released version, if there is none (first release), no need to bump anything
- consult your changes with semver
- set the new version of the package ignoring the "alpha/beta" tag (this is to ensure that once the package is out of beta, it'll still follow semver compared to the previously released stable version)

sdk/communication/Shared/src/CommunicationIdentifierSerializer.cs
```diff
@@ -84,6 +59,26 @@ static void AssertMaximumOneNestedModel(CommunicationIdentifierModel identifier)
             }
         }
 
+        public static CommunicationIdentifierModelKind GetKind(CommunicationIdentifierModel identifier)
```
petrsvihlik, line 62:
```suggestion
        private static CommunicationIdentifierModelKind GetKind(CommunicationIdentifierModel identifier)
```

petrsvihlik (changes requested):

petrsvihlik (approved):

AikoBB (approved):

sdk/communication/Shared/src/CommunicationIdentifierSerializer.cs
```diff
@@ -15,56 +15,31 @@ public static CommunicationIdentifier Deserialize(CommunicationIdentifierModel i
...
-                    AssertNotNull(teamsUser.IsAnonymous, nameof(teamsUser.IsAnonymous), nameof(MicrosoftTeamsUserIdentifierModel)),
-                    Deserialize(AssertNotNull(teamsUser.Cloud, nameof(teamsUser.Cloud), nameof(MicrosoftTeamsUserIdentifierModel))),
-                    rawId);
+                      AssertNotNull(user.UserId, nameof(user.UserId), nameof(MicrosoftTeamsUserIdentifierModel)),
+                      AssertNotNull(user.IsAnonymous, nameof(user.IsAnonymous), nameof(MicrosoftTeamsUserIdentifierModel)),
+                      Deserialize(AssertNotNull(user.Cloud, nameof(user.Cloud), nameof(MicrosoftTeamsUserIdentifierModel))),
+                      rawId);
             }
 
             return new UnknownIdentifier(rawId);
```
AikoBB, line 45: for other platforms(java&js) we are checking for the non-nullability of rawId first, we have to unify this logic across all platforms

ostoliarova-msft: @petrsvihlik @AikoBB here RawId is not used for creating `MicrosoftTeamsUserIdentifier` and `CommunicationUserIdentifier` checking it for null before trying to create any of these identifiers will break the existing logic

ostoliarova-msft: @AikoBB  I've re-checked: there is already a validation of the rawId on the top of the method
https://github.com/Azure/azure-sdk-for-net/blob/8ce74c287bc545f1ea518a3364fe174d3d94e431/sdk/communication/Shared/src/CommunicationIdentifierSerializer.cs#L14
            `string rawId = AssertNotNull(identifier.RawId, nameof(identifier.RawId), nameof(CommunicationIdentifierModel));`

PR #27383, opened 2022-03-07
[ACR] Add upload/download code for blobs and manifests

Mohit-Chakraborty:
ApiView diff - https://apiview.dev/Assemblies/Review/2d89f5fb7766481c928e41e65fbd4740?diffOnly=True&diffRevisionId=500d32ab2730498084dedadf61d740fc

We are integrating the changes that were made in `feature/containerregistry/oci-blob-api` and released earlier.
The branch diff shows that all changes have been picked up - https://github.com/Azure/azure-sdk-for-net/compare/mohitc/BlobUpdateToMain...feature/containerregistry/oci-blob-api

PR #26268, opened 2022-01-11
[ACR] Rename `ArtifactManifestProperties.Size`->`SizeInBytes`

Mohit-Chakraborty:

PR #31920, opened 2022-10-20
Use EditorBrowsableState.Never on ACR `GetTagPropertiesCollection` method.

annelo-msft:
We decided to GA the ACR library using the `GetAll<resources>` naming convention for pageable methods (used when the resource name is already plural), instead of the `Get<resources>Collection` convention.  The rationale for this is described in [this PR comment](https://github.com/Azure/azure-sdk-for-net/pull/26268#issuecomment-1010434537).

Unfortunately, the `GetTagPropertiesCollection` method was reintroduced with a bad merge in [this PR](https://github.com/Azure/azure-sdk-for-net/pull/27383).

This PR adds the `[EditorBrowsable(EditorBrowsableState.Never)]` attribute to these methods on the ACR client so they won't appear in Intellisense or docs, to guide callers to use the method consistent with .NET naming conventions.


pallavit:
Is this considered a breaking change, if so, how do we version it? Do we add this in the readme changelog?

annelo-msft:
> Is this considered a breaking change, if so, how do we version it? Do we add this in the readme changelog?

My understanding is that this is *NOT* a breaking change because it's not removing anything, just adding an attribute to an existing method.  I believe it is a strategy we use commonly to hide methods in our APIs that we would in principle want to deprecate, but can't do that because of our no breaking change policy.

@tg-msft can correct/fill in any details I may have missed here?

pallavit:
Do we mention this in the changelog?

annelo-msft:
It doesn't look like storage or textanalytics have done this:
- https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/storage/Azure.Storage.Blobs/CHANGELOG.md
- https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/textanalytics/Azure.AI.TextAnalytics/CHANGELOG.md

My instinct would be that changes to docs only wouldn't need to be called out in the CHANGELOG, unless they were substantial improvements, but if we have evidence of a team precedent or a strong feeling the other way, I'm fine adding it.

ShivangiReja (approved):
LGTM

PR #31926, opened 2022-10-21
[ContainerRegistry] Fix RegistryUri of ContainerRegistryImportSource

### LINKED ISSUE #31007: Wrong type for RegistryUri on ContainerRegistryImportSource
### Library name and version

Azure.ResourceManager.ContainerRegistry 1.0.0

### Describe the bug

The API of the Container Registry expects something like "mcr.microsoft.com" or "docker.io", however the type on the ContainerRegistryImportSource is System.Uri, which always has a scheme like http:// or https://. So when you try to submit a request you always get the following error: 

Invalid value https://docker.io/ for property parameters.Source.RegistryUri for resource name ******. For more information, please visit https://aka.ms/acr/import.
Status: 400 (Bad Request)
ErrorCode: InvalidImportImageParameter

Content:
{"error":{"code":"InvalidImportImageParameter","message":"Invalid value https://docker.io/ for property parameters.Source.RegistryUri for resource name ******. For more information, please visit https://aka.ms/acr/import."},"status":"Failed"}

### Expected behavior

Allow a source registry uri without a schems, e.g. docker.io

### Actual behavior

Only a uri with scheme can be given

### Reproduction Steps

Try to import an image from a public registry into an ACR like in this example: https://docs.microsoft.com/en-us/azure/container-registry/container-registry-import-images?tabs=azure-cli#import-from-docker-hub, but using the .NET SDK.

### Environment

_No response_

fengzhou-msft:
Resolve #31007


sdk/containerregistry/Azure.ResourceManager.ContainerRegistry/src/Custom/Models/ContainerRegistryImportSource.Serialization.cs
```diff
@@ -1,13 +1,12 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-// <auto-generated/>
-
 #nullable disable
 
 using System.Text.Json;
 using Azure.Core;
```
Yao725, line 8: I don't prefer to totally remove this model to the customized code, it might let us miss some new properties if the service team update this model in the future. How about we rename the property to `RegistryAddress` and write some code as shown below in the customized code:
```csharp
[Obsolete("RegistryUri is deprecated, use RegistryAddress instead")]
public Uri RegistryUri {
    get
    {
        if (RegistryAddress != null && Uri.TryCreate(RegistryAddress, UriKind.RelativeOrAbsolute, out var registryUri))
            return registryUri;
        return null;
    }
    set
    {
        RegistryAddress = value == null ? null : value.ToString();
    }
}
```

fengzhou-msft: This seems a good workaround to avoid customizing the `write` method in serialization class (and the generator does not support only customize `write` method, that's why I need to customize the whole class).

fengzhou-msft: Relative Uri is discouraged with https://github.com/Azure/azure-sdk-for-net/blob/main/eng/BannedSymbols.txt, will modify to only use AbsoluteUri.

Yao725 (approved):
LGTM

PR #31927, opened 2022-10-21
Split test pipelines for EventGrid

### LINKED ISSUE #31881: [Event Grid] Split client and management test pipelines
Since the mgmt tests were merged in https://github.com/Azure/azure-sdk-for-net/commit/0ea6efbe000c1d3a3a16719c85ebf83230ae555f, the live test runs have been timing out. We should either split into two separate pipelines and use an adequate timeout for the mgmt pipeline, or disable the mgmt live tests if they cannot be run in a reasonable amount of time.

ArthurMa1978:
Fix https://github.com/Azure/azure-sdk-for-net/issues/31881


sdk/eventgrid/ci.data.yml
```diff
@@ -0,0 +1,43 @@
...
+    include:
+    - main
+    - feature/*
+    - hotfix/*
+    - release/*
+  paths:
+    include:
+    - sdk/eventgrid/ci.data.yml
+    - sdk/eventgrid/Microsoft.Azure.EventGrid/
+    - sdk/eventgrid/Microsoft.Azure.Messaging.EventGrid.CloudNativeCloudEvents/
```
JoshLove-msft, line 28: This one is actually maintained by our team.

sdk/eventgrid/ci.data.yml
```diff
@@ -0,0 +1,43 @@
...
+    - sdk/eventgrid/Microsoft.Azure.WebJobs.Extensions.EventGrid/
```
JoshLove-msft, line 29: Also maintained by our team.

sdk/eventgrid/ci.data.yml
ArthurMa1978: moved to ci.yml

sdk/eventgrid/ci.data.yml
ArthurMa1978: moved to ci.yml

sdk/eventgrid/ci.data.yml
```diff
@@ -0,0 +1,39 @@
...
+    SdkType: data
+    ServiceDirectory: eventgrid
+    ArtifactName: packages
+    Artifacts:
+    - name: Microsoft.Azure.EventGrid
+      safeName: MicrosoftAzureEventGrid
+    - name: Microsoft.Azure.WebJobs.Extensions.EventGrid
+      safeName: MicrosoftAzureWebJobsExtensionsEventGrid
+    - name: Microsoft.Azure.Messaging.EventGrid.CloudNativeCloudEvents
+      safeName: MicrosoftAzureMessagingEventGridCloudNativeCloudEvents
```
JoshLove-msft, line 39: Need to move these as well.

sdk/eventgrid/ci.yml
```diff
@@ -21,21 +23,17 @@ pr:
...
   parameters:
+    SdkType: client
     ServiceDirectory: eventgrid
     ArtifactName: packages
     Artifacts:
     - name: Azure.Messaging.EventGrid
       safeName: AzureMessagingEventGrid
-    - name: Microsoft.Azure.EventGrid
-      safeName: MicrosoftAzureEventGrid
-    - name: Microsoft.Azure.WebJobs.Extensions.EventGrid
```
JoshLove-msft, line 38: Should add these back.

sdk/eventgrid/ci.data.yml
ArthurMa1978: fixed

sdk/eventgrid/ci.yml
ArthurMa1978: fixed

JoshLove-msft:
Seems the build is now failing

JoshLove-msft:
> Seems the build is now failing
> 
> https://dev.azure.com/azure-sdk/public/_build/results?buildId=1940713&view=logs&j=3dc8fd7e-4368-5a92-293e-d53cefc8c4b3&t=e28e7f78-806d-598e-4e2c-9dfd5946e132&l=161

I'm not sure what the issue is here - it is failing on the Build Doc Fx step. /cc @weshaggard any idea?

sdk/eventgrid/service.projects
```diff
@@ -1,5 +1,34 @@
...
+  SDKType values
+   - "all" will build all projects
+   - "client" will build the new track 2 client/data libraries
+   - "mgmtclient" will build the track 2 mgmt libraries
+   - "data" will build the track 1 data plane libraries
+-->
+
 <Project>
-  <ItemGroup>
-    <ProjectReference Remove="$(MSBuildThisFileDirectory)Microsoft.Azure.*\**\*.csproj" Condition="'$(AZURE_EVENTGRID_TEST_MODE)' == 'Live'" />
```
weshaggard, line 3: It looks like we were excluding the track 1 library tests during live mode in the past so I assume we probably want to exclude them still.

sdk/eventgrid/tests.data.yml
```diff
@@ -0,0 +1,11 @@
+trigger: none
```
weshaggard, line 1: You didn't add a tests.mgmt.yml does that mean you don't want live test runs for the mgmt library?

sdk/eventgrid/service.projects
```diff
@@ -1,5 +1,34 @@
...
+  <ItemGroup Condition="'$(SDKType)' == 'client'">
+    <ProjectsToIncludeBySDKType Include="$(MSBuildThisFileDirectory)Azure.Messaging.*\**\*.csproj" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(SDKType)' == 'mgmtclient'">
+    <ProjectsToIncludeBySDKType Include="$(MSBuildThisFileDirectory)Azure.ResourceManager.*\**\*.csproj" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(SDKType)' == 'data'">
+    <ProjectsToIncludeBySDKType Include="$(MSBuildThisFileDirectory)Microsoft.Azure.EventGrid.*\**\*.csproj" />
```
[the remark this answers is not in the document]
weshaggard: @JoshLove-msft I believe this is the reason for the current docs failure. These libraries are listed in the artifact list but we aren't building them in ci.yml because they aren't correctly listed under the client sdktype section.

sdk/eventgrid/ci.data.yml
```diff
@@ -0,0 +1,35 @@
+# NOTE: Please refer to https://aka.ms/azsdk/engsys/ci-yaml before editing this file.
```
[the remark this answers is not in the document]
JoshLove-msft: From my perspective, it isn't necessary. We never had any issues with the T1 data-plane, but am not opposed to doing it.

sdk/eventgrid/tests.data.yml
ArthurMa1978: Yes, I think it isn't necessary to run live test for mgmt. plane sdk

sdk/eventgrid/ci.data.yml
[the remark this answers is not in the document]
ArthurMa1978: I merged Track 1 and Track 2 into one pipeline

sdk/eventgrid/service.projects
ArthurMa1978: This is for the old pipeline which exclude Track 1, now the pipeline is using SDKType to define the scope for testing.

sdk/eventgrid/tests.data.yml
weshaggard: Is that true for all the mgmt plane libraries? Or just this one? Do we ever plan to run live tests for the mgmt libraries?

sdk/eventgrid/ci.yml
```diff
@@ -21,21 +24,26 @@ pr:
...
+    - sdk/eventgrid/ci.yml
+    - sdk/eventgrid/Azure.Messaging.EventGrid/
+    - sdk/eventgrid/Microsoft.Azure.EventGrid/
+    - sdk/eventgrid/Microsoft.Azure.Messaging.EventGrid.CloudNativeCloudEvents/
+    - sdk/eventgrid/Microsoft.Azure.WebJobs.Extensions.EventGrid/
 
 extends:
   template: /eng/pipelines/templates/stages/archetype-sdk-client.yml
   parameters:
+    SdkType: alldata
```
weshaggard, line 36: I'd like to scope this PR to simply splitting the mgmt and data. For now I'd suggest keeping the name client here and mgmtclient in the ci.mgmt.yml. While I agree it isn't 100% correct I'd like to avoid adding new SDKType for this one case.

sdk/eventgrid/service.projects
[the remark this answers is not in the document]
weshaggard: Lets just have client and mgmtclient which I think will solve your issue. I'd like to avoid splitting this more in isolation from other service directories.

sdk/eventgrid/tests.yml
```diff
@@ -3,6 +3,9 @@ trigger: none
 extends:
   template: ../../eng/pipelines/templates/stages/archetype-sdk-tests.yml
   parameters:
+    MaxParallel: 4
```
weshaggard, line 6: Is there a reason for explicitly setting this to 4?

ArthurMa1978: this is copy from servicebus setting, remove this as this might not mandatory

sdk/eventgrid/tests.data.yml
ArthurMa1978: Yes, the mgmt code is auto-gened from swagger, so mostly it's not worth to run regular live testing because actually the tests are testing service not test sdk.
We only got ask from ASH team which want to have live test.

sdk/eventgrid/service.projects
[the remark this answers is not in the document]
ArthurMa1978: fixed

sdk/eventgrid/tests.yml
```diff
@@ -4,5 +4,7 @@ extends:
   template: ../../eng/pipelines/templates/stages/archetype-sdk-tests.yml
   parameters:
     ServiceDirectory: eventgrid
-    EnvVars:
-      AZURE_EVENTGRID_TEST_MODE: Live
+    SDKType: client
+    TimeoutInMinutes: 240
```
weshaggard, line 8: I'm not sure we actually need any of these changes and I'd suggest undoing all the changes except for the SDKType addition for this file. I would prefer the library owner do any specific timeout or cloud configuration changes so they can be validated first.

sdk/eventgrid/tests.data.yml
weshaggard: OK. What is the process for creating and updating the recordings for the tests then?

sdk/eventgrid/tests.yml
ArthurMa1978: Sure, revert except the `EnvVars` which is replaced by SDKType

sdk/eventgrid/tests.data.yml
ArthurMa1978: Required to create / update the recordings before release a stable package

JoshLove-msft (approved):

weshaggard (approved):

sdk/eventgrid/service.projects
JoshLove-msft: Looks like this is causing issues in the live test pipeline - I think we need to add this line back.
https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1946344&view=results

PR #31937, opened 2022-10-21
Disable project reference test in ci

### LINKED ISSUE #31905: Test failed in net - resourcemanager - mgmt - ci due to project reference
In the tests of Azure.ResourceManager.EventHubs, we have a package reference to [Azure.ResourceManager.ManagedServiceIdentitie](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Azure.ResourceManager.EventHubs/tests/Azure.ResourceManager.EventHubs.Tests.csproj#L9). But our pipeline will take turns using project reference and package reference for tests and EventHubs just happens to belong to the group that used project reference.

Currently we have a PR https://github.com/Azure/azure-sdk-for-net/pull/31887 to upgrade the API version for ManagedServiceIdentitie and it will cause the failure of EventHubs tests because of the API version mismatch in the test recording. Considering that we have a GAed Azure.ResourceManager.ManagedServiceIdentitie, we should probably respect the package reference in test project because instead of using project reference to override the settings.

ArthurMa1978:
fix https://github.com/Azure/azure-sdk-for-net/issues/31905

### DIFF
--- a/eng/pipelines/templates/stages/platform-matrix.json
+++ b/eng/pipelines/templates/stages/platform-matrix.json
@@ -39,8 +39,7 @@
       }
     },
     "AdditionalTestArguments": [
-        "/p:UseProjectReferenceToAzureClients=false",
-        "/p:UseProjectReferenceToAzureClients=true"
+        "/p:UseProjectReferenceToAzureClients=false"
     ],
     "BuildConfiguration": [
         "Debug",
--- a/sdk/resourcemanager/ci.mgmt.yml
+++ b/sdk/resourcemanager/ci.mgmt.yml
@@ -1,4 +1,5 @@
 # NOTE: Please refer to https://aka.ms/azsdk/engsys/ci-yaml before editing this file.
+# Base pipeline for all Track 2 Management SDKs
 
 trigger:
   branches:


weshaggard (changes requested):
This impacts all projects in the repo and not just the mgmt pipelines. Why do we need to disable this option? If a project reference doesn't work that is an indication there is a breaking change.

ArthurMa1978:
> This impacts all projects in the repo and not just the mgmt pipelines. Why do we need to disable this option? If a project reference doesn't work that is an indication there is a breaking change.

Yes, I have the same concern as you, but currently we already have APICompatible check why still need this job to test? The problem here is, when upgrade sdk **A** to a new version, then the ci of another sdk **B** will be failed if its test depend on **A** because the version of REST call is different.
I'm not sure how does data plane sdk handle this scenario, but this is a really a blocking issue for mgmt. sdk to release new versions.

weshaggard:
APICompat only helps with public API breaking changes it doesn't help with behavior breaking changes in the implementation. Doing this project reference helps to find those sort of issues in testing as well. 

>REST API version in recorded test data is still old but expect new version.

Can you help me better understand why that causes the test to fail? Are the tests tightly coupled or is there actually a break in the library? If there is a break I suggest we fix those avoid the break, if the tests are tightly coupled we should figure out how to fix the tests to allow this scenario. In either way the "project reference" can be thought of as vNext from the "package reference" and vNext should always still work and if it doesn't there is a break somewhere.

ArthurMa1978:
> APICompat only helps with public API breaking changes it doesn't help with behavior breaking changes in the implementation. Doing this project reference helps to find those sort of issues in testing as well.
> 
> > REST API version in recorded test data is still old but expect new version.
> 
> Can you help me better understand why that causes the test to fail? Are the tests tightly coupled or is there actually a break in the library? If there is a break I suggest we fix those avoid the break, if the tests are tightly coupled we should figure out how to fix the tests to allow this scenario. In either way the "project reference" can be thought of as vNext from the "package reference" and vNext should always still work and if it doesn't there is a break somewhere.

Here is a [sample](https://github.com/Azure/azure-sdk-for-net/runs/8988760558), we updated the MSI to a preview version `2022-01-31-preview`, and the EventHubs test depend on it by package reference which the API version is `2018-11-30`, that fine, but in ci pipeline it's failed on the `project reference` job because currently the MSI API version is deferent to the version in the test record:
```
Unable to find a record for the request PUT https://management.azure.com/subscriptions/326100e2-f69d-4268-8503-075374f62b6e/resourceGroups/testeventhubRG-6572/providers/Microsoft.ManagedIdentity/userAssignedIdentities/identity18127?api-version=2022-01-31-preview
Uri doesn't match:
    request <https://management.azure.com/subscriptions/326100e2-f69d-4268-8503-075374f62b6e/resourceGroups/testeventhubRG-6572/providers/Microsoft.ManagedIdentity/userAssignedIdentities/identity18127?api-version=2022-01-31-preview>
    record  <https://management.azure.com/subscriptions/326100e2-f69d-4268-8503-075374f62b6e/resourceGroups/testeventhubRG-6572/providers/Microsoft.ManagedIdentity/userAssignedIdentities/identity18127?api-version=2018-11-30>
```

This sample is small impact, we had this [pr](https://github.com/Azure/azure-sdk-for-net/pull/31930) to workaround it, but if we upgrade Compute then that will be a massive mess.

The project reference test doesn't introduce this kind of issue for Azure Core, but impact RPs, so how about just disable this for mgmt. sdk?

weshaggard:
@ArthurMa1978 do you know if there is a reason the api-version is included in the test recording name? Is that something we can remove? 

@scbedd is this a scenario that we have seen in other cases where the api-version is changing between package and project references and causing the playback tests to break?

ArthurMa1978:
> @ArthurMa1978 do you know if there is a reason the api-version is included in the test recording name? Is that something we can remove?
> 
> @scbedd is this a scenario that we have seen in other cases where the api-version is changing between package and project references and causing the playback tests to break?

I don't know why, but I think this api-version should be recorded as different version may have different behavior

m-nash:
I agree we don't want to turn off project reference testing its a critical part of our compatibility story prior to publishing.

@ArthurMa1978  There are a couple options I see here.

1. Libraries should not record portions of the setup that are run by dependencies.  This will make the test immune to any change in api version in the set up steps
2. If its required that we record these steps then Libraries should lock the version of the dependency to be a specific version to be resilient to swapping between package and project refs.
 
Even if a library keeps the version latest project reference will point at latest +1 since by definition it has not published yet.  One of the above two options should maintain our compatibility coverage while eliminating the noise here from the recordings.

ArthurMa1978:
Offline synced, we will fix the test instead of change the pipeline

PR #31952, opened 2022-10-21
Define ConversationsClientOptions.Audience for AAD

### LINKED ISSUE #29672: Add `Audience` property to `ConversationsClientOptions`
You'll need to add an `Audience` property to your `ClientOptions` and provide an extensible enum with well known cloud audiences. 

You can use TextAnalytics as a guild. They added the [Audience](https://github.com/Azure/azure-sdk-for-net/blob/7091cb1413ae9639a2f620fbaceacbbd2ba014e2/sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClientOptions.cs#L59) property to `TextAnalyticsClientOptions` and added the extensible enum [TextAnalyticsAudience](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsAudience.cs).

_Originally posted by @schaabs in https://github.com/Azure/azure-sdk-for-net/pull/29665#discussion_r915213824_

heaths:
Resolves #29672

### DIFF
--- a/sdk/cognitivelanguage/Azure.AI.Language.Conversations/src/ConversationsAudience.cs
+++ b/sdk/cognitivelanguage/Azure.AI.Language.Conversations/src/ConversationsAudience.cs
@@ -1,0 +1,59 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.ComponentModel;
+using Azure.Core;
+
+namespace Azure.AI.Language.Conversations
+{
+    /// <summary> Cloud audiences available for Conversations. </summary>
+    public readonly partial struct ConversationsAudience : IEquatable<ConversationsAudience>
+    {
+        private readonly string _value;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ConversationsAudience"/> object.
+        /// </summary>
+        /// <param name="value">The Azure Active Directory audience to use when forming authorization scopes.For the Language service, this value corresponds to a URL that identifies the Azure cloud where the resource is located. For more information: <see href="https://docs.microsoft.com/azure/azure-government/documentation-government-cognitiveservices" />.</param>
+        /// <exception cref="ArgumentNullException"> <paramref name="value"/> is null. </exception>
+        /// <remarks>Please use one of the constant members over creating a custom value unless you have special needs for doing so.</remarks>
+        public ConversationsAudience(string value)
+        {
+            Argument.AssertNotNullOrEmpty(value, nameof(value));
+            _value = value;
+        }
+
+        private const string AzureChinaValue = "https://cognitiveservices.azure.cn";
+        private const string AzureGovernmentValue = "https://cognitiveservices.azure.us";
+        private const string AzurePublicCloudValue = "https://cognitiveservices.azure.com";
+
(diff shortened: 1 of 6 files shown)


jsquire (approved):
LGTM and is consistent with the approach naming and approach used by other Cognitive packages.

sdk/cognitivelanguage/Azure.AI.Language.Conversations/src/ConversationsAudience.cs
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        private const string AzureChinaValue = "https://cognitiveservices.azure.cn";
+        private const string AzureGovernmentValue = "https://cognitiveservices.azure.us";
+        private const string AzurePublicCloudValue = "https://cognitiveservices.azure.com";
+
+        /// <summary> Azure China. </summary>
+        public static ConversationsAudience AzureChina { get; } = new ConversationsAudience(AzureChinaValue);
+
+        /// <summary> Azure Government. </summary>
```
pallavit, line 34: I see some SDKs also mention Germany as a known cloud? Was this miss intentional?

jsquire: What mentions Germany?  That's a new one on me.   The other Cognitive packages and our test matrix don't reference Germany to my knowledge... should we be looking to add it?

pallavit: Not sure actually - I see some packages like ACR using public, gov, china and mooncake(germany) as known ones, so this was me mostly asking if we should be consistent here or we are intentionally different.

jsquire: Mooncake is China.

pallavit: Ahh my bad - blackforest or whatever was the codename. I understand we can still point the audience to that cloud ourselves but was curious if we have different known types in different SDKs?

pallavit: I will grep through the repo and see if we have gaps and check if we should be consistent. It is totally possible we have a legit reason in some scenarios. I wish we had a base class with support for known types in core or something similar

heaths: FWIW, I merely copied the "enum" from other Language SDKs. Given it's an extensible enum, people can also always provide their own endpoints. We could possibly update all of them to be consistent across all our track2s, but what's the status on the more comprehensive configuration features for the SDKs? I haven't heard anything in a while.

pallavit: @heaths agreed 100%. Grep shows we do not have consistency across our SDKs and it is worth talking about whether we should be. The intention is not to block the PR or the release in anyway. I will take a todo on myself for it and if any action item comes from it we can always add later.

Issue #31771, opened 2022-10-12
[FEATURE REQ] ServiceBus and EventHubs processing: use parent instead of link when possible


lmolkova:
### Library name

Azure.Messaging.ServiceBus, Azure.Messaging.EventHubs

### Please describe the feature.

When we trace EH/SB.Process call, we create a link in all cases, even when one message is received. We receive feedback indicating that links are confusing, e.g.
https://github.com/Azure/azure-functions-dotnet-worker/issues/1089 and through support channels.

The suggestion here is that if we process one message/event, we should use the incoming message context as a parent for the new span rather than a link. With this, the producer and consumer will have one trace-id - it's supported by all tracing backends and easy to query.

This is not breaking as all tracing backends should support parent-child relationships.

JoshLove-msft:
Here is the current structure for Service Bus:
```
Receiver.Receive --Link to Message
|
Processor.Process --Link to Message
```
@lmolkova are you saying it should be 
```
Message
|
Receiver.Receive
|
Processor.Process
```

lmolkova:
I think today it's:
Trace0: Message
Trace1: Receiver.Receive --Link to Message
Trace2: Receiver.Process --Link to Message

But should rather be

Trace0: Message
Trace1: Receiver.Receive -- Link to Message
Trace0: Receiver.Process, child of Message


(for single message case)

PR #31957, opened 2022-10-24
Set parent activity instead of using links when processing single message/event

JoshLove-msft:
https://github.com/Azure/azure-sdk-for-net/issues/31771

### DIFF
--- a/sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/EventProcessor.cs
+++ b/sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/EventProcessor.cs
@@ -593,7 +593,8 @@ namespace Azure.Messaging.EventHubs.Primitives
 
             if ((diagnosticScope.IsEnabled) && (eventBatch.Any()))
             {
-                bool isBatch = (EventBatchMaximumCount > 1);
+                var isBatch = (EventBatchMaximumCount > 1);
+                
                 foreach (var eventData in eventBatch)
                 {
                     if (EventDataInstrumentation.TryExtractDiagnosticId(eventData, out string diagnosticId))
@@ -604,7 +605,9 @@ namespace Azure.Messaging.EventHubs.Primitives
                             {
                                 { DiagnosticProperty.EnqueuedTimeAttribute, eventData.EnqueuedTime.ToUnixTimeMilliseconds().ToString(CultureInfo.InvariantCulture) }
                             };
-                            // only use links when we have a batch of events
+
+                            // Use links only when the batch size is not set to a single event.
+                            
                             diagnosticScope.AddLink(diagnosticId, null, attributes);
                         }
                         else
--- a/sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/EventProcessor.cs
+++ b/sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/EventProcessor.cs
@@ -594,7 +594,7 @@ namespace Azure.Messaging.EventHubs.Primitives
             if ((diagnosticScope.IsEnabled) && (eventBatch.Any()))
             {
                 var isBatch = (EventBatchMaximumCount > 1);
-                
+
                 foreach (var eventData in eventBatch)
                 {
                     if (EventDataInstrumentation.TryExtractDiagnosticId(eventData, out string diagnosticId))
@@ -607,7 +607,7 @@ namespace Azure.Messaging.EventHubs.Primitives
                             };
 
                             // Use links only when the batch size is not set to a single event.
-                            
+
                             diagnosticScope.AddLink(diagnosticId, null, attributes);
                         }
                         else
--- a/sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/EventProcessor.cs
+++ b/sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/EventProcessor.cs
@@ -593,16 +593,27 @@ namespace Azure.Messaging.EventHubs.Primitives
 
             if ((diagnosticScope.IsEnabled) && (eventBatch.Any()))
             {
+                bool isBatch = (eventBatch.Count > 1);
                 foreach (var eventData in eventBatch)
                 {
                     if (EventDataInstrumentation.TryExtractDiagnosticId(eventData, out string diagnosticId))
                     {
-                        var attributes = new Dictionary<string, string>(1)
+                        if (isBatch)
                         {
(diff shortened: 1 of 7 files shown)


sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/EventProcessor.cs
```diff
@@ -593,16 +593,27 @@ protected EventProcessor()
...
+                            {
+                                { DiagnosticProperty.EnqueuedTimeAttribute, eventData.EnqueuedTime.ToUnixTimeMilliseconds().ToString(CultureInfo.InvariantCulture) }
+                            };
+                            // only use links when we have a batch of events
+                            diagnosticScope.AddLink(diagnosticId, null, attributes);
+                        }
+                        else
+                        {
+                            diagnosticScope.SetTraceparent(diagnosticId);
+                            diagnosticScope.AddAttribute(
```
JoshLove-msft, line 613: @lmolkova is this the correct handling for attributes in the single event case?

sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/EventProcessor.cs
```diff
@@ -593,16 +593,27 @@ protected EventProcessor()
 
             if ((diagnosticScope.IsEnabled) && (eventBatch.Any()))
             {
+                bool isBatch = (eventBatch.Count > 1);
```
jsquire, line 596: Why declare an interim variable if we're only checking the condition once?

sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/EventProcessor.cs
```diff
@@ -593,16 +593,27 @@ protected EventProcessor()
...
                     {
-                        var attributes = new Dictionary<string, string>(1)
+                        if (isBatch)
                         {
-                            { DiagnosticProperty.EnqueuedTimeAttribute, eventData.EnqueuedTime.ToUnixTimeMilliseconds().ToString(CultureInfo.InvariantCulture) }
-                        };
-
-                        diagnosticScope.AddLink(diagnosticId, null, attributes);
+                            var attributes = new Dictionary<string, string>(1)
+                            {
```
jsquire, line 604: I wonder if this is really the right approach, and I'd like to talk things through and better understand the intent before this moves forward.

Events are should always be read in batches, even when they're being dispatched for processing individually.  Under normal circumstances, if a processor is reading a single event, it's a sign that the prefetch is not tuned properly or there are network issues.

We do not recommend using a batch size of 1, outside of some very unusual edge cases, as it would have a large performance impact.

jsquire (changes requested):

sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/EventProcessor.cs
JoshLove-msft: We are checking the condition in each iteration of the loop.

sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/EventProcessor.cs
JoshLove-msft: One scenario where this will come into play is using the Event Hub trigger for a function that receives a single event.

JoshLove-msft: I guess I see your point if it's possible that some events are delivered in batches and some are not in the lifetime of a single EventProcessor. If that could happen, this would probably cause more confusion than help. @lmolkova - what do you think?

lmolkova: I agree that being consistent is good, i.e. if as a user I sometimes receive batches, it's best to have the same tracing experience when I accidentally receive a single message.

But if I only receive a single message (either with EventHubPoricessor or Function), having parent-child relationship is better.

So, if `MaxBatchSize` is the setting that controls it, I agree we should use it to control the behavior and not the actual batch size.

lmolkova: BTW, in any case there should not be any problem on Azure Monitor with either of these. AzMon supports both cases just fine and Track1 SDKs used parent-child relationships.

jsquire: Josh and I chatted, and we're going to keep the basic structure and just key off of the requested batch size.   That'll keep things consistent per-read.

lmolkova (approved):
LGTM! Thank you @JoshLove-msft

sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/EventProcessor.cs
```diff
@@ -593,16 +593,27 @@ protected EventProcessor()
 
             if ((diagnosticScope.IsEnabled) && (eventBatch.Any()))
             {
+                bool isBatch = (EventBatchMaximumCount > 1);
                 foreach (var eventData in eventBatch)
```
jsquire, line 597:
```suggestion
                var isBatch = (EventBatchMaximumCount > 1);
                
                foreach (var eventData in eventBatch)
```

sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/EventProcessor.cs
```diff
@@ -593,16 +593,27 @@ protected EventProcessor()
...
-                            { DiagnosticProperty.EnqueuedTimeAttribute, eventData.EnqueuedTime.ToUnixTimeMilliseconds().ToString(CultureInfo.InvariantCulture) }
-                        };
-
-                        diagnosticScope.AddLink(diagnosticId, null, attributes);
+                            var attributes = new Dictionary<string, string>(1)
+                            {
+                                { DiagnosticProperty.EnqueuedTimeAttribute, eventData.EnqueuedTime.ToUnixTimeMilliseconds().ToString(CultureInfo.InvariantCulture) }
+                            };
+                            // only use links when we have a batch of events
+                            diagnosticScope.AddLink(diagnosticId, null, attributes);
```
jsquire, line 608:
```suggestion

                            // Use links only when the batch size is not set to a single event.
                            
                            diagnosticScope.AddLink(diagnosticId, null, attributes);
```

jsquire (approved):

PR #31964, opened 2022-10-24
[Event Hubs Extensions] Update EventHubClientFactory

shenchan-msft:
# Support duplicate eventHubName for multiple EventHub triggers/output bindings
For now, the `eventHubName` can not be the same for different EventHub triggers and output bindings as the `EventHubClientFactory ` uses `eventhubName` as the key for consumer and producer clients cache.
To fix this issue, we can switch to use $"{FullyQualifiedNamespace}/{EventHubName}" as the key for producer clients cache.
And $"{FullyQualifiedNamespace}/{EventHubName}/{ConsumerGroup}" as the key for consumer clients cache.

### DIFF
--- a/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/EventHubClientFactory.cs
+++ b/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/EventHubClientFactory.cs
@@ -48,14 +48,14 @@ namespace Microsoft.Azure.WebJobs.EventHubs
             eventHubName = _nameResolver.ResolveWholeString(eventHubName);
             connection = _nameResolver.ResolveWholeString(connection);
 
-            return _producerCache.GetOrAdd(eventHubName, key =>
+            if (!string.IsNullOrWhiteSpace(connection))
             {
-                if (!string.IsNullOrWhiteSpace(connection))
-                {
-                    var info = ResolveConnectionInformation(connection);
+                var info = ResolveConnectionInformation(connection);
 
-                    if (info.FullyQualifiedEndpoint != null &&
-                        info.TokenCredential != null)
+                if (info.FullyQualifiedEndpoint != null &&
+                    info.TokenCredential != null)
+                {
+                    return _producerCache.GetOrAdd(info.FullyQualifiedEndpoint + eventHubName, key =>
                     {
                         return new EventHubProducerClient(
                             info.FullyQualifiedEndpoint,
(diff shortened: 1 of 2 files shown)


shenchan-msft:

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/EventHubClientFactory.cs
```diff
@@ -48,14 +48,14 @@ internal EventHubProducerClient GetEventHubProducerClient(string eventHubName, s
...
-                {
-                    var info = ResolveConnectionInformation(connection);
+                var info = ResolveConnectionInformation(connection);
 
-                    if (info.FullyQualifiedEndpoint != null &&
-                        info.TokenCredential != null)
+                if (info.FullyQualifiedEndpoint != null &&
+                    info.TokenCredential != null)
+                {
+                    return _producerCache.GetOrAdd(info.FullyQualifiedEndpoint + eventHubName, key =>
```
JoshLove-msft, line 58: We should probably do something like:
$"{fullyQualifiedNamespace}/{entityPath}"

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/EventHubClientFactory.cs
```diff
@@ -66,19 +66,24 @@ internal EventHubProducerClient GetEventHubProducerClient(string eventHubName, s
...
-                            ConnectionOptions = _options.ConnectionOptions
-                        });
+                    });
+                }
+                else
+                {
+                    return _producerCache.GetOrAdd(NormalizeConnectionString(info.ConnectionString, eventHubName), key =>
+                    {
+                        return new EventHubProducerClient(
+                            NormalizeConnectionString(info.ConnectionString, eventHubName),
```
JoshLove-msft, line 76: Can we cache the value into local variable instead of computing it twice?

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/EventHubClientFactory.cs
```diff
@@ -48,14 +48,14 @@ internal EventHubProducerClient GetEventHubProducerClient(string eventHubName, s
             eventHubName = _nameResolver.ResolveWholeString(eventHubName);
             connection = _nameResolver.ResolveWholeString(connection);
 
-            return _producerCache.GetOrAdd(eventHubName, key =>
```
JoshLove-msft, line 51: Do we need the same logic for the consumerCache?

JoshLove-msft:
Thanks @shenchan-msft! Can we add some tests here - https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/tests/EventHubsClientFactoryTests.cs?

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/EventHubClientFactory.cs
shenchan-msft: Fixed

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/EventHubClientFactory.cs
shenchan-msft: Fixed

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/EventHubClientFactory.cs
shenchan-msft: Hmm, I did not dig into the `EventHubTriggerAttribute`, but I think a function can only have one eventhub trigger.  so the `_consumerCache` is bug free?

shenchan-msft:
> Thanks @shenchan-msft! Can we add some tests here - https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/tests/EventHubsClientFactoryTests.cs?

Sure, added a new test case to cover this.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/EventHubClientFactory.cs
JoshLove-msft: I think the EventHubClientFactory is a singleton across an instance of a function app, so the same instance would be used for multiple functions. We should probably add a test to confirm this, but I suspect that if you have two functions using different namespaces, but same hub name you would hit the same issue.

shenchan-msft: Fixed with the same logic.

JoshLove-msft (approved):
Thanks @shenchan-msft!

PR #31980, opened 2022-10-24
[Text Analytics] Re-enable and re-record AAD tests, disable statistics tests

### LINKED ISSUE #31854: [Text Analytics] Re-enable AAD tests for version 2022-10-01-preview and newer
The new service API version 2022-10-01-preview does not currently have support for AAD enabled in production. I'm temporarily setting the `ServiceVersion.Max` attribute of the tests that use AAD to the previous version: 2022-05-01. Once AAD is available, this attribute should be removed.

joseharriaga:
Resolves: #31854

Now that the service has enabled AAD for version 2022-10-01-preview, I can re-enable and re-record these tests.

However, I'm also taking the chance to disable the healthcare tests that use statistics (a.k.a. `showStats`), which are failing in some older versions due to a known service regression. More specifically:
- In version 2022-05-01, no stats are returned.
- In version 3.1, the request fails with a 500 Internal Server Error.

### DIFF
--- a/sdk/textanalytics/Azure.AI.TextAnalytics/tests/AnalyzeHealthcareEntitiesTests.cs
+++ b/sdk/textanalytics/Azure.AI.TextAnalytics/tests/AnalyzeHealthcareEntitiesTests.cs
@@ -305,16 +305,17 @@ namespace Azure.AI.TextAnalytics.Tests
         }
 
         [RecordedTest]
-        // TODO: https://github.com/Azure/azure-sdk-for-net/issues/31978.
-        [ServiceVersion(Min = TextAnalyticsClientOptions.ServiceVersion.V2022_10_01_Preview)]
         public async Task AnalyzeHealthcareEntitiesBatchConvenienceWithStatisticsTest()
         {
             TextAnalyticsClient client = GetClient();
 
-            AnalyzeHealthcareEntitiesOptions options = new AnalyzeHealthcareEntitiesOptions()
+            AnalyzeHealthcareEntitiesOptions options = new AnalyzeHealthcareEntitiesOptions();
+
+            // TODO: https://github.com/Azure/azure-sdk-for-net/issues/31978.
+            if (ServiceVersion is not TextAnalyticsClientOptions.ServiceVersion.V3_1)
             {
-                IncludeStatistics = true
-            };
+                options.IncludeStatistics = true;
+            }
 
             AnalyzeHealthcareEntitiesOperation operation = await client.StartAnalyzeHealthcareEntitiesAsync(s_batchConvenienceDocuments, "en", options);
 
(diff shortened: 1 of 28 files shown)


sdk/textanalytics/Azure.AI.TextAnalytics/tests/AnalyzeHealthcareEntitiesTests.cs
```diff
@@ -307,6 +305,8 @@ public async Task AnalyzeHealthcareEntitiesBatchConvenienceTest()
         }
 
         [RecordedTest]
+        // TODO: https://github.com/Azure/azure-sdk-for-net/issues/31978.
+        [ServiceVersion(Min = TextAnalyticsClientOptions.ServiceVersion.V2022_10_01_Preview)]
```
heaths, line 309: I think it's good to track the issue, but you could set `showStats` based on the client versioned passed in. If your base class doesn't already save the `ServiceVersion` under test, you can do so as a protected member and then conditionally set `showStats` based on that. You only need to worry about the latest version with recorded tests; v3.1 will only run under nightly, so you don't have to worry about recordings being wrong.

The fact that you caught this regression proves testing `showStats` is worth covering rather than disabling entirely.

heaths (approved):

heaths (approved):

sdk/textanalytics/Azure.AI.TextAnalytics/tests/AnalyzeHealthcareEntitiesTests.cs
joseharriaga: I made the change. With this, I can 1) keep checking for request-level stats in version 2022-05-01 (only document-level stats are broken in this version), and 2) keep the test enabled for version 3.0 (which was not affected by the regression).

sdk/textanalytics/Azure.AI.TextAnalytics/tests/Infrastructure/TextAnalyticsClientLiveTestBase.cs
```diff
@@ -25,7 +25,7 @@ public class TextAnalyticsClientLiveTestBase : RecordedTestBase<TextAnalyticsTes
         protected TextAnalyticsClientOptions.ServiceVersion ServiceVersion { get; }
 
         public TextAnalyticsClientLiveTestBase(bool isAsync, TextAnalyticsClientOptions.ServiceVersion serviceVersion)
-            : base(isAsync, RecordedTestMode.Live)
+            : base(isAsync)
```
heaths, line 28: I actually recommend making that argument something like:

```c#
..., null /* RecordingMode.Record /* remove "null /*" to force recording /*)
```

Like we do in most files. Makes really easy to just remove `null /*` and you can force recording when necessary. Automatic re-recording only happens for recording mismatches.

joseharriaga: I undid the change for now as it was accidental. Will consider this in the future. Thanks!

Issue #31982, opened 2022-10-24
Re-record all tests on 7.4-preview.1


heaths:
Be sure to update all clients' options to include support for 7.4-preview.1.

PR #31984, opened 2022-10-25
Add KeyVaultSettingsClient

### LINKED ISSUE #31847: [KeyVault] - Add support for new Settings endpoints
In order to support the new Settings endpoints added to Key Vault, the following changes need to be made:

- Create a new client called `KeyVaultSettingsClient` in the Key Vault Administration library.
- Have said client support all 3 operations outlined in the service's swagger [specification](https://github.com/Azure/azure-rest-api-specs/blob/fead0dec636e7554fb8401370418085136d4f052/specification/keyvault/data-plane/Microsoft.KeyVault/preview/7.4-preview.1/settings.json):
  - GetSettingValue
  - GetSettings
  - UpdateSettings

The need for ETag support has not yet been confirmed, but if required it could look like what is being done in the App Configuration library.

heaths:
- [x] Add tests for `KeyVaultSettingsClient`
- [x] Get approval on overloads to set type from @KrzysztofCwalina and @tg-msft; should we still then even make the `KeyVaultSetting` public since it has a string value? /cc @vcolin7 
- [x] Re-record all tests for the Administration package for 7.4-preview.1. Re-recording tests for other packages is tracked in #31982


sdk/keyvault/Azure.Security.KeyVault.Administration/api/Azure.Security.KeyVault.Administration.netstandard2.0.cs
```diff
@@ -281,4 +283,41 @@ public partial class KeyVaultSelectiveKeyRestoreResult
...
+        public virtual System.Threading.Tasks.Task<Azure.Response<Azure.Security.KeyVault.Administration.KeyVaultSetting[]>> GetSettingsAsync(System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
+        public virtual Azure.Response<Azure.Security.KeyVault.Administration.KeyVaultSetting> UpdateSetting(string name, bool value, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
+        public virtual System.Threading.Tasks.Task<Azure.Response<Azure.Security.KeyVault.Administration.KeyVaultSetting>> UpdateSettingAsync(string name, bool value, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
+    }
+    [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
+    public readonly partial struct SettingType : System.IEquatable<Azure.Security.KeyVault.Administration.SettingType>
```
vcolin7, line 307: Should this be named `KeyVaultSettingType` if the public type is `KeyVaultSetting`?

heaths: We have `KeyType` to `KeyVaultKey` so I think `SettingType` makes more sense...if we even keep it.

vcolin7: We actually have `KeyVaultRoleType` in Java, I can't recall the exact reasoning behind the name but I think it's because all other classes it relates to start with `KeyVaultRole`. Maybe it was because RoleType is something that can be found elsewhere for any library that offers RBAC functionality. In any case, I think for Java we will stick to precedent and .NET can probably do the same. I'll talk to Jonathan or Srikanta about this offline.

sdk/keyvault/Azure.Security.KeyVault.Administration/tests/KeyVaultSettingsClientLiveTests.cs
```diff
@@ -0,0 +1,26 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Threading.Tasks;
+using Azure.Core.TestFramework;
+using NUnit.Framework;
+
+namespace Azure.Security.KeyVault.Administration.Tests
+{
+    public class KeyVaultSettingsClientLiveTests : SettingsTestBase
```
christothes, line 10: Do we need to test the other operations as well?

sdk/keyvault/Azure.Security.KeyVault.Administration/src/KeyVaultAdministrationModelFactory.cs
```diff
@@ -120,5 +120,13 @@ public static KeyVaultRoleDefinition KeyVaultRoleDefinition(string id, string na
         /// <returns>A new <see cref="BackupResult"/> instance.</returns>
         public static KeyVaultRestoreResult SelectiveKeyRestoreResult(DateTimeOffset startTime, DateTimeOffset endTime) =>
             new KeyVaultRestoreResult(startTime, endTime);
+
+        /// <summary> Initializes a new instance of KeyVaultSetting. </summary>
+        /// <param name="name"> The account setting to be updated. </param>
+        /// <param name="value"> The value of the pool setting. </param>
+        /// <param name="type"> The type specifier of the value. </param>
+        /// <returns> A new <see cref="Administration.KeyVaultSetting"/> instance for mocking. </returns>
+        public static KeyVaultSetting KeyVaultSetting(string name, string value, SettingType? type = null) =>
```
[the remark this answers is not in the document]
heaths: Given the factory is used for testing, at the very least, a setting type will expand over time. This was meant to provide an "escape" hatch if a different type is needed that isn't supported by swagger. The actual properties are not defined by swagger so, technically, the service could define a new property of a different type at any time. The setting type itself is `modelAsString: true`.

sdk/keyvault/Azure.Security.KeyVault.Administration/tests/KeyVaultSettingsClientLiveTests.cs
heaths: They are in the sample that's also a recorded test. Didn't figure we need to test 2/3<sup>rds</sup> of the operations twice. Do you feel we should for some other reason?

christothes (approved):

sdk/keyvault/Azure.Security.KeyVault.Administration/tests/KeyVaultSettingsClientLiveTests.cs
heaths: Or was it just that it wasn't obvious? I could add a comment pointing to the sample if that would help. KV tests already take long enough, I didn't want to compound it further unnecessarily. Sure, this is quick, but it all adds up.

sdk/keyvault/Azure.Security.KeyVault.Administration/src/KeyVaultAdministrationModelFactory.cs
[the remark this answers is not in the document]
heaths: @KrzysztofCwalina what do you think? Based on our conversation last week about hiding possible future types in the future, do you think the factory should allow for it? I feel it's a nice escape hatch, but also contrary to our conversation. Then again, that's what our factories generally allow in the name of testing.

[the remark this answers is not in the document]
KrzysztofCwalina: I think we can add an overload once service supports other types, and for now just support bool.

sdk/keyvault/Azure.Security.KeyVault.Administration/tests/KeyVaultSettingsClientLiveTests.cs
heaths: I added a comment. Let me know if you want more, though.

PR #31993, opened 2022-10-25
Fixing Repatability headers + Tests + Bugs for HMAC in custom headers

minwoolee-msft:
- Fixing RepeatabilityHeaders: Allow null to be passed (Do not send repeataiblity headers)
- Auto-create repeatabilityHeaders by default
- Updated CustomeHMACHeaders to allow rawId in HMAC authentication (See this: https://github.com/Azure/azure-sdk-for-net/commit/5c0a1b81511258720eda60b3ad0d0b491bc4c181)
- Updated all tests and re-recorded based on changes


sdk/communication/Azure.Communication.CallAutomation/src/CallConnection.cs
```diff
@@ -110,11 +110,10 @@ public virtual async Task<Response> HangUpAsync(HangUpOptions options, Cancellat
                 if (options == null)
                     throw new ArgumentNullException(nameof(options));
 
+                options.RepeatabilityHeaders?.GenerateIfRepeatabilityHeadersNotProvided();
```
yyw-msft, line 113: hang up call is already idempotent because it's a DELETE method so can move this under` if（options.ForEveryone)`. see Fang's design doc ref under `HangupCall`  section https://microsoft.sharepoint-df.com/teams/ServerCalling/_layouts/15/guestaccess.aspx?e=ml5jfd&share=EQ-3bH7vK3pBsos2I0l_4yoBgv7ak3VnUHudVSAFz6BPPQ

sdk/communication/Azure.Communication.CallAutomation/src/CallConnection.cs
```diff
@@ -163,11 +162,10 @@ public virtual Response HangUp(HangUpOptions options,CancellationToken cancellat
...
+
```
yyw-msft, line 166: same comment as above

sdk/communication/Azure.Communication.CallAutomation/src/Models/AddParticipantsOptions.cs
```diff
@@ -18,6 +18,7 @@ public class AddParticipantsOptions
         public AddParticipantsOptions(IEnumerable<CommunicationIdentifier> participantsToAdd)
         {
             ParticipantsToAdd = participantsToAdd;
+            RepeatabilityHeaders = new RepeatabilityHeaders();
```
yyw-msft, line 21: is this needed? If so, don't we need it for other `Options` classes too?

minwoolee-msft: Yes - thank you for pointing out. we need for all options

sdk/communication/Azure.Communication.CallAutomation/src/CallConnection.cs
minwoolee-msft: I did not know this - thank you!

yyw-msft (approved):

PR #32002, opened 2022-10-25
Share hash fix

jaschrep-msft:
Fixes a bug where `ShareFileClient` `Download()` and `DownloadAsync()` when `TransferValidation` is enabled would return the network stream already consumed to validate the checksum, instead of the contents that were buffered on validation. This bug does not affect `OpenRead()`.


seanmcc-msft (approved):
Nit - update changelog

amnguye (approved):

PR #32005, opened 2022-10-25
[Event Hubs] Live test fixes

jsquire:
# Summary 

The focus of these changes is to make tests more deterministic and resilient to the service returning duplicates, which should be expected behavior.

Previously, tests used the count of events read as the condition to signal that reading should cease and validation begin.  When duplicates were present, this caused some expected events to not appear in the final set being validated.  The approach for validating reads from enqueue time were also adjusted to account for clock skew and a starting point that may not be precise, where the service begins reading a bit earlier than expected, which would also cause additional events to be read and throw off validation.

### DIFF
--- a/sdk/eventhub/Azure.Messaging.EventHubs/tests/Consumer/EventHubConsumerClientLiveTests.cs
+++ b/sdk/eventhub/Azure.Messaging.EventHubs/tests/Consumer/EventHubConsumerClientLiveTests.cs
@@ -158,7 +158,7 @@ namespace Azure.Messaging.EventHubs.Tests
 
                     // Read the events and validate the resulting state.
 
-                    var readState = await ReadEventsFromPartitionAsync(consumer, partition, 1, cancellationSource.Token);
+                    var readState = await ReadEventsFromPartitionAsync(consumer, partition, new HashSet<string> { singleEvent.MessageId }, cancellationSource.Token);
 
                     Assert.That(cancellationSource.IsCancellationRequested, Is.False, "The cancellation token should not have been signaled.");
                     Assert.That(readState.Events.Count, Is.EqualTo(1), "A single event was sent.");
@@ -192,7 +192,7 @@ namespace Azure.Messaging.EventHubs.Tests
 
                     // Read the events and validate the resulting state.
 
-                    var readState = await ReadEventsFromPartitionAsync(consumer, partition, 1, cancellationSource.Token);
+                    var readState = await ReadEventsFromPartitionAsync(consumer, partition, new HashSet<string> { singleEvent.MessageId }, cancellationSource.Token);
 
                     Assert.That(cancellationSource.IsCancellationRequested, Is.False, "The cancellation token should not have been signaled.");
                     Assert.That(readState.Events.Count, Is.EqualTo(1), "A single event was sent.");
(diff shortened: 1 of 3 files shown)


sdk/eventhub/Azure.Messaging.EventHubs/tests/Consumer/EventHubConsumerClientLiveTests.cs
```diff
@@ -2585,8 +2600,14 @@ Task<bool> readCallback(ReadState currentState)
         ///
         private class ReadState
         {
-            public readonly ConcurrentDictionary<string, PartitionEvent> Events = new ConcurrentDictionary<string, PartitionEvent>();
+            public readonly HashSet<string> RemainingEvents;
+            public readonly ConcurrentDictionary<string, PartitionEvent> Events = new();
             public long EmptyCount = 0;
+
+            public ReadState(IEnumerable<string> expectedEvents) =>
```
christothes, line 2607: nit: Unless your tests require that this collection be copied, you could check for `is HashSet<string> hash` and assign it directly to `RemainingEvents` (or just create an additional ctor overload).

sdk/eventhub/Azure.Messaging.EventHubs/tests/Primitives/PartitionReceiverLiveTests.cs
```diff
@@ -2246,17 +2278,21 @@ await using (var producer = new EventHubProducerClient(connectionString))
...
 
-                        var eventId = eventData.Properties[EventGenerator.IdPropertyName].ToString();
+                        result.RemainingEvents?.Remove(eventData.MessageId);
+                        result.Events.TryAdd(eventData.MessageId, eventData);
+
+                        // If there's a callback registered per-tick, invoke it and respect its
+                        // decision on whether iteration should continue.
 
-                        if ((result.Events.TryAdd(eventId, eventData)) && (result.Events.Count >= expectedEventCount))
+                        if ((iterationCallback != null) && (!(await iterationCallback(result).ConfigureAwait(false))))
```
christothes, line 2287: nit: Based on its usage, it seems like `iterationCallback` could be renamed to `shouldReadNextEventBatchCallback` and this whole block could be simplified to:
```suggestion
                        shouldReadNextEventBatch = result.RemainingEvents?.Count switch {
                            not null => false,
                            null when shouldReadNextEventBatchCallback is not null => shouldReadNextEventBatchCallback(result).ConfigureAwait(false),
                            _ => true
                        };
                      
```

christothes (approved):

sdk/eventhub/Azure.Messaging.EventHubs/tests/Consumer/EventHubConsumerClientLiveTests.cs
jsquire: The copying is intentional in this case; I need an independent set for each insanitation.   Some of the tests have multiple readers working against the same set of events, where each is keeping track of which they haven't yet received.

sdk/eventhub/Azure.Messaging.EventHubs/tests/Primitives/PartitionReceiverLiveTests.cs
jsquire: I think that changes the logic; reads are done when the count is 0 and the callback should always run if provided, regardless of whether there are events left to read, as there are side-effects that take place in some scenarios (for example, setting a completion source).   I probably could create a compound switch clause to cover things, but I'm going to keep the simple form.

PR #32013, opened 2022-10-25
Hotfix/share consumed stream

seanmcc-msft:


seanmcc-msft:
Continued with https://github.com/Azure/azure-sdk-for-net/pull/32014

PR #32023, opened 2022-10-26
[Text Analytics] Add support for extractive summarization

### LINKED ISSUE #31268: [Text Analytics] Add support for extractive summarization
Revisit previous PR that removed it from a preview before version 2022-05-01.

# Tasks
- [ ] Implementation
- [ ] Tests
- [ ] Docs

joseharriaga:
Resolves: #31268

Features added:
- Added `TextAnalyticsActions.ExtractSummaryActions` to perform extractive summarization in a batch of actions.
- Added `TextAnalyticsClient.StartExtractSummary` and `StartExtractSummaryAsync` to perform extractive summarization on a collection of documents.


sdk/textanalytics/Azure.AI.TextAnalytics/src/ExtractSummaryAction.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// <summary>
+        /// Gets or sets a value that, if set, indicates the version of the text
+        /// analytics model that will be used to generate the result. For supported
+        /// model versions, see operation-specific documentation, for example:
+        /// <see href="https://docs.microsoft.com/azure/cognitive-services/text-analytics/concepts/model-versioning#available-versions"/>.
+        /// </summary>
+        public string ModelVersion { get; set; }
+
+        /// <summary>
+        /// The default value of this property is 'false'. This means, the Language service logs your input text for 48 hours,
```
heaths, line 48: Nit: If you're going to copy/paste or retype the summary, I'd put programmatic elements like `'false'` like so `<c>false</c>` to avoid looking like a string.

sdk/textanalytics/Azure.AI.TextAnalytics/src/ExtractSummaryAction.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// </summary>
+        public bool? DisableServiceLogs { get; set; }
+
+        /// <summary>
+        /// The name of this action. If not provided, the service will generate one.
+        /// </summary>
+        public string ActionName { get; set; }
+
+        /// <summary>
+        /// The maximum number of sentences to be returned in the result. The default value is 3.
```
heaths, line 64: I wouldn't document the default. The service could change it. Just say if it's null it will use the service default.

sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/ServiceClient.cs
```diff
@@ -145,6 +145,15 @@ internal abstract class ServiceClient
 
         #endregion
 
+        #region Extract Summary
+
+        public abstract ExtractSummaryOperation StartExtractSummary(IEnumerable<string> documents, string language = default, ExtractSummaryOptions options = default, CancellationToken cancellationToken = default);
+        public abstract ExtractSummaryOperation StartExtractSummary(IEnumerable<TextDocumentInput> documents, ExtractSummaryOptions options = default, CancellationToken cancellationToken = default);
+        public abstract Task<ExtractSummaryOperation> StartExtractSummaryAsync(IEnumerable<string> documents, string language = default, ExtractSummaryOptions options = default, CancellationToken cancellationToken = default);
+        public abstract Task<ExtractSummaryOperation> StartExtractSummaryAsync(IEnumerable<TextDocumentInput> documents, ExtractSummaryOptions options = default, CancellationToken cancellationToken = default);
```
heaths, line 153: Tip: these are already declared as virtual, so you might as well define them as virtual and throw here instead of in _LegacyServiceClient.cs_. Then just override in _LanguageServiceClient.cs_, which you have to anyway. Makes things easier moving forward.

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
```diff
@@ -2040,6 +2040,138 @@ public virtual async Task<AnalyzeHealthcareEntitiesOperation> StartAnalyzeHealth
...
         #endregion
 
+        #region Extract Summary
+
+        /// <summary>
+        /// Performs extractive summarization on the given documents, which consists of extracting
+        /// sentences that collectively represent the most important or relevant information within
+        /// the original content.
+        /// For a list of languages supported by this operation, see
+        /// <see href="https://learn.microsoft.com/en-us/azure/cognitive-services/language-service/summarization/language-support?tabs=document-summarization"/>.
```
heaths, line 2050: Never include the locale in links because it won't use the users' browser's preference for language and auto-route:
```suggestion
        /// <see href="https://learn.microsoft.com/azure/cognitive-services/language-service/summarization/language-support?tabs=document-summarization"/>.
```
Please make sure to check for any other instances throughout the PR. I won't repeat for brevity.

heaths: I'd also elide the query parameters. These aren't required. Just let the default behavior happen.

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
```diff
@@ -2040,6 +2040,138 @@ public virtual async Task<AnalyzeHealthcareEntitiesOperation> StartAnalyzeHealth
...
+        /// <param name="language">The language that the document is written in.</param>
+        /// <param name="options">The additional configurable options<see cref="ExtractSummaryOptions"/></param>
+        /// <param name="cancellationToken">A <see cref="CancellationToken"/> controlling the request lifetime.</param>
+        /// <returns>
+        /// An <see cref="ExtractSummaryOperation"/> that can be used to monitor the status of this long-running operation.
+        /// Upon completion, the <see cref="ExtractSummaryOperation"/> will contain the collections of extracted summary
+        /// sentences for each document that was successfully analyzed.
+        /// </returns>
+        /// <exception cref="NotSupportedException">This method is only supported in service API version 2022-10-01-preview and newer.</exception>
+        /// <exception cref="RequestFailedException">Service returned a non-success status code.</exception>
```
heaths, line 2067: There are also `ArgumentException`s and derivatives that could throw. Be sure to document those as well.

heaths (approved):

sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/LegacyServiceClient.cs
```diff
@@ -1593,6 +1597,19 @@ private Legacy.JobManifestTasks CreateTasks(TextAnalyticsActions actions)
 
         #endregion
 
+        #region Extract Summary
```
kinelski, line 1600: Not really related to this PR but I see you're adding this region because `LegacyServiceClient` inherits from `ServiceClient`. I think it's a bit awkward to update this class only with methods that are not supported. Have we already included that in a GA release?

kinelski (approved):
LGMT! Left a comment but not blocking for this PR.

mssfang (approved):

sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/ServiceClient.cs
joseharriaga: Done. I had to move the `private readonly TextAnalyticsClientOptions _options` member variable in both `LegacyServiceClient` and `LanguageServiceClient` classes to being a protected property on the parent `ServiceClient` class. I only cleaned up the methods for extractive summarization for now, but I will make this change to the rest of the methods in the `ServiceClient` class in a future PR.

heaths (approved):

PR #32035, opened 2022-10-26
Polish the README and CHANGELOG files (first batch)

ArthurMa1978:


m-nash (approved):
Can we add a follow up work item to enhance our analyzer to be able to detect certain of these mandatory paragraphs always exist?

Also it appears this wasn't universally applied to all 100ish mgmt libraries are there more PRs coming in the future?

Overall the changes look good.

ArthurMa1978:
> Can we add a follow up work item to enhance our analyzer to be able to detect certain of these mandatory paragraphs always exist?
> 
> Also it appears this wasn't universally applied to all 100ish mgmt libraries are there more PRs coming in the future?
> 
> Overall the changes look good.

Yes, the README is not following the guideline, so we need to fix all RPs by using the Acrolinx.
Not sure if there is a ci job can help on verify the document, let me chat with Wes.

PR #32039, opened 2022-10-26
Add Parse/TryParse methods to ResourceIdentifier

### LINKED ISSUE #28988: [QUERY][Azure.Core] Validate resource Id with ResourceIdentifier
### Library name and version

Azure.Core 1.24.0

### Query/Question

What is the correct way to verify a string is valid resource ID?

current we need to 
```csharp
var rid = new ResourceIdentifier(input);
var _ = rid.Name; // Accessing one property verify FormatException
```

Is there a canonical way to do this?

### Environment

_No response_

m-nash:
Fixes https://github.com/Azure/azure-sdk-for-net/issues/28988

### DIFF
--- a/sdk/core/Azure.Core/src/ResourceIdentifier.cs
+++ b/sdk/core/Azure.Core/src/ResourceIdentifier.cs
@@ -113,8 +113,6 @@ namespace Azure.Core
                     _resourceGroupName = resourceName;
                     break;
                 case SpecialType.Subscription:
-                    if (!Guid.TryParse(resourceName, out _))
-                        throw new FormatException($"The GUID for subscription is invalid {resourceName}.");
                     _subscriptionId = resourceName;
                     break;
                 case SpecialType.Provider:
@@ -138,13 +136,13 @@ namespace Azure.Core
             _initialized = true;
         }
 
-        private void Parse()
+        private string? Parse()
         {
             ReadOnlySpan<char> remaining = _stringValue.AsSpan();
 
             if (!remaining.StartsWith(SubscriptionStart.AsSpan(), StringComparison.OrdinalIgnoreCase) &&
                 !remaining.StartsWith(ProviderStart.AsSpan(), StringComparison.OrdinalIgnoreCase))
-                throw new FormatException($"The ResourceIdentifier must start with {SubscriptionStart} or {ProviderStart}.");
+                return $"The ResourceIdentifier must start with {SubscriptionStart} or {ProviderStart}.";
 
             //trim trailing '/' off the end if it exists
             remaining = remaining[remaining.Length - 1] == Separator ? remaining.Slice(1, remaining.Length - 2) : remaining.Slice(1);
@@ -152,23 +150,46 @@ namespace Azure.Core
             ReadOnlySpan<char> nextWord = PopNextWord(ref remaining);
 
             //we know we at least have 1 ResourceIdentifier in the tree here
-            ResourceIdentifierParts nextParts = GetNextParts(Root, ref remaining, ref nextWord);
+            var errorMessage = GetNextParts(Root, ref remaining, ref nextWord, out ResourceIdentifierParts? result);
+            if (errorMessage is not null)
+                return errorMessage;
+            ResourceIdentifierParts nextParts = result!.Value;
+
             while (!nextWord.IsEmpty)
             {
                 //continue to get the next ResourceIdentifier in the tree until we reach the end which will be 'this'
                 ResourceIdentifier newParent = new ResourceIdentifier(nextParts.Parent, nextParts.ResourceType, nextParts.ResourceName, nextParts.IsProviderResource, nextParts.SpecialType);
-                nextParts = GetNextParts(newParent, ref remaining, ref nextWord);
+                if (nextParts.SpecialType == SpecialType.Subscription)
+                {
+                    errorMessage = newParent.CheckSubscriptionFormat();
+                    if (errorMessage is not null)
+                        return errorMessage;
+                }
+                errorMessage = GetNextParts(newParent, ref remaining, ref nextWord, out result);
+                if (errorMessage is not null)
+                    return errorMessage;
+                nextParts = result!.Value;
             }
 
             //initialize ourselves last
             Init(nextParts.Parent, nextParts.ResourceType, nextParts.ResourceName, nextParts.IsProviderResource, nextParts.SpecialType);
+            return nextParts.SpecialType == SpecialType.Subscription ? CheckSubscriptionFormat() : null;
+        }
+
+        private string? CheckSubscriptionFormat()
+        {
+            if (_subscriptionId is not null && !Guid.TryParse(_subscriptionId, out _))
+                return $"The GUID for subscription is invalid {_subscriptionId}.";
+            return null;
         }
 
         private T GetValue<T>(ref T value)
         {
             if (!_initialized)
             {
-                Parse();
+                var error = Parse();
+                if (error is not null)
+                    throw new FormatException(error);
             }
 
             return value;
@@ -194,8 +215,9 @@ namespace Azure.Core
             return parent.ResourceType.AppendChild(resourceTypeName.ToString());
         }
 
-        private static ResourceIdentifierParts GetNextParts(ResourceIdentifier parent, ref ReadOnlySpan<char> remaining, ref ReadOnlySpan<char> nextWord)
+        private static string? GetNextParts(ResourceIdentifier parent, ref ReadOnlySpan<char> remaining, ref ReadOnlySpan<char> nextWord, out ResourceIdentifierParts? parts)
         {
+            parts = null;
             ReadOnlySpan<char> firstWord = nextWord;
             ReadOnlySpan<char> secondWord = PopNextWord(ref remaining);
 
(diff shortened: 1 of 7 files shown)


sdk/core/Azure.Core/src/ResourceIdentifier.cs
```diff
@@ -470,6 +469,45 @@ public override int GetHashCode()
...
+        {
+            try
+            {
+                result = Parse(input);
+                return true;
+            }
+            catch
+            {
+                result = null;
+                return false;
```
JoshLove-msft, line 507: One of the benefits to providing the TryX pattern is that it could improve perf as it isn't expected to throw. It looks like we are still throwing, but just catching the exception here. Is there a way to avoid throwing if Parse fails?

sdk/core/Azure.Core/src/ResourceIdentifier.cs
```diff
@@ -470,6 +469,45 @@ public override int GetHashCode()
...
+        /// </param>
+        /// <returns> True if the parse operation was successful; otherwise, false. </returns>
+        public static bool TryParse(string input, out ResourceIdentifier? result)
+        {
+            try
+            {
+                result = Parse(input);
+                return true;
+            }
+            catch
```
JoshLove-msft, line 504: If there is no way to avoid Parse throwing, we should probably catch the specific exception that it throws.

sdk/core/Azure.Core/src/ResourceIdentifier.cs
```diff
@@ -470,6 +469,45 @@ public override int GetHashCode()
             return left is null ? right is null : left.CompareTo(right) >= 0;
         }
 
+        /// <summary>
+        /// Converts the string representation of a ResourceIdentifier to the equivalent <see cref="ResourceIdentifier"/> structure.
+        /// </summary>
+        /// <param name="input"> The id string to convert. </param>
```
christothes, line 475:
```suggestion
        /// <param name="input"> The id string to parse. </param>
```

sdk/core/Azure.Core/src/ResourceIdentifier.cs
```diff
@@ -470,6 +469,45 @@ public override int GetHashCode()
...
+        /// <returns> A class that contains the value that was parsed. </returns>
```
christothes, line 476:
```suggestion
        /// <returns> The parsed <see cref="ResourceIdentifier" />. </returns>
```

sdk/core/Azure.Core/src/ResourceIdentifier.cs
```diff
@@ -470,6 +469,45 @@ public override int GetHashCode()
...
+            result.Parse();
+            return result;
+        }
+
+        /// <summary>
+        /// Converts the string representation of a ResourceIdentifier to the equivalent <see cref="ResourceIdentifier"/> structure.
+        /// </summary>
+        /// <param name="input"> The id string to convert. </param>
+        /// <param name="result">
+        /// The structure that will contain the parsed value.
```
christothes, line 492: Do we need this line?

sdk/core/Azure.Core/src/ResourceIdentifier.cs
```diff
@@ -470,6 +469,45 @@ public override int GetHashCode()
             return left is null ? right is null : left.CompareTo(right) >= 0;
         }
 
+        /// <summary>
+        /// Converts the string representation of a ResourceIdentifier to the equivalent <see cref="ResourceIdentifier"/> structure.
```
christothes, line 473:
```suggestion
        /// Parses the string representation of an Azure resource identifier to the equivalent <see cref="ResourceIdentifier"/> structure.
```

christothes (approved):

sdk/core/Azure.Core/src/ResourceIdentifier.cs
```diff
@@ -470,6 +469,45 @@ public override int GetHashCode()
...
+        /// The structure that will contain the parsed value.
+        /// If the method returns true result contains a valid ResourceIdentifier.
+        /// If the method returns false, result will be null.
+        /// </param>
+        /// <returns> True if the parse operation was successful; otherwise, false. </returns>
+        public static bool TryParse(string input, out ResourceIdentifier? result)
+        {
+            try
+            {
+                result = Parse(input);
```
KrzysztofCwalina, line 501: We should inplement Parse using TryParse, not the other way.

KrzysztofCwalina (changes requested):

m-nash:
Updated the algorithm to be able to not need a try/catch block.  It did add some time to the happy path, here is an example benchmark from before and after

Before
```
CreateGetName | 364.58 ns | 2.127 ns | 1.885 ns | 0.0129 |     720 B |
```
After
```
CreateGetName |   418.21 ns |  1.707 ns |  1.597 ns | 0.0129 |     720 B |

```

It isn't huge in total magnitude but it is 14% increase.  Let me know if anyone sees what may have caused.

sdk/core/Azure.Core/src/ResourceIdentifier.cs
```diff
@@ -195,25 +215,27 @@ private static ResourceType ChooseResourceType(ReadOnlySpan<char> resourceTypeNa
             return parent.ResourceType.AppendChild(resourceTypeName.ToString());
         }
 
-        private static ResourceIdentifierParts GetNextParts(ResourceIdentifier parent, ref ReadOnlySpan<char> remaining, ref ReadOnlySpan<char> nextWord)
+        private static string? GetNextParts(ResourceIdentifier parent, ref ReadOnlySpan<char> remaining, ref ReadOnlySpan<char> nextWord, out ResourceIdentifierParts? parts)
```
JoshLove-msft, line 218: nit: why not have the error message be the out param and return nullable ResourceIdentifierParts?

JoshLove-msft:
> Updated the algorithm to be able to not need a try/catch block. It did add some time to the happy path, here is an example benchmark from before and after
> 
> Before
> 
> ```
> CreateGetName | 364.58 ns | 2.127 ns | 1.885 ns | 0.0129 |     720 B |
> ```
> 
> After
> 
> ```
> CreateGetName |   418.21 ns |  1.707 ns |  1.597 ns | 0.0129 |     720 B |
> ```
> 
> It isn't huge in total magnitude but it is 14% increase. Let me know if anyone sees what may have caused.

That is really interesting. I guess we could compute the cost of a null check by doing a benchmark of 1 null check vs 2 null checks.

JoshLove-msft:
> > Updated the algorithm to be able to not need a try/catch block. It did add some time to the happy path, here is an example benchmark from before and after
> > Before
> > ```
> > CreateGetName | 364.58 ns | 2.127 ns | 1.885 ns | 0.0129 |     720 B |
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > After
> > ```
> > CreateGetName |   418.21 ns |  1.707 ns |  1.597 ns | 0.0129 |     720 B |
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It isn't huge in total magnitude but it is 14% increase. Let me know if anyone sees what may have caused.
> 
> That is really interesting. I guess we could compute the cost of a null check by doing a benchmark of 1 null check vs 2 null checks.

Here are the numbers I got:

|        Method |     Mean |    Error |   StdDev | Allocated |
|-------------- |---------:|---------:|---------:|----------:|
|  OneNullCheck | 249.6 ns | 10.35 ns | 29.18 ns |         - |
| TwoNullChecks | 257.3 ns |  9.54 ns | 27.06 ns |         - |

So based on this, a rough estimate is that a single null check can add 8ns to the runtime.

christothes:
> Updated the algorithm to be able to not need a try/catch block. It did add some time to the happy path, here is an example benchmark from before and after
> 
> Before
> 
> ```
> CreateGetName | 364.58 ns | 2.127 ns | 1.885 ns | 0.0129 |     720 B |
> ```
> 
> After
> 
> ```
> CreateGetName |   418.21 ns |  1.707 ns |  1.597 ns | 0.0129 |     720 B |
> ```
> 
> It isn't huge in total magnitude but it is 14% increase. Let me know if anyone sees what may have caused.

Should I read into the fact that the numbers are separated in different tables that these were taken from two different runs? If so, I'm not sure if they are as comparable as if run in the same benchmark session (both code paths co-exist and can be run in the same session).

sdk/core/Azure.Core/src/ResourceIdentifier.cs
m-nash: I did this for consistency since all the other methods return string

sdk/core/Azure.Core/src/ResourceIdentifier.cs
m-nash: @christothes for all of these comments I simply copied Guid.Parse, let me know if copying that is the right way to go or make the changes you suggest here.

KrzysztofCwalina (approved):

JoshLove-msft (approved):

sdk/core/Azure.Core/src/ResourceIdentifier.cs
m-nash: ok, I will leave now for consistency we can change later if we feel its necessary

PR #32044, opened 2022-10-26
[Messaging] Clarify batch references to buffers

jsquire:
# Summary

The focus of these changes is to update the documentation for the `EventDataBatch` and `ServiceBusMessageBatch` to clarify that the batch holds a reference to the memory buffers used for event/message bodies, calling out that those buffers need to remain available and unchanged.

# References and Related

- [EventDataBatch.TryAdd does not copy buffer to batch as documentation indicates (#31395)](https://github.com/Azure/azure-sdk-for-net/issues/31395)
- [[Event Hubs] Data body copy for EventDataBatch (#31845)](https://github.com/Azure/azure-sdk-for-net/issues/31845)

### DIFF
--- a/sdk/eventhub/Azure.Messaging.EventHubs/src/Producer/EventDataBatch.cs
+++ b/sdk/eventhub/Azure.Messaging.EventHubs/src/Producer/EventDataBatch.cs
@@ -146,9 +146,13 @@ namespace Azure.Messaging.EventHubs.Producer
         /// <returns><c>true</c> if the event was added; otherwise, <c>false</c>.</returns>
         ///
         /// <remarks>
-        ///   When an event is accepted into the batch, its content and state are frozen; any
-        ///   changes made to the event will not be reflected in the batch nor will any state
-        ///   transitions be reflected to the original instance.
+        ///   When an event is accepted into the batch, changes made to its properties
+        ///   will not be reflected in the batch nor will any state transitions be reflected
+        ///   to the original instance.
+        ///
+        ///   An important consideration to note is that any <see cref="ReadOnlyMemory{T}" /> or byte array
+        ///   passed to the event or a <see cref="BinaryData" /> instance associated with the event
+        ///   is referenced by the batch and must remain valid and unchanged until the batch is disposed.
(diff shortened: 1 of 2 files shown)


jsquire:
//fyi:  @danielmarbach, @Stabzs, @Stabzs

danielmarbach (approved):

JoshLove-msft (approved):

sdk/eventhub/Azure.Messaging.EventHubs/src/Producer/EventDataBatch.cs
```diff
@@ -146,9 +146,13 @@ public sealed class EventDataBatch : IDisposable
...
-        ///   When an event is accepted into the batch, its content and state are frozen; any
-        ///   changes made to the event will not be reflected in the batch nor will any state
-        ///   transitions be reflected to the original instance.
+        ///   When an event is accepted into the batch, changes made to its properties
+        ///   will not be reflected in the batch nor will any state transitions be reflected
+        ///   to the original instance.
+        ///
+        ///   An important consideration to note is that any <see cref="ReadOnlyMemory{T}" /> or byte array
+        ///   passed to the event or a <see cref="BinaryData" /> instance associated with the event
+        ///   is referenced by the batch and must remain valid and unchanged until the batch is disposed.
```
christothes, line 155: totally fine as is, but I wanted to offer an alternative in case you think it may be more clear
```suggestion
        ///   Note:  Any <see cref="ReadOnlyMemory{T}" />, byte array, or
        ///   <see cref="BinaryData" /> instance associated with the event and
        ///   referenced by the batch and must remain valid and unchanged until the batch is disposed.
```

christothes (approved):

sdk/eventhub/Azure.Messaging.EventHubs/src/Producer/EventDataBatch.cs
jsquire: Yeah, that definitely reads better.   Going to apply locally so that I can hit both areas in one commit.

PR #32059, opened 2022-10-27
Update ACR UploadBlob method to upload a blob in chunks

### LINKED ISSUE #32339: Enable support for large blob upload in the BlobClient

annelo-msft:
Fixes https://github.com/Azure/azure-sdk-for-net/issues/32339

The current implementation of [ContainerRegistryBlobClient](https://apiview.dev/Assemblies/Review/7ebfcb36ff6149ba96170bd059094513#Azure.Containers.ContainerRegistry.Specialized.ContainerRegistryBlobClient).UploadBlob() tries to upload an artifact layer in a single chunk, and if the layer size is large, that can cause the service to close the connection during upload, and the upload to fail.

This PR reimplements `UploadBlob()` to upload the layer in chunks, and optionally allows the caller to specify the chunk size.


sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs
```diff
@@ -368,6 +373,98 @@ public virtual async Task<Response<UploadBlobResult>> UploadBlobAsync(Stream str
             }
         }
 
+        private async Task<ChunkedUploadResult> UploadInChunks(string location, Stream stream, int maxChunkSize, bool async = true, CancellationToken cancellationToken = default)
+        {
+            long length = stream.Length;
```
pallavit, line 378: Not every stream is seekable and this method can throw NotSupportedException. Do we want to build an experience around it? Not sure, if there is a specialized seekable stream type and should we consider using it.

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs
```diff
@@ -368,6 +373,98 @@ public virtual async Task<Response<UploadBlobResult>> UploadBlobAsync(Stream str
...
+                }
+                else
+                {
+                    sha256.TransformFinalBlock(buffer, 0, chunkLength);
+                    digest = "sha256:" + OciBlobDescriptor.BytesToString(sha256.Hash);
+                }
+
+                using (Stream chunk = new MemoryStream(buffer, 0, chunkLength))
+                {
+                    uploadChunkResult = async ?
```
pallavit, line 421: >uploadChunkResult [](http://example.com/codeflow?start=20&length=17)

what happens if we fail to upload one of the chunks? Does this rest API call do the retries? 

Also, wouldn't the last uploadChunkResult be used? Is there a chance that the last one may not be the correct representative for all the chunks?

pallavit, line 421: >uploadChunkResult [](http://example.com/codeflow?start=20&length=17)

Given this is costly, should we consider an experience where we do not compute the digest?

This was one question I had previously as well. Shouldn't the service be doing the digest computation? I vaguely remember you mentioned that the service team explicitly asked us to support digest but I am curious do they do it and want us to support it as well or they never do it?

sdk/containerregistry/Azure.Containers.ContainerRegistry/api/Azure.Containers.ContainerRegistry.netstandard2.0.cs
```diff
@@ -244,8 +244,8 @@ public partial class ContainerRegistryBlobClient
...
-        public virtual Azure.Response<Azure.Containers.ContainerRegistry.Specialized.UploadBlobResult> UploadBlob(System.IO.Stream stream, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
-        public virtual System.Threading.Tasks.Task<Azure.Response<Azure.Containers.ContainerRegistry.Specialized.UploadBlobResult>> UploadBlobAsync(System.IO.Stream stream, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
+        public virtual Azure.Response<Azure.Containers.ContainerRegistry.Specialized.UploadBlobResult> UploadBlob(System.IO.Stream stream, Azure.Containers.ContainerRegistry.Specialized.UploadBlobOptions options = null, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
```
pallavit, line 247: I had gotten feedback in Java last time I had shown this to architects that they would like us to support BinaryData variant too - is that something we would like to do for .NET as well?

pallavit:
>     public partial class ContainerRegistryBlobClient

This is also a perfect client candidate for perf testing for the SDK.

---
Refers to: sdk/containerregistry/Azure.Containers.ContainerRegistry/api/Azure.Containers.ContainerRegistry.netstandard2.0.cs:230 in 2cda954. [](commit_id = 2cda95409a38db24b99885c5526ac9720953ea7b, deletion_comment = False)

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs
annelo-msft: Thank you!  Let me look into whether we can get that value differently.

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs
annelo-msft: The pipeline will automatically handle retries according to the [RetryPolicy](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/Pipeline/RetryPolicy.cs).  The caller can configure the RetryPolicy in [RetryOptions](https://apiview.dev/Assemblies/Review/77bdbcef49a3465fa96fac3fb745bf57#Azure.Core.RetryOptions), or starting with Azure.Core 1.26.0, they can implement their own retry policy as well.

>  Is there a chance that the last one may not be the correct representative for all the chunks?

I'm not sure I understand the question ... the API requires that chunks are uploaded [in order](https://docs.docker.com/registry/spec/api/#chunked-upload), and will return a 416 response if an out-of-order chunk is uploaded.  

What do you mean by representative?

annelo-msft: For upload, we're required to send the digest in the final PUT request sent to complete the upload, after all the chunks have been sent.  The registry will also compute the digest of the blob, and will [compare the two](https://docs.docker.com/registry/spec/api/#completed-upload) to confirm that the blob hasn't been tampered with during its transfer across the wire.  So, I don't think we can avoid computing the digest for upload.

sdk/containerregistry/Azure.Containers.ContainerRegistry/api/Azure.Containers.ContainerRegistry.netstandard2.0.cs
annelo-msft: We can certainly add an overload for that, no problem.  I'll need to look into how BinaryData handles streams and whether they load the data fully into memory - if they do, we may want to give greater consideration to this before adding the overload.

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs
```diff
@@ -368,6 +373,98 @@ public virtual async Task<Response<UploadBlobResult>> UploadBlobAsync(Stream str
...
+                digest = OciBlobDescriptor.ComputeDigest(stream);
+
+                uploadChunkResult = async ?
+                    await _blobRestClient.UploadChunkAsync(location, stream, cancellationToken: cancellationToken).ConfigureAwait(false) :
+                    _blobRestClient.UploadChunk(location, stream, cancellationToken: cancellationToken);
+
+                return new ChunkedUploadResult(digest,
+                    uploadChunkResult.Headers.Location,
+                    length);
+            }
```
ShivangiReja, line 394: If user passes really large `maxChunkSize` let's say 1GB and the stream length is little less than that something like 999MB, then this condition will be true and we will endup uploading the whole stream at once for a really large stream. 

We might consider setting up some upper limit for `maxChunkSize` to break the stream in chunks.

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs
```diff
@@ -368,6 +373,98 @@ public virtual async Task<Response<UploadBlobResult>> UploadBlobAsync(Stream str
...
+
+            byte[] buffer = new byte[maxChunkSize];
```
ShivangiReja, line 396: Same issue as above, if user passes some arbitrary large number for `maxChunkSize`, then we will endup creating really large chunks and the user can run out of memory.

So I think we might consider setting up some upper limit for `maxChunkSize` to break the stream in chunks.

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/UploadBlobResult.cs
```diff
@@ -8,14 +8,20 @@ namespace Azure.Containers.ContainerRegistry.Specialized
...
 
         /// <summary>
         /// The digest of the uploaded blob, calculated by the registry.
         /// </summary>
         public string Digest { get; }
+
+        /// <summary>
+        /// The size of the uploaded blob.
+        /// </summary>
+        public long Size { get; }
```
nisha-bhatia, line 25: I am not very familiar with ACR and had a question on chunk sizing. I am a little confused how the size of the uploaded blob could differ from MaxChunkSize passed in the UploadBlobOptions. Are we letting the user know what size chunk they uploaded (but wouldn't they know already)? If MaxChunkSize is not given, the size of the uploaded blob would be 4 Mb. If MaxChunkSize is passed in as 0, I think nothing will be uploaded and there wouldn't be a result?

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/OciBlobDescriptor.cs
```diff
@@ -38,7 +38,7 @@ internal static string ComputeDigest(Stream stream)
                 {
                     stream.Position = 0;
```
JoshLove-msft, line 39: Similar to the comment about accessing stream length, we can't set the stream position if it is not seekable (based on CanSeek property).

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs
```diff
@@ -368,6 +373,98 @@ public virtual async Task<Response<UploadBlobResult>> UploadBlobAsync(Stream str
...
+        /// <returns>A string formatted to suit the Content-Range header.</returns>
+        private static string GetContentRangeString(HttpRange range)
+        {
+            var endRange = "";
+            if (range.Length.HasValue && range.Length != 0)
+            {
+                endRange = (range.Offset + range.Length.Value - 1).ToString(CultureInfo.InvariantCulture);
+            }
+
+            return FormattableString.Invariant($"{range.Offset}-{endRange}");
```
JoshLove-msft, line 445: Should we use the Core [HttpRange](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/HttpRange.cs#L15) type?

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs
```diff
@@ -368,6 +373,98 @@ public virtual async Task<Response<UploadBlobResult>> UploadBlobAsync(Stream str
...
         }
 
+        private async Task<ChunkedUploadResult> UploadInChunks(string location, Stream stream, int maxChunkSize, bool async = true, CancellationToken cancellationToken = default)
+        {
+            long length = stream.Length;
+
+            string digest = default;
+            ResponseWithHeaders<ContainerRegistryBlobUploadChunkHeaders> uploadChunkResult = null;
+
+            if (length < maxChunkSize)
```
JoshLove-msft, line 383: Since we can't compute the length up front, I imagine we would just upload in maxChunkSize chunks until we reach the end of the steram.

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs
```diff
@@ -368,6 +373,98 @@ public virtual async Task<Response<UploadBlobResult>> UploadBlobAsync(Stream str
...
+                    stream.Read(buffer, 0, (int)range.Length);
+
+                if (i < ranges.Count - 1)
+                {
+                    sha256.TransformBlock(buffer, 0, chunkLength, buffer, 0);
+                }
+                else
+                {
+                    sha256.TransformFinalBlock(buffer, 0, chunkLength);
+                    digest = "sha256:" + OciBlobDescriptor.BytesToString(sha256.Hash);
```
JoshLove-msft, line 416: nit: interpolated string

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs
JoshLove-msft: Nevermind, I see the ref docs that explain this doesn't use the standard format for content-range.

JoshLove-msft:
> The current implementation of [ContainerRegistryBlobClient](https://apiview.dev/Assemblies/Review/7ebfcb36ff6149ba96170bd059094513#Azure.Containers.ContainerRegistry.Specialized.ContainerRegistryBlobClient).UploadBlob() tries to upload an artifact layer in a single chunk, and if the layer size is large, that means the client can run out of memory or a request can fail if the transport can't send that large of a stream.

From looking at the previous implementation (before this PR), it doesn't look like we were buffering the stream into a memory stream. Do we know where the Out of Memory was occurring? Was it happening in the Transport? If HttpClient is setting the Transfer-Encoding header to chunked then I don't think we should see out of memory issues. Do we have a test session record for a large upload of a non-seekable stream that we can look at?

I can imagine we would hit the NetworkTimeout of 100s though without chunking ourselves, but this is configurable by users.

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/OciBlobDescriptor.cs
JoshLove-msft: Actually.. looks like we don't even support non-seekable streams - https://github.com/Azure/azure-sdk-for-net/blob/a64584452f7552864cba37a43deaa0a424992677/sdk/core/Azure.Core/src/RequestContent.cs#L129

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs
annelo-msft: Thanks!  Do we have a guideline against these and/or are they bad for perf?  (What do we know?)  What is the recommended action?

annelo-msft:
> From looking at the previous implementation (before this PR), it doesn't look like we were buffering the stream into a memory stream. Do we know where the Out of Memory was occurring? Was it happening in the Transport?

Yeah, great question.  I reached out to the service team asking for a specific repro last week and waiting to get the exact exception.

Can you point me to more information about not buffering the stream?  Is this something we've typically done in other libraries for large files?

nisha-bhatia:
> > From looking at the previous implementation (before this PR), it doesn't look like we were buffering the stream into a memory stream. Do we know where the Out of Memory was occurring? Was it happening in the Transport?
> 
> Yeah, great question. I reached out to the service team asking for a specific repro last week and waiting to get the exact exception.
> 
> Can you point me to more information about not buffering the stream? Is this something we've typically done in other libraries for large files?

Ingestion is going to use an ArrayPool (which is better for large amounts of data): https://learn.microsoft.com/en-us/dotnet/api/system.buffers.arraypool-1?view=net-6.0. Maybe that would work here as well? :)

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs
JoshLove-msft: I don't know if we have specific guidance, but I believe interpolated strings are preferred as the runtime is able to optimize as needed to use stringbuilder. I don't think there is any realistic perf difference - it is more of a style comment.

JoshLove-msft:
> Can you point me to more information about not buffering the stream? Is this something we've typically done in other libraries for large files?

By buffering the stream, I mean reading the whole stream into a memory stream. I could see how this could cause OOM issues. But we weren't doing this in the previous implementation, so I was curious where the memory issue was actually occurring. I would imagine it is probably in the transport layer if the transport ends up attempting to send a single request.

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs
annelo-msft: No, that's great.  Thank you!

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs
```diff
@@ -368,6 +372,66 @@ public virtual async Task<Response<UploadBlobResult>> UploadBlobAsync(Stream str
             }
         }
 
+        private async Task<ChunkedUploadResult> UploadInChunks(string location, Stream stream, int chunkSize, bool async = true, CancellationToken cancellationToken = default)
+        {
+            // TODO: don't allocate a large buffer unless we need it.  How will we know?
+            byte[] buffer = new byte[chunkSize];
+            using SHA256 sha256 = SHA256.Create();
```
ShivangiReja, line 379: I am just curious that what would happen if user passes chunkSize of 0? Or a really large number (100MB, 1GB) leading to OOM issues? Do we need to have upper and lower bounds for this value?

In case you missed it, see my previous comments here:
* https://github.com/Azure/azure-sdk-for-net/pull/32059#discussion_r1013369562
* https://github.com/Azure/azure-sdk-for-net/pull/32059#discussion_r1013368925

annelo-msft: Hi @ShivangiReja - sorry for my delay in responding! I was out sick for a few days, and I'm still a little slow catching up.

Yes, you are absolutely right.  If the caller passes [UploadBlobOptions](https://apiview.dev/Assemblies/Review/ce4598afc1f94854b136ea2183e05af6#Azure.Containers.ContainerRegistry.Specialized.UploadBlobOptions) with MaxChunkSize of 0, a zero-length byte array would be allocated, and the call would fail because all the stream.Read calls wouldn't have anywhere to copy the data from the stream to, and my while loop would be infinite.  Yikes!  I was planning to add argument validation to the UploadBlob method to prevent this case.

I wanted to do a little more research to see what the precedent was in other of our libraries like storage to understand whether we place limitation on the max size a caller can specify.  I'd personally be cautious about hard-coding something that makes an assumption about the memory limits on a user's system and might not age well.  An argument could be made that it's the user's responsibility to provide a reasonable limit if they want to override the default, and if they ask for more memory than they have, they have written a bug.  But if our team has discussed this and made an informed decision, I'm happy to follow precedent, I just need to figure out what that is.

Does that address your concern?

ShivangiReja: >An argument could be made that it's the user's responsibility to provide a reasonable limit if they want to override the default, and if they ask for more memory than they have, they have written a bug

Why are we asking users to pass the maxChunkSize? Can we not set it internally? 

I see storage is no giving users any option to pass the max size. See Upload API [here](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/storage/Azure.Storage.Blobs/src/BlockBlobClient.cs#L583) and [BlobUploadOptions](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/storage/Azure.Storage.Blobs/src/Models/BlobUploadOptions.cs).

annelo-msft: Huh!  Honestly?  I had seen it in the PartitionedUploader API and thought it was exposed internally.  Good call - I can fix this value in the client.  Thanks!

JoshLove-msft: Not sure if we are talking about the same thing, but this is in the [StorageTransferOptions](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/storage/Azure.Storage.Common/src/StorageTransferOptions.cs#L29).

annelo-msft: Thanks @JoshLove-msft, I did see it in a public API then.

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs
```diff
@@ -368,6 +372,66 @@ public virtual async Task<Response<UploadBlobResult>> UploadBlobAsync(Stream str
...
+                }
+
+                blobLength += bytesRead;
+                chunkCount++;
+
+                // Read next chunk into buffer
+                bytesRead = async ?
+                    await stream.ReadAsync(buffer, 0, chunkSize, cancellationToken).ConfigureAwait(false) :
+                    stream.Read(buffer, 0, chunkSize);
+            }
```
ShivangiReja, line 413: Have we considered wrapping stream into [BufferedStream](https://learn.microsoft.com/en-us/dotnet/api/system.io.bufferedstream?view=net-7.0)? I was just reading these [docs](https://learn.microsoft.com/en-us/dotnet/api/system.io.bufferedstream?view=net-7.0#remarks) and seems like we can optimize read performance using BufferedStream.

annelo-msft: Can you share its advantages over MemoryStream?

annelo-msft: Thanks, Shivangi!  I'll have a quick look, and consider this as a possible improvement for a later PR.  If you're able to show how I could make concrete performance improvements with this approach, that would help accelerate this work.  Thanks!

annelo-msft:
> This is also a perfect client candidate for perf testing for the SDK.

@pallavit, tracking with https://github.com/Azure/azure-sdk-for-net/issues/32415

sdk/containerregistry/Azure.Containers.ContainerRegistry/api/Azure.Containers.ContainerRegistry.netstandard2.0.cs
annelo-msft: Tracking with https://github.com/Azure/azure-sdk-for-net/issues/32416

annelo-msft:
> By buffering the stream, I mean reading the whole stream into a memory stream. I could see how this could cause OOM issues. But we weren't doing this in the previous implementation, so I was curious where the memory issue was actually occurring. I would imagine it is probably in the transport layer if the transport ends up attempting to send a single request.

I still haven't heard back from the service team on this, but looking at the information they've given us, it looks like the OOM issue they were experience was on download of a large blob.  The only thing I've been able to repro on upload is the service closing the connection, presumably because we're sending too much data/data too slowly:

----> System.Net.Http.HttpRequestException : Error while copying content to a stream.
----> System.IO.IOException : Unable to write data to the transport connection: An existing connection was forcibly closed by the remote host..
----> System.Net.Sockets.SocketException : An existing connection was forcibly closed by the remote host.

Uploading in smaller chunks does mitigate this.

I'll update the PR description to more accurately reflect the issue as well - thanks, @JoshLove-msft!

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs
annelo-msft: Discussed here: https://github.com/Azure/azure-sdk-for-net/pull/32059#discussion_r1018332277

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs
annelo-msft: Discussed here: https://github.com/Azure/azure-sdk-for-net/pull/32059#discussion_r1018332277

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs
ShivangiReja: I did a small perf test between these 2 approaches but it gave me almost the same result.

Later I found [this](https://learn.microsoft.com/en-us/archive/blogs/brada/filestream-and-bufferedstream) doc:
>No, there is zero benefit from wrapping a BufferedStream around a FileStream. We copied BufferedStream’s buffering logic into FileStream about 4 years ago to encourage better default performance ([Pit of Success](https://learn.microsoft.com/en-us/archive/blogs/brada/the-pit-of-success)). **In fact, I don’t think there are any Streams in the .NET Framework that require it, but it might be needed by custom Stream implementations if they do not do buffering by default.**

Seems like all the streams in .NET framework do buffer by default. So in this case we don't need to wrap stream into BufferedStream.

annelo-msft: Thanks so much for the investigation, @ShivangiReja!

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/UploadBlobResult.cs
JoshLove-msft: Also curious about this - it doesn't look like this is returned from the service so we are essentially just returning the stream.Length to the user. Not sure if this is valuable information to include here.

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs
```diff
@@ -267,23 +230,8 @@ public virtual async Task<Response<UploadManifestResult>> UploadManifestAsync(St
...
-                    ValidateDigest(stream, response.Headers.DockerContentDigest);
-
-                if (!valid)
-                {
-                    throw _clientDiagnostics.CreateRequestFailedException(response,
-                        new ResponseError(null, "The digest in the response does not match the digest of the uploaded manifest."));
-                }
-
-                return Response.FromValue(new UploadManifestResult(response.Headers.DockerContentDigest), response.GetRawResponse());
+                using Stream stream = await CopyStreamAsync(manifestStream, true).ConfigureAwait(false);
```
JoshLove-msft, line 233: So we do buffer the stream into memory for manifests, but are these usually smaller than blobs?

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs
```diff
@@ -282,6 +240,51 @@ public virtual async Task<Response<UploadManifestResult>> UploadManifestAsync(St
             }
         }
 
+        private async Task<Response<UploadManifestResult>> UploadManifestInternalAsync(Stream stream, UploadManifestOptions options, bool async, CancellationToken cancellationToken)
+        {
+            Debug.Assert(stream is MemoryStream, "Should only be called on internally allocated, seekable streams.");
```
JoshLove-msft, line 245: Why not just make this method take a MemoryStream?

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs
```diff
@@ -282,6 +240,51 @@ public virtual async Task<Response<UploadManifestResult>> UploadManifestAsync(St
...
+
+            ResponseWithHeaders<ContainerRegistryCreateManifestHeaders> response = async ?
+                // TODO: media type should be configurable to support non-OCI types.
+                await _restClient.CreateManifestAsync(_repositoryName, tagOrDigest, stream, ManifestMediaType.OciManifest.ToString(), cancellationToken).ConfigureAwait(false) :
+                _restClient.CreateManifest(_repositoryName, tagOrDigest, stream, ManifestMediaType.OciManifest.ToString(), cancellationToken);
+
+            if (!ValidateDigest(digest, response.Headers.DockerContentDigest))
+            {
+                throw _clientDiagnostics.CreateRequestFailedException(response,
+                    new ResponseError(null, "The digest in the response does not match the digest of the uploaded manifest."));
```
JoshLove-msft, line 258: Isn't ResponseError meant for service returned errors? Why not just pass the message?

JoshLove-msft (approved):

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs
annelo-msft: Yes, I'm doing that so we don't dispose the caller's passed-in stream.

annelo-msft:
> Ingestion is going to use an ArrayPool (which is better for large amounts of data): https://learn.microsoft.com/en-us/dotnet/api/system.buffers.arraypool-1?view=net-6.0. Maybe that would work here as well? :)

@nisha-bhatia, I plan to re-work the implementation to possibly use ArrayPool when I implement chunked/parallel download.  Thanks again for the suggestion!

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/OciBlobDescriptor.cs
```diff
@@ -28,6 +29,8 @@ public partial class OciBlobDescriptor
 
         internal static string ComputeDigest(Stream stream)
         {
+            Debug.Assert(stream.CanSeek, "Should only be called on seekable streams.");
```
JoshLove-msft, line 32: Similar comment here - could we just take a MemoryStream instead?

annelo-msft: For this one, I looked at that and decided to handle it differently for this case, but I changed it to test CanSeek instead of the stream type.  Although I reworked that again this morning - let me check whether your suggestion works now in light of that refactoring ...

annelo-msft: I still have a dependency on this being Stream from the Download case ... would you be ok with me refactoring this when I address that?  I'll add a note to the issue, I'm tracking a few things there: https://github.com/Azure/azure-sdk-for-net/issues/32414

JoshLove-msft: Sure!

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/UploadBlobResult.cs
annelo-msft: > so we are essentially just returning the stream.Length to the user

Yes, that's fair.  My thinking was that this is useful to have during an image push operation, to author the image manifest.  It's true that the caller might already have the size of their stream.  If for some reason they're passing through a non-seekable stream, they might not be able to call Length on it?

JoshLove-msft: I don't think we support non-seekable streams 😄

annelo-msft: >  I am a little confused how the size of the uploaded blob could differ from MaxChunkSize passed in the UploadBlobOptions. 

The chunk size represents fragments of the blob and MaxChunkSize lets callers say how big those fragments should be so they can tune the upload operation, e.g. based on characteristics of their system.  The size of the uploaded blob represents the sum of all the chunk sizes for the entire upload operation.

PR #32062, opened 2022-10-27
Generate toc for dotnet

sima-zhu:
This is the PR to generate toc for dotnet pipeline.

Here is the testing pipeline: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1977544&view=results
Here is the toc looks like: https://github.com/Azure/azure-docs-sdk-dotnet/blob/daily/2022-11-09/docs-ref-toc/reference-unified.yml

### DIFF
--- a/eng/pipelines/docindex.yml
+++ b/eng/pipelines/docindex.yml
@@ -5,7 +5,8 @@ jobs:
 
   - job: UpdateDocsMsBuildConfig
     pool:
-      vmImage: ubuntu-20.04
+      name: azsdk-pool-mms-ubuntu-2004-general
+      vmImage: MMSUbuntu20.04Compliant
     variables:
       DocRepoLocation: $(Pipeline.Workspace)/docs
       DocRepoOwner: Azure
@@ -22,19 +23,34 @@ jobs:
           Paths:
             - bundlepackages/
             - metadata/
+            - docs-ref-toc/
           Repositories:
             - Name: $(DocRepoOwner)/$(DocRepoName)
               WorkingDirectory: $(DocRepoLocation)
-
+      # Install the tool for toc step to use
+      - task: NuGetToolInstaller@1
+        inputs:
+          versionSpec: 6.3.1
       # Call update docs ci script to onboard packages
+      - template: /eng/common/pipelines/templates/steps/set-default-branch.yml
       - task: Powershell@2
         inputs:
           pwsh: true
           filePath: eng/common/scripts/Update-DocsMsPackages.ps1
           arguments: -DocRepoLocation $(DocRepoLocation)
         displayName: Update Docs Onboarding
+        condition: and(succeeded(), or(eq(variables['Build.Reason'], 'Schedule'), eq(variables['Force.MainUpdate'], 'true')))
 
-      - template: /eng/common/pipelines/templates/steps/set-default-branch.yml
+      - task: Powershell@2
+        inputs:
+          pwsh: true
+          filePath: eng/common/scripts/Update-DocsMsToc.ps1
+          arguments: >-
+            -DocRepoLocation $(DocRepoLocation)
+            -OutputLocation $(DocRepoLocation)/docs-ref-toc/reference-unified.yml
+            -ReadmeFolderRoot "api/overview/azure"
+        displayName: Generate ToC for main branch
+        condition: and(succeeded(), or(eq(variables['Build.Reason'], 'Schedule'), eq(variables['Force.MainUpdate'], 'true')))
 
       - template: /eng/common/pipelines/templates/steps/git-push-changes.yml
         parameters:
@@ -56,6 +72,13 @@ jobs:
           git status
         displayName: Checkout daily branch if it exists
         workingDirectory: $(DocRepoLocation)
+
+      - pwsh: |
+          $publicFeedUrl = "https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-for-net/nuget/v3/index.json"
+          Write-Host "##vso[task.setvariable variable=PackageSourceOverride]$publicFeedUrl"
+        displayName: Set package source variable
+        workingDirectory: $(DocRepoLocation)
+        
       - task: Powershell@2
         inputs:
           pwsh: true
@@ -64,8 +87,19 @@ jobs:
           # updated packages.
           arguments: >
             -DocRepoLocation $(DocRepoLocation)
-            -PackageSourceOverride 'https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-for-net/nuget/v3/index.json'
+            -PackageSourceOverride $(PackageSourceOverride)
         displayName: Update Docs Onboarding for Daily branch
+      - task: Powershell@2
+        inputs:
+          pwsh: true
+          filePath: eng/common/scripts/Update-DocsMsToc.ps1
+          arguments: >-
+            -DocRepoLocation $(DocRepoLocation)
+            -OutputLocation $(DocRepoLocation)/docs-ref-toc/reference-unified.yml
+            -ReadmeFolderRoot "api/overview/azure"
+            -PackageSourceOverride $(PackageSourceOverride)
+        displayName: Generate ToC for Daily Branches
+
       - template: /eng/common/pipelines/templates/steps/git-push-changes.yml
         parameters:
           BaseRepoBranch: $(DailyDocsBranchName)
@@ -75,6 +109,7 @@ jobs:
           TargetRepoOwner: $(DocRepoOwner)
           WorkingDirectory: $(DocRepoLocation)
           ScriptDirectory: $(Build.SourcesDirectory)/eng/common/scripts
+          PushArgs: -f
 
       - task: PowerShell@2
         displayName: Queue Docs CI build
--- a/eng/scripts/Language-Settings.ps1
+++ b/eng/scripts/Language-Settings.ps1
@@ -405,11 +405,17 @@ function EnsureCustomSource($package) {
   }
 
   Write-Host "Checking custom package source for $($package.Name)"
-  $existingVersions = Find-Package `
-    -Name $package.Name `
-    -Source CustomPackageSource `
-    -AllVersions `
-    -AllowPrereleaseVersions
+  try {
+    $existingVersions = Find-Package `
+      -Name $package.Name `
+      -Source CustomPackageSource `
+      -AllVersions `
+      -AllowPrereleaseVersions
+  }
+  catch {
+    Write-Error $_ -ErrorAction Continue
+    return $package
+  }
   
   # Matches package version against output: 
   # "Azure.Security.KeyVault.Secrets 4.3.0-alpha.20210915.3"
--- a/eng/scripts/Language-Settings.ps1
+++ b/eng/scripts/Language-Settings.ps1
@@ -38,7 +38,7 @@ function Get-AllPackageInfoFromRepo($serviceDirectory)
       $defaultDll = Get-ChildItem "$dllFolder/Release/netstandard2.0/*" -Filter "$pkgName.dll" -Recurse
       if ($defaultDll -and (Test-Path $defaultDll)) {
         Write-Verbose "Here is the dll file path: $($defaultDll.FullName)"
-        $namespaces = @(Get-NamepspacesFromDll $defaultDll.FullName)
+        $namespaces = @(Get-NamespacesFromDll $defaultDll.FullName)
       }
     }
     $pkgProp = [PackageProps]::new($pkgName, $pkgVersion, $pkgPath, $serviceDirectory)
--- a/eng/scripts/docs/Docs-ToC.ps1
+++ b/eng/scripts/docs/Docs-ToC.ps1
@@ -28,7 +28,7 @@ function DownloadNugetPackage($package, $version, $destination) {
     # If it is empty then the property is not overridden.
     $customPackageSource = Get-Variable -Name 'PackageSourceOverride' -ValueOnly -ErrorAction 'Ignore'
     # Download package from nuget, if no package found in nuget, then search devops public feeds if set.
-    Write-Host "Downloading $package with version $version..."
+    Write-Host "Downloading source: '$customPackageSource'. If it is empty, we default to use nuget gallery."
     if ($customPackageSource) {
         nuget install -FallbackSource $customPackageSource $package -Version $version -DirectDownload -DependencyVersion Ignore -OutputDirectory $destination
     }
--- a/eng/scripts/docs/Docs-ToC.ps1
+++ b/eng/scripts/docs/Docs-ToC.ps1
@@ -28,7 +28,7 @@ function DownloadNugetPackage($package, $version, $destination) {
     # If it is empty then the property is not overridden.
     $customPackageSource = Get-Variable -Name 'PackageSourceOverride' -ValueOnly -ErrorAction 'Ignore'
     # Download package from nuget, if no package found in nuget, then search devops public feeds if set.
-    Write-Host "Downloading $package with version $version..."
+    LogDebug "Downloading $package with version $version..."
     if ($customPackageSource) {
         nuget install -FallbackSource $customPackageSource $package -Version $version -DirectDownload -DependencyVersion Ignore -OutputDirectory $destination
     }
@@ -42,41 +42,42 @@ function Fetch-NamespacesFromNupkg ($package, $version) {
     if (!(Test-Path $tempLocation)) {
         New-Item -ItemType Directory -Path $tempLocation -Force | Out-Null
     }
-    Write-Host "Downloading nupkg package $package with version $version to $tempLocation ...."
+    LogDebug "Downloading nupkg package $package with version $version to $tempLocation ...."
     DownloadNugetPackage -package $package -version $version -destination $tempLocation | Out-Null
     $packageFolder = "$package.$version"
     if ($LASTEXITCODE -ne 0 -or !$(Test-Path $tempLocation/$packageFolder)) {
-        Write-Host "Did not download the package correctly. Skipping..."
+        LogDebug "Did not download the package correctly. Skipping..."
         return @()
     }
-    Write-Host "Searching for dll file..."
+    LogDebug "Searching for dll file..."
     $dllFiles = @()
     $dllFileName = ""
     if (Test-Path "$tempLocation/$packageFolder/lib/netstandard2.0/"){
         $dllFiles = @(Get-ChildItem "$tempLocation/$packageFolder/lib/netstandard2.0/*" -Filter '*.dll' -Recurse)
     }
     if (!$dllFiles) {
-        Write-Warning "Can't find any dll file from $tempLocation/$packageFolder with target netstandard2.0."
+        LogWarning "Can't find any dll file from $tempLocation/$packageFolder with target netstandard2.0."
         $dllFiles = Get-ChildItem "$tempLocation/$packageFolder/lib/*" -Filter '*.dll' -Recurse
         if (!$dllFiles) {
-            Write-Error "Can't find any dll file from $tempLocation/$packageFolder."
+            LogError "Can't find any dll file from $tempLocation/$packageFolder."
             return @()
         }
     }
     elseif ($dllFiles.Count -gt 1) {
-        Write-Warning "There are multiple dll files in target netstandard2.0. "
+        LogWarning "There are multiple dll files in target netstandard2.0. "
         if (Test-Path "$tempLocation/$packageFolder/lib/netstandard2.0/$package.dll") {
-            Write-Host "Use the dll file $package.dll"
+            LogDebug "Use the dll file $package.dll"
             $dllFileName = "$tempLocation/$packageFolder/lib/netstandard2.0/$package.dll"
         }
     }
     if (!$dllFileName) {
+        $dllFiles = $dllFiles | Get-Unique | Sort-Object
         $dllFileName = $dllFiles[0].FullName
     }
-    Write-Host "Dll file found: $dllFileName"
+    LogDebug "Dll file found: $dllFileName"
     $namespaces = Get-NamespacesFromDll $dllFileName
     if (!$namespaces) {
-        Write-Error "Can't find namespaces from dll file $dllFileName." -ErrorAction Continue
+        LogError "Can't find namespaces from dll file $dllFileName." -ErrorAction Continue
         return @()
     }
     return $namespaces
@@ -94,7 +95,7 @@ function Get-dotnet-OnboardedDocsMsPackages($DocRepoLocation) {
         foreach ($spec in $onboardingSpec) {
             $packageInfo = $spec -split ","
             if (!$packageInfo -or ($packageInfo.Count -lt 2)) {
-                Write-Error "Please check the package info in csv file $file. Please have at least one package and follow the format {name index, package name, version(optional)}" -ErrorAction Continue
+                LogError "Please check the package info in csv file $file. Please have at least one package and follow the format {name index, package name, version(optional)}" -ErrorAction Continue
                 return $null
             }
             $packageName = $packageInfo[1].Trim() -replace "\[.*\](.*)", '$1'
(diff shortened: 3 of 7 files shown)


eng/scripts/Language-Settings.ps1
```diff
@@ -405,11 +405,17 @@ function EnsureCustomSource($package) {
   }
 
   Write-Host "Checking custom package source for $($package.Name)"
-  $existingVersions = Find-Package `
-    -Name $package.Name `
-    -Source CustomPackageSource `
-    -AllVersions `
-    -AllowPrereleaseVersions
+  try {
```
sima-zhu, line 408: Some of the management track 1 packages do not exist in public feeds anymore. 
image

eng/scripts/docs/Docs-ToC.ps1
```diff
@@ -21,3 +21,200 @@ function Get-NamepspacesFromDll($dllPath) {
     }
     return $namespaces | Sort-Object -Unique
 }
+
+function DownloadNugetPackage($package, $version, $destination) {
+    # $PackageSourceOverride is a global variable provided in
+    # Update-DocsMsPackages.ps1. Its value can set a "customSource" property.
+    # If it is empty then the property is not overridden.
+    $customPackageSource = Get-Variable -Name 'PackageSourceOverride' -ValueOnly -ErrorAction 'Ignore'
```
sima-zhu, line 29: We setup `PackageSourceOverride` variable at `Update Docs Onboarding for Daily branch` step

eng/scripts/Language-Settings.ps1
sima-zhu: Use the try catch to skip the failures

eng/pipelines/docindex.yml
```diff
@@ -5,7 +5,8 @@ jobs:
 
   - job: UpdateDocsMsBuildConfig
     pool:
-      vmImage: ubuntu-20.04
+      name: azsdk-pool-mms-ubuntu-2004-general
+      vmImage: MMSUbuntu20.04Compliant
```
weshaggard, line 9: Do we need to add the word Compliant at the end here?

eng/pipelines/docindex.yml
```diff
@@ -56,6 +69,11 @@ jobs:
           git status
         displayName: Checkout daily branch if it exists
         workingDirectory: $(DocRepoLocation)
+      - pwsh: |
+          $publicFeedUrl = "https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-for-net/nuget/v3/index.json"
+          Write-Host "##vso[task.setvariable variable=PackageSourceOverride]$publicFeedUrl"
+        displayName: Set package source variable
+        workingDirectory: $(DocRepoLocation)
```
weshaggard, line 76: nit: spacing
```suggestion

      - pwsh: |
          $publicFeedUrl = "https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-for-net/nuget/v3/index.json"
          Write-Host "##vso[task.setvariable variable=PackageSourceOverride]$publicFeedUrl"
        displayName: Set package source variable
        workingDirectory: $(DocRepoLocation)
        
```

Is there a reason this needs to be set as a variable dynamically here as opposed to hardcode below?

eng/pipelines/docindex.yml
```diff
@@ -75,6 +103,7 @@ jobs:
           TargetRepoOwner: $(DocRepoOwner)
           WorkingDirectory: $(DocRepoLocation)
           ScriptDirectory: $(Build.SourcesDirectory)/eng/common/scripts
+          PushArgs: -f
```
weshaggard, line 106: Are we certain we want to force push now?

eng/scripts/Language-Settings.ps1
weshaggard: I suspect they may have never been in our devops feed. If we need namespaces for them we might need to either add them or use nuget.org.

sima-zhu: It doesn't affect main/live site. Will add some fallback logic here to make sure we download the right package.

eng/pipelines/docindex.yml
sima-zhu: That's what we have in the pool.
image

eng/pipelines/docindex.yml
sima-zhu: It is needed for toc generation. That's how we know whether to use public feeds for daily docs or not. 

Our common script doesn't provide the $PackageSourceOverride parameters, as other languages have their way to use this.
Java: we have a step to configure .m2 settings. [Link](https://github.com/Azure/azure-sdk-for-java/blob/main/eng/pipelines/docindex.yml#L96)

.NET has no way to pass this to the agent. So use variable for .NET only and read variable from script.

eng/pipelines/docindex.yml
sima-zhu: We might have nightly pipeline pushes commits during docindex run, so it complains about the commits behind sometime.

By force pushing the commit, we might lose one or two commits from nightly pipeline. The miss on the commits is acceptable to me as following reasons.  First of all, it doesn't occur very often. Second, it doesn't always affect the same package. Last, this is daily refresh docs, so it will show up the right thing in the next day.

eng/pipelines/docindex.yml
sima-zhu: Here is the place to use the variable.
https://github.com/Azure/azure-sdk-for-net/pull/32062/files#r1008278055

eng/pipelines/docindex.yml
weshaggard: @mikeharder do you think we should stick with the image name we have configured in the pool? Or since we are only allowed one image per pool should we use a general name like "windows" for the vmImage name in our yml files?

eng/pipelines/docindex.yml
weshaggard: Ok so is more about setting this as an env variable. Can you please add a comment to that effect here so we don't accidently remove it later and break that usage.

eng/pipelines/docindex.yml
weshaggard: As long as we don't end up overwriting any manual commits, we might have that I suppose this will be ok.

eng/pipelines/docindex.yml
weshaggard: @sima-zhu in the meantime yes keep the name matching the image.

eng/scripts/docs/Docs-ToC.ps1
```diff
@@ -21,3 +21,196 @@ function Get-NamepspacesFromDll($dllPath) {
...
+
+function DownloadNugetPackage($package, $version, $destination) {
+    # $PackageSourceOverride is a global variable provided in
+    # Update-DocsMsPackages.ps1. Its value can set a "customSource" property.
+    # If it is empty then the property is not overridden.
+    $customPackageSource = $env:PACKAGESOURCEOVERRIDE
+    $defaultDownloadUri = "https://www.nuget.org/api/v2/package/$package/$version"
+    try {
+        if ($customPackageSource){
+            $publicFeedsUrl = "https://pkgs.dev.azure.com/azure-sdk/public/_apis/packaging/feeds/azure-sdk-for-net/nuget/packages/$package/versions/$version/content?api-version=6.0-preview.1"
```
weshaggard, line 33: Should we use the custom feed as the first part of this string? As it stands it seems like we are hardcoding this and it will need to be in sync with the other location we are setting. 'https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-for-net/nuget/v3/index.json' If we ever pass in another value there then this will be broken.

eng/scripts/docs/Docs-ToC.ps1
```diff
@@ -21,3 +21,196 @@ function Get-NamepspacesFromDll($dllPath) {
     }
```
weshaggard, line 21: We should add some tests for these new functions.

eng/pipelines/docindex.yml
sima-zhu: I have this `vmImage` as convention. Based on my testing, it is ok to use pool name only without specifying `vmImage` name.

eng/scripts/docs/Docs-ToC.ps1
sima-zhu: The location we hard-coded before can't use for download. I did not find how we utilize the public feeds for downloading purpose.

We can add download-devOpsPackages in `Invoke-DevOps.ps1` so we do not hardcoded anymore. A little extra work as the changes are inside of common script.

weshaggard: What I want to make sure is that we don't hard code this download path here but we actually pass in another feed in our customPackageSource property. So we need to somehow make those be in-sync.

eng/scripts/docs/Docs-ToC.ps1
```diff
@@ -21,3 +21,195 @@ function Get-NamepspacesFromDll($dllPath) {
...
     return $namespaces | Sort-Object -Unique
 }
+
+function DownloadNugetPackage($package, $version, $destination) {
+    # $PackageSourceOverride is a global variable provided in
+    # Update-DocsMsToc.ps1. Its value can set a "customSource" property.
+    # If it is empty then the property is not overridden.
+    $customPackageSource = Get-Variable -Name 'PackageSourceOverride' -ValueOnly -ErrorAction 'Ignore'
+    $defaultDownloadUri = "https://www.nuget.org/api/v2/package/$package/$version"
```
weshaggard, line 30: I don't think this is correct any longer I believe you want the real nuget feed url instead of the package/version url. YOu should be able to use the same nuget install command with either feed and the command should be the same.

eng/scripts/docs/Docs-ToC.ps1
```diff
@@ -21,3 +21,195 @@ function Get-NamepspacesFromDll($dllPath) {
...
+
+function Fetch-NamespacesFromNupkg ($package, $version) {
+    $tempLocation = (Join-Path ([System.IO.Path]::GetTempPath()) "extractNupkg/$package-$version")
+    if (!(Test-Path $tempLocation)) {
+        New-Item -ItemType Directory -Path $tempLocation -Force | Out-Null
+    }
+    $nupkgFilePath = "$tempLocation/$package-$version.nupkg"
+    Write-Host "Downloading nupkg packge to $nupkgFilePath ...."
+    DownloadNugetPackage -package $package -version $version -destination $nupkgFilePath
+    if (!(Test-Path $nupkgFilePath)) {
```
weshaggard, line 60: Part of the value of the nuget directdownload is you also don't need to extract the package as it does it for you already.

eng/scripts/docs/Docs-ToC.ps1
```diff
@@ -21,3 +21,181 @@ function Get-NamepspacesFromDll($dllPath) {
...
+function DownloadNugetPackage($package, $version, $destination) {
+    # $PackageSourceOverride is a global variable provided in
+    # Update-DocsMsToc.ps1. Its value can set a "customSource" property.
+    # If it is empty then the property is not overridden.
+    $customPackageSource = Get-Variable -Name 'PackageSourceOverride' -ValueOnly -ErrorAction 'Ignore'
+    if ($customPackageSource) {
+        # Download package from devop public feeds
+        Write-Host "Download from public feeds: $customPackageSource"
+        # Invoke the download link and store it into destination
+        nuget install -FallbackSource $customPackageSource $package -Version $version -DirectDownload -DependencyVersion Ignore -OutputDirectory $destination
```
sima-zhu, line 34: Go ahead with `$package.$version` as folder name since we might need to run both preview and latest. Make the script robust.

eng/scripts/docs/Docs-ToC.ps1
```diff
@@ -21,3 +21,178 @@ function Get-NamepspacesFromDll($dllPath) {
...
+    if (!(Test-Path $tempLocation)) {
+        New-Item -ItemType Directory -Path $tempLocation -Force | Out-Null
+    }
+    Write-Host "Downloading nupkg packge to $tempLocation ...."
+    DownloadNugetPackage -package $package -version $version -destination $tempLocation | Out-Null
+    if ($LASTEXITCODE -ne 0) {
+        Write-Host "Did not download the package correctly. Skipping..."
+        return @()
+    }
+    Write-Host "Searching ddl file..."
```
weshaggard, line 51:
```suggestion
    Write-Host "Searching for dll file..."
```

eng/scripts/docs/Docs-ToC.ps1
```diff
@@ -21,3 +21,178 @@ function Get-NamepspacesFromDll($dllPath) {
...
+        nuget install $package -Version $version -DirectDownload -DependencyVersion Ignore -OutputDirectory $destination
+    }
+}
+
+function Fetch-NamespacesFromNupkg ($package, $version) {
+    $tempLocation = (Join-Path ([System.IO.Path]::GetTempPath()) "extractNupkg")
+    if (!(Test-Path $tempLocation)) {
+        New-Item -ItemType Directory -Path $tempLocation -Force | Out-Null
+    }
+    Write-Host "Downloading nupkg packge to $tempLocation ...."
```
weshaggard, line 45:
```suggestion
    Write-Host "Downloading nupkg package $package with version $version to $tempLocation ...."
```

eng/scripts/docs/tests/Toc-Generation-Tests.ps1
```diff
@@ -0,0 +1,76 @@
+<#
+```
+Invoke-Pester -Output Detailed $PSScriptRoot\tests\Toc-Generation-Tests.ps1
```
weshaggard, line 3: Lets see if we can also add some tests for the Toc functions as well.

weshaggard (commented):
The changes look reasonable to me but I'd like to wait for @danieljurek to review for final sign-off as he understands the structure of the Toc files better.

eng/scripts/docs/tests/Toc-Generation-Tests.ps1
sima-zhu: `Get-Toc-Children` is worth to have more tests as they have some complicated logic. I will leave the `Get-dotnet-DocsMsTocData` as it is since the logic is pretty straight forward and it is the same structure as other languages.

eng/scripts/docs/Docs-ToC.ps1
```diff
@@ -21,3 +21,178 @@ function Get-NamepspacesFromDll($dllPath) {
...
 }
+
+function DownloadNugetPackage($package, $version, $destination) {
+    # $PackageSourceOverride is a global variable provided in
+    # Update-DocsMsToc.ps1. Its value can set a "customSource" property.
+    # If it is empty then the property is not overridden.
+    $customPackageSource = Get-Variable -Name 'PackageSourceOverride' -ValueOnly -ErrorAction 'Ignore'
+    # Download package from nuget, if no package found in nuget, then search devops public feeds if set.
+    Write-Host "Downloading $package with version $version..."
+    if ($customPackageSource) {
```
benbp, line 32: Should we handle `$LASTEXITCODE` for these nuget calls?

eng/scripts/docs/Docs-ToC.ps1
```diff
@@ -21,3 +21,178 @@ function Get-NamepspacesFromDll($dllPath) {
...
+        Write-Host "Did not download the package correctly. Skipping..."
+        return @()
+    }
+    Write-Host "Searching for dll file..."
+    $dllFile = @()
+    if (Test-Path "$tempLocation/$packageFolder/lib/netstandard2.0/"){
+        $dllFile = @(Get-ChildItem "$tempLocation/$packageFolder/lib/netstandard2.0/*" -Filter '*.dll' -Recurse)
+    }
+    if (!$dllFile -or ($dllFile.Count -gt 1)) {
+        Write-Warning "Can't find any dll file from $tempLocation/$packageFolder with target netstandard2.0."
```
benbp, line 58: Is this supposed to be a `-lt` check? If `-gt` the error message could be a bit confusing.

eng/scripts/docs/Docs-ToC.ps1
sima-zhu: Don't want to check the error code in each individual nuget command, so I have error handler when we call the Download function.

eng/scripts/docs/Docs-ToC.ps1
sima-zhu: Actually, the dllFile should only has one in each target framework (e.g. netstandard2.0). The ($dllFile.Count -gt 1) is redundant for normal world.

benbp: If there are > 1 dll files, and the log warning says "can't find any dll file" I think that could be misleading.

sima-zhu: After the test, I found we do have package which includes multiple dlls in one target. Added elseif the the case.
https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1971666&view=logs&j=dc056dfc-c0cf-5958-c8c4-8da4f91a3739&t=98c76ae6-047b-5869-f98c-bcf6099a854a&l=623 
image

```
/tmp/extractNupkg/Azure.Cosmos.4.0.0-preview3/lib/netstandard2.0/Azure.Cosmos.dll
/tmp/extractNupkg/Azure.Cosmos.4.0.0-preview3/lib/netstandard2.0/Microsoft.Azure.Cosmos.Core.dll 
/tmp/extractNupkg/Azure.Cosmos.4.0.0-preview3/lib/netstandard2.0/Microsoft.Azure.Cosmos.Direct.dll 
/tmp/extractNupkg/Azure.Cosmos.4.0.0-preview3/lib/netstandard2.0/Microsoft.Azure.Cosmos.Serialization.HybridRow.dll

```

eng/scripts/docs/Docs-ToC.ps1
```diff
@@ -21,3 +21,188 @@ function Get-NamepspacesFromDll($dllPath) {
...
+    DownloadNugetPackage -package $package -version $version -destination $tempLocation | Out-Null
+    $packageFolder = "$package.$version"
+    if ($LASTEXITCODE -ne 0 -or !$(Test-Path $tempLocation/$packageFolder)) {
+        Write-Host "Did not download the package correctly. Skipping..."
+        return @()
+    }
+    Write-Host "Searching for dll file..."
+    $dllFile = @()
+    if (Test-Path "$tempLocation/$packageFolder/lib/netstandard2.0/"){
+        $dllFile = @(Get-ChildItem "$tempLocation/$packageFolder/lib/netstandard2.0/*" -Filter '*.dll' -Recurse)
```
weshaggard, line 55: Shouldn't this be $dllFiles as this is a list?

eng/scripts/docs/Docs-ToC.ps1
```diff
@@ -21,3 +21,188 @@ function Get-NamepspacesFromDll($dllPath) {
...
+        Write-Warning "Can't find any dll file from $tempLocation/$packageFolder with target netstandard2.0."
+        $dllFiles = Get-ChildItem "$tempLocation/$packageFolder/lib/*" -Filter '*.dll' -Recurse
+        if (!$dllFiles) {
+            Write-Error "Can't find any dll file from $tempLocation/$packageFolder." -ErrorAction Continue
+            return @()
+        }
+        $dllFile = $dllFiles[0]
+    }
+    elseif ($dllFile.Count -gt 1) {
+        Write-Warning "There are multiple dll files in target netstandard2.0. Choose the one with package name."
```
weshaggard, line 67: You should move this warning after you select the dll so you can include the name of the dll in the output.

eng/scripts/docs/Docs-ToC.ps1
```diff
@@ -21,3 +21,188 @@ function Get-NamepspacesFromDll($dllPath) {
...
+            $dllFile = [System.IO.DirectoryInfo]"$tempLocation/$packageFolder/lib/netstandard2.0/$package.dll"
+        }
+        else {
+            $dllFile = $dllFile[0]
+        }
+    }
+    Write-Host "Dll file found: $($dllFile.FullName)"
+    $namespaces = Get-NamespacesFromDll $dllFile.FullName
+    if (!$namespaces) {
+        Write-Error "Can't find namespaces from dll file $($dllFile.FullName)." -ErrorAction Continue
```
weshaggard, line 79: Why are you setting ErrorAction Continue on these Write-Error messages? Also should we should probably be using the LogError helpers in these cases so they output correctly as errors in DevOps.

eng/scripts/docs/tests/Toc-Generation-Tests.ps1
```diff
@@ -0,0 +1,115 @@
...
+    Remove-Item "$PSScriptRoot/outputs" -Recurse
+}
+# Test plan:
+# 1. Tests on Fetch-NamespacesFromNupkg from nuget source. 
+# 2. Tests on Fetch-NamespacesFromNupkg from public feeds. 
+# 3. Tests on Get-Toc-Children for latest
+# 4. Tests on Get-Toc-Children for preview
+Describe "Fetch-NamespacesFromNupkg-Nuget" -Tag "UnitTest" {
+    # Passed cases
+    It "Fetch namespaces from package downloads from nuget" -TestCases @(
```
weshaggard, line 40: We should have the cosmos packages that were causing issues added as test cases.

kurtzeborn: If cosmos has been problematic, it should definitely be added to your list of test cases, shouldn't it?

eng/scripts/docs/Docs-ToC.ps1
sima-zhu: I have one testing failed in the middle. I switched to logging.ps1 apis if it's non stop behavior.

eng/scripts/docs/tests/Toc-Generation-Tests.ps1
sima-zhu: This is newly found case. Will add to test.

sima-zhu: I will use Azure.Cosmos one for testing as this is deterministic. `Microsoft.Azure.Cosmos` will pick the first one which is flaky from test perspective.

eng/scripts/docs/Docs-ToC.ps1
sima-zhu: Tested in local. The tests failed at Log function in logging.ps1 since we did not import the logging script in test script. I switched back the Write-Error

eng/scripts/docs/tests/Toc-Generation-Tests.ps1
weshaggard: If we think Microsoft.Azure.Cosmos is flaky that means that we also have a flaky implementation that will give indeterministic results. If that is the case we should fix the implementation to order the dlls so that the result is more stable.

eng/scripts/docs/Docs-ToC.ps1
sima-zhu: https://github.com/Azure/azure-sdk-for-net/pull/32062/files#diff-4b590925689e1df66d5c28ddc67d15ded4db6d686eb7c7878e0c8238feb57e25R76

I have printed at end. The message here shows the route to find dll

eng/scripts/docs/tests/Toc-Generation-Tests.ps1
sima-zhu: Then we need to order the dll first and then fetch the first.

eng/scripts/docs/Docs-ToC.ps1
```diff
@@ -21,3 +21,188 @@ function Get-NamepspacesFromDll($dllPath) {
...
     return $namespaces | Sort-Object -Unique
 }
+
+function DownloadNugetPackage($package, $version, $destination) {
+    # $PackageSourceOverride is a global variable provided in
+    # Update-DocsMsToc.ps1. Its value can set a "customSource" property.
+    # If it is empty then the property is not overridden.
+    $customPackageSource = Get-Variable -Name 'PackageSourceOverride' -ValueOnly -ErrorAction 'Ignore'
+    # Download package from nuget, if no package found in nuget, then search devops public feeds if set.
+    Write-Host "Downloading source: '$customPackageSource'. If it is empty, we default to use nuget gallery."
```
weshaggard, line 31: It is still worth having the package name and version for diagnostic purposes.

sima-zhu: Actually, we have the $package $version information logged out. No need to log here it is a redundant information. 
https://github.com/Azure/azure-sdk-for-net/pull/32062/files#diff-4b590925689e1df66d5c28ddc67d15ded4db6d686eb7c7878e0c8238feb57e25R45

eng/scripts/docs/tests/Toc-Generation-Tests.ps1
sima-zhu: Done with the change

eng/scripts/docs/Docs-ToC.ps1
```diff
@@ -21,3 +21,189 @@ function Get-NamepspacesFromDll($dllPath) {
...
+        }
+    }
+    if (!$dllFileName) {
+        $dllFiles = $dllFiles | Get-Unique | Sort-Object
+        $dllFileName = $dllFiles[0].FullName
+    }
+    LogDebug "Dll file found: $dllFileName"
+    $namespaces = Get-NamespacesFromDll $dllFileName
+    if (!$namespaces) {
+        LogError "Can't find namespaces from dll file $dllFileName." -ErrorAction Continue
```
weshaggard, line 80: Looks like this ErrorAction is leftover and should be removed. Looks like there might be a few of them left over.

weshaggard (approved):

PR #32073, opened 2022-10-27
AzmonExporter - Limit OTel version to be specific

cijothomas:
AzMonExporter has a dependency on non-stable version of OTel SDK. As the OTel SDK can have breaking changes, we should restrict upper bound, else exporter might break if user updates to a newer version of OTel SDK, which has breaking changes.

Once the dependency is updated to stable version, we should still cap upper bound at 2.0], instead of leaving upper bound open.

### DIFF
--- a/eng/Packages.Data.props
+++ b/eng/Packages.Data.props
@@ -116,7 +116,7 @@
     <PackageReference Update="System.IdentityModel.Tokens.Jwt" Version="6.5.0" />
 
     <!-- OpenTelemetry dependency approved for Azure.Monitor.OpenTelemetry.Exporter package only -->
-    <PackageReference Update="OpenTelemetry" Version="1.4.0-beta.2"  Condition="'$(MSBuildProjectName)' == 'Azure.Monitor.OpenTelemetry.Exporter'" />
+    <PackageReference Update="OpenTelemetry" Version="[1.4.0-beta.2]"  Condition="'$(MSBuildProjectName)' == 'Azure.Monitor.OpenTelemetry.Exporter'" />
     <PackageReference Update="OpenTelemetry.Extensions.PersistentStorage" Version="1.0.0-beta.1" Condition="'$(MSBuildProjectName)' == 'Azure.Monitor.OpenTelemetry.Exporter'" />
   </ItemGroup>
 


weshaggard:
@cijothomas thanks for the contribute but I'm not sure this will do what you expect. Can you give me more details on how this helps? In generally we try to avoid ranges in our library dependency an if there is any issue it should be pinned inside of the application. .NET will never try to use a higher version anyway unless something in the application is trying to pull it to a higher version.

cijothomas:
> @cijothomas thanks for the contribute but I'm not sure this will do what you expect. Can you give me more details on how this helps? In generally we try to avoid ranges in our library dependency an if there is any issue it should be pinned inside of the application. .NET will never try to use a higher version anyway unless something in the application is trying to pull it to a higher version.

If user brings a newer version of OTel SDK (direct or indirect via other deps), like 1.4.0-beta.6, which can contain breaking changes, then AzMonExporter breaks. Restricting the nuget dependency to specific version can avoid that (as the upgrade will be blocked by nuget).

TimothyMothra (approved):

cijothomas:
> > @cijothomas thanks for the contribute but I'm not sure this will do what you expect. Can you give me more details on how this helps? In generally we try to avoid ranges in our library dependency an if there is any issue it should be pinned inside of the application. .NET will never try to use a higher version anyway unless something in the application is trying to pull it to a higher version.
> 
> If user brings a newer version of OTel SDK (direct or indirect via other deps), like 1.4.0-beta.6, which can contain breaking changes, then AzMonExporter breaks. Restricting the nuget dependency to specific version can avoid that (as the upgrade will be blocked by nuget).

There is a incoming breaking change already for next beta: https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/CHANGELOG.md#unreleased

vishweshbankwar (approved):

reyang (approved):

cijothomas:
@weshaggard Do you have objections merging this?

weshaggard:
>If user brings a newer version of OTel SDK (direct or indirect via other deps), like 1.4.0-beta.6, which can contain breaking changes, then AzMonExporter breaks. Restricting the nuget dependency to specific version can avoid that (as the upgrade will be blocked by nuget).

You are just trading one break off for another break. If something pulls in 1.4.0.beta.6 and a package with this version range pinned it will also break the application. I'm not a big fan of this type of pinning, especially for libraries, but given this is only used by the one library (Azure.Monitor.OpenTelemetry.Exporter) that I can see, and it is still in beta as well I'm OK with taking this change. We of course will not want to be doing this once we ship a stable/GA version.

Do we know when we will have a stable version of this dependency that we can use to avoid these types of breaks?

cijothomas:
> > If user brings a newer version of OTel SDK (direct or indirect via other deps), like 1.4.0-beta.6, which can contain breaking changes, then AzMonExporter breaks. Restricting the nuget dependency to specific version can avoid that (as the upgrade will be blocked by nuget).
> 
> You are just trading one break off for another break. If something pulls in 1.4.0.beta.6 and a package with this version range pinned it will also break the application. I'm not a big fan of this type of pinning, especially for libraries, but given this is only used by the one library (Azure.Monitor.OpenTelemetry.Exporter) that I can see, and it is still in beta as well I'm OK with taking this change. We of course will not want to be doing this once we ship a stable/GA version.
> 
> Do we know when we will have a stable version of this dependency that we can use to avoid these types of breaks?

This is *only* while we depend on beta versions. OTel follows semantic version, and guarantees no breaking change (unless major version bump).

_". If something pulls in 1.4.0.beta.6 and a package with this version range pinned it will also break the application."_
I dont see how. Because nuget will block the upgrade. The alternative w/o pinning is app crash at runtime.

cijothomas:
_Do we know when we will have a stable version of this dependency_

By end of Nov 2022, OTel 1.4.0 stable will release, so we can get rid of this :)

cijothomas:
Proceeding to merge, with the understanding that this is only for the beta duration, and can be removed when OTel releases 1.4.0 stable (Nov end)

weshaggard:
>I dont see how. Because nuget will block the upgrade. The alternative w/o pinning is app crash at runtime.

You are correct that Nuget will block it but that turns it into a break at build/restore time over runtime which I do agree is a little nicer for developers but it still break them nonetheless.

PR #32085, opened 2022-10-28
[PrivateDns] Add customization code for RecordSet

dvbb:


Yao725 (commented):
I noticed three things in this PR:
1. There are many models and resources that have the duplicated name compared with DNS, is it allowed? Or should we rename them?
2. The type of ifMatch is not unified, sometimes it's string but sometimes it's ETag. Could you please double comfirm?
3. The parameter name `MXRecordName` need to be corrected.

ArthurMa1978:
> I noticed three things in this PR:
> 
> 1. There are many models and resources that have the duplicated name compared with DNS, is it allowed? Or should we rename them?
> 2. The type of ifMatch is not unified, sometimes it's string but sometimes it's ETag. Could you please double comfirm?
> 3. The parameter name `MXRecordName` need to be corrected.

Good point for the model name question, the most simple way is prepend `PrivateDns` prefix, but looks a little bit weird, let's check with @m-nash .

sdk/privatedns/Azure.ResourceManager.PrivateDns/tests/Azure.ResourceManager.PrivateDns.Tests.csproj
```diff
@@ -1,4 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
+  <ItemGroup>
```
Yao725, line 2: Why do we need this?

sdk/privatedns/Azure.ResourceManager.PrivateDns/tests/Scenario/RecordTests.cs
```diff
@@ -0,0 +1,482 @@
...
+            Assert.IsTrue(getResponse.Value.Data.PrivateDnsSoaRecord.RetryTimeInSeconds > 0);
+            Assert.IsTrue(getResponse.Value.Data.PrivateDnsSoaRecord.ExpireTimeInSeconds > 0);
+            Assert.IsTrue(getResponse.Value.Data.PrivateDnsSoaRecord.MinimumTtlInSeconds > 0);
+
+            // GetAll
+            var list = await collection.GetAllAsync().ToEnumerableAsync();
+            Assert.IsNotEmpty(list);
+            ValidateRecordBaseInfo(list.First(item => item.Data.Name == soaRecordName).Data, soaRecordName);
+
+            // Delete - Type SOA cannot delete
```
Yao725, line 355: Do you know what will happen if we call delete method for SOA? If the service will return 400 or some other errors, how about we just remove this method from `PrivateDnsSoaRecordResource`?

sdk/privatedns/Azure.ResourceManager.PrivateDns/src/Customization/PrivateDnsRecordData.cs
```diff
@@ -40,66 +39,50 @@ public RecordSetData()
...
-                CnameRecord.Cname = value;
+                if (PrivateDnsCnameRecordInfo is null)
+                    PrivateDnsCnameRecordInfo = new PrivateDnsCnameRecordInfo();
+                PrivateDnsCnameRecordInfo.Cname = value;
             }
         }
 
         /// <summary> The list of MX records in the record set. </summary>
-        public IList<MxRecord> MxRecords { get; }
+        public IList<PrivateDnsMXRecordInfo> MxRecords { get; }
```
Yao725, line 78: I assume the property should be called `MXRecords`

sdk/privatedns/Azure.ResourceManager.PrivateDns/src/Generated/PrivateZoneData.cs
```diff
@@ -56,7 +56,7 @@ internal PrivateZoneData(ResourceIdentifier id, string name, ResourceType resour
         /// <summary> The maximum number of record sets that can be created in this Private DNS zone. This is a read-only property and any attempt to set this value will be ignored. </summary>
         public long? MaxNumberOfRecordSets { get; }
```
Yao725, line 57: It seems that we changed `RecordSets` to `Records` in other places, do we also need to change this one?

sdk/privatedns/Azure.ResourceManager.PrivateDns/api/Azure.ResourceManager.PrivateDns.netstandard2.0.cs
```diff
@@ -377,36 +421,65 @@ public partial class VirtualNetworkLinkResource : Azure.ResourceManager.ArmResou
...
+        public string Target { get { throw null; } set { } }
+        public int? Weight { get { throw null; } set { } }
+    }
+    public partial class PrivateDnsTxtRecordInfo
+    {
+        public PrivateDnsTxtRecordInfo() { }
+        public System.Collections.Generic.IList<string> Values { get { throw null; } }
+    }
     [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
     public readonly partial struct ProvisioningState : System.IEquatable<Azure.ResourceManager.PrivateDns.Models.ProvisioningState>
```
Yao725, line 484: Too generic, we should rename this model

dvbb: `ProvisioningState` rename to `PrivateDnsProvisioningState `

sdk/privatedns/Azure.ResourceManager.PrivateDns/tests/Scenario/RecordTests.cs
dvbb: It will return `400(Bad Request){"code":"BadRequest","message":"Record sets of type 'SOA' with name '@' cannot be deleted."}`
I remove that method in customization, please double confirm. If we confirm that change, I will remove the delete methods in dns.

Yao725 (approved):

ArthurMa1978 (approved):

PR #32166, opened 2022-10-28
[Event Hubs] Function bindings test tweaks

jsquire:
# Summary

The focus of these changes is to address some flakiness in the Functions tests and to perform some minor refactoring.

### DIFF
--- a/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/tests/EventHubEndToEndTests.cs
+++ b/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/tests/EventHubEndToEndTests.cs
@@ -644,7 +644,7 @@ namespace Microsoft.Azure.WebJobs.Host.EndToEndTests
             // so that we get coverage for receiving events from the same partition in multiple chunks
             private const int EventsPerPartition = 15;
             private const int PartitionCount = 5;
-            private const int TotalEventsCount = EventsPerPartition * PartitionCount;
+            private const string PartitionKeyPrefix = "test_pk";
 
             public static async Task SendEvents_TestHub(
                 string input,
@@ -660,7 +660,7 @@ namespace Microsoft.Azure.WebJobs.Host.EndToEndTests
                 for (int i = 0; i < PartitionCount; i++)
                 {
                     evt = new EventData(Encoding.UTF8.GetBytes(input));
-                    tasks.Add(client.SendAsync(Enumerable.Repeat(evt, EventsPerPartition), new SendEventOptions() { PartitionKey = "test_pk" + i }));
+                    tasks.Add(client.SendAsync(Enumerable.Repeat(evt, EventsPerPartition), new SendEventOptions() { PartitionKey = PartitionKeyPrefix + i }));
                 }
 
                 await Task.WhenAll(tasks);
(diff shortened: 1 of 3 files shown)


sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/tests/EventHubEndToEndTests.cs
```diff
@@ -31,18 +31,19 @@ public class EventHubEndToEndTests : WebJobsEventHubTestBase
...
         public void SetUp()
         {
             _results = new List<string>();
             _eventWait = new ManualResetEvent(initialState: false);
         }
 
+        [TearDown]
+        public void TearDown()
+        {
+            _eventWait?.Dispose();
```
jsquire, line 44:
```suggestion
            _eventWait.Dispose();
```

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/tests/EventHubEndToEndTests.cs
```diff
@@ -712,14 +706,18 @@ public class EventHubTestInitialOffsetFromEnqueuedTimeJobs
 
             public static void ProcessMultipleEvents([EventHubTrigger(TestHubName, Connection = TestHubName)] EventData[] events)
             {
-                Assert.LessOrEqual(events.Length, ExpectedEventsCount);
+                Assert.GreaterOrEqual(events.Length, ExpectedEventsCount);
+
+                // there's potentially some level of rewind due to clock differences; allow a small delta when validating.
+                var earliestAllowedOffset = _initialOffsetEnqueuedTimeUTC.AddMilliseconds(-500);
```
m-redding, line 712: Where does this variable get used?

m-redding (approved):

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/tests/EventHubEndToEndTests.cs
```diff
@@ -372,8 +371,8 @@ public async Task EventHub_InitialOffsetFromEnd()
                 });
             using (jobHost)
             {
-                // We don't expect to get signalled as there should be no messages received with a FromEnd initial offset
-                bool result = _eventWait.WaitOne(Timeout);
+                // We don't expect to get signaled as there should be no messages received with a FromEnd initial offset
+                bool result = _eventWait.WaitOne(1000);
```
JoshLove-msft, line 375: Why do we hardcode this one? What was Timeout?

jsquire: The flow basically says "wait for some period of time to make sure that we DON'T get something" then send something and verify that we get it.    The overall timeout that we apply for tests is several minutes; waiting for a second is enough to prove that nothing will be received.  We don't need to wait the full standard test timeout to confirm that something doesn't happen.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/tests/EventHubEndToEndTests.cs
jsquire: It doesn't, but should have been used for the assertion on 725.   Fixed!

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/tests/EventHubEndToEndTests.cs
JoshLove-msft: Ah, would it make sense to wait 5 seconds just to make it more likely the test would fail  when it should?

JoshLove-msft (approved):

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/tests/EventHubEndToEndTests.cs
jsquire: Shouldn't hurt.   Shifted to 5 seconds and added a constant to make the intent more clear.

PR #32170, opened 2022-10-28
[AzureMonitorExporter] Add support for UpDownCounter and ObservableUpDownCounter

TimothyMothra:
### Changes
- Add support for UpDownCounter and ObservableUpDownCounter
- Update TestExtensions to return the MetricReader.
  This is needed to invoke collection of Observable Metrics.

### DIFF
--- a/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Customizations/Models/MetricsData.cs
+++ b/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Customizations/Models/MetricsData.cs
@@ -41,6 +41,8 @@ namespace Azure.Monitor.OpenTelemetry.Exporter.Models
                 MetricType.LongGauge => true,
                 MetricType.LongSum => true,
                 MetricType.Histogram => true,
+                MetricType.DoubleSumNonMonotonic => true,
+                MetricType.LongSumNonMonotonic => true,
                 _ => false
             };
     }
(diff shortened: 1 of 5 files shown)


sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Customizations/Models/MetricsData.cs
```diff
@@ -41,6 +41,8 @@ metricType switch
                 MetricType.LongGauge => true,
                 MetricType.LongSum => true,
                 MetricType.Histogram => true,
+                MetricType.DoubleSumNonMonotonic => true,
```
cijothomas, line 44: the whole IsSupportedType seems not relevant? Might be good to clean up later. Not blocker.

cijothomas (approved):

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Customizations/Models/MetricsData.cs
TimothyMothra: I had the same question. It's being used to skip some extra processing.

https://github.com/Azure/azure-sdk-for-net/blob/71ce2a14eff9871622d20914415a297b188e3e72/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/MetricHelper.cs#L22-L39

vishweshbankwar: This was added initially to skip long types. But since, we added that later this is no longer needed

PR #32177, opened 2022-10-29
[Text Analytics] Add support for NER resolutions

### LINKED ISSUE #31276: [Text Analytics] Add support for NER resolutions

# Tasks
- [ ] Implementation
- [ ] Tests
- [ ] Docs

joseharriaga:
Resolves: #31276

Features added:
- Added the `CategorizedEntity.Resolutions` property.
- Added the following classes: `AgeResolution`, `AreaResolution`, `BaseResolution`, `BooleanResolution`, `CurrencyResolution`, `DateTimeResolution`, `InformationResolution`, `LengthResolution`, `NumberResolution`, `NumericRangeResolution`, `OrdinalResolution`, `SpeedResolution`, `TemperatureResolution`, `TemporalSpanResolution` `VolumeResolution`, and `WeightResolution`.
- Added the following enums: `AgeUnit`, `AreaUnit`, `DateTimeSubKind`, `InformationUnit`,`LengthUnit`, `NumberKind`, `RangeKind`, `RelativeTo`, `SpeedUnit`, `TemperatureUnit`,`TemporalModifier`, `VolumeUnit`, and `WeightUnit`.

### DIFF
--- a/sdk/textanalytics/Azure.AI.TextAnalytics/api/Azure.AI.TextAnalytics.netstandard2.0.cs
+++ b/sdk/textanalytics/Azure.AI.TextAnalytics/api/Azure.AI.TextAnalytics.netstandard2.0.cs
@@ -1596,6 +1596,7 @@ namespace Azure.AI.TextAnalytics
     }
     public static partial class TextAnalyticsModelFactory
     {
+        public static Azure.AI.TextAnalytics.AgeResolution AgeResolution(Azure.AI.TextAnalytics.AgeUnit unit, double value) { throw null; }
         [System.ComponentModel.EditorBrowsableAttribute(System.ComponentModel.EditorBrowsableState.Never)]
         public static Azure.AI.TextAnalytics.AnalyzeActionsResult AnalyzeActionsResult(System.Collections.Generic.IEnumerable<Azure.AI.TextAnalytics.ExtractKeyPhrasesActionResult> extractKeyPhrasesActionResult, System.Collections.Generic.IEnumerable<Azure.AI.TextAnalytics.RecognizeEntitiesActionResult> recognizeEntitiesActionResults, System.Collections.Generic.IEnumerable<Azure.AI.TextAnalytics.RecognizePiiEntitiesActionResult> recognizePiiEntitiesActionResults, System.Collections.Generic.IEnumerable<Azure.AI.TextAnalytics.RecognizeLinkedEntitiesActionResult> recognizeLinkedEntitiesActionsResults, System.Collections.Generic.IEnumerable<Azure.AI.TextAnalytics.AnalyzeSentimentActionResult> analyzeSentimentActionsResults) { throw null; }
         [System.ComponentModel.EditorBrowsableAttribute(System.ComponentModel.EditorBrowsableState.Never)]
@@ -1619,7 +1620,9 @@ namespace Azure.AI.TextAnalytics
         public static Azure.AI.TextAnalytics.AnalyzeSentimentResult AnalyzeSentimentResult(string id, Azure.AI.TextAnalytics.TextAnalyticsError error) { throw null; }
         public static Azure.AI.TextAnalytics.AnalyzeSentimentResult AnalyzeSentimentResult(string id, Azure.AI.TextAnalytics.TextDocumentStatistics statistics, Azure.AI.TextAnalytics.DocumentSentiment documentSentiment) { throw null; }
         public static Azure.AI.TextAnalytics.AnalyzeSentimentResultCollection AnalyzeSentimentResultCollection(System.Collections.Generic.IEnumerable<Azure.AI.TextAnalytics.AnalyzeSentimentResult> list, Azure.AI.TextAnalytics.TextDocumentBatchStatistics statistics, string modelVersion) { throw null; }
+        public static Azure.AI.TextAnalytics.AreaResolution AreaResolution(Azure.AI.TextAnalytics.AreaUnit unit, double value) { throw null; }
         public static Azure.AI.TextAnalytics.AssessmentSentiment AssessmentSentiment(Azure.AI.TextAnalytics.TextSentiment sentiment, double positiveScore, double negativeScore, string text, bool isNegated, int offset, int length) { throw null; }
+        public static Azure.AI.TextAnalytics.BooleanResolution BooleanResolution(bool value) { throw null; }
         [System.ComponentModel.EditorBrowsableAttribute(System.ComponentModel.EditorBrowsableState.Never)]
         public static Azure.AI.TextAnalytics.CategorizedEntity CategorizedEntity(string text, string category, string subCategory, double score) { throw null; }
         [System.ComponentModel.EditorBrowsableAttribute(System.ComponentModel.EditorBrowsableState.Never)]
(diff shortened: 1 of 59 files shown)


sdk/textanalytics/Azure.AI.TextAnalytics/src/GeneratorStubs/Resolutions.cs
```diff
@@ -0,0 +1,582 @@
...
+    public readonly partial struct VolumeUnit { }
+
+    [CodeGenModel("WeightUnit")]
+    public readonly partial struct WeightUnit { }
+
+    #endregion
+
+    #region Resolutions
+
+    [CodeGenModel("AgeResolution")]
```
joseharriaga, line 56: @heaths : I'm not a fan of how I'm exposing these Resolution types, but I couldn't figure out how to compose an autorest transform that would turn them into public read-only types (i.e., types with internal constructors, and only getters with no setters). Do you know what the trick might be? Or, perhaps, is the current approach acceptable?

heaths: By default, if they are only used in responses they should already be read-only. Is that not what you're seeing?

See https://github.com/Azure/azure-sdk-for-net/search?q=x-accessibility&type=code for some examples of how to add the `x-accessibility` via transform. Your `where` clause could be something like `$.definitions.*` and then your transform could check against a static list of which models you want to make public, for example.

joseharriaga: I talked to Michael and ended up filing a related bug: https://github.com/Azure/autorest.csharp/issues/2823. I will keep it like this for now. Since it's only a handful of simple classes, I don't think it's problematic.

mssfang (approved):

sdk/textanalytics/Azure.AI.TextAnalytics/src/CategorizedEntity.cs
```diff
@@ -67,5 +72,10 @@ internal CategorizedEntity(string text, string category, string subcategory, dou
         /// Gets the length of the matching text in the input document.
         /// </summary>
         public int Length { get; }
+
+        /// <summary>
+        /// Gets the collection of entity resolutions.
```
kristapratico, line 77: Feel free to add this aka link to the docs: https://aka.ms/azsdk/language/ner-resolutions

kristapratico (approved):

sdk/textanalytics/Azure.AI.TextAnalytics/api/Azure.AI.TextAnalytics.netstandard2.0.cs
```diff
@@ -926,6 +1176,25 @@ public partial class RecognizePiiEntitiesResultCollection : System.Collections.O
         public Azure.AI.TextAnalytics.TextDocumentBatchStatistics Statistics { get { throw null; } }
     }
     [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
+    public readonly partial struct RelativeTo : System.IEquatable<Azure.AI.TextAnalytics.RelativeTo>
+    {
+        private readonly object _dummy;
+        private readonly int _dummyPrimitive;
```
deyaaeldeen, line 1182: I am curious, what are these things?

sdk/textanalytics/Azure.AI.TextAnalytics/api/Azure.AI.TextAnalytics.netstandard2.0.cs
```diff
@@ -1,5 +1,32 @@
 namespace Azure.AI.TextAnalytics
 {
+    public partial class AgeResolution : Azure.AI.TextAnalytics.BaseResolution
```
deyaaeldeen, line 3: Does it make sense to put these things in a new sub namespace to not pollute the main one?

sdk/textanalytics/Azure.AI.TextAnalytics/tests/RecognizeCustomEntitiesTests.cs
```diff
@@ -359,6 +359,8 @@ private void ValidateInDocumentResult(CategorizedEntityCollection entities, List
                 {
                     Assert.IsNotEmpty(entity.SubCategory);
                 }
+
+                Assert.IsNotNull(entity.Resolutions);
```
deyaaeldeen, line 363: are resolutions part of the custom entity recognition feature too?

deyaaeldeen (approved):
Looks good! Thanks for creating such a comprehensive coverage!

sdk/textanalytics/Azure.AI.TextAnalytics/api/Azure.AI.TextAnalytics.netstandard2.0.cs
joseharriaga: Normally yes, but in the particular case of Text Analytics, it appears that the current design puts all its models in the topmost namespace. I do have a note to discuss this in the arch review. I'm curious if all these new types are a good excuse to revisit this previous design decision.

sdk/textanalytics/Azure.AI.TextAnalytics/tests/RecognizeCustomEntitiesTests.cs
joseharriaga: According to the swagger spec, NER and custom NER share the same type "EntityWithResolution", which is called `CategorizedEntity` in the .NET library. However, custom NER never actually returns any resolutions, so in the case the property is always an empty collection (but not null, which is what I'm checking here).

sdk/textanalytics/Azure.AI.TextAnalytics/src/CategorizedEntity.cs
joseharriaga: Thank you!

sdk/textanalytics/Azure.AI.TextAnalytics/tests/RecognizeCustomEntitiesTests.cs
deyaaeldeen: should it be null though?

joseharriaga: As per the .NET framework design guidelines, we shouldn't return null values for collection properties, and to return an empty collection instead.

https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/guidelines-for-collections

deyaaeldeen: I agree with the rule in general, but this seems like a special case. Maybe an empty collection signals to the customer it could also be non-empty? Would returning null here means do not even expect to have a collection empty or not?

joseharriaga: My feeling is that the fact that the property exists at all signals to the customer that it serves some purpose, which in the case of custom NER, it unfortunately doesn't. I think that returning null wouldn't be enough to decisively override this assumption and worry that instead it could add to the confusion because users can tell it's different but not why ("is this special in some way or is this just a bug?"). Ideally, NER and custom NER would return different types, but this is not possible without a breaking change. For the record, I also have this question in my notes to discuss during the arch review.

sdk/textanalytics/Azure.AI.TextAnalytics/api/Azure.AI.TextAnalytics.netstandard2.0.cs
joseharriaga: I had the same question. These are auto-generated, private, and show up in other places that pre-date me (including in other libraries too). 

I asked for more info. From Heath:
> Those are just artifacts from the metadata reader we get from structs.

sdk/textanalytics/Azure.AI.TextAnalytics/tests/RecognizeCustomEntitiesTests.cs
deyaaeldeen: On another thought, could the entity class stay the same and add one that inherits from it to be used with the prebuilt entity recognition result only?

sdk/textanalytics/Azure.AI.TextAnalytics/api/Azure.AI.TextAnalytics.netstandard2.0.cs
```diff
@@ -926,6 +1176,25 @@ public partial class RecognizePiiEntitiesResultCollection : System.Collections.O
         public Azure.AI.TextAnalytics.TextDocumentBatchStatistics Statistics { get { throw null; } }
     }
     [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
+    public readonly partial struct RelativeTo : System.IEquatable<Azure.AI.TextAnalytics.RelativeTo>
```
kinelski, line 1179: I think the name `RelativeTo` may be a bit generic. We should consider renaming it later to avoid conflicts with other libraries.

sdk/textanalytics/Azure.AI.TextAnalytics/api/Azure.AI.TextAnalytics.netstandard2.0.cs
```diff
@@ -216,6 +288,42 @@ public partial class ClassifyDocumentResultCollection : System.Collections.Objec
...
+        public string Iso4217 { get { throw null; } }
+        public string Unit { get { throw null; } }
+        public double Value { get { throw null; } }
+    }
+    public partial class DateTimeResolution : Azure.AI.TextAnalytics.BaseResolution
+    {
+        internal DateTimeResolution() { }
+        public Azure.AI.TextAnalytics.DateTimeSubKind DateTimeSubKind { get { throw null; } }
+        public Azure.AI.TextAnalytics.TemporalModifier? Modifier { get { throw null; } }
+        public string Timex { get { throw null; } }
```
kinelski, line 303: Silly question but I'm assuming the `x` comes from `extended` or something of the sort, and `Timex` is not a typo. Is that right?

kinelski (approved):
Two minor comments but definitely not blocking. LGTM!

sdk/textanalytics/Azure.AI.TextAnalytics/api/Azure.AI.TextAnalytics.netstandard2.0.cs
joseharriaga: I agree! I have this one in my notes to be discussed in a future arch review.

sdk/textanalytics/Azure.AI.TextAnalytics/api/Azure.AI.TextAnalytics.netstandard2.0.cs
joseharriaga: Correct, it's not a typo; this is what the service team called it in their swagger. That being said, I'm personally expecting that we will give it a better name in the future as we prepare it for an arch review.

PR #32197, opened 2022-10-31
[Text Analytics] Add support for script detection

### LINKED ISSUE #31272: [Text Analytics] Add support for script detection

# Tasks
- [ ] Implementation
- [ ] Tests
- [ ] Docs

joseharriaga:
Resolves: #31272

Features added:
- Added `Script` property to `DetectedLanguage`.
- Added `ScriptKind` enum.

### DIFF
--- a/sdk/textanalytics/Azure.AI.TextAnalytics/tests/DetectLanguageTests.cs
+++ b/sdk/textanalytics/Azure.AI.TextAnalytics/tests/DetectLanguageTests.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 using Azure.Core.TestFramework;
 using NUnit.Framework;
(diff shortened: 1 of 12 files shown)


sdk/textanalytics/Azure.AI.TextAnalytics/tests/DetectLanguageTests.cs
```diff
@@ -266,11 +266,27 @@ public void DetectLanguageBatchDisableServiceLogsThrows()
...
         }
 
+        [RecordedTest]
+        [ServiceVersion(Min = TextAnalyticsClientOptions.ServiceVersion.V2022_10_01_Preview)]
+        public async Task DetectLanguageBatchWithScriptTest()
+        {
+            string document = "Tumhara naam kya hai?";
+
+            TextAnalyticsClient client = GetClient();
+            TextAnalyticsRequestOptions options = new TextAnalyticsRequestOptions() { ModelVersion = "2022-04-10-preview" };
```
heaths, line 276: I'd add a `BUGBUG` comment to this tracking when this should be changed. Ideally we don't have to pin to such an old - even pre-GA (2022-05-01) - version.

sdk/textanalytics/Azure.AI.TextAnalytics/tests/DetectLanguageTests.cs
```diff
@@ -266,11 +266,27 @@ public void DetectLanguageBatchDisableServiceLogsThrows()
...
+
+            DetectLanguageResultCollection results = await client.DetectLanguageBatchAsync(new List<string>() { document }, options: options);
+
+            Assert.AreEqual(ScriptKind.Latin, results[0].PrimaryLanguage.Script);
```
heaths, line 280: Perhaps dubious value since we're testing the service, but given other issues we've seen perhaps pass in another non-Latin script and verify we get back "other".

heaths (approved):

sdk/textanalytics/Azure.AI.TextAnalytics/tests/DetectLanguageTests.cs
kinelski: Could you remind me why we need to pin the model version? Is it to make sure the behavior of the test won't change?

kinelski (approved):
LGTM!

sdk/textanalytics/Azure.AI.TextAnalytics/tests/DetectLanguageTests.cs
joseharriaga: @heaths: Good call. I created an issue and added a comment here.

@kinelski: The reason in this case is that currently model version 2022-04-10-preview is the only one that supports this feature, which is kind of unfortunate and a little awkward.

PR #32201, opened 2022-11-01
Exclude T1 Event Grid live tests

JoshLove-msft:


sdk/eventgrid/ci.yml
```diff
@@ -24,11 +22,9 @@ pr:
...
   paths:
     include:
-    - sdk/eventgrid/ci.yml
-    - sdk/eventgrid/Azure.Messaging.EventGrid/
-    - sdk/eventgrid/Microsoft.Azure.EventGrid/
-    - sdk/eventgrid/Microsoft.Azure.Messaging.EventGrid.CloudNativeCloudEvents/
-    - sdk/eventgrid/Microsoft.Azure.WebJobs.Extensions.EventGrid/
+      - sdk/eventgrid/
+    exclude:
+      - sdk/eventgrid/Azure.ResourceManager.EventGrid/
```
jsquire, line 27: Wouldn't we also need to explicitly exclude the T1 paths as well?

sdk/eventgrid/service.projects
```diff
@@ -19,7 +19,7 @@
 
   <ItemGroup Condition="'$(SDKType)' == 'client'">
     <ProjectsToIncludeBySDKType Include="$(MSBuildThisFileDirectory)Azure.Messaging.EventGrid\**\*.csproj" />
-    <ProjectsToIncludeBySDKType Include="$(MSBuildThisFileDirectory)Microsoft.Azure.EventGrid\**\*.csproj" />
+    <ProjectsToIncludeBySDKType Include="$(MSBuildThisFileDirectory)Microsoft.Azure.EventGrid\**\*.csproj" Condition="'$(AZURE_EVENTGRID_TEST_MODE)' != 'Live'" />
```
jsquire, line 22: Can we just split the pipeline and put the T1 package in its own?  Excluding only when it's a live run just seems to muddy things.

sdk/eventgrid/ci.yml
```diff
@@ -24,11 +22,9 @@ pr:
...
 
 extends:
   template: /eng/pipelines/templates/stages/archetype-sdk-client.yml
```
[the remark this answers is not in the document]
JoshLove-msft: No, T1 uses the same ci file.

sdk/eventgrid/service.projects
JoshLove-msft: The goal here was to revert to the state prior to this PR - https://github.com/Azure/azure-sdk-for-net/pull/31927

I can look into just excluding the T1 path directly in the tests.yml.

JoshLove-msft: Actually, I don't think there is a way to do it purely in tests.yml - the exclude condition controls what will trigger a pipeline run, not what a pipeline run will actually execute.

JoshLove-msft: @weshaggard if you have any suggestions on how to implement this in a better way. I'm trying to get the EG test pipeline passing again and reverting to the previous state of this file.

jsquire: I'd prefer that we split T1 and T2 into separate pipelines entirely.

jsquire (approved):

JoshLove-msft:
@hallipr would it be possibly to get the Event Grid test pipeline (tests.yml) added to the live test dashboard?

PR #32206, opened 2022-11-01
automation script to generate sdk from cadl

### LINKED ISSUE #4434: CADL DPG Integration in SDK Automation/Generation Pipeline - .Net
Please follow doc: https://gist.github.com/dw511214992/de785084657716b59e19a99970800544

chunyu3:

Fix https://github.com/Azure/azure-sdk-tools/issues/4434

### DIFF
--- a/eng/scripts/Invoke-GenerateAndBuildV2.ps1
+++ b/eng/scripts/Invoke-GenerateAndBuildV2.ps1
@@ -134,6 +134,8 @@ if ($relatedCadlProjectFolder) {
             $namespace = $directories[-2];
         }
         New-CADLPackageFolder -service $service -namespace $namespace -sdkPath $sdkPath -cadlInput $caldFolder/main.cadl -outputJsonFile $newpackageoutput
+        $newPackageOutputJson = Get-Content $newPackageOutput | Out-String | ConvertFrom-Json
+        $relativeSdkPath = $newPackageOutputJson.path
         # node $swaggerDir/node_modules/@cadl-lang/compiler/cmd/cadl.js compile --emit @azure-tools/cadl-csharp --output-path $sdkPath .\main.cadl
 
         # node $swaggerDir/node_modules/@cadl-lang/compiler/cmd/cadl.js compile --output-path $sdkPath --emit @azure-tools/cadl-csharp ./main.cadl
@@ -142,7 +144,7 @@ if ($relatedCadlProjectFolder) {
         if ( !$?) {
             Throw "Failed to generate sdk for cadl. exit code: $?"
         }
-        GeneratePackage -projectFolder $projectFolder -sdkRootPath $sdkPath -path $projectFolder -downloadUrlPrefix "$downloadUrlPrefix" -skipGenerate -generatedSDKPackages $generatedSDKPackages
+        GeneratePackage -projectFolder $projectFolder -sdkRootPath $sdkPath -path $relativeSdkPath -downloadUrlPrefix "$downloadUrlPrefix" -skipGenerate -generatedSDKPackages $generatedSDKPackages
         Pop-Location
     }
 }
--- a/eng/scripts/Invoke-GenerateAndBuildV2.ps1
+++ b/eng/scripts/Invoke-GenerateAndBuildV2.ps1
@@ -22,8 +22,8 @@ please refer to eng/scripts/automation/unified-pipeline-test.md for more test sc
 
 #>
 param (
-  [string]$inputJsonFile,
-  [string]$outputJsonFile
+  [string]$inputJsonFile="./eng/scripts/tests/firstOnboardInput.json",
+  [string]$outputJsonFile="output.json"
 )
 
 . (Join-Path $PSScriptRoot ".." "common" "scripts" "Helpers" PSModule-Helpers.ps1)
@@ -41,6 +41,7 @@ $commitid = $inputJson.headSha
 $repoHttpsUrl = $inputJson.repoHttpsUrl
 $downloadUrlPrefix = $inputJson.installInstructionInput.downloadUrlPrefix
 $autorestConfig = $inputJson.autorestConfig
+$relatedCadlProjectFolder = $inputJson.relatedCadlProjectFolder
 
 $autorestConfigYaml = ""
 if ($autorestConfig) {
@@ -108,6 +109,39 @@ if ($inputFileToGen) {
     UpdateExistingSDKByInputFiles -inputFilePaths $inputFileToGen -sdkRootPath $sdkPath -headSha $commitid -repoHttpsUrl $repoHttpsUrl -downloadUrlPrefix "$downloadUrlPrefix" -generatedSDKPackages $generatedSDKPackages
 }
 
+# generate sdk from cadl file
+if ($relatedCadlProjectFolder) {
+    Push-Location $swaggerDir
+    npm install @azure-tools/cadl-csharp
+    Pop-Location
+    for ($i = 0; $i -le $relatedCadlProjectFolder.Count - 1; $i++) {
+        $caldFolder = (Join-Path $swaggerDir $relatedCadlProjectFolder[$i]) -replace "\\", "/"
+        $newPackageOutput = "newPackageOutput.json"
+
+        Push-Location $caldFolder
+        $cadlProjectYaml = Get-Content -Path "$caldFolder/cadl-project.yaml" -Raw
+
+        Install-ModuleIfNotInstalled "powershell-yaml" "0.4.1" | Import-Module
+        $yml = ConvertFrom-YAML $cadlProjectYaml
+        $sdkFolder = $yml["emitters"]["@azure-tools/cadl-csharp"]["sdk-folder"]
+        $projectFolder = (Join-Path $sdkPath $sdkFolder)
+        $projectFolder = $projectFolder -replace "\\", "/"
+        if ($projectFolder) {
+            $directories = $projectFolder.Split("/");
+            $count = $directories.Count
+            $projectFolder = $directories[0 .. ($count-2)] -join "/"
+            $service = $directories[-3];
+            $namespace = $directories[-2];
+        }
+        New-CADLPackageFolder -service $service -namespace $namespace -sdkPath $sdkPath -cadlInput $caldFolder/main.cadl -outputJsonFile $newpackageoutput
+        # node $swaggerDir/node_modules/@cadl-lang/compiler/cmd/cadl.js compile --emit @azure-tools/cadl-csharp --output-path $sdkPath .\main.cadl
+
+        node $swaggerDir/node_modules/@cadl-lang/compiler/cmd/cadl.js compile --output-path $sdkPath --emit @azure-tools/cadl-csharp ./main.cadl
+
+        GeneratePackage -projectFolder $projectFolder -sdkRootPath $sdkPath -path $projectFolder -downloadUrlPrefix "$downloadUrlPrefix" -skipGenerate -generatedSDKPackages $generatedSDKPackages
+        Pop-Location
+    }
+}
 $outputJson = [PSCustomObject]@{
     packages = $generatedSDKPackages
 }
--- a/eng/scripts/Invoke-GenerateAndBuildV2.ps1
+++ b/eng/scripts/Invoke-GenerateAndBuildV2.ps1
@@ -137,7 +137,8 @@ if ($relatedCadlProjectFolder) {
         # node $swaggerDir/node_modules/@cadl-lang/compiler/cmd/cadl.js compile --emit @azure-tools/cadl-csharp --output-path $sdkPath .\main.cadl
 
         # node $swaggerDir/node_modules/@cadl-lang/compiler/cmd/cadl.js compile --output-path $sdkPath --emit @azure-tools/cadl-csharp ./main.cadl
-        npx cadl compile --output-path $sdkPath --emit @azure-tools/cadl-csharp ./main.cadl
+        npm install
+        npx cadl compile --output-path $sdkPath --emit @azure-tools/cadl-csharp .
         GeneratePackage -projectFolder $projectFolder -sdkRootPath $sdkPath -path $projectFolder -downloadUrlPrefix "$downloadUrlPrefix" -skipGenerate -generatedSDKPackages $generatedSDKPackages
         Pop-Location
     }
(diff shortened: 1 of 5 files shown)


eng/scripts/Invoke-GenerateAndBuildV2.ps1
```diff
@@ -108,6 +109,39 @@ if ($inputFileToGen) {
     UpdateExistingSDKByInputFiles -inputFilePaths $inputFileToGen -sdkRootPath $sdkPath -headSha $commitid -repoHttpsUrl $repoHttpsUrl -downloadUrlPrefix "$downloadUrlPrefix" -generatedSDKPackages $generatedSDKPackages
 }
 
+# generate sdk from cadl file
+if ($relatedCadlProjectFolder) {
+    for ($i = 0; $i -le $relatedCadlProjectFolder.Count - 1; $i++) {
+        $caldFolder = (Join-Path $swaggerDir $relatedCadlProjectFolder[$i]) -replace "\\", "/"
```
heaths, line 115: Why not just use a foreach. for statements in PowerShell are often not the best choice:

```powershell
foreach ($cadlFolder in $relatedCadlProjectFolder) {
  $cadlFolder = join-path $swaggerDir $cadlFolder)
```

Also, if you're having to replace backslashes with forward slashes to satisfy the tool, I would argue that's a bug in the tool. It should normalize and canonicalize folders itself as most tools do. Using it outside this script means it'll be a problem for people to type `{tools} foo{tab}` and use tab complete to complete directory/file names. That's a productivity issue.

eng/scripts/Invoke-GenerateAndBuildV2.ps1
```diff
@@ -108,6 +109,39 @@ if ($inputFileToGen) {
...
+        New-CADLPackageFolder -service $service -namespace $namespace -sdkPath $sdkPath -cadlInput $caldFolder/main.cadl -outputJsonFile $newpackageoutput
+        $newPackageOutputJson = Get-Content $newPackageOutput | Out-String | ConvertFrom-Json
+        $relativeSdkPath = $newPackageOutputJson.path
+        npm install
+        npx cadl compile --output-path $sdkPath --emit @azure-tools/cadl-csharp .
+        if ( !$?) {
+            Throw "Failed to generate sdk for cadl. exit code: $?"
+        }
+        GeneratePackage -projectFolder $projectFolder -sdkRootPath $sdkPath -path $relativeSdkPath -downloadUrlPrefix "$downloadUrlPrefix" -skipGenerate -generatedSDKPackages $generatedSDKPackages
+        Pop-Location
```
heaths, line 142: If anything fails above and `$ErrorActionPreference` is `Stop`, you'll not pop the location. I recommend against this pattern. Use the full directories, and if tools can't handle that they need to be fixed. For example, I can run `msbuild` or `dotnet build` on a .csproj in any directory from any other directory and it works. Tools should do this. If `npm` is the reason above, it takes full paths and output directories you can set. If that's not otherwise reasonable, then on line 118 you should do something like:
```powershell
push-location $cadlFolder
trap { pop-location }
```
This will make sure that any error pops the location regardless of where it happens.

eng/scripts/Invoke-GenerateAndBuildV2.ps1
heaths, line 115: You also misspelled `cadlFolder`. I recommend fixing the mistake so someone later doesn't spell it correctly, but break the script.

eng/scripts/Invoke-GenerateAndBuildV2.ps1
```diff
@@ -108,6 +109,39 @@ if ($inputFileToGen) {
...
+            $directories = $projectFolder.Split("/");
+            $count = $directories.Count
+            $projectFolder = $directories[0 .. ($count-2)] -join "/"
+            $service = $directories[-3];
+            $namespace = $directories[-2];
+        }
+        New-CADLPackageFolder -service $service -namespace $namespace -sdkPath $sdkPath -cadlInput $caldFolder/main.cadl -outputJsonFile $newpackageoutput
+        $newPackageOutputJson = Get-Content $newPackageOutput | Out-String | ConvertFrom-Json
+        $relativeSdkPath = $newPackageOutputJson.path
+        npm install
```
heaths, line 136: Is there a reason you run `npm install` before `npx`? If it's just for cadl tools, it's unnecessary. `npx` will fetch modules as needed and run the specified package's entry point. If it's only for cadl, I imagine `npm install` will install much more than needed as well.

eng/scripts/Invoke-GenerateAndBuildV2.ps1
```diff
@@ -108,6 +109,39 @@ if ($inputFileToGen) {
...
+        $projectFolder = $projectFolder -replace "\\", "/"
+        if ($projectFolder) {
+            $directories = $projectFolder.Split("/");
+            $count = $directories.Count
+            $projectFolder = $directories[0 .. ($count-2)] -join "/"
+            $service = $directories[-3];
+            $namespace = $directories[-2];
+        }
+        New-CADLPackageFolder -service $service -namespace $namespace -sdkPath $sdkPath -cadlInput $caldFolder/main.cadl -outputJsonFile $newpackageoutput
+        $newPackageOutputJson = Get-Content $newPackageOutput | Out-String | ConvertFrom-Json
```
heaths, line 134: The `out-string` is unnecessary. The content is already a string and `convertfrom-json` handles both a string and string array correctly (unless you're on *really* older Windows PowerShell, which you shouldn't be using...it's effectively defunct.

eng/scripts/automation/GenerateAndBuildLib.ps1
```diff
@@ -357,6 +357,100 @@ function New-MgmtPackageFolder() {
     return $projectFolder
 }
 
+function New-CADLPackageFolder() {
+    param(
+        [string]$service,
+        [string]$namespace,
+        [string]$sdkPath = "",
+        [string]$cadlInput ="",
+        [string]$outputJsonFile = "output.json"
```
heaths, line 366: This should be a full path. You could use `"$PWD\output.json"`, for example. Don't require pushing and popping paths as I noted above. It's fragile.

eng/scripts/automation/GenerateAndBuildLib.ps1
```diff
@@ -357,6 +357,100 @@ function New-MgmtPackageFolder() {
...
+        Invoke-Expression $dotnetNewCmd
+
+        $projFile = (Join-Path $projectFolder "src" "$namespace.csproj")
+        $fileContent = Get-Content -Path $projFile
+        $fileContent = $fileContent -replace "<Version>*.*.*-*.*</Version>", "<Version>1.0.0-beta.1</Version>"
+        $startNum = ($fileContent | Select-String -Pattern '</PropertyGroup>').LineNumber[0]
+        $fileContent[$startNum - 2] += ([Environment]::NewLine + "<AutoRestInput>$cadlInput</AutoRestInput>")
+        $fileContent | Out-File $projFile
+        # (Get-Content $projFile) -replace "<Version>*.*.*-*.*</Version>", "<Version>1.0.0-beta.1</Version>" | -replace "<AutoRestInput>*</AutoRestInput>", "<AutoRestInput>$cadlInput</AutoRestInput>" |Set-Content $projFile
+        Pop-Location
```
heaths, line 426: As noted above, don't use this pattern. Errors above will leave the script in the previously pushed location. Use paths everywhere, fix the tool to handle the current working directory correctly (generally good anyway), or at the very least use the `trap { pop-location }` pattern I mentioned above.

eng/scripts/automation/GenerateAndBuildLib.ps1
```diff
@@ -357,6 +357,100 @@ function New-MgmtPackageFolder() {
...
+
+            $dotnetNewCmd = $dotnetNewCmd + " --includeCI false"
+        }
+        # dotnet new azsdkdpg --name $namespace --clientName $clientName --groupName $groupName --serviceDirectory $service --swagger $inputfile --securityScopes $securityScope --securityHeaderName $securityHeaderName --includeCI true --force
+        Write-Host "Invoke dotnet new command: $dotnetNewCmd"
+        Invoke-Expression $dotnetNewCmd
+
+        $projFile = (Join-Path $projectFolder "src" "$namespace.csproj")
+        $fileContent = Get-Content -Path $projFile
+        $fileContent = $fileContent -replace "<Version>*.*.*-*.*</Version>", "<Version>1.0.0-beta.1</Version>"
```
heaths, line 421: This might work, but is not safe. This is not a proper regex, which is what is expected here. Instead, probably this would be safer (and generally use single quotes in a literal string, especially when regex is involved):
```suggestion
        $fileContent = $fileContent -replace '<Version>[^<]+</Version>', '<Version>1.0.0-beta.1</Version>'
```
This would work even if the XML content were compressed (all on one line) where yours may not since it's greedy: `.*` in regex will match *everything* until the very last `</Version>` which, if there's more than one in the file, would replace *far* more content than you want.

eng/scripts/automation/Invoke-GenerateAndBuild.ps1
```diff
@@ -41,27 +42,59 @@ if ($autorestConfig -ne "") {
...
-    $readme = "$repoHttpsUrl/blob/$commitid/$readmeFile"
+if ($readmeFile) {
+  Write-Host "swaggerDir:$swaggerDir, readmeFile:$readmeFile"
+
+  $readme = ""
+  if ($commitid -ne "") {
+    if ($repoHttpsUrl -ne "") {
+      $readme = "$repoHttpsUrl/blob/$commitid/$readmeFile"
+    } else {
+      $readme = "https://github.com/$org/azure-rest-api-specs/blob/$commitid/$readmeFile"
```
heaths, line 60: Is this not intended to support cadl in the private repo i.e., Azure/azure-rest-api-specs-pr?

Also, what other org is there? Seems like an odd variable to have, or is this meant to also support forks?

eng/scripts/automation/Invoke-GenerateAndBuild.ps1
```diff
@@ -41,27 +42,59 @@ if ($autorestConfig -ne "") {
...
+    }
   } else {
-    $readme = "https://github.com/$org/azure-rest-api-specs/blob/$commitid/$readmeFile"
+    $readme = (Join-Path $swaggerDir $readmeFile)
   }
-} else {
-  $readme = (Join-Path $swaggerDir $readmeFile)
+  Invoke-GenerateAndBuildSDK -readmeAbsolutePath $readme -sdkRootPath $sdkPath -autorestConfigYaml "$autorestConfigYaml" -downloadUrlPrefix "$downloadUrlPrefix" -generatedSDKPackages $generatedSDKPackages
```
heaths, line 65: Nit: you don't need to quote variables when they are the only parameter value.

eng/scripts/automation/Invoke-GenerateAndBuild.ps1
```diff
@@ -41,27 +42,59 @@ if ($autorestConfig -ne "") {
...
+  New-CADLPackageFolder -service $service -namespace $namespace -sdkPath $sdkPath -cadlInput $caldFolder/main.cadl -outputJsonFile $newpackageoutput
+  $newPackageOutputJson = Get-Content $newPackageOutput | Out-String | ConvertFrom-Json
+  $relativeSdkPath = $newPackageOutputJson.path
+  npm install
+  npx cadl compile --output-path $sdkPath --emit @azure-tools/cadl-csharp .
+  if ( !$?) {
+      Throw "Failed to generate sdk for cadl. exit code: $?"
+  }
+  GeneratePackage -projectFolder $projectFolder -sdkRootPath $sdkPath -path $relativeSdkPath -downloadUrlPrefix "$downloadUrlPrefix" -skipGenerate -generatedSDKPackages $generatedSDKPackages
+  Pop-Location
```
heaths, line 96: Another instance where you should either eliminate this pattern, or put `trap { pop-location }` *immediately* after `push-location`.

heaths (changes requested):

eng/scripts/Invoke-GenerateAndBuildV2.ps1
chunyu3: `npx cadl compile` only to fetch cadl compile (@cadl-liang/compiler) modules, but each cadl project will depend on a group of modules and may different for each cadl project. There is an package.json in the cadl project, we need to `npm install` to install all the depend modules this cadl project needed.

eng/scripts/Invoke-GenerateAndBuildV2.ps1
chunyu3: Thanks @heaths  we need to install all the modules defined in `package.json`, but `npm install` cannot specify the file, so we need to enter the directory which contains the `package.json`. 
I will use `trap` to make sure `pop-location` will be executed when script is exit in unnormal way.

eng/scripts/Invoke-GenerateAndBuildV2.ps1
chunyu3: Thanks @heaths updated

eng/scripts/Invoke-GenerateAndBuildV2.ps1
chunyu3: resolved

eng/scripts/automation/GenerateAndBuildLib.ps1
chunyu3: uddate

eng/scripts/automation/Invoke-GenerateAndBuild.ps1
chunyu3: Current `cadl compile` can only accept file-path, it cannot load cadl file via url. So we need to clone the swagger repo locally .

eng/scripts/automation/GenerateAndBuildLib.ps1
chunyu3: Thanks @heaths . I updated.

eng/scripts/automation/Invoke-GenerateAndBuild.ps1
chunyu3: Thanks @heaths . I removed the unused quote.

eng/scripts/automation/GenerateAndBuildLib.ps1
chunyu3: resolved

heaths (approved):

PR #32207, opened 2022-11-01
Remove LastException from ProcessingContext

JoshLove-msft:


christothes (approved):

sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -1024,8 +1023,8 @@ public abstract partial class RetryPolicy : Azure.Core.Pipeline.HttpPipelinePoli
         protected internal virtual System.Threading.Tasks.ValueTask OnSendingRequestAsync(Azure.Core.HttpMessage message) { throw null; }
         public override void Process(Azure.Core.HttpMessage message, System.ReadOnlyMemory<Azure.Core.Pipeline.HttpPipelinePolicy> pipeline) { }
         public override System.Threading.Tasks.ValueTask ProcessAsync(Azure.Core.HttpMessage message, System.ReadOnlyMemory<Azure.Core.Pipeline.HttpPipelinePolicy> pipeline) { throw null; }
-        protected internal virtual bool ShouldRetry(Azure.Core.HttpMessage message) { throw null; }
-        protected internal virtual System.Threading.Tasks.ValueTask<bool> ShouldRetryAsync(Azure.Core.HttpMessage message) { throw null; }
+        protected internal virtual bool ShouldRetry(Azure.Core.HttpMessage message, System.Exception? exception) { throw null; }
```
annelo-msft, line 1026: Would it make sense to provide a default for `exception`?

JoshLove-msft: I think it is probably better for it to be a required parameter. If customers override ShouldRetry and call into the base method, we wouldn't want them to forego passing the exception.

sdk/core/Azure.Core/src/HttpMessage.cs
```diff
@@ -87,10 +87,6 @@ public Response Response
 
         internal int RetryNumber { get; set; }
 
-        internal int ResponseNumber { get; set; }
```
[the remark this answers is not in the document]
JoshLove-msft: This was left over from the initial PR - it is no longer used anywhere.

PR #32208, opened 2022-11-01
Changes for updating SDK to new preview version and AlwaysServe flag.

imsiddhartha:


sdk/trafficmanager/Azure.ResourceManager.TrafficManager/src/Generated/RestOperations/EndpointsRestOperations.cs
```diff
@@ -32,11 +33,11 @@ public EndpointsRestOperations(HttpPipeline pipeline, string applicationId, Uri
         {
             _pipeline = pipeline ?? throw new ArgumentNullException(nameof(pipeline));
             _endpoint = endpoint ?? new Uri("https://management.azure.com");
-            _apiVersion = apiVersion ?? "2018-08-01";
+            _apiVersion = apiVersion ?? "2022-04-01-preview";
             _userAgent = new TelemetryDetails(GetType().Assembly, applicationId);
         }
 
-        internal HttpMessage CreateUpdateRequest(string subscriptionId, string resourceGroupName, string profileName, string endpointType, string endpointName, TrafficManagerEndpointData data)
+        internal HttpMessage CreateUpdateRequest(string subscriptionId, string resourceGroupName, string profileName, EndpointType endpointType, string endpointName, TrafficManagerEndpointData data)
```
m-nash, line 40: This is a breaking change, can you paste the spec PR that caused this change?

m-nash, line 40: This is a breaking change, can you paste the spec PR that caused this change?

m-nash (changes requested):

sdk/trafficmanager/Azure.ResourceManager.TrafficManager/src/Generated/RestOperations/EndpointsRestOperations.cs
imsiddhartha: Spec PR: https://github.com/Azure/azure-rest-api-specs/pull/21308 

It was there in the earlier version as well https://github.com/Azure/azure-rest-api-specs/blob/main/specification/trafficmanager/resource-manager/Microsoft.Network/stable/2018-08-01/trafficmanager.json#L41

Nothing really changed in this version: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/trafficmanager/resource-manager/Microsoft.Network/preview/2022-04-01-preview/trafficmanager.json#L37

imsiddhartha: Removed breaking changes

eng/mgmt/mgmtmetadata/trafficmanager_resource-manager.txt
```diff
@@ -5,7 +5,7 @@ Generating CSharp code
 Executing AutoRest command
 cmd.exe /c autorest.cmd https://github.com/Azure/azure-rest-api-specs/blob/master/specification/trafficmanager/resource-manager/readme.md --csharp --version=v2 --reflect-api-versions --opt-in-extensible-enums --csharp-sdks-folder=D:\GitHub\Repos\AllenCalMinChild\azure-sdk-for-net\sdk
 Autorest CSharp Version: 2.3.82
-2021-05-07 03:07:52 UTC
+2022-11-04 03:07:52 UTC
```
Yao725, line 8: Please revert this change as we don't need it.

imsiddhartha:
There were some pipeline failures while generating .NET SDK for TM, thanks @Yao725 for the detailed analysis:

 

Earlier when the stable version 1.0.0 was released, this [commit](https://github.com/Azure/azure-rest-api-specs/blob/7384176da46425e7899708f263e0598b851358c2/specification/trafficmanager/resource-manager/readme.md) was used as the source readme file from swagger repo, which means generator script will use the API file with same commit id, and you can see at that time the [endpointType](https://github.com/Azure/azure-rest-api-specs/blob/7384176da46425e7899708f263e0598b851358c2/specification/trafficmanager/resource-manager/Microsoft.Network/stable/2018-08-01/trafficmanager.json#L41) is just a string. 

 

And if you navigate to the latest API file, it becomes the [enum](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/trafficmanager/resource-manager/Microsoft.Network/stable/2018-08-01/trafficmanager.json#L41). This is the root cause of this change, if you changed a path parameter to an enum, SDK generator will automatically split them to a new resource. 

 

For instance, previously we have a resource called TrafficManagerEndpoint, after the endpoint type explicitly marked as enum, generator will use three new resources(aka TrafficManagerAzureEndpoint, TrafficManagerExternalEndpoint and TrafficManagerNestedEndpoint) to replace the old resource. So we have the breaking changes.

 

This is the PR responsible for the errors: [trafficmanager: adding the missing enum values for traffic manager endpoint by tombuildsstuff · Pull Request #16472 · Azure/azure-rest-api-specs (github.com)](https://github.com/Azure/azure-rest-api-specs/pull/16472/files)

I wonder why .NET SDK was not updated after this PR and was this accidental or it was supposed to be like that

 

 

To minimize the breaking changes and keep backward compatibility, we removed the enum definition in the autorest.md configuration and added the supported endpoint type to the parameter description, please take a look at [this commit](https://github.com/Azure/azure-sdk-for-net/pull/32208/commits/5cbd3ca2eaef43a8a8e680b4aca86da2d31f97de) and let me know if you are ok with this change.

Yao725 (approved):

sdk/trafficmanager/Azure.ResourceManager.TrafficManager/src/Generated/RestOperations/EndpointsRestOperations.cs
Yao725: Hi @m-nash, the breaking changes has been removed, could you please take a look?

m-nash (approved):

PR #32213, opened 2022-11-01
[Event Hubs] Generate SAS example

jsquire:
# Summary

The focus of these changes is to add an example of generating a SAS token using C# to the authorization sample.


christothes (approved):

jsquire:
Failure is due to time outs when provisioning a macOS host in the pipeline.   Merging, as this is sample code.

PR #32221, opened 2022-11-01
Add new library Azure.Core.Expressions.DataFactory

m-nash:
Implements this feature https://github.com/Azure/azure-sdk/issues/4964

### DIFF
--- a/sdk/core/Azure.Core.Expressions.DataFactory/api/Azure.Core.Expressions.DataFactory.net461.cs
+++ b/sdk/core/Azure.Core.Expressions.DataFactory/api/Azure.Core.Expressions.DataFactory.net461.cs
@@ -1,0 +1,16 @@
+namespace Azure.Core.Expressions.DataFactory
+{
+    public sealed partial class DataFactoryExpression<T>
+    {
+        public DataFactoryExpression(T literal) { }
+        public bool HasLiteral { get { throw null; } }
+        public T? Literal { get { throw null; } }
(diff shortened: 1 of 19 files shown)


sdk/core/Azure.Core.Expressions.DataFactory/api/Azure.Core.Expressions.DataFactory.net461.cs
```diff
@@ -0,0 +1,16 @@
+namespace Azure.Core.Expressions.DataFactory
+{
+    public sealed partial class DataFactoryExpression<T>
+    {
+        public DataFactoryExpression(T literal) { }
+        public bool HasLiteral { get { throw null; } }
+        public T? Literal { get { throw null; } }
+        [System.ComponentModel.EditorBrowsableAttribute(System.ComponentModel.EditorBrowsableState.Never)]
+        public override bool Equals(object? obj) { throw null; }
+        public static Azure.Core.Expressions.DataFactory.DataFactoryExpression<T> FromExpression(string expression) { throw null; }
```
JoshLove-msft, line 10: This is necessary to disambiguate an expression string from a primitive when dealing with an `Expression<string>`. Unfortunately, it leads us to have a constructor for literals, and a factory method for expressions. An alternative would be to just have two factory methods and no constructors, but not sure if that improves things, especially if the literal is the more common scenario.

m-nash:
latest commit LGTM

JoshLove-msft (approved):

PR #32229, opened 2022-11-01
[Event Hubs] AMQP logging improvements

jsquire:
# Summary

The focus of these changes is to capture the total duration that receive and publish operations took when logging the "Complete" status.


JoshLove-msft (approved):
Are there any tests we can update?

jsquire:
> Are there any tests we can update?

No; we currently don't have tests for logging in the lower-level clients.   There hasn't been justification to invest the time to add them up to this point.

PR #32230, opened 2022-11-01
[EngSys] Moq version bump

jsquire:
# Summary

The focus of these changes is to bump the version of Moq referenced by tests.  The previous version was from 2018, and there have been many enhancements since.  Several of our tests which mock ETW sources occasionally fail with an `IndexOutOfRange` exception related to the event source, the root cause of which has not been found.  My hope is that using a newer Moq version may resolve the issue.

### NOTE

If there are specific test pipelines that you would like to validate against, please feel free to kick off a run or mention them in a comment.

# References and Resources

- [Moq Changelog](https://github.com/moq/moq4/blob/main/CHANGELOG.md)
- [Sample Moq + ETW failure](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1958029&view=ms.vss-test-web.build-test-results-tab&runId=36512202&resultId=100654&paneView=debug) _(Microsoft internal)_

### DIFF
--- a/eng/Packages.Data.props
+++ b/eng/Packages.Data.props
@@ -239,7 +239,7 @@
     <PackageReference Update="Microsoft.Rest.ClientRuntime.Azure.TestFramework" Version="[1.7.7, 2.0.0)" />
     <PackageReference Update="Microsoft.ServiceFabric.Data" Version="3.3.624" />
     <PackageReference Update="Microsoft.Spatial" Version="7.5.3" />
-    <PackageReference Update="Moq" Version="4.10.1" />
+    <PackageReference Update="Moq" Version="4.18.2" />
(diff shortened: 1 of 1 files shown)


heaths (approved):
With as much as [changed](https://github.com/moq/moq4/blob/main/CHANGELOG.md) since then, should we run a full gamut of playback tests? Don't know how much use this is getting. If it's just what you're testing now, 👌.

JoshLove-msft:
> With as much as [changed](https://github.com/moq/moq4/blob/main/CHANGELOG.md) since then, should we run a full gamut of playback tests? Don't know how much use this is getting. If it's just what you're testing now, 👌.

This should trigger Core CI

heaths (approved):

benbp (approved):

Issue #32260, opened 2022-11-03
Managed HSM is returning 400 BadParameter for release policy version


heaths:
For several tests where 7.4-preview.1 support was added, anything with a release policy that does define a version (as they did before) is returning:

```
Status: 400 (Bad Request)
ErrorCode: BadParameter
Content:
{"error":{"code":"BadParameter","message":"Key Release Policy: missing 'version' property (Activity ID: 31b04ff2-5b0a-11ed-94f7-0022488df0c4)"}}
```

Re-running these tests on 7.3 works. For now, I'll add a `Max` version for live tests and have a thread going with the service team.

heaths:
This no longer appears to be a problem. `Max` was removed from tests and they pass just fine with 7.4.

PR #32261, opened 2022-11-03
Add OKP support to Key Vault 

### LINKED ISSUE #26193: [KeyVault] - Add support for OKP key type
In order to support OKP key types and EdDSA algorithm, the following changes need to be made:

1. A new Key Type is added - OKP and OKP-HSM
2. A new Key Curve is added - Ed25519
3. Sign and verify using the EdDSA algorithm is supported by cryptography client
4. SignData and VerifyData using the EdDSA algorithm is supported by cryptography client using sha256 for hashing. We should probably document that we use sha256 somewhere? 

In addition, we should add a new `createOkpKey` convenience method similar to `createRsaKey` (with similar support for the `hsm` option)

Acceptance criteria:
1. `createKey` supports creating OKP / OKP-HSM keys 
2. `createOkpKey` is added to KeyClient and can be used to create OKP / OKP-HSM keys (controlled by the `hsm` option)
3. Cryptography client can be used to sign and verify using the OKP key type and the EdDSA algorithm (both sign/verify and signData/verifyData)

heaths:
Resolves #26193


heaths:
Note to reviewers: if you [ignore the JSON recordings](https://github.com/Azure/azure-sdk-for-net/pull/32261/files?file-filters%5B%5D=.cs&file-filters%5B%5D=.md&show-viewed-files=true), it's not so bad.

sdk/keyvault/Azure.Security.KeyVault.Keys/src/Cryptography/SignatureAlgorithm.cs
```diff
@@ -35,55 +36,60 @@ public SignatureAlgorithm(string value)
...
         /// <summary>
-        /// Gets an ECDSA with a secp256k1 curve <see cref="SignatureAlgorithm"/>.
+        /// Gets an ECDSA with a secp256k1 curve <see cref="SignatureAlgorithm"/> as described in <see href="https://tools.ietf.org/html/rfc7518"/>.
         /// </summary>
         public static SignatureAlgorithm ES256K { get; } = new SignatureAlgorithm(ES256KValue);
 
+        /// <summary>
+        /// Gets an ECDSA with a secp256k1 curve <see cref="SignatureAlgorithm"/> as described in <see href="https://tools.ietf.org/html/rfc8032"/>.
+        /// </summary>
+        public static SignatureAlgorithm EdDSA { get; } = new SignatureAlgorithm(EdDSAValue);
```
heaths, line 91: @KrzysztofCwalina, @tg-msft: I realize this is a departure from our typically casing, but prior you allowed `ToRSA` and `ToECDsa` because it matched the .NET types. There is a (long-standing) ask to add EdDSA (as its cased in the industry) which I imagine would follow a similar naming convention. As such, is this fine as "EdDSA" or should it be "EdDsa"?

For context, it stands for "Edwards-curve Digital Signature Algorithm".

heaths:
This adds a couple of workarounds for #32260, which is tracking removal of the workaround once resolved by the service.

sdk/keyvault/Azure.Security.KeyVault.Keys/src/Cryptography/SignatureAlgorithm.cs
```diff
@@ -35,55 +36,60 @@ public SignatureAlgorithm(string value)
         }
 
         /// <summary>
-        /// Gets an RSA SHA-256 <see cref="SignatureAlgorithm"/>.
+        /// Gets an RSA SHA-256 <see cref="SignatureAlgorithm"/> as described in <see href="https://tools.ietf.org/html/rfc7518"/>.
```
christothes, line 39: should we make this an aka.ms link?

christothes (approved):

sdk/keyvault/Azure.Security.KeyVault.Keys/src/Cryptography/SignatureAlgorithm.cs
KrzysztofCwalina: I am fine with EdDSA for the same reason: this is how .NET cases it.

sdk/keyvault/Azure.Security.KeyVault.Keys/src/Cryptography/SignatureAlgorithm.cs
heaths: I don't think that's necessary. These IETF links have been this way for as long as I can remember, and I'm old.

PR #32266, opened 2022-11-03
[Text Analytics] Use dynamically-created test resources, except when testing custom text features

### LINKED ISSUE #32265: [Text Analytics] Use dynamically-created test resources, except when testing custom text features
The dynamically-created test resource must also be in a region that supports the new features of API version 2022-10-01-preview.

joseharriaga:
Resolves: #32265


sdk/textanalytics/test-resources.json
```diff
@@ -35,18 +35,13 @@
         },
         "location": {
             "type": "string",
-            "defaultValue": "[resourceGroup().location]",
+            "defaultValue": "eastus",
```
deyaaeldeen, line 38: could this be controlled using a parameter in `tests.yml` instead? similar to [this](https://github.com/Azure/azure-sdk-for-js/blob/ceb5483caf27456c05c6ecb11ca5dcbd9ce339aa/sdk/cognitivelanguage/ai-language-text/tests.yml#L1-L5)

sdk/textanalytics/test-resources.json
```diff
@@ -81,22 +78,41 @@
...
         "TEXT_ANALYTICS_API_KEY": {
             "type": "string",
-            "value": "[listKeys(resourceId('Microsoft.CognitiveServices/accounts',variables('uniqueSubDomainName')), '2017-04-18').key1]"
+            "value": "[listKeys(resourceId('Microsoft.CognitiveServices/accounts', variables('uniqueSubDomainName')), variables('cognitiveServicesApiVersion')).key1]"
         },
         "TEXT_ANALYTICS_ENDPOINT": {
             "type": "string",
             "value": "[variables('endpointValue')]"
+        },
+        "TEXTANALYTICS_ENDPOINT_SUFFIX": {
```
deyaaeldeen, line 113: what is this needed for?

sdk/textanalytics/Azure.AI.TextAnalytics/tests/Infrastructure/TextAnalyticsClientLiveTestBase.cs
```diff
@@ -54,7 +57,9 @@ protected TextAnalyticsClient GetClient(AzureKeyCredential credential = default,
             }
             else
             {
-                credential ??= new AzureKeyCredential(TestEnvironment.ApiKey);
+                credential ??= useStaticResource
+                    ? new AzureKeyCredential(TestEnvironment.StaticApiKey)
+                    : new AzureKeyCredential(TestEnvironment.ApiKey);
```
deyaaeldeen, line 62: nitpick: could the conditional be moved inside the `AzureKeyCredential` constructor?

joseharriaga: The `AzureKeyCredential` class and its constructor are part of Azure Core. Since this dual need for static and dynamic test resources is very particular to Text Analytics, I was thinking that it is more appropriate to have this logic in the Text Analytics test infra.

sdk/textanalytics/test-resources.json
joseharriaga: Awesome. Thanks!

sdk/textanalytics/Azure.AI.TextAnalytics/tests/Infrastructure/TextAnalyticsClientLiveTestBase.cs
deyaaeldeen: ops sorry for being unclear, I meant moved inside the constructor _call_, e.g. 
```csharp
credential ??= new AzureKeyCredential(useStaticResource ? TestEnvironment.StaticApiKey : TestEnvironment.ApiKey);
```

deyaaeldeen (approved):
Looks good, I defer my approval to @heaths

sdk/textanalytics/tests.yml
```diff
@@ -1,9 +1,16 @@
...
+  type: string
+  default: eastus
+
 trigger: none
 
 extends:
   template: ../../eng/pipelines/templates/stages/archetype-sdk-tests.yml
   parameters:
     ServiceDirectory: textanalytics
+    Location: "${{ parameters.Location }}"
```
benbp, line 13: FYI if you set this at the top level, the sovereign clouds (usgov/china) will not work because `eastus` is not a valid location for those clouds. I would nest `Location` under the `Public` object on line 16 instead.

sdk/textanalytics/test-resources.json
```diff
@@ -81,22 +78,37 @@
...
+        },
         "TEXT_ANALYTICS_API_KEY": {
             "type": "string",
-            "value": "[listKeys(resourceId('Microsoft.CognitiveServices/accounts',variables('uniqueSubDomainName')), '2017-04-18').key1]"
+            "value": "[listKeys(resourceId('Microsoft.CognitiveServices/accounts', variables('uniqueSubDomainName')), variables('cognitiveServicesApiVersion')).key1]"
         },
         "TEXT_ANALYTICS_ENDPOINT": {
             "type": "string",
             "value": "[variables('endpointValue')]"
         }
```
heaths, line 112: Any automatically-added variables will actually use "TEXTANALYTICS" (no underscore) so if you changed these to follow, they will be consistent. In fact, at that point you don't even need the prefix since we look it up by default in `TestEnvironment` based on the service directory, "textanalytics". You could just define `API_KEY` and `ENDPOINT`. That said, I typically like the prefixes explicit because I tend to define long-lived apps as user env vars on my system. Just a tip.

sdk/textanalytics/Azure.AI.TextAnalytics/tests/Infrastructure/TextAnalyticsTestEnvironment.cs
```diff
@@ -9,6 +9,11 @@ public class TextAnalyticsTestEnvironment: TestEnvironment
     {
         public string Endpoint => GetRecordedVariable("TEXT_ANALYTICS_ENDPOINT");
         public string ApiKey => GetRecordedVariable("TEXT_ANALYTICS_API_KEY", options => options.IsSecret());
+
+        // The following variables refer to the static test resource, which is
+        // used to test the custom text analysis features.
+        public string StaticEndpoint => GetRecordedVariable("TEXT_ANALYTICS_STATIC_ENDPOINT");
+        public string StaticApiKey => GetRecordedVariable("TEXT_ANALYTICS_STATIC_API_KEY", options => options.IsSecret());
```
heaths, line 16: I commented below as well, but you might - for consistency and ease of look-up (see how we use the service directory e.g., "textanalytics", in `TestEnvironment`) - use "TEXTANALYTICS" as a consistent prefix.

heaths (approved):

heaths:
Also, did you share this with @deyaaeldeen, @kristapratico, and @mssfang? Ideally, everyone has a similar copy of _test-resources.json_ (and associated files as needed) so you're not all having to maintain completely separate definitions. Maybe for these resources it's not a huge win, but it's definitely been helpful for CLU, QA, and KV.

kinelski (approved):
LGTM!

mssfang:
PR looks good!

I experienced these transient failures as well when I ran the live tests in java without enabling dynamic resource allocation. I think you can try with rerun the failed test a couple of times( < 2 times) it might resolve the issue.

mssfang (approved):

kristapratico (approved):

joseharriaga:
Thank you, @mssfang !

PR #32280, opened 2022-11-03
Use HasPath/HasQuery instead of PathLength/QueryLength

AlexanderSher:
Also add .net 6.0 to the list of target frameworks


sdk/core/Azure.Core/src/Azure.Core.csproj
```diff
@@ -8,7 +8,9 @@
     <PackageTags>Microsoft Azure Client Pipeline</PackageTags>
     <Nullable>enable</Nullable>
     <DefineConstants>$(DefineConstants);AZURE_NULLABLE;HAS_INTERNALS_VISIBLE_CORE</DefineConstants>
-    <TargetFrameworks>$(RequiredTargetFrameworks);net461;netcoreapp2.1;net5.0</TargetFrameworks>
+    <TargetFrameworks>$(RequiredTargetFrameworks);net461;netcoreapp2.1;net5.0;net6.0</TargetFrameworks>
```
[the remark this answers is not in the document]
AlexanderSher: Should we remove `netcoreapp2.1` from it?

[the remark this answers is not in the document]
AlexanderSher: Discussed offline

sdk/core/Azure.Core.Experimental/src/JsonData.cs
```diff
@@ -193,7 +193,7 @@ private void InitFromElement(JsonElement element)
         /// <param name="propertyName">The name of the property to get.</param>
         /// <returns>The value for a given property</returns>
         /// <remarks>If the <see cref="Kind"/> property is not <see cref="JsonValueKind.Object"/> this method throws <see cref="InvalidOperationException"/>.</remarks>
-        public T Get<T>(string propertyName)
+        public T? Get<T>(string propertyName)
```
JoshLove-msft, line 196: This probably makes sense, but why are we changing this here?

JoshLove-msft (approved):

sdk/core/Azure.Core/src/Azure.Core.csproj
```diff
@@ -8,7 +8,9 @@
...
+    <!-- Remove this line when Azure.Core is released for net6.0  -->
```
christothes, line 12: Let's create an issue to track this.

sdk/core/Azure.Core/src/RequestUriBuilder.cs
```diff
@@ -117,13 +117,15 @@ public string Path
             }
         }
 
-        private bool HasQuery => _queryIndex != -1;
+        /// <summary> Gets whether or not this instance of <see cref="RequestUriBuilder"/> has path. </summary>
```
christothes, line 120:
```suggestion
        /// <summary> Gets whether or not this instance of <see cref="RequestUriBuilder"/> has a path. </summary>
```

sdk/core/Azure.Core/src/RequestUriBuilder.cs
```diff
@@ -117,13 +117,15 @@ public string Path
...
+        protected bool HasPath => PathLength > 0;
 
-        /// <summary> Gets the length of the path in this instance of <see cref="RequestUriBuilder"/>. </summary>
-        protected int PathLength => HasQuery ? _queryIndex : _pathAndQuery.Length;
+        /// <summary> Gets whether or not this instance of <see cref="RequestUriBuilder"/> has query. </summary>
```
christothes, line 123:
```suggestion
        /// <summary> Gets whether or not this instance of <see cref="RequestUriBuilder"/> has a query. </summary>
```

sdk/core/Azure.Core.Experimental/src/JsonData.cs
christothes: Agreed - let's make these changes in a separate PR - do they need to ship this month?

AlexanderSher: > This probably makes sense, but why are we changing this here?

.net 6.0 complains about this. So I need to either fix it or disable nullability checks for this file.

christothes: Since this type is migrating to Core anyway, I think we should make the smallest change necessary and disable nullable checks for now.

JoshLove-msft: Well, that is somewhat dangerous as it introduces the risk that we forget to re-enable. I don't think we have tooling to check for this.

JoshLove-msft: I think the smallest _safe_ change here would be to just not target net6.0 in experimental yet.

christothes: This type will disappear from the project, so the nullable change will go with it.

JoshLove-msft: Yes, but it will be copy pasted into Core, hence the risk.

christothes: No - there is already a PR in progress.

JoshLove-msft: Ah ok - seems fine then.

AlexanderSher: Disabling nullable causes more subsequent changes, so minimal change is indeed not adding .net 6.0 for now.

sdk/core/Azure.Core/src/Azure.Core.csproj
AlexanderSher: Done: https://github.com/Azure/azure-sdk-for-net/issues/32329

sdk/core/Azure.Core/src/RequestUriBuilder.cs
AlexanderSher: Fixed

sdk/core/Azure.Core/src/RequestUriBuilder.cs
AlexanderSher: Fixed

christothes (approved):

PR #32288, opened 2022-11-04
Search project directory for test-resources files first

### LINKED ISSUE #4450: Improved metadata logging

### LINKED ISSUE #4450: Support test-resources.json at library level

benbp:
Resolves https://github.com/Azure/azure-sdk-tools/issues/4450

ServiceDirectory is equivalent to the test resources search path in the deploy script, so to add a secondary parameter would be redundant aside from affecting resource naming. I think it's better to handle this at a higher level for now.

### DIFF
--- a/eng/pipelines/templates/jobs/live.tests.yml
+++ b/eng/pipelines/templates/jobs/live.tests.yml
@@ -102,12 +102,22 @@ jobs:
           SubscriptionConfiguration: ${{ parameters.CloudConfig.SubscriptionConfiguration }}
           SubscriptionConfigurations: ${{ parameters.CloudConfig.SubscriptionConfigurations }}
 
+      - pwsh: |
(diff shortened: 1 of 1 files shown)


benbp:
I'm just realizing from testing that the naming DOES matter because of the environment variables consumed, so back to the drawing board and closing for now.

benbp:
Re-opening. Depends on https://github.com/Azure/azure-sdk-tools/pull/4623

PR #32304, opened 2022-11-04
Search project directory for test-resources files first

### LINKED ISSUE #4450: Support test-resources.json at library level

benbp:
Partially resolves https://github.com/Azure/azure-sdk-tools/issues/4450
Depends on https://github.com/Azure/azure-sdk-tools/pull/4623
Replaces https://github.com/Azure/azure-sdk-for-net/pull/32288 which won't re-open for some reason.

I decided it would be better to add detection at the language repository yaml level for template existence in the project directory, rather than enable the deploy script to fall back to the service directory if the project directory did not have a bicep template. This way we can fail fast locally when people intend to specify a TestResourcesDirectory override but make a mistake with the file path/existence. The downside is we have to add extra yaml handling in the languages.

### DIFF
--- a/eng/pipelines/templates/jobs/live.tests.yml
+++ b/eng/pipelines/templates/jobs/live.tests.yml
@@ -102,12 +102,23 @@ jobs:
           SubscriptionConfiguration: ${{ parameters.CloudConfig.SubscriptionConfiguration }}
           SubscriptionConfigurations: ${{ parameters.CloudConfig.SubscriptionConfigurations }}
 
+      - pwsh: |
+          $project = "${{ parameters.Project }}"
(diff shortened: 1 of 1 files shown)


benbp:
Example run with a sub-project bicep file: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1967774&view=results

JoshLove-msft (approved):

Issue #29589, opened 2022-06-30
Update Key Vault documentation


vcolin7:
### Library name

Azure.Security.KeyVault.Keys

### Please describe the feature.

We need to update the Key Vault documentation in a few places:

- [ ] **Authentication in Samples:** Our samples should show the use of Managed Identity instead of Service Principals with Client Secrets as the default authentication method. We can still show how to use the former for scenarios where the use of Managed Identity is not possible.
- [x] **Cryptographic operations using CBC and CBCPAD:** We should make it clear that there are some considerations to have when using these algorithms for local cryptographic operations using our clients:
  - We should add a notice similar to the following:
    > Microsoft recommends you not use CBC without first ensuring the integrity of the ciphertext using an HMAC, for example. See https://docs.microsoft.com/dotnet/standard/security/vulnerabilities-cbc-mode for more information.
  - When it comes to a user provided IV, we should make it clear that if they pass their own IV, they need to make sure they use a cryptographically random, non-repeating IV. Here's an example:
    > Optional initialization vector (IV). If you pass your own IV, make sure you use a cryptographically random, non-repeating IV. If null, a cryptographically random IV will be choosing using {RandomNumberGenerator | whatever cryptorng your language provides}.
[...]

PR #32309, opened 2022-11-04
Fix markdownlint issues in Key Vault markdown

heaths:


heaths:
I have some content changes coming up for #29589 and want to get all these markdownlint issues finally resolved so they don't mask actual content-related markdown issues the linter should catch. These are almost entirely whitespace changes with a couple of actual fixes to bookmarks.

pallavit (approved):
:shipit:

PR #32326, opened 2022-11-05
Fix namespace typo (Identitiy)

antkmsft:
Even though that namespace has only the internal class, it still shows up in IntelliSense as you type `using Azure.Ident`, and that is a bit confusing.


jsquire (approved):

jsquire:
Nice catch, @antkmsft.   Thanks!

PR #32335, opened 2022-11-07
[Update autorest to version 20221109.2]: Unify discriminator implementations across generator types

archerzz:

Found the following two issues while doing this PR.

- https://github.com/Azure/azure-sdk-for-net/issues/32351
- https://github.com/Azure/azure-sdk-for-net/issues/32355


m-nash:
@archerzz lets get an answer to [this](https://apiview.dev/Assemblies/Review/e3b03066906c4d6d81a16ad65db08448/db741d5f11d34559b6af870006a01ff8?diffRevisionId=81ff90db2ca74aa685f6c132aa2bea16&doc=False&diffOnly=True#dd3927a1da4d40a89913071f5f3900da) before we push this in.

sdk/cognitivelanguage/Azure.AI.Language.Conversations/tests/Generated/Models/BaseExtraInformation.cs
```diff
@@ -7,7 +7,11 @@
 
 namespace Azure.AI.Language.Conversations
 {
-    /// <summary> The abstract base object for entity extra information. </summary>
+    /// <summary>
+    /// The abstract base object for entity extra information.
+    /// Please note <see cref="BaseExtraInformation"/> is the base class. According to the scenario, a derived class of the base class might need to be assigned here, or this property needs to be casted to one of the possible derived classes.
+    /// The available derived classes include <see cref="EntitySubtype"/>, <see cref="ListKey"/> and <see cref="RegexKey"/>.
+    /// </summary>
     public partial class BaseExtraInformation
```
heaths, line 15: Isn't this supposed to actually be an abstract class as we discussed offline?

heaths (changes requested):

sdk/cognitivelanguage/Azure.AI.Language.Conversations/tests/Generated/Models/BaseExtraInformation.cs
m-nash: Its a GA'd type so we can't change it to abstract.

JoshLove-msft (approved):
Event Grid changes look good.

sdk/cognitivelanguage/Azure.AI.Language.Conversations/tests/Generated/Models/BaseExtraInformation.cs
heaths: Maybe we shouldn't call them "abstract base object" in doc comments at least, then. I'll switch my sign-off status, but I think we should still fix that. It's misleading. Given that, though, why all the extra work to duplicate deserialization in the Unknown* classes?

heaths (approved):

sdk/cognitivelanguage/Azure.AI.Language.Conversations/tests/Generated/Models/BaseExtraInformation.cs
m-nash: I think the comment still makes sense.  The intent is this is an abstract base class that you should never construct.

heaths: Offline, you pointed out that first line mentioning "abstract" is from swagger. The generated only added the subsequent 2 lines. In that case, fine. I figured if the C#-specific code generator was doing this, we should probably change it. If it's in swagger, *them* calling it abstract makes sense. It would just be misleading if the C# generator did.

PR #32343, opened 2022-11-07
[AzureMonitorExporter] customize repositoryurl

TimothyMothra:
### Changes
- customize RepositoryURL in Nupkg to link to this project

image


TimothyMothra:
@vishweshbankwar This is a super easy change, but no one else in this repo is doing this. We need to double check that we're following best practices.

lmolkova:
Why is it needed?

It seems Project Url  on Nuget already points to README of exporter and repo Url points to the repo https://www.nuget.org/packages/Azure.Monitor.OpenTelemetry.Exporter

`PackageProjectUrl` is set here
https://github.com/Azure/azure-sdk-for-net/blob/5b43f35884e3469f63d705bf26ba331e8c76689e/eng/Directory.Build.Common.targets#L23

vishweshbankwar:
> Why is it needed?
> 
> It seems Project Url on Nuget already points to README of exporter and repo Url points to the repo https://www.nuget.org/packages/Azure.Monitor.OpenTelemetry.Exporter
> 
> `PackageProjectUrl` is set here
> 
> https://github.com/Azure/azure-sdk-for-net/blob/5b43f35884e3469f63d705bf26ba331e8c76689e/eng/Directory.Build.Common.targets#L23

Thanks @lmolkova - the intention was to make it easy for users to find the source code. Given we have project link, I will follow up to see if it is still needed.

PR #32344, opened 2022-11-07
upgrade Personalizer RL.NET to 0.2.0

homezcx:
[upgrade Personalizer RL.NET to 0.2.0](https://github.com/Azure/azure-sdk-for-net/commit/22e5726b42dad6c973bd5a97090ac92746faa808)

### DIFF
--- a/eng/Packages.Data.props
+++ b/eng/Packages.Data.props
@@ -81,7 +81,7 @@
     <PackageReference Update="Microsoft.CSharp" Version="4.7.0" />
 
     <!-- Azure SDK packages -->
-	  <PackageReference Update="Azure.Communication.Identity" Version="1.2.0" />
+    <PackageReference Update="Azure.Communication.Identity" Version="1.2.0" />
     <PackageReference Update="Azure.Communication.Common" Version="1.2.1" />
     <PackageReference Update="Azure.Core" Version="1.25.0" />
     <PackageReference Update="Azure.Core.Amqp" Version="1.2.0" />
@@ -106,6 +106,7 @@
     <PackageReference Update="Microsoft.Azure.WebPubSub.Common" Version="1.1.0" />
     <PackageReference Update="Microsoft.Identity.Client" Version="4.46.0" />
     <PackageReference Update="Microsoft.Identity.Client.Extensions.Msal" Version="2.23.0" />
+	<PackageReference Update="RL.Net"  VersionOverride="0.2.0"/>
     <!--
       TODO: This package needs to be released as GA and arch-board approved before taking a dependency in any stable SDK library.
       Currently, it is referencd by Azure.Identity.BrokeredAuthentication which is still in beta
--- a/eng/Packages.Data.props
+++ b/eng/Packages.Data.props
@@ -81,7 +81,7 @@
     <PackageReference Update="Microsoft.CSharp" Version="4.7.0" />
 
     <!-- Azure SDK packages -->
-<PackageReference Update="Azure.Communication.Identity" Version="1.2.0" />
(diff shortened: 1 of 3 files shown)


orenmichaely (approved):

jackgerrits (approved):

eng/Packages.Data.props
```diff
@@ -106,6 +106,7 @@
     <PackageReference Update="Microsoft.Azure.WebPubSub.Common" Version="1.1.0" />
     <PackageReference Update="Microsoft.Identity.Client" Version="4.46.0" />
     <PackageReference Update="Microsoft.Identity.Client.Extensions.Msal" Version="2.23.0" />
+    <PackageReference Update="RL.Net"  VersionOverride="0.2.0"/>
```
weshaggard, line 109:
```suggestion
    <PackageReference Update="RL.Net" Version="0.2.0"/>
```

Is this package dependency approved by ArchBoard? /cc @tg-msft @KrzysztofCwalina

weshaggard (changes requested):
We should not merge this change until @tg-msft approves this dependency from ArchBoard prospective. We also might want to limit the scope of this dependency to only allow it only for Personalizer and not all the other libraries.

homezcx:
> We should not merge this change until @tg-msft approves this dependency from ArchBoard prospective. We also might want to limit the scope of this dependency to only allow it only for Personalizer and not all the other libraries.

Reverted the change and only update the nuget version in existing project files.

sdk/personalizer/Azure.AI.Personalizer/src/Azure.AI.Personalizer.csproj
```diff
@@ -21,7 +21,7 @@
 
   <ItemGroup>
     <PackageReference Include="Azure.Core" />
-    <PackageReference Include="RL.Net" VersionOverride="0.1.0-preview3" />
+    <PackageReference Include="RL.Net" Version="0.2.0" />
```
weshaggard, line 24: If you are doing this directly in the project it will need to be VersionOverride. I definitely think this is better then moving this preview dependency to the common props file but we still do need to get this dependency reviewed and approved and we should avoid shipping any stable versions of this package with this prerelease dependency.

homezcx: Thanks. Updated. Yes. Let's wait for the dependency review.

weshaggard (approved):

sdk/personalizer/Azure.AI.Personalizer/src/Azure.AI.Personalizer.csproj
homezcx: @tg-msft -> Could you help check the comments to review this dependency?

tg-msft: When we last discussed this in https://github.com/Azure/azure-sdk/issues/3877#issuecomment-1043638770, I believe the conclusion was you'd bring your dependency back to the Arch Board when ready because there were open questions like how to make this pay-for-play, what this looks like across our platform matrix, whether this works on other platforms, etc.  That still blocks your GA release per comments in https://apiview.dev/Assemblies/Review/35ddf25e338f4bb88d114ab43933c003.

That said, I have no concerns with folks updating the package in the repo to continue working on it though.  As @weshaggard suggested, this should definitely be scoped to just Personalizer.  (I'm not sure why anyone else would _want_ to take a dependency on it, but I'd rather we get a heads up if they tried.)

PR #32348, opened 2022-11-07
Updated emailRecipients to allow zero to recipients.

anmolbohra97:

### DIFF
--- a/sdk/communication/Azure.Communication.Email/src/Models/EmailRecipients.cs
+++ b/sdk/communication/Azure.Communication.Email/src/Models/EmailRecipients.cs
@@ -22,10 +22,10 @@ namespace Azure.Communication.Email.Models
         /// <param name="bcc">Email bcc recipients. </param>
         /// <exception cref="ArgumentNullException"> <paramref name="to"/> is null. </exception>
         public EmailRecipients(
-            IEnumerable<EmailAddress> to,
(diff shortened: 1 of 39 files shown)


AliRafiee (commented):
Please take a look at the live tests and maybe add a new test that only adds cc and bcc without to.

sdk/communication/Azure.Communication.Email/src/Models/EmailRecipients.cs
```diff
@@ -22,10 +22,10 @@ public partial class EmailRecipients
         /// <param name="bcc">Email bcc recipients. </param>
         /// <exception cref="ArgumentNullException"> <paramref name="to"/> is null. </exception>
         public EmailRecipients(
```
AliRafiee, line 24: Feels like we should remove the constructor for EmailRecipients in the generated file.
I'm not a 100% sure how to do that without changing the SWAGGER.

Are we not going to change the SWAGGER to not make To not required?

anmolbohra97: Yeah, we are going to update the swagger. Komivi mentioned that there are other changes coming too so plan is to update all of them together. This change is a temporary workaround to unblock customers.

kagbakpem (approved):

pallavit:
@joseharriaga could you please help them through the review.

PR #32349, opened 2022-11-07
Add interactive sample

heaths:


heaths:
This is effectively what .NET does for our samples: we write samples as tests and extract from them snippets we include in our packages' _samples/_ directory. This makes sure that our samples actually compile and run - something they didn't always do when I first started a few years ago.

Alternatively, I long ago prototype an idea about using Jupyter Notebooks, but you had to install additional kernels (think: language processors). It wasn't a good experience for anything other than Python examples. However, the Polyglot Notebooks (formerly .NET Interactive Notebooks) extension for VSCode makes this easier, and GitHub's rendering has improved as well. See https://github.com/Azure/azure-sdk-for-net/blob/56530de2487b10ea2cb0d6a0ccda56b2b3ab1636/sdk/keyvault/Azure.Security.KeyVault.Secrets/samples/InteractiveSample.ipynb for the file herein displayed as users might see it online. Combining this with VSCode for the Web (or perhaps a .devcontainer), our samples could be interactive while explaining, and this could be done across multiple languages.

/cc @jongio

PR #32359, opened 2022-11-08
[Redis] Minimize the breaking changes

Yao725:


sdk/redis/Azure.ResourceManager.Redis/src/ApiCompatBaseline.txt
```diff
@@ -1,4 +0,0 @@
-Compat issues with assembly Azure.ResourceManager.Redis:
```
ArcturusZhang, line 1: What is this file?

Yao725: It's a file that we can use it to suppress the api compat error. That why the previous PR with breaking changes could pass the CI.

m-nash (approved):

PR #32364, opened 2022-11-08
Support LRO rehydration with `ISerializable` interface

### LINKED ISSUE #2158: Implement rehydration for LROs

fengzhou-msft:
Fix https://github.com/Azure/autorest.csharp/issues/2158

This PR is to show a prototype to implement rehydration of LRO. It depends on https://github.com/Azure/autorest.csharp/pull/2843


sdk/containerservice/Azure.ResourceManager.ContainerService/src/Custom/LongRunningOperation/ContainerServiceArmOperation.cs
```diff
@@ -65,5 +99,58 @@ internal ContainerServiceArmOperation(ClientDiagnostics clientDiagnostics, HttpP
 
         /// <inheritdoc />
         public override ValueTask<Response> WaitForCompletionResponseAsync(TimeSpan pollingInterval, CancellationToken cancellationToken = default) => _operation.WaitForCompletionResponseAsync(pollingInterval, cancellationToken);
+
+        // TODO: use this class as a temporary workaround before https://github.com/Azure/autorest.csharp/issues/2231 is resolved so that we can deserialize to the original HttpWebResponseImplementation in LRO helper classes.
+        // No all methods and properties are implemented/populated as it's just for prototype.
+        internal class ContainerServiceResponse: Response
```
fengzhou-msft, line 105: This class is temporary for prototype purpose as we cannot deserialize the response to the abstract class of `Azure.Core.Response`. The final solution is likely to depend on https://github.com/Azure/autorest.csharp/issues/2231. Then we can deserialize to the original `HttpWebResponseImplementation`.

sdk/keyvault/Azure.Security.KeyVault.Certificates/src/DeleteCertificateOperation.cs
```diff
@@ -118,5 +118,10 @@ async ValueTask<OperationState> IOperation.UpdateStateAsync(bool async, Cancella
                     return OperationState.Failure(response, ex);
             }
         }
+
+        /// <summary>
+        /// To get the Id of the operation.
+        /// </summary>
+        public string GetOperationId() => Id;
```
heaths, line 125: We already expose this. What advantage is there of doing it two different ways, potentially causing confusion? These LROs - the whole of Key Vault packages - are HLCs.

sdk/textanalytics/Azure.AI.TextAnalytics/src/AbstractSummaryOperation.cs
```diff
@@ -297,5 +297,10 @@ async ValueTask<OperationState<AsyncPageable<AbstractSummaryResultCollection>>>
 
             return OperationState<AsyncPageable<AbstractSummaryResultCollection>>.Failure(rawResponse, requestFailedException);
         }
+
+        /// <summary>
+        /// To get the Id of the operation.
+        /// </summary>
+        public string GetOperationId() => Id;
```
heaths, line 304: Text Analytics are also HLCs and already expose the operation ID via a property. Why duplicate that?

heaths (changes requested):
I don't see any reason to duplicate getting the ID. Can you clarify? I would think this only leads to confusion.

sdk/resourcemanager/Azure.ResourceManager/tests/Scenario/LroRehydrationTests.cs
```diff
@@ -0,0 +1,84 @@
...
+            };
+            SubscriptionResource subscription = await Client.GetDefaultSubscriptionAsync().ConfigureAwait(false);
+            // The creation of a resource group is a fake LRO
+            var rgOp = await subscription.GetResourceGroups().Construct(AzureLocation.WestUS2, tags).CreateOrUpdateAsync(rgName);
+            var rgOpId = rgOp.Id;
+            var rehydration = new ResourcesArmOperationRehydration<ResourceGroupResource>(Client, rgOpId);
+            var rehydratedLro = await rehydration.RehydrateAsync(WaitUntil.Completed);
+            Assert.AreEqual(rehydratedLro.Id, rgOpId);
+            ResourceGroupResource rg = rehydratedLro.Value;
+            Assert.AreEqual(rgName, rg.Data.Name);
```
m-nash, line 36: can we add full body comparison, potentially in raw json format if thats easiest

sdk/keyvault/Azure.Security.KeyVault.Certificates/src/DeleteCertificateOperation.cs
fengzhou-msft: Yeah, I noticed this doe not make sense for data plane SDKs as the LRO implementation class usually both inherits `Operation<T>` and implements `IOperation`, making `Operation.Id` and `IOperation.GetOperationId()` duplicate on the same class.

This was originally added for management plane as its LRO implementation class only inherits `Operation<T>` and it has a private `_operation` property to do real work which eventually relies on an `IOperation` class. Therefore, it only has `Operation.Id` exposed and relies on the internal `IOperation.GetOperationId()`.

I will check with Michael how we can improve it.

heaths: Could we make the relatively newer management plane operations look like data plane and try to unify the APIs a bit e.g., an `Id` property instead of a `GetOperationId()` method? /cc @m-nash

heaths (commented):
This is prototyping, so not blocking.

sdk/keyvault/Azure.Security.KeyVault.Certificates/src/DeleteCertificateOperation.cs
m-nash: @heaths this is all prototyping so nothing is set in stone as of yet.  The issue with a property of `ID` is that the value can change at any moment during the execution.

The user scenario as I understand it is as follows:

- Start an LRO and retrieve an instance to an Operation
- save the ID from that LRO instance somewhere
- later construct an Operation passing in an instance to a client an the ID saved from above
- Check status / wait for completely / save and rehydrate again at a later date

The value of ID can change with each iteration of the poller and definitely changes as you cross from 3 high level states `Inital`, `Polling`, `Final`.  This is leading us towards having a method GetOperationId instead of a property as this doesn't give off a false impression that the ID is a known static value that doesn't change for a given instance of Operation.

sdk/resourcemanager/Azure.ResourceManager/src/Resources/Custom/Models/ResourceGroupExportResult.Serialization.cs
```diff
@@ -0,0 +1,19 @@
...
+#nullable disable
+
+using System;
+using System.Text.Json;
+using Azure;
+using Azure.Core;
+
+namespace Azure.ResourceManager.Resources.Models
+{
+    public partial class ResourceGroupExportResult : IOperationSourceProvider<ResourceGroupExportResult>
```
m-nash, line 15: We will need to only add this if net7 or higher if we end up going this route as this will fail to compile if a customer uses net6 and adds a dependency to our package

sdk/resourcemanager/Azure.ResourceManager/src/ArmOperationOfT.cs
```diff
@@ -39,6 +39,7 @@ internal ArmOperation(IOperationSource<T> source, ClientDiagnostics clientDiagno
         /// <summary> Initializes a new instance of ArmOperation. </summary>
         public ArmOperation(ArmClient client, string id)
         {
+#if NET7_0_OR_GREATER
```
KrzysztofCwalina, line 42: Why is this NET7+? This code does not seem to be using the interface with static members.

sdk/resourcemanager/Azure.ResourceManager/src/ArmOperationOfT.cs
```diff
@@ -39,6 +39,7 @@ internal ArmOperation(IOperationSource<T> source, ClientDiagnostics clientDiagno
...
             var method = typeof(T).GetMethods(BindingFlags.NonPublic | BindingFlags.Static).FirstOrDefault(m => m.Name.EndsWith("GetOperationSource"));
```
KrzysztofCwalina, line 43: Why don't you use GetMethod, instead of GetMethods? It's cheaper.

sdk/resourcemanager/Azure.ResourceManager/src/ArmOperationOfT.cs
fengzhou-msft: The method invoked through reflection is using the static method from the interface.

KrzysztofCwalina: If we are using reflection to invoke it, we don't need the interface, do we? i.e. if we are using reflection, we can just call a static method with the name/signature of our choosing. We don't need to bother with interfaces with static methods, and then this whole feature would work in NetStandard2.0. The whole point of adding this interfaces was to avoid reflection.

KrzysztofCwalina:
Sorry, I clicked wrong button "ready for review".

sdk/core/Azure.Core/src/Shared/StreamConverter.cs
```diff
@@ -0,0 +1,78 @@
...
+                throw new InvalidOperationException($"The response is not fully buffered.");
+            }
+
+            if (memoryContent.TryGetBuffer(out ArraySegment<byte> segment))
+            {
+                var data = new BinaryData(segment.AsMemory());
+#if NET6_0_OR_GREATER
+                writer.WriteRawValue(data);
+#else
+                JsonSerializer.Serialize(writer, JsonDocument.Parse(data.ToString()).RootElement);
```
JoshLove-msft, line 45:
```suggestion
                JsonSerializer.Serialize(writer, JsonDocument.Parse(data.ToMemory()).RootElement);
```

sdk/core/Azure.Core/src/Shared/StreamConverter.cs
```diff
@@ -0,0 +1,78 @@
...
+#endif
+            }
+            else
+            {
+                var data = new BinaryData(memoryContent.ToArray());
+#if NET6_0_OR_GREATER
+                writer.WriteRawValue(data);
+#else
+                JsonSerializer.Serialize(writer, JsonDocument.Parse(data.ToString()).RootElement);
```
JoshLove-msft, line 54:
```suggestion
                JsonSerializer.Serialize(writer, JsonDocument.Parse(data.ToMemory()).RootElement);
```

This avoids allocating again vs ToString

sdk/core/Azure.Core/src/Shared/OperationInternalOfT.cs
```diff
@@ -279,6 +280,22 @@ protected override async ValueTask<Response> UpdateStatusAsync(bool async, Cance
             }
         }
 
+        public virtual string GetOperationId()
+        {
+            var id = _operation.GetOperationId();
+            if (string.IsNullOrEmpty(id))
+            {
+                var serializeOptions = new JsonSerializerOptions { Converters = { new StreamConverter() } };
```
JoshLove-msft, line 288: Would probably be better to define a ResponseConverter so that we have more control over what gets serialized. For instance, Response has both ContentStream and a Content properties, but we only need to serialize one of them. I'd suggest just using the Content property which is already BinaryData.

eng/Directory.Build.Common.props
```diff
@@ -116,14 +116,14 @@
     <SupportsNetStandard20 Condition="'$(SupportsNetStandard20)' == '' and '$(IsClientLibrary)' == 'true'">true</SupportsNetStandard20>
     <SupportsNetStandard20 Condition="'$(SupportsNetStandard20)' == ''">false</SupportsNetStandard20>
 
-    <RequiredTargetFrameworks>net452;netstandard1.4;net461;netstandard2.0</RequiredTargetFrameworks>
-    <RequiredTargetFrameworks Condition="'$(SupportsNetStandard20)' == 'true'">netstandard2.0</RequiredTargetFrameworks>
+    <RequiredTargetFrameworks>net7.0;net452;netstandard1.4;net461;netstandard2.0</RequiredTargetFrameworks>
+    <RequiredTargetFrameworks Condition="'$(SupportsNetStandard20)' == 'true'">net7.0;netstandard2.0</RequiredTargetFrameworks>
```
jsquire, line 120: This impacts all libraries; what's your test plan for this change?

eng/Directory.Build.Common.props
```diff
@@ -116,14 +116,14 @@
...
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(IsTestProject)' == 'true' or '$(IsTestSupportProject)' == 'true' or '$(IsSamplesProject)' == 'true' or '$(IsPerfProject)' == 'true' or '$(IsStressProject)' == 'true'">
     <IsPackable>false</IsPackable>
     <!-- List newest targets first so that recordings are made with latest, running tests from editor runs latest, etc. -->
-    <RequiredTargetFrameworks>net6.0;netcoreapp3.1</RequiredTargetFrameworks>
+    <RequiredTargetFrameworks>net7.0;net6.0;netcoreapp3.1</RequiredTargetFrameworks>
```
jsquire, line 126: @heaths:  Is this supplanting your PR for updating the SDK and targets?   Seems we have some overlap.

sdk/confidentialledger/Azure.Security.ConfidentialLedger/src/PostLedgerEntryOperation.cs
```diff
@@ -88,5 +88,10 @@ async ValueTask<OperationState> IOperation.UpdateStateAsync(bool async, Cancella
 
         /// <inheritdoc />
         public override bool HasCompleted => _operationInternal.HasCompleted;
+
+        /// <summary>
+        /// To get the Id of the operation.
```
jsquire, line 93:
```suggestion
        /// Gets the Id of the operation.
```

jsquire: Let's please update all instances of this comment to follow the common form.   It appears in our public docs and should be consistent with other types/members.

eng/Directory.Build.Common.props
heaths: Yes, please don't do this yet. There's more work involved and I should be done this week. There's also a horrible regression in 7.0.100 that will cause all pipelines to run _way_ more than they need to (VSTest is broken; fixed in 7.0.101).

eng/Directory.Build.Common.props
heaths: I'm working on this and all the necessary changes in another PR. There are other issues to resolve, so please do not upgrade this to net7.0 in this PR.

heaths (changes requested):

heaths:
...besides, it's generally best to keep PRs focused on their intent in case we have to roll anything back, investigate via `git log` when/why something was changed, etc. Updating our target framework is a major milestone.

eng/Directory.Build.Common.props
fengzhou-msft: Sure. I saw your PR and I am aware that more work needs to be done for bumping to .NET 7. This is for prototyping now. Using static abstract interface method available from .NET 7 is one of the solutions we want to explore.

heaths: Last I talked to @KrzysztofCwalina and @tg-msft - our architects - late last year, all APIs have to be present in all versions we support with rare exception. Given static interface members require runtime changes, please talk with them.

Even if accepted, this must still wait. Updating all projects just so some can use net7.0 will have adverse effects, especially given that 7.0.100 regresses important behavior that will greatly increase build times.

KrzysztofCwalina: Yes, when I talked to Michael about it initially, we discussed using this only as an implementation detail, i.e. we cannot have public API differences as it's viral and prevents customers from building portable libraries.

heaths: Why multi-target just to use an implementation detail that we'd have to accommodate targeting netstandard2.0 anyway? Is the "work around" for netstandard2.0 less performant such that we need to multi-target? This adds complexity and, though minor, does increase nupkg size: build times in e.g., CIs, will increase as a result.

sdk/resourcemanager/Azure.ResourceManager/api/Azure.ResourceManager.net7.0.cs
```diff
@@ -79,11 +80,12 @@ public abstract partial class ArmCollection
     public partial class ArmOperation : Azure.Operation
     {
         protected ArmOperation() { }
-        public ArmOperation(Azure.ResourceManager.ArmClient client, string id) { }
         public override bool HasCompleted { get { throw null; } }
         [System.ComponentModel.EditorBrowsableAttribute(System.ComponentModel.EditorBrowsableState.Never)]
         public override string Id { get { throw null; } }
         public override Azure.Response GetRawResponse() { throw null; }
+        public static Azure.ResourceManager.ArmOperation Rehydrate(Azure.ResourceManager.ArmClient client, string id) { throw null; }
+        public static Azure.ResourceManager.ArmOperation<R> Rehydrate<R>(Azure.ResourceManager.ArmClient client, string id) where R : Azure.ResourceManager.ISerializable<R> { throw null; }
```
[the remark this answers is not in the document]
KrzysztofCwalina: Also, we should close in a larger group (cross-language) if we want to officially and publicly call the concept "rehydration". It's been a term that we used internally so far, and this is the first time (I think) it is making its way into public APIs.

sdk/resourcemanager/Azure.ResourceManager/api/Azure.ResourceManager.net7.0.cs
```diff
@@ -132,6 +133,16 @@ public abstract partial class ArmResource
...
     }
+    public enum AzureStackProfile
+    {
+        Profile20200901Hybrid = 0,
+    }
+    public partial interface ISerializable<T>
+    {
+        T Deserialize(System.IO.Stream data);
+        T Deserialize(System.ReadOnlyMemory<byte> data);
+        void Serialize(System.Span<byte> buffer);
```
KrzysztofCwalina, line 144: this method should by "bool TrySerialize(Span<byte>, out in bytesWritten)"

sdk/resourcemanager/Azure.ResourceManager/api/Azure.ResourceManager.net7.0.cs
```diff
@@ -132,6 +133,16 @@ public abstract partial class ArmResource
         public virtual Azure.ResourceManager.Resources.TagResource GetTagResource() { throw null; }
         protected virtual bool TryGetApiVersion(Azure.Core.ResourceType resourceType, out string apiVersion) { throw null; }
     }
+    public enum AzureStackProfile
+    {
+        Profile20200901Hybrid = 0,
+    }
+    public partial interface ISerializable<T>
+    {
+        T Deserialize(System.IO.Stream data);
```
KrzysztofCwalina, line 142: This needs to take a cancellation token

sdk/resourcemanager/Azure.ResourceManager/api/Azure.ResourceManager.net7.0.cs
```diff
@@ -132,6 +133,16 @@ public abstract partial class ArmResource
...
+        T Deserialize(System.ReadOnlyMemory<byte> data);
```
KrzysztofCwalina, line 143: Why does it take memory and not span?

sdk/resourcemanager/Azure.ResourceManager/api/Azure.ResourceManager.net7.0.cs
KrzysztofCwalina: Overall, I was thinking we should have something like the following:
```csharp
public interface IBufferSerializable
{
    bool TrySerialize(Span<byte> buffer, out int bytesWritten, StandardFormat format = default);
}

public interface IBufferDeserializable : IBufferSerializable
{
    bool TryDeserialize(ReadOnlySpan<byte> data, out int bytesConsumed, StandardFormat format = default);
}
```

sdk/resourcemanager/Azure.ResourceManager/src/ArmOperation.cs
```diff
@@ -55,6 +55,22 @@ public static ArmOperation Rehydrate(ArmClient client, string id)
...
+#if NET7_0_OR_GREATER
+            IOperationSource<R> source = new GenericOperationSource<R>(client);
+            var nextLinkOperation = NextLinkOperationImplementation.Create(source, client.Pipeline, id, out string finalResponse);
+            // TODO: Do we need more specific OptionsNamespace, ProviderNamespace and OperationTypeName and possibly from id?
+            var clientDiagnostics = new ClientDiagnostics("Azure.ResourceManager", "Microsoft.Resources", client.Diagnostics);
+            DecodedResponse response;
+            var operation = finalResponse == null ? new OperationInternal<R>(clientDiagnostics, nextLinkOperation, null, operationTypeName: null, fallbackStrategy: new ExponentialDelayStrategy()) : OperationInternal<R>.Succeeded(response = JsonSerializer.Deserialize<DecodedResponse>(finalResponse)!, source.CreateResult(response, CancellationToken.None));
+            return new ArmOperation<R>(operation);
+#else
+            throw new InvalidOperationException("LRO rehydration is not supported in this version of .NET. Please upgrade to .NET 7.0 or later.");
```
KrzysztofCwalina, line 70: Why don't we support?

KrzysztofCwalina (changes requested):

PR #32365, opened 2022-11-08
[MySql] Update renaming config to make them more robust to swagger changes

fengzhou-msft:

### DIFF
--- a/sdk/mysql/Azure.ResourceManager.MySql/src/autorest.md
+++ b/sdk/mysql/Azure.ResourceManager.MySql/src/autorest.md
@@ -88,7 +88,6 @@ prepend-rp-prefix:
   - MinimalTlsVersionEnum
   - GeoRedundantBackup
   - InfrastructureEncryption
-  - NameAvailabilityRequest
   - PerformanceTierListResult
   - PerformanceTierServiceLevelObjectives
   - PrivateEndpointProvisioningState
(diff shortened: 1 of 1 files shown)


sdk/mysql/Azure.ResourceManager.MySql/src/autorest.md
```diff
@@ -149,6 +148,7 @@ rename-mapping:
   ConfigurationListResult: MySqlConfigurations
   LogFile.properties.type: LogFileType
   ConfigurationListResult.value: Values
+  NameAvailabilityRequest: MySqlNameAvailabilityContent
```
Yao725, line 151: Code change looks good to me, but as we will automatically change the suffix from `Request` to `Content`, I'd like to know why we need this change in autorest.md.

Yao725 (approved):

sdk/mysql/Azure.ResourceManager.MySql/src/autorest.md
fengzhou-msft: There are conditions for the generator to change suffix from `Request` to `Content`. In this swagger [preview PR](https://github.com/azure-sdk/azure-sdk-for-net-pr/pull/740/files#diff-39f48986688b855f6bcea1c945e00cd0b32105a5e8180273f166be393c24f150),  the auto-renaming didn't happen. I believe it's bacause the `MySqlFlexibleServerNameAvailabilityRequest` model is used as an input parameter in a new `POST` method.

PR #32452, opened 2022-11-11
Fix empty string validation for GetEntity

### LINKED ISSUE #32447: [BUG] Breaking change in Azure.Data.Table makes it no longer possible to retrieve an entity with an empty partition or row key
### Library name and version

Azure.Data.Tables 12.7.0

### Describe the bug

Due to legacy reasons, we have a scenario where we might have entities that are stored with `string.Empty` as a row key and we need to be able to load those. The service accepts such entities without a problem. Here is a simple reproduction code

```csharp
using Azure.Data.Tables;

var partitionKey = Guid.NewGuid().ToString();
var entity = new TableEntity
{
    RowKey = string.Empty,
    PartitionKey = partitionKey
};
var tableServiceClient = new TableServiceClient(Environment.GetEnvironmentVariable("AzureTableConnectionString"));
var tableClient = tableServiceClient.GetTableClient("AzureDataTableRepro");
await tableClient.CreateIfNotExistsAsync();
await tableClient.AddEntityAsync(entity);

var foundEntity = await tableClient.GetEntityAsync<TableEntity>(partitionKey, string.Empty);

Console.WriteLine($"{foundEntity.Value.PartitionKey}");
```
This prints the PartitionKey previously generated when using 12.6.1.

When we upgrade to 12.7.0 we get the following stack trace

```
Unhandled exception. System.ArgumentException: Value cannot be an empty string. (Parameter 'rowKey')
   at Azure.Core.Argument.AssertNotNullOrEmpty(String value, String name)
   at Azure.Data.Tables.TableRestClient.QueryEntityWithPartitionAndRowKeyAsync(String table, String partitionKey, String rowKey, Nullable`1 timeout, String format, String select, String filter, RequestContext context)
   at Azure.Data.Tables.TableClient.GetEntityInternalAsync[T](Boolean async, String partitionKey, String rowKey, Boolean noThrow, IEnumerable`1 select, CancellationToken cancellationToken)
   at Azure.Data.Tables.TableClient.GetEntityAsync[T](String partitionKey, String rowKey, IEnumerable`1 select, CancellationToken cancellationToken)
   at Program.<Main>$(String[] args) in /home/danielmarbach/Projects/AzureDataTableRepro/ConsoleApp1/ConsoleApp1/Program.cs:line 16
   at Program.<Main>(String[] args)
```

While historically null was never permitted because the client prevented it with validating against null 12.7.0 introduced validation against empty which is a breaking change which also makes it impossible to retrieve already persisted entities that are using a row key with `string.Empty`

### Expected behavior

Being able to store and retrieve entities with rowKey = string.Empty

### Actual behavior

An ArgumentException is thrown

For example `QueryEntityWithPartitionAndRowKeyAsync` has been changed from

```csharp
        public async Task<ResponseWithHeaders<IReadOnlyDictionary<string, object>, TableQueryEntityWithPartitionAndRowKeyHeaders>> QueryEntityWithPartitionAndRowKeyAsync(string table, string partitionKey, string rowKey, int? timeout = null, QueryOptions queryOptions = null, CancellationToken cancellationToken = default)
        {
            if (table == null)
            {
                throw new ArgumentNullException(nameof(table));
            }
            if (partitionKey == null)
            {
                throw new ArgumentNullException(nameof(partitionKey));
            }
            if (rowKey == null)
            {
                throw new ArgumentNullException(nameof(rowKey));
            }
```

to

```csharp
            Argument.AssertNotNullOrEmpty(table, nameof(table));
            Argument.AssertNotNullOrEmpty(partitionKey, nameof(partitionKey));
            Argument.AssertNotNullOrEmpty(rowKey, nameof(rowKey));
```

### Reproduction Steps

```csharp
using Azure.Data.Tables;

var partitionKey = Guid.NewGuid().ToString();
var entity = new TableEntity
{
    RowKey = string.Empty,
    PartitionKey = partitionKey
};
var tableServiceClient = new TableServiceClient(Environment.GetEnvironmentVariable("AzureTableConnectionString"));
var tableClient = tableServiceClient.GetTableClient("AzureDataTableRepro");
await tableClient.CreateIfNotExistsAsync();
await tableClient.AddEntityAsync(entity);

var foundEntity = await tableClient.GetEntityAsync<TableEntity>(partitionKey, string.Empty);

Console.WriteLine($"{foundEntity.Value.PartitionKey}");
```

### Environment

_No response_

christothes:
fixes #32447

related to https://github.com/Azure/autorest.csharp/issues/2856


jsquire (approved):

sdk/tables/Azure.Data.Tables/tests/TableClientLiveTests.cs
```diff
@@ -1286,6 +1286,24 @@ public async Task GetEntityIfExistsContainsValueWhenExists()
             Assert.That(entityResults, Is.Not.Null, "The entity should not be null.");
         }
 
+        [RecordedTest]
+        public async Task GetEntityAllowsEmptyRowKey()
```
danielmarbach, line 1290: Shouldn't the empty partition key scenario also be tested to prevent regression?

sdk/tables/Azure.Data.Tables/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 ### Breaking Changes
 
 ### Bugs Fixed
+- Removed client side validation which prevented `GetEntity` and `GetEntityAsync` from getting an entity with an empty string as its RowKey value. ([#32447](https://github.com/Azure/azure-sdk-for-net/issues/32447))
```
danielmarbach, line 10: The fix also enables empty partition key again which is not mentioned here

christothes: I'll update this when I prepare the release next week.

danielmarbach:
And I forgot to mention. Thank you for doing this ☺️

aascione-mw:
Hi, I tested the AddEntityAsync with an empty/null PartitionKey, but it is still throwing exception. I am on the version 12.7.1 of Azure.Data.Tables and also tested the latest version with no luck. We also have the legacy data in the storage account that is getting migrated  that has an empty partiotion key value. And this is a breaking change for us. Can somebody give me some clarification?

PR #32458, opened 2022-11-13
[AzureMonitorExporter] updating track activity events method to also track non-exception events

fernandoescolar:
Currently this exporter only logs exception events:

https://github.com/Azure/azure-sdk-for-net/blob/edd266f12d23344706e36da6e3b8eaab08de1a41/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TraceHelper.cs#L180-L200

As we can see in the official Application Insights API documentation (https://learn.microsoft.com/en-us/azure/azure-monitor/app/api-custom-events-metrics#api-summary) we can create different tracking records. I would like to use the [TrackEvent](https://learn.microsoft.com/en-us/azure/azure-monitor/app/api-custom-events-metrics#trackevent) method to store the other non-captured events.

Currently the official Application Insights Library for dotnet creates a new `EventTelemetry`:

https://github.com/microsoft/ApplicationInsights-dotnet/blob/7191788b3fb7169394e9aed64982d1128c9c9fd6/BASE/src/Microsoft.ApplicationInsights/DataContracts/EventTelemetry.cs

And it stores the properties and the metrics inside this object:

https://github.com/microsoft/ApplicationInsights-dotnet/blob/7191788b3fb7169394e9aed64982d1128c9c9fd6/BASE/src/Microsoft.ApplicationInsights/TelemetryClient.cs#L115-L130

In the other hand, I have found in the Go-Lang exporter it saves the tags as `Measurements` when it is a number and as `Properties` in other case:

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/azuremonitorexporter/trace_to_envelope.go#L647-L666

So, I have created this PR to add those currently ignored events to the "customEvents" collection:

https://github.com/Azure/azure-sdk-for-net/blob/e801e77feaebad625e9b9095746fd9ada443d96d/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TraceHelper.cs#L180-L197

To do that I have modified the `TelemetryItem` to understand items of type "Event":

https://github.com/Azure/azure-sdk-for-net/blob/e801e77feaebad625e9b9095746fd9ada443d96d/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Customizations/Models/TelemetryItem.cs#L58-L61

I have added the same logic we can find in the Go-Lang exporter to create these events:

https://github.com/Azure/azure-sdk-for-net/blob/33e40783989a3cd6212692ff89daecc05e667578/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TraceHelper.cs#L278-L298

And I have created some tests to validate it works well.

### DIFF
--- a/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TraceHelper.cs
+++ b/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TraceHelper.cs
@@ -32,7 +32,7 @@ namespace Azure.Monitor.OpenTelemetry.Exporter.Internals
                 // Check for Exceptions events
                 if (activity.Events.Any())
                 {
-                    AddExceptionTelemetryFromActivityExceptionEvents(activity, telemetryItem, telemetryItems);
+                    AddExceptionAndEventTelemetryFromActivityEvents(activity, telemetryItem, telemetryItems);
                 }
 
                 switch (activity.GetTelemetryType())
@@ -175,7 +175,7 @@ namespace Azure.Monitor.OpenTelemetry.Exporter.Internals
             return activity.DisplayName;
         }
 
-        private static void AddExceptionTelemetryFromActivityExceptionEvents(Activity activity, TelemetryItem telemetryItem, List<TelemetryItem> telemetryItems)
+        private static void AddExceptionAndEventTelemetryFromActivityEvents(Activity activity, TelemetryItem telemetryItem, List<TelemetryItem> telemetryItems)
         {
             foreach (var evnt in activity.Events)
             {
@@ -185,7 +185,7 @@ namespace Azure.Monitor.OpenTelemetry.Exporter.Internals
                     {
                         TryAddExceptionTelemetryFromActivityExceptionEvents(activity, telemetryItem, telemetryItems, evnt);
                     }
-                    else
+                    else if (!string.IsNullOrWhiteSpace(evnt.Name))
                     {
                         TryAddEventTelemetryFromActivityEvents(activity, telemetryItem, telemetryItems, evnt);
                     }
@@ -213,7 +213,7 @@ namespace Azure.Monitor.OpenTelemetry.Exporter.Internals
             var eventData = GetEventDataDetailsOnTelemetryItem(evnt.Name, evnt.Tags);
             if (eventData != null)
             {
-                var eventTelemetryItem = new TelemetryItem(telemetryItem, activity.SpanId, activity.Kind, evnt.Timestamp);
+                var eventTelemetryItem = new TelemetryItem(telemetryItem, activity.SpanId, activity.Kind, evnt.Timestamp, isAnEvent: true);
                 eventTelemetryItem.Data = eventData;
                 telemetryItems.Add(eventTelemetryItem);
             }
@@ -278,7 +278,7 @@ namespace Azure.Monitor.OpenTelemetry.Exporter.Internals
         internal static MonitorBase GetEventDataDetailsOnTelemetryItem(string eventName, IEnumerable<KeyValuePair<string, object>> activityEventTags)
         {
             var eventData = new TelemetryEventData(Version, eventName);
-            foreach(var tag in activityEventTags)
+            foreach (var tag in activityEventTags)
             {
                 eventData.Properties.Add(tag.Key, tag.Value.ToString());
             }
--- a/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TraceHelper.cs
+++ b/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TraceHelper.cs
@@ -280,7 +280,14 @@ namespace Azure.Monitor.OpenTelemetry.Exporter.Internals
             var eventData = new TelemetryEventData(Version, eventName);
             foreach (var tag in activityEventTags)
             {
-                eventData.Properties.Add(tag.Key, tag.Value.ToString());
+                if (double.TryParse(tag.Value.ToString(), out var value))
+                {
+                    eventData.Measurements.Add(tag.Key, value);
+                }
+                else
+                {
+                    eventData.Properties.Add(tag.Key, tag.Value.ToString());
+                }
             }
 
             return new MonitorBase
--- a/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TraceHelper.cs
+++ b/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TraceHelper.cs
@@ -179,23 +179,43 @@ namespace Azure.Monitor.OpenTelemetry.Exporter.Internals
         {
             foreach (var evnt in activity.Events)
             {
-                if (evnt.Name == SemanticConventions.AttributeExceptionEventName)
+                try
                 {
-                    try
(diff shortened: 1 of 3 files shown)


fernandoescolar:

JorTurFer (approved):

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TraceHelper.cs
```diff
@@ -175,27 +175,47 @@ internal static string GetOperationName(Activity activity, ref AzMonList MappedT
...
+                telemetryItems.Add(exceptionTelemetryItem);
+            }
+        }
+
+        private static void TryAddEventTelemetryFromActivityEvents(Activity activity, TelemetryItem telemetryItem, List<TelemetryItem> telemetryItems, ActivityEvent evnt)
+        {
+            var eventData = GetEventDataDetailsOnTelemetryItem(evnt.Name, evnt.Tags);
+            if (eventData != null)
+            {
+                var eventTelemetryItem = new TelemetryItem(telemetryItem, activity.SpanId, activity.Kind, evnt.Timestamp, isAnEvent: true);
```
cijothomas, line 216: SpanEvents are better mapped to TraceTelemetry, not customEvents.

cijothomas (changes requested):
See https://github.com/Azure/azure-sdk-for-net/pull/32458/files#r1021798616
Please describe the design and outcome of this PR in the description itself, so we can tell if this is in the right direction.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TraceHelper.cs
JorTurFer: Do you mean it's better to send the SpanEvents to Trace table than CustomEvents table?

cijothomas: SpanEvent is a log, embedded inside Span. https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/glossary.md#embedded-log

JorTurFer: Yes it is, but Exceptions are a subset of events and the SDK uses a different table for them (Exceptions), so it could make sense to use the customEvents table for the other event types, couldn't?

fernandoescolar: In the Application Insights (AI) official documentation, we can read we can create different tracking records: https://learn.microsoft.com/en-us/azure/azure-monitor/app/api-custom-events-metrics#api-summary

As you said "SpanEvent is a log, embedded inside Span" and it is tracked in AI using this method:

```csharp
void TrackTrace(string message, SeverityLevel severityLevel, IDictionary<string, string> properties)
```

But if I want to track an event, I should use this:

```csharp
void TrackEvent(string eventName, IDictionary<string, string> properties = null, IDictionary<string, double> metrics = null)
```

I think the second one better fits the information we are using in an open telemetry activity event:
- `TrackTrace` contains a `Message` but `ActivityEvent` does not
- The `TrackTrace` contains a `SeverityLevel` but the `ActivityEvent` and `TrackEvent` do not
- TrackEvent` contains a `Name` as `ActivityEvent`

cijothomas: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/event-api.md#events-api-interface This is what would likely become the customEvent in Application Insights.

Its true TraceTelemetry has severity which SpanEvent does not provide, but SpanEvents are just logs, so they should be mapped to TraceTelemetry. "Events" like "button-clicked" , "added-to-cart" would come in OTel via the Event-API, and then it'll be mapped to CustomEvent table in application insights.

fernandoescolar: This API proposes to create an event object that wraps an `OTEL.LogRecord` with its severity level and so on. I think that if we want to translate it to Application Insights (AI) we should create two records: one of type `AI.Event` with the property `OTEL.Event.event_name` and another of type `AI.Trace` with the information from `OTEL.Event.LogRecord`.

Each dotnet application when sending its telemetry to AI translates the logs as `AI.Traces`. And this is the expected default behavior.

If you want, I could add in this PR support to this c# code:

```csharp
var testEvent = new EventId(1, "Event 1"); // create an OTEL event-api compatible event.
logger. Log(LogLevel.Critical, testEvent, "Hello"); // send the event to the dotnet logger
logger.LogCritical(testEvent, "Bye"); // another way to do the same.
```

Creating both logs in AI: one as `AI.Event` and the other as `AI.Trace`.

And likewise, in the actual code of the PR it should be fine if I create an `OTEL.Activity` containing an `OTEL.ActivityEvent` and I store them as an `AI.Request` or even `AI.Dependency` for the `OTEL.Activity` and an `AI.Event` for the `OTEL.ActivityEvent`.

Semantically the `OTEL.Event` and the `OTEL.ActivityEvent` are events, and these types of records are stored in AI as `AI.Events`. As `OTEL.Activity` could become `AI.Request` or `AI.Dependency` depending on its content. And as `OTEL.LogRecord` is stored as `AI.Trace`.

cijothomas: Though it appears that SpanEvent is a better fit as AI's Events, it was decided by the product team that Events serve a different purpose, and OTel SpanEvents must be mapped to TraceTelemetry table. The only official implementation is in Java - https://github.com/microsoft/ApplicationInsights-Java/pull/1834/files. 
That can be used as a reference. (Note: It contains links to some Microsoft internal repos which you'll not have access to.)

If you have a desire to use OTel APIs to produce Application Insights' CustomEvent, then there is no support for that today. If this is an issue, will need to engage with the product owners. I'll let someone know to reply here.

JorTurFer: We need the custom events available because we use it from business to generate stuff. If we need to use logger events to map them to custom metrics, we will do it. I mean, during the migration from AI SDK to Otel, we already have to modify the code, so we could do it, but we need to support customEvents somehow.

fernandoescolar: Yes @JorTurFer

In addition, if I want to store an OTel AtivityEvent as an AI Trace I should invent the severity level property value and I should not...

Maybe, we should close this PR and open another one with a new approach to creating AI Events, once we have discussed it with the product owners. @cijothomas Could you provide us with someone's contact, please?

cijothomas: _new approach to creating AI Events, once we have discussed it with the product owners_

What I conveyed is the product teams decision. i.e OTel's SpanEvents are to be mapped as TraceTelemetry in Application Insights.

There is no current support for producing Application Insights' CustomEvent using OTel today.

I have let PM owner to reply here, if you have additional questions/suggestions.

PR #32462, opened 2022-11-14
[WebPubSub] Adding feature support

vicancy:

Supporting new features in Web PubSub
1. Add "groups" parameter when generating client access URI
2. Add a helper class `SearchFilter` for generating the filter string, referencing the practice of Cognitive Search
3. Add more tests

### DIFF
--- a/sdk/webpubsub/Azure.Messaging.WebPubSub/api/Azure.Messaging.WebPubSub.netstandard2.0.cs
+++ b/sdk/webpubsub/Azure.Messaging.WebPubSub/api/Azure.Messaging.WebPubSub.netstandard2.0.cs
@@ -1,5 +1,10 @@
 namespace Azure.Messaging.WebPubSub
 {
+    public static partial class SearchFilter
+    {
+        public static string Create(System.FormattableString filter) { throw null; }
+        public static string Create(System.FormattableString filter, System.IFormatProvider formatProvider) { throw null; }
+    }
     public enum WebPubSubPermission
     {
         SendToGroup = 1,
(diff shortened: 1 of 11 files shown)


sdk/webpubsub/Azure.Messaging.WebPubSub/api/Azure.Messaging.WebPubSub.netstandard2.0.cs
```diff
@@ -1,5 +1,10 @@
 namespace Azure.Messaging.WebPubSub
 {
+    public static partial class SearchFilter
```
ArthurMa1978, line 3: The name is too generic, consider adding a prefix

sdk/webpubsub/Azure.Messaging.WebPubSub/api/Azure.Messaging.WebPubSub.netstandard2.0.cs
```diff
@@ -33,10 +38,14 @@ public partial class WebPubSubServiceClient
...
-        public virtual System.Threading.Tasks.Task<System.Uri> GetClientAccessUriAsync(System.DateTimeOffset expiresAt, string userId = null, System.Collections.Generic.IEnumerable<string> roles = null, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
-        public virtual System.Threading.Tasks.Task<System.Uri> GetClientAccessUriAsync(System.TimeSpan expiresAfter = default(System.TimeSpan), string userId = null, System.Collections.Generic.IEnumerable<string> roles = null, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
+        public virtual System.Uri GetClientAccessUri(System.DateTimeOffset expiresAt, string userId = null, System.Collections.Generic.IEnumerable<string> roles = null, System.Collections.Generic.IEnumerable<string> groups = null, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
```
ArthurMa1978, line 41: Here introduced breaking changes which is not allowed if not increase the major version.
2 solutions:
1. Increase the version to 2.0.0
2. Keep old methods and add new overloads

sdk/webpubsub/Azure.Messaging.WebPubSub/src/SearchFilter.cs
```diff
@@ -0,0 +1,113 @@
...
+        /// </summary>
+        /// <param name="text">The text to quote.</param>
+        /// <returns>The quoted text.</returns>
+        private static string Quote(string text)
+        {
+            if (text == null)
+            { return "null"; }
+
+            // Optimistically allocate an extra 5% for escapes
+            StringBuilder builder = new StringBuilder(2 + (int)(text.Length * 1.05));
```
ArthurMa1978, line 99: How about use the [String.Replace](https://learn.microsoft.com/en-us/dotnet/api/system.string.replace?view=net-7.0) directly?

sdk/webpubsub/Azure.Messaging.WebPubSub/src/WebPubSubServiceClient_helpers.cs
```diff
@@ -6,6 +6,7 @@
 using System.Linq;
 using System.Text;
 using System.Text.Json;
+using System.Text.RegularExpressions;
```
ArthurMa1978, line 9: I didn't find any regex functionality be used in this file

sdk/webpubsub/Azure.Messaging.WebPubSub/api/Azure.Messaging.WebPubSub.netstandard2.0.cs
vicancy: I don't think it is a breaking change though, the previous method (L36) is changed to L42 with all the default values removed. I thought it is the recommended practice to add the parameter `groups` to the method having default values. Please correct me if I am wrong.

sdk/webpubsub/Azure.Messaging.WebPubSub/api/Azure.Messaging.WebPubSub.netstandard2.0.cs
vicancy: Referenced https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/search/Azure.Search.Documents/src/SearchFilter.cs#L19

sdk/webpubsub/Azure.Messaging.WebPubSub/src/SearchFilter.cs
vicancy: Referenced https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/search/Azure.Search.Documents/src/SearchFilter.cs#L19

sdk/webpubsub/Azure.Messaging.WebPubSub/src/WebPubSubServiceClient_helpers.cs
vicancy: nice catch

sdk/webpubsub/Azure.Messaging.WebPubSub/api/Azure.Messaging.WebPubSub.netstandard2.0.cs
ArthurMa1978: Got confirmed from @m-nash, this is not breaking changes, case close :)

ArthurMa1978 (approved):

vicancy:
Hi @jsquire do you have any expert looking into the CI failure on "Build Test Windows2019_NET461_ProjectRef_Release" complaining about some assembly missing? This is strange and should not be related to my change because my previous commit passed successfully however the next commit containing only a simple rename failed with this CI failure.

jsquire:
@vicancy:  I don't see any failures.   I'd suggest raising issues around pipelines in the "Engineering System" channel on Teams so that the right folks can assist.

vicancy:
Thanks for the review @ArthurMa1978 ! Could you also help approve the APIview https://apiview.dev/Assemblies/Review/c02c4162593a42a7b404ca7d927e4cef?

PR #32471, opened 2022-11-14
Restore RunApiCompat in Azure.Core

AlexanderSher:
Also, as @joseharriaga found, we don't need explicit reference of .net6.0 in test projects cause it is added automatically:
https://github.com/Azure/azure-sdk-for-net/blob/main/eng/Directory.Build.Common.props#L123-L125


annelo-msft (approved):
LGTM

PR #32481, opened 2022-11-15
Some performance improvements in Azure.Core for .Net 6.0

AlexanderSher:
Performance gains in .Net 6.0 is about 10%

### DIFF
--- a/global.json
+++ b/global.json
@@ -3,7 +3,7 @@
     "Microsoft.Build.Traversal": "3.0.23"
   },
   "sdk": {
-    "version": "6.0.100",
+    "version": "6.0.403",
     "rollForward": "feature"
   }
 }
--- a/sdk/core/Azure.Core/src/Pipeline/HttpClientTransport.cs
+++ b/sdk/core/Azure.Core/src/Pipeline/HttpClientTransport.cs
@@ -207,25 +207,61 @@ namespace Azure.Core.Pipeline
 
         internal static bool TryGetHeader(HttpHeaders headers, HttpContent? content, string name, [NotNullWhen(true)] out string? value)
         {
+#if NET6_0_OR_GREATER
+            if (headers.NonValidated.TryGetValues(name, out HeaderStringValues values) ||
+                content is not null && content.Headers.NonValidated.TryGetValues(name, out values))
+            {
+                value = JoinHeaderValues(values);
+                return true;
+            }
+#else
             if (TryGetHeader(headers, content, name, out IEnumerable<string>? values))
             {
                 value = JoinHeaderValues(values);
                 return true;
             }
-
+#endif
             value = null;
             return false;
         }
 
         internal static bool TryGetHeader(HttpHeaders headers, HttpContent? content, string name, [NotNullWhen(true)] out IEnumerable<string>? values)
         {
+#if NET6_0_OR_GREATER
+            if (headers.NonValidated.TryGetValues(name, out HeaderStringValues headerStringValues) ||
+                content != null &&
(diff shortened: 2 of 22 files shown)


sdk/core/Azure.Core/src/Pipeline/HttpClientTransport.cs
```diff
@@ -238,28 +274,43 @@ internal static IEnumerable<HttpHeader> GetHeaders(HttpHeaders headers, HttpCont
...
         internal static bool RemoveHeader(HttpHeaders headers, HttpContent? content, string name)
         {
             // .Remove throws on invalid header name so use TryGet here to check
+#if NET6_0_OR_GREATER
+            if (headers.NonValidated.Contains(name) && headers.Remove(name))
+            {
+                return true;
+            }
+
+            return content is not null && content.Headers.NonValidated.Contains(name) && content.Headers.Remove(name);
```
christothes, line 290: Is the `Contains` call necessary here, or is it faster to check existence first in NonValidated?
```suggestion
            return headers.Remove(name) || content is not null && content.Headers.Remove(name);
```

sdk/storage/Azure.Storage.Common/src/Azure.Storage.Common.csproj
```diff
@@ -1,6 +1,8 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>$(RequiredTargetFrameworks)</TargetFrameworks>
+    <TargetFrameworks>$(RequiredTargetFrameworks);net6.0</TargetFrameworks>
```
christothes, line 3: Do we need this in the specific client libraries also?
Can we add this to the storage test projects?

sdk/core/Azure.Core/src/Pipeline/HttpClientTransport.cs
AlexanderSher: It depends on what we do more often - remove existing or non-existing header. If header is added, then yes - simple call of `headers.Remove` is faster. And header is missing, then `headers.NonValidated.Contains` validates faster than `headers.Remove`. Probably we can use your approach.

sdk/storage/Azure.Storage.Common/src/Azure.Storage.Common.csproj
AlexanderSher: I'm going to revert changes in `Azure.Storage` to minimize the size of the PR and send those as a separate one.

sdk/core/Azure.Core/src/Pipeline/HttpClientTransport.cs
AlexanderSher: It looks like `headers.Remove` still thrown on invalid header, so we must call  `headers.NonValidated.Contains` first

global.json
```diff
@@ -3,7 +3,7 @@
     "Microsoft.Build.Traversal": "3.0.23"
   },
   "sdk": {
-    "version": "6.0.100",
+    "version": "6.0.403",
```
christothes, line 6: It doesn't look like we've added net6.0 to the TFM for `Azure.Core.Tests.csproj` yet. I think we should add it

AlexanderSher: .net6.0 is added to all test projects: https://github.com/Azure/azure-sdk-for-net/blob/main/eng/Directory.Build.Common.props#L123-L125

christothes: Oh, right 🤦 😄

christothes (approved):

PR #32482, opened 2022-11-15
First cadl version of AnomalyDetector

### LINKED ISSUE #2840: Generate .NET client from Anomaly Detector (aka MVAD) Cadl API description
This issue tracks validating that a DPG generator can generate the MVAD client, models, convinience APIs and protocol methods from https://github.com/Azure/azure-rest-api-specs/tree/feature/anomaly-detector-cadl/specification/cognitiveservices/data-plane/AnomalyDetector/stable/v1.1/cadl, to achieve parity with what DPG v1.0 can generate from MVAD OpenAPI description.

Inputs:
https://github.com/Azure/azure-rest-api-specs/tree/feature/anomaly-detector-cadl/specification/cognitiveservices/data-plane/AnomalyDetector/stable/v1.1/cadl

pshao25:
Fix https://github.com/Azure/autorest.csharp/issues/2840


chunyu3 (commented):
Which cadl file do you use to generated this sdk? 
Shall we use the one in the PR https://github.com/Azure/azure-rest-api-specs/pull/21525 to regenerate?

pshao25:
> 

Yes, the repo info and the commit info is in sdk\anomalydetector\Azure.AI.AnomalyDetector\src\cadl-location.yaml

sdk/anomalydetector/Azure.AI.AnomalyDetector/src/Generated/Models/TimeSeriesPoint.Serialization.cs
```diff
@@ -17,12 +17,25 @@ void IUtf8JsonSerializable.Write(Utf8JsonWriter writer)
...
+                    writer.WritePropertyName("timestamp");
+                    writer.WriteStringValue(Timestamp.Value, "O");
+                }
+                else
+                {
+                    writer.WriteNull("timestamp");
+                }
             }
-            writer.WritePropertyName("value");
-            writer.WriteNumberValue(Value);
```
pshao25, line 24: These two lines are removed.

archerzz (approved):

sdk/anomalydetector/Azure.AI.AnomalyDetector/src/Generated/Models/TimeSeriesPoint.Serialization.cs
pshao25: Will fix in https://github.com/Azure/autorest.csharp/issues/2877

weshaggard:
@pshao25 why did we merge this PR while the spec PR is still in progress? Is there a reason this needed to be merged early?

pshao25:
> @pshao25 why did we merge this PR while the spec PR is still in progress? Is there a reason this needed to be merged early?

Yes, only after the code is merged, arch's comments in APIview could be reserved. Otherwise, any change in PR would make us lose those comments. So we merge it first and fix the comments from arch gradually later. That is our common practice per @m-nash.

weshaggard:
@praveenkuttappan is it true that APIView comments coming from a PR get lost if there are changes pushed to the PR? 

@pshao25 @m-nash thanks for the context. We are working towards a model where we will create PR's in the SDK repos but we should not merge them before the spec is actually merged into the specs repo. We can iterate on them as much as needed but we should try to keep them in PR form until everything is ready. If there are blocking issues to this workflow I'd definitely like to understand them.

pshao25:
> @praveenkuttappan is it true that APIView comments coming from a PR get lost if there are changes pushed to the PR?
> 
> @pshao25 @m-nash thanks for the context. We are working towards a model where we will create PR's in the SDK repos but we should not merge them before the spec is actually merged into the specs repo. We can iterate on them as much as needed but we should try to keep them in PR form until everything is ready. If there are blocking issues to this workflow I'd definitely like to understand them.

I heard about there would be a feature to keep all the comments in APIView for PR. Once that is done, we are totally OK with the workflow you mentioned above.

weshaggard:
I spoke with @praveenkuttappan about this yesterday and I believe he will be filling an issue for that.

Issue #31914, opened 2022-10-20
[FormRecognizer] Service is throttling


kinelski:
## Description

This issue collects multiple failures we've seen in the Form Recognizer test pipeline for the last couple of months. All these issues are believed to have the same underlying cause: throttling on the service side. They have all been reported to the Form Recognizer service team and are under investigation.

## Failures

All failures described below are not happening deterministically and affect multiple different tests.

### Content is not accessible
- **API version:** 2022-08-31
- **Thrown when:** in the LRO POST request when calling `BuildModel` (in `DocumentModelAdministrationClient`).
- **Frequency:** daily
- **Error message example:**
```
Azure.RequestFailedException : Invalid request.
Status: 400 (Bad Request)
ErrorCode: InvalidRequest

Content:
{
  "error": {
    "code": "InvalidRequest",
    "message": "Invalid request.",
    "innererror": {
      "code": "ContentSourceNotAccessible",
      "message": "Content is not accessible: Could not retrieve build data within 60 seconds."
    }
  }
}
```
- **Request ID reference:** 0a177c1a-3ba9-4086-af67-5401153eab4c
- [Example in the test pipeline](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1928370&view=logs&j=01d5a77e-6e85-5280-0e9f-af5629bd443f&t=1e514895-521a-57d8-4030-b9d8f4f46b20&l=290)

### Generic error during training
- **API version:** 2.1
- **Thrown when:** in the LRO GET request when polling `CreateCustomFormModelOperation`.
- **Frequency:** daily
- **Error message example:**
```
Azure.RequestFailedException : Invalid model created with ID 94ae7f2b-d7c1-4509-9df1-00f7e802d956
Status: 200 (OK)
ErrorCode: 3014

Additional Information:
error-0: 3014: Generic error during training.

Content:
```
- **Request ID reference:** 7fd3ef92-58dd-4115-b724-b5a3dfbcac55
- [Example in the test pipeline](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1928370&view=logs&j=01d5a77e-6e85-5280-0e9f-af5629bd443f&t=1e514895-521a-57d8-4030-b9d8f4f46b20&l=457)

### Could not access Azure blob storage account
- **API version:** 2.1
- **Thrown when:** in the LRO POST request when calling `StartTraining` (in `FormTrainingClient`).
[...]

Issue #28923, opened 2022-05-23
[FormRecognizer] Remove IgnoreServiceError attribute for "ContentSourceNotAccessible"


kinelski:
## Error information

Error code: ContentSourceNotAccessible
Error message: "Content is not accessible: Invalid data URL"
When: when starting a model-building LRO.
API version: v2022-01-30-preview
apim-request-id: 7d771404-5269-4099-9e51-e8ba8ca0b4d2
Example: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1575344&view=logs&j=01d5a77e-6e85-5280-0e9f-af5629bd443f&t=1e514895-521a-57d8-4030-b9d8f4f46b20&l=272

## Additional details

This is a flaky service error that doesn't happen often, but it's frequent enough to happen on almost every test pipeline run. The service team has already been notified about it ([discussion here](https://teams.microsoft.com/l/message/19:88f2d9dea2344075b5d4bc34d82a3d1c@thread.skype/1652910631452?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&parentMessageId=1652910631452&teamName=Azure%20SDK&channelName=Service%20-%20Form%20Recognizer&createdTime=1652910631452)).

This error won't make our pipeline fail anymore since it's been suppressed with the `IgnoreServiceError` attribute in our test project. The attribute is set on the class level (instead of single method) because it can happen on any test that builds a model, which includes most of our tests.

## Action items

Once the service has fixed this issue on their side, we must remove the `IgnoreServiceError` attribute from the following classes:
- `DocumentModelAdministrationClientLiveTests`

prodramp:
I am hitting the same error when I build the model with V3.0 custom model. When I access the same PDF/JSON documents from V2.1 the model does build however I want to use Table layout so V3.0 is most appropriate for me. 
When I can expect the fix of this issue? Are there any workaround to build my model? 

Thanks

kinelski:
@prodramp The bug described in this issue does not occur frequently and only happens when we put the service under a heavy load of requests. My understanding is that you were seeing it every time you try to build a model, is that right? In that case it may have a different cause.

I'm closing this issue because it will be merged into https://github.com/Azure/azure-sdk-for-net/issues/31914, but feel free to continue the discussion there if you are still blocked by it.

Issue #28913, opened 2022-05-22
[FormRecognizer] Remove IgnoreServiceError attribute for "Generic error during training"


kinelski:
## Error information

Error code: 3014
Error message: Generic error during training.
When: after a model-training LRO finishes.
API version: v2.0 ([confirmed](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1592631&view=logs&j=4fdea57c-dd88-576f-325a-03d0fa6bffa4&t=a4ecc073-dd68-51ff-f619-b7e68a44a8f8&l=565)) and possibly v2.1 (needs further investigation).
apim-request-id: bdefd7ca-7b2b-4fc6-aca0-b168ba6f9a43
Example: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1575344&view=logs&j=01d5a77e-6e85-5280-0e9f-af5629bd443f&t=1e514895-521a-57d8-4030-b9d8f4f46b20&l=85060

## Additional details

This is a flaky service error that doesn't happen often, but it's frequent enough to happen on every test pipeline run. The service team has already been notified about it ([discussion here](https://teams.microsoft.com/l/message/19:88f2d9dea2344075b5d4bc34d82a3d1c@thread.skype/1652910631452?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&parentMessageId=1652910631452&teamName=Azure%20SDK&channelName=Service%20-%20Form%20Recognizer&createdTime=1652910631452)).

This error won't make our pipeline fail anymore since it's been suppressed with the `IgnoreServiceError` attribute in our test project. The attribute is set on the class level (instead of single method) because it can happen on any test that trains a model, which includes most of our tests.

## Action items

Once the service has fixed this issue on their side, we must remove the `IgnoreServiceError` attribute from the following classes:
- `FormRecognizerSamples`
- `FormTrainingClientLiveTests`
- `RecognizeCustomFormsLiveTests`

kinelski:
Merged into https://github.com/Azure/azure-sdk-for-net/issues/28913.

PR #32487, opened 2022-11-15
[FormRecognizer] Updated FormRecongizer samples to use RecordedTestBase

kinelski:
This ensures the `IgnoreServiceError` attribute will work with samples, making our pipeline green again.

For more context, this issue describes the service errors we've been seeing in the last couple of months: https://github.com/Azure/azure-sdk-for-net/issues/31914

Issues https://github.com/Azure/azure-sdk-for-net/issues/28913 and https://github.com/Azure/azure-sdk-for-net/issues/28923 track our uses of the `IgnoreServiceError` attribute, to make sure we don't forget to remove them when the service fixes these errors.

### DIFF
--- a/sdk/formrecognizer/Azure.AI.FormRecognizer/tests/samples/DocumentAnalysisSamples.cs
+++ b/sdk/formrecognizer/Azure.AI.FormRecognizer/tests/samples/DocumentAnalysisSamples.cs
@@ -1,0 +1,14 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Azure.AI.FormRecognizer.DocumentAnalysis.Tests;
+using Azure.Core.TestFramework;
+
+namespace Azure.AI.FormRecognizer.DocumentAnalysis.Samples
+{
+    [LiveOnly]
+    [IgnoreServiceError(400, "InvalidRequest", Message = "Content is not accessible: Invalid data URL", Reason = "https://github.com/Azure/azure-sdk-for-net/issues/28923")]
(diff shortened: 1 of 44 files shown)


pallavit:
@kinelski is 'IgnoreServiceError' attribute the only thing that is available only in recorded tests? If not, what other features are different so we can identify which SDKs will likely benefit from making this change?

One other thing, part of the reason why samples were run in live test mode is to ensure samples are always current. There can potentially come a time when someone forgot to update the sample recordings and samples may become stale.

I am also curious how does this work when recordings are moved out of the repo? Do we have provision in engsys to move our sample recordings as well?

kinelski:
> There can potentially come a time when someone forgot to update the sample recordings and samples may become stale.

We're marking samples with the `LiveOnly` attribute which means that, even though they're now deriving from `RecordedTestBase`, they will only run in Live mode and won't generate recordings.

> @kinelski is 'IgnoreServiceError' attribute the only thing that is available only in recorded tests?

As far as I know it's the only thing that samples could benefit from, at least today. Other features such as automatically re-recording a test won't affect samples since they're live-only.

These features are all implemented as part of the `RecordedTest` attribute ([here](https://github.com/Azure/azure-sdk-for-net/blob/e0311f1a46aee4d33d39ad2d29e63986bd26fedf/sdk/core/Azure.Core.TestFramework/src/RecordedTestAttribute.cs#L46)), which only works with test classes deriving from `RecordedTestBase`.

Another possibility would be trying to make the `IgnoreServiceError` attribute work with the `LiveTestBase` class as well, but @JoshLove-msft and I were wondering if there's any point in having two separate "TestBase" classes in the first place. It may be worth discussing if we want to merge them.

pallavit:
Got it, thank you for the detailed response.

pallavit (approved):

PR #32493, opened 2022-11-15
[Text Analytics] Move `Warnings` to `ExtractSummaryResult` and remove `SummarySentenceCollection` 

### LINKED ISSUE #32434: [Text Analytics] Move Warnings property from `SummarySentenceCollection` to `ExtractSummaryResult`
The `ExtractSummaryResult` class has a property called `Sentences` of type `SummarySentenceCollection`, which in turn has a property called `Warnings` of type `IReadOnlyCollection<TextAnalyticsWarning>`. In order to align with the existing patterns of the library, the `Warnings` property must be moved up into the `ExtractSummaryResult` class.

joseharriaga:
Resolves: https://github.com/Azure/azure-sdk-for-net/issues/32434

Also taking the chance to brush up the documentation related to extractive summarization.


kristapratico (approved):
LGTM

sdk/textanalytics/Azure.AI.TextAnalytics/api/Azure.AI.TextAnalytics.netstandard2.0.cs
```diff
@@ -1736,9 +1732,9 @@ public static partial class TextAnalyticsModelFactory
...
         public static Azure.AI.TextAnalytics.ExtractSummaryActionResult ExtractSummaryActionResult(Azure.AI.TextAnalytics.ExtractSummaryResultCollection result, string actionName, System.DateTimeOffset completedOn) { throw null; }
         public static Azure.AI.TextAnalytics.ExtractSummaryActionResult ExtractSummaryActionResult(string actionName, System.DateTimeOffset completedOn, string code, string message) { throw null; }
-        public static Azure.AI.TextAnalytics.ExtractSummaryResult ExtractSummaryResult(string id, Azure.AI.TextAnalytics.TextDocumentStatistics statistics, Azure.AI.TextAnalytics.SummarySentenceCollection sentences) { throw null; }
+        public static Azure.AI.TextAnalytics.ExtractSummaryResult ExtractSummaryResult(string id, Azure.AI.TextAnalytics.TextDocumentStatistics statistics, System.Collections.Generic.IList<Azure.AI.TextAnalytics.SummarySentence> sentences, System.Collections.Generic.IList<Azure.AI.TextAnalytics.TextAnalyticsWarning> warnings = null) { throw null; }
```
mssfang, line 1735: Should this use the IEnumerable type?

joseharriaga: In this case, this parameter is used to the `Sentences` property of the `ExtractSummaryResult` model. Because this is the property of a model, it uses `IList` instead of `IEnumerable` as per the guidelines here:
https://azure.github.io/azure-sdk/dotnet_introduction.html#dotnet-service-return-model-collections

mssfang: So in the output model, .NET uses either of IReadOnlyList or IList. But can use the more generic IEnumerable type for the input model?

mssfang (approved):

sdk/textanalytics/Azure.AI.TextAnalytics/api/Azure.AI.TextAnalytics.netstandard2.0.cs
joseharriaga: For both, input and output model properties, .NET prefers `IReadOnlyList` or `IList`. But for the method parameters in any other classes (e.g., client classes, helper classes, etc.), the use of `IEnumerable` is allowed if appropriate.

sdk/textanalytics/Azure.AI.TextAnalytics/src/ExtractSummaryAction.cs
```diff
@@ -4,29 +4,22 @@
...
         }
 
         /// <summary>
-        /// Initializes a new instance of the <see cref="ExtractSummaryAction"/>
-        /// class based on the values of a <see cref="ExtractSummaryOptions"/>.
-        /// It sets the <see cref="ModelVersion"/>, <see cref="DisableServiceLogs"/>,
-        /// <see cref="MaxSentenceCount"/>, and <see cref="OrderBy"/> properties.
+        /// Initializes a new instance of the <see cref="ExtractSummaryAction"/> class based on the given
+        /// <see cref="ExtractSummaryOptions"/>. It sets the <see cref="ModelVersion"/>,
+        /// <see cref="DisableServiceLogs"/>, <see cref="MaxSentenceCount"/>, and <see cref="OrderBy"/> properties.
```
heaths, line 22: IMO, no need to talk about the implementation that may change and the comment falls out of date. Generally, "Initialize a new instance..." doesn't go into details, not even for the .NET BCL itself.

sdk/textanalytics/Azure.AI.TextAnalytics/src/ExtractSummaryOptions.cs
```diff
@@ -4,14 +4,13 @@
 namespace Azure.AI.TextAnalytics
 {
     /// <summary>
-    /// Options that allow callers to specify details about how the operation
-    /// is run and what information is returned from it by the service.
+    /// A set of options used to configure extractive summarization, including the display name to use, the maximum
```
heaths, line 7: Nit:
```suggestion
    /// A set of options used to configure extractive summarization including the display name to use, the maximum
```

heaths (approved):

sdk/textanalytics/Azure.AI.TextAnalytics/src/ExtractSummaryOptions.cs
joseharriaga: I think the comma is correct in this case because everything that comes after it could technically be removed without changing the meaning.

PR #32499, opened 2022-11-15
Remove model generation from CLU tests

heaths:


heaths:
I don't often need to regenerate anyway, so hopefully this allays some concerns.

m-nash (approved):

PR #32507, opened 2022-11-16
Add net6.0 target framework to Azure.Storage

AlexanderSher:
Tracking issue to restore ApiCompat https://github.com/Azure/azure-sdk-for-net/issues/32508


jaschrep-msft:
Odd live tests failures. Tests are unrelated to changes and haven't been updated for over 2 years. @seanmcc-msft are these tests we've seen flake before in live? It doesn't seem like they should, as they're just using SAS without the required permission to induce an error.

christothes (approved):

PR #32520, opened 2022-11-16
Fix instance metadata validation error

### LINKED ISSUE #32498: Azure.Identity Question
### Library name and version

Azure.Identity 1.8.0

### Query/Question

I am able to use v1.7.0 of Azure.Identity with no problems.  My situation is I have an AKS cluster using pod identity which is configured to a user assigned managed identity.  I use the Microsoft.Extensions.Azure library to configure all of my Azure clients such as Blob client, ServiceBus client, etc. 

While using v1.7.0, occasionally my application may attempt to bootstrap before the identity is available to the pod to use and gets the occasional error only during startup which is fine because the pod won't be considered ready until it successfully starts.

However, with v1.8.0 I get stuck in an infinite error state and get stuck with millions of logged failures that look like the following:

`ManagedIdentityCredential authentication failed: The json containing instance metadata does not contain details about the authority in use: login.microsoftonline.us. See https://aka.ms/msal-net-custom-instance-metadata for more details. 
See the troubleshooting guide for more information. https://aka.ms/azsdk/net/identity/managedidentitycredential/troubleshoot The json containing instance metadata does not contain details about the authority in use: login.microsoftonline.us. See https://aka.ms/msal-net-custom-instance-metadata for more details.`

The links are of no help.  I'm not sure what changed between 1.7.0 and 1.8.0, but could someone please explain what was changed and why this may be occurring?

### Environment

_No response_

christothes:
fixes #32498


sdk/identity/Azure.Identity/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### Bugs Fixed
 - Fixed error message parsing in `AzureCliCredential` which would misinterpret AAD errors with the need to login with `az login`.
 - `ManagedIdentityCredential` will no longer fail when a response received from the endpoint is invalid JSON. It now treats this scenario as if the credential is unavailable.
+- Fixed an issue when using ManagedIdentityCredential in combindation with authorities other than Azure public cloud that resulted in a bogus instance metadata validation error [#32498](https://github.com/Azure/azure-sdk-for-net/issues/32498).
```
scottaddie, line 12:
```suggestion
- Fixed an issue when using `ManagedIdentityCredential` in combination with authorities other than Azure public cloud that resulted in a bogus instance metadata validation error [#32498](https://github.com/Azure/azure-sdk-for-net/issues/32498).
```

amirschw:
Thanks for fixing this, @christothes! Any idea when there's going to be a new release that includes this fix? We're waiting for it to unblock our migration from AppAuthentication to Azure.Identity (we can't use the latest version because of this issue nor can we use previous versions since they don't have internal token caching for ManagedIdentityCredential).

christothes:
> Thanks for fixing this, @christothes! Any idea when there's going to be a new release that includes this fix? We're waiting for it to unblock our migration from AppAuthentication to Azure.Identity (we can't use the latest version because of this issue nor can we use previous versions since they don't have internal token caching for ManagedIdentityCredential).

Should be in January.

schaabs (approved):

snakefoot:
Seems #32498 has [not been fixed completely](https://github.com/Azure/azure-sdk-for-net/issues/32498#issuecomment-1382669582)

PR #32525, opened 2022-11-17
WIP: Use operator precedence to avoid unnecessary parentheses and nesting

gliljas:
When the LINQ expression is converted to a query string, lots of parentheses are included, for good measure. It's certainly correct, but:

1. It increases the length, sometimes significantly. 
2. The parser on the receiving end doesn't like the nesting.

This PR was initiated by what happened when we moved from the old Cosmos library to this. We had a very large, dynamically generated query (one partitionkey, lots of suggested rowkeys), which utilized the old TableQuery class to generate the query string. We converted this to a LINQ expression but ran into issues in production, when the table service responded with a Bad Request: "Recursion depth exceeded allowed limit". 
We realized this was due to all the "RowKey='a' or (RowKey='b' or (RowKey='c' .....", where every binary expression is encapsulated in parentheses. An OData "in" would have been nicer, but the table service doesn't support it.

The PR attempts to check operator precedence in order to insert parentheses only when it's necessary.

### DIFF
--- a/sdk/tables/Azure.Data.Tables/src/Queryable/ExpressionPrecedenceComparer.cs
+++ b/sdk/tables/Azure.Data.Tables/src/Queryable/ExpressionPrecedenceComparer.cs
@@ -1,0 +1,110 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Collections.Generic;
+using System.Linq.Expressions;
+
+namespace Azure.Data.Tables.Queryable
+{
+    internal sealed class ExpressionPrecedenceComparer : IComparer<Expression>
+    {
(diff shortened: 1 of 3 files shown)


gliljas:
I didn't know what to do with the recorded tests, so.... they fail.

christothes:
> I didn't know what to do with the recorded tests, so.... they fail.

In order to setup a live test environment, you can run the [New-TestResources script](https://github.com/Azure/azure-sdk-for-net/blob/main/eng/common/TestResources/New-TestResources.ps1). More details [here](https://github.com/Azure/azure-sdk-for-net/blob/main/eng/common/TestResources/README.md).

From the root of the enlistment, you'd run it like so:
```
.\eng\common\TestResources\New-TestResources.ps1 tables
```

gliljas:
I'll get back to this now. I think it's a good fix to do.

christothes:
> I'll get back to this now. I think it's a good fix to do.

Thanks for spending time on this - please let me know if you run into any issues with the tests.

PR #32527, opened 2022-11-17
Support generic list in DataFactoryExpression

JoshLove-msft:


sdk/core/Azure.Core.Expressions.DataFactory/src/DataFactoryExpression.cs
```diff
@@ -14,21 +14,22 @@ namespace Azure.Core.Expressions.DataFactory
...
+        internal T? LiteralInternal { get; }
+        internal string? Expression { get; }
 
         /// <summary>
         /// Initializes a new instance of DataFactoryExpression with a literal value.
         /// </summary>
-        /// <param name="literal"> The literal value. </param>
-        public DataFactoryExpression(T literal)
+        /// <param name="literalInternal"> The literal value. </param>
+        public DataFactoryExpression(T literalInternal)
```
m-nash, line 29: not sure we want this rename on this public parameter, lets keep it `literal`.

sdk/core/Azure.Core.Expressions.DataFactory/src/DataFactoryExpression.cs
```diff
@@ -14,21 +14,22 @@ namespace Azure.Core.Expressions.DataFactory
...
-    public sealed partial class DataFactoryExpression<T> : IUtf8JsonSerializable
+    [JsonConverter(typeof(DataFactoryExpressionJsonConverter))]
+    public sealed class DataFactoryExpression<T>
 #pragma warning restore SA1649 // File name should match first type name
     {
-        private string? _type;
-        private T? _literal;
-        private string? _expression;
+        internal string? Type { get; }
+        internal T? LiteralInternal { get; }
```
m-nash, line 22: I would still just call this `Literal`

sdk/core/Azure.Core.Expressions.DataFactory/src/DataFactoryExpressionJsonConverter.cs
```diff
@@ -11,32 +15,157 @@ internal class DataFactoryExpressionJsonConverter : JsonConverter<object>
...
+            writer.WritePropertyName("value");
+            writer.WriteStringValue(value);
+            writer.WriteEndObject();
+        }
+
+        internal static DataFactoryExpression<List<T>> DeserializeList<T>(JsonElement element)
+        {
+            if (element.ValueKind == JsonValueKind.Array)
+            {
+                var list = new List<T>();
```
m-nash, line 131: Consider if we should use changetrackinglist here

sdk/core/Azure.Core.Expressions.DataFactory/src/DataFactoryExpressionJsonConverter.cs
```diff
@@ -11,32 +15,157 @@ internal class DataFactoryExpressionJsonConverter : JsonConverter<object>
...
+                foreach (var item in element.EnumerateArray())
+                {
+                    list.Add(JsonSerializer.Deserialize<T>(item.ToString()!)!);
+                }
+
+                return new DataFactoryExpression<List<T>>(list);
+            }
+
+            throw new InvalidOperationException("Can only be called on an array");
```
m-nash, line 140: "Attempted to deserialize a {element.ValueKind} as a List" ?

sdk/core/Azure.Core.Expressions.DataFactory/src/DataFactoryExpressionJsonConverter.cs
```diff
@@ -11,32 +15,157 @@ internal class DataFactoryExpressionJsonConverter : JsonConverter<object>
...
+        }
+
+        internal static DataFactoryExpression<List<T>> DeserializeList<T>(JsonElement element)
+        {
+            if (element.ValueKind == JsonValueKind.Array)
+            {
+                var list = new List<T>();
+                foreach (var item in element.EnumerateArray())
+                {
+                    list.Add(JsonSerializer.Deserialize<T>(item.ToString()!)!);
```
m-nash, line 134: Are the `!`'s safe?  Is it possible an item.ToString() is `null`?  Is the second `!` even need since the api doesn't allow for nullable reference?

sdk/core/Azure.Core.Expressions.DataFactory/src/DataFactoryExpressionJsonConverter.cs
```diff
@@ -11,32 +15,157 @@ internal class DataFactoryExpressionJsonConverter : JsonConverter<object>
...
+                return Deserialize<double>(document.RootElement);
             if (typeToConvert == typeof(DataFactoryExpression<Array>))
-                return DataFactoryExpression<Array>.DeserializeDataFactoryExpression(document.RootElement);
+                return Deserialize<Array>(document.RootElement);
             if (typeToConvert == typeof(DataFactoryExpression<bool>))
-                return DataFactoryExpression<bool>.DeserializeDataFactoryExpression(document.RootElement);
+                return Deserialize<bool>(document.RootElement);
+            if (IsGenericDataFactoryList(typeToConvert))
+            {
+                var methodInfo = typeof(DataFactoryExpressionJsonConverter)
```
m-nash, line 41: maybe share this this code with the Write method only diff seems to the be the name of the method

sdk/core/Azure.Core.Expressions.DataFactory/tests/DataFactoryExpressionsTests.cs
```diff
@@ -259,10 +274,27 @@ private static void AssertBoolDfe(DataFactoryExpression<bool> dfe)
         public void DeserailizationOfStringValue()
         {
             var doc = JsonDocument.Parse(StringJson);
-            var dfe = DataFactoryExpression<string>.DeserializeDataFactoryExpression(doc.RootElement);
+            var dfe = DataFactoryExpressionJsonConverter.Deserialize<string>(doc.RootElement);
             AssertStringDfe(dfe, StringValue);
         }
 
+        [Test]
+        public void DeserializationOfListOfT()
```
m-nash, line 282: Can we add a case to validate the case where we call DataFactoryExpression<List<TestModel>>.FromExpression("blah")?

I assume the json should either be the literal array OR the normal expression object?  Should we validate this with the service team?

sdk/core/Azure.Core.Expressions.DataFactory/tests/DataFactoryExpressionsTests.cs
```diff
@@ -189,6 +191,19 @@ public void SerailizationOfStringValue()
             Assert.AreEqual(StringJson, actual);
         }
 
+        [Test]
+        public void SerailizationOfListOfT()
+        {
+            var elements = new List<TestModel>
+            {
+                new TestModel { A = 1, B = true },
+                new TestModel { A = 2, B = false }
```
m-nash, line 200: can we add a test case for Empty list, null list, and possible a list with one of the items null?

sdk/core/Azure.Core.Expressions.DataFactory/src/DataFactoryExpression.cs
JoshLove-msft: woops this was unintended

sdk/core/Azure.Core.Expressions.DataFactory/src/DataFactoryExpressionJsonConverter.cs
JoshLove-msft: It seems like we probably should - this is used as the implementation for all generated models exposing `IList<T>` or `IReadOnlyList<T>`. We intend for this type to be used in generated models. I think the main purpose of the type is to allow exposing libraries not to have to add setters for nullable collection properties.

sdk/core/Azure.Core.Expressions.DataFactory/src/DataFactoryExpression.cs
JoshLove-msft: There is already a public literal property. The difference is that this property does not check `HasLiteral`

sdk/core/Azure.Core.Expressions.DataFactory/src/DataFactoryExpressionJsonConverter.cs
JoshLove-msft: I don't think ToString can be null in this case since we would be dealing with JSON objects in the array.

sdk/core/Azure.Core.Expressions.DataFactory/tests/DataFactoryExpressionsTests.cs
JoshLove-msft: Checked with service team - the expression object would have the same structure. Will add a test.

sdk/core/Azure.Core.Expressions.DataFactory/src/DataFactoryExpressionJsonConverter.cs
JoshLove-msft: It seems it can be null, so I've updated to use a `List<T?>`

sdk/core/Azure.Core.Expressions.DataFactory/src/DataFactoryExpressionJsonConverter.cs
```diff
@@ -2,41 +2,238 @@
...
+                   type.GenericTypeArguments[0].GenericTypeArguments[0].GetCustomAttributes().Any(a => a.GetType() == typeof(JsonConverterAttribute));
+        }
+
+        private static bool IsGenericListType(Type type) => type == typeof(IList<>) || type == typeof(IReadOnlyList<>);
+
+        private static void Serialize<T>(Utf8JsonWriter writer, DataFactoryExpression<T?> expression)
+        {
+            if (expression.HasLiteral)
+            {
+                writer.WriteObjectValue(expression.LiteralInternal!);
```
JoshLove-msft, line 115: The nullability in WriteObjectValue is apparently incorrect - it does support nulls: https://github.com/Azure/autorest.csharp/blob/feature/v3/src/assets/Generator.Shared/Utf8JsonWriterExtensions.cs#L69

sdk/core/Azure.Core.Expressions.DataFactory/src/DataFactoryExpressionJsonConverter.cs
```diff
@@ -2,41 +2,238 @@
...
+            {
+                Type typeToConvert = value.GetType();
+                if (IsGenericDataFactoryList(typeToConvert))
+                {
+                    string methodName = typeToConvert.GenericTypeArguments[0].GetGenericTypeDefinition() == typeof(IList<>)
+                        ? nameof(SerializeList)
+                        : nameof(SerializeReadOnlyList);
+                    var methodInfo = GetGenericSerializationMethod(typeToConvert, methodName);
+                    methodInfo!.Invoke(null, new object[] { writer, value });
+                }
```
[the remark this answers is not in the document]
JoshLove-msft: Yeah I considered that initially, essentially just passing back the type via an out param. It seemed a little weird, but you are right it would make the overall if/else pattern cleaner.

sdk/core/Azure.Core.Expressions.DataFactory/src/DataFactoryExpressionJsonConverter.cs
```diff
@@ -2,41 +2,238 @@
...
+                return null;
+            }
+
+            if (element.ValueKind == JsonValueKind.Object)
+            {
+                foreach (var property in element.EnumerateObject())
+                {
+                    if (property.NameEquals("value"))
+                    {
+                        expression = property.Value.GetString();
```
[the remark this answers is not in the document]
JoshLove-msft: Eh, I don't know what STJ behavior is for duplicate key values.

sdk/core/Azure.Core.Expressions.DataFactory/src/DataFactoryExpressionJsonConverter.cs
```diff
@@ -2,41 +2,210 @@
...
+        internal static DataFactoryExpression<T>? Deserialize<T>(JsonElement element)
+        {
+            T? value = default;
+
+            if (element.ValueKind == JsonValueKind.Null)
+            {
+                return null;
+            }
+
+            if (element.ValueKind == JsonValueKind.Object)
```
JoshLove-msft, line 183: For now at least, we have the invariant that Json Object => expression. In theory this could change so we would probably need to inspect the payload to determine if it is an expression or not.

christothes (approved):

sdk/core/Azure.Core.Expressions.DataFactory/src/DataFactoryExpressionJsonConverter.cs
JoshLove-msft: It turns out we probably don't need this. The only point of this type would be if we wanted to avoid sending an entire JSON property for a `DFE<IList<T>>` if the list empty. The service team wants an empty array in this case so just using `List<T>` makes more sense.

sdk/core/Azure.Core.Expressions.DataFactory/src/DataFactoryExpressionJsonConverter.cs
m-nash: Lets leave a TODO here please

m-nash (approved):

PR #32540, opened 2022-11-17
Support cadl generation from dotnet language repo

m-nash:


eng/scripts/Cadl-Project-Sync.ps1
```diff
@@ -0,0 +1,101 @@
...
+        {
+            $source = "$specCloneRoot/$additionalDir"
+            Write-Host "Copying spec from $source"
+            Copy-Item -Path $source -Destination $dest -Recurse -Force
+        }
+    }
+}
+
+function UpdateSparseCheckoutFile([string]$mainSpecDir, [string[]]$specAdditionalSubDirectories) {
+    $sparseCheckoutFile = ".git/info/sparse-checkout"
```
benbp, line 48: This isn't used in the underlying function, and I think it's ok to keep hardcoding the path in the underlying function as well.

Also, are we guaranteed that this will always run in a sparse git directory? If not, we should only append to the file if it exists.

eng/scripts/Cadl-Project-Sync.ps1
```diff
@@ -0,0 +1,101 @@
...
+    Pop-Location
+    $sparseSpecCloneDir = "$root/../temp/$projectName"
+    New-Item $sparseSpecCloneDir -Type Directory -Force | Out-Null
+    $createResult = Resolve-Path $sparseSpecCloneDir
+    return $createResult
+}
+
+$cadlConfigurationFile = Resolve-Path "$ProjectDirectory/src/cadl-location.yaml"
+Write-Host "Reading configuration from $cadlConfigurationFile"
+$configuration = Get-Content -Path $cadlConfigurationFile -Raw | ConvertFrom-Yaml
```
benbp, line 76: This is often not installed on machines or pipeline agents by default. You can put something like this at the top of the file after parameter declarations:

```
. $PSScriptRoot/../common/scripts/Helpers/PSModule-Helpers.ps1

Install-ModuleIfNotInstalled "powershell-yaml" "0.4.1" | Import-Module
```

eng/scripts/Cadl-Project-Sync.ps1
```diff
@@ -0,0 +1,101 @@
...
+
+$pieces = $cadlConfigurationFile.Path.Split("\")
```
benbp, line 78: Just to handle all OS paths that may show up:

```suggestion
$pieces = $cadlConfigurationFile.Path.Replace("\","/").Split("\")
```

eng/scripts/Cadl-Project-Sync.ps1
```diff
@@ -0,0 +1,101 @@
...
+    {
+        AddSparseCheckoutPath $subDir $sparseCheckoutFile
+    }
+}
+
+function InitializeSparseGitClone([string]$repo) {
+    if (!(Test-Path ".git")) {
+        git init
+        git remote add origin $repo
+        git config core.sparseCheckout true
```
benbp, line 60:
```suggestion
        git init
        if ($LASTEXITCODE) { exit $LASTEXITCODE }
        git remote add origin $repo
        if ($LASTEXITCODE) { exit $LASTEXITCODE }
        git config core.sparseCheckout true
        if ($LASTEXITCODE) { exit $LASTEXITCODE }
```

eng/scripts/Cadl-Project-Sync.ps1
```diff
@@ -0,0 +1,101 @@
...
+$specCloneDir = GetSpecCloneDir $projectName
+
+Write-Host "Setting up sparse clone for $projectName"
+Push-Location $specCloneDir.Path
+InitializeSparseGitClone $configuration["repo"]
+
+$specSubDirectory = $configuration["directory"]
+UpdateSparseCheckoutFile $specSubDirectory $configuration["additionalDirectories"]
+
+$result = (git pull origin $configuration["commit"])
```
benbp, line 90: Does this work out of the box, or am I missing something?

```
$ git init
$ git remote add origin git@github.com:azure/azure-sdk-for-net.git
$ git config core.sparseCheckout true
$ "eng" >> ./.git/info/sparse-checkout
$ git pull origin 66d1dab3e7
fatal: couldn't find remote ref 66d1dab3e7
```

Also, related, I think it may be more future-proof to use the git sparse-checkout subcommands over manipulating the config (this will also error if the script is using an older git client that doesn't support these features):

```
git sparse-checkout init
git sparse-checkout add $subDirectory
```

eng/scripts/Cadl-Project-Sync.ps1
```diff
@@ -0,0 +1,101 @@
...
+Write-Host $result
+Pop-Location
```
benbp, line 92: Generally with push/pop it is good to use a `try/finally` so when we error we don't drop the user or subsequent scripts into the wrong location.

eng/scripts/Cadl-Project-Sync.ps1
```diff
@@ -0,0 +1,101 @@
...
+    Push-Location $workingDirectory
+
+    $currentDur = Resolve-Path "."
+    Write-Host "Generating from $currentDur"
+    if (Test-Path ".\package.json") {
+        Remove-Item -Path ".\package.json" -Force
+    }
+    if (Test-Path ".\.npmrc") {
+        Remove-Item -Path ".\.npmrc" -Force
+    }
```
benbp, line 26:
```suggestion
    if (Test-Path "package.json") {
        Remove-Item -Path "package.json" -Force
    }
    if (Test-Path ".npmrc") {
        Remove-Item -Path ".npmrc" -Force
    }
```

eng/scripts/Cadl-Project-Sync.ps1
```diff
@@ -0,0 +1,101 @@
...
+    npm install "@azure-tools/cadl-csharp"
```
benbp, line 27:
```suggestion
    npm install "@azure-tools/cadl-csharp"
    if ($LASTEXITCODE) { exit $LASTEXITCODE }
```

eng/scripts/Cadl-Project-Sync.ps1
```diff
@@ -0,0 +1,101 @@
...
+
+$result = (git pull origin $configuration["commit"])
+Write-Host $result
+Pop-Location
+
+$specDir = Resolve-Path "$specCloneDir/$specSubDirectory"
+
+$tempCadlDir = "$ProjectDirectory/temp"
+New-Item $tempCadlDir -Type Directory -Force | Out-Null
+CopySpecToProjectIfNeeded $result $specCloneDir $specSubDirectory $tempCadlDir $configuration["additionalDirectories"]
```
benbp, line 98:
```suggestion
CopySpecToProjectIfNeeded `
    -pullRequest $result `
    -specCloneRoot $specCloneDir `
    -mainSpecDir $specSubDirectory `
    -dest $tempCadlDir `
    -specAdditionalSubDirectories $configuration["additionalDirectories"]
```

eng/scripts/Cadl-Project-Sync.ps1
```diff
@@ -0,0 +1,101 @@
...
+
+$innerFolder = Split-Path $specDir -Leaf
+NpmInstallForProject (Resolve-Path "$tempCadlDir\$innerFolder")
```
benbp, line 101:
```suggestion
NpmInstallForProject (Resolve-Path "$tempCadlDir/$innerFolder")
```

eng/scripts/Cadl-Project-Sync.ps1
```diff
@@ -0,0 +1,101 @@
...
+$ErrorActionPreference = "Stop"
+
+function AddSparseCheckoutPath([string]$subDirectory, [string]$file) {
+    if (!(Test-Path $file) -or !((Get-Content $file).Contains($subDirectory))) {
+        Write-Output $subDirectory >> .git/info/sparse-checkout
+    }
+}
+
+function NpmInstallForProject([string]$workingDirectory) {
+    Push-Location $workingDirectory
```
heaths, line 17: Whenever pushing locations, right below it use:

```powershell
trap { Pop-Location }
```

This is an old school try..finally mechanism in PowerShell that makes sure, if errors terminate the current scope, code runs. If you don't and errors occur, your script will be stuck in this directory and bad things can happen.

eng/scripts/Cadl-Project-Sync.ps1
heaths: For now, yes. Just an FYI, though: this is now automatic in PowerShell 7.3: "native" commands still set `$LASTEXITCODE` but that still works like `$?` and friends now, so the `$ErrorActionPreference` should work regardless. Again, though, for now this is a good suggestion.

eng/scripts/Cadl-Project-Sync.ps1
m-nash: it does work out of the box.  Did you try running the script or just those commands directly?

benbp: Yep I should have put the full commit. Script works for me, outside of the git SSH-based auth/remote issues we discussed.

eng/scripts/Cadl-Project-Sync.ps1
```diff
@@ -0,0 +1,123 @@
...
+
+Write-Host "Setting up sparse clone for $projectName"
+Push-Location $specCloneDir.Path
+try {
+    InitializeSparseGitClone $configuration["repo"]
+
+    $specSubDirectory = $configuration["directory"]
+    UpdateSparseCheckoutFile $specSubDirectory $configuration["additionalDirectories"]
+
+    $result = (git pull origin $configuration["commit"])
```
benbp, line 104: So just playing around with this locally with a larger repo like azure-sdk-for-net, running the pull command still downloads all the git objects and takes a very long time. For the specs repo it still downloads 137MB of objects even if it only shows the directories you specify for sparse.

The way we do it in our pipelines is much faster, by specifing to only get the commits from the latest tree when we set up the remote. See [here](https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/pipelines/templates/steps/sparse-checkout.yml#L43)


```
git clone --no-checkout --filter=tree:0 -c remote.origin.fetch="$($configuration.commit):refs/remotes/origin/$($configuration.commit" $configuration.repo .
git sparse-checkout init
git sparse-checkout add <sub directories>
```

eng/scripts/Cadl-Project-Sync.ps1
```diff
@@ -0,0 +1,152 @@
+[CmdletBinding()]
```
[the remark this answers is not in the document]
m-nash: No we would need to add in some "offline" mode.  Right now it will always try to pull the changes in for the commit sha since it doesn't know if the local files are up to date or not.

we could do this by some flag in the cadl-location.yaml such as `commit: local` to indicate don't sync just use my local files.

Do you mind filing an issue for this and we can address after this PR unless you think its a requirement for v1.  Trying to keep this to MVP as the dpg timelines are very sensitive right now.

eng/scripts/Cadl-Project-Sync.ps1
m-nash: @benbp I don't think this perf improvement is needed for v1 as the current process does limit the sync and works.  If its ok with you given our timelines for dpg it would be helpful if you can open an issue to track this and we can get any additional perf improvements in a later PR.

m-nash: Was able to get this working in this commit https://github.com/Azure/azure-sdk-for-net/pull/32540/commits/f252df07651d2239aa2f78b43a13cf1e0ea7cd41

.gitignore
```diff
@@ -172,4 +172,7 @@ msbuild.binlog
 BenchmarkDotNet.Artifacts
 
 artifacts
-.assets
\ No newline at end of file
+.assets
+
+# Temporary cadl folders for cadl generation
+TempCadlFiles/
```
weshaggard, line 178: Any reason to put these in a new top level folder instead of under artifacts like everything else?

eng/scripts/Cadl-Project-Sync.ps1
```diff
@@ -0,0 +1,121 @@
...
+    }
+    finally {
+        Pop-Location
+    }
+
+    return $result
+}
+
+function InitializeSparseGitClone([string]$repo) {
+    git clone --no-checkout --filter=tree:0 $repo .
```
weshaggard, line 68: In order to get this to work for the private repo we are going to either need to use a PAT from our bot account or perhaps figure out how to use the credentials from the agent. @benbp did you ever do any investigation with the persistCredentials option https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/steps-checkout?view=azure-pipelines while doing your sparse checkout work?

eng/scripts/Cadl-Project-Sync.ps1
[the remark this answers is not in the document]
weshaggard: We could potentially add some configuration option that allows folks to point at a local specs repo clone as I do think that will be an interesting scenario even outside of the offline scenario.

sdk/adp/Azure.AI.AutonomousDevelopmentPlatform/src/cadl-location.yaml
```diff
@@ -0,0 +1,6 @@
+directory: specification/adp/DataManagement
+additionalDirectories:
```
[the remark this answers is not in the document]
m-nash: it won't always be in the same service directory.  We will need the ability to point at additional sub directories.

.gitignore
m-nash: not sure I follow, they are currently inside each project folder, I have the ignore so they are never accidentally checked in.

sdk/adp/Azure.AI.AutonomousDevelopmentPlatform/src/cadl-location.yaml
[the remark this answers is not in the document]
weshaggard: My assumption is that we should block source references outside of the same service folder but you are correct that we might not be able to do that in all cases.

eng/scripts/Cadl-Project-Sync.ps1
benbp: I think the approach currently used in test proxy where we filter on the PAT existing and change the repo url accordingly could work here as well. I recall initially trying to get pipeline sparse checkout with persist credentials working, but the format/encoding of the token was set up in some way where I couldn't figure out how to consume it. Some working examples may have been posted [here](https://github.com/Azure/azure-sdk-tools/issues/3222#issuecomment-1111116452).

eng/scripts/Cadl-Project-Generate.ps1
```diff
@@ -0,0 +1,58 @@
...
+
+Write-Host "Reading configuration from $cadlConfigurationFile"
+$configuration = Get-Content -Path $cadlConfigurationFile -Raw | ConvertFrom-Yaml
+
+$specSubDirectory = $configuration["directory"]
+$innerFolder = Split-Path $specSubDirectory -Leaf
+
+$tempFolder = "$ProjectDirectory/TempCadlFiles"
+$npmWorkingDir = Resolve-Path $tempFolder/$innerFolder
+$mainCadlFile = Resolve-Path "$npmWorkingDir/main.cadl"
```
pshao25, line 42: We might need another parameter to indicate the entry cadl file. It doesn't have to be `main.cadl`. `client.cadl` is a common practice.

eng/scripts/Cadl-Project-Sync.ps1
```diff
@@ -0,0 +1,120 @@
...
+Install-ModuleIfNotInstalled "powershell-yaml" "0.4.1" | Import-Module
+$sparseCheckoutFile = ".git/info/sparse-checkout"
+
+function AddSparseCheckoutPath([string]$subDirectory) {
+    if (!(Test-Path $sparseCheckoutFile) -or !((Get-Content $sparseCheckoutFile).Contains($subDirectory))) {
+        Write-Output $subDirectory >> .git/info/sparse-checkout
+    }
+}
+
+function CopySpecToProjectIfNeeded([string]$specCloneRoot, [string]$mainSpecDir, [string]$dest, [string[]]$specAdditionalSubDirectories) {
```
pshao25, line 19: I think we should clean the folder before copying.

eng/scripts/Cadl-Project-Generate.ps1
```diff
@@ -0,0 +1,58 @@
...
+$resolvedProjectDirectory = Resolve-Path "$ProjectDirectory/src"
+
+try {
+    Push-Location $npmWorkingDir
+    NpmInstallForProject $npmWorkingDir
+    Write-Host("npx cadl compile $mainCadlFile --emit `"`@azure-tools/cadl-csharp`" --output-path `"$resolvedProjectDirectory`" --option @azure-tools/cadl-csharp.clear-output-folder=true")
+    npx cadl compile $mainCadlFile --emit "@azure-tools/cadl-csharp" --output-path "$resolvedProjectDirectory" --option @azure-tools/cadl-csharp.clear-output-folder=true
```
pshao25, line 49: Why the options are hard-coded here? Shouldn't it be in the cadl-project.yaml?

pshao25: And my understanding is there should be logic some where to move all the parts only leaving "@azure-tools/cadl-csharp" part. Is this feature included?

.gitignore
archerzz: It's because `cleanup: false` in `cadl-location.yaml`. I remove that property in another PR, and `TempCadlFiles` will be removed after `dotnet build /t:GenerateFiles`

PR #32541, opened 2022-11-17
Switch to script over pwsh to enable command logging

weshaggard:
If we use the script block instead of pwsh the DevOPs task will output the script command line in the logs which is very helpful when trying to reproduce a test failure locally.

As part of this also need to switch to double quotes as single quotes don't work in the script block correctly.

### DIFF
--- a/eng/pipelines/templates/jobs/ci.tests.yml
+++ b/eng/pipelines/templates/jobs/ci.tests.yml
@@ -87,11 +87,11 @@ jobs:
             Write-Host "##vso[task.setvariable variable=CodeCoverageFilePattern]$(Build.SourcesDirectory)\sdk\${{parameters.ServiceDirectory}}\**\coverage.cobertura.xml"
           }
         displayName: Set variable for the project list file and coverage directory
-      - pwsh: >-
+      - script: >-
           dotnet test eng/service.proj
-          --filter '(TestCategory!=Manually) & (TestCategory!=Live) & ($(AdditionalTestFilters))'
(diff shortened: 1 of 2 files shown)


eng/pipelines/templates/jobs/ci.tests.yml
```diff
@@ -87,11 +87,11 @@ jobs:
             Write-Host "##vso[task.setvariable variable=CodeCoverageFilePattern]$(Build.SourcesDirectory)\sdk\${{parameters.ServiceDirectory}}\**\coverage.cobertura.xml"
           }
         displayName: Set variable for the project list file and coverage directory
-      - pwsh: >-
+      - script: >-
           dotnet test eng/service.proj
-          --filter '(TestCategory!=Manually) & (TestCategory!=Live) & ($(AdditionalTestFilters))'
+          --filter "(TestCategory!=Manually) & (TestCategory!=Live) & ($(AdditionalTestFilters))"
```
benbp, line 92: I assume the double quotes won't break here if AdditionalTestFilters is not defined, because it's no longer powershell?

benbp (approved):

eng/pipelines/templates/jobs/ci.tests.yml
weshaggard: Yes based on being able to see the full command line now in the logs it looks correct to me.

weshaggard:
The failing storage test pipeline is unrelated to this change, I just triggered it to verify the AdditionaFilters parameter works correctly.

PR #32545, opened 2022-11-18
Redirect customers from sources to samples

### LINKED ISSUE #32264: Remove links to test sources from sample markdowns
Some samples' markdown files include links to the source files, but these add no value and may only confuse customers, leading to them think they may need to set up their functions similarly. Remove any links to such source files across all SDKs.

### LINKED ISSUE #32267: Add README to sample source directories to redirect to sample README
For customers discovering our sample source under test/samples, we should add a simple README that points back to the samples/README.md with text similar to:

```markdown
Source files in this directory are written as tests from which samples are extracted.
They are not intended to be viewed directly and help ensure our samples compile and work correctly.
See our [list of samples](../../samples/README.md) for more explanation about how to use this client library.
```

heaths:
Fixes #32267 by adding a README.md to each sample source directory redirecting back to the packages' _samples_ directory, if any. Also fixes #32264. There are a few left where the sample is less of a test and would be involved to replace with snippets.

### DIFF
--- a/sdk/eventhub/Azure.Messaging.EventHubs/samples/Sample10_AzureEventSourceListener.md
+++ b/sdk/eventhub/Azure.Messaging.EventHubs/samples/Sample10_AzureEventSourceListener.md
@@ -79,7 +79,7 @@ finally
 
 ## Apply filtering logic to logs
 
-This examples demonstrates using a callback with the listener to allow custom logic, such as filtering, of the log messages.  This can help to reduce log spam when troubleshooting by capturing just the log entries that are helpful.   
+This examples demonstrates using a callback with the listener to allow custom logic, such as filtering, of the log messages.  This can help to reduce log spam when troubleshooting by capturing just the log entries that are helpful.
 
 In the snippet below, the `Verbose` messages for the `Azure-Identity` event source are captured and written to `Trace` output.  Log messages for the `Azure-Messaging-EventHubs` event source are filtered to capture only a specific set to aid in debugging publishing, which are then written to the standard `Console` output.
 
@@ -127,7 +127,7 @@ finally
 
 ## Capture filtered logs to a file
 
-For scenarios where capturing logs to `Trace` or `Console` outputs isn't ideal, log information can be streamed into a variety of targets, such as Azure Storage, databases, and files for durable persistence.    This example demonstrates capturing error logs to a text file so that they can be analyzed later, while capturing non-error information to `Console` output.  
+For scenarios where capturing logs to `Trace` or `Console` outputs isn't ideal, log information can be streamed into a variety of targets, such as Azure Storage, databases, and files for durable persistence.    This example demonstrates capturing error logs to a text file so that they can be analyzed later, while capturing non-error information to `Console` output.
 
 ```C# Snippet:EventHubs_Sample10_CustomListenerWithFile
 var connectionString = "<< CONNECTION STRING FOR THE EVENT HUBS NAMESPACE >>";
@@ -175,15 +175,15 @@ finally
 
 ## Finding the desired events
 
-The Event Hubs client library logs using several event sources for different areas of functionality.  Each event source contains multiple log events, with most grouped into logical sets that follow the pattern:  
+The Event Hubs client library logs using several event sources for different areas of functionality.  Each event source contains multiple log events, with most grouped into logical sets that follow the pattern:
 
 - `{ Operation Name }Start`
 - `{ Operation Name }Error`
 - `{ Operation Name }Complete`
 
-Each operation will always emit its "Start" and "Complete" log events, and will only emit its "Error" event as needed.  For AMQP operations, the "Complete" event will detail the number of retries that were used for that operation.  
+Each operation will always emit its "Start" and "Complete" log events, and will only emit its "Error" event as needed.  For AMQP operations, the "Complete" event will detail the number of retries that were used for that operation.
 
-Unfortunately, there is currently way to easily view all of the log events offered.  To discover the available log events, inspecting the associated source code is the best option. 
+Unfortunately, there is currently way to easily view all of the log events offered.  To discover the available log events, inspecting the associated source code is the best option.
 
 ### Event Source: "Azure-Messaging-EventHubs"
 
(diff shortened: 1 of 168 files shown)


heaths:
I'll also update our sample guidelines to recommend against linking to sample sources. This seems to be generating too much confusion.

/cc @achandmsft @meeraharidasa

heaths:
@benbp @weshaggard who owns the README validation? Seems it's too broadly scoped:

```
/sdk/appconfiguration/Azure.ResourceManager.AppConfiguration/tests/Samples/README.md is missing headers with patterns
```

This is basically meant as a "homepage" for the directory in GitHub so that customers (hopeflly) see this isn't the directory for samples they are looking for and links to the appropriate directory. I would think this validator should only be checking README.md files directly under package directories.

petrsvihlik (approved):
communication identity 👍

sdk/eventhub/Azure.Messaging.EventHubs/samples/Sample10_AzureEventSourceListener.md
```diff
@@ -127,7 +127,7 @@ finally
 
 ## Capture filtered logs to a file
 
-For scenarios where capturing logs to `Trace` or `Console` outputs isn't ideal, log information can be streamed into a variety of targets, such as Azure Storage, databases, and files for durable persistence.    This example demonstrates capturing error logs to a text file so that they can be analyzed later, while capturing non-error information to `Console` output.  
+For scenarios where capturing logs to `Trace` or `Console` outputs isn't ideal, log information can be streamed into a variety of targets, such as Azure Storage, databases, and files for durable persistence.    This example demonstrates capturing error logs to a text file so that they can be analyzed later, while capturing non-error information to `Console` output.
```
jsquire, line 130:
```suggestion
For scenarios where capturing logs to `Trace` or `Console` outputs isn't ideal, log information can be streamed into a variety of targets, such as Azure Storage, databases, and files for durable persistence.  This example demonstrates capturing error logs to a text file so that they can be analyzed later, while capturing non-error information to `Console` output.
```

sdk/eventhub/Azure.Messaging.EventHubs/samples/Sample10_AzureEventSourceListener.md
```diff
@@ -79,7 +79,7 @@ finally
 
 ## Apply filtering logic to logs
 
-This examples demonstrates using a callback with the listener to allow custom logic, such as filtering, of the log messages.  This can help to reduce log spam when troubleshooting by capturing just the log entries that are helpful.
```
jsquire, line 82: wow.  I suck.

jsquire (approved):

sdk/template/Azure.Template/samples/Sample1.HelloWorldAsync.md
```diff
@@ -22,6 +22,3 @@ SecretBundle secret = await client.GetSecretValueAsync("TestSecret");
 
 Console.WriteLine(secret.Value);
 ```
-
-To see the full example source files, see:
```
annelo-msft, line 26: Do we know why this was included in the template in first place?  Curious why we authored this given the current concerns - what changed in our practices to change the guidance?

sdk/translation/Azure.AI.Translation.Document/tests/samples/README.md
```diff
@@ -0,0 +1,3 @@
+Source files in this directory are written as tests from which samples are extracted.
```
annelo-msft, line 1: Proliferating copies of the same text could result in drift if someone changes one and they get out of sync ... is there a way to practice DRY for this README at all?

sdk/eventhub/Azure.Messaging.EventHubs/samples/Sample10_AzureEventSourceListener.md
heaths: I was trying to keep this PR focused on just this change. If you want me to go through and fix up a lot of the markdown problems as found by the Microsoft Docs extension pack, it may take several days. If that's worthwhile, can we do that in another PR?

Also, the double-space in the second recommendation - the archaic typewriter style - will actually be folded into a single space by markdown renderers. Same with the trailing space, though those are not recommended by git. It actually highlights them in an initial diff.

sdk/template/Azure.Template/samples/Sample1.HelloWorldAsync.md
heaths: When I first prototyped our new samples-as-tests with Pavel and Krzysztof there was some debate about how extracts may not be as useful as a full end-to-end sample. At one point we even discussed doing standalone apps much like Python does, but making sure we can test these - and, hence, relying on the MSTest framework (using NUnit to execute tests - was preferred.

It was a point in time thing, but after hearing customer feedback how they were confused by our samples (and turns out they were looking at these tests sources), we want to deemphasize the test sources as much as possible and lead people to the markdowns.

sdk/eventhub/Azure.Messaging.EventHubs/samples/Sample10_AzureEventSourceListener.md
jsquire: Double-space 4Lyfe!

sdk/translation/Azure.AI.Translation.Document/tests/samples/README.md
heaths: In order to show in all relevant directories, they need to actually exist in those directories. Even if text drifts a little, the message is only "these are not the samples you're looking for" with a link to the sample sources.

sdk/eventhub/Azure.Messaging.EventHubs/samples/Sample10_AzureEventSourceListener.md
heaths: As discussed offline, this isn't necessary now. Perhaps a separate pass to resolve markdown linting rules would help, but unless people actually have the Microsoft Docs extension pack, more issues will only be introduced later. I can mention it in my "PSA" talking about this change to the team.

heaths:
Besides the management plane timeout, the other errors are irrelevant and I've opened bugs to track:

* #32568 
* #32569 
* #32570

PR #32564, opened 2022-11-18
Use Rooms connection string for tests

anujissarMS:
Rooms require different connection string in INT cloud from what is being for other communication services.
Use the Communication_Connection_string_Rooms secret as env variable for Rooms tests

### DIFF
--- a/sdk/communication/Azure.Communication.Rooms/tests/RoomsClientTestEnvironment.cs
+++ b/sdk/communication/Azure.Communication.Rooms/tests/RoomsClientTestEnvironment.cs
@@ -1,11 +1,21 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System;
 using Azure.Communication.Tests;
+using Azure.Core.TestFramework;
 
 namespace Azure.Communication.Rooms.Tests
 {
     public class RoomsClientTestEnvironment : CommunicationTestEnvironment
     {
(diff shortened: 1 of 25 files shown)


elavarasidc:
Please tag an user story related to these changes.

sdk/communication/Azure.Communication.Rooms/tests/RoomsClientTestEnvironment.cs
```diff
@@ -1,11 +1,21 @@
...
 
+using System;
 using Azure.Communication.Tests;
+using Azure.Core.TestFramework;
 
 namespace Azure.Communication.Rooms.Tests
 {
     public class RoomsClientTestEnvironment : CommunicationTestEnvironment
     {
+        public const string CommunicationConnectionStringRoomsVariableName = "COMMUNICATION_CONNECTION_STRING_ROOMS";
```
elavarasidc, line 12: Does it mean to the INT environment, now we have added a new environment variable COMMUNICATION_CONNECTION_STRING_ROOMS with the rooms specific connection string?

anujissarMS: There is no differentiation for INT or Public cloud in tests. This variable is present in both INT and Public cloud key vaults. 
In INT it points to our PPE but in public it points to the same resource as the previous env variable.

elavarasidc (approved):

Issue #31624, opened 2022-10-06
[FEATURE REQ] ServiceBusProcessor allow to adjust the prefetch count forUpdateConcurrency


danielmarbach:
### Library name

Azure.Messaging.ServiceBus

### Please describe the feature.

The ServiceBusProcessor allows to dynamically adjust the concurrency by calling `UpdateConcurrency` which is a pretty nice feature!

Sometimes though when you are for example downsizing the concurrency it also makes sense to adjust the prefetch count to avoid running into too many MessageLockLostExceptions during the time window when running with limited concurrency. 

Since the prefetch count is a link setting, it would require to introduce an `ValueTask UpdateConcurrencyAsync(int maxConcurrentCalls, int? prefetchCount = default)` that when the prefetch count is set to a new value stops the behind-the-scenes receiver and re-establishes a new link with the new prefetch count. 

Currently, we are solving this scenario by stopping the receiver, creating a new one and starting the new one. Having this possibility on the processor would be :nail_care:

jsquire:
Hi @danielmarbach.   Interesting idea, and I think that it makes sense conceptually.  If you're reading messages more slowly there's a bigger risk of having locks expire while sitting in prefetch.   @JoshLove-msft - thoughts?

JoshLove-msft:
I think it makes sense. We have a similar request that is asking for the ability to modify this in the functions extensions - https://github.com/Azure/azure-sdk-for-net/issues/25929.

danielmarbach:
By the way, I'd be happy to give this ago after my vacation sometime in late October, in case you find this valuable.

JoshLove-msft:
> By the way, I'd be happy to give this ago after my vacation sometime in late October, in case you find this valuable.

Yes, that would be great! One thing we may want to think about is whether we should expose this for the regular receive and function extensions as well.

danielmarbach:
I have raised last Friday https://github.com/Azure/azure-sdk-for-net/pull/32572 to get the discussion started.

PR #32572, opened 2022-11-18
Allow updating prefetch count in UpdateConcurrency

danielmarbach:
Spike for #31624 to get the discussion started

This is a first attempt to allow adjusting the prefetch count with an UpdateConcurrency overload. 

It could very well be that I misunderstood the reconciliation mechanism and this PR is not even valid. But this is all I could do for a Friday evening coding session, so please be patient with me.

### DIFF
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ReceiverManager.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ReceiverManager.cs
@@ -39,7 +39,7 @@ namespace Azure.Messaging.ServiceBus
             _receiverOptions = new ServiceBusReceiverOptions
             {
                 ReceiveMode = ProcessorOptions.ReceiveMode,
-                PrefetchCount = ProcessorOptions.PrefetchCount,
+                PrefetchCount = processor.PrefetchCount,
                 // Pass None for subqueue since the subqueue has already
                 // been taken into account when computing the EntityPath of the processor.
                 SubQueue = SubQueue.None,
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessor.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessor.cs
@@ -103,7 +103,7 @@ namespace Azure.Messaging.ServiceBus
         /// The prefetch count is specified using <see cref="ServiceBusProcessorOptions.PrefetchCount"/>
         /// and has a default value of 0.
         /// </value>
-        public virtual int PrefetchCount { get; }
+        public virtual int PrefetchCount => _prefetchCount;
 
         /// <summary>
         /// Gets whether or not this processor is currently processing messages.
@@ -114,7 +114,7 @@ namespace Azure.Messaging.ServiceBus
         /// </value>
         public virtual bool IsProcessing => ActiveReceiveTask != null;
 
-        internal ServiceBusProcessorOptions Options { get; }
+        internal ServiceBusProcessorOptions Options { get; } // TODO contains stale values??
 
         /// <summary>
         ///   The active connection to the Azure Service Bus service, enabling client communications for metadata
@@ -131,6 +131,7 @@ namespace Azure.Messaging.ServiceBus
         /// </value>
         public virtual int MaxConcurrentCalls => _maxConcurrentCalls;
         private volatile int _maxConcurrentCalls;
+        private volatile int _prefetchCount;
         private int _currentConcurrentCalls;
 
         internal int MaxConcurrentSessions => _maxConcurrentSessions;
@@ -248,7 +249,7 @@ namespace Azure.Messaging.ServiceBus
             Identifier = string.IsNullOrEmpty(Options.Identifier) ? DiagnosticUtilities.GenerateIdentifier(EntityPath) : Options.Identifier;
 
             ReceiveMode = Options.ReceiveMode;
-            PrefetchCount = Options.PrefetchCount;
+            _prefetchCount = Options.PrefetchCount;
             MaxAutoLockRenewalDuration = Options.MaxAutoLockRenewalDuration;
             _maxConcurrentCalls = Options.MaxConcurrentCalls;
             MaxReceiveWaitTime = Options.MaxReceiveWaitTime;
@@ -636,7 +637,7 @@ namespace Azure.Messaging.ServiceBus
             }
         }
 
-        private void ReconcileReceiverManagers(int maxConcurrentSessions)
+        private void ReconcileReceiverManagers(int maxConcurrentSessions, int prefetchCount)
         {
             if (_receiverManagers.Count == 0)
             {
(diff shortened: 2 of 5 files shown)


sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ReceiverManager.cs
```diff
@@ -39,7 +39,7 @@ internal class ReceiverManager
             _receiverOptions = new ServiceBusReceiverOptions
             {
                 ReceiveMode = ProcessorOptions.ReceiveMode,
-                PrefetchCount = ProcessorOptions.PrefetchCount,
```
danielmarbach, line 42: I really don't like this but while doing some spiking and fiddling around I discovered the options are never aligned anymore after the initial cloning. I wonder if the whole approach needs an overhaul. For example it would be possible to use the options as a base and then clone and reassign them on every modification. Then they would properly represent the adjusted values also for `MaxConcurrentCalls` when they are updated with `UpdateConcurrency`

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessor.cs
```diff
@@ -114,7 +114,7 @@ public class ServiceBusProcessor : IAsyncDisposable
         /// </value>
         public virtual bool IsProcessing => ActiveReceiveTask != null;
 
-        internal ServiceBusProcessorOptions Options { get; }
+        internal ServiceBusProcessorOptions Options { get; } // TODO contains stale values??
```
danielmarbach, line 117: See also https://github.com/Azure/azure-sdk-for-net/pull/32572/files#r1026946748

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessor.cs
```diff
@@ -1035,6 +1058,24 @@ public async ValueTask DisposeAsync()
...
+        /// <summary>
+        ///
+        /// </summary>
+        /// <param name="maxConcurrentCalls"></param>
+        /// <param name="prefetchCount"></param>
+        public void UpdateConcurrency(int maxConcurrentCalls, int prefetchCount)
+        {
+            Argument.AssertAtLeast(maxConcurrentCalls, 1, nameof(maxConcurrentCalls));
+            Argument.AssertAtLeast(prefetchCount, 0, nameof(prefetchCount));
+            lock (_maxConcurrencySyncLock)
```
danielmarbach, line 1070: The sync lock name is probably now also no longer good and needs a more descriptive name

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusSessionProcessor.cs
```diff
@@ -335,7 +335,7 @@ public async ValueTask DisposeAsync()
         /// <see cref="ServiceBusSessionProcessor.MaxConcurrentCallsPerSession"/>.</param>
         public void UpdateConcurrency(int maxConcurrentSessions, int maxConcurrentCallsPerSession)
         {
-            InnerProcessor.UpdateConcurrency(maxConcurrentSessions, maxConcurrentCallsPerSession);
+            InnerProcessor.UpdateConcurrencyInternal(maxConcurrentSessions, maxConcurrentCallsPerSession);
```
danielmarbach, line 338: The hole session path is completely ignored for now

danielmarbach:
Ping @jsquire @JoshLove-msft

jsquire:
> Ping @jsquire @JoshLove-msft

For this one, I'm going to defer to @JoshLove-msft.   I'm happy to help review if this is a feature that he believes we should include.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ReceiverManager.cs
JoshLove-msft: It shouldn't matter if the options are not aligned for the concurrency settings as those are consumed in the processor rather than the manager. With Prefetch it is different as that is consumed here since it gets passed to the receiver options. I think we can probably avoid needing to re-recreate the receiver for prefetch changes by copying what was done in the old library here - https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/servicebus/Microsoft.Azure.ServiceBus/src/Core/MessageReceiver.cs#L245. We could make the Prefetch property in ServiceBusReceiver internally settable so that it can be modified by the processor.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessor.cs
```diff
@@ -1035,6 +1058,24 @@ public async ValueTask DisposeAsync()
             GC.SuppressFinalize(this);
         }
 
+        /// <summary>
+        ///
+        /// </summary>
+        /// <param name="maxConcurrentCalls"></param>
+        /// <param name="prefetchCount"></param>
+        public void UpdateConcurrency(int maxConcurrentCalls, int prefetchCount)
```
JoshLove-msft, line 1066: Any reason to not have a separate method UpdatePrefetch?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessor.cs
JoshLove-msft: Yeah this isn't ideal, but it also the values that can get stale are not re-read. We could update to only cache the specific values that we need to reference repeatedly.

JoshLove-msft:
> > Ping @jsquire @JoshLove-msft
> 
> For this one, I'm going to defer to @JoshLove-msft. I'm happy to help review if this is a feature that he believes we should include.

I like the feature and think it will be a good addition. I added a couple comments around potentially using the same mechanism that was used in the T1 library for modifying prefetch as well as an API question.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ReceiverManager.cs
danielmarbach: Will wrap my head around it and report back

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessor.cs
danielmarbach: My thought was that concurrency and prefetch should ideally be neatly balanced in relation to each other. That thought made me think you want to tweak those two settings while processing together. 

From an implementation detail perspective having two methods under the assumption both would be mostly used together (if that is true) would require unnecessary double synchronisation

But maybe my assumptions are incorrect when it comes to those two values should be balanced. That being said I always treated it that way in my practice.

What are your thoughts?

danielmarbach: Happy to adjust to whatever you feel is best

JoshLove-msft: Yeah I would just go with a separate method, UpdatePrefetch.

danielmarbach:
I apologise for pinging and then not being responsive on my end. End of year and some other things causes me to not have as much spare time as I wished/hope to have to focus on this.

JoshLove-msft:
> I apologise for pinging and then not being responsive on my end. End of year and some other things causes me to not have as much spare time as I wished/hope to have to focus on this.

No worries at all. We are grateful for the contribution!

danielmarbach:
I have sent in a follow up PR using a different approach https://github.com/Azure/azure-sdk-for-net/pull/33162

PR #32576, opened 2022-11-18
More performance improvements

AlexanderSher:
This PR addresses 3 issues:
- In `HttpRange.ToString()` it replaces `FormattableString.Invariant` with `string.Create`
- In `HttpClientTransport.JoinHeaderValues` which is still used for response headers, it replaces `string.Join` with `DefaultInterpolatedStringHandler`.
- Caches the value returned by `AppContextSwitchHelper.GetConfigValue`

**Benchmarks:**
- FormattableString vs StringCreate:
```cs
[MemoryDiagnoser]
[MarkdownExporterAttribute.GitHub]
[SimpleJob(RuntimeMoniker.Net70)]
public class FormattableStringVsStringCreate
{
    private static string _bytes = "bytes";
    private static long _start = 10343;
    private static long _end = 49305;

    [Benchmark(Baseline = true)]
    public string UseFormattableString() => FormattableString.Invariant($"{_bytes}={_start}-{_end}");

    [Benchmark]
    public string UseStringCreate() => string.Create(CultureInfo.InvariantCulture, $"{_bytes}={_start}-{_end}");
}
```

  Results:
```
|            Method |      Mean |    Error |   StdDev | Ratio |   Gen0 | Allocated | Alloc Ratio |
|------------------ |----------:|---------:|---------:|------:|-------:|----------:|------------:|
| FormattableString | 140.83 ns | 0.780 ns | 0.729 ns |  1.00 | 0.0234 |     184 B |        1.00 |
|     String.Create |  69.38 ns | 0.416 ns | 0.368 ns |  0.49 | 0.0070 |      56 B |        0.30 |
```

- String.Join vs UseDefaultInterpolatedStringHandler:
```cs
[MemoryDiagnoser]
[SimpleJob(RuntimeMoniker.Net70)]
public class StringJoinVsDefaultInterpolatedStringHandler
{
    private CustomEnumerable _enumerable;
    private string _valueToCopy;

    [Params(1,2,5)]
    public int Count { get; set; }

    [GlobalSetup]
    public void Setup() 
    {
        var input = new[] { "abc", "asdsads", "dfdsdgserg", "dfsr4evsdvs", "dfsefsgsegfg" };
        _enumerable = new CustomEnumerable(input.Take(Count));
        _valueToCopy = string.Join(',', _enumerable.Take(Count));
    }

    [Benchmark(Baseline = true)]
    public string UseStringJoin() => string.Join(',', _enumerable);

    [Benchmark]
    public string UseDefaultInterpolatedStringHandler() 
    {
        var count = _enumerable.Count;
        var interpolatedStringHandler = new DefaultInterpolatedStringHandler(count-1, count);
        var isFirst = true;

        foreach (var str in _enumerable)
        {
            if (isFirst)
            {
                isFirst = false;
            }
            else
            {
                interpolatedStringHandler.AppendLiteral(",");
            }
            interpolatedStringHandler.AppendFormatted(str);
        }
        return string.Create(null, ref interpolatedStringHandler);
    }
    
    [Benchmark]
    public string CreateCopyOfString() => string.Copy(_valueToCopy);

    private class CustomEnumerable : IEnumerable<string>
    {
        private readonly List<string> _values;

        public CustomEnumerable(IEnumerable<string> values) 
        {
            _values = values.ToList();
        }

        public int Count => _values.Count;
        [MethodImpl(MethodImplOptions.NoInlining)]
        public List<string>.Enumerator GetEnumerator() => _values.GetEnumerator();
        [MethodImpl(MethodImplOptions.NoInlining)]
        IEnumerator<string> IEnumerable<string>.GetEnumerator() => _values.GetEnumerator();
        [MethodImpl(MethodImplOptions.NoInlining)]
        IEnumerator IEnumerable.GetEnumerator() => _values.GetEnumerator();
    }
}
```

  Results
```
|                           Method | Count |       Mean |     Error |    StdDev | Ratio |   Gen0 | Allocated | Alloc Ratio |
|--------------------------------- |------ |-----------:|----------:|----------:|------:|-------:|----------:|------------:|
|                      String.Join |     1 |  39.668 ns | 0.2842 ns | 0.2520 ns |  1.00 | 0.0051 |      40 B |        1.00 |
| DefaultInterpolatedStringHandler |     1 |  52.187 ns | 0.1106 ns | 0.0981 ns |  1.32 | 0.0041 |      32 B |        0.80 |
|               CreateCopyOfString |     1 |   8.961 ns | 0.0302 ns | 0.0252 ns |  0.23 | 0.0041 |      32 B |        0.80 |
|                                  |       |            |           |           |       |        |           |             |
|                      String.Join |     2 |  79.848 ns | 0.1984 ns | 0.1657 ns |  1.00 | 0.0112 |      88 B |        1.00 |
| DefaultInterpolatedStringHandler |     2 |  51.448 ns | 1.0575 ns | 1.1755 ns |  0.64 | 0.0061 |      48 B |        0.55 |
|          Copy of string constant |     2 |  10.010 ns | 0.2156 ns | 0.1800 ns |  0.13 | 0.0061 |      48 B |        0.55 |
|                                  |       |            |           |           |       |        |           |             |
|                      String.Join |     5 | 148.205 ns | 3.0479 ns | 3.6283 ns |  1.00 | 0.0203 |     160 B |        1.00 |
| DefaultInterpolatedStringHandler |     5 |  75.206 ns | 0.4523 ns | 0.4009 ns |  0.51 | 0.0153 |     120 B |        0.75 |
|          Copy of string constant |     5 |  15.742 ns | 0.2206 ns | 0.1956 ns |  0.11 | 0.0153 |     120 B |        0.75 |
```


sdk/core/Azure.Core/src/HttpRange.cs
```diff
@@ -52,13 +52,21 @@ public HttpRange(long offset = 0, long? length = default)
...
+            span.AppendChar(ref position, '-');
+
             if (Length.HasValue && Length != 0)
             {
-                endRange = (Offset + Length.Value - 1).ToString(CultureInfo.InvariantCulture);
+                span.AppendLong(ref position, Offset + Length.Value - 1, provider: CultureInfo.InvariantCulture);
             }
 
-            return FormattableString.Invariant($"{Unit}={Offset}-{endRange}");
+            return span.Slice(0, position).ToString();
```
AlexanderSher, line 69: `HttpRange.ToString` is called very often, but usually on different instances, so caching the value doesn't provide any effect.

sdk/core/Azure.Core/src/Internal/SpanOfCharsExtensions.cs
```diff
@@ -0,0 +1,41 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+
+namespace Azure.Core
+{
+    internal static class SpanOfCharsExtensions
```
AlexanderSher, line 8: Alternative to static helpers can be something like [`ValueStringBuilder`](https://github.com/dotnet/runtime/blob/main/src/libraries/Common/src/System/Text/ValueStringBuilder.cs), but I would prefer to make code that uses these methods look more low level to avoid possible misuse of array on stack

sdk/core/Azure.Core/src/Pipeline/HttpClientTransport.cs
```diff
@@ -326,12 +327,40 @@ internal static void CopyHeaders(HttpHeaders from, HttpHeaders to)
...
-            };
+            var count = values.Count;
+            // Explicitly handle most common cases
+            switch (count) {
+                case 1:
+                    return values.ToString();
+                case 0:
+                    return string.Empty;
+                case 2:
+                    using (var enumerator = values.GetEnumerator())
```
AlexanderSher, line 338: This is a very common case. Unfortunately, the only way to access elements of the `HeaderStringValues` is to use enumerator.

sdk/storage/Azure.Storage.Blobs/perf/Azure.Storage.Blobs.Perf/Options/PartitionedTransferOptions.cs
```diff
@@ -98,5 +116,203 @@ private void UpdateStorageTransferOptions()
...
+/*
+#if NET5_0_OR_GREATER
+                    && x.VersionPolicy == y.VersionPolicy
+                    && EqualOptions(x.Options, y.Options)
+#else
+                    && EqualOptions(x.Properties, y.Properties)
+#endif
+                    && EqualHeaders(x.Headers, y.Headers)
+                    && EqualContent(x.Content, y.Content);
+*/
```
christothes, line 261: remove?

sdk/storage/Azure.Storage.Blobs/perf/Azure.Storage.Blobs.Perf/Options/PartitionedTransferOptions.cs
```diff
@@ -98,5 +116,203 @@ private void UpdateStorageTransferOptions()
...
+            private static bool EqualOptions(IDictionary<string, object> xOptions, IDictionary<string, object> yOptions)
+            {
+                if (xOptions.Count != yOptions.Count)
+                {
+                    return false;
+                }
+
+                foreach (var kvp in xOptions)
+                {
+                    if (!xOptions.TryGetValue(kvp.Key, out var yValue) || !Equals(kvp.Value, yValue))
```
christothes, line 285: Looks like we are comparing xOptions to xOptions here
```suggestion
                    if (!yOptions.TryGetValue(kvp.Key, out var yValue) || !Equals(kvp.Value, yValue))
```

sdk/core/Azure.Core/src/Internal/SpanOfCharsExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        public static void AppendString(this ref Span<char> span, ref int offset, string str)
+        {
+#if NET6_0_OR_GREATER
```
danielmarbach, line 12: Just out of curiosity why net simply `#if NET`?

annelo-msft (commented):
For perf-related PRs, it'd be helpful to share a benchmark showing the improvement in the PR description.

sdk/core/Azure.Core/src/Internal/SpanOfCharsExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+            str.CopyTo(span[offset..]);
+#else
+            str.AsSpan().CopyTo(span.Slice(offset));
+#endif
+            offset += str.Length;
+        }
+
+        public static void AppendChar(this ref Span<char> span, ref int offset, char value)
+        {
+            span[offset] = value;
```
annelo-msft, line 22: Is it the caller's responsibility to do bounds-checking, or would it make sense for us to add parameter validation?

sdk/core/Azure.Core/src/Internal/SpanOfCharsExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+            offset++;
+        }
+
+        public static void AppendLong(this ref Span<char> span, ref int offset, long value, string? format = null, IFormatProvider? provider = null)
```
annelo-msft, line 26: Curious: why add this overload, and not require the caller to pass a string?  Would we consider adding overloads for all types that can be formatted as string?

sdk/core/Azure.Core/src/Pipeline/HttpClientTransport.cs
```diff
@@ -326,12 +327,40 @@ internal static void CopyHeaders(HttpHeaders from, HttpHeaders to)
...
-            return values.Count switch
-            {
-                0 => string.Empty,
-                1 => values.ToString(),
-                _ => string.Join(",", values)
-            };
+            var count = values.Count;
+            // Explicitly handle most common cases
+            switch (count) {
+                case 1:
```
annelo-msft, line 333: Curious: why are cases not in numerical order?

sdk/storage/Azure.Storage.Blobs/src/BlobExtensions.cs
```diff
@@ -1481,9 +1481,7 @@ internal static PageRangeItem[] ToPageBlobRanges(this ResponseWithHeaders<PageLi
             string operationName,
             string parameterName)
         {
-            if (AppContextSwitchHelper.GetConfigValue(
-                Constants.DisableRequestConditionsValidationSwitchName,
-                Constants.DisableRequestConditionsValidationEnvVar))
+            if (CompatSwitches.DisableRequestConditionsValidation)
```
annelo-msft, line 1484: Does removing this change the functionality?

sdk/storage/Azure.Storage.Common/src/Shared/CompatSwitches.cs
```diff
@@ -0,0 +1,15 @@
...
+
+using Azure.Core;
+
+namespace Azure.Storage
+{
+    internal static class CompatSwitches
+    {
+        private static bool? _disableRequestConditionsValidation;
+
+        public static bool DisableRequestConditionsValidation => _disableRequestConditionsValidation
```
annelo-msft, line 12: Does making this static potentially cause problems if multiple libraries access this value from different environments?

sdk/storage/Azure.Storage.Common/src/Shared/StorageSharedKeyPipelinePolicy.cs
```diff
@@ -42,45 +42,46 @@ public override void OnSendingRequest(HttpMessage message)
...
             var signature = StorageSharedKeyCredentialInternals.ComputeSasSignature(_credentials, stringToSign);
 
             var key = new AuthenticationHeaderValue(Constants.HeaderNames.SharedKey, _credentials.AccountName + ":" + signature).ToString();
-            message.Request.Headers.SetValue(Constants.HeaderNames.Authorization, key);
+            request.Headers.SetValue(Constants.HeaderNames.Authorization, key);
         }
 
         // If you change this method, make sure live tests are passing before merging PR.
-        private string BuildStringToSign(HttpMessage message)
+        private string BuildStringToSign(RequestMethod requestMethod, Uri uri, RequestContent content, Dictionary<string, string> headers)
```
annelo-msft, line 62: Curious: how does this improve perf?

sdk/storage/Azure.Storage.Blobs/perf/Azure.Storage.Blobs.Perf/Options/PartitionedTransferOptions.cs
```diff
@@ -50,6 +59,9 @@ public class PartitionedTransferOptions : SizeOptions, IBlobClientOptionsProvide
         [Option("client-encryption")]
         public string EncryptionVersionString { get; set; }
 
+        [Option("cache-responses")]
+        public bool CacheResponses { get; set; }
```
annelo-msft, line 63: Curious: would this need to be exposed in a public APIs for customers to get the perf benefit of this, or is it just an improvement to boost perf on our own perf tests?

sdk/storage/Azure.Storage.Blobs/perf/Azure.Storage.Blobs.Perf/Options/PartitionedTransferOptions.cs
AlexanderSher: Yep, the whole type is temporary

sdk/core/Azure.Core/src/Internal/SpanOfCharsExtensions.cs
AlexanderSher: I would prefer caller to be responsible for bounds-checking. This is a very low-level extension method, caller must be aware of what it is doing.

sdk/core/Azure.Core/src/Internal/SpanOfCharsExtensions.cs
AlexanderSher: > why add this overload, and not require the caller to pass a string?

Pass a string where?

> Would we consider adding overloads for all types that can be formatted as string?

Yeah, that sounds like a good idea. And I'll add tests for them as well.

sdk/core/Azure.Core/src/Pipeline/HttpClientTransport.cs
AlexanderSher: `1` is the most common case. But since IL creates `switch` instruction anyway, it can be ordered in numerical order

sdk/storage/Azure.Storage.Common/src/Shared/CompatSwitches.cs
AlexanderSher: No, this is an internal shared file, so each library will have its own copy of the `CompatSwitches` type

sdk/storage/Azure.Storage.Blobs/perf/Azure.Storage.Blobs.Perf/Options/PartitionedTransferOptions.cs
AlexanderSher: This whole type must be redone, I'll remove it from this PR.

sdk/storage/Azure.Storage.Common/src/Shared/StorageSharedKeyPipelinePolicy.cs
christothes: It would be great to see the diffs in Benchmark.NET

sdk/core/Azure.Core/src/Pipeline/HttpClientTransport.cs
```diff
@@ -327,12 +328,40 @@ internal static void CopyHeaders(HttpHeaders from, HttpHeaders to)
...
+                    return values.ToString();
+                case 2:
+                    using (var enumerator = values.GetEnumerator())
+                    {
+                        enumerator.MoveNext();
+                        var value1 = enumerator.Current;
+                        enumerator.MoveNext();
+                        var value2 = enumerator.Current;
+                        return string.Concat(value1, ",", value2);
```
danielmarbach, line 345: Could you shim in string.Create? It seems it would be ideal also for the default

sdk/core/Azure.Core/src/Internal/SpanOfCharsExtensions.cs
AlexanderSher: `DefaultInterpolatedStringHandler` is good enough. Removing this custom helpers.

sdk/core/Azure.Core/src/Pipeline/HttpClientTransport.cs
AlexanderSher: `DefaultInterpolatedStringHandler` is also good

sdk/storage/Azure.Storage.Blobs/perf/Azure.Storage.Blobs.Perf/Options/PartitionedTransferOptions.cs
AlexanderSher: Removed

sdk/core/Azure.Core/src/Internal/SpanOfCharsExtensions.cs
AlexanderSher: Harder to read (you need to remember that `NET` is 5.0+), also some features are available only in 6.0+ and we don't release for 5.0

sdk/core/Azure.Core/src/Pipeline/HttpClientTransport.cs
AlexanderSher: Done

sdk/storage/Azure.Storage.Blobs/perf/Azure.Storage.Blobs.Perf/Options/PartitionedTransferOptions.cs
AlexanderSher: Change is undone

sdk/storage/Azure.Storage.Common/src/Shared/StorageSharedKeyPipelinePolicy.cs
AlexanderSher: Removed because https://github.com/Azure/azure-sdk-for-net/pull/34195 solved the problem

sdk/core/Azure.Core/src/Pipeline/HttpClientTransport.cs
AlexanderSher: Replaced with `DefaultInterpolatedStringHandler`.

sdk/core/Azure.Core/src/Internal/SpanOfCharsExtensions.cs
AlexanderSher: Whole type is removed in favor of `DefaultInterpolatedStringHandler`

sdk/core/Azure.Core/src/Internal/SpanOfCharsExtensions.cs
AlexanderSher: Whole type is removed in favor of `DefaultInterpolatedStringHandler`

sdk/storage/Azure.Storage.Blobs/perf/Azure.Storage.Blobs.Perf/Options/PartitionedTransferOptions.cs
AlexanderSher: Changes in this file are reverted

sdk/core/Azure.Core/src/HttpRange.cs
AlexanderSher: Removed in favor of `string.Create` with `DefaultInterpolatedStringHandler`

annelo-msft (commented):
Would you be willing to add a description of the changes to the PR, as well as benchmark numbers to help us understand their effectiveness?

christothes (approved):

christothes (commented):
Would be nice if we had some benchmarks.

AlexanderSher:
> Would be nice if we had some benchmarks.

```
|            Method |      Mean |    Error |   StdDev | Ratio |   Gen0 | Allocated | Alloc Ratio |
|------------------ |----------:|---------:|---------:|------:|-------:|----------:|------------:|
| FormattableString | 140.83 ns | 0.780 ns | 0.729 ns |  1.00 | 0.0234 |     184 B |        1.00 |
|     String.Create |  69.38 ns | 0.416 ns | 0.368 ns |  0.49 | 0.0070 |      56 B |        0.30 |
```

mikeharder:
@AlexanderSher: Do you know if any of these performance improvements are large enough to be detected in our end-to-end performance tests like storage-blob upload/download/list-blobs?

AlexanderSher:
> @AlexanderSher: Do you know if any of these performance improvements are large enough to be detected in our end-to-end performance tests like storage-blob upload/download/list-blobs?

@mikeharder I expect to see some visible improvements in our throughput tests cause these fixes were addressing headers and appcontext switches, both affecting storage blob performance

mikeharder:
> @mikeharder I expect to see some visible improvements in our throughput tests cause these fixes were addressing headers and appcontext switches, both affecting storage blob performance

Did you run any end-to-end tests on your PR to compare with main?  If not, we can just see if the results from main appear to improve on the dashboard.

PR #32578, opened 2022-11-19
Update sample guidance

heaths:
These changes correspond with the [changes to guidelines](https://github.com/Azure/azure-sdk/pull/5152) and the recent PR #32545

### DIFF
--- a/samples/README.MD
+++ b/samples/README.MD
@@ -3,43 +3,107 @@
 Developers like to learn by looking at code, and so the Azure SDK comes with a myriad of code samples in the form of short code snippets, sample applications, and how-to guides. This document describes where to find all these resources.
 
 ## Structure of the Repository
+
 The Azure SDK repository is organized in the following folder structure, with the main sample locations highlighted using **bold** font.
 
-`/samples` (this folder)<br>
-&nbsp;&nbsp;&nbsp;&nbsp;`README.md` (this file)<br>
-`/sdk` (folder containing sources for all SDK packages)<br>
-&nbsp;&nbsp;&nbsp;&nbsp;`/<service>` (e.g. storage)<br>
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`/<package>` (e.g. Azure.Storage.Blobs)<br>
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;**`README.md`** (package READMEs contain hello world samples)<br>
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;**`/samples`** (package-specific samples)<br>
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`/src`<br>
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`/tests`<br>
+`/samples` (this folder)\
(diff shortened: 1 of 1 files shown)


jsquire (changes requested):
I think that we should consider defining the intended audience for this README.   Some content appears to be targeting developers who use the Azure SDK while other content seems to be instructions for package authors.

A lot of the new content being added seems to fall into the latter category.  Should we consider moving these parts to the contributing guide, so that we can keep unified guidance for package authors and let this README guide developers looking for help to use the SDK?

heaths:
There is some corresponding changes I made to guidance in Azure/azure-sdk#5152 but this documentation I updated here already seemed to be geared toward package authors.

samples/README.MD
```diff
@@ -3,43 +3,107 @@
...
-`/sdk` (folder containing sources for all SDK packages)<br>
-&nbsp;&nbsp;&nbsp;&nbsp;`/<service>` (e.g. storage)<br>
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`/<package>` (e.g. Azure.Storage.Blobs)<br>
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;**`README.md`** (package READMEs contain hello world samples)<br>
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;**`/samples`** (package-specific samples)<br>
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`/src`<br>
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`/tests`<br>
-
-##  Getting Started (a.k.a. `Hello World`) Samples
+`/samples` (this folder)\
```
annelo-msft, line 9: This looks quite similar to what we have in the DPG library author docs: https://github.com/Azure/azure-sdk-for-net/blob/main/doc/DataPlaneCodeGeneration/Create_Samples_Guidance.md

Would it make sense for these docs to link to each other?  How can we make sure they are both updated if one is updated?

@jsquire's point about Samples guidance in the CONTRIBUTING.md (https://github.com/Azure/azure-sdk-for-net/blob/main/CONTRIBUTING.md#updating-sample-snippets) is good as well ... it looks like we've got samples guidance in at least four places now.  Would it be worth picking a single place to discuss all of the different concepts and then linking to the central guidance from each of the necessary places?

annelo-msft:
> I think that we should consider defining the intended audience for this README. Some content appears to be targeting developers who use the Azure SDK while other content seems to be instructions for package authors.
> 
> A lot of the new content being added seems to fall into the latter category. Should we consider moving these parts to the contributing guide, so that we can keep unified guidance for package authors and let this README guide developers looking for help to use the SDK?

@jsquire, I like this call out re: target audience.  Do we have a clear-cut way we think about target audience in our repo docs?  Is it (`sdk/<service>/README`, samples, and anything linked from the README) = library **user** customer?  And most all else library **author** customer?  I think it would be good to standardize our understanding on this as a team.

heaths:
These files I'm targeting here are worded very much like they are geared towards package devs. We could better clarify that at the top of the /samples/README.md file. Admitted, it's a little confusing right now:

> Developers like to learn by looking at code, and so the Azure SDK comes with a myriad of code samples in the form of short code snippets, sample applications, and how-to guides. This document describes where to find all these resources.

That first sentence seems to be addressing package devs, but that second sentence seems to be for users. How about something like,

> Developers like to learn by looking at code, so the Azure SDK has a lot of code samples as short code snippets, sample applications, and how-to guides. The following shows package developers where they should put any of those samples to provide the best experience for users of our packages.

I also want to clean up the text a little to remove words like "myriad" that may not be as well understood - make the wording more inclusive.

/cc @KrzysztofCwalina @pallavit

heaths:
This has lower priority than other work at the moment.

PR #32593, opened 2022-11-21
[Text Analytics] Add support for dynamic classification

### LINKED ISSUE #31271: [Text Analytics] Add support for dynamic classification

# Tasks
- [ ] Implementation
- [ ] Tests
- [ ] Docs

joseharriaga:
Resolves: #31271

Features added:
- Added the following methods: `TextAnalyticsClient.DynamicClassify` and `TextAnalyticsClient.DynamicClassifyAsync`.
- Added the following methods: `TextAnalyticsClient.DynamicClassifyBatch` and `TextAnalyticsClient.DynamicClassifyBatchAsync`.
- Added the following classes: `DynamicClassifyDocumentResultCollection` and `DynamicClassifyOptions`.
- Added the `ClassificationType` enum.


sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsModelFactory.cs
```diff
@@ -736,6 +736,9 @@ public static ClassificationCategoryCollection ClassificationCategoryCollection(
         }
         #endregion
 
+        #region Dynamic Classify
+        #endregion
```
kinelski, line 740: Just double checking, but you intend to add factory methods in a future PR. Is that right?

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
```diff
@@ -2098,6 +2098,192 @@ public virtual async Task<AnalyzeHealthcareEntitiesOperation> StartAnalyzeHealth
...
+        /// <exception cref="NotSupportedException">This method is only supported in service API version 2022-10-01-preview and newer.</exception>
+        /// <exception cref="RequestFailedException">The service returned a non-success status code.</exception>
+        /// <exception cref="ArgumentException"><paramref name="document"/> is an empty string.</exception>
+        /// <exception cref="ArgumentNullException">
+        /// <para><paramref name="document"/> is null.</para>
+        /// <para><paramref name="options"/> is null.</para>
+        /// </exception>
+        public virtual Response<ClassificationCategoryCollection> DynamicClassify(
+            string document,
+            DynamicClassifyOptions options,
```
joseharriaga, line 2127: @heaths, @kinelski: I wanted to ask you what you think about this one: Right now, `DynamicClassifyOptions` has the required `Categories` property, which is just a list of strings to specify which categories participate in the classification. Here's their example, where they use the categories "Health", "Politics", "Music", and "Sports":
🔗 https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cognitiveservices/data-plane/Language/preview/2022-10-01-preview/examples/text/SuccessfuDynamicClassificationRequest.json

From the Guidelines: "The Options class is designed similarly to .NET custom attributes, where required service method parameters are modeled as Options class constructor parameters and get-only properties, and optional parameters are get-set properties." 

(I just realized I need to make `Categories` get-only if I wanted to follow this accurately. But before I do that...)

I was wondering if in this case it makes more sense to move the `Categories` property out of the Options class and make it a parameter that must be passed in all these methods (kind of like `document`/`documents` is its own parameter).

sdk/textanalytics/Azure.AI.TextAnalytics/api/Azure.AI.TextAnalytics.netstandard2.0.cs
```diff
@@ -423,6 +441,18 @@ public partial class DocumentSentiment
...
+    public partial class DynamicClassifyDocumentResultCollection : System.Collections.ObjectModel.ReadOnlyCollection<Azure.AI.TextAnalytics.ClassifyDocumentResult>
+    {
+        internal DynamicClassifyDocumentResultCollection() : base (default(System.Collections.Generic.IList<Azure.AI.TextAnalytics.ClassifyDocumentResult>)) { }
+        public string ModelVersion { get { throw null; } }
+        public Azure.AI.TextAnalytics.TextDocumentBatchStatistics Statistics { get { throw null; } }
+    }
+    public partial class DynamicClassifyOptions : Azure.AI.TextAnalytics.TextAnalyticsRequestOptions
+    {
+        public DynamicClassifyOptions(System.Collections.Generic.IList<string> categories) { }
+        public System.Collections.Generic.IList<string> Categories { get { throw null; } }
```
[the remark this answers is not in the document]
joseharriaga: I think it's similar but not entirely the same: My understanding is that `CategoriesFilter` in `RecognizePiiEntitiesOptions` is an optional property that filters down the result; is this correct? On the other hand, `Categories` in `DynamicClassifyOptions` is a required property that specifies the entire set of categories that will be used during the classification.

sdk/textanalytics/Azure.AI.TextAnalytics/api/Azure.AI.TextAnalytics.netstandard2.0.cs
```diff
@@ -423,6 +441,18 @@ public partial class DocumentSentiment
...
     }
+    public partial class DynamicClassifyDocumentResultCollection : System.Collections.ObjectModel.ReadOnlyCollection<Azure.AI.TextAnalytics.ClassifyDocumentResult>
+    {
+        internal DynamicClassifyDocumentResultCollection() : base (default(System.Collections.Generic.IList<Azure.AI.TextAnalytics.ClassifyDocumentResult>)) { }
+        public string ModelVersion { get { throw null; } }
+        public Azure.AI.TextAnalytics.TextDocumentBatchStatistics Statistics { get { throw null; } }
+    }
+    public partial class DynamicClassifyOptions : Azure.AI.TextAnalytics.TextAnalyticsRequestOptions
+    {
+        public DynamicClassifyOptions(System.Collections.Generic.IList<string> categories) { }
```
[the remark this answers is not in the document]
joseharriaga: I passed it as a constructor parameter because `Categories` is a required property, but see my following comment because I think there's a better way and want to know what you think: https://github.com/Azure/azure-sdk-for-net/pull/32593#discussion_r1028375249

kristapratico (approved):
Dynamic classification changes LGTM. Just one question - since you share ClassifyDocumentResult between DC and custom single/multi label classification, do you plan to put the detectedLanguage on this result type (even though DC can't do auto-detect language)? I don't recall if .NET decided to separate result types which can't do auto-detect or not.

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsModelFactory.cs
joseharriaga: Oh, funny. I suspect I lost this change because I might have forgotten to save the file before I rebased my branch. 🤔 I have added it back. Thank you for catching it!

mssfang (approved):

sdk/textanalytics/Azure.AI.TextAnalytics/api/Azure.AI.TextAnalytics.netstandard2.0.cs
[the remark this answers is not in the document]
heaths: What's not clear is whether the property should be read-only henceforth. We don't want customers clearing it before sending if it's required (by default we should not typically be serializing an empty collection). `ICollection<T>` defines `Clear`, so `IList<T>` would have it to. I would talk with @KrzysztofCwalina and @tg-msft about this. Perhaps if it's passed in as a parameter, backing properties should be read-only which means customers can't mutate it after. If anything, that is equivalent to other constructor parameters now that should be read-only (i.e. no `set` accessor) for those backing constructor parameters.

sdk/textanalytics/Azure.AI.TextAnalytics/src/DynamicClassifyOptions.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// <summary>
+        /// The type of classification to perform indicating whether the result considers a single category or multiple
+        /// categories. If not set, the service default is used.
+        /// </summary>
+        public ClassificationType? ClassificationType { get; set; }
+
+        /// <summary>
+        /// The categories that a given document can be classified with.
+        /// </summary>
+        public IList<string> Categories { get; internal set; } = new List<string>();
```
heaths, line 30: 1. As mentioned in one of my first comments, this should probably be `IReadOnlyList<T>` (or `IReadOnlyCollection<T>`, if more consistent throughout). This jives with our guidelines for constructor parameters (required properties).
2. Do not initialize to a `new List<string>` here. Do so in the constructor since you're taking args anyway.. This is a wasted allocation.
3. If this is for a required parameter, `categories` should never be null. Throw an exception. Use `Argument.AssertNotNullOrEmpty` because it's required, which means it has to be non-empty.

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
heaths: Depends if it might grow. I'd talk with @tg-msft or @KrzysztofCwalina about this. There's some nuance to this. I've linked the guidelines above as well but it's better to get a final say from them.

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
```diff
@@ -2098,6 +2098,192 @@ public virtual async Task<AnalyzeHealthcareEntitiesOperation> StartAnalyzeHealth
...
+        /// <returns>
+        /// The collection of categories used to classify each document that was successfully analyzed.
+        /// </returns>
+        /// <exception cref="NotSupportedException">This method is only supported in service API version 2022-10-01-preview and newer.</exception>
+        /// <exception cref="RequestFailedException">The service returned a non-success status code.</exception>
+        /// <exception cref="ArgumentException"><paramref name="document"/> is an empty string.</exception>
+        /// <exception cref="ArgumentNullException">
+        /// <para><paramref name="document"/> is null.</para>
+        /// <para><paramref name="options"/> is null.</para>
+        /// </exception>
```
heaths, line 2124: Nit: 2 paragraphs just to document this way seems unnecessary and long. This won't grow. It's more typical just to say something like:

```c#
/// <exception cref="ArgumentNullException"><paramref name="document"/> or <paramref name="options"/> is null.</exception>
```

That also jives with official .NET documentation.

heaths (changes requested):
I have enough concerns around the options class I'd like to re-review when this changes.

sdk/textanalytics/Azure.AI.TextAnalytics/api/Azure.AI.TextAnalytics.netstandard2.0.cs
```diff
@@ -1548,6 +1578,12 @@ public partial class TextAnalyticsClient
...
         public virtual System.Threading.Tasks.Task<Azure.Response<Azure.AI.TextAnalytics.DetectLanguageResultCollection>> DetectLanguageBatchAsync(System.Collections.Generic.IEnumerable<Azure.AI.TextAnalytics.DetectLanguageInput> documents, Azure.AI.TextAnalytics.TextAnalyticsRequestOptions options = null, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
         public virtual System.Threading.Tasks.Task<Azure.Response<Azure.AI.TextAnalytics.DetectLanguageResultCollection>> DetectLanguageBatchAsync(System.Collections.Generic.IEnumerable<string> documents, string countryHint = null, Azure.AI.TextAnalytics.TextAnalyticsRequestOptions options = null, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
+        public virtual Azure.Response<Azure.AI.TextAnalytics.ClassificationCategoryCollection> DynamicClassify(string document, Azure.AI.TextAnalytics.DynamicClassifyOptions options, string language = null, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
```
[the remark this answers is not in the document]
joseharriaga: The reason why it's optional is because it currently contains a property that is required for the request to be valid. This is why I find this pattern awkward and why I want to move that property out of the Options class and make it a method parameter instead (see my initial comment here: https://github.com/Azure/azure-sdk-for-net/pull/32593#discussion_r1028375249)

sdk/textanalytics/Azure.AI.TextAnalytics/api/Azure.AI.TextAnalytics.netstandard2.0.cs
```diff
@@ -314,6 +314,24 @@ public partial class ClassificationCategoryCollection : System.Collections.Objec
         internal ClassificationCategoryCollection() : base (default(System.Collections.Generic.IList<Azure.AI.TextAnalytics.ClassificationCategory>)) { }
         public System.Collections.Generic.IReadOnlyCollection<Azure.AI.TextAnalytics.TextAnalyticsWarning> Warnings { get { throw null; } }
     }
+    [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
+    public readonly partial struct ClassificationType : System.IEquatable<Azure.AI.TextAnalytics.ClassificationType>
+    {
+        private readonly object _dummy;
+        private readonly int _dummyPrimitive;
+        public ClassificationType(string value) { throw null; }
+        public static Azure.AI.TextAnalytics.ClassificationType Multi { get { throw null; } }
```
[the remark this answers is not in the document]
joseharriaga: I have added this one to my notes to discuss it in the future before reviewing APIView. The precedent of "multi label" might need to be considered, but in general, I would prefer "multiple".

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
joseharriaga: We're going with the option of moving the `Categories` property out of the `DynamicClassifyOptions` class and making it a parameter. I think it looks better now.

kinelski (approved):
LGTM!

heaths (approved):
Approving to unblock, but I have some minor questions/concerns.

sdk/textanalytics/Azure.AI.TextAnalytics/src/ServiceClients/ServiceClient.cs
```diff
@@ -155,6 +153,56 @@ public ServiceClient(TextAnalyticsClientOptions options)
...
+
+        public virtual Response<ClassificationCategoryCollection> DynamicClassify(
+            string document,
+            IEnumerable<string> categories,
+            string language = default,
+            DynamicClassifyOptions options = default,
+            CancellationToken cancellationToken = default) =>
+            throw Validation.NotSupported($"{nameof(TextAnalyticsClient)}.{nameof(TextAnalyticsClient.DynamicClassify)}", ServiceVersion.V2022_10_01_Preview, ServiceVersion);
+
+        public virtual Response<DynamicClassifyDocumentResultCollection> DynamicClassifyBatch(
```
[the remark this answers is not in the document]
joseharriaga: The difference is between the LRO features (e.g. `StartAnalyzeHealthcareEntities`/`StartAnalyzeHealthcareEntitiesAsync`) and the non-LRO features (e.g., `ExtractKeyPhrases`/`ExtractKeyPhrasesAsync` and `ExtractKeyPhrasesBatch`/`ExtractKeyPhrasesBatchAsync`). Dynamic classification is the latter, although I can see it being a little confusing because custom classification is the former.

PR #32598, opened 2022-11-22
Run code coverage on latest test target framework

### LINKED ISSUE #31768: List test runtimes in priority order and set runtime for coverage explicitly
Currently, the `<RequiredTargetFrameworks>` property for tests lists `net461` first so that the last (a comment says it's `netcoreapp3.1`, but that comment wasn't updated when `net6.0` was added last) is used for coverage, but then that means when you record `net461` runs first and actually exacerbates other issues e.g., #27695. Instead, maybe we could list the desired order we want for tests to run (if sequential, which they are by default) and only record in explicitly-listed runtimes for coverage (multiple may be good since some libraries multi-target or use reflection-based proxies for light-up support).

heaths:


heaths:
I needed to use a property function. If I use a separate property like so:

```xml
<LatestTargetFramework>net6.0</LatestTargetFramework>
<RequiredFrameworks>$(LatestTargetFramework);netcoreapp3.1</RequiredTargetFrameworks>

<CodeCoverageTargetFramework>$(LatestTargetFramework)</CodeCoverageTargetFramework>
```

VSTest doesn't seem to actually recurse down into property evaluations enough, so the first literal string it picked up was `netcoreapp3.1`, which is what it then used.

Instead, this way, it does use the literal `net6.0`.

benbp (approved):

nrz77 (approved):

jsquire (approved):
Thanks, @heaths!

heaths:
Adding `"CollectCoverage": true` where I did now means we collect coverage on multiple agents when we don't need to. I don't think it'll hurt much since we rarely vary code coverage based on `Configugration` or other parameters, but it's possible. Is there a way, @benbp, we can do this on just one?

Maybe we add back a single `include`, but this seems like an opportunity to reduce our agent count across all the SDKs (by 1, but it adds up). /cc @weshaggard if that's even worth the hassle if applying this to only one agent matrix slice isn't supported now.

benbp:
That's been the challenge in the past as well. A lot of the matrix configs kept one of their version tests to only run via the included coverage step, but because NET has the desire to run more version+OS combinations, that approach doesn't work as well.

heaths:
Per offline discussion with @benbp, I think we need to add the `include` back and use another agent like we were before. Separately, I think I'm going about this wrong. Currently, _CodeCoverage.targets_ is included in all projects to work around an issue before netcoreapp3.1: https://github.com/coverlet-coverage/coverlet/blob/master/Documentation/DeterministicBuild.md. We could exclude that now, but need to make sure we're setting [`DeterministicSourcePaths=true`](https://github.com/dotnet/sourcelink/tree/main/docs#deterministicsourcepaths). Looking at a package we built as part of this PR, it did not have deterministic sources so I turn that on explicitly. The workaround was probably still applying even though we didn't need it to.

Though, I do remember the reason we only collect in the latest is because results overwrite each other. Maybe we can prevent that another way.

heaths:
With all the different attempts to filter to the latest - whether run by CI or by a user cross-building - perhaps its better not to attempt to and just merge any reports together. This would also give total coverage over all target frameworks where code branches.

eng/CodeCoverage.targets
```diff
@@ -4,6 +4,8 @@
     <_IsCodeCoverable Condition="'$(IsClientLibrary)' == 'true'">true</_IsCodeCoverable>
   </PropertyGroup>
 
+  <!-- BUGBUG: When a specific framework is passed IsCrossTargetingBuild is unset but IsInnerBuild is still true. Cannot clean up past TestResults, but seeminly an edge case. -->
```
weshaggard, line 7:
```suggestion
  <!-- BUGBUG: When a specific framework is passed IsCrossTargetingBuild is unset but IsInnerBuild is still true. Cannot clean up past TestResults, but seemingly an edge case. -->
```

eng/CodeCoverage.targets
```diff
@@ -91,7 +98,7 @@
   <Target Name="CoverletGetPathMap"
           DependsOnTargets="InitializeSourceRootMappedPaths"
           Returns="@(_LocalTopLevelSourceRoot)"
-          Condition="'$(DeterministicSourcePaths)' == 'true'">
+          Condition="'$(DeterministicSourcePaths)' == 'true' and $([MSBuild]::VersionLessThan('$(TargetFrameworkVersion)', '3.1'))">
```
weshaggard, line 101: Why is this interesting? Or is this temporary until you remove 3.1 support?

heaths: We still build for `net461` (and `net47` in some cases) that need this support. Not to mention this also needs to impact the shipping nupkgs which we target `netstandard2.0` and soon `net461` in addition.

eng/pipelines/templates/stages/platform-matrix.json
```diff
@@ -20,7 +20,7 @@
         "OSVmImage": "MMS2019",
         "Pool": "azsdk-pool-mms-win-2019-general",
         "TestTargetFramework": "net6.0"
-      },
+},
```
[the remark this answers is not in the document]
heaths: I went ahead and reformatted the whole file. There were a few other incosistencies.

eng/Directory.Build.Common.props
```diff
@@ -93,6 +93,8 @@
 
   <PropertyGroup Condition="'$(IsShippingClientLibrary)' == 'true' and '$(TF_BUILD)' == 'true'">
     <ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
+    <!-- See https://github.com/dotnet/sourcelink/tree/main/docs#deterministicsourcepaths for details. -->
+    <DeterministicSourcePaths>true</DeterministicSourcePaths>
```
weshaggard, line 97: This seems unrelated. Is there a reason for this update?

weshaggard (approved):

eng/Directory.Build.Common.props
heaths: It's related but I'm definitely going to make sure the right thing keeps happening. That block I conditioned on < netcoreapp3.1? The point is to make symbols work for nupkgs but >= netcoreapp3.1 does the right thing so long as `DeterministicSourcePaths` is true. According to docs, as long as `ContinuousIntegrationBuild` and `Deterministic` is set, this property gets set to `true`. While I know `ContinuousIntegrationBuild` gets set to `true` in our CIs, I can't find for sure what sets `Deterministic` to true and want to make sure, now that I'm conditioning this older block, it's always set correctly.

heaths:
Looking at a release build, the nupkg is showing it's non-deterministic. While this seems to be the case for any CI-produced nupkg, release builds have been deterministic. Need to investigate more.

eng/pipelines/templates/steps/dotnet-diagnostics.yml
```diff
@@ -0,0 +1,13 @@
+parameters:
+- name: LogFilePath
+  type: string
+  default: $(Build.ArtifactStagingDirectory)/msbuild.binlog
+
+steps:
+- pwsh: |
+    if ('${{ eq(variables['System.Debug'], 'true') }}' -eq 'true') {
```
benbp, line 8: It still won't work if you're trying `${{ }}` because it gets parsed/rendered too early.

```suggestion
    if ('$(System.Debug)' -eq 'true') {
```

weshaggard (approved):

heaths:
Given Key Vault is old and kind of funky because of samples (though those conditions should only affect samples), I tried a build using Cognitive Language using both this PR merge branch and main and still got the same results in NuGet Explorer for both unsigned and signed nupkgs:

image

There's nothing in my PR that I can think would affect this, and pouring through binlogs hasn't revealed anything other than various properties like `ContinuousIntegrationBuild`, `Deterministic`, and `DeterministicSourcePaths` don't appear to be defined for *every* project, but I'm wondering if this is more a way that Structured Log Viewer is just displaying results. Looking at our .props and .targets, all these projects absolutely should define `IsClientLibrary=true` and `IsShippingLibrary=true` but only some do.

Given this PR doesn't seem to regress any behavior, @weshaggard @hallipr @benbp should we merge it and keep an eye on builds?

weshaggard:
>Given this PR doesn't seem to regress any behavior, @weshaggard @hallipr @benbp should we merge it and keep an eye on builds?

Yes lets move forward with merging this for the coverage update and I suggest watching some builds to make sure things on this front are still working as expected. It is possible that they might have broken or changed at some point as well.

heaths:
After talking with a PM on the NuGet team, this has probably long not been a problem: turns out, for those 3 bottom checks (the red ones) you either need the symbols in the package or in a snupkg next to the nupkg, or the snupkg published like we do for previously-released packages, which is why previous builds showed green. Sure enough, when I look at the packages from release builds against both main (back) and this PR (fore):

image

PR #32604, opened 2022-11-22
Add Cloud property to CommunicationUserIdentifier

maximrytych-ms:
Added Cloud property to CommunicationUserIdentifier. ACS is (or will be) supported not only in public cloud, so it makes sense to explicitly surface the value. This will allow to identify the cloud without the need to parse MRI. Consumers won't need to keep their custom parsing logic - it will be part of common package.

### DIFF
--- a/sdk/communication/Azure.Communication.Common/src/CommunicationCloudEnvironment.cs
+++ b/sdk/communication/Azure.Communication.Common/src/CommunicationCloudEnvironment.cs
@@ -19,6 +19,9 @@ namespace Azure.Communication
         private const string PublicValue = "public";
         private const string DodValue = "dod";
         private const string GcchValue = "gcch";
+        private const string AirGap08Value = "airgap08";
+        private const string AirGap09Value = "airgap09";
+        private const string GallatinValue = "gallatin";
 
         /// <summary> public. </summary>
         public static CommunicationCloudEnvironment Public { get; } = new CommunicationCloudEnvironment(PublicValue);
@@ -26,6 +29,12 @@ namespace Azure.Communication
         public static CommunicationCloudEnvironment Dod { get; } = new CommunicationCloudEnvironment(DodValue);
         /// <summary> gcch. </summary>
         public static CommunicationCloudEnvironment Gcch { get; } = new CommunicationCloudEnvironment(GcchValue);
+        /// <summary> airgap08. </summary>
+        public static CommunicationCloudEnvironment AirGap08 { get; } = new CommunicationCloudEnvironment(AirGap08Value);
+        /// <summary> airgap09. </summary>
+        public static CommunicationCloudEnvironment AirGap09 { get; } = new CommunicationCloudEnvironment(AirGap09Value);
+        /// <summary> gallatin. </summary>
+        public static CommunicationCloudEnvironment Gallatin { get; } = new CommunicationCloudEnvironment(GallatinValue);
         /// <summary> Determines if two <see cref="CommunicationCloudEnvironment"/> values are the same. </summary>
         public static bool operator ==(CommunicationCloudEnvironment left, CommunicationCloudEnvironment right) => left.Equals(right);
         /// <summary> Determines if two <see cref="CommunicationCloudEnvironment"/> values are not the same. </summary>
--- a/sdk/communication/Azure.Communication.Common/src/CommunicationCloudEnvironment.cs
+++ b/sdk/communication/Azure.Communication.Common/src/CommunicationCloudEnvironment.cs
@@ -19,9 +19,6 @@ namespace Azure.Communication
         private const string PublicValue = "public";
         private const string DodValue = "dod";
         private const string GcchValue = "gcch";
-        private const string AirGap08Value = "airgap08";
-        private const string AirGap09Value = "airgap09";
-        private const string GallatinValue = "gallatin";
 
         /// <summary> public. </summary>
         public static CommunicationCloudEnvironment Public { get; } = new CommunicationCloudEnvironment(PublicValue);
@@ -29,12 +26,6 @@ namespace Azure.Communication
         public static CommunicationCloudEnvironment Dod { get; } = new CommunicationCloudEnvironment(DodValue);
         /// <summary> gcch. </summary>
         public static CommunicationCloudEnvironment Gcch { get; } = new CommunicationCloudEnvironment(GcchValue);
-        /// <summary> airgap08. </summary>
-        public static CommunicationCloudEnvironment AirGap08 { get; } = new CommunicationCloudEnvironment(AirGap08Value);
-        /// <summary> airgap09. </summary>
-        public static CommunicationCloudEnvironment AirGap09 { get; } = new CommunicationCloudEnvironment(AirGap09Value);
-        /// <summary> gallatin. </summary>
-        public static CommunicationCloudEnvironment Gallatin { get; } = new CommunicationCloudEnvironment(GallatinValue);
         /// <summary> Determines if two <see cref="CommunicationCloudEnvironment"/> values are the same. </summary>
         public static bool operator ==(CommunicationCloudEnvironment left, CommunicationCloudEnvironment right) => left.Equals(right);
         /// <summary> Determines if two <see cref="CommunicationCloudEnvironment"/> values are not the same. </summary>
--- a/sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
+++ b/sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
@@ -10,6 +10,18 @@ namespace Azure.Communication
     /// <summary>Represents an identifier in Azure Communication Services.</summary>
     public abstract class CommunicationIdentifier : IEquatable<CommunicationIdentifier>
     {
+        internal const string TEAMS_USER_VISITOR_PREFIX = "8:teamsvisitor:";
+        internal const string TEAMS_USER_ORGID_PREFIX = "8:orgid:";
+        internal const string TEAMS_USER_DOD_PREFIX = "8:dod:";
+        internal const string TEAMS_USER_GCCH_PREFIX = "8:gcch:";
+        internal const string ACS_USER_ACS_PREFIX = "8:acs:";
+        internal const string ACS_USER_SPOOL_PREFIX = "8:spool:";
+        internal const string ACS_USER_GCCH_PREFIX = "8:gcch-acs:";
+        internal const string ACS_USER_AIR_GAP_08_PREFIX = "8:ag08-acs:";
+        internal const string ACS_USER_AIR_GAP_09_PREFIX = "8:ag09-acs:";
+        internal const string ACS_GALLATIN_PREFIX = "8:gal-acs:";
+        internal const string ACS_USER_DOD_PREFIX = "8:dod-acs:";
+
         /// <summary>
         /// Returns the canonical string representation of the <see cref="CommunicationIdentifier"/>.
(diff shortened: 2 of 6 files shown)


sdk/communication/Azure.Communication.Common/src/CommunicationUserIdentifier.cs
```diff
@@ -27,10 +30,26 @@ public class CommunicationUserIdentifier : CommunicationIdentifier
...
+        /// <exception cref="System.ArgumentNullException">
+        /// Thrown when the <paramref name="id"/> is null.
+        /// </exception>
+        /// <exception cref="System.ArgumentException">
+        /// Thrown when the <paramref name="id"/> is empty.
+        /// </exception>
+        /// <param name="cloud">The cloud that the ACS user belongs to.</param>
+        public CommunicationUserIdentifier(string id, CommunicationCloudEnvironment cloud)
         {
             Argument.AssertNotNullOrEmpty(id, nameof(id));
```
ostoliarova-msft, line 50: Do you allow to have null for `cloud` or should there also be the same verification as with `id`?

sdk/communication/Azure.Communication.Common/src/MicrosoftTeamsUserIdentifier.cs
```diff
@@ -22,19 +22,19 @@ public override string RawId
...
                     }
                     else if (Cloud == CommunicationCloudEnvironment.Gcch)
                     {
-                        _rawId = $"8:gcch:{UserId}";
+                        _rawId = $"{TEAMS_USER_GCCH_PREFIX}{UserId}";
                     }
                     else
                     {
-                        _rawId = $"8:orgid:{UserId}";
+                        _rawId = $"{TEAMS_USER_ORGID_PREFIX}{UserId}";
```
ostoliarova-msft, line 37: Should we also handle other prefixes (like airgap08 etc) that you've added in this PR and have more `else` branches (and maybe look into using `switch`/refactoring into a separate method)?

ostoliarova-msft:
Should there also be a change to (de)serialization in `CommunicationIdentifierSerializer`?

sdk/communication/Azure.Communication.Common/src/CommunicationUserIdentifier.cs
maximrytych-ms: Good point. Cloud shouldn't be null. Will update.

sdk/communication/Azure.Communication.Common/src/MicrosoftTeamsUserIdentifier.cs
maximrytych-ms: If I understand it correctly, this is for teams user. I am working now with ACS ids, they have different prefix. I think this logic is specific to teams user...

petrsvihlik: The question is whether any of the newly added cloud envs is relevant to `MicrosoftTeamsUserIdentifier`...and whether this logic should be extended too... That's what Olena is asking about.

sdk/communication/Azure.Communication.Common/src/CommunicationUserIdentifier.cs
```diff
@@ -27,10 +30,26 @@ public class CommunicationUserIdentifier : CommunicationIdentifier
...
+        /// </summary>
+        /// <param name="id">Id of the communication user.</param>
+        /// <exception cref="System.ArgumentNullException">
+        /// Thrown when the <paramref name="id"/> is null.
+        /// </exception>
+        /// <exception cref="System.ArgumentException">
+        /// Thrown when the <paramref name="id"/> is empty.
+        /// </exception>
+        /// <param name="cloud">The cloud that the ACS user belongs to.</param>
+        public CommunicationUserIdentifier(string id, CommunicationCloudEnvironment cloud)
```
petrsvihlik, line 48: What if Contoso passes `("8:dod-acs:bbbcbc1e-9f06-482a-b5d8-20e3f26ef0cd_45ab2481-1c1c-4005-be24-0ffb879b1130", CommunicationCloudEnvironment.Gcch)`?

I think, unlike in the `MicrosoftTeamsUserIdentifier`, we should infer the value of `cloud` from the `id` and discard this constructor.

petrsvihlik: This will likely require extraction of the cloud-parsing logic (`string` -> `CommunicationCloudEnvironment`) into a separate method. This method will then be also useful for the `CommunicationIdentifier.FromRawId()` method.
Make sure this logic is resilient to potentially newly added env values (utilize the extensible enum capability of accepting a `string` in the constructor.

sdk/communication/Azure.Communication.Common/src/CommunicationCloudEnvironment.cs
```diff
@@ -19,13 +19,22 @@ public CommunicationCloudEnvironment(string value)
         private const string PublicValue = "public";
         private const string DodValue = "dod";
         private const string GcchValue = "gcch";
+        private const string AirGap08Value = "airgap08";
+        private const string AirGap09Value = "airgap09";
+        private const string GallatinValue = "gallatin";
```
petrsvihlik, line 24: Can you please get a confirmation from an authority that this list is complete (to this date)?

sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
```diff
@@ -68,11 +80,16 @@ public static CommunicationIdentifier FromRawId(string rawId)
...
+                TEAMS_USER_VISITOR_PREFIX => new MicrosoftTeamsUserIdentifier(suffix, true),
+                TEAMS_USER_ORGID_PREFIX => new MicrosoftTeamsUserIdentifier(suffix, false, CommunicationCloudEnvironment.Public),
+                TEAMS_USER_DOD_PREFIX => new MicrosoftTeamsUserIdentifier(suffix, false, CommunicationCloudEnvironment.Dod),
+                TEAMS_USER_GCCH_PREFIX => new MicrosoftTeamsUserIdentifier(suffix, false, CommunicationCloudEnvironment.Gcch),
+                ACS_USER_ACS_PREFIX or ACS_USER_SPOOL_PREFIX => new CommunicationUserIdentifier(rawId, CommunicationCloudEnvironment.Public),
+                ACS_USER_GCCH_PREFIX => new CommunicationUserIdentifier(rawId, CommunicationCloudEnvironment.Gcch),
+                ACS_USER_AIR_GAP_08_PREFIX => new CommunicationUserIdentifier(rawId, CommunicationCloudEnvironment.AirGap08),
+                ACS_USER_AIR_GAP_09_PREFIX => new CommunicationUserIdentifier(rawId, CommunicationCloudEnvironment.AirGap09),
+                ACS_GALLATIN_PREFIX => new CommunicationUserIdentifier(rawId, CommunicationCloudEnvironment.Gallatin),
+                ACS_USER_DOD_PREFIX => new CommunicationUserIdentifier(rawId),
```
petrsvihlik, line 92: this seems incorrect - `ACS_USER_DOD_PREFIX => new CommunicationUserIdentifier(rawId),`

anyway, this will likely not be needed at all...see my other comment about inferring the cloud env from the id

sdk/communication/Azure.Communication.Common/tests/CommunicationIdentifierTest.cs
```diff
@@ -32,6 +32,18 @@ public void RawIdTakesPrecendenceInEqualityCheck()
         public void MicrosoftTeamsUserIdentifier_DefaultCloudIsPublic()
             => Assert.AreEqual(CommunicationCloudEnvironment.Public, new MicrosoftTeamsUserIdentifier("45ab2481-1c1c-4005-be24-0ffb879b1130", isAnonymous: true, rawId: "Raw Id").Cloud);
 
+        [Test]
+        public void RawIdDeserializesToCorrectAcsCloudType()
+        {
+            Assert.AreEqual(CommunicationCloudEnvironment.Public, (CommunicationIdentifier.FromRawId("8:acs:bbbcbc1e-9f06-482a-b5d8-20e3f26ef0cd_45ab2481-1c1c-4005-be24-0ffb879b1130") as CommunicationUserIdentifier)?.Cloud);
```
petrsvihlik, line 38: I wonder why the `?`. Doesn't the `FromRawId` method always return a value?

petrsvihlik (changes requested):

petrsvihlik (changes requested):
plus, please adjust the logic of

`azure-sdk-for-net\sdk\communication\Shared\src\CommunicationIdentifierSerializer.cs`

sdk/communication/Azure.Communication.Common/src/MicrosoftTeamsUserIdentifier.cs
maximrytych-ms: I believe that now with removed clouds I originally added this logic doesn't need changes.

sdk/communication/Azure.Communication.Common/src/CommunicationCloudEnvironment.cs
maximrytych-ms: Confirmed with Auth team. These clouds won't be supported.

sdk/communication/Azure.Communication.Common/src/CommunicationUserIdentifier.cs
maximrytych-ms: I removed the constructor altogether. Cloud property is now set in original constructor using dedicated private method.

sdk/communication/Azure.Communication.Common/src/CommunicationUserIdentifier.cs
```diff
@@ -11,6 +11,9 @@ public class CommunicationUserIdentifier : CommunicationIdentifier
         /// <summary>The id of the communication user.</summary>
         public string Id { get; }
 
+        /// <summary> The cloud that the identifier belongs to.</summary>
```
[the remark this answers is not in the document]
maximrytych-ms: Fixed.

sdk/communication/Azure.Communication.Common/src/CommunicationUserIdentifier.cs
maximrytych-ms: I added private method for `RawId` -> `CommunicationCloudEnvironment` transformation to `CommunicationUserIdentifier`. I don't see how can I utilize it in the `CommunicationIdentifier.FromRawId()` method. Also not sure about the comment on the enum capability. Can you elaborate a bit more, pls?

sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
maximrytych-ms: This should be ok. Following logic will assign default, i.e. `CommunicationCloudEnvironment.Public` value. Or we can use `UnknownIdentifier`. Not sure what makes more sense.

sdk/communication/Azure.Communication.Common/tests/CommunicationIdentifierTest.cs
maximrytych-ms: It does. But it returns instance of `CommunicationIdentifier` type that doesn't contain `Cloud` property, so I need to cast it to `CommunicationUserIdentifier` that does. I guess it's then not able to infer that the property can't be `null`. Anyway, without the check I am getting a compiler error, so this was the way how to get rid of it :)

ostoliarova-msft:
The feature is in the backlog, commenting to leave the PR open

PR #32607, opened 2022-11-22
[Docs] Fixing broken links

jsquire:
# Summary

The focus of these changes is to fix a set of broken links to the MS Docs resources which moved during their recent view reorganization.

# References and Resources

- [Link verification run](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2005181&view=logs&j=d41e9505-d81b-5b3c-5b50-be71bc9b0ccb&t=2521f384-3a28-5bc9-41e5-2f3e6ef99ecd) _(Microsoft internal)_

### DIFF
--- a/sdk/eventgrid/README.md
+++ b/sdk/eventgrid/README.md
@@ -32,4 +32,4 @@ Use the following library to work with the Azure Event Grid resource provider:
 
 | Nuget Package | Reference |
 |--------------------------------------|---------------------------------------------------------------|
-| [Microsoft.Azure.Management.EventGrid](https://www.nuget.org/packages/Microsoft.Azure.Management.EventGrid)  | [API Reference for Microsoft.Azure.Management.EventGrid](https://docs.microsoft.com/dotnet/api/overview/azure/eventgrid/management?view=azure-dotnet)  |
+| [Microsoft.Azure.Management.EventGrid](https://www.nuget.org/packages/Microsoft.Azure.Management.EventGrid)  | [API Reference for Microsoft.Azure.Management.EventGrid](https://learn.microsoft.com/dotnet/api/overview/azure/eventgrid/management/management-eventgrid?view=azure-dotnet)  |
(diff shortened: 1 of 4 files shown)


sdk/eventgrid/README.md
```diff
@@ -32,4 +32,4 @@ Use the following library to work with the Azure Event Grid resource provider:
 
 | Nuget Package | Reference |
 |--------------------------------------|---------------------------------------------------------------|
-| [Microsoft.Azure.Management.EventGrid](https://www.nuget.org/packages/Microsoft.Azure.Management.EventGrid)  | [API Reference for Microsoft.Azure.Management.EventGrid](https://docs.microsoft.com/dotnet/api/overview/azure/eventgrid/management?view=azure-dotnet)  |
\ No newline at end of file
+| [Microsoft.Azure.Management.EventGrid](https://www.nuget.org/packages/Microsoft.Azure.Management.EventGrid)  | [API Reference for Microsoft.Azure.Management.EventGrid](https://learn.microsoft.com/dotnet/api/overview/azure/eventgrid/management/management-eventgrid?view=azure-dotnet)  |
```
jsquire, line 35: It is intentional that link differs from the pattern for the other two libraries; this is the structure in MS Docs.  I believe the difference is because there is only a T1 management package for Event Grid and the other two have both T1 and T2 packages.

sdk/monitor/Azure.Monitor.Ingestion/src/Azure.Monitor.Ingestion.csproj
```diff
@@ -4,7 +4,7 @@
     <AssemblyTitle>Azure Monitor Ingestion client library</AssemblyTitle>
     <Version>1.0.0-beta.5</Version>
     <PackageTags>Azure Monitor Ingestion</PackageTags>
-    <TargetFrameworks>$(RequiredTargetFrameworks);net461;net5.0</TargetFrameworks>
+    <TargetFrameworks>$(RequiredTargetFrameworks)</TargetFrameworks>
```
jsquire, line 7: CI was no longer passing due to `net5.0` being out of support.   These look like they were added as part of a POC for GZip support and should no longer be neded.

_([ref](https://github.com/Azure/azure-sdk-for-net/pull/29426))_

pallavit:
Should we run link verification step per SDK too?

jsquire:
> Should we run link verification step per SDK too?

That's part of a different pipeline that we can't trigger from a PR, unfortunately.

JoshLove-msft (approved):

PR #32613, opened 2022-11-22
Update JsonData in Experimental

annelo-msft:
Addresses https://github.com/Azure/azure-sdk-for-net/issues/30717.

This PR does the following:
- Moves JsonData into Azure.Core.Dynamic namespace
- Removes public constructors and members from JsonData, with the exception of overloaded operators
- Adds a `ToDynamic()` extension method to BinaryData, to enable `response.Content.ToDynamic()`
- Returns null instead of throwing when an absent property is accessed (to mimic optional properties on Azure SDK models)
- Enables access to indexers through dynamic interface
- Disposes JsonDocument in calls where one is created
- Adds a test project and tests that only have access to the JsonData public API surface
- Adds tests for JSON object, array, leaf nodes separately

The following will be handled in a later PR:
- [ ] Making exception types consistent: https://github.com/Azure/azure-sdk-for-net/pull/32613#discussion_r1038645125
- [ ] Improving exception messages: https://github.com/Azure/azure-sdk-for-net/pull/32613/files#r1044805511
- [ ] Letting caller access properties as CamelCase: https://github.com/Azure/azure-sdk-for-net/pull/32613#discussion_r1038651832
- [ ] Deciding whether to support Azure.Core ObjectSerializer: https://github.com/Azure/azure-sdk-for-net/pull/32613#discussion_r1038652140
- [ ] Support for writing to a stream: https://github.com/Azure/azure-sdk-for-net/pull/32613#discussion_r1038652662
- [ ] Handling non-JSON data: https://github.com/Azure/azure-sdk-for-net/pull/32613#discussion_r1039893046
- [ ] Adding full set of equality operators: https://github.com/Azure/azure-sdk-for-net/pull/32613#discussion_r1039898091
- [ ] Perf improvements: 
    - [ ] https://github.com/Azure/azure-sdk-for-net/pull/32613#discussion_r1038652986
    - [ ] https://github.com/Azure/azure-sdk-for-net/pull/32613#discussion_r1039911220
- [ ] Remove inline TODOs


sdk/core/Azure.Core.Experimental/api/Azure.Core.Experimental.netstandard2.0.cs
```diff
@@ -106,91 +106,39 @@ namespace Azure
...
-        public void Set(string propertyName, long value) { }
-        public Azure.Core.JsonData Set(string propertyName, object? serializable) { throw null; }
-        public Azure.Core.JsonData Set(string propertyName, object? serializable, System.Text.Json.JsonSerializerOptions options) { throw null; }
-        public void Set(string propertyName, float value) { }
-        public void Set(string propertyName, string? value) { }
-        public Azure.Core.JsonData SetEmptyArray(string propertyName) { throw null; }
-        public Azure.Core.JsonData SetEmptyObject(string propertyName) { throw null; }
-        public Azure.Core.JsonData Set<T>(string propertyName, T[] serializable) { throw null; }
-        public Azure.Core.JsonData Set<T>(string propertyName, T[] serializable, System.Text.Json.JsonSerializerOptions options) { throw null; }
+        public static bool operator ==(Azure.Core.Dynamic.JsonData? left, int right) { throw null; }
```
annelo-msft, line 122: Is this the right list of types to support with equality operators?
- string
- bool
- int
- double
- long
- float

sdk/core/Azure.Core.Experimental/tests/public/JsonDataObjectTests.cs
```diff
@@ -0,0 +1,133 @@
...
+        [Test]
+        public void CannotConvertObjectToLeaf()
+        {
+            dynamic data = JsonDataTestHelpers.CreateFromJson(@"{ ""value"": 5 }");
+
+            // TODO: Standardize Exception types.
+            Assert.Throws<InvalidOperationException>(() => { var i = (int)data; });
+            Assert.Throws<InvalidOperationException>(() => { var b = (bool)data; });
+            Assert.Throws<InvalidOperationException>(() => { var s = (string)data; });
+            Assert.Throws<JsonException>(() => { var time = (DateTime)data; });
```
annelo-msft, line 23: We should provide a consistent exception experience to aid developers working with the type through the dynamic interface.

sdk/core/Azure.Core.Experimental/tests/public/Azure.Core.Experimental.Tests.Public.csproj
```diff
@@ -0,0 +1,28 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
annelo-msft, line 1: Is this a good way to add tests for Core types where we don't want InternalsVisibleTo?  I'm super open to approaches here.  @JoshLove-msft, @christothes, and thoughts on this?

sdk/core/Azure.Core.Experimental/tests/public/JsonDataPublicTests.cs
```diff
@@ -0,0 +1,400 @@
...
+        {
+            dynamic jsonData = new BinaryData(new { value = "foo" }).ToDynamic();
+
+            List<string> list = new();
+
+            // TODO: Should we add an implicit cast to string?
+            // Can we do it dynamically?
+            // Alternatively, we could provide a different type for string leaf
+            // nodes, like we do for numbers.
+            list.Add(jsonData.value);
```
annelo-msft, line 135: This fails with a:

`Microsoft.CSharp.RuntimeBinder.RuntimeBinderException : The best overloaded method match for 'System.Collections.Generic.List<string>.Add(string)' has some invalid arguments`

because jsonData.value's underlying type is JsonData, and there is no implicit conversion to string.  The DynamicMetaObject's BindConvert method is never invoked here, so we do not have the opportunity there to do the conversion.  Are we ok requiring a cast to string to make this succeed?  The cons to requiring the cast are that it is inconsistent with the usage of Azure SDK model types, and the compiler doesn't enforce it so developers won't discover the need for the cast until an exception is thrown at runtime.

sdk/core/Azure.Core.Experimental/tests/public/JsonDataPublicTests.cs
```diff
@@ -0,0 +1,400 @@
...
+        }
+
+        [Test]
+        public void EqualsForStringNUnit()
+        {
+            dynamic foo = new BinaryData("{ \"value\": \"foo\" }").ToDynamic();
+            var value = foo.Value;
+
+            // TODO: Outstanding question regarding whether we want these to succeed without a cast.
+            Assert.AreEqual(value, "foo");
```
annelo-msft, line 393: This is the same question about whether we should require a cast to a primitive type.  NUnit does not use the provided `==` operator and BindConvert is never invoked.

sdk/core/Azure.Core.Experimental/tests/public/JsonDataPublicTests.cs
```diff
@@ -0,0 +1,400 @@
...
+            Assert.AreEqual(5, list[0]);
+        }
+
+        [Test]
+        public void GetMemberIsCaseInsensitive()
+        {
+            dynamic jsonData = new BinaryData("{ \"primitive\":\"Hello\", \"nested\": { \"nestedPrimitive\":true } }").ToDynamic();
+
+            Assert.AreEqual("Hello", (string)jsonData.Primitive);
+            Assert.AreEqual(true, (bool)jsonData.Nested.NestedPrimitive);
```
annelo-msft, line 162: If we want to make this configurable (see discussion [here](https://github.com/Azure/azure-sdk-for-net/pull/31769#discussion_r1017018043)), we would probably expose this as letting callers pass an options type to BinaryData's ToDynamic() method.  This raises several questions, for example, do we want to create an options type for JsonData/DynamicData to control this, or would we want to piggyback on JsonSerializerOptions PropertyNameCaseInsensitive property?

sdk/core/Azure.Core.Experimental/src/JsonData.cs
```diff
@@ -65,7 +76,7 @@ public JsonData(object? value) : this(value, DefaultJsonSerializerOptions)
         /// <param name="value">The value to convert.</param>
         /// <param name="options">Options to control the conversion behavior.</param>
         /// <param name="type">The type of the value to convert. </param>
-        public JsonData(object? value, JsonSerializerOptions options, Type? type = null)
+        internal JsonData(object? value, JsonSerializerOptions options, Type? type = null)
```
annelo-msft, line 79: Would we want to also expose a way to use this with Azure.Core's ObjectSerializer?

sdk/core/Azure.Core.Experimental/src/JsonData.cs
```diff
@@ -879,38 +423,11 @@ public IEnumerable<string> Properties
...
-            return writer.BytesCommitted;
-        }
-
-        /// <summary>
-        /// Writes the UTF-8 encoded string representation of this instance.
-        /// </summary>
-        /// <param name="stream">The stream to write to.</param>
-        /// <param name="cancellationToken">A token to monitor for cancelation requests.</param>
-        /// <returns>The number of bytes written into the stream.</returns>
-        public async Task<long> WriteToAsync(Stream stream, CancellationToken cancellationToken)
```
annelo-msft, line 906: I removed this before understanding the requirement to get the JsonData as a stream.  We no longer want this as a public API on JsonData, but I will add this capability back in a different way.

sdk/core/Azure.Core.Experimental/src/JsonData.cs
```diff
@@ -116,40 +129,6 @@ private JsonData(JsonElement element)
...
-                    _arrayRepresentation.Add(new JsonData((object?)null));
-                }
-                else
-                {
-                    _arrayRepresentation.Add(item);
-                }
-            }
-        }
-
         private void InitFromElement(JsonElement element)
```
annelo-msft, line 132: This approach to creating the type is a clear performance concern, and will be addressed in this or a later PR.

sdk/core/Azure.Core.Experimental/tests/public/Azure.Core.Experimental.Tests.Public.csproj
christothes: Do you have an example of something that requires it to test properly?

sdk/core/Azure.Core.Experimental/api/Azure.Core.Experimental.netstandard2.0.cs
annelo-msft: Cool, I'll add these.

sdk/core/Azure.Core.Experimental/tests/public/Azure.Core.Experimental.Tests.Public.csproj
annelo-msft: There are a couple of categories.
			
One is that we have a lot of internal methods on JsonData, so in test projects where we have InternalsVisibleTo, it's easy to accidentally write a test to mimic a user scenario that calls an internal method when a library user wouldn't actually be able to call this.  Adding a test project that only has access to Azure.Core public APIs gives us compiler checks to prevent this.

The second category is where we're using dynamic APIs going through the DynamicMetaObject interface and a member is available internally but not publicly.  The DMO interface obfuscates access modifier rules and it's another case where it's easy to write something you think is available publicly that's not, and it's nice to have automated validation of this both now, and for future modifications to the type.

I'm working on getting a clear test that both compiles and fails in public tests but succeeds in internal tests.  I had one with JsonData was internal with public members, now that it is public with internal members I'm having a harder time finding one, but trying to pin down enough specifics to do this.  I hope I can get this before next week when I'm back from jury duty so we can move forward on the approach in the interim.

@christothes can you share your specific concerns with this approach and what information might motivate you to invest in further thinking in this area?

sdk/core/Azure.Core.Experimental/api/Azure.Core.Experimental.netstandard2.0.cs
```diff
@@ -106,91 +106,39 @@ namespace Azure
...
-namespace Azure.Core
+namespace Azure.Core.Dynamic
 {
+    public static partial class BinaryDataExtensions
+    {
+        public static dynamic ToDynamic(this System.BinaryData data) { throw null; }
+    }
     [System.Diagnostics.DebuggerDisplayAttribute("{DebuggerDisplay,nq}")]
-    public partial class JsonData : System.Dynamic.IDynamicMetaObjectProvider, System.IEquatable<Azure.Core.JsonData>
+    public partial class JsonData : System.Dynamic.IDynamicMetaObjectProvider, System.IEquatable<Azure.Core.Dynamic.JsonData>
```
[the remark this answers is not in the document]
annelo-msft: True.  For merging to Experimental, I think we should keep JsonData.  We had also previously discussed that we could implement dynamic XML support separately in a different named type.  Let me know how strongly you feel about this for this specific PR.

sdk/core/Azure.Core.Experimental/tests/public/JsonDataLeafTests.cs
```diff
@@ -0,0 +1,300 @@
...
+            );
+        }
+        #endregion
+
+        #region Array behaviors tests
+
+        [Test]
+        public void CannotEnumerateLeaf()
+        {
+            // TODO: This exception says "exepcted kind to be object" - improve this.
```
annelo-msft, line 279: Improve exception messages.

sdk/core/Azure.Core.Experimental/tests/public/Azure.Core.Experimental.Tests.Public.csproj
annelo-msft: Thanks @JoshLove-msft for the help troubleshooting this!

sdk/core/Azure.Core.Experimental/src/JsonData.cs
```diff
@@ -1247,6 +821,27 @@ public override DynamicMetaObject BindSetMember(SetMemberBinder binder, DynamicM
                 DynamicMetaObject setProperty = new DynamicMetaObject(setPropertyCall, restrictions);
                 return setProperty;
             }
+
+            public override DynamicMetaObject BindSetIndex(SetIndexBinder binder, DynamicMetaObject[] indexes, DynamicMetaObject value)
+            {
+                var targetObject = Expression.Convert(Expression, LimitType);
+                var arguments = new Expression[2] {
```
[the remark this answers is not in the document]
annelo-msft: I'll do another perf pass over this in a later PR.  I've flagged this comment to follow up on in the PR description notes.

sdk/core/Azure.Core.Experimental/src/JsonData.cs
```diff
@@ -972,6 +319,7 @@ public override int GetHashCode()
         }
 
         private string? GetString() => (string?)EnsureValue();
+
         private int GetIn32()
```
christothes, line 323: nit: should this be named `GetInt`, since it is not resticted to 32 bit?

sdk/core/Azure.Core.Experimental/src/JsonData.cs
```diff
@@ -972,6 +319,7 @@ public override int GetHashCode()
...
         {
             var value = EnsureNumberValue().AsLong();
```
christothes, line 325: Is there an int version of this method, or does it not matter for our scenario?

sdk/core/Azure.Core.Experimental/src/JsonData.cs
```diff
@@ -1058,19 +408,32 @@ private JsonData GetPropertyValue(string propertyName)
         /// </summary>
         /// <param name="propertyName">The name of the property to get the value of.</param>
         /// <returns></returns>
-        private object GetDynamicProperty(string propertyName)
+        private object? GetDynamicProperty(string propertyName)
```
christothes, line 411: nit: should this be named `GetDynamicPropertyValue`?

sdk/core/Azure.Core.Experimental/src/JsonData.cs
```diff
@@ -1058,19 +408,32 @@ private JsonData GetPropertyValue(string propertyName)
...
+        {
+            switch (index)
+            {
+                case string propertyName:
+                    return GetPropertyValue(propertyName);
+                case int arrayIndex:
+                    return GetValueAt(arrayIndex);;
+            }
+
+            throw new InvalidOperationException($"Tried to access indexer with unsupport index type: {index}");
```
christothes, line 436:
```suggestion
            throw new InvalidOperationException($"Tried to access indexer with an unsupported index type: {index}");
```

sdk/core/Azure.Core.Experimental/src/JsonData.cs
```diff
@@ -1095,6 +458,19 @@ private List<JsonData> EnsureArray()
...
+        {
+            switch (index)
+            {
+                case string propertyName:
+                    return SetValue(propertyName, value);
+                case int arrayIndex:
+                    return SetValueAt(arrayIndex, value);
+            }
+
+            throw new InvalidOperationException($"Tried to access indexer with unsupport index type: {index}");
```
christothes, line 471:
```suggestion
            throw new InvalidOperationException($"Tried to access indexer with an unsupported index type: {index}");
```

sdk/core/Azure.Core.Experimental/tests/public/Azure.Core.Experimental.Tests.Public.csproj
```diff
@@ -0,0 +1,28 @@
...
+    <PackageReference Include="Microsoft.NET.Test.Sdk" />
+  </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="$(AzureCoreTestFramework)" />
+    <ProjectReference Include="..\..\src\Azure.Core.Experimental.csproj" />
+  </ItemGroup>
+  
+  <ItemGroup>
+    <Compile Remove="perf/**/*.*" />
+  </ItemGroup>
```
christothes, line 20: I don't think this is needed since we are in a peer directory of `perf/`

sdk/core/Azure.Core.Experimental/src/JsonData.cs
annelo-msft: This method validates that the value is in range for a 32-bit int and casts it to Int32, so I feel the name is appropriate.  What did you mean by it not being restricted to Int32?

sdk/core/Azure.Core.Experimental/tests/public/Azure.Core.Experimental.Tests.Public.csproj
annelo-msft: Good catch - thanks!

sdk/core/Azure.Core.Experimental/src/JsonData.cs
christothes: This method thunks down to `var value = EnsureNumberValue().AsLong();` , which I assumed would handle 64 bit int also.

sdk/core/Azure.Core.Experimental/src/JsonData.cs
christothes: I meant the implementation - which is the same as the `GetLong` method. Both seem to return a long, but I don't know how `AsLong` is implemented.

annelo-msft: The original implementation of [Number](https://github.com/Azure/azure-sdk-for-net/pull/32613/files#diff-3b95abc3214f39c93aa150b4b83ca41600e9c97b111d8860284e37e266d5c88fR525-L1202) keeps track of whether the value is an integer or a real by storing a long and a double and a couple of boolean values to indicate which one is used as the backing store for a given instance.  I didn't make any modifications to this logic, because it seemed reasonable to me.  Number.AsLong() returns the `long`, which is returned directly by GetLong(), but is checked to make sure it's in an Int32 range by GetInt32 before it is cast to `int`.  

Do you have any concern about this implementation?

sdk/core/Azure.Core.Experimental/src/JsonData.cs
annelo-msft: After it retrieves the value as a long, it checks that it is in range for a 32-bit int and then casts it to an int.  I believe it would throw OverflowException if the value was between Int32.MaxValue and Int64.MaxValue.

sdk/core/Azure.Core.Experimental/src/JsonData.cs
christothes: Seems fine as long as we get a reasonable exception if the case of an overflow.

christothes (approved):

sdk/core/Azure.Core.Experimental/src/JsonData.cs
ellismg: FYI, @annelo-msft it looks like the "t" got dropped from this method.  I think `GetInt32` is what you intended.

sdk/core/Azure.Core.Experimental/tests/public/JsonDataArrayTests.cs
```diff
@@ -0,0 +1,294 @@
...
+        [Test]
+        public void CanConvertArrayToIEnumerable()
+        {
+            dynamic data = JsonDataTestHelpers.CreateFromJson("[1, 2, 3]");
+
+            // TODO: This is interesting - what is commented out below fails,
+            // because apparently the type system has decided the IEnumerable
+            // is holding JsonData, not dynamic, so the DMO callback is never called
+            // and the explicit cast to int is not visible b/c JsonData is internal.
+            //IEnumerable value = (IEnumerable)data;
```
jsquire, line 23: I believe this may be due to using `IEnumerable` which is non-generic and assumes `object` for the type.   Does using `IEnumerable<dynamic>` as the cast work?

sdk/core/Azure.Core.Experimental/api/Azure.Core.Experimental.netstandard2.0.cs
```diff
@@ -106,91 +106,39 @@ namespace Azure
...
+        public static implicit operator int (Azure.Core.Dynamic.JsonData json) { throw null; }
+        public static implicit operator long (Azure.Core.Dynamic.JsonData json) { throw null; }
+        public static implicit operator bool? (Azure.Core.Dynamic.JsonData json) { throw null; }
+        public static implicit operator double? (Azure.Core.Dynamic.JsonData json) { throw null; }
+        public static implicit operator int? (Azure.Core.Dynamic.JsonData json) { throw null; }
+        public static implicit operator long? (Azure.Core.Dynamic.JsonData json) { throw null; }
+        public static implicit operator float? (Azure.Core.Dynamic.JsonData json) { throw null; }
+        public static implicit operator float (Azure.Core.Dynamic.JsonData json) { throw null; }
+        public static implicit operator string (Azure.Core.Dynamic.JsonData json) { throw null; }
+        public static bool operator !=(Azure.Core.Dynamic.JsonData? left, int right) { throw null; }
```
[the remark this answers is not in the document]
annelo-msft: These will be addressed in a follow-up PR.  My goal was to merge this to make something available early for CloudMachine to start working with.

sdk/core/Azure.Core.Experimental/tests/JsonDataDynamicMutableTests.cs
```diff
@@ -63,9 +63,9 @@ public void NewObjectPropertiesCanBeAssignedWithArrays()
         [Test]
         public void NewObjectPropertiesCanBeAssignedWithObject()
         {
-            var json = JsonData.FromString("{}");
+            var json = JsonData.Parse("{}");
             dynamic jsonData = json;
-            jsonData.a = JsonData.EmptyObject();
+            jsonData.a = JsonData.Parse("{}");
```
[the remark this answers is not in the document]
annelo-msft: This is only for internal test purposes.  It simulates what would be returned by a protocol method response without providing a client.  Per our requirement that customers will modify JsonData instances only for round-trips, I don't think we need a formal API to create an empty instance.  I can wrap the method in the test to make the intent more clear - that the caller would obtain the JsonData instance from a client method.

sdk/core/Azure.Core.Experimental/tests/perf/DynamicReadingBenchmark.cs
```diff
@@ -13,9 +14,10 @@ namespace Azure.Data.AppConfiguration.Performance
     public class DynamicReadingBenchmark
     {
         private static string _json = "{\"a\":{\"b\":5}}";
+        private static BinaryData _binaryData = new BinaryData(_json);
 
         private static dynamic _expandoObject = new ExpandoObject();
-        private static dynamic _jsonData = JsonData.FromString(_json);
+        private static dynamic _jsonData = _binaryData.ToDynamic();
```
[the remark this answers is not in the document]
annelo-msft: Cool, that makes sense.  We could have it be ToDynamicJson() ... although that makes me wonder if we should change the typename to DynamicJson rather than JsonData ... is there a strong need for Data to be in the name?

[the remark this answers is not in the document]
KrzysztofCwalina: BinaryData has a method `FromObjectAsJson`, so maybe we call this one `ToDynamicFromJson`, but such name is on the longish side. I would prefer something shorter. We could also take a parameter that selects between formats that we support. For now, it would be JSON but in the future we could support XML, protobuff, etc. 

I don't think calling the type DynamicJson would help. This type name is not visible when `ToDynamic` is called. 

Lastly, that would close the door to reusing this type for other formats besides JSON.

sdk/core/Azure.Core.Experimental/tests/public/JsonDataArrayTests.cs
annelo-msft: It does not ... it fails with a RuntimeBinderException whereas just IEnumerable fails with an InvalidCastException.

sdk/core/Azure.Core.Experimental/tests/perf/DynamicReadingBenchmark.cs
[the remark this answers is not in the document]
annelo-msft: Per offline conversation, we will leave this as-is for now.

weshaggard:
@annelo-msft looks like this change also is the reason some of the purview tests are failing see https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2055180&view=logs&j=19ae9ca6-cc95-59cd-da1c-d5a57cd99bf5&t=d4ee6916-5fd3-538d-4187-7e8f2bed8e1c&l=49. It looks like it is caused by the move of JsonData to a new namespace.

PR #32615, opened 2022-11-22
[Text Analytics] Add support for automatic language detection

### LINKED ISSUE #32193: [Text Analytics] Add support for automatic language detection

# Tasks
- [ ] Implementation
- [ ] Tests
- [ ] Docs

joseharriaga:
Resolves: #32193

Features added:
- Added the `AbstractSummaryResult.DetectedLanguage` property.
- Added the `AnalyzeHealthcareEntitiesResult.DetectedLanguage` property.
- Added the `AnalyzeSentimentResult.DetectedLanguage` property.
- Added the `ClassifyDocumentResult.DetectedLanguage` property.
- Added the `ExtractKeyPhrasesResult.DetectedLanguage` property.
- Added the `ExtractSummaryResult.DetectedLanguage` property.
- Added the `RecognizeEntitiesResult.DetectedLanguage` property.
- Added the `RecognizeLinkedEntitiesResult.DetectedLanguage` property.
- Added the `RecognizePiiEntitiesResult.DetectedLanguage` property.


sdk/textanalytics/Azure.AI.TextAnalytics/CHANGELOG.md
```diff
@@ -17,6 +17,15 @@
...
 - Added the following enums: `AgeUnit`, `AreaUnit`, `DateTimeSubKind`, `InformationUnit`,`LengthUnit`, `NumberKind`, `RangeKind`, `RelativeTo`, `SpeedUnit`, `TemperatureUnit`,`TemporalModifier`, `VolumeUnit`, and `WeightUnit`.
+- Added the `AbstractSummaryResult.DetectedLanguage` property.
+- Added the `AnalyzeHealthcareEntitiesResult.DetectedLanguage` property.
+- Added the `AnalyzeSentimentResult.DetectedLanguage` property.
+- Added the `ClassifyDocumentResult.DetectedLanguage` property.
+- Added the `ExtractKeyPhrasesResult.DetectedLanguage` property.
+- Added the `ExtractSummaryResult.DetectedLanguage` property.
+- Added the `RecognizeEntitiesResult.DetectedLanguage` property.
+- Added the `RecognizeLinkedEntitiesResult.DetectedLanguage` property.
+- Added the `RecognizePiiEntitiesResult.DetectedLanguage` property.
```
kristapratico, line 28: should RecognizeCustomEntitiesResult be included?

sdk/textanalytics/Azure.AI.TextAnalytics/src/AnalyzeHealthcareEntitiesResult.cs
```diff
@@ -47,6 +53,12 @@ public partial class AnalyzeHealthcareEntitiesResult : TextAnalyticsResult
         /// </summary>
         public IReadOnlyCollection<TextAnalyticsWarning> Warnings { get; } = new List<TextAnalyticsWarning>();
 
+        /// <summary>
+        /// The language of the input document as detected by the service when requested to perform automatic language
+        /// detection, which is possible by specifying "auto" as the language of the input document.
+        /// </summary>
+        public DetectedLanguage? DetectedLanguage { get; }
```
kristapratico, line 60: Should this be string for healthcare to match how the service returns it for this api version?

sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.cs
```diff
@@ -145,9 +159,23 @@ internal static ExtractKeyPhrasesResultCollection ConvertToExtractKeyPhrasesResu
...
-            {
-                keyPhrases.Add(new ExtractKeyPhrasesResult(docKeyPhrases.Id, docKeyPhrases.Statistics ?? default, ConvertToKeyPhraseCollection(docKeyPhrases)));
+            foreach (KeyPhraseResultDocumentsItem document in results.Documents)
+            {
+                DetectedLanguage? detectedLanguage = (document.DetectedLanguage is not null)
+                    ? new DetectedLanguage(
+                        document.DetectedLanguage.Value.Name,
+                        document.DetectedLanguage.Value.Iso6391Name,
+                        document.DetectedLanguage.Value.ConfidenceScore,
+                        document.DetectedLanguage.Value.Script,
```
kristapratico, line 169: just curious, if someone sets the api version to an older version in which script is not supported/sent by the service, will this code set it to null?

sdk/textanalytics/Azure.AI.TextAnalytics/CHANGELOG.md
joseharriaga: RecognizeCustomEntities returns a collection of RecognizeEntitiesResult, so there's no RecognizeCustomEntitiesResult class.

sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.cs
joseharriaga: My understanding is that both automatic language detection and script detection are both new features in version 2022-10-01-preview, which means that if a version older than 2022-10-01-preview is used, the request would fail if "auto" is passed in the request (or at a minimum, detectedLanguage would be null in the response). Is this correct?

kristapratico: I haven't tested "auto" in an older version yet so not sure what the service behavior is. In python, if an older version is selected, the generated model received by the transform code won't have a detectedLanguage property and so we set it to null/None on the exposed model. I guess I'm wondering in .NET's case if you would still create the DetectedLanguage model, but just set each of its properties to null.

sdk/textanalytics/Azure.AI.TextAnalytics/src/AnalyzeHealthcareEntitiesResult.cs
joseharriaga: Yeah, you're probably right: It might be less confusing if it's simply a string for now. I changed it.

sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.Legacy.cs
```diff
@@ -674,24 +699,26 @@ internal static List<HealthcareEntity> ConvertToHealthcareEntityCollection(IRead
...
-                    ConvertToWarnings(documentHealthcareEntities.Warnings),
-                    default));
+                healthcareEntititesResults.Add(
+                    new AnalyzeHealthcareEntitiesResult(
+                        documentHealthcareEntities.Id,
+                        ConvertToDocumentStatistics(documentHealthcareEntities.Statistics),
+                        ConvertToHealthcareEntityCollection(documentHealthcareEntities.Entities),
+                        ConvertToHealthcareEntityRelationsCollection(documentHealthcareEntities.Entities, documentHealthcareEntities.Relations),
+                        default,
+                        default,
```
mssfang, line 720: what are the `default` means in .NET?

sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.cs
```diff
@@ -199,16 +241,30 @@ internal static RecognizeCustomEntitiesResultCollection ConvertToRecognizeCustom
...
             {
-                recognizeEntities.Add(new RecognizeEntitiesResult(docEntities.Id, docEntities.Statistics ?? default, ConvertToCategorizedEntityCollection(docEntities)));
+                DetectedLanguage? detectedLanguage = (document.DetectedLanguage is not null)
+                    ? new DetectedLanguage(
+                        document.DetectedLanguage.Value.Name,
+                        document.DetectedLanguage.Value.Iso6391Name,
+                        document.DetectedLanguage.Value.ConfidenceScore,
+                        document.DetectedLanguage.Value.Script,
+                        default)
+                    : null;
```
mssfang, line 260: Saw this section of code being used multiple places, should we create an utility method which can be reused?

sdk/textanalytics/Azure.AI.TextAnalytics/tests/RecognizeCustomEntitiesTests.cs
```diff
@@ -46,7 +46,7 @@ public RecognizeCustomEntitiesTests(bool isAsync, TextAnalyticsClientOptions.Ser
             "petróleo",
         };
 
-        private static readonly List<string> e_batchConvenienceDocuments = new List<string>
+        private static readonly List<string> s_englishBatchConvenienceDocuments = new List<string>
```
mssfang, line 49: is "e_" stands for 'English' and "s_" prefix stands for 'Spanish'?

joseharriaga: I think that was the original intention, but "e_" is not a valid prefix per C# standards. On the other hand, "s_" is used to prefix static classes.

kristapratico (approved):

mssfang (approved):

PR #32661, opened 2022-11-24
feat(dataplane): unify abstract base class

archerzz:
Right now, dataplane does not default to abstract for base classes of polymorphic sets. We'll apply the same logic for mgmt plane to data plane in https://github.com/Azure/autorest.csharp/pull/2878

For backward compatibility, we use `suppress-abstract-base-class` directive to keep the existing behavior for officially released data plane SDKs.


heaths (approved):
Text Analytics and Video Analyzer LGTM

ShivangiReja (approved):
Search LGTM

PR #32749, opened 2022-11-29
introduce the modelfactory for resourcemanager

ArcturusZhang:


sdk/resourcemanager/Azure.ResourceManager/src/Resources/Custom/ResourceManagerModelFactory.cs
```diff
@@ -0,0 +1,27 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Azure.Core;
+
+namespace Azure.ResourceManager.Resources.Models
```
m-nash, line 6: For the common library I think we want all models that live in this library to go in a single place of `Azure.ResourceManager.Models` to avoid any namespace conflicts with other libraries model factories

ArcturusZhang: make sense. Will change this.

m-nash (approved):

Issue #32418, opened 2022-11-10
[ACR] Document upload/download artifacts samples


lmolkova:
### Library name

Azure.Containers.ContainerRegistry

### Please describe the feature.

- [ ]  Add sample code
- [ ]  Update Readme
- [ ] Nice-to-have: Live tests

annelo-msft:
Per @lmolkova, the plan is that we will discuss these across language at the next ACR stand-up.

PR #32758, opened 2022-11-29
Add Upload and Download Samples to ACR README and samples collection

annelo-msft:
Add samples showing how to upload and download OCI Artifacts to/from an Azure Container Registry instance.

Addresses https://github.com/Azure/azure-sdk-for-net/issues/32418

See sample provided for Java in https://github.com/Azure/azure-sdk-for-java/issues/31598.


eng/Packages.Data.props
```diff
@@ -247,7 +247,7 @@
     <PackageReference Update="Newtonsoft.Json" Version="10.0.3" />
     <PackageReference Update="NSubstitute" Version="3.1.0" />
     <PackageReference Update="NUnit" Version="3.13.2" />
-    <PackageReference Update="NUnit3TestAdapter" Version="4.3.0" />
+    <PackageReference Update="NUnit3TestAdapter" Version="4.3.1" />
```
annelo-msft, line 250: This will merge in a separate PR before this PR goes in.

sdk/containerregistry/Azure.Containers.ContainerRegistry/samples/Sample05_UploadArtifact.md
```diff
@@ -0,0 +1,58 @@
...
+
+The following sample illustrates how to upload an OCI Artifact to an Azure Container Registry instance.
+
+## Overview
+
+An OCI artifact consists of a manifest file, config file, and one or more layer files.  To upload an image, you will upload each of these files, ending with the manifest file.  For more details regarding the OCI artifact format, please see the [OCI Image Format Specification](https://github.com/opencontainers/image-spec/blob/main/spec.md).
+
+## Upload Image
+
+Create a `ContainerRegistryBlobClient`.
```
annelo-msft, line 11: TODO: Update the text to make it useful.

sdk/containerregistry/Azure.Containers.ContainerRegistry/samples/Sample05_UploadArtifact.md
```diff
@@ -0,0 +1,58 @@
...
+// Create a new ContainerRegistryBlobClient
+ContainerRegistryBlobClient client = new ContainerRegistryBlobClient(endpoint, repository, new DefaultAzureCredential(), new ContainerRegistryClientOptions()
+{
+    Audience = ContainerRegistryAudience.AzureResourceManagerPublicCloud
+});
+
+// Create a manifest to list files in this artifact
+OciManifest manifest = new()
+{
+    SchemaVersion = 2
```
annelo-msft, line 29: TODO: Can we default this for callers so they don't have to set a magic number?

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/autorest.md
```diff
@@ -158,3 +158,47 @@ directive:
...
     delete $["x-accessibility"]
 ```
+
+# Add OciMediaType extensible enum
+``` yaml
+directive:
+  from: swagger-document
+  where: $.definitions
+  transform: >
+    $["OciMediaType"] = {
```
ShivangiReja, line 168: Why are we adding this type via configuration for .NET SDK only? shouldn't we update the swagger if it's a common type for all the languages?

ShivangiReja (approved):

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/autorest.md
YalinLi0312: +1 Should Python follow the same settings here?

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/OciManifest.cs
```diff
@@ -8,5 +8,11 @@ namespace Azure.Containers.ContainerRegistry.Specialized
     [CodeGenModel("OCIManifest")]
     public partial class OciManifest
     {
+        /// <summary> Initializes a new instance of OciManifest. </summary>
+        public OciManifest()
+        {
+            Layers = new ChangeTrackingList<OciBlobDescriptor>();
+            SchemaVersion = 2;
```
annelo-msft, line 15: @lmolkova, I have moved setting the default value to the required value into the constructor so callers don't have to set it themselves, per the OCI spec: https://github.com/opencontainers/image-spec/blob/main/manifest.md#image-manifest-property-descriptions

Let me know if you have any concerns with this - if you agree, we can make sure all languages implement it this way.

sdk/containerregistry/Azure.Containers.ContainerRegistry/samples/Sample05_UploadArtifact.md
annelo-msft: Moved this into type constructor.

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/autorest.md
annelo-msft: @YalinLi0312, let's discuss this in our next stand-up, whether we want this at all, and if so, if we should put it to a common place (per @ShivangiReja's good comment)

PR #32770, opened 2022-11-30
Support additional types for DataFactoryExpression

JoshLove-msft:


sdk/core/Azure.Core.Expressions.DataFactory/src/DataFactoryExpression.cs
```diff
@@ -6,14 +6,17 @@
...
 
 namespace Azure.Core.Expressions.DataFactory
 {
     /// <summary>
     /// A class representing either a primitive value or an expression.
     /// For details on DataFactoryExpressions see https://learn.microsoft.com/en-us/azure/data-factory/control-flow-expression-language-functions#expressions.
     /// </summary>
-    /// <typeparam name="T"> Can be one of <see cref="string"/>, <see cref="bool"/>, <see cref="int"/>, <see cref="double"/>, <see cref="Array"/>, or <see cref="IList{TElement}"/>.</typeparam>
+    /// <typeparam name="T"> Can be one of <see cref="string"/>, <see cref="bool"/>, <see cref="int"/>, <see cref="double"/>, <see cref="TimeSpan"/>,
+    /// <see cref="DateTimeOffset"/>, <see cref="Uri"/>, <see cref="IList{String}"/>, <see cref="IList{TElement}"/>,
```
christothes, line 18: Is there something special about `IList<string>` that isn't covered by `IList<TElement>`?

sdk/core/Azure.Core.Expressions.DataFactory/src/DataFactoryExpression.cs
```diff
@@ -63,16 +66,30 @@ internal DataFactoryExpression(string expression, string type)
...
+                if (_literal == null)
+                    return null;
+
+                if (typeof(T).IsGenericType && typeof(T).GetGenericTypeDefinition() == typeof(IList<>))
                 {
-                    return $"[{string.Join(",", literalArray.OfType<object>().Select(item => item?.ToString()))}]";
+                    var methodInfo = typeof(DataFactoryExpression<>).MakeGenericType(typeof(IList<>)).GetGenericTypeDefinition().MakeGenericType(typeof(T).GenericTypeArguments[0])
+                        .GetMethod(nameof(ToStringList), BindingFlags.Static | BindingFlags.NonPublic)!
+                        .MakeGenericMethod(typeof(T).GenericTypeArguments[0]);
+                    return (string?) methodInfo!.Invoke(null, new object?[] {_literal})!;
```
christothes, line 77: Is it worth caching this `methodInfo`?

christothes (approved):

sdk/core/Azure.Core.Expressions.DataFactory/src/DataFactoryExpression.cs
JoshLove-msft: Yeah TElement is actually for model types that have a JsonConverter defined.

sdk/core/Azure.Core.Expressions.DataFactory/src/DataFactoryExpression.cs
JoshLove-msft: Just for readability.

sdk/core/Azure.Core.Expressions.DataFactory/src/DataFactoryExpression.cs
JoshLove-msft: Will update.

m-nash: same with IDctionary<string, string>

sdk/core/Azure.Core.Expressions.DataFactory/src/DataFactoryExpression.cs
```diff
@@ -6,14 +6,17 @@
...
     /// A class representing either a primitive value or an expression.
     /// For details on DataFactoryExpressions see https://learn.microsoft.com/en-us/azure/data-factory/control-flow-expression-language-functions#expressions.
     /// </summary>
-    /// <typeparam name="T"> Can be one of <see cref="string"/>, <see cref="bool"/>, <see cref="int"/>, <see cref="double"/>, <see cref="Array"/>, or <see cref="IList{TElement}"/>.</typeparam>
+    /// <typeparam name="T"> Can be one of <see cref="string"/>, <see cref="bool"/>, <see cref="int"/>, <see cref="double"/>, <see cref="TimeSpan"/>,
+    /// <see cref="DateTimeOffset"/>, <see cref="Uri"/>, <see cref="IList{String}"/>, <see cref="IList{TElement}"/>,
+    /// or <see cref="IDictionary{String,String}"/>.</typeparam>
 #pragma warning disable SA1649 // File name should match first type name
     [JsonConverter(typeof(DataFactoryExpressionJsonConverter))]
     public sealed class DataFactoryExpression<T>
```
m-nash, line 22: Have we thought about mockability here?

I am ok if that is a follow up feature but if we choose to do that can we create an issue to track?

sdk/core/Azure.Core.Expressions.DataFactory/src/DataFactoryExpression.cs
m-nash: I wonder if we should use our own serialization here instead?

sdk/core/Azure.Core.Expressions.DataFactory/src/DataFactoryExpression.cs
JoshLove-msft: > nit: should it be `IList{string}` to match the cref to `string`?
> 
> Dotnet has always confused me on `String` vs `string`

This doesn't compile, for some reason it wants it to be String.

sdk/core/Azure.Core.Expressions.DataFactory/src/DataFactoryExpression.cs
JoshLove-msft: Do you mean get a string representation of the serialized JSON? I'm not sure we want to tie it that closely to a serialization format.

sdk/core/Azure.Core.Expressions.DataFactory/src/DataFactoryExpression.cs
JoshLove-msft: Ah I didn't even notice that this was sealed. I think it should be simple enough to actually create instances of this type for testing though rather than relying on mocks.

sdk/core/Azure.Core.Expressions.DataFactory/src/DataFactoryExpression.cs
m-nash: Aren't we re-inventing a different way to do the same thing here?  Id prefer to avoid the reflection if possible here.

m-nash (approved):
Lets put in the change for the ToString other than that this is ready to go.

Issue #32792, opened 2022-12-01
Key Vault samples should be rewritten as recorded tests


heaths:
Now that we have the `SNIPPET` symbol on which to condition snippets vs. running samples, we should rewrite the Key Vault samples - at least Keys - using the `RecordedTestBase`. This is because Managed HSM tests - which also run the Keys packages tests - can occasionally fail because a backup/restore job is running and there's no effective way to make sure those run serially: tests are `[NonParallelizable]`, but assemblies run in parallel because that's how we run them upstream via _eng/service.proj_.

To best take advantage of the `[IgnoreServiceError]` attribute we need to use recorded tests. It's not worth porting to support non-recorded tests, probably, since doing recorded tests may give us a little more coverage anyway. At the very least it'll better test PRs.

The `[IgnoreServiceError]` assembly can and should be put on the `SampleFixture` base class so any keys method that fails with those specific-as-possible errors are ignored (or marked inconclusive).

pallavit:
I think this issue is similar to issue #28953 and something that we can address as part of the test initiative we are taking with KV

heaths:
This is separate. The samples aren't recorded and would need some fundamental rewrites. It's not expensive, but just hasn't been priority. It'll take some work, though, as they predate most of our contemporary samples procedures.

PR #32793, opened 2022-12-01
Ignore some occasional KV service errors

heaths:
Related to #32792, which is a more extensive fix for Keys samples.

### DIFF
--- a/sdk/keyvault/Azure.Security.KeyVault.Certificates/tests/CertificateClientLiveTests.cs
+++ b/sdk/keyvault/Azure.Security.KeyVault.Certificates/tests/CertificateClientLiveTests.cs
@@ -934,7 +934,15 @@ namespace Azure.Security.KeyVault.Certificates.Tests
             CertificateOperation operation = await Client.StartCreateCertificateAsync(name, policy);
             RegisterForCleanup(name);
 
-            await operation.WaitForCompletionAsync(DefaultCertificateOperationPollingInterval, default);
+            try
+            {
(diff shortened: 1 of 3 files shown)


sdk/keyvault/Azure.Security.KeyVault.Certificates/tests/CertificateClientLiveTests.cs
```diff
@@ -934,7 +934,15 @@ public async Task DownloadECDsaCertificateSignRemoteVerifyLocal([EnumValues] Cer
             CertificateOperation operation = await Client.StartCreateCertificateAsync(name, policy);
             RegisterForCleanup(name);
 
-            await operation.WaitForCompletionAsync(DefaultCertificateOperationPollingInterval, default);
+            try
+            {
+                await operation.WaitForCompletionAsync(DefaultCertificateOperationPollingInterval, default);
+            }
+            catch (InvalidOperationException ex) when (ex.Message.Contains("An internal error has occurred"))
```
JoshLove-msft, line 941: This is probably due to an issue with the Test Proxy. Can we file an issue with repro instructions in the tools repo?

heaths: As discussed offline, this may not be a test proxy issue but I'll consider adding an inner exception to what's thrown so we have more details.

JoshLove-msft (approved):

PR #32800, opened 2022-12-01
Apply for remove duplicate description

pshao25:
Apply for https://github.com/Azure/autorest.csharp/pull/2903

### DIFF
--- a/sdk/confidentialledger/Azure.Security.ConfidentialLedger/src/Generated/ConfidentialLedgerCertificateClient.cs
+++ b/sdk/confidentialledger/Azure.Security.ConfidentialLedger/src/Generated/ConfidentialLedgerCertificateClient.cs
@@ -14,7 +14,7 @@ using Azure.Security.ConfidentialLedger;
 
 namespace Azure.Security.ConfidentialLedger.Certificate
 {
-    // Data plane generated client. The ConfidentialLedgerCertificate service client.
+    // Data plane generated client.
     /// <summary> The ConfidentialLedgerCertificate service client. </summary>
     public partial class ConfidentialLedgerCertificateClient
     {
--- a/sdk/agrifood/Azure.Verticals.AgriFood.Farming/src/Generated/ApplicationDataClient.cs
+++ b/sdk/agrifood/Azure.Verticals.AgriFood.Farming/src/Generated/ApplicationDataClient.cs
@@ -16,7 +16,7 @@ using Azure.Core.Pipeline;
 
 namespace Azure.Verticals.AgriFood.Farming
 {
-    // Data plane generated client. The ApplicationData service client.
+    // Data plane generated client.
     /// <summary> The ApplicationData service client. </summary>
     public partial class ApplicationDataClient
     {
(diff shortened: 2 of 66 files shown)


sdk/confidentialledger/Azure.Security.ConfidentialLedger/src/Generated/ConfidentialLedgerCertificateClient.cs
```diff
@@ -14,7 +14,7 @@
 
 namespace Azure.Security.ConfidentialLedger.Certificate
 {
-    // Data plane generated client. The ConfidentialLedgerCertificate service client.
+    // Data plane generated client.
```
christothes, line 17: Do we need this comment and the XML comments? Usually, we only apply the XML comments to public types.

annelo-msft: Having this as a search string allows us to identify DPG clients vs. HLC clients.

annelo-msft: @pshao25, if the DPG search-string comment isn't part of public documentation, it might be worth making it look more automated to make it clear it's just for search purposes.  Or, taking the opposite approach, and making it useful/nice enough to find a way to include it in the public XML comments.  It does look a bit out of place here right before the public XML comment.

heaths (approved):
CLU, QA, and the template LGTM

ShivangiReja (approved):
AppConfig, DevCenter changes LGTM!

sdk/confidentialledger/Azure.Security.ConfidentialLedger/src/Generated/ConfidentialLedgerCertificateClient.cs
pshao25: How about:
```
// Search purpose: Data plane generated client.
```

archerzz: We can follow the convention [here](https://github.com/Azure/azure-sdk-for-net/blob/ecd180864503e78a65e10dcb5482cdf57c74e9c3/sdk/advisor/Azure.ResourceManager.Advisor/src/Generated/Extensions/AdvisorExtensions.cs#L4).

> // <data-plane-generated-client/>

sebrenna (approved):

vicancy (approved):

sdk/confidentialledger/Azure.Security.ConfidentialLedger/src/Generated/ConfidentialLedgerCertificateClient.cs
m-nash: Lets keep the string as is for now.  I think we will want to make this accessible via some other mean going forward as its just one of many attributes people might want to search for, but until we have some more clarity around that I don't want to break existing use cases.

pshao25:
This change is included in the main branch. So close it.

Issue #32722, opened 2022-11-28
[Core] Allow disabling auto redirect on a message


annelo-msft:
### API Change Needed

Yes

### Background and motivation

Per the below requests for the ACR library, customers need the ability to obtain the Location header from the service's redirect response.  However, because the caller wishes to download the blob from the URI provided in the Location header, we also need the ability to tell Core to disable auto-redirects, so that the request to initiate blob download will not be sent multiple times.

https://github.com/Azure/azure-sdk-for-net/issues/28735
https://github.com/Azure/azure-sdk-for-net/issues/28871

### API / Feature Proposal

```C#
namespace Azure.Core.Pipeline
{
    public class RedirectPolicy
    {
        public static void AllowAutoRedirect(HttpMessage message, bool allowAutoRedirect);
    }
}
```

### Time Constraints and Dependencies

Target the January or February Core release to enable the ACR feature request along the same timeline.

annelo-msft:
Fixed via https://github.com/Azure/azure-sdk-for-net/pull/32810

PR #32810, opened 2022-12-01
Add the ability to disable automatically following redirects

annelo-msft:
Addresses https://github.com/Azure/azure-sdk-for-net/issues/32722


sdk/core/Azure.Core/src/Pipeline/Internal/RedirectPolicy.cs
```diff
@@ -8,17 +8,31 @@
 
 namespace Azure.Core.Pipeline
 {
-    internal sealed class RedirectPolicy : HttpPipelinePolicy
+    /// <summary>
+    /// A pipeline policy that detects a redirect response code and resends the request to the
+    /// location specified by the response.
+    /// </summary>
+    public sealed class RedirectPolicy : HttpPipelinePolicy
```
annelo-msft, line 15: If we decide this is the approach we want to take in the Core API, I will move this file out of the `Internal` folder.  I wanted to leave it here for a clean diff in the initial PR.

sdk/core/Azure.Core/src/Pipeline/Internal/RedirectPolicy.cs
```diff
@@ -36,7 +50,9 @@ internal async ValueTask ProcessAsync(HttpMessage message, ReadOnlyMemory<HttpPi
             Request request = message.Request;
             Response response = message.Response;
 
-            while ((redirectUri = GetUriForRedirect(request, message.Response)) != null)
+            bool allowAutoRedirect = GetAllowAutoRedirectValue(message);
+
+            while (allowAutoRedirect && (redirectUri = GetUriForRedirect(request, message.Response)) != null)
```
JoshLove-msft, line 55:
```suggestion
            if (!allowAutoRedirect) return;
            while ((redirectUri = GetUriForRedirect(request, message.Response)) != null)
```

sdk/core/Azure.Core/src/Pipeline/Internal/RedirectPolicy.cs
```diff
@@ -8,17 +8,31 @@
...
         {
             _maxAutomaticRedirections = 50;
         }
 
+        /// <summary>
+        /// Sets a value that indicates whether redirects will be automatically followed for this message.
+        /// </summary>
+        /// <param name="message"></param>
+        /// <param name="allowAutoRedirect"></param>
+        public static void AllowAutoRedirect(HttpMessage message, bool allowAutoRedirect)
```
JoshLove-msft, line 31: Should this be called something like `SetAllowAutoRedirect` or `SetAutoRedirect`?

annelo-msft: I considered that ... I was worried that the former was wordy, and the latter didn't mirror the .NET API naming, but you're right, it should have a Set prefix.  I'll update it, and thanks!

sdk/core/Azure.Core/src/Pipeline/Internal/RedirectPolicy.cs
```diff
@@ -164,11 +180,20 @@ private static bool RequestRequiresForceGet(int statusCode, RequestMethod reques
         internal static bool IsSecureWebSocketScheme(string scheme) =>
             string.Equals(scheme, "wss", StringComparison.OrdinalIgnoreCase);
 
+        /// <summary>
```
JoshLove-msft, line 183: nit: either fill out or inheritdoc

JoshLove-msft (approved):

sdk/containerregistry/Azure.Containers.ContainerRegistry/api/Azure.Containers.ContainerRegistry.netstandard2.0.cs
```diff
@@ -138,6 +138,7 @@ public partial class ContainerRegistryClient
         public ContainerRegistryClient(System.Uri endpoint, Azure.Core.TokenCredential credential) { }
         public ContainerRegistryClient(System.Uri endpoint, Azure.Core.TokenCredential credential, Azure.Containers.ContainerRegistry.ContainerRegistryClientOptions options) { }
         public virtual System.Uri Endpoint { get { throw null; } }
+        public virtual Azure.Core.Pipeline.HttpPipeline Pipeline { get { throw null; } }
```
christothes, line 141: Could you provide some context on why this needs to be public? If this behavior for a specific client call is intended to be exposed to customers, I'd expect it to be exposed via some client options or parameter on the specific methods that require it. If for the client itself, I'd expect it to be internal.

annelo-msft: Sure, there are a couple of reasons.  

The first is that our team is moving toward a strategy of giving lower-level access to library users by exposing the Pipeline as a public member on the client.  This is generated automatically in DPG clients, so this is following that precedent, just implementing it manually in this case.

The second is that this scenario is sufficiently uncommon, and not fully documented by the service team, that we don't want to expose it in a supported API on a GA client, but we do want to make it available to callers who understand the details and are willing to use an advanced API to accomplish it.

sdk/core/Azure.Core/src/Pipeline/Internal/RedirectPolicy.cs
annelo-msft: This is completed.

sdk/containerregistry/Azure.Containers.ContainerRegistry/api/Azure.Containers.ContainerRegistry.netstandard2.0.cs
annelo-msft: Discussed offline.

PR #32822, opened 2022-12-01
Regen to get diff

JoshLove-msft:


sdk/anomalydetector/Azure.AI.AnomalyDetector/TempCadlFiles/cadl/cadl-project.yaml
```diff
@@ -0,0 +1,17 @@
+emitters:
```
m-nash, line 1: These files should not be here.  I wonder if there is a path mismatch when using service.proj vs pointing at a project directory?  We will need to investigate this.

sdk/search/Azure.Search.Documents/src/Generated/Models/CognitiveServicesAccount.cs
```diff
@@ -12,7 +12,7 @@ namespace Azure.Search.Documents.Indexes.Models
     /// Please note <see cref="CognitiveServicesAccount"/> is the base class. According to the scenario, a derived class of the base class might need to be assigned here, or this property needs to be casted to one of the possible derived classes.
     /// The available derived classes include <see cref="CognitiveServicesAccountKey"/> and <see cref="DefaultCognitiveServicesAccount"/>.
     /// </summary>
-    public partial class CognitiveServicesAccount
+    public abstract partial class CognitiveServicesAccount
```
m-nash, line 15: this is breaking if its GAd

m-nash (approved):

archerzz (approved):

Issue #32046, opened 2022-10-26
[Core] Regional Fallback


ronniegeraghty:
### Library name

Azure.Core

### Please describe the feature.

Add regional fallback to azure core.

jsquire:
To my knowledge, this is something that the archboard determined is not feasible for a global approach in Core.   Closing this out.

PR #32827, opened 2022-12-02
Prototype GeoRedundantPolicy

JoshLove-msft:
Addresses https://github.com/Azure/azure-sdk-for-net/issues/32046


sdk/core/Azure.Core/src/Shared/GeoRedundantReadFallbackPolicy.cs
```diff
@@ -0,0 +1,100 @@
...
+
+                // subsequent attempt
+
+                // only fallback if no response was received
+                if (message.HasResponse)
+                    return;
+
+                // advance the index
+                _fallbackIndex ??= -1;
+                _fallbackIndex++;
```
christothes, line 82: I think we need to check that the current host matches the one pointed to by `_fallbackIndex` before incrementing. Consider the following scenario:

Thread 1: 
 - traverse up policy stack to the transport
 - send request to primary and fail

Thread 2: 
 - traverse up policy stack to the transport
 - send request to primary and fail

Thread 1:
- Unroll the policy stack to `GeoRedundantReadFallbackPolicy`
- acquire lock 
- increment `_fallbackIndex` (points to secondary)
- release lock
- request send to secondary

Thread 2:
- Unroll the policy stack to `GeoRedundantReadFallbackPolicy`
- acquire lock 
- increment `_fallbackIndex` (points to primary, since it's already pointed at the last index)
- release lock
- request sent to primary again

sdk/core/Azure.Core/src/Shared/GeoRedundantReadFallbackPolicy.cs
```diff
@@ -0,0 +1,100 @@
...
+
+                // reset to null to indicate primary should be used when we reach the end
+                if (_fallbackIndex >= _readFallbackHosts.Length)
+                    _fallbackIndex = null;
```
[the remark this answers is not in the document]
JoshLove-msft: We wouldn't need to change it in that case because it is either already set to primary, or it is set to a fallback but hostAffinity is true.

sdk/core/Azure.Core/src/Shared/GeoRedundantReadFallbackPolicy.cs
```diff
@@ -0,0 +1,100 @@
...
+                //
+                // var secondMessage = _pipeline.CreateMessage();
+                // secondMessage.Request.Uri.Host = firstMessage.Request.Uri.Host;
+                // GeoRedundantReadFallbackPolicy.SetHostAffinity(secondMessage, true);
+                // secondMessage.ProcessingContext.HostAffinity = true;
+                // ...
+                // _pipeline.Send(secondMessage);
+
+                if (GetHostAffinity(message))
+                    return;
```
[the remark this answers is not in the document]
JoshLove-msft: Yeah definitely. Another common case that should be handled outside the lock is for the first attempt. We can check the index outside the lock and then check it again in the lock if it was set.

sdk/core/Azure.Core/src/Shared/GeoRedundantReadFallbackPolicy.cs
```diff
@@ -0,0 +1,134 @@
...
+                SetPrimaryHost(message);
+                // avoid locking for common path - there is a chance that the index would be incremented here by another thread but we prioritize
+                // throughput over correctness in this case
+                if (_fallbackIndex == null)
+                    return;
+            }
+
+            // must be under lock so that this can be synchronized across client calls - each call should leverage the fallback information
+            // from all other calls
+            lock (_syncLock)
```
KrzysztofCwalina, line 71: Do we need to lock, or we could use Interlocked?

sdk/core/Azure.Core/src/Shared/GeoRedundantReadFallbackPolicy.cs
```diff
@@ -0,0 +1,134 @@
...
+
+            if (message.HasResponse || GetHostAffinity(message))
+                return;
+
+            if (message.ProcessingContext.RetryNumber == 0)
+            {
+                SetPrimaryHost(message);
+                // avoid locking for common path - there is a chance that the index would be incremented here by another thread but we prioritize
+                // throughput over correctness in this case
+                if (_fallbackIndex == null)
```
KrzysztofCwalina, line 65: I think you need to read the filed under the lock, which brings me to the point below, i.e. can/should we use Interlocek as this read will be executed on every send.

sdk/core/Azure.Core/src/Shared/GeoRedundantReadFallbackPolicy.cs
```diff
@@ -0,0 +1,134 @@
...
+
+                if (!shouldAdvance)
+                    return;
+
+                // advance the index
+                _fallbackIndex ??= -1;
+                _fallbackIndex++;
+
+                // reset to null to indicate primary should be used when we reach the end
+                if (_fallbackIndex >= _readFallbackHosts.Length)
```
KrzysztofCwalina, line 107: Don't we want to fall back to the primary after some time, not only when the secondary fails.

sdk/core/Azure.Core/src/Shared/GeoRedundantReadFallbackPolicy.cs
```diff
@@ -0,0 +1,134 @@
...
+                if (message.ProcessingContext.RetryNumber == 0)
+                {
+                    UpdateHost(message);
+                    return;
+                }
+
+                // subsequent attempt
+
+                // we should only advance if another thread hasn't already done so
+                bool shouldAdvance = (_fallbackIndex == null && message.Request.Uri.Host == GetPrimaryHost(message)) ||
```
KrzysztofCwalina, line 96: Can/should we do these comparisons using ordinal comparison?

sdk/core/Azure.Core/src/Shared/GeoRedundantReadFallbackPolicy.cs
JoshLove-msft: As mentioned in the comment, this is intentionally outside of the lock to prioritize perf.

sdk/core/Azure.Core/src/Shared/GeoRedundantReadFallbackPolicy.cs
JoshLove-msft: I think we need some form of lock to handle Increment index + update host to use that index. Using interlocked can allow us to increment index atomically, but it might get incremented again before updating the host.

sdk/core/Azure.Core/src/Shared/GeoRedundantReadFallbackPolicy.cs
JoshLove-msft: Yes! I'm working on that now.

sdk/core/Azure.Core/src/Shared/GeoRedundantReadFallbackPolicy.cs
```diff
@@ -0,0 +1,134 @@
...
+
+namespace Azure.Core.Shared
+{
+    /// <summary>
+    ///
+    /// </summary>
+    internal class GeoRedundantReadFallbackPolicy : HttpPipelineSynchronousPolicy
+    {
+        private readonly string[] _readFallbackHosts;
+        private int? _fallbackIndex;
```
KrzysztofCwalina, line 14: you should use int/long and just treat -1 as null

JoshLove-msft: Yep, just realized that this will be necessary to use Interlocked.

sdk/core/Azure.Core/tests/RetryPolicyTestBase.cs
```diff
@@ -384,6 +385,219 @@ public async Task RetryingEmitsEventSourceEvent()
             AssertRetryEvent(listener, request, 2);
         }
 
+        [Test]
```
christothes, line 388: Could we create a test that exercises the thread-safe scenario where the index is not incremented because the host value is not the same as the current fallback value?

christothes (approved):

sdk/core/Azure.Core/src/Shared/GeoRedundantFallbackPolicy.cs
```diff
@@ -0,0 +1,151 @@
...
+            // var firstMessage = _pipeline.CreateMessage();
+            // ...
+            // _pipeline.Send(firstMessage);
+            //
+            // var secondMessage = _pipeline.CreateMessage();
+            // secondMessage.Request.Uri.Host = firstMessage.Request.Uri.Host;
+            // GeoRedundantReadFallbackPolicy.SetHostAffinity(secondMessage, true);
+            // ...
+            // _pipeline.Send(secondMessage);
+            if (message.HasResponse || GetHostAffinity(message) || message.Request.Uri.Host == null)
```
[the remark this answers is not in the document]
JoshLove-msft: the property is nullable so best to be safe.

sdk/core/Azure.Core.TestFramework/src/MockTransport.cs
```diff
@@ -77,6 +77,8 @@ private async Task ProcessCore(HttpMessage message)
             if (!(message.Request is MockRequest request))
                 throw new InvalidOperationException("the request is not compatible with the transport");
 
+            message.Response = null;
```
JoshLove-msft, line 80: Use ClearResponse

JoshLove-msft: Can't as this is the test framework which does not have access to Core internals.

KrzysztofCwalina (approved):

sdk/core/Azure.Core/src/Shared/GeoRedundantFallbackPolicy.cs
```diff
@@ -93,50 +82,59 @@ public override void OnSendingRequest(HttpMessage message)
...
-            }
-            else
+            if (Volatile.Read(ref fallbackTicks) != -1)
             {
-                message.Request.Uri.Host = _writeFallbackIndex != -1 ? _writeFallbackHosts[_writeFallbackIndex] : GetPrimaryHost(message);
+                long nowTicks = DateTimeOffset.UtcNow.Ticks;
+                if (nowTicks - fallbackTicks >= _primaryCoolDown.Ticks)
+                {
+                    Interlocked.Exchange(ref fallbackIndex, -1);
+                    Interlocked.Exchange(ref fallbackTicks, nowTicks);
```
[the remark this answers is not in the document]
JoshLove-msft: Yeah this should have been setting to -1

sdk/core/Azure.Core/src/Shared/GeoRedundantFallbackPolicy.cs
```diff
@@ -12,13 +12,14 @@ namespace Azure.Core.Shared
...
-    internal class GeoRedundantFallbackPolicy : HttpPipelineSynchronousPolicy, IDisposable
+    internal class GeoRedundantFallbackPolicy : HttpPipelineSynchronousPolicy
     {
         private readonly string[] _readFallbackHosts;
         private readonly string[] _writeFallbackHosts;
         private volatile int _readFallbackIndex = -1;
         private volatile int _writeFallbackIndex = -1;
-        private readonly Timer _timer;
+        private long _readFallBackTicks = -1;
+        private long _writeFallbackTicks = -1;
```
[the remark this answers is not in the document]
JoshLove-msft: Will address in follow-up PR.

PR #32831, opened 2022-12-02
Add `SetApiVersionsFromProfile` in `ArmClientOptions`

### LINKED ISSUE #31874: Provide a profile mechanism to support ASH
ASH uses old service APIs, currently it can set the version for each service but not so convivence.
The good solution is more like we can provide a profile mechanism which can work for multiple services.

fengzhou-msft:
Resolve #31874


sdk/resourcemanager/Azure.ResourceManager/src/ArmClient.cs
```diff
@@ -111,15 +110,43 @@ internal virtual async Task<bool> CanUseTagResourceAsync(CancellationToken cance
 
         private void CopyApiVersionOverrides(ArmClientOptions options)
         {
+            if (options.ApiVersionProfile != null)
+            {
+                // Refer to https://github.com/Azure/azure-rest-api-specs/tree/main/profile#file-structure for data structure.
+                var allProfile = options.ApiVersionProfile.ToObjectFromJson() as Dictionary<string, object>;
+                var armProfile = allProfile["resource-manager"] as Dictionary<string, object>;
+                foreach (var keyValuePair in armProfile)
```
ArcturusZhang, line 118: could we change this in this way so that this is more readable and simple?
```suggestion
                foreach ((string namespaceName, Dictionary<string, object> profile) in armProfile)
```
I this way I believe we do not need the cast

sdk/resourcemanager/Azure.ResourceManager/src/ArmClient.cs
```diff
@@ -27,8 +27,7 @@ public partial class ArmClient
         private bool? _canUseTagResource;
 
         internal virtual Dictionary<ResourceType, string> ApiVersionOverrides { get; } = new Dictionary<ResourceType, string>();
-
-        internal ConcurrentDictionary<string, Dictionary<string, string>> ResourceApiVersionCache { get; } = new ConcurrentDictionary<string, Dictionary<string, string>>();
+        internal ConcurrentDictionary<string, Dictionary<string, string>> ResourceApiVersionCache { get; } = new ConcurrentDictionary<string, Dictionary<string, string>>(StringComparer.OrdinalIgnoreCase);
```
ArcturusZhang, line 30: This ConcurrentDictionary has a ignore case comparer - do we need a ignore case comparer in the Dictionary as value? Sometimes the casing of the resource types is not consistent and I believe it is better we could make this thing more robust

fengzhou-msft: Good point. I've made the changes and updated test cases.

sdk/resourcemanager/Azure.ResourceManager/src/ArmClient.cs
fengzhou-msft: Unfortunately, the .NET Standard 2.0 framework the SDK uses [does not support Deconstruct](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.keyvaluepair-2.deconstruct?view=net-7.0#applies-to) method of `KeyValuePairs`.

sdk/resourcemanager/Azure.ResourceManager/src/ArmClient.cs
```diff
@@ -111,15 +110,43 @@ internal virtual async Task<bool> CanUseTagResourceAsync(CancellationToken cance
 
         private void CopyApiVersionOverrides(ArmClientOptions options)
         {
+            if (options.ApiVersionProfile != null)
+            {
+                // Refer to https://github.com/Azure/azure-rest-api-specs/tree/main/profile#file-structure for data structure.
+                var allProfile = options.ApiVersionProfile.ToObjectFromJson() as Dictionary<string, object>;
```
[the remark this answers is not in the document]
fengzhou-msft: While it looks more straightforward to use the cast way which can just parse dictionary keys while the generic type way has to use JsonElement parser, with a benchmark test, I think it makes sense to change to use generic type.

|                     Method |     Mean |   Error |  StdDev |  Gen 0 |  Gen 1 | Allocated |
|--------------------------- |---------:|--------:|--------:|-------:|-------:|----------:|
| ToObjectFromJson<Dictionary<string, object>>| 581.5 us | 9.11 us | 8.52 us | 9.7656 | 1.9531 |    352 KB |
| ToObjectFromJson() as Dictionary<string, object> | 694.5 us | 11.76 us | 11.00 us | 12.6953 | 2.9297 |    465 KB |

sdk/resourcemanager/Azure.ResourceManager/src/ArmClient.cs
m-nash: If the reason for this is that the profile files sometimes have mis-cased resource types I don't think this is the right place to normalize as it opens up possibilities for mis matches in other areas that may not be resilient to it.  These are used in URIs which are case sensitive by design.

Instead the api that reads in the profile could normalize to the real resource type if possible OR we may need to fix the profiles.  Given that these were never used in a way that required them to be correct before it makes sense there might be a few mistakes in there that need to be corrected.

sdk/resourcemanager/Azure.ResourceManager/src/ArmClientOptions.cs
```diff
@@ -22,6 +22,11 @@ public sealed class ArmClientOptions : ClientOptions
         /// </summary>
         public ArmEnvironment? Environment { get; set; }
 
+        /// <summary>
+        /// Gets or sets API version profile. See https://github.com/Azure/azure-rest-api-specs/tree/main/profile#file-structure for data structure.
+        /// </summary>
+        public BinaryData ApiVersionProfile { get; set; }
```
m-nash, line 28: I wonder if this will create ambiguity with the `SetApiVersion` api since its not clear they do the same thing just in batch.

Perhaps we add an overload (ish) method like
`public void SetApiVersions(Uri profileUri)`
We might also consider a more explicit name like
`public void SetApiVersionsFromAzureStackProfile(Uri profileUri)`

I don't like the long winded name, but it makes it more clear we are expecting a very specific format here and you can't just point at any profile format you may have created yourself.

We might also consider a local file version of the API to support having the file already downloaded locally?

This method could read the profile and call `SetApiVersion` in a loop for each entry.

sdk/resourcemanager/Azure.ResourceManager/src/ArmClient.cs
fengzhou-msft: The `Equals` method of `ResourceType` is using ignore-case comparison, here the dictionary keys use the split namespace and type from ResourceType correspondingly, therefore I think adding the `OrdinalIgnoreCase` is consistent with the `Equals` method of `ResourceType`.  As a comparison, `ApiVersionOverrides` uses `ResourceType` as the key, and when invoking `TryGetValue(key)`, the underlying resource type string for the key is case-insensitive.
https://github.com/Azure/azure-sdk-for-net/blob/3bc773a598f2a358bd0147daf4609dd0e47bc909/sdk/core/Azure.Core/src/ResourceType.cs#L120-L123

sdk/resourcemanager/Azure.ResourceManager/src/AzureStackProfile.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// <summary>
+    /// AzureStackProfile represents the information of an Azure Stack profile.
+    /// </summary>
+    public enum AzureStackProfile
+    {
+        /// <summary> The 2020-09-01-hybrid profile. </summary>
+        Profile20200901Hybrid
+    }
+
+#pragma warning disable SA1649 // File name should match first type name
```
m-nash, line 17: lets add this to its own class or our internal extensions class so we don't need this warn disable

sdk/resourcemanager/Azure.ResourceManager/src/ArmClientOptions.cs
```diff
@@ -35,5 +38,48 @@ public void SetApiVersion(ResourceType resourceType, string apiVersion)
...
             ResourceApiVersionOverrides[resourceType] = apiVersion;
         }
+
+        /// <summary>
+        /// Sets the api versions from an Azure Stack profile.
+        /// </summary>
+        public void SetApiVersionsFromProfile(AzureStackProfile profile)
+        {
+            var assembly = Assembly.GetExecutingAssembly();
+            var profileFileFullName = $"Azure.ResourceManager.Assets.Profile.{profile.GetFileName()}";
```
m-nash, line 48: should we just change this to get `GetManifestName` and put this additional logic inside the extension method as well?

m-nash (approved):

PR #32843, opened 2022-12-02
TableClient.CreateIfNotExists / TableClient.CreateIfNotExistsAsync documentation corrected

radekfojtik:
TableClient.CreateIfNotExists / TableClient.CreateIfNotExistsAsync documentation corrected (methods do not return `null` if the table already exists)

Anyway, this change #30849 is, in my opinion, questionable. For example 
- BlobContainerClient.CreateIfNotExists still returns NULL if container already exists
- QueueClient.CreateIfNotExists still returns NULL if the queue already exists

	
It would be nice to have this unified (as it has been in the past).

### DIFF
--- a/sdk/tables/Azure.Data.Tables/CHANGELOG.md
+++ b/sdk/tables/Azure.Data.Tables/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Fixed an issue where LINQ predicates containing New expressions, such as `ent => ent.TimeStamp > new DateTimeOffset(...)`, threw an exception.
 
 ### Other Changes
+- `TableClient.CreateIfNotExists` / `TableClient.CreateIfNotExistsAsync` documentation corrected (methods do not return `null` if the table already exists)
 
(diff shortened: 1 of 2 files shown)


christothes:
> TableClient.CreateIfNotExists / TableClient.CreateIfNotExistsAsync documentation corrected (methods do not return `null` if the table already exists)
> 
> Anyway, this change #30849 is, in my opinion, questionable. For example
> 
> * BlobContainerClient.CreateIfNotExists still returns NULL if container already exists
> * QueueClient.CreateIfNotExists still returns NULL if the queue already exists
> 
> It would be nice to have this unified (as it has been in the past).

Thanks for this PR! 

Regarding the inconsistent behavior with Blob and Queue clients, they should actually conform this this new behavior. See https://github.com/Azure/azure-sdk-for-net/issues/25626 for details.

radekfojtik:
> 

radekfojtik:
@christothes thanks for the information

christothes (approved):

PR #32856, opened 2022-12-02
Ignore import/export throttling

heaths:


heaths:
The service team should be increasing the limit, but since that's not happened yet I figured I'd hedge bets.

sdk/cognitivelanguage/Azure.AI.Language.Conversations/tests/Samples/Sample5_ImportProject.cs
```diff
@@ -15,6 +15,7 @@
 
 namespace Azure.AI.Language.Conversations.Tests.Samples
 {
+    [IgnoreServiceError(429, "429", Reason = "Exceeded rate limit of S pricing tier given number of tests run")]
```
kristapratico, line 18: this is a very nice little decorator (not sure if that's the right .NET term)!

kristapratico (approved):

sdk/cognitivelanguage/Azure.AI.Language.Conversations/tests/Samples/Sample5_ImportProject.cs
heaths: "Attribute", but thanks! I created this a while back for service issues that were not necessarily endemic of real-world usage and probably wouldn't be fixed.

PR #32867, opened 2022-12-05
regenerate for the mgmt model factories

ArcturusZhang:


m-nash:
Issue found in existing dataplane library https://apiview.dev/Assemblies/Review/c09fb5af8732460eb1f9ae0a39d4265c/31f697dca0894f869dc847444135e952#d064b460307341e0b6ae9f80bd733611

m-nash:
Discriminator doesn't seem to follow what we had discussed https://apiview.dev/Assemblies/Review/cdc330ff2b674defa3e4f8f43d9549fc/90a71fa1c00a43999694effa12b19145?diffRevisionId=85cc237be8af4594803a601f2c2cd8f8&doc=False&diffOnly=True#f52784bd565b479eaf4c90eab9d1d7f8

ArcturusZhang:
> Issue found in existing dataplane library https://apiview.dev/Assemblies/Review/c09fb5af8732460eb1f9ae0a39d4265c/31f697dca0894f869dc847444135e952#d064b460307341e0b6ae9f80bd733611

I will investigate this

ArcturusZhang:
> Tweak to base library https://apiview.dev/Assemblies/Review/52ab7894c62641ffa603d3c91b38ac67/5e063790d669464198f157faa1f3e55d?diffRevisionId=5af6165c04474afabb0f2a67db03f5c0&doc=False&diffOnly=True#dc60ba9768e2401da209e0b157df8ecb

I will fix this

ArcturusZhang:
> Discriminator doesn't seem to follow what we had discussed https://apiview.dev/Assemblies/Review/cdc330ff2b674defa3e4f8f43d9549fc/90a71fa1c00a43999694effa12b19145?diffRevisionId=85cc237be8af4594803a601f2c2cd8f8&doc=False&diffOnly=True#f52784bd565b479eaf4c90eab9d1d7f8

This does not follow the way we discussed because this class is not a type with discriminator. Details can be found in the apiview comment

sdk/communication/Azure.Communication.Sms/src/autorest.md
```diff
@@ -12,4 +12,5 @@ require:
     -  https://raw.githubusercontent.com/Azure/azure-rest-api-specs/896d05e37dbb00712726620b8d679cc3c3be09fb/specification/communication/data-plane/Sms/readme.md
 payload-flattening-threshold: 10
 generation1-convenience-client: true
+model-factory-for-hlc: true
```
m-nash, line 15: nit: seems we do this quite a bit would it be better to make this an "opt out"?

sdk/dataprotection/Azure.ResourceManager.DataProtectionBackup/src/Generated/Models/BackupInstancePolicyInfo.cs
```diff
@@ -45,6 +45,8 @@ internal BackupInstancePolicyInfo(ResourceIdentifier policyId, string policyVers
         /// Gets or sets the DataStore Parameters
         /// Please note <see cref="DataStoreSettings"/> is the base class. According to the scenario, a derived class of the base class might need to be assigned here, or this property needs to be casted to one of the possible derived classes.
         /// The available derived classes include <see cref="OperationalDataStoreSettings"/>.
+        /// Please note <see cref="DataStoreSettings"/> is the base class. According to the scenario, a derived class of the base class might need to be assigned here, or this property needs to be casted to one of the possible derived classes.
```
m-nash, line 48: why are we doubling the comment here?

sdk/communication/Azure.Communication.Sms/src/autorest.md
ArcturusZhang: Well this is already an "opt out" for our new model factories. This is to restore the old behaviour of the model factory in hlc.
FYI the old model factory in hlc works like this:
it only has the models that have at least one readonly property.
it will always put discriminator in the method.

I really do not want to add customization code for them therefore I added this configuration for them so that these libraries could keep their old code. When they want to adopt our new style of model factories, they could just remove this config

sdk/dataprotection/Azure.ResourceManager.DataProtectionBackup/src/Generated/Models/BackupInstancePolicyInfo.cs
ArcturusZhang: Well I did not find the root cause for this yet.
This only happens once and it is doubling something instead of missing something therefore I am planning to finish this feature with this issue and we fix this later
what do you think?

ArcturusZhang: OK, this has been fixed.

sdk/communication/Azure.Communication.Sms/src/autorest.md
m-nash: @ArcturusZhang can we explore whether or not filtering models by ones with readonly properties would work for us as well?

I guess one issue would be if something goes from readonly to not we wouldn't want the factory to disappear?

annelo-msft (commented):
ACR data plane changes LGTM

sdk/cognitivelanguage/Azure.AI.Language.QuestionAnswering/src/Generated/QuestionAnsweringModelFactory.cs
```diff
@@ -10,7 +10,7 @@
 
 namespace Azure.AI.Language.QuestionAnswering
 {
-    /// <summary> Model factory for read-only models. </summary>
+    /// <summary> Model factory for generated models. </summary>
```
heaths, line 13: 1. Why does it matter to the user that these are generated models? That's an implementation detail that should, in fact, be hidden from the user.
2. Might be more apt to say, "Model factory for response models with read-only properties."
3. We may hand-code these to augment them. That's why this is a partial class. Also saying these are "generated models" would be incorrect.

ArcturusZhang: I believe `Model factory for models with read-only properties.` should be better because we could also include input models.
On the other hand, since we could add customized code to this class, we might add some models without readonly properties, shall we use just `Model factory for models`?

sdk/communication/Azure.Communication.Sms/src/autorest.md
ArcturusZhang: I have changed this. Now we always only generate models with at least one readonly properties.

sdk/cognitivelanguage/Azure.AI.Language.QuestionAnswering/src/Generated/QuestionAnsweringModelFactory.cs
heaths: That works too, though with correct punctuation: "Model factory for models." The main point was to leave implementation details out of it since it doesn't really matter to the user.

heaths (approved):

m-nash (approved):

ArcturusZhang:
Close this PR since the same changes have been introduced in this PR: https://github.com/Azure/azure-sdk-for-net/pull/34104

PR #32874, opened 2022-12-05
allow http scheme for localhost SAS requests

### LINKED ISSUE #32861: [FEATURE REQ] Allow HTTP SAS URLs for localhost
### Library name

Azure.Data.Tables

In Azure.Data.Tables (tested (12.7.1), it's possible to create a SAS URL that cannot be used. The following exception is thrown:

Message: **Cannot use TokenCredential without HTTPS.**

```
System.ArgumentException: Cannot use TokenCredential without HTTPS. (Parameter 'endpoint')
   at Azure.Data.Tables.TableServiceClient..ctor(Uri endpoint, TableClientOptions options)
   at Program.<Main>$(String[] args) in C:\z\Trash\2022-12-04\ConsoleApp1\ConsoleApp1\Program.cs:line 18
   at Program.<Main>(String[] args)

```
Here's a repro:
```csharp
using Azure.Data.Tables;
using Azure.Data.Tables.Sas;
using Microsoft.WindowsAzure.Storage.Table;

// Make the SAS
var tableServiceClient = new TableServiceClient("UseDevelopmentStorage=true");
var table = tableServiceClient.GetTableClient("test");
var sasUri = table.GenerateSasUri(TableSasPermissions.Read, DateTimeOffset.UtcNow.AddDays(7));

Console.WriteLine(sasUri.AbsoluteUri.ToString());

// WindowsAzure.Storage
var older = new CloudTable(sasUri);
Console.WriteLine("Using older SDK...");
Console.WriteLine(await older.ExistsAsync());

// Azure.Data.Tables
var newer = new TableServiceClient(sasUri);
Console.WriteLine("Using newer SDK...");
Console.WriteLine(await newer.GetPropertiesAsync());
```

This restriction makes total sense for production scenarios, but for testing scenarios, HTTPS is a painful hurdle. I propose that the HTTPS restriction is disabled when the host is `localhost` or `127.0.0.1`.

Azurite uses HTTP by default. To make Azurite use HTTPS, it's not easy:
https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azurite?tabs=visual-studio#certificate-configuration-https. There's several challenges like making your PC trust the Azurite cert and telling Azurite to use a specific cert which is impossible/hard for the Azurite VS Code extension.

image

christothes:
closes #32861

### DIFF
--- a/sdk/tables/Azure.Data.Tables/src/TableServiceClient.cs
+++ b/sdk/tables/Azure.Data.Tables/src/TableServiceClient.cs
@@ -115,9 +115,9 @@ namespace Azure.Data.Tables
         public TableServiceClient(Uri endpoint, TableClientOptions options = null)
             : this(endpoint, default, default, options)
         {
-            if (endpoint.Scheme != Uri.UriSchemeHttps)
+            if (endpoint.Scheme != Uri.UriSchemeHttps && !Uri.IsLoopback)
             {
-                throw new ArgumentException("Cannot use TokenCredential without HTTPS.", nameof(endpoint));
+                throw new ArgumentException("Cannot use a SAS token without HTTPS.", nameof(endpoint));
             }
             if (string.IsNullOrEmpty(endpoint.Query))
             {
@@ -141,9 +141,9 @@ namespace Azure.Data.Tables
         public TableServiceClient(Uri endpoint, AzureSasCredential credential, TableClientOptions options = null)
             : this(endpoint, default, credential, options)
         {
-            if (endpoint.Scheme != Uri.UriSchemeHttps)
+            if (endpoint.Scheme != Uri.UriSchemeHttps && !Uri.IsLoopback)
             {
-                throw new ArgumentException("Cannot use TokenCredential without HTTPS.", nameof(endpoint));
+                throw new ArgumentException("Cannot use AzureSasCredential without HTTPS.", nameof(endpoint));
             }
 
             Argument.AssertNotNull(credential, nameof(credential));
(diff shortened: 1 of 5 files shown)


sdk/tables/Azure.Data.Tables/src/TableServiceClient.cs
```diff
@@ -141,9 +141,9 @@ public TableServiceClient(Uri endpoint, TableClientOptions options = null)
         public TableServiceClient(Uri endpoint, AzureSasCredential credential, TableClientOptions options = null)
             : this(endpoint, default, credential, options)
         {
-            if (endpoint.Scheme != Uri.UriSchemeHttps)
+            if (endpoint.Scheme != Uri.UriSchemeHttps && !Uri.IsLoopback)
             {
-                throw new ArgumentException("Cannot use TokenCredential without HTTPS.", nameof(endpoint));
+                throw new ArgumentException("Cannot use AzureSasCredential without HTTPS.", nameof(endpoint));
```
jsquire, line 146: nit:  Should we consider `nameof` here?

sdk/tables/Azure.Data.Tables/tests/TableServiceClientTests.cs
```diff
@@ -33,6 +33,8 @@ public TableServiceClientTests(bool isAsync) : base(isAsync)
         private static readonly Uri _urlWithTableName = new Uri($"https://someaccount.table.core.windows.net/{TableName}");
         private static readonly Uri _devUrlWIthTableName = new Uri($"https://10.0.0.1:10002/{AccountName}/{TableName}/");
         private readonly Uri _urlHttp = new Uri($"http://someaccount.table.core.windows.net");
+        private readonly Uri _localHttp = new Uri("http://127.0.0.1:10002/accountName");
```
jsquire, line 36: Should we include an IPv6 loopback test as well?

jsquire (approved):

PR #32880, opened 2022-12-05
Generate new compute import for SDK. Includes OS Image profile and maxSurge in Microsoft.Compute

johnzhang-microsoft:

Swagger PR: https://github.com/Azure/azure-rest-api-specs/pull/21253

Adding new features to compute service:
  - OS Image Scheduled Events to the Scheduled Events Profile in Virtual Machines and Virtual Machine Scale Sets. This allows users to configure if a VM has scheduled events pop up when it's about to undergo an OS-related operation like a reimage or an OS upgrade.
  - Max Surge setting to the Rolling Upgrade Policy in Virtual Machine Scale Sets. Enabling Max Surge would create new VMs to upgrade the Scale Set, rather than updating the existing VMs. Existing VMs will be deleted once the new virtual machines are created for each batch.


johnzhang-microsoft:
@grizzlytheodore FYI. Had some issues with some commits and versioning, had to create new version of the branch.

grizzlytheodore (approved):

PR #32882, opened 2022-12-05
[Event Hubs] Event Source Fixes

jsquire:
# Summary

The focus of these changes is to correct a typo for message parameter within the event source and to attempt to improve reliability for tests that mock the event source by removing base class delegation for calls.

### DIFF
--- a/sdk/eventhub/Azure.Messaging.EventHubs/src/Diagnostics/EventHubsEventSource.cs
+++ b/sdk/eventhub/Azure.Messaging.EventHubs/src/Diagnostics/EventHubsEventSource.cs
@@ -2601,7 +2601,7 @@ namespace Azure.Messaging.EventHubs.Diagnostics
         {
             if (IsEnabled())
             {
-                WriteEvent(125, partitionId ?? string.Empty, identifier ?? string.Empty, eventHubName ?? string.Empty, consumerGroup ?? string.Empty, errorMessage ?? string.Empty);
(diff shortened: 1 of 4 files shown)


sdk/eventhub/Azure.Messaging.EventHubs/src/Diagnostics/EventHubsEventSource.cs
```diff
@@ -2601,7 +2601,7 @@ public void UnexpectedException(string errorMessage)
         {
             if (IsEnabled())
             {
-                WriteEvent(125, partitionId ?? string.Empty, identifier ?? string.Empty, eventHubName ?? string.Empty, consumerGroup ?? string.Empty, errorMessage ?? string.Empty);
+                WriteEvent(125, partitionId ?? string.Empty, identifier ?? string.Empty, eventHubName ?? string.Empty, consumerGroup ?? string.Empty, operationId ?? string.Empty, errorMessage ?? string.Empty);
```
christothes, line 2604: Do we have a test that validates that operationId makes it to the event source message?

christothes (approved):

m-redding (approved):

sdk/eventhub/Azure.Messaging.EventHubs/src/Diagnostics/EventHubsEventSource.cs
jsquire: We do not currently have any live tests that sniff the logs as they're written.   The validation that we do is over mocks, where the parameter was correctly being passed in.   We had considered adding a set, but I don't know that the ROI is there to justify it at this point.

PR #32888, opened 2022-12-05
Handle AzureSasCredential.Update on retries

### LINKED ISSUE #31442: AzureSasCredentialSynchronousPolicy does not properly handle Refresh between retry attempts
### Library name

Azure.Core

If `AzureSasCredential.Update()` is called and there are requests currently in the pipeline that have been sent through the `AzureSasCredential.Update` at least once, the policy does not properly handle validation of the presence of the SAS token in the URI.

The following code would incorrectly concatenate the old SAS token with the new one.

https://github.com/Azure/azure-sdk-for-net/blob/ff7ec416b45ef3a026f096131c820476d944f7a2/sdk/core/Azure.Core/src/Shared/AzureSasCredentialSynchronousPolicy.cs#L32-L35

### LINKED ISSUE #32514: Azure SAS credential synchronous policy is producing a wrong query string
### Library name and version

Azure.Storage.Blobs 12.10.0

### Describe the bug

We noticed that the Azure SAS credential synchronous policy is producing a query URI with duplicated parameters when the SAS credential is updated just before a request retry operation.

The Azure Storage backend is refusing systematically a request with duplicated SAS query parameters and the .NET SDK is throwing the following exception:
```
Azure.RequestFailedException: Server failed to authenticate the request. Make sure the value of the Authorization header is formed correctly including the signature.
RequestId: 22b0de9f-941d-40d0-899a-f0cc5d92dcf5
Time: 2022-11-16T09:45:10,123Z
Status: 403 (Server failed to authenticate the request. Make sure the value of the Authorization header is formed correctly including the signature.) 
ErrorCode: AuthorizationFailure
```

### Expected behavior

The query URI should be produced without duplicated parameters. The Azure SAS credential synchronous policy should override existant parameters’ values if new SAS credentials are provided.

### Actual behavior

The query URI is containing duplicated parameters after an update SAS credential operation triggered before a request retry operation.

### Reproduction Steps

Here is an example of a unit test that can be used to reproduce the issue:

```csharp
// The unit test can be added to the file AzureSasCredentialSynchronousPolicyTests.cs
[Test]
public async Task VerifyRetryAfterSasCredentialUpdateOperation()
{
  // Arrange
  const string INITIAL_QUERY_URI = "?foo=bar"; 
  const string FIRST_SIGNATURE_VALUE = "?sig=first_signature_value"; 
  const string SECOND_SIGNATURE_VALUE = "?sig=second_signature_value";
  var transport = new MockTransport(new MockResponse(403), new MockResponse(200));
  var azureSasCredential = new AzureSasCredential(FIRST_SIGNATURE_VALUE);
  var sasPolicy = new AzureSasCredentialSynchronousPolicy(azureSasCredential);

  // Act + Assert
  using (Request request = transport.CreateRequest())
  {
    request.Method = RequestMethod.Get;
    request.Uri.Query = INITIAL_QUERY_URI;
    var pipeline = new HttpPipeline(transport, new[] { sasPolicy });

    await pipeline.SendRequestAsync(request, CancellationToken.None).ConfigureAwait(false);
    Assert.AreEqual("?foo=bar&sig=first_signature_value", transport.Requests[0].Uri.Query);

    azureSasCredential.Update(SECOND_SIGNATURE_VALUE);

    await pipeline.SendRequestAsync(request, CancellationToken.None).ConfigureAwait(false);
    Assert.AreEqual("?foo=bar&sig=second_signature_value", transport.Requests[1].Uri.Query);
  }
}
```

The execution of the unit test is failing with the following error message:
```
Expected string length 35 but was 61. Strings differ at index 13.
Expected: "?foo=bar&sig=second_signature_value"
But was:  "?foo=bar&sig=first_signature_value&sig=second_signature_value"
———————-——————————————-—^
```

### Environment

```
.NET SDK (reflecting any global.json):
  Version: 6.0.400
  Commit: 7771abd614

Runtime Environment:
  OS Name: Windows
  OS Version: 10.0.19042
  OS Platform: Windows
  RID: win10-x64
```

christothes:
fixes #32514
fixes #31442


sdk/core/Azure.Core/src/AzureSasCredential.cs
```diff
@@ -25,6 +29,23 @@ public string Signature
             private set => Volatile.Write(ref _signature, value);
         }
 
+        /// <summary>
+        /// The ticks value of the DateTimeOffset when the shared access signature was last set.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public long SignatureUpdated
```
JoshLove-msft, line 36: Use DateTimeOffset?

sdk/core/Azure.Core/src/Shared/AzureSasCredentialSynchronousPolicy.cs
```diff
@@ -31,7 +31,24 @@ public override void OnSendingRequest(HttpMessage message)
             }
             if (!query.Contains(signature))
             {
-                query = string.IsNullOrEmpty(query) ? '?' + signature : query + '&' + signature;
+                bool setSignature = false;
+                // check if the signature has updated since we started processing this message
+                if (message.ProcessingStartTime.Ticks < _credential.SignatureUpdated)
+                {
+                    foreach (string previousSig in _credential.SignatureHistory.ToArray())
```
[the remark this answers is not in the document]
christothes: To take a snapshot of the list for enumeration. The alternative is to use a concurrent collection.

sdk/core/Azure.Core/src/Shared/AzureSasCredentialSynchronousPolicy.cs
```diff
@@ -31,7 +31,24 @@ public override void OnSendingRequest(HttpMessage message)
...
+                    {
+                        if (query.Contains(previousSig))
+                        {
+                            query = query.Replace(previousSig, signature);
+                            setSignature = true;
+                            break;
+                        }
+                    }
+                }
+                if (!setSignature)
```
JoshLove-msft, line 48: How could this occur? I thought the library would need to set it initially?

sdk/core/Azure.Core/src/AzureSasCredential.cs
christothes: Volatile doesn't work with DateTimeOffset.

sdk/core/Azure.Core/src/Shared/AzureSasCredentialSynchronousPolicy.cs
christothes: - Request sent
 - SasCredential.Update called
 - Request fails and retries
 - Second request (first retry) goes through this path because [this check](https://github.com/Azure/azure-sdk-for-net/pull/32888/files/ae1a7c48a39a1bae1063b80353db143adb3334c0#diff-b60f5438f5cab373d46faf5bdaafbf4aae06f264d8c9de9004dbeae4e33b1c1fR32) doesn't detect the current signature

sdk/core/Azure.Core/src/AzureSasCredential.cs
JoshLove-msft: Could do a Volatile.Read on the Ticks property, but I guess it doesn't matter much since this is hidden anyways.

JoshLove-msft (approved):

weshaggard:
@christothes could you have a look at failures at https://dev.azure.com/azure-sdk/public/_build/results?buildId=2057402&view=ms.vss-test-web.build-test-results-tab&runId=37955336&paneView=debug&resultId=101867.

It looks like this might have introduced some inconsistent tests:
```
Expected string length 35 but was 61. Strings differ at index 13.
Expected: "?foo=bar&sig=second_signature_value"
But was:  "?foo=bar&sig=first_signature_value&sig=second_signature_value"
```

christothes:
Fixed here https://github.com/Azure/azure-sdk-for-net/pull/33078

PR #32889, opened 2022-12-05
Improve original, current error messages for recorded tests

### LINKED ISSUE #31903: Suggestions to improve recording framework exceptions
I was re-recording tests after rev'ing the default service version and adding it to the test class, and several tests failed. Most had the familiar "the test was re-recorded and should pass", but a few just said the records didn't match (in this case, only on the api-version). They weren't re-recorded. Turns out some service issues were actually returning an error so the framework didn't re-recorded. This behavior makes sense, but it would be nice if the error logged to test output said that instead of just the original "records didn't match" error. If possible, perhaps show both: if not the api-version mismatching, perhaps the original reason the records didn't match would be handy.

heaths:
Fixes #31903

### DIFF
--- a/sdk/core/Azure.Core.TestFramework/src/RecordedTestAttribute.cs
+++ b/sdk/core/Azure.Core.TestFramework/src/RecordedTestAttribute.cs
@@ -70,19 +70,27 @@ namespace Azure.Core.TestFramework
                         // If the recording succeeded, set an error result.
                         if (context.CurrentResult.ResultState.Status == TestStatus.Passed)
                         {
-                            context.CurrentResult.SetResult(ResultState.Error,
-                                "Test failed playback, but was successfully re-recorded (it should pass if re-run)." +
+                            string message = "Test failed playback, but was successfully re-recorded. It should pass if re-run." +
                                 Environment.NewLine +
                                 Environment.NewLine +
-                                originalResult.Message);
+                                originalResult.Message;
+
+                            context.CurrentResult.SetResult(ResultState.Error, message);
                         }
                         else
                         {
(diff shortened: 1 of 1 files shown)


heaths:
Apart from fixing the grammar a bit, I also put the current error for re-recording failures first since that one is probably more meaningful than a rather generic "the recording didn't match" error. Now the failure case will look like:

```
    The [RecordedTest] attribute attempted to re-record, but failed:

Azure.RequestFailedException : The specified deployment cannot be found within the provided subscription.
Status: 404 (Not Found)
ErrorCode: NotFound

Content:
{
  "error": {
    "code": "NotFound",
    "message": "The specified deployment cannot be found within the provided subscription."
  }
}

Headers:
Date: Mon, 05 Dec 2022 22:54:34 GMT
Server: Kestrel
Cache-Control: no-store, proxy-revalidate, no-cache, max-age=0, private
Pragma: no-cache
Request-Id: 7fec8391-8b99-4c66-a152-ade2905ad888
apim-request-id: REDACTED
x-envoy-upstream-service-time: REDACTED
Strict-Transport-Security: REDACTED
X-Content-Type-Options: REDACTED
x-ms-region: REDACTED
Content-Length: 136
Content-Type: application/json


Original failure:

Azure.Core.TestFramework.TestRecordingMismatchException : Unable to find a record for the request POST https://languagesdkresource.cognitiveservices.azure.com/language/:analyze-conversations?api-version=2022-10-01-preview
Header differences:
Body differences:
Request and record bodies do not match at index 122:
     request: "ipantId":"2"}},"parameters":{""
     record:  "ipantId":"1"}},"parameters":{""
```

christothes (approved):

JoshLove-msft (approved):

PR #32890, opened 2022-12-05
[dmlib] Audit XML comments in `Azure.Storage.DataMovement` dmlib library

vincenttran-msft:
This PR contains the XML comment changes for the `Azure.Storage.DataMovement` library.

### DIFF
--- a/sdk/storage/Azure.Storage.DataMovement/src/JobPartInternal.cs
+++ b/sdk/storage/Azure.Storage.DataMovement/src/JobPartInternal.cs
@@ -20,12 +20,12 @@ namespace Azure.Storage.DataMovement
         public QueueChunkDelegate QueueChunk { get; internal set; }
 
         /// <summary>
-        /// Part Number
+        /// Part number of the current job part.
         /// </summary>
         public int PartNumber;
 
         /// <summary>
-        /// DataTransfer communicate when the transfer has finished and the progress
+        /// DataTransfer object that communicates when the transfer completes and it's current progress.
         /// </summary>
         internal DataTransfer _dataTransfer { get; set; }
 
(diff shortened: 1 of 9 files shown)


sdk/storage/Azure.Storage.DataMovement/src/LocalTransferCheckpointer.cs
```diff
@@ -191,7 +191,7 @@ public override Task<bool> TryRemoveStoredTransferAsync(string id, CancellationT
         /// <summary>
         /// Lists all the transfers contained in the checkpointer.
         /// </summary>
-        /// <param name="cancellationToken"></param>
+        /// <param name="cancellationToken">The cancellation token of the current transfer.</param>
```
vincenttran-msft, line 194: Is it accurate to describe this as cancellation token of the current transfer? Or is it really the cancellation token of the checkpointer? Will change pending feedback

sdk/storage/Azure.Storage.DataMovement/src/JobPartInternal.cs
```diff
@@ -20,12 +20,12 @@ internal abstract class JobPartInternal
         public QueueChunkDelegate QueueChunk { get; internal set; }
```
vincenttran-msft, line 20: I realized this was internal so I stopped -- but leaving it here just to show what my changes look like.

sdk/storage/Azure.Storage.DataMovement/src/LocalTransferCheckpointer.cs
amnguye: For everywhere that has cancellation token let's use
```
<param name="cancellationToken">
        /// Optional <see cref="CancellationToken"/> to propagate
        /// notifications that the operation should be cancelled.
        /// </param>
        ````

File: sdk/storage/Azure.Storage.DataMovement/src/LocalTransferCheckpointer.cs [hunk 3]
```diff
@@ -96,15 +96,15 @@ public override Task TryAddTransferAsync(string id, CancellationToken cancellati
...
         /// </summary>
-        /// <param name="id"></param>
-        /// <param name="partNumber"></param>
-        /// <param name="offset"></param>
-        /// <param name="buffer"></param>
-        /// <param name="cancellationToken"></param>
+        /// <param name="id">The transfer ID.</param>
+        /// <param name="partNumber">The part number of the current transfer.</param>
+        /// <param name="offset">The offset of the current transfer as a long.</param>
+        /// <param name="buffer">The buffer of the current transfer in bytes.</param>
```
amnguye, line 106: Maybe something like "The buffer to write data from to the checkpoint."

amnguye (approved):

sdk/storage/Azure.Storage.DataMovement/src/StorageTransferProgress.cs
```diff
@@ -7,27 +7,27 @@
...
     {
         /// <summary>
-        /// Number of Files that succeeded in transfer
+        /// Number of files that were transferred successfully.
         /// </summary>
         public int CompletedCount { get; internal set; }
 
         /// <summary>
-        /// Number of files that were skipped in transfer due to overwrite being set to not overwrite files, but the files exists already
+        /// Number of files that were skipped in transfer due to files already existing in the destination and overwite policy set to not overwrite existing files.
```
amnguye, line 20: Maybe we should rewrite this to be

Number of files that were skipped in transfer due to files already existing in the destination. Files will be skipped, if the overwrite policy is set to not overwrite existing files.

Issue #32456, opened 2022-11-13
EUAP region pages should contain EUAP "Early Updates Access Program" meaning or mention that they are canary regions


vs-li:
The pages for CentralUSEUAP and EastUS2EUAP should contain some short description that they are canary regions or "Early Updates Access Program" regions.

Ex. instead of "Name of the Azure Central US EUAP region in the Azure Cosmos DB service." have "Name of the Azure Central US Early Updates Access Program (EUAP) region in the Azure Cosmos DB service."

See: https://stackoverflow.com/questions/50647007/what-is-an-euap-region-in-azure

---
#### Document Details

⚠ *Do not edit this section. It is required for learn.microsoft.com ➟ GitHub issue linking.*

* ID: 1dbf611e-ee34-9f8d-f9de-df57119912f7
* Version Independent ID: a2e6ff2e-62f5-e4f8-f167-df8e2280b82a
* Content: [LocationNames.CentralUSEUAP Field (Microsoft.Azure.Documents) - Azure for .NET Developers](https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.documents.locationnames.centraluseuap?view=azure-dotnet&source=docs)
* Content Source: [xml/Microsoft.Azure.Documents/LocationNames.xml](https://github.com/Azure/azure-docs-sdk-dotnet/blob/master/xml/Microsoft.Azure.Documents/LocationNames.xml)
* Service: **cosmos-db**
* GitHub Login: @rloutlaw
* Microsoft Alias: **routlaw**

PramodValavala-MSFT:
@vs-li Thanks for the feedback! We will update accordingly.

PR #32891, opened 2022-12-06
Fix failing SIP Routing tests in Communication/PhoneNumbers 

jannovak-msft:
# Description
The SIP Routing tests started to fail when run in parallel by pipelines.
Used dynamic resources instead of a static one. So each pipeline works with own resource.
Code altered to use randomized domain FQDNs based on UUID. (Each domain can be assigned only to one resource at the time)

[Issue #3047678 Fix Sip Routing Tests - .NET SDK](https://skype.visualstudio.com/SPOOL/_workitems/edit/3047678)
[Successful CI run #2034017](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2034017&view=results)

Related PR:
[Pull Request #23812 - JavaScript SDK](https://github.com/Azure/azure-sdk-for-js/pull/23812)
[Pull Request #32456- Java SDK](https://github.com/Azure/azure-sdk-for-java/pull/32456)

### DIFF
--- a/sdk/communication/Azure.Communication.PhoneNumbers/tests/SipRouting/Infrastructure/TestData.cs
+++ b/sdk/communication/Azure.Communication.PhoneNumbers/tests/SipRouting/Infrastructure/TestData.cs
@@ -3,39 +3,49 @@
 
 using System;
 using System.Collections.Generic;
+using Azure.Core.TestFramework;
 
 namespace Azure.Communication.PhoneNumbers.SipRouting.Tests
 {
-    internal static class TestData
+    public class TestData
     {
-        private static readonly Guid Domain = Guid.NewGuid();
-        public static readonly List<string> Fqdns = new List<string>(){ "sbs1." + Domain + ".com", "sbs2." + Domain + ".com" };
-        public static readonly int[] TrunkPorts = { 1122, 1123 };
+        private readonly string Domain;
+        public readonly List<string> Fqdns;
+        public readonly int[] TrunkPorts = { 1122, 1123 };
 
-        public static readonly List<SipTrunk> TrunkList = new List<SipTrunk>
+        public readonly List<SipTrunk> TrunkList;
+        public readonly SipTrunk NewTrunk;
+
+        public readonly SipTrunkRoute RuleNavigateToTrunk1;
+        public readonly SipTrunkRoute RuleNavigateToAllTrunks;
+        public readonly SipTrunkRoute RuleNavigateToNewTrunk;
+
+        public TestData(RecordedTestMode? mode)
         {
-            new SipTrunk(Fqdns[0], TrunkPorts[0]),
-            new SipTrunk(Fqdns[1], TrunkPorts[1])
-        };
+            Domain = mode != RecordedTestMode.Playback ? ("." + Guid.NewGuid()) : "";
+            Fqdns = new List<string>() { "sbs1" + Domain + ".com", "sbs2" + Domain + ".com" };
+            TrunkList = new List<SipTrunk>
+            {
+                new SipTrunk(Fqdns[0], TrunkPorts[0]),
+                new SipTrunk(Fqdns[1], TrunkPorts[1])
+            };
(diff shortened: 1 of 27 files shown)


sdk/communication/Azure.Communication.PhoneNumbers/tests/SipRouting/Infrastructure/TestData.cs
```diff
@@ -1,39 +1,51 @@
...
+        public readonly SipTrunkRoute RuleNavigateToNewTrunk;
 
-        public static readonly SipTrunkRoute RuleNavigateToAllTrunks = new SipTrunkRoute(
-            name: "Last rule",
-            description: "Handle all other numbers'",
-            numberPattern: @"\+[1-9][0-9]{3,23}",
-            trunks: Fqdns);
+        public TestData(RecordedTestMode? mode)
+        {
+            Domain = mode != RecordedTestMode.Playback ? ("." + Guid.NewGuid()) : "";
```
christothes, line 25: Do you want this to be truly random, or recording random? (e.g `Recording.Random.NewGuid()`)

sdk/communication/Azure.Communication.PhoneNumbers/tests/SipRouting/SipRoutingClientLiveTests.cs
```diff
@@ -28,6 +28,20 @@ public SipRoutingClient InitializeTest()
...
+        public void ClearConfiguration()
+        {
+            var client = CreateClient();
+            if (client.GetRoutesAsync().Result.Value.Count > 0)
+            {
+                client.SetRoutesAsync(new List<SipTrunkRoute>()).Wait();
+            }
+            if (client.GetTrunksAsync().Result.Value.Count > 0)
+            {
+                client.SetTrunksAsync(new List<SipTrunk>()).Wait();
```
christothes, line 41:
```suggestion
                client.SetRoutesAsync(new List<SipTrunkRoute>()).EnsureCompleted();
            }
            if (client.GetTrunksAsync().Result.Value.Count > 0)
            {
                client.SetTrunksAsync(new List<SipTrunk>()).EnsureCompleted();
```

sdk/communication/Azure.Communication.PhoneNumbers/tests/SipRouting/Infrastructure/TestData.cs
jannovak-msft: Used `Recording.Random.NewGuid()` instead of `Guid.NewGuid()`
Domain sanitizer was removed, since it is no longer needed.

`Guid.NewGuid()` was called directly in the `TestData` constructor.
Now the Guid parameter is passed into the constructor from the test.
(In the test the `Recording` variable is accessible via parent)

`TestData = new TestData(Recording.Random.NewGuid())` is called from a `[SetUp]` method now because the `Recording` variable is not setup until then.

sdk/communication/Azure.Communication.PhoneNumbers/tests/SipRouting/SipRoutingClientLiveTests.cs
jannovak-msft: Used `EnsureCompleted()`.
To be able to run async tests the `Wait()` call was still needed.

var task = client.SetRoutesAsync(new List<SipTrunkRoute>());
task.Wait();
task.EnsureCompleted();

christothes: Since this is just in the teardown which is not an async method - why not just use the sync versions (`SetTrunk(...)` and `GetTrunks(...)`) and avoid sync over async?

jannovak-msft: As discussed offline, in the end used a solution with async teardown method.

christothes (approved):

PR #32895, opened 2022-12-06
[Storage] [DataMovement] Remove Delegates after finished use from internal events

### LINKED ISSUE #32737: [DataMovement] [Blobs] Add way to remove delegates from events
### Library name and version

Azure.Storage.DataMovement; Azure.Storage.DataMovement.Blobs

### Query/Question

https://github.com/Azure/azure-sdk-for-net/pull/32483/files#r1026862432

### Environment

_No response_

amnguye:
Resolves https://github.com/Azure/azure-sdk-for-net/issues/32737


jalauzon-msft (commented):
Left a couple of questions on the pattern here but totally could be me just not fully understanding what's going on.

sdk/storage/Azure.Storage.DataMovement/src/CommitChunkHandler.cs
```diff
@@ -73,17 +74,31 @@ public struct Behaviors
...
+        {
+            CleanUp();
+        }
+
+        public void CleanUp()
+        {
+            if (_transferType == TransferType.Sequential)
             {
-                AddQueueBlockEvent();
+                _commitBlockHandler -= AddQueueBlockEvent();
```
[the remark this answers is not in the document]
amnguye: To unsubscribe from the event, you have to remove the delegate from the event.

I was just mostly following this.
https://stackoverflow.com/questions/4526829/why-and-how-to-avoid-event-handler-memory-leaks

sdk/storage/Azure.Storage.DataMovement/src/CommitChunkHandler.cs
```diff
@@ -73,17 +74,31 @@ public struct Behaviors
...
             }
-            AddCommitBlockEvent();
+            _commitBlockHandler -= AddCommitBlockEvent();
         }
 
-        public void AddCommitBlockEvent()
+        public SyncAsyncEventHandler<StageChunkEventArgs> AddCommitBlockEvent()
```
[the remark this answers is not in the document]
amnguye: Renamed it to GetCommitBlockEvent and also to the rest of the names too.

sdk/storage/Azure.Storage.DataMovement/src/CopyStatusHandler.cs
```diff
@@ -49,37 +49,44 @@ public CopyStatusHandler(Behaviors behaviors)
...
-            _getStatusHandler += async (CopyStatusEventArgs args) =>
+            Cleanup();
+        }
+
+        public void Cleanup()
+        {
+            _getStatusHandler -= GetStatusEvent;
+        }
+
+        public async Task GetStatusEvent(CopyStatusEventArgs args)
```
jalauzon-msft, line 65: nit: I see you turned the function `private` above. Did you want to do that for all of these?

sdk/storage/Azure.Storage.DataMovement/src/DownloadChunkHandler.cs
```diff
@@ -117,83 +117,79 @@ public struct Behaviors
...
-            _downloadChunkEventHandler += async (DownloadRangeEventArgs args) =>
+            CleanUp();
+        }
+
+        public void CleanUp()
+        {
+            _downloadChunkEventHandler -= AddDownloadChunkEvent;
+        }
+
+        public async Task AddDownloadChunkEvent(DownloadRangeEventArgs args)
```
jalauzon-msft, line 133: Did you want to change this name too?

jalauzon-msft (approved):
Thanks! This approach looks better to me but honestly, I'm not too familiar with delegates/events so may be worth having someone else take a look if you want a more thorough review of the logic.

Left a couple more minor comments on naimg.

sdk/storage/Azure.Storage.DataMovement/src/CommitChunkHandler.cs
```diff
@@ -73,70 +74,73 @@ public struct Behaviors
...
             {
-                AddQueueBlockEvent();
+                _commitBlockHandler -= GetQueueBlockEvent;
             }
-            AddCommitBlockEvent();
+            _commitBlockHandler -= GetCommitBlockEvent;
         }
 
-        public void AddCommitBlockEvent()
+        private async Task GetCommitBlockEvent(StageChunkEventArgs args)
```
jalauzon-msft, line 99: Sorry to bring up naming again (feel free to ignore) but with this approach maybe it's better to just call this `CommitBlockEvent` since it **is** the event? Like what you did below with `JobPartEvent`.

This would apply to all delegate functions in this review.

Issue #32858, opened 2022-12-03
[DataMovement] [Blobs] Generate and merge all test recordings


amnguye:
### Library name and version

Azure.Storage.DataMovement; Azure.Storage.DataMovement.Blobs

### Query/Question

Undo live tests bracket and add all necessary json test recorded files

note: we don't want to add the large testing files. So maybe we need to add some tests with smaller blob sizes, so we dont generate such enormous json test recorded files.

https://github.com/Azure/azure-sdk-for-net/pull/32483

PR #32898, opened 2022-12-06
[Storage] [DataMovement] Local Resource Tests

amnguye:
Part of https://github.com/Azure/azure-sdk-for-net/issues/32858

### DIFF
--- a/sdk/storage/Azure.Storage.DataMovement/tests/LocalFileStorageResourceTests.cs
+++ b/sdk/storage/Azure.Storage.DataMovement/tests/LocalFileStorageResourceTests.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Drawing;
 using System.IO;
 using System.Linq;
 using System.Runtime.InteropServices;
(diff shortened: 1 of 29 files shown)


sdk/storage/Azure.Storage.DataMovement/tests/LocalFileStorageResourceTests.cs
```diff
@@ -82,60 +81,79 @@ public void Ctor_string()
...
             string path = await CreateRandomFileAsync(Path.GetTempPath(), size:0);
-            var data = GetRandomBuffer(size);
-            using (var stream = new MemoryStream(data))
+            try
             {
-                using FileStream fileStream = new FileStream(path, FileMode.OpenOrCreate);
-                await stream.CopyToAsync(fileStream);
-            }
+                var data = GetRandomBuffer(size);
+                using (var stream = new MemoryStream(data))
```
jalauzon-msft, line 92: Not sure the type returned by GetRandomBuffer but I _think_ you could consider using `File.WriteAllBytes()` to write the data to the local file instead of using a `MemoryStream` and copying it?

Just a thought as its probably doing the same thing. And if this is an established pattern in the file/package, then also fine to leave it for that reason.

jalauzon-msft: Also see this is only part of the diff because of indent so totally fine to leave it :)

jalauzon-msft (approved):
LGTM!

PR #32902, opened 2022-12-06
Add auto-generate sample for ProviderHub and Resources

dvbb:


sdk/providerhub/Azure.ResourceManager.ProviderHub/src/autorest.md
```diff
@@ -49,6 +49,12 @@ rename-rules:
   URI: Uri
   Etag: ETag|etag
 
+request-path-to-resource-name:
+  /subscriptions/{subscriptionId}/providers/Microsoft.ProviderHub/providerRegistrations/{providerNamespace}/resourcetypeRegistrations/{resourceType}/skus/{sku}: NestedResourceTypeSku
```
fengzhou-msft, line 53: Should this be `ResourceTypeSku`?

dvbb: It already has a model called [ResourceTypeSku](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/providerhub/Azure.ResourceManager.ProviderHub/src/Generated/Models/ResourceTypeSku.cs#L16)

fengzhou-msft: Can we rename that model to something like `ResourceTypeSkuInfo`? I don't think `{providerNamespace}/resourcetypeRegistrations/{resourceType}` is considered a nested resource. We should count the levels starting from `{providerNamespace}`. Therefore, this is a top level resource after it's registered altough we're registering it under `Microsoft.ProviderHub/providerRegistrations`. And the description of the `CreateOrUpdate` method of this request path also says `Creates or updates the resource type skus in the given resource type.`.

dvbb: fixed, change `NestedResourceTypeSku` => `ResourceTypeSkuInfo`

fengzhou-msft: I mean we use `ResourceTypeSku` for the resource name and rename the existing  `ResourceTypeSku` model to `ResourceTypeSkuInfo`. This way the resource names are relatively consistent.

fengzhou-msft (approved):

PR #32923, opened 2022-12-07
[Perf] Add pipeline parameter ToolsRepoCommitish

mikeharder:
- Allows manually running a perf pipeline against a branch or PR of the tools repo


weshaggard (approved):
Looks reasonable. At some point we may want to actually turn this into a repository ref instead of passing it in as a parameter. That will allow us to have automation to pin this to tags instead of defaulting to main.

mikeharder:
Closing for now.  These files will have more changes as part of #3530.

PR #32929, opened 2022-12-07
[Storage] [DataMovement] Rename encodedPath to path

### LINKED ISSUE #32924: [Storage] [DataMovement] [Blobs] Rename encodedPath to path
### Library name and version

Azure.Storage.DataMovement; Azure.Storage.DataMovement.Blobs

### Query/Question

A lot of places in the API randomly have encodedPath and path. We need to just have one name, probably path.

### Environment

_No response_

amnguye:
Resolves https://github.com/Azure/azure-sdk-for-net/issues/32924


amnguye:
I guess this is unrelated to the PR but looks like I need to rerecord the GetChildStorageResource tests

jalauzon-msft (approved):

PR #32937, opened 2022-12-07
Update the LoadTesting for stable release.

ArthurMa1978:


Yao725 (approved):

sdk/loadtestservice/Azure.ResourceManager.LoadTesting/src/Generated/Models/LoadTestResourcePageList.Serialization.cs
```diff
@@ -16,7 +16,7 @@ internal partial class LoadTestResourcePageList
     {
         internal static LoadTestResourcePageList DeserializeLoadTestResourcePageList(JsonElement element)
```
prativen, line 17: Should the suffix be PageList or ListResult? There are other places where ListResult is used as suffix for similar concept eg. LoadTestingQuotaListResult. The suffix should be consistent?

sdk/loadtestservice/Azure.ResourceManager.LoadTesting/src/autorest.md
```diff
@@ -20,7 +20,8 @@ irregular-plural-words:
   quota: quota
 
 rename-mapping:
-  LoadTestResource: LoadTesting
+  Resource: LoadTestingBaseResource
+  LoadTestResource: LoadTestingResource
   QuotaResource: LoadTestingQuota
   CheckQuotaAvailabilityResponse: LoadTestingQuotaAvailabilityResponse
```
prativen, line 26: CheckQuotaAvailabilityResponse: LoadTestingQuotaAvailability ?

sdk/loadtestservice/Azure.ResourceManager.LoadTesting/src/autorest.md
```diff
@@ -20,7 +20,8 @@ irregular-plural-words:
...
   EncryptionProperties: CustomerManagedKeyEncryptionProperties
```
prativen, line 27: QuotaBucketRequest: LoadTestingQuotaBucket?

prativen, line 27: OutboundEnvironmentEndpointCollection: OutboundEnvironmentEndpointPageList ?

prativen, line 27: QuotaResourceList: LoadTestingQuotaPageList?

sdk/loadtestservice/Azure.ResourceManager.LoadTesting/src/Generated/Models/LoadTestResourcePageList.Serialization.cs
ArthurMa1978: This is internal function that don't need to change the name

sdk/loadtestservice/Azure.ResourceManager.LoadTesting/src/autorest.md
ArthurMa1978: This is a common naming pattern for all CheckAvailability method and related models

sdk/loadtestservice/Azure.ResourceManager.LoadTesting/src/autorest.md
ArthurMa1978: cThis is a common naming pattern for a model which is used as a method parameter, end with `Content` suffix.

ArthurMa1978: Here the `Collection` is a reserved suffix for ResourceCollection, so here rename it, and the naming pattern is change the suffix from `Collection` to `ListResult`.

ArthurMa1978: Similar as above, the suffix is `ListResult` as it used as a result / response of a method.

prativen: @ArthurMa1978 in that case, should LoadTestResourcePageList be renamed to LoadTestResourceListResult

ArthurMa1978: The `LoadTestResourcePageList` is internal class that we don't need to rename, the rename function only works for all public class / interface / method.

PR #32939, opened 2022-12-07
initial commit for review with SDK team

imakhlin:
The intent of this PR is to:
1. Share and communicate the issues/questions/concerns with core SDK dev team
2. Demonstrate SDK development lifecycle


sdk/adp/tests.yml
```diff
@@ -0,0 +1,7 @@
+trigger: none
+
+extends:
+  template: /eng/pipelines/templates/stages/archetype-sdk-tests.yml
+  parameters:
+    ServiceDirectory: adp
+    SupportedClouds: 'Public'
```
benbp, line 7:
```suggestion
```

sdk/adp/test-resources.json
```diff
@@ -0,0 +1,85 @@
...
+                "description": "The location of the resource. By default, this is the same as the resource group."
+            }
+        }
+    },
+    "variables": {
+        "secretValue": "Very secret value"
+    },
+    "resources": [
+        {
+            "type": "Microsoft.KeyVault/vaults",
```
benbp, line 39: @jsquire @heaths is deploying a keyvault a live test template default or something?

jsquire: The template project uses one in its live tests, but it shouldn't be here unless it's needed.

PR #32943, opened 2022-12-07
Add DocumentModelAdministrationClientBuilderExtensions

rgylesbedford:
Allow DocumentModelAdministrationClient to be registered with IAzureClientFactoryBuilder


kinelski:
Hello @rgylesbedford, and thank you for your contribution.

Originally we didn't include ASP.NET Core integration for the `DocumentModelAdministrationClient` because, according to our [design guidelines](https://azure.github.io/azure-sdk/dotnet_implementation.html#integration-with-aspnet-core):
> ✅ DO provide integration extension methods for every top level client type users are expected to start with in the main namespace. Do not include integration extension methods for secondary clients, child clients, or clients in advanced namespaces.

`DocumentModelAdministrationClient` is considered a secondary client since most champion scenarios don't require it. We expect model management to be done via [Form Recognizer Studio](https://formrecognizer.appliedai.azure.com/studio) by most devs.

Do you have a specific scenario you want to support?

PR #32945, opened 2022-12-07
Use PackageReference to Azure.Core for ShareLink

### LINKED ISSUE #32065: Update packageref to Azure.Core in sharelink sample
PR #31790 added types which haven't shipped in Azure.Core yet. I was able to update the sharelink sample so that it at least builds in the repo using a `ProjectReference` to Azure.Core, but the out-of-repo configuration in _sharelink.csproj_ will need to be updated to the latest Azure.Core with the require `NullResponse<T>` once shipped.

heaths:
Resolves #32065

### DIFF
--- a/sdk/keyvault/samples/sharelink/ShareLink.csproj
+++ b/sdk/keyvault/samples/sharelink/ShareLink.csproj
@@ -18,6 +18,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Azure.Core" />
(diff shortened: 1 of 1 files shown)


heaths:
-

heaths:
This now builds correctly once a customer adds Azure.Core to their project as documented, but there's a bunch of nullable errors i.e., C# nullable. Not sure if we want to fix these at some point, but that's a whole separate issue from this re: `NullableResponse<T>`.

pallavit (approved):

PR #32953, opened 2022-12-07
Refactor fallback policy

JoshLove-msft:


heaths (dismissed):

sdk/core/Azure.Core/src/Shared/GeoRedundantFallbackPolicy.cs
```diff
@@ -81,72 +79,97 @@ public override void OnSendingRequest(HttpMessage message)
...
-            {
-                int next = current + 1;
+            fallback.AdvanceIfNeeded(message);
+            UpdateHostIfNeeded(message, fallback);
+        }
 
-                // reset to null to indicate primary should be used when we reach the end
-                if (next >= fallbackHosts.Length)
-                    next = -1;
+        private static void UpdateHostIfNeeded(HttpMessage message, Fallback fallback)
```
christothes, line 100: could this also go inside `Fallback`?

christothes (approved):

sdk/core/Azure.Core/src/Shared/GeoRedundantFallbackPolicy.cs
JoshLove-msft: It could, but I wanted to only include methods in Fallback when mutating the fallback as opposed to mutating the message itself.

PR #32996, opened 2022-12-09
Fix issues related to cancellation in ResponseBodyPolicy

### LINKED ISSUE #32577: Cancelling HttpPipeline.Send() can fail by disposing stream during read
Run the following app (or clone https://github.com/mikeharder/http-pipeline-send-cancel):

```c#
using Azure.Core;
using Azure.Core.Pipeline;
using Microsoft.AspNetCore.Builder;
using Microsoft.AspNetCore.Hosting;
using Microsoft.AspNetCore.Http;
using Microsoft.Extensions.Options;
using System.Net;
using System.Threading;

namespace HttpPipelineSendCancel
{
    internal class Program
    {
        // Config
        private const int MinDelayMilliseconds = 1;
        private const int MaxDelayMilliseconds = 20;
        private const int ResponseSize = 10_000_000;

        private static readonly string _response = new string('a', ResponseSize);
        private static readonly HttpPipeline _pipeline = HttpPipelineBuilder.Build(new TestClientOptions());

        static async Task Main()
        {
            Console.WriteLine("Starting web server...");
            await new WebHostBuilder()
                .UseKestrel()
                .Configure(app => app.Run(async context =>
                {
                    await context.Response.WriteAsync(_response);
                }))
                .Build()
                .StartAsync();

            var random = new Random();
            while (true)
            {
                var delay = random.Next(MinDelayMilliseconds, MaxDelayMilliseconds);
                using (var cts = new CancellationTokenSource(millisecondsDelay: delay))
                {
                    try
                    {
                        var message = CreateMessage();
                        _pipeline.Send(message, cts.Token);
                        Console.WriteLine($"[{DateTime.Now:hh:mm:ss.fff}] completed");
                    }
                    catch (TaskCanceledException)
                    {
                        Console.WriteLine($"[{DateTime.Now:hh:mm:ss.fff}] cancelled");
                    }
                }
            }
        }

        private static HttpMessage CreateMessage()
        {
            var message = _pipeline.CreateMessage();
            message.Request.Uri.Reset(new Uri("https://localhost:5001"));
            return message;
        }

        private class TestClientOptions : ClientOptions { }
    }
}
```

The app should fail pretty quickly with the following exception.  If it doesn't fail, try modifying `MinDelayMilliseconds`, `MaxDelayMilliseconds`, or `ResponseSize` as appropriate for your dev machine.

```
Unhandled exception. System.NotSupportedException:  This method may not be called when another read operation is pending.
   at System.Net.Security.SslStream.ReadAsyncInternal[TIOAdapter](Memory`1 buffer, CancellationToken cancellationToken)
   at System.Runtime.CompilerServices.PoolingAsyncValueTaskMethodBuilder`1.StateMachineBox`1.System.Threading.Tasks.Sources.IValueTaskSource<TResult>.GetResult(Int16 token)
   at System.Net.Security.SslStream.Read(Byte[] buffer, Int32 offset, Int32 count)
   at System.IO.Stream.Read(Span`1 buffer)
   at System.Net.Http.HttpConnection.Read(Span`1 destination)
   at System.Net.Http.HttpConnection.ChunkedEncodingReadStream.Read(Span`1 buffer)
   at System.Net.Http.HttpBaseStream.Read(Byte[] buffer, Int32 offset, Int32 count)
   at Azure.Core.Pipeline.ResponseBodyPolicy.CopyTo(Stream source, Stream destination, CancellationTokenSource cancellationTokenSource)
   at Azure.Core.Pipeline.ResponseBodyPolicy.ProcessAsync(HttpMessage message, ReadOnlyMemory`1 pipeline, Boolean async)
   at Azure.Core.Pipeline.TaskExtensions.EnsureCompleted(ValueTask task)
   at Azure.Core.Pipeline.ResponseBodyPolicy.Process(HttpMessage message, ReadOnlyMemory`1 pipeline)
   at Azure.Core.Pipeline.HttpPipelinePolicy.ProcessNext(HttpMessage message, ReadOnlyMemory`1 pipeline)
   at Azure.Core.Pipeline.LoggingPolicy.Process(HttpMessage message, ReadOnlyMemory`1 pipeline)
   at Azure.Core.Pipeline.HttpPipelinePolicy.ProcessNext(HttpMessage message, ReadOnlyMemory`1 pipeline)
   at Azure.Core.Pipeline.RedirectPolicy.ProcessAsync(HttpMessage message, ReadOnlyMemory`1 pipeline, Boolean async)
```

We believe the bug is our code can call Dispose() concurrently with Read():

https://github.com/Azure/azure-sdk-for-net/blob/785c355bde40ae4b59d922611666f52b16074e6e/sdk/core/Azure.Core/src/Pipeline/Internal/ResponseBodyPolicy.cs#L80

Ideally, this would throw an `ObjectDisposedException`, but it throws the "nested read" exception instead due to https://github.com/dotnet/runtime/issues/78586.

JoshLove-msft:
Fixes https://github.com/Azure/azure-sdk-for-net/issues/32577

- Handle NotSupportedException being thrown by Stream when disposing due to cancellation
- Register cancellation delegate if the underlying user token can be cancelled

### DIFF
--- a/sdk/core/Azure.Core/src/Pipeline/Internal/ResponseBodyPolicy.cs
+++ b/sdk/core/Azure.Core/src/Pipeline/Internal/ResponseBodyPolicy.cs
@@ -75,7 +75,9 @@ namespace Azure.Core.Pipeline
 
             if (message.BufferResponse)
             {
-                if (networkTimeout != Timeout.InfiniteTimeSpan)
+                // If there is a network timeout or the user passed a cancellation token, then we will dispose the stream when
+                // cancellation is triggered.
+                if (networkTimeout != Timeout.InfiniteTimeSpan || oldToken.CanBeCanceled)
                 {
                     cts.Token.Register(state => ((Stream?)state)?.Dispose(), responseContentStream);
                 }
@@ -96,19 +98,8 @@ namespace Azure.Core.Pipeline
                     bufferedStream.Position = 0;
                     message.Response.ContentStream = bufferedStream;
                 }
-                // We dispose stream on timeout so catch and check if cancellation token was cancelled
-                catch (ObjectDisposedException ex)
-                {
-                    ThrowIfCancellationRequestedOrTimeout(oldToken, cts.Token, ex, networkTimeout);
-                    throw;
-                }
-                // We dispose stream on timeout so catch and check if cancellation token was cancelled
-                catch (IOException ex)
-                {
-                    ThrowIfCancellationRequestedOrTimeout(oldToken, cts.Token, ex, networkTimeout);
-                    throw;
-                }
-                catch (OperationCanceledException ex)
+                // We dispose stream on timeout or user cancellation so catch and check if cancellation token was cancelled
+                catch (Exception ex) when (ex is ObjectDisposedException or IOException or OperationCanceledException or NotSupportedException)
                 {
                     ThrowIfCancellationRequestedOrTimeout(oldToken, cts.Token, ex, networkTimeout);
                     throw;
--- a/sdk/core/Azure.Core/src/Pipeline/Internal/ResponseBodyPolicy.cs
+++ b/sdk/core/Azure.Core/src/Pipeline/Internal/ResponseBodyPolicy.cs
@@ -99,7 +99,11 @@ namespace Azure.Core.Pipeline
                     message.Response.ContentStream = bufferedStream;
                 }
                 // We dispose stream on timeout or user cancellation so catch and check if cancellation token was cancelled
-                catch (Exception ex) when (ex is ObjectDisposedException or IOException or OperationCanceledException or NotSupportedException)
+                catch (Exception ex)
+                    when (ex is ObjectDisposedException
+                              or IOException
+                              or OperationCanceledException
+                              or NotSupportedException)
                 {
                     ThrowIfCancellationRequestedOrTimeout(oldToken, cts.Token, ex, networkTimeout);
                     throw;
--- a/sdk/core/Azure.Core/src/Pipeline/Internal/ResponseBodyPolicy.cs
+++ b/sdk/core/Azure.Core/src/Pipeline/Internal/ResponseBodyPolicy.cs
@@ -75,8 +75,8 @@ namespace Azure.Core.Pipeline
 
             if (message.BufferResponse)
             {
-                // If there is a network timeout or the user passed a cancellation token, then we will dispose the stream when
-                // cancellation is triggered.
+                // If cancellation is possible (whether due to network timeout or a user cancellation token being passed), then
+                // register callback to dispose the stream on cancellation.
                 if (networkTimeout != Timeout.InfiniteTimeSpan || oldToken.CanBeCanceled)
                 {
                     cts.Token.Register(state => ((Stream?)state)?.Dispose(), responseContentStream);
--- a/sdk/core/Azure.Core/tests/ResponseBodyPolicyTests.cs
+++ b/sdk/core/Azure.Core/tests/ResponseBodyPolicyTests.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Runtime.CompilerServices;
 using System.Threading;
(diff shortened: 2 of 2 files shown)


sdk/core/Azure.Core/src/Pipeline/Internal/ResponseBodyPolicy.cs
```diff
@@ -96,19 +98,8 @@ private async ValueTask ProcessAsync(HttpMessage message, ReadOnlyMemory<HttpPip
...
-                }
-                // We dispose stream on timeout so catch and check if cancellation token was cancelled
-                catch (IOException ex)
-                {
-                    ThrowIfCancellationRequestedOrTimeout(oldToken, cts.Token, ex, networkTimeout);
-                    throw;
-                }
-                catch (OperationCanceledException ex)
+                // We dispose stream on timeout or user cancellation so catch and check if cancellation token was cancelled
+                catch (Exception ex) when (ex is ObjectDisposedException or IOException or OperationCanceledException or NotSupportedException)
```
annelo-msft, line 102: Super nit: any option to break this into multiple lines for easier reading/maintenance?

sdk/core/Azure.Core/src/Pipeline/Internal/ResponseBodyPolicy.cs
```diff
@@ -75,7 +75,9 @@ private async ValueTask ProcessAsync(HttpMessage message, ReadOnlyMemory<HttpPip
 
             if (message.BufferResponse)
             {
-                if (networkTimeout != Timeout.InfiniteTimeSpan)
+                // If cancellation is possible (whether due to network timeout or a user cancellation token being passed), then
+                // register callback to dispose the stream on cancellation.
+                if (networkTimeout != Timeout.InfiniteTimeSpan || oldToken.CanBeCanceled)
```
annelo-msft, line 80: Can you share why this fixes the issue (for my edification) ?  Is it that now that we've registered the callback, the stream will be disposed when the CT is cancelled, so it won't be disposed during read?

JoshLove-msft: There are two issues being fixed here:
1. We weren't translating NotSupportedException into TaskCanceledException
2. We were only registering the callback if the network timeout was not infinite even though the point of this code is to dispose the stream on cancellation.

christothes, line 80: Should we do this to cover any tokens linked to this source?
```suggestion
                if (networkTimeout != Timeout.InfiniteTimeSpan || cts.Token.CanBeCanceled)
```

JoshLove-msft: Nope, that one will always return true because it is a new linked token that we created.

annelo-msft: How did the Dispose concurrent with Read issue work?  And how does this prevent that?

christothes (approved):

sdk/core/Azure.Core/src/Pipeline/Internal/ResponseBodyPolicy.cs
JoshLove-msft: Similar to how we translate ObjectDisposedExceptions when the cancellation token is cancelled, we are going to translate NotSupportedExceptions into TaskCanceledExceptions when the user cancels. So we are not actually preventing the race condition, but rather recovering from it in the same way that we do for ObjectDisposedException. The reason we can't prevent these is we intentionally dispose the stream on cancellation to make a best effort attempt to get the processing to stop as soon as the user cancels.

annelo-msft (approved):

sdk/core/Azure.Core/src/Pipeline/Internal/ResponseBodyPolicy.cs
JoshLove-msft: Correct - we throw what would be the expected exception under this scenario because the cancellation token has been cancelled.

PR #33001, opened 2022-12-11
Replace ClientDiagnostics-derived classes with RequestFailedDetailsParser implementations

AlexanderSher:
This PR is a fix for https://github.com/Azure/azure-sdk-for-net/issues/27490

It replaces all types derived from `ClientDiagnostics` that override `ExtractFailureContent` with `RequestFailedDetailsParser` implementations and removes `ExtractFailureContent` itself to avoid new overrides


sdk/core/Azure.Core/src/Shared/ClientDiagnostics.cs
```diff
@@ -101,16 +86,23 @@ internal static HttpMessageSanitizer CreateMessageSanitizer(DiagnosticsOptions d
 
         public async ValueTask<RequestFailedException> CreateRequestFailedExceptionAsync(Response response, ResponseError? error = null, IDictionary<string, string>? additionalInfo = null, Exception? innerException = null)
         {
+            if (GetType() == typeof(ClientDiagnostics) && error is null && additionalInfo is null)
+            {
+                return new RequestFailedException(response, innerException);
```
christothes, line 91: Do we still need all the helper methods at the bottom of this file?

christothes (approved):

sdk/core/Azure.Core/src/RequestFailedException.cs
```diff
@@ -168,16 +168,12 @@ internal static (string FormattedError, string? ErrorCode, IDictionary<string, s
                 }
             }
 
-            if (response.ContentStream is MemoryStream)
+            if (response.ContentStream is MemoryStream && ContentTypeUtilities.TryGetTextEncoding(response.Headers.ContentType, out Encoding _))
```
annelo-msft, line 171: What does this check accomplish if we are throwing the encoding value away?

AlexanderSher: To be sure that we are not trying to decode a non-text. Encoding itself isn't important here, cause `BinaryData` only supports UTF8.

sdk/core/Azure.Core/src/Shared/ClientDiagnostics.cs
```diff
@@ -101,16 +86,23 @@ internal static HttpMessageSanitizer CreateMessageSanitizer(DiagnosticsOptions d
 
         public async ValueTask<RequestFailedException> CreateRequestFailedExceptionAsync(Response response, ResponseError? error = null, IDictionary<string, string>? additionalInfo = null, Exception? innerException = null)
         {
+            if (GetType() == typeof(ClientDiagnostics) && error is null && additionalInfo is null)
```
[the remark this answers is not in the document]
AlexanderSher: I want to remove the whole method at the final stage after changes are made in autorest.csharp

JoshLove-msft (approved):

seanmcc-msft:
It appears that all the client-side encryption unit tests are failing.

jaschrep-msft (approved):
Looks good from the storage side

AlexanderSher:
> It appears that all the client-side encryption unit tests are failing.

Yes, but they are failing on `storage Preview macos11_NET60` configuration since 11/21

jaschrep-msft:
> It appears that all the client-side encryption unit tests are failing.

This is a macos specific issue regarding setting up GCM support on a test machine.

However, file share livetests seem to be failing regardless. `Azure.Storage.Files.Shares.Tests.ShareClientTests.DeleteIfExistsAsync_Error` and `Azure.Storage.Files.Shares.Tests.ShareClientTests.ExistsAsync_Error` are failing in all pipelines.

PR #33014, opened 2022-12-12
[Event Hubs] Include Inner Exception Details

### LINKED ISSUE #33009: [BUG] Azure.Messaging.EventHubs EventHubsException.ToString() does not outputs the innerException.
### Library name and version

Azure.Messaging.EventHubs.Processor 5.7.5 

### Describe the bug

Azure.Messaging.EventHubs EventHubsException.ToString() overrides the default System.Exception.ToString() and does not outputs the InnerException details.

### Expected behavior

To output the InnerException details.

### Actual behavior

Only this is output, and no InnerException detail.

Azure.Messaging.EventHubs.EventHubsException(GeneralError): An error was encountered while executing developer-provided code to process events.  On most hosts, this will fault the task responsible for partition processing, causing it to be restarted from the last checkpoint.  On some hosts, it may crash the process.  It is very strongly advised that all developer-provided code include a try/catch wrapper and ensure that no exceptions are allowed to propagate up the stack.  Exception details can be found in the inner exception. (vdqgadeviothub).  For troubleshooting information, see https://aka.ms/azsdk/net/eventhubs/exceptions/troubleshoot

### Reproduction Steps

Throw a Azure.Messaging.EventHubs EventHubsException with an InnerException

### Environment

.NET 6.0

jsquire:
# Summary

The focus of these changes is to include the details of any inner exception as part of the `ToString` output of `EventHubsException`.

# References and Related

- [[Event Hubs] Include inner exception output for EventHubsException.ToString() (#33009)](https://github.com/Azure/azure-sdk-for-net/issues/33009)

### DIFF
--- a/sdk/eventhub/Azure.Messaging.EventHubs/src/EventHubsException.cs
+++ b/sdk/eventhub/Azure.Messaging.EventHubs/src/EventHubsException.cs
@@ -201,7 +201,17 @@ namespace Azure.Messaging.EventHubs
         /// <returns>A <see cref="System.String" /> that represents this instance.</returns>
         ///
         public override string ToString() =>
-            $"{ typeof(EventHubsException).FullName }({ Reason }): { Message }{ Environment.NewLine }{ StackTrace }";
+            $"{ typeof(EventHubsException).FullName }({ Reason }): { Message }{ Environment.NewLine }{ StackTrace }{ FormatInnerException() }";
+
+        /// <summary>
+        ///   Formats the <see cref="Exception.InnerException"/> for inclusion in the <see cref="ToString" />
+        ///   details.
+        /// </summary>
+        ///
+        /// <returns>The text to include for the inner exception, if any.</returns>
+        ///
+        private string FormatInnerException() =>
+            (InnerException == null) ? string.Empty : $"{ Environment.NewLine }{ InnerException }";
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


christothes (approved):

sdk/eventhub/Azure.Messaging.EventHubs/src/EventHubsException.cs
```diff
@@ -201,7 +201,17 @@ public override string Message
         /// <returns>A <see cref="System.String" /> that represents this instance.</returns>
         ///
         public override string ToString() =>
-            $"{ typeof(EventHubsException).FullName }({ Reason }): { Message }{ Environment.NewLine }{ StackTrace }";
+            $"{ typeof(EventHubsException).FullName }({ Reason }): { Message }{ Environment.NewLine }{ StackTrace }{ FormatInnerException() }";
```
JoshLove-msft, line 204: I think [Exception.ToString](https://github.com/microsoft/referencesource/blob/master/mscorlib/system/exception.cs#L450) already does this. Why do we not just override Message?

jsquire: I don't recall why we didn't originally include the reason as a suffix added to the message.  In hindsight, that may have been a better idea.   At this point, I don't want to change the current format so I'm going to leave this as-is.

PR #33033, opened 2022-12-13
Chunk Handler isolating bytesTransferred updates / Channels exiting gracefully. 

amnguye:


sdk/storage/Azure.Storage.DataMovement/src/CommitChunkHandler.cs
```diff
@@ -100,28 +135,63 @@ private async Task CommitBlockEvent(StageChunkEventArgs args)
...
+                    {
+                        await _currentBytesSemaphore.WaitAsync(_cancellationToken).ConfigureAwait(false);
+                    }
+                    catch (OperationCanceledException)
+                    {
+                        // We should not continue if waiting on the semaphore has cancelled out.
+                        return;
+                    }
+
+                    Interlocked.Add(ref _bytesTransferred, args.BytesTransferred);
```
jalauzon-msft, line 165: Take my comment here with a grain of salt since again, not too familiar with .NET concurrency topics but, I'm wondering if we need the Semaphore, Interlocked Add, and the channel all for safely updating `_bytesTransferred`?

Since this is a single reader channel, would there only be one thread executing this code at once? More curious, it's probably fine to leave all of them to be extra safe.

jalauzon-msft (approved):

sdk/storage/Azure.Storage.DataMovement/src/CommitChunkHandler.cs
amnguye: https://github.com/Azure/azure-sdk-for-net/issues/33049

PR #33035, opened 2022-12-13
Fix bug involving MaxConcurrentCallsPerSession setting

JoshLove-msft:


sdk/servicebus/Azure.Messaging.ServiceBus/tests/Processor/SessionProcessorLiveTests.cs
```diff
@@ -1714,6 +1715,62 @@ async Task ProcessMessage(ProcessSessionMessageEventArgs args)
...
+                    // add a delay to simulate processing
+                    await Task.Delay(100);
+                }
+
+                await tcs.Task;
+                await processor.StopProcessingAsync();
+                stopWatch.Stop();
+
+                Assert.AreEqual(totalMessages, messageCt);
+                Assert.Less(stopWatch.Elapsed, TimeSpan.FromSeconds(10));
```
JoshLove-msft, line 1770: I verified that without the fix this test fails.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessor.cs
```diff
@@ -836,10 +836,6 @@ private async Task RunReceiveTaskAsync(CancellationToken cancellationToken)
                         {
                             await _messageHandlerSemaphore.WaitAsync(linkedHandlerTcs.Token).ConfigureAwait(false);
                             messageHandlerLockAcquired = true;
-                            if (IsSessionProcessor)
```
JoshLove-msft, line 839: This was causing us to wait unnecessarily when MaxCallsPerSession > 1 but there were outstanding accept session calls in progress.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/SessionReceiverManager.cs
```diff
@@ -127,8 +127,11 @@ private async Task CreateAndInitializeSessionReceiver(CancellationToken processo
 
         private async Task CreateReceiver(CancellationToken processorCancellationToken)
         {
+            bool releaseSemaphore = false;
             try
             {
+                await _concurrentAcceptSessionsSemaphore.WaitAsync(processorCancellationToken).ConfigureAwait(false);
```
jsquire, line 133: Does it make sense to attempt to acquire synchronously here first with a `Wait(0)` and then fall back to async?

jsquire (approved):

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/SessionReceiverManager.cs
JoshLove-msft: Yeah good catch - I can also add this back to the other semaphore wait in ServiceBusProcessor.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessor.cs
```diff
@@ -836,10 +836,6 @@ private async Task RunReceiveTaskAsync(CancellationToken cancellationToken)
                         {
                             await _messageHandlerSemaphore.WaitAsync(linkedHandlerTcs.Token).ConfigureAwait(false);
                             messageHandlerLockAcquired = true;
```
JoshLove-msft, line 838: This is unnecessary since it would never be true in the one place we are checking it - the catch block.

PR #33038, opened 2022-12-13
JsonData Updates: Add WriteTo(), ==operators, and PR FB

annelo-msft:
This is a follow-up to https://github.com/Azure/azure-sdk-for-net/pull/32613 that addresses:

- [x] Support for writing to a stream: https://github.com/Azure/azure-sdk-for-net/pull/32613#discussion_r1038652662
- [x] Adding equality operators to match cast operators: https://github.com/Azure/azure-sdk-for-net/pull/32613#discussion_r1047430154
- [x] Misc test clean-up and other missed PR feedback.

### DIFF
--- a/sdk/core/Azure.Core.Experimental/src/DynamicData.cs
+++ b/sdk/core/Azure.Core.Experimental/src/DynamicData.cs
@@ -1,0 +1,26 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Text.Json;
+
+namespace Azure.Core.Dynamic
+{
+    /// <summary>
+    /// A dynamic abstraction over content data that may have an underlying
+    /// format of JSON or other format.
+    /// </summary>
+    public abstract class DynamicData
+    {
+        /// <summary>
+        /// Writes the data to the provided writer as a JSON value.
+        /// </summary>
+        /// <param name="writer">The writer to which to write the document.</param>
+        /// <param name="data">The dynamic data value to write.</param>
+        public static void WriteTo(Utf8JsonWriter writer, DynamicData data)
+        {
+            data.WriteTo(writer);
+        }
+
+        internal abstract void WriteTo(Utf8JsonWriter writer);
+    }
+}
--- a/sdk/core/Azure.Core.Experimental/src/DynamicData.cs
+++ b/sdk/core/Azure.Core.Experimental/src/DynamicData.cs
@@ -6,8 +6,10 @@ using System.Text.Json;
 namespace Azure.Core.Dynamic
 {
     /// <summary>
-    /// A dynamic abstraction over content data that may have an underlying
-    /// format of JSON or other format.
+    /// A dynamic abstraction over content data.  Deriving types are implemented
+    /// for a specific format, such as JSON or XML.
+    ///
+    /// This and related types are not intended to be mocked.
     /// </summary>
     public abstract class DynamicData
     {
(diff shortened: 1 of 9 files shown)


sdk/core/Azure.Core.Experimental/src/DynamicData.cs
```diff
@@ -0,0 +1,26 @@
...
+        /// Writes the data to the provided writer as a JSON value.
+        /// </summary>
+        /// <param name="writer">The writer to which to write the document.</param>
+        /// <param name="data">The dynamic data value to write.</param>
+        public static void WriteTo(Utf8JsonWriter writer, DynamicData data)
+        {
+            data.WriteTo(writer);
+        }
+
+        internal abstract void WriteTo(Utf8JsonWriter writer);
```
JoshLove-msft, line 24: Why is this internal?

annelo-msft: I didn't want it to be visible in the public API, because we are trying to make JsonData have no public APIs (with the exception of operator overloads) so everything must be accessed through the dynamic interface.

sdk/core/Azure.Core.Experimental/src/DynamicData.cs
```diff
@@ -0,0 +1,26 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Text.Json;
+
+namespace Azure.Core.Dynamic
+{
+    /// <summary>
+    /// A dynamic abstraction over content data that may have an underlying
+    /// format of JSON or other format.
```
jsquire, line 10:
```suggestion
    /// A dynamic abstraction over content data that may have different underlying
    /// formats, such as JSON or XML.
```

sdk/core/Azure.Core.Experimental/src/DynamicData.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// </summary>
+    public abstract class DynamicData
+    {
+        /// <summary>
+        /// Writes the data to the provided writer as a JSON value.
+        /// </summary>
+        /// <param name="writer">The writer to which to write the document.</param>
+        /// <param name="data">The dynamic data value to write.</param>
+        public static void WriteTo(Utf8JsonWriter writer, DynamicData data)
```
jsquire, line 19: How can this be mocked for testing?   The combination of static here and internal for the abstract method seems to signal a testability issue for customers.

If our intention is that this class should not be extended outside of Azure.Core, we should probably call that out loudly in the class-level comments.   Inevitably, someone will open an issue about this.   😄

jsquire (approved):

sdk/core/Azure.Core.Experimental/src/DynamicData.cs
annelo-msft: Oh wow, that is interesting!  I'll update the comments for now to say this type is not mockable, but we can return to it if needed either in Experimental or when we move it to Core proper.  I'm not sure I see a need to mock this type, especially since most of the APIs are internal, but I'm interested in understanding reasons we would want to.  How do we think about mocking Core types in general?  The scenarios I think about for mocking Azure SDK types are primarily around clients and models.

jsquire: I don't disagree with our intent, but based on what I've seen in issue triage - for any given public type, at some point an issue will be opened asking how to mock it.

annelo-msft: Cool - I added the comment to proactively address this.  Thanks!

sdk/core/Azure.Core.Experimental/src/DynamicData.cs
annelo-msft: If you have a specific concern, I'm happy to discuss, and will address in a later PR.

PR #33040, opened 2022-12-13
Prototype making DelayStrategy public

JoshLove-msft:
Attempts to unify the DelayStrategy into something that can be used in both LROs and custom Retry policies. 

ServerDelayHint is the retry-after header if populated.

Additional background - https://github.com/Azure/azure-sdk/issues/3841

**User Scenarios that we have heard of:**
- Customize exponential delay (https://github.com/Azure/azure-sdk-for-net/issues/34611)
  - We are not exposing ExponentialDelayStrategy type, so this is not as simple as overriding that type, but this is achievable using a custom DelayStrategy/RetryPolicy.
- Per-method retry customization
  - Users can do this with a custom DelayStrategy/RetryPolicy that looks at the Request.Uri to determine what the retry logic should be.

Additional notes about the existing implementation (prior to this PR):
- The current ConstantDelayStrategy is used as the fallback strategy for LROs. Right now this is only used for dataplane LROs. Mgmt LROs specify the ExponentialDelayStrategy (renamed to SequentialDelayStrategy in this PR), so the fallback is not used. The ExponentialDelayStrategy ignores suggestedInterval, whereas the ConstantDelayStrategy respects it. **Should we be consistent here and apply the suggestedInterval (aka clientDelayHint) in all strategies?** 
- All LRO strategies are currently wrapped with the RetryAfterStrategy which makes it so the Retry-After header is always applied and not overridable within a strategy passed to OperationPoller. **Should users be able to override this for either LRO or retries?**
- Fixed retry mode and ConstantDelayStrategy as well as what was formerly called ExponentialDelayStrategy do NOT use jitter. Only the Exponential mode retry uses jitter. **Should all strategies use jitter by default? Should we allow them to opt out of using jitter?**
- The DelayStrategy GetNextDelay signatures use `Response` rather than `HttpMessage` because the existing LRO API uses `Response`. This is a problem if users want to store state in the message - there is no way to do this on the response without mutating headers. In HttpMessage there are Set/GetProperties methods they can use. We can consider adding the same to Response.


sdk/core/Azure.Core/src/Pipeline/RetryPolicy.cs
```diff
@@ -283,6 +286,11 @@ private TimeSpan CalculateNextDelayInternal(HttpMessage message)
                 delay = serverDelay;
             }
 
+            if (_delayStrategy != null)
```
christothes, line 289: Should `CalculateNextDelayInternal` be implemented as a default `DelayStrategy`? If so, we could set it in the ctor based on the RetryOptions (`ConstantDelayStrategy`, `ExponentialDelayStrategy`, etc)

sdk/core/Azure.Core/src/ClientOptions.cs
```diff
@@ -99,6 +99,11 @@ public HttpPipelineTransport Transport
         /// </summary>
         public HttpPipelinePolicy? RetryPolicy { get; set; }
 
+        /// <summary>
+        /// The strategy to use for all delays.
+        /// </summary>
+        public DelayStrategy? Delay { get; set; }
```
annelo-msft, line 105: Is it only delays in the retry policy?  Would it make sense to put it on RetryOptions instead of at the top level?  Would it make sense to call it RetryDelayStrategy since Delay is [pretty heavily used](https://learn.microsoft.com/en-us/dotnet/api/?term=delay) in .NET APIs?

sdk/core/Azure.Core/src/DelayStrategy.cs
```diff
@@ -10,18 +10,25 @@ namespace Azure.Core
...
         /// Get the interval of next delay iteration.
         /// </summary>
         /// <remarks> Note that the value could change per call. </remarks>
         /// <param name="response"> Server response. </param>
+        /// <param name="attempt"></param>
         /// <param name="suggestedInterval"> Suggested pollingInterval. It is up to strategy
         /// implementation to decide how to deal with this parameter. </param>
         /// <returns> Delay interval of next iteration. </returns>
-        public abstract TimeSpan GetNextDelay(Response response, TimeSpan? suggestedInterval);
+        public abstract TimeSpan GetNextDelay(Response response, int attempt, TimeSpan? suggestedInterval);
```
annelo-msft, line 24: nit: should this be RetryNumber to match the public API on ProcessingContext?

sdk/core/Azure.Core/src/Shared/RetryAfterDelayStrategy.cs
```diff
@@ -34,9 +34,10 @@ public RetryAfterDelayStrategy(DelayStrategy? fallbackStrategy = null)
         /// If retry-after header is not found, adopt the value of last retry-after header.
         /// </summary>
         /// <param name="response">Service response which might carry retry-after header.</param>
+        /// <param name="attempt"></param>
```
annelo-msft, line 37: Would the shared implementations move into the public API as well?  Or would they be sealed and made internal and moved out of Shared?

sdk/core/Azure.Core/src/ClientOptions.cs
JoshLove-msft: No this could be used for more than just retries, e.g. LROs.

sdk/core/Azure.Core/src/Pipeline/RetryPolicy.cs
JoshLove-msft: Yeah it can be - I was mostly just trying to get it to build to generate an API

sdk/core/Azure.Core/src/DelayStrategy.cs
JoshLove-msft: This would not be retry specific

sdk/core/Azure.Core/src/Shared/RetryAfterDelayStrategy.cs
JoshLove-msft: They would still need to be shared since they are still in other libraries.

sdk/core/Azure.Core/src/ClientOptions.cs
JoshLove-msft: I think we also want to avoid putting it on RetryOptions because that is meant to be protocol-agnostic. We could probably have the dependency go the other way - you can pass in a RetryOptions when constructing the Delay.

sdk/core/Azure.Core/src/Shared/RetryAfterDelayStrategy.cs
annelo-msft: What would our longer-term strategy be for them?

sdk/core/Azure.Core/src/DelayStrategy.cs
annelo-msft: Would there ever be a scenario where we wanted delays in LRO polling to have a different strategy than retry delays in the same client?

JoshLove-msft: Yeah I think that is possible. We may want to call the options property `RetryStrategy` and potentially have a separate `LroStrategy`.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -395,6 +395,13 @@ public abstract partial class ClientOptions
         public static bool operator !=(Azure.Core.ContentType left, Azure.Core.ContentType right) { throw null; }
         public override string ToString() { throw null; }
     }
+    public abstract partial class DelayStrategy
+    {
+        protected DelayStrategy() { }
+        public abstract System.TimeSpan GetNextDelay(Azure.Response response, int attempt, System.TimeSpan? suggestedInterval);
+        protected internal static System.TimeSpan GetServerDelay(Azure.Core.HttpMessage message) { throw null; }
+        protected static System.TimeSpan Max(System.TimeSpan t1, System.TimeSpan t2) { throw null; }
```
JoshLove-msft, line 403: This can probably be removed.

sdk/core/Azure.Core/src/DelayStrategy.cs
JoshLove-msft: For now I'm limiting the DelayStrategy to being exposed in RetryPolicy constructor. We can consider adding additional properties in ClientOptions for LRO and Retry to be used outside of a custom RetryPolicy.

sdk/core/Azure.Core/src/Shared/DelayStrategy.cs
```diff
@@ -1,27 +0,0 @@
...
-#nullable enable
-
-using System;
-
-namespace Azure.Core
-{
-    /// <summary>
-    /// Strategy to control delay behavior.
-    /// </summary>
-    internal abstract class DelayStrategy
```
christothes, line 13: You'll have to fixup all these references https://grep.app/search?q=DelayStrategy&filter[repo][0]=Azure/azure-sdk-for-net&filter[lang][0]=XML

This may also affect the generator, but not sure.

sdk/core/Azure.Core/src/Shared/RetryAfterDelayStrategy.cs
```diff
@@ -34,9 +34,10 @@ public RetryAfterDelayStrategy(DelayStrategy? fallbackStrategy = null)
...
         /// <param name="suggestedInterval">Suggested pollingInterval.</param>
         /// <returns>Max value of retry-after header and <paramref name="suggestedInterval"/>.</returns>
-        public override TimeSpan GetNextDelay(Response response, TimeSpan? suggestedInterval)
+        public override TimeSpan GetNextDelay(Response response, int attempt, TimeSpan? suggestedInterval)
```
christothes, line 40: Can this just call `GetServerDelay` in the base now?

sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -395,6 +395,13 @@ public abstract partial class ClientOptions
         public static bool operator !=(Azure.Core.ContentType left, Azure.Core.ContentType right) { throw null; }
         public override string ToString() { throw null; }
     }
+    public abstract partial class DelayStrategy
+    {
+        protected DelayStrategy() { }
```
JoshLove-msft, line 400: Thoughts on the naming here - DelayStrategy vs DelayPolicy vs something else?

/cc @christothes @annelo-msft

christothes: I like the name as is

sdk/core/Azure.Core/src/DelayStrategy.cs
```diff
@@ -0,0 +1,75 @@
...
+        public abstract TimeSpan GetNextDelay(Response response, int attempt, TimeSpan? suggestedDelay);
+
+        /// <summary>
+        /// Gets the server specified delay. If the message has no response, <see cref="TimeSpan.Zero"/> is returned.
+        /// This method can be used to help calculate the next delay when overriding <see cref="GetNextDelay"/>, i.e.
+        /// implementors may want to add the server delay to their own custom delay.
+        /// </summary>
+        /// <param name="message">The message to inspect for the server specified delay.</param>
+        /// <returns>The server specified delay.</returns>
+        protected internal static TimeSpan GetServerDelay(HttpMessage message)
```
JoshLove-msft, line 36: This should probably take in the Response for consistency.

JoshLove-msft: Though I suppose that would mean all callers would need to do the HasResponse check themselves so maybe it is okay to leave this as HttpMessage.

annelo-msft: Methods in our public APIs that take HttpMessage usually do so because we want to be able to pull values from either the request or the response.  Since this depends on a value from the server exclusively, i.e. a response, it would probably make sense to scope it directly to that.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
annelo-msft: I feel like `Policy` is largely taken, given we use those for our pipeline, but I like the way it describes the concept.  Strategy feels a little weird to me - is there any .NET/BCL naming precedent for "a thing that is a placeholder for an implementation"?  "Policy" is kind of that.

sdk/core/Azure.Core/src/DelayStrategy.cs
annelo-msft, line 36: Nit - I just noticed that this method is added and wasn't in the shared type.  Does it make sense to put it here?  Doesn't it only apply to RetryPolicy, and not LROs?  Would anyone ever override it?

sdk/core/Azure.Core/api/Azure.Core.net461.cs
annelo-msft: Actually, this makes me curious - have we already said what the motivation for moving this into the public API is?  (Apologies if I'm forgetting)

JoshLove-msft: Yeah this is covered here - https://github.com/Azure/azure-sdk/issues/3841#issuecomment-1029433201

sdk/core/Azure.Core/src/DelayStrategy.cs
JoshLove-msft: No, this would apply to LROs as well since responses may have the Retry-After header. This isn't intended to be overriden, but rather as a convenience to extract the server delay just like we had initially in the RetryPolicy.

sdk/core/Azure.Core/src/Shared/DelayStrategy.cs
JoshLove-msft: There will need to be some generator changes for this if we replace the shared strategies with new public ones.

sdk/core/Azure.Core/src/Shared/RetryAfterDelayStrategy.cs
JoshLove-msft: I've removed the shared strategies for exponential and constant and replaced with public strategies.

sdk/core/Azure.Core/src/ExponentialDelayStrategy.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// <param name="delayHint"></param>
+        /// <returns></returns>
+        /// <exception cref="NotImplementedException"></exception>
+        public override TimeSpan GetNextDelay(Response? response, int attempt, TimeSpan? delayHint)
+        {
+            return
+                Max(
+                    delayHint ?? TimeSpan.Zero,
+                    Max(
+                        GetServerDelay(response),
```
m-nash, line 44: I would consider flipping this back around.  If done this way we introduce a scenario where customers could write their own delay strategy and bypass the RetryAfter headers.  The RetryAfter headers are there to protect azure services from being spammed more than to help the customer get the fastest response.

Originally we had a RetryAfterDelayStrategy which took a fallback strategy incase it didn't exist or the customer opted into waiting longer.  This could still exist and be internal which would allow the customer to create their own strategy but not introduce a scenario would the customer is spamming our services and causing issues.

sdk/core/Azure.Core/src/ExponentialDelayStrategy.cs
```diff
@@ -0,0 +1,51 @@
...
+                        TimeSpan.FromMilliseconds(
```
m-nash, line 45: Might be slightly more efficient if we treat this as a SequentialDelay that we pre-calculate in the ctor based on the inputs.  This was the idea with the hard coded array in SequentialDelay only difference is this could vary based on ctor arguments.

sdk/core/Azure.Core/src/Shared/OperationInternalOfT.cs
```diff
@@ -91,7 +91,8 @@ internal class OperationInternal<T> : OperationInternalBase
         /// parameter <paramref name="operation"/>.
         /// </param>
         /// <param name="scopeAttributes">The attributes to use during diagnostic scope creation.</param>
-        /// <param name="fallbackStrategy">The fallback delay strategy when Retry-After header is not present.  When it is present, the longer of the two delays will be used. Default is <see cref="ConstantDelayStrategy"/>.</param>
+        /// <param name="fallbackStrategy">The fallback delay strategy when Retry-After header is not present.  When it is present, the longer of the two delays will be used.
+        /// Default is <see cref="FixedDelayStrategy"/>.</param>
```
m-nash, line 95: Should we add an overload `WaitForCompletion(DelayStrategy delayStrategy, CancellationToken cancellationToken)` which allows users to pass in their stategy in an LRO scenario?

sdk/core/Azure.Core/src/Shared/DelayStrategy.cs
m-nash: Might need to leave these here temporarily until we can make the autorest.csharp changes then can remove in a follow up

sdk/core/Azure.Core/src/ExponentialDelayStrategy.cs
JoshLove-msft: This will depend on whether we want to allow customers to bypass the retry-after headers or not. The approach taken here is to provide an API that implementors can use, `GetServerDelay`, if they want to override GetNextDelay but still respect the server delay. If we don't want to allow this level of control, then I agree with your proposed approach.

@KrzysztofCwalina, @tg-msft what do you think about this issue? Should customers be allowed to bypass the retry-after header, or should they be limited to Max(Retry-After, CustomDelay)?

JoshLove-msft: @m-nash provided more context around the LRO use case. The impact of a customer ignoring the server delay in a custom strategy is larger for LROs because we expect to poll the server multiple times for each LRO. For retries, it is a bit different because retry-after header would only come into play for error scenarios.

sdk/core/Azure.Core/src/DelayStrategy.cs
JoshLove-msft: I've added an overload to `Operation<T>` based on @m-nash's suggestion. We could also consider a client option specific to LROs.

sdk/core/Azure.Core/src/ExponentialDelayStrategy.cs
JoshLove-msft: Pinging @KrzysztofCwalina @tg-msft, what do you think the ideal behavior should be for the retry case? Should we allow users to ignore the server delay or would it be better to always wrap their delay strategy so that we use the max returned by their strategy and the retry header.

KrzysztofCwalina: I think we cannot prevent customers from ignoring the header. They can write their own client after all, or they can change the retry policy to only do one try and then "retry" in the client calling code. 

But I do agree that it would be good to design the extensibility APIs such that the user has to explicitly ignore the header as opposed to just not knowing that they should honor it. Maybe we can add "suggestedRetryAfter" parameter to the extensibility APIs.

JoshLove-msft: Just to clarify - @m-nash is suggesting that we don't allow customers to ignore the header for LRO scenarios, because the penalty of doing so would be quite severe. Are you alright with us diverging between LRO and retries?

christothes: One difference between general retries and LROs is that Retries usually have a finite retry limit. I think that makes ignoring the header less impactful to the service. That said, I agree that the default should be to respect to header with some way to explicitly override it.

JoshLove-msft: Yes, also retries only occur in error scenarios, where trying again is expected for LROs. The default is to respect the header, but we are discussing whether it should be possible to override the header for LROs and for retries. @m-nash has stated that it should not be overrideable for LROs.

KrzysztofCwalina: Do we have two different policies for LROS and not LROS?

JoshLove-msft: We don't exactly have policies for LROs, but we do have delay strategies (currently internal). The delay strategies are applied outside of the pipeline, whereas retry policy is within the policy. So an LRO would use the configured retry policy for individual requests, but between requests it would use the LRO delay strategy.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -398,6 +400,13 @@ public abstract partial class ClientOptions
...
+        public abstract System.TimeSpan GetNextDelay(Azure.Response? response, int attempt, System.TimeSpan? delayHint);
```
JoshLove-msft, line 406: @KrzysztofCwalina's idea: clientDelayHint and serverDelayHint params. clientDelayHint would be based on user specified delay/retry policy calculation. serverDelayHint would be based on retry-after header. Since users can always call Update themselves in a loop for LROs, they can always choose to bypass the retry-after header, so it may be sufficient to just present them with the value and let them decide what to do.

JoshLove-msft: @m-nash thoughts on the above?

sdk/core/Azure.Core/api/Azure.Core.net5.0.cs
```diff
@@ -398,6 +400,16 @@ public abstract partial class ClientOptions
         public static bool operator !=(Azure.Core.ContentType left, Azure.Core.ContentType right) { throw null; }
         public override string ToString() { throw null; }
     }
+    public abstract partial class DelayStrategy
+    {
+        protected DelayStrategy(Azure.Core.RetryMode? mode = default(Azure.Core.RetryMode?), System.TimeSpan? initialDelay = default(System.TimeSpan?), System.TimeSpan? maxDelay = default(System.TimeSpan?)) { }
+        public static Azure.Core.DelayStrategy CreateExponentialDelayStrategy(System.TimeSpan? initialDelay = default(System.TimeSpan?), System.TimeSpan? maxDelay = default(System.TimeSpan?)) { throw null; }
+        public static Azure.Core.DelayStrategy CreateFixedDelayStrategy(System.TimeSpan? delay = default(System.TimeSpan?)) { throw null; }
+        public static Azure.Core.DelayStrategy CreateSequentialDelayStrategy(System.Collections.Generic.IEnumerable<System.TimeSpan>? sequence = null) { throw null; }
+        public System.TimeSpan? GetClientDelayHint(Azure.Response? response, int attempt) { throw null; }
```
JoshLove-msft, line 409: Should we add a GetServerDelayHint for symmetry? Right now this is being made public with ResponseHeaders.RetryAfter.

sdk/core/Azure.Core/src/DelayStrategy.cs
JoshLove-msft: I ended up changing this to RetryNumber as per @annelo-msft's initial suggestion - it felt odd to  use attempt number given that it would always start at 2 when the overrides are called.

sdk/core/Azure.Core/src/ExponentialDelayStrategy.cs
m-nash: There is a big difference between retry and lro-polling in that during a retry scenario the response is not expected and in most cases can / should be remedied quickly.  In an lro-polling scenario the response is expected and the retry-after protects not only azure services but it protects the customer from accidentally locking their subscription by hitting their api rate limit.  On top of this there is no expectation that this should get "fixed quickly".  For one, nothing is broken to fix, and for two some of these LROs take a very long time to finish.

My suggestion is that when you pass in a DelayStrategy to WaitForCompletion that this is treated as the fallback strategy to the RetryAfterDelayStrategy that exists today.  If someone writes their own rest client to spam themselves into api rate limiting thats on them, but our client shouldn't make this the default path.  Even adding a `useSuggestedRetryAfter` won't necessarily be understood by customers and I can't come up with a single valid reason for anyone to need to do that.

JoshLove-msft: > If someone writes their own rest client to spam themselves into api rate limiting thats on them, but our client shouldn't make this the default path.

This wouldn't be the default path. In order for customers to do this, they'd need to create their own custom delay strategy, i.e. not use one of the out-of-the-box ones for Exponential, Fixed, or Sequential, and in their custom delay strategy they would need to override GetNextDelay and choose to ignore the serverDelayHint.

sdk/core/Azure.Core/api/Azure.Core.net5.0.cs
```diff
@@ -398,6 +400,15 @@ public abstract partial class ClientOptions
         public static bool operator !=(Azure.Core.ContentType left, Azure.Core.ContentType right) { throw null; }
         public override string ToString() { throw null; }
     }
+    public abstract partial class DelayStrategy
+    {
+        protected DelayStrategy() { }
+        public static Azure.Core.DelayStrategy CreateExponentialDelayStrategy(System.TimeSpan? initialDelay = default(System.TimeSpan?), System.TimeSpan? maxDelay = default(System.TimeSpan?), double factor = 2, double minJitterFactor = 0.8, double maxJitterFactor = 1.2) { throw null; }
+        public static Azure.Core.DelayStrategy CreateFixedDelayStrategy(System.TimeSpan? delay = default(System.TimeSpan?)) { throw null; }
```
JoshLove-msft, line 407: All should probably use jitter.

sdk/core/Azure.Core/src/ExponentialDelayStrategy.cs
m-nash: I think the point still stands that they will have no idea what it is the are opting into or out of.

What is the legitimate scenario where a customer needs the ability to spam the service instead of waiting the suggested amount of time?

christothes: > What is the legitimate scenario where a customer needs the ability to spam the service instead of waiting the suggested amount of time?

Bugs in the service that result in bogus retry-after header values is one I can imagine.

sdk/core/Azure.Core/api/Azure.Core.net5.0.cs
```diff
@@ -398,6 +400,15 @@ public abstract partial class ClientOptions
...
+        public static Azure.Core.DelayStrategy CreateSequentialDelayStrategy(System.Collections.Generic.IEnumerable<System.TimeSpan>? sequence = null) { throw null; }
+        public abstract System.TimeSpan GetNextDelay(Azure.Response? response, int retryNumber, System.TimeSpan? clientDelayHint, System.TimeSpan? serverDelayHint);
```
JoshLove-msft, line 409: Flip serverDelay and clientDelayHint order.

potentially introduce protected GetNextDelayCore(no hint parameters)
GetNextDelay would not be abstract => GetNextDelayCore.

sdk/core/Azure.Core/src/ExponentialDelayStrategy.cs
JoshLove-msft: @tg-msft had a suggestion that we introduce GetNextDelayCore as abstract which would not take either hint. We could then have GetNextDelay as virtual which has a default implementation that uses the max of the hints. If customers really want to, they can override this, but it makes it that much more hidden.

sdk/core/Azure.Core/src/DelayStrategy.cs
```diff
@@ -0,0 +1,128 @@
...
+using System;
+using System.Collections.Generic;
+using Azure.Core.Pipeline;
+
+namespace Azure.Core
+{
+    /// <summary>
+    /// Strategy to control delay behavior.
+    /// </summary>
+    public abstract class DelayStrategy
```
annelo-msft, line 15: Curious: Have we given any consideration to whether/how these would apply to RetriableStream if set?

JoshLove-msft: Hmm, it looks like the only overlap would be in the maxRetries as RetriableStream doesn't do any kind of backoff. It looks like all libraries that use RetriableStream hardcode the maxRetries so I don't think we'd need to do anything special.

annelo-msft: Yeah, I noticed it was pretty bare-bones in RetriableStream - I guess the question would be whether if a customer made changes to RetryPolicy configuration details, if they would have the expectation that those would be applied in a stream-retry case as well.  RetriableStream has a small enough footprint that I wouldn't imagine it has come up yet, but it might be something we want to think about as part of the effort to move it out of shared source when the time for that comes.  Thanks!

JoshLove-msft: At least for the current implementations, the retriable stream maxRetries is independent of the user's RetryOptions, so this would already be an existing gap if we end up deciding it should respect the RetryOptions. My worry about correlating them would be if you could somehow get in a situation where N^2 retries are being applied instead of N*M where M is the hard-coded constant. I'd need to dig into the details more to see if this is a concern.

annelo-msft: Oh, interesting.  I'm looking into using RetriableStream for ACR, and was planning to pass in the max retries value from RetryOptions.  I'll have a quick look to see if it looks like that would be an issue...

sdk/core/Azure.Core/src/DelayStrategy.cs
```diff
@@ -0,0 +1,128 @@
...
+        /// <remarks> Note that the value could change per call. </remarks>
+        /// <param name="response"> Server response. </param>
+        /// <param name="retryNumber"></param>
+        /// <param name="serverDelayHint"></param>
+        /// <param name="clientDelayHint"></param>
+        /// <returns> Delay interval of next iteration. </returns>
+        public virtual TimeSpan GetNextDelay(Response? response, int retryNumber, TimeSpan? serverDelayHint, TimeSpan? clientDelayHint)
+        {
+            return
+                Max(
```
JoshLove-msft, line 110: Should be min

sdk/core/Azure.Core/src/Shared/OperationPoller.cs
```diff
@@ -17,49 +17,50 @@ internal sealed class OperationPoller
     {
         private readonly DelayStrategy _delayStrategy;
 
-        public OperationPoller(DelayStrategy? fallbackStrategy = null)
+        public OperationPoller(DelayStrategy? strategy = null)
         {
-            _delayStrategy = new RetryAfterDelayStrategy(fallbackStrategy);
+            _delayStrategy = strategy ?? DelayStrategy.CreateSequentialDelayStrategy();
```
JoshLove-msft, line 22: This should be FixedDelayStrategy.

sdk/core/Azure.Core/src/DelayStrategy.cs
JoshLove-msft: Remove virtual for now - can consider extensible options in the future if needed.

sdk/core/Azure.Core/api/Azure.Core.net5.0.cs
```diff
@@ -398,6 +400,16 @@ public abstract partial class ClientOptions
         public static bool operator !=(Azure.Core.ContentType left, Azure.Core.ContentType right) { throw null; }
         public override string ToString() { throw null; }
     }
+    public abstract partial class DelayStrategy
+    {
+        protected DelayStrategy(System.TimeSpan? maxDelay = default(System.TimeSpan?), double minJitterFactor = 0.8, double maxJitterFactor = 1.2) { }
+        public static Azure.Core.DelayStrategy CreateExponentialDelayStrategy(System.TimeSpan? initialDelay = default(System.TimeSpan?), System.TimeSpan? maxDelay = default(System.TimeSpan?), double factor = 2, double minJitterFactor = 0.8, double maxJitterFactor = 1.2) { throw null; }
```
JoshLove-msft, line 406: Remove factor

JoshLove-msft: Remove jitter

JoshLove-msft: We can ask architects about the jitter value defaults.

JoshLove-msft:
Delay vs DelayStrategy

sdk/core/Azure.Core/src/DelayStrategy.cs
```diff
@@ -0,0 +1,135 @@
...
+        private readonly double _maxJitterFactor;
+        private readonly TimeSpan _maxDelay;
+
+        /// <summary>
+        ///
+        /// </summary>
+        /// <param name="maxDelay"></param>
+        /// <param name="minJitterFactor"></param>
+        /// <param name="maxJitterFactor"></param>
+        protected DelayStrategy(TimeSpan? maxDelay = default, double minJitterFactor = 0.8, double maxJitterFactor = 1.2)
```
christothes, line 29: Do you think allowing fine grained control of min and max jitter is needed, or could we simplify to a single `jitterFactor` representing the factor that would be added or subtracted. In this case, we'd default to "0.2" to indicate "20%".

sdk/core/Azure.Core/src/FixedDelayStrategy.cs
```diff
@@ -0,0 +1,31 @@
...
+    internal class FixedDelayStrategy : DelayStrategy
+    {
+        private readonly TimeSpan _delay;
+
+        /// <summary>
+        ///
+        /// </summary>
+        /// <param name="delay"></param>
+        public FixedDelayStrategy(
+            TimeSpan delay) : base(TimeSpan.FromMilliseconds(delay.TotalMilliseconds * 1.2), 0.8, 1.2)
```
christothes, line 22: Should we be explicit about `MaxDelay` here rather than implicitly applying the jitter?

sdk/core/Azure.Core/src/Pipeline/RetryPolicy.cs
```diff
@@ -14,30 +14,25 @@ namespace Azure.Core.Pipeline
...
+        /// </summary>
+        protected DelayStrategy DelayStrategy => _delayStrategy;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="RetryPolicy"/> class.
         /// </summary>
-        /// <param name="options">The set of options to use for configuring the policy.</param>
-        protected RetryPolicy(RetryOptions? options = default)
+        /// <param name="maxRetries"></param>
+        /// <param name="delayStrategy">The delay strategy to use</param>
```
christothes, line 31:
```suggestion
        /// <param name="delayStrategy">The delay strategy to use. Defaults to <see cref="ExponentialDelayStrategy" /></param>
```

sdk/core/Azure.Core/src/Pipeline/RetryPolicy.cs
```diff
@@ -14,30 +14,25 @@ namespace Azure.Core.Pipeline
...
+        public RetryPolicy(int? maxRetries = default, DelayStrategy? delayStrategy = default)
```
christothes, line 32: Should we just be explicit about the maxRetries default value?
```suggestion
        public RetryPolicy(int maxRetries = 3, DelayStrategy? delayStrategy = default)
```

sdk/core/Azure.Core/src/Pipeline/RetryPolicy.cs
```diff
@@ -127,7 +122,13 @@ private async ValueTask ProcessAsync(HttpMessage message, ReadOnlyMemory<HttpPip
 
                 if (shouldRetry)
                 {
-                    TimeSpan delay = async ? await CalculateNextDelayAsync(message).ConfigureAwait(false) : CalculateNextDelay(message);
+                    if (message.RetryNumber == 0)
+                    {
+                        message.SetProperty(typeof(RetryPolicyPropertiesKey), new Dictionary<string, object?>());
+                    }
+                    message.TryGetProperty(typeof(RetryPolicyPropertiesKey), out object? value);
+                    var context = (IDictionary<string, object?>)value!;
```
christothes, line 130: Is this used anywhere?

sdk/core/Azure.Core/src/Pipeline/RetryPolicy.cs
```diff
@@ -127,7 +122,13 @@ private async ValueTask ProcessAsync(HttpMessage message, ReadOnlyMemory<HttpPip
 
                 if (shouldRetry)
                 {
-                    TimeSpan delay = async ? await CalculateNextDelayAsync(message).ConfigureAwait(false) : CalculateNextDelay(message);
+                    if (message.RetryNumber == 0)
+                    {
+                        message.SetProperty(typeof(RetryPolicyPropertiesKey), new Dictionary<string, object?>());
```
christothes, line 127: Is this intended to be a generic property bag to be used by implementors that is opaque to us?

sdk/core/Azure.Core/src/Shared/ExponentialDelayStrategy.cs
```diff
@@ -13,7 +13,7 @@ namespace Azure.Core
     /// the sequence {1, 1, 1, 2, 4, ...32}.
     /// </summary>
     /// <remarks>Polling interval always follows the given sequence.</remarks>
-    internal class ExponentialDelayStrategy : DelayStrategy
+    internal class SequentialDelayStrategy : DelayStrategy
```
christothes, line 16: Is this intentional? We already have a `SequentialDelayStrategy` below.

sdk/core/Azure.Core.TestFramework/src/Instrumentation/ZeroPollingStrategy.cs
```diff
@@ -13,6 +15,9 @@ namespace Azure.Core
     /// </summary>
     internal class ZeroPollingStrategy : DelayStrategy
     {
-        public override TimeSpan GetNextDelay(Response response, TimeSpan? suggestedInterval) => TimeSpan.Zero;
+        protected override TimeSpan GetNextDelayCore(Response? response, int retryNumber, IDictionary<string, object?> context) => TimeSpan.Zero;
+
+        protected override ValueTask<TimeSpan> GetNextDelayCoreAsync(Response? response, int retryNumber,
+            IDictionary<string, object?> context) => new(TimeSpan.Zero);
```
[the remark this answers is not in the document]
JoshLove-msft: Curious - what would the benefit be? This is already allocation-free.

sdk/core/Azure.Core/src/DelayStrategy.cs
JoshLove-msft: I think that is reasonable.

sdk/core/Azure.Core/src/FixedDelayStrategy.cs
JoshLove-msft: Yeah we probably could do that, but the issue I was trying to avoid was having the jitter be ignored half the time.

sdk/core/Azure.Core/src/Pipeline/RetryPolicy.cs
JoshLove-msft: Yes, it is passed into GetNextDelay

sdk/core/Azure.Core/src/Shared/ExponentialDelayStrategy.cs
JoshLove-msft: Will delete.

sdk/core/Azure.Core/src/FixedDelayStrategy.cs
christothes: Yeah, that is what made me think about how this should be specified. It seems that we should either be explicit or document that the `maxDelay` is always subject to jitter adjustments and expose that with a default.

sdk/core/Azure.Core.TestFramework/src/Instrumentation/ZeroPollingStrategy.cs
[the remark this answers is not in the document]
christothes: Actually, it is surprisingly slower to use a static value!

|         Method |     Mean |     Error |    StdDev | Ratio | RatioSD | Allocated | Alloc Ratio |
|--------------- |---------:|----------:|----------:|------:|--------:|----------:|------------:|
| StaticValuTask | 4.966 ns | 0.0634 ns | 0.0562 ns |  3.63 |    0.08 |         - |          NA |
|    NewValuTask | 1.367 ns | 0.0375 ns | 0.0333 ns |  1.00 |    0.00 |         - |          NA | 

```c#
public class ValueTaskBenchmark
{ 
 	private static ValueTask<TimeSpan> ts = new(TimeSpan.Zero);

	[Benchmark]
	public void StaticValuTask() => FooB();

	[Benchmark(Baseline = true)]
	public void NewValuTask() => FooA();
	
	private ValueTask<TimeSpan> FooA() => new(TimeSpan.Zero);
	private ValueTask<TimeSpan> FooB() => ts;
}
```

sdk/core/Azure.Core/src/FixedDelayStrategy.cs
JoshLove-msft: I changed to keep it maxDelay as absolute.

sdk/core/Azure.Core/api/Azure.Core.net5.0.cs
```diff
@@ -398,6 +400,18 @@ public abstract partial class ClientOptions
         public static bool operator !=(Azure.Core.ContentType left, Azure.Core.ContentType right) { throw null; }
         public override string ToString() { throw null; }
     }
+    public abstract partial class DelayStrategy
```
JoshLove-msft, line 403: I considered using Delay rather than DelayStrategy but I felt it becomes too confusing when passing in TimeSpan delay values and in docs. I'm open to reconsidering, but left as DelayStrategy for now.

JoshLove-msft: /cc @tg-msft @m-nash

sdk/core/Azure.Core/api/Azure.Core.net6.0.cs
```diff
@@ -398,6 +400,18 @@ public abstract partial class ClientOptions
         public static bool operator !=(Azure.Core.ContentType left, Azure.Core.ContentType right) { throw null; }
         public override string ToString() { throw null; }
     }
+    public abstract partial class Delay
```
JoshLove-msft, line 403: This is temporarily renamed to avoid collisions with the internal DelayStrategy which cannot be removed until after this PR is merged, and an Autorest update is made.

sdk/core/Azure.Core/src/Delay.cs
```diff
@@ -0,0 +1,132 @@
...
+        }
+
+        /// <summary>
+        ///
+        /// </summary>
+        /// <param name="response"></param>
+        /// <param name="retryNumber"></param>
+        /// <param name="context"></param>
+        /// <returns></returns>
+        protected abstract TimeSpan GetNextDelayCore(Response? response, int retryNumber, IDictionary<string, object?> context);
```
m-nash, line 69: Instead of the IDictionary<string, object?> here I wonder if we can achieve per set and global state in both LRO and Retry by adding 2 ctors to RetryPolicy

`public RetryPolicy(..., Delay delay = null)`
and
`public RetryPolicy(..., Func<Delay> delayFactory)`

If a user wanted to have per set state in a retry they would just pass in a factory

```c#
options.RetryPolicy = new RetryPolicy(..., () => new MyCustomDelay());
```

I think this gives us a universal way to handle state across retry and lro and lets the customer deal with a typed state vs always needing to cast from `object?` out of the dictionary.

JoshLove-msft: I would be fine with this approach. The factory approach would also be additive so we could wait to add the new ctor if we wanted.

sdk/core/Azure.Core.Experimental/src/Azure.Core.Experimental.csproj
```diff
@@ -20,4 +20,8 @@
     <Compile Include="$(AzureCoreSharedSources)NullableAttributes.cs" />
   </ItemGroup>
 
+  <ItemGroup>
```
annelo-msft, line 23: Curious: what is this used for in this PR?

sdk/core/Azure.Core/api/Azure.Core.net6.0.cs
annelo-msft: Will it go back to being named `DelayStrategy` at that time?  Wondering if we considered `RetryDelay` as a type name, or if the delay is intended to be used more widely than retries?

sdk/core/Azure.Core/api/Azure.Core.net6.0.cs
```diff
@@ -398,6 +400,18 @@ public abstract partial class ClientOptions
...
+        public static Azure.Core.Delay CreateFixedDelay(System.TimeSpan? delay = default(System.TimeSpan?)) { throw null; }
+        public System.TimeSpan GetNextDelay(Azure.Response? response, int retryNumber, System.TimeSpan? serverDelayHint, System.Collections.Generic.IDictionary<string, object?> context) { throw null; }
+        public System.Threading.Tasks.ValueTask<System.TimeSpan> GetNextDelayAsync(Azure.Response? response, int retryNumber, System.TimeSpan? serverDelayHint, System.Collections.Generic.IDictionary<string, object?> context) { throw null; }
+        protected abstract System.TimeSpan GetNextDelayCore(Azure.Response? response, int retryNumber, System.Collections.Generic.IDictionary<string, object?> context);
+        protected abstract System.Threading.Tasks.ValueTask<System.TimeSpan> GetNextDelayCoreAsync(Azure.Response? response, int retryNumber, System.Collections.Generic.IDictionary<string, object?> context);
+        protected static System.TimeSpan Max(System.TimeSpan t1, System.TimeSpan t2) { throw null; }
```
annelo-msft, line 412: Curious if there's a way to move these out of the public API.  They feel like TimeSpan utilities that could be used more broadly than Delay, and might better live in internal shared source?

sdk/core/Azure.Core/api/Azure.Core.net6.0.cs
```diff
@@ -398,6 +400,18 @@ public abstract partial class ClientOptions
         public static bool operator !=(Azure.Core.ContentType left, Azure.Core.ContentType right) { throw null; }
         public override string ToString() { throw null; }
     }
+    public abstract partial class Delay
+    {
+        protected Delay(System.TimeSpan? maxDelay = default(System.TimeSpan?), double jitterFactor = 0.2) { }
+        public static Azure.Core.Delay CreateExponentialDelay(System.TimeSpan? initialDelay = default(System.TimeSpan?), System.TimeSpan? maxDelay = default(System.TimeSpan?)) { throw null; }
+        public static Azure.Core.Delay CreateFixedDelay(System.TimeSpan? delay = default(System.TimeSpan?)) { throw null; }
+        public System.TimeSpan GetNextDelay(Azure.Response? response, int retryNumber, System.TimeSpan? serverDelayHint, System.Collections.Generic.IDictionary<string, object?> context) { throw null; }
```
annelo-msft, line 408: If this is only called by types in Core, does it need to be public?

sdk/core/Azure.Core/api/Azure.Core.net6.0.cs
```diff
@@ -398,6 +400,18 @@ public abstract partial class ClientOptions
         public static bool operator !=(Azure.Core.ContentType left, Azure.Core.ContentType right) { throw null; }
         public override string ToString() { throw null; }
     }
+    public abstract partial class Delay
+    {
+        protected Delay(System.TimeSpan? maxDelay = default(System.TimeSpan?), double jitterFactor = 0.2) { }
```
annelo-msft, line 405: nit: does "factor" come from something specific?  Why not just `jitter`?

sdk/core/Azure.Core/api/Azure.Core.net6.0.cs
```diff
@@ -398,6 +400,18 @@ public abstract partial class ClientOptions
...
+        public static Azure.Core.Delay CreateExponentialDelay(System.TimeSpan? initialDelay = default(System.TimeSpan?), System.TimeSpan? maxDelay = default(System.TimeSpan?)) { throw null; }
+        public static Azure.Core.Delay CreateFixedDelay(System.TimeSpan? delay = default(System.TimeSpan?)) { throw null; }
+        public System.TimeSpan GetNextDelay(Azure.Response? response, int retryNumber, System.TimeSpan? serverDelayHint, System.Collections.Generic.IDictionary<string, object?> context) { throw null; }
+        public System.Threading.Tasks.ValueTask<System.TimeSpan> GetNextDelayAsync(Azure.Response? response, int retryNumber, System.TimeSpan? serverDelayHint, System.Collections.Generic.IDictionary<string, object?> context) { throw null; }
+        protected abstract System.TimeSpan GetNextDelayCore(Azure.Response? response, int retryNumber, System.Collections.Generic.IDictionary<string, object?> context);
```
annelo-msft, line 410: Should the nullable parameters take default values?  I feel like I'm more used to seeing nullable parameters come at the end of the signature.

sdk/core/Azure.Core/api/Azure.Core.net6.0.cs
JoshLove-msft: Yes, it comes from the fact that it is the multiplicative "factor" that is used against the delay to determine the jitter. So 0.2 means that for a delay of 10s, the jittered delay could end up being anywhere from 8s to 12s.

sdk/core/Azure.Core/api/Azure.Core.net6.0.cs
JoshLove-msft: This would also be called by code in shared source (OperationPoller) as well as being usable by customer code in custom policies or RetryPolicy.

sdk/core/Azure.Core/api/Azure.Core.net6.0.cs
annelo-msft: Is there a way to provide them without putting them on the public API?  Are there other places in Core where it might be useful to use the same logic to compare TimeSpans?

JoshLove-msft: Well it depends on what the goal is. If we want customers to have these conveniences (which was my intent) then it would need to be a public API.

JoshLove-msft: Well, I don't think we would end up releasing public utility functions for this, so was hoping to sneak it in here 😆 

I'm not sure if there is a specific guideline here, but am okay with removing them from the public API if that is the consensus.

sdk/core/Azure.Core/api/Azure.Core.net6.0.cs
JoshLove-msft: The reason that Response is nullable is that it is possible that there is no response received from the service. It is still a required parameter because GetNextDelay will always attempt to pass the response if it is there, and implementers expect the value to contain the response, if any.

JoshLove-msft: In terms of moving to the end of the signature, it felt more natural to have it be the first param as it is the primary data users are going to interact with. I suppose it could be swapped with retryNumber though - no strong preference here.

sdk/core/Azure.Core/api/Azure.Core.net6.0.cs
JoshLove-msft: Yeah the final name is probably still TBD. It would be used in LROs as well so my thinking is that it is better to avoid using `Retry` in the name.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -398,6 +400,18 @@ public abstract partial class ClientOptions
...
+        public static Azure.Core.Delay CreateExponentialDelay(System.TimeSpan? initialDelay = default(System.TimeSpan?), System.TimeSpan? maxDelay = default(System.TimeSpan?)) { throw null; }
+        public static Azure.Core.Delay CreateFixedDelay(System.TimeSpan? delay = default(System.TimeSpan?)) { throw null; }
+        public System.TimeSpan GetNextDelay(Azure.Response? response, int retryNumber, System.TimeSpan? serverDelayHint, System.Collections.Generic.IDictionary<string, object?> context) { throw null; }
+        public System.Threading.Tasks.ValueTask<System.TimeSpan> GetNextDelayAsync(Azure.Response? response, int retryNumber, System.TimeSpan? serverDelayHint, System.Collections.Generic.IDictionary<string, object?> context) { throw null; }
+        protected abstract System.TimeSpan GetNextDelayCore(Azure.Response? response, int retryNumber, System.Collections.Generic.IDictionary<string, object?> context);
+        protected abstract System.Threading.Tasks.ValueTask<System.TimeSpan> GetNextDelayCoreAsync(Azure.Response? response, int retryNumber, System.Collections.Generic.IDictionary<string, object?> context);
```
m-nash, line 411: It might be overkill to offer an async version of calculating how long to wait next.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -149,8 +149,10 @@ public abstract partial class Operation<T> : Azure.Operation where T : notnull
         protected Operation() { }
         public abstract bool HasValue { get; }
         public abstract T Value { get; }
+        public virtual Azure.Response<T> WaitForCompletion(Azure.Core.Delay delayStrategy, System.Threading.CancellationToken cancellationToken) { throw null; }
```
m-nash, line 152: Any reason this overload wasn't added to `Operation.WaitForCompletionResponse`?

sdk/core/Azure.Core/api/Azure.Core.net461.cs
JoshLove-msft: Yes, that is fair, but we offer the async version in the RetryPolicy since it derives from HttpPipelinePolicy. If we left the async off for now, we would not be able to be add it as abstract later if we need to, so I would err towards leaving it.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
JoshLove-msft: Will add.

sdk/core/Azure.Core/src/Operation.cs
```diff
@@ -90,6 +90,25 @@ public virtual async ValueTask<Response> WaitForCompletionResponseAsync(TimeSpan
...
+        /// The interval between status requests to the server.
+        /// The interval can change based on information returned from the server.
+        /// For example, the server might communicate to the client that there is not reason to poll for status change sooner than some time.
+        /// </param>
+        /// <param name="cancellationToken">A <see cref="CancellationToken"/> used for the periodical service calls.</param>
+        /// <returns>The last HTTP response received from the server.</returns>
+        /// <remarks>
+        /// This method will periodically call UpdateStatusAsync till HasCompleted is true, then return the final response of the operation.
+        /// </remarks>
+        public virtual async ValueTask<Response> WaitForCompletionResponseAsync(Delay delay, CancellationToken cancellationToken = default)
```
m-nash, line 106: I think we are still missing `WaitForCompletionResponse`?

JoshLove-msft: Woops - added.

m-nash:
@JoshLove-msft I assume we want to track this separately so I opened an issue for it https://github.com/Azure/azure-sdk-for-net/issues/35373

sdk/core/Azure.Core.Experimental/src/Azure.Core.Experimental.csproj
JoshLove-msft: Removed

heaths:
> All LRO strategies are currently wrapped with the RetryAfterStrategy which makes it so the Retry-After header is always applied and not overridable within a strategy passed to OperationPoller. Should users be able to override this for either LRO or retries?

@JoshLove-msft I've often wondered if the service has determine the best value to pass back. Some I've seen just return 10 (10s) and it seems rather random. Perhaps we don't make it incidentally easy to ignore it, but I wonder if the pipeline should at least make it possible if someone wrote their own retry policy to ignore it.

JoshLove-msft:
> @JoshLove-msft not sure if you noticed these. They didn't show up as fatal Checks errors, but looks like something broke inadvertently:
> 
> image

Hmm where are you seeing this?

heaths:
@JoshLove-msft in the Files tab, not the Checks.

JoshLove-msft:
> @JoshLove-msft in the Files tab, not the Checks.

Seems to not be an issue anymore.

JoshLove-msft:
> > All LRO strategies are currently wrapped with the RetryAfterStrategy which makes it so the Retry-After header is always applied and not overridable within a strategy passed to OperationPoller. Should users be able to override this for either LRO or retries?
> 
> @JoshLove-msft I've often wondered if the service has determine the best value to pass back. Some I've seen just return 10 (10s) and it seems rather random. Perhaps we don't make it incidentally easy to ignore it, but I wonder if the pipeline should at least make it possible if someone wrote their own retry policy to ignore it in case a better value works for their workload after their own investigation.

We ended up not allowing users to use a value that is less than the service provided value. If we hear of legitimate use cases we can add this ability in the future.

JoshLove-msft:
Once this is merged, will need to update DownloadSharedSource.ps1 in AutoRest.

sdk/core/Azure.Core/src/Shared/SequentialDelayStrategy.cs
```diff
@@ -1,19 +1,12 @@
...
 
 #nullable enable
 
 namespace Azure.Core
 {
-    /// <summary>
-    /// Polling interval is based on the specified sequence.
-    /// Defaults to {1s, 1s, 1s, 2s, 4s, 8s, 16s, 32s}.
-    /// </summary>
-    /// <remarks>Polling interval always follows the given sequence.</remarks>
```
heaths: Any reason to remove useful documentation? As this is shared source, we SDK devs have to call this. Frankly, I've been a bit annoyed how little internal source is documented. It means I have to pour into source if it's not clear what a member does. It's not helpful to be not helpful.

heaths (approved):

sdk/core/Azure.Core/src/Shared/SequentialDelayStrategy.cs
JoshLove-msft: This documentation is no longer accurate as there is no sequence that you can pass. It is hard-coded to the sequence in the code, so I figured there was no point in duplicating it in two places.

JoshLove-msft: Also, this is used purely by code-gen. Library authors never need to use this.

JoshLove-msft: I went ahead and added some ref docs for the internal strategies.

m-nash (approved):

christothes (approved):

PR #33061, opened 2022-12-15
Filter out package references for the dependency testing for core pipelines

### LINKED ISSUE #32785: [FEATURE REQ] Expand the sparse matrix for core and mgmt core tests
Right now the sparse matrix pins each project to either package reference or project reference.  We saw recently with this issue https://github.com/Azure/azure-sdk-for-net/issues/32733 that this can allow bugs to get checked into main which make debugging and finding the issue a lot more difficult.  It would be better to catch these at PR time so we can fix them and disrupt everyone in the repo.

One option is to expand the parse matrix slightly so that every library runs on both project and package reference variants and it would be very useful if both of those could pick different platforms to take advantage of the free extra coverage since we already need to run it twice.

weshaggard:
Fixes https://github.com/Azure/azure-sdk-for-net/issues/32785


annelo-msft (approved):
LGTM

annelo-msft:
Does this also add back in the project reference pipelines, so we have those to detect breaks?

weshaggard:
> Does this also add back in the project reference pipelines, so we have those to detect breaks?

Yes this should make all the dependency tests use project references and no package references in that set of tests.

benbp (approved):

weshaggard:
Test failure is being fixed by https://github.com/Azure/azure-sdk-for-net/pull/33078 and is unrelated.

PR #33063, opened 2022-12-15
Prototype moving JsonData to be backed by JsonDocument

annelo-msft:
### Summary

In this PR, "JsonData" becomes a collection of types:
- `MutableJsonDocument`, a wrapper around [JsonDocument](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsondocument) that enables mutations to the JSON
- `MutableJsonElement`, the parallel of [JsonElement](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsonelement?view=net-7.0)
- `DynamicJson`, a dynamic layer around `MutableJsonDocument` and `MutableJsonElement`
- `DynamicData`, an abstract class proving the interface for `DynamicJson`, and future types like `DynamicXml`, etc.

Functional features and performance improvements remain to be added, and are tracked in https://github.com/Azure/azure-sdk-for-net/issues/33856.

For details of the currently implemented API, please see the [api changes file](https://github.com/Azure/azure-sdk-for-net/pull/33063/files#diff-62f3bc8e067ac5ebd48d94a2f93253252b108bf6dbe6baa9c7563943efe03f8dR115).

<br>

<br>

<br>

#### Prior PR Description

Allocating many Dictionaries to store a JsonData object for each node in the tree is very inefficient.

This prototype explores what it would look like to have each JsonData backed by a JsonElement instead.

Mutations will be tracked in a "Patch List" such as the one described in [Mutable "No" Alloc JsonDocument (github.com)](https://gist.github.com/KrzysztofCwalina/885d9f062222f6c0a2ec3e1fb6488823)

#### Initial Perf Benchmarks (Prior to changes)
```
ParseAndReadLargePayloadBenchmark

|          Method |     Mean |    Error |   StdDev |   Median | Ratio | RatioSD |  Gen 0 | Allocated |
|---------------- |---------:|---------:|---------:|---------:|------:|--------:|-------:|----------:|
| ReadJsonElement | 11.05 us | 0.376 us | 1.079 us | 10.73 us |  1.00 |    0.00 | 1.0071 |      4 KB |
|    ReadJsonData | 20.56 us | 0.402 us | 0.494 us | 20.52 us |  1.99 |    0.15 | 4.2114 |     17 KB |


ReadLargePayloadBenchmark

|          Method |      Mean |    Error |   StdDev | Ratio | RatioSD |  Gen 0 | Allocated |
|---------------- |----------:|---------:|---------:|------:|--------:|-------:|----------:|
| ReadJsonElement | 131.72 ns | 2.439 ns | 3.338 ns |  1.00 |    0.00 |      - |         - |
|    ReadJsonData |  85.29 ns | 1.374 ns | 1.285 ns |  0.65 |    0.02 | 0.0114 |      48 B |
```


sdk/core/Azure.Core.Experimental/src/JsonData.cs
```diff
@@ -79,116 +79,19 @@ internal JsonData(object? value) : this(value, DefaultJsonSerializerOptions)
...
-            }
-        }
-
-        private JsonData(JsonElement element)
-        {
-            _kind = element.ValueKind;
-            InitFromElement(element);
-        }
+            if (value is JsonDocument)
+                throw new InvalidOperationException("Calling wrong constructor.");
```
annelo-msft, line 83: I'll change this if we go with this approach - it's for debugging for now.

sdk/core/Azure.Core.Experimental/src/JsonData.cs
```diff
@@ -79,116 +79,19 @@ internal JsonData(object? value) : this(value, DefaultJsonSerializerOptions)
...
+        private JsonData(JsonElement element)
         {
-            if (EnsureObject().TryGetValue(propertyName, out JsonData value))
-            {
-                return value;
-            }
-
-            return null;
+            // Note: you can't call the line below unless element is of type string.
+            //_utf8 = element.GetBytesFromBase64();
```
annelo-msft, line 93: I'll find a way to set `_utf8` lazily when/if we need it.

sdk/core/Azure.Core.Experimental/src/JsonData.cs
```diff
@@ -226,56 +129,63 @@ internal string ToJsonString()
...
+        //}
 
         /// <inheritdoc />
         public override bool Equals(object? obj)
         {
-            if (obj is string)
-            {
-                return this == ((string?)obj);
-            }
+            //if (obj is string)
```
annelo-msft, line 185: This will be added back in.

sdk/core/Azure.Core.Experimental/src/JsonData.cs
```diff
@@ -286,116 +196,61 @@ public override bool Equals(object? obj)
...
+        private int GetInt32() => _element.GetInt32();
 
-        private int GetInt32()
-        {
-            var value = EnsureNumberValue().AsLong();
-            if (value > int.MaxValue || value < int.MinValue)
-            {
-                throw new OverflowException();
-            }
-            return (int)value;
```
annelo-msft, line 330: This means we won't get `InvalidCastExceptions` anymore, but will instead standardize on exceptions thrown by JsonElement, which is nice for consistency with BCL.

sdk/core/Azure.Core.Experimental/src/JsonData.cs
```diff
@@ -430,95 +285,74 @@ internal override void WriteTo(Utf8JsonWriter writer)
...
             throw new InvalidOperationException($"Tried to access indexer with an unsupported index type: {index}");
         }
 
-        private JsonData SetValue(string propertyName, object value)
-        {
-            if (!(value is JsonData json))
-            {
-                json = new JsonData(value);
-            }
+        //private JsonData SetValue(string propertyName, object value)
```
annelo-msft, line 294: Will add mutability with a changelist.

sdk/core/Azure.Core.Experimental/src/JsonData.cs
```diff
@@ -430,95 +285,74 @@ internal override void WriteTo(Utf8JsonWriter writer)
...
         /// <inheritdoc />
         DynamicMetaObject IDynamicMetaObjectProvider.GetMetaObject(Expression parameter) => new MetaObject(parameter, this);
 
-        private IEnumerable GetDynamicEnumerable()
-        {
-            if (_kind == JsonValueKind.Array)
-            {
-                return EnsureArray();
-            }
+        //private IEnumerable GetDynamicEnumerable()
```
annelo-msft, line 347: Will add this back once array strategy is identified.

sdk/core/Azure.Core.Experimental/src/JsonData.cs
```diff
@@ -674,62 +508,62 @@ public override DynamicMetaObject BindSetIndex(SetIndexBinder binder, DynamicMet
...
-                else if (_jsonData.Kind == JsonValueKind.Array)
-                {
-                    foreach (var property in _jsonData.Items)
-                    {
-                        yield return property;
-                    }
-                }
-            }
-        }
+        //internal class JsonDataDebuggerProxy
```
annelo-msft, line 511: These will be added back last.  The methods that only debugger view calls into don't need to be as performant as those used by applications at runtime.

sdk/core/Azure.Core.Experimental/tests/public/JsonDataPublicTests.cs
```diff
@@ -301,24 +303,6 @@ public void CanCastToIEnumerableOfT()
             }
         }
 
-        [Test]
-        public void CanGetDynamicFromBinaryData()
```
annelo-msft, line 305: This is testing IEnumerable functionality, which is covered in other tests.

sdk/core/Azure.Core.Experimental/src/JsonData.cs
```diff
@@ -20,15 +18,204 @@ namespace Azure.Core.Dynamic
...
+                return false;
+            }
+
+            for (int i = _changes.Count - 1; i >= 0; i--)
+            {
+                var change = _changes[i];
+                if (change.Property == path)
+                {
+                    // TODO: does this do boxing?
+                    value = ((JsonDataChange<T>)change).Value;
```
KrzysztofCwalina, line 48: This is unboxing. The list is of interface instances, i.e. the list stores boxed values. This is not ideal. Why do you need this interface?

sdk/core/Azure.Core.Experimental/src/JsonData.cs
```diff
@@ -20,15 +18,204 @@ namespace Azure.Core.Dynamic
...
+
+        internal bool TryGetChange<T>(ReadOnlySpan<byte> path, out T? value)
+        {
+            if (_changes == null)
+            {
+                value = default(T);
+                return false;
+            }
+
+            // TODO: re-enable optimizations
```
KrzysztofCwalina, line 65: Yeah, we will need to fix this. The lines below allocate way too much.

sdk/core/Azure.Core.Experimental/src/JsonData.cs
```diff
@@ -20,15 +18,204 @@ namespace Azure.Core.Dynamic
...
+            // (System.Text.Unicode.Utf8 isn't available to us here)
+            var pathStr = Encoding.UTF8.GetString(path.ToArray());
+
+            //Span<char> utf16 = stackalloc char[path.Length];
+            //OperationStatus status = System.Text.Unicode.Utf8.ToUtf16(path, utf16, out _, out int written, false, true);
```
KrzysztofCwalina, line 70: This should be available as a NuGet package soon. I think @grabyourpitchforks is working on it.

sdk/core/Azure.Core.Experimental/src/JsonData.cs
```diff
@@ -20,15 +18,204 @@ namespace Azure.Core.Dynamic
...
+                }
+            }
+
+            value = default;
+            return false;
+        }
+
+        internal void Set<T>(string path, T value)
+        {
+            // TODO: why was this here?
```
KrzysztofCwalina, line 93: I think it's a mistake, i.e. I looked into my prototype and I cannot tell why added it.

sdk/core/Azure.Core.Experimental/src/JsonData.cs
KrzysztofCwalina: But throwing STJ exceptions is leaking implementation details.

sdk/core/Azure.Core.Experimental/src/JsonDataChange.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Azure.Core.Dynamic
+{
+    internal struct JsonDataChange<T> : IJsonDataChange
```
KrzysztofCwalina, line 10: Why did you make change record to be generic? Is it to avoid boxing the T value? If yes, then I think you should go back to doing it like in the prototype, i.e. use Value type (which will need to be productized too). The reason is that I don't think you can make the changelist non-boxing with generics. Generics force you to add the interface, and then then in turn casing a struct to an interface will box the struct. The Value type can store all primitive value types without boxing.

sdk/core/Azure.Core.Experimental/src/JsonData.cs
annelo-msft: @KrzysztofCwalina is that NuGet package something we could take a dependency on in Core, when this type moves there?

sdk/core/Azure.Core.Experimental/src/JsonDataChange.cs
annelo-msft: Is there a way to accomplish the changelist without productizing the `Value` type?  I don't want JsonData to require a dependency on this.

sdk/core/Azure.Core.Experimental/tests/JsonDataChangeListTests.cs
```diff
@@ -0,0 +1,301 @@
...
+
+            var a = jd.RootElement.GetIndexElement(0);
+            jd.RootElement.GetIndexElement(0).Set(5);
+
+            // This is wicked because 'a' keeps a reference to the root
+            // with the changelist.  Would we detach that?  How would we know to?
+            a.GetProperty("Foo").Set(6);
+
+            Assert.AreEqual(5, jd.RootElement.GetIndexElement(0).GetInt32());
+            Assert.AreEqual(6, a.GetProperty("Foo").GetIndexElement(0).GetInt32());
```
KrzysztofCwalina, line 263: line 260 set `a.Foo` to the number 6. Numbers don't have indexers so a.Foo[0] throws, as it should.

KrzysztofCwalina: I changed the line above to `Assert.AreEqual(6, a.GetProperty("Foo").GetInt32());` and my test passes

annelo-msft: The original pseudocode we were discussing was the following:

```csharp
var json = DynamicJson.Parse("""[ { "Bar" : 4 } ]""");
dynamic a = json[0]; 
json[0] = 5;
a.Bar = 6;
Console.WriteLine(json[0]);
Console.WriteLine(a);
json[0] = a;
Console.WriteLine(json[0]);
Console.WriteLine(json.ToPatch());
```

I had interpreted this as `a` is now a JsonDataElement representing the JSON object `{"Bar":4}`, even if `json[0]` is reassigned in the original object.  But I'm happy to discuss the interpretation if you think these aren't the right semantics for the type.

sdk/core/Azure.Core.Experimental/src/JsonData.cs
```diff
@@ -20,15 +19,67 @@ namespace Azure.Core.Dynamic
     /// <summary>
     /// A mutable representation of a JSON value.
     /// </summary>
-    [DebuggerDisplay("{DebuggerDisplay,nq}")]
-    [DebuggerTypeProxy(typeof(JsonDataDebuggerProxy))]
+    //[DebuggerDisplay("{DebuggerDisplay,nq}")]
```
annelo-msft, line 22: TODO: Add back debugger view

sdk/core/Azure.Core.Experimental/src/DynamicJson.DynamicMetaObject.cs
```diff
@@ -48,11 +51,21 @@ internal object GetProperty(string name)
             return null;
         }
 
+        internal T ConvertTo<T>()
+        {
+            // TODO: Respect user-provided serialization options
+            // TODO: Could we optimize this by serializing from the byte array instead?  We don't currently slice into this in WriteTo(), but could look at storing that.
+            return JsonSerializer.Deserialize<T>(_element.ToString() , MutableJsonDocument.DefaultJsonSerializerOptions);
```
KrzysztofCwalina, line 58: Can you use this overload? https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializer.deserialize?view=net-7.0#system-text-json-jsonserializer-deserialize(system-text-json-jsonelement-system-type-system-text-json-serialization-jsonserializercontext). This will avoid creating a string.

sdk/core/Azure.Core.Experimental/src/DynamicJson.DynamicMetaObject.cs
```diff
@@ -83,6 +119,14 @@ public override DynamicMetaObject BindSetMember(SetMemberBinder binder, DynamicM
                 BindingRestrictions restrictions = BindingRestrictions.GetTypeRestriction(Expression, LimitType);
                 return new DynamicMetaObject(setCall, restrictions);
             }
+
+            private static Dictionary<Type, MethodInfo> GetCastFromOperators()
+            {
+                return typeof(DynamicJson)
+                    .GetMethods(BindingFlags.Public | BindingFlags.Static)
+                    .Where(method => method.Name == "op_Explicit" || method.Name == "op_Implicit")
```
KrzysztofCwalina, line 127: We should cache these MethodInfos in a static.

sdk/core/Azure.Core.Experimental/src/DynamicJson.DynamicMetaObject.cs
annelo-msft: Much better!  Yes, thank you, I absolutely will.

sdk/core/Azure.Core.Experimental/src/DynamicJson.DynamicMetaObject.cs
annelo-msft: I believe I am doing what you're suggesting above here: https://github.com/Azure/azure-sdk-for-net/pull/33063/files#diff-9fa0e6f4c4c212bff6545aca2d81f128cf2ac7c6f68db22ef905ecbe484ce1d7R81

But let me know if you had something else in mind ...

annelo-msft: If you have suggestions to refactor to make this more clear, I'm happy to take them.

sdk/core/Azure.Core.Experimental/src/DynamicJson.DynamicMetaObject.cs
annelo-msft: Oh, stink, no, it looks like this is only available in .NET 6 and 7.

annelo-msft: Added multitargeting to address.

sdk/core/Azure.Core.Experimental/src/DynamicJson.DynamicMetaObject.cs
annelo-msft: Refactored for clarity.

sdk/core/Azure.Core.Experimental/src/MutableJsonElement.cs
```diff
@@ -405,6 +403,15 @@ private byte[] GetRawBytes()
             return changedElementStream.ToArray();
         }
 
+        internal static Utf8JsonReader GetReaderForElement(JsonElement element)
+        {
+            using MemoryStream stream = new();
+            Utf8JsonWriter writer = new(stream);
+            element.WriteTo(writer);
+            writer.Flush();
+            return new Utf8JsonReader(stream.ToArray());
```
KrzysztofCwalina, line 412: you can call GetBuffer().AsSpan().Slice(0, stream.Position) to get the array without a new allocation/copy

annelo-msft: Excellent - thank you!

sdk/core/Azure.Core.Experimental/src/DynamicJson.DynamicMetaObject.cs
```diff
@@ -0,0 +1,181 @@
...
+        private static readonly MethodInfo GetPropertyMethod = typeof(DynamicJson).GetMethod(nameof(GetProperty), BindingFlags.NonPublic | BindingFlags.Instance)!;
+        private static readonly MethodInfo SetMethod = typeof(DynamicJson).GetMethod(nameof(Set), BindingFlags.NonPublic | BindingFlags.Instance, null, new Type[] { typeof(object) }, null)!;
+        private static readonly MethodInfo GetViaIndexerMethod = typeof(DynamicJson).GetMethod(nameof(GetViaIndexer), BindingFlags.NonPublic | BindingFlags.Instance)!;
+        private static readonly MethodInfo SetViaIndexerMethod = typeof(DynamicJson).GetMethod(nameof(SetViaIndexer), BindingFlags.NonPublic | BindingFlags.Instance)!;
+
+        /// <summary>
+        /// Gets a value indicating whether the current instance has a valid value of its underlying type.
+        /// </summary>
+        // TODO: Decide if this is the API we want for this.  We could also expose ValueKind.
+        public bool HasValue => _element.ValueKind != JsonValueKind.Null;
```
annelo-msft, line 23: @KrzysztofCwalina, curious about your perspective on this.  We could expose ValueKind instead, but anything that is a property access on DynamicJson has the following concerns:
1. Potential conflict with property in the wrapped JSON
2. Caller has to be aware of DynamicJson public API, which we're trying to avoid.

We could also remove this entirely and force callers to rely on a cast to check for null.  

What's implemented is my favorite of the options, since the usage would look most like a strongly-typed model.

KrzysztofCwalina: We should not expose ValueKind. ValueKind is specific to JSON and moreover to System.Text.Json. As we discussed in the past, the meta object should be independent of the format or a specific format reader technology. 

Secondly, I might not be understanding the issue (why we need HasValue at all), but I suspect it's because some JSOn properties are optional and the caller might not know if they happen to exist or not. Let me know if this is not the issue. 

If it is, I thought it would work the following way: you can cast any node of the object graph to either: 

1. A reference type, e.g. `(string)foo.Bar`, in which case the user can simply check if the reference is null.
2. A value type, e.g. `(int)foo.Bar`, in which case we would throw if Bar property does not exist.
3. A nullable variant of a value type, e.g. `(int?)foo.Bar`, in which case the user can just check HasValue on the nullable.

annelo-msft: I like this path forward.  We'll see it in action when I update the CLU samples -- the motivating example was here: https://github.com/Azure/azure-sdk-for-net/pull/32501/files#diff-d55b2105899a17c3fafc1a079b6bb5a777c542671abad5f96c12223977a26597R188

KrzysztofCwalina (approved):

sdk/core/Azure.Core.Experimental/api/Azure.Core.Experimental.netstandard2.0.cs
```diff
@@ -115,52 +115,90 @@ public static partial class BinaryDataExtensions
...
+        public static implicit operator int? (Azure.Core.Dynamic.DynamicJson value) { throw null; }
+        public static implicit operator long? (Azure.Core.Dynamic.DynamicJson value) { throw null; }
+        public static implicit operator float? (Azure.Core.Dynamic.DynamicJson value) { throw null; }
+        public static implicit operator float (Azure.Core.Dynamic.DynamicJson value) { throw null; }
+        public static implicit operator string (Azure.Core.Dynamic.DynamicJson value) { throw null; }
         System.Dynamic.DynamicMetaObject System.Dynamic.IDynamicMetaObjectProvider.GetMetaObject(System.Linq.Expressions.Expression parameter) { throw null; }
         public override string ToString() { throw null; }
+        [System.Diagnostics.DebuggerDisplayAttribute("{Current,nq}")]
+        [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
+        public partial struct ArrayEnumerator : System.Collections.Generic.IEnumerable<Azure.Core.Dynamic.DynamicJson>, System.Collections.Generic.IEnumerator<Azure.Core.Dynamic.DynamicJson>, System.Collections.IEnumerable, System.Collections.IEnumerator, System.IDisposable
```
annelo-msft, line 139: I will look into whether this can be made internal and accessed through the dynamic layer.

KrzysztofCwalina: If we are planning to ship the MutableJsonDocument publicly, then this enumerator adds value (as public). It will allow enumerating arrays without the need to allocate (box) the enumerator

annelo-msft: I agree it makes sense to have MutableJsonElement.ArrayEnumerator as a public type.  Do you feel strongly about also exposing DynamicJson.ArrayEnumerator as a public type?

sdk/core/Azure.Core.Experimental/src/JsonData.cs
annelo-msft: Tracking with https://github.com/Azure/azure-sdk-for-net/issues/33856

sdk/core/Azure.Core.Experimental/src/JsonData.cs
annelo-msft: Tracking with https://github.com/Azure/azure-sdk-for-net/issues/33856

sdk/core/Azure.Core.Experimental/src/JsonData.cs
```diff
@@ -20,15 +18,204 @@ namespace Azure.Core.Dynamic
...
+                stream.Flush();
+                return;
+            }
+            WriteTheHardWay(writer);
+        }
+
+        // TODO: if we keep this, make it an extension on stream.
+        private static void Write(Stream stream, ReadOnlySpan<byte> buffer)
+        {
+            byte[] sharedBuffer = ArrayPool<byte>.Shared.Rent(buffer.Length);
```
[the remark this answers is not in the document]
annelo-msft: Tracking with https://github.com/Azure/azure-sdk-for-net/issues/33856

sdk/core/Azure.Core.Experimental/tests/JsonDataChangeListTests.cs
annelo-msft: Addressed, and tested in the reference semantics tests in `MutableJsonDocumentTests`

sdk/core/Azure.Core.Experimental/src/JsonData.cs
annelo-msft: Tracking with https://github.com/Azure/azure-sdk-for-net/issues/33856

sdk/core/Azure.Core.Experimental/api/Azure.Core.Experimental.netstandard2.0.cs
annelo-msft: Tracking with https://github.com/Azure/azure-sdk-for-net/issues/33856

sdk/core/Azure.Core.Experimental/src/JsonData.cs
annelo-msft: Tracking with https://github.com/Azure/azure-sdk-for-net/issues/33856

PR #33073, opened 2022-12-15
Polish Azure.ResourceManager.ManagedServices

ArthurMa1978:


sdk/managedservices/Azure.ResourceManager.ManagedServices/src/Generated/Models/ManagedServicesEligibleAuthorization.cs
```diff
@@ -40,12 +35,12 @@ internal EligibleAuthorization(string principalId, string principalIdDisplayName
         }
 
         /// <summary> The identifier of the Azure Active Directory principal. </summary>
-        public string PrincipalId { get; set; }
+        public Guid PrincipalId { get; set; }
         /// <summary> The display name of the Azure Active Directory principal. </summary>
         public string PrincipalIdDisplayName { get; set; }
         /// <summary> The identifier of the Azure built-in role that defines the permissions that the Azure Active Directory principal will have on the projected scope. </summary>
-        public string RoleDefinitionId { get; set; }
+        public Guid RoleDefinitionId { get; set; }
```
ArcturusZhang, line 42: Well this actually corresponds to the Name in authorizationRoleDefinition, which is a string. Therefore I do not recommend that we make this a guid despite it is really a guid

ArcturusZhang (approved):

PR #33078, opened 2022-12-15
Fix flaky Core test

christothes:

### DIFF
--- a/sdk/core/Azure.Core/tests/AzureSasCredentialSynchronousPolicyTests.cs
+++ b/sdk/core/Azure.Core/tests/AzureSasCredentialSynchronousPolicyTests.cs
@@ -108,8 +108,8 @@ namespace Azure.Core.Tests
                 if (callCount++ == 0)
                 {
                     Assert.AreEqual("?foo=bar&sig=first_signature_value", req.Uri.Query);
-                    azureSasCredential.Update(SECOND_SIGNATURE_VALUE);
                     Thread.Sleep(100);
+                    azureSasCredential.Update(SECOND_SIGNATURE_VALUE);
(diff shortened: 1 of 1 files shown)


sdk/core/Azure.Core/tests/AzureSasCredentialSynchronousPolicyTests.cs
```diff
@@ -108,8 +108,8 @@ public async Task VerifyRetryAfterSasCredentialUpdateOperation()
                 if (callCount++ == 0)
                 {
                     Assert.AreEqual("?foo=bar&sig=first_signature_value", req.Uri.Query);
-                    azureSasCredential.Update(SECOND_SIGNATURE_VALUE);
                     Thread.Sleep(100);
+                    azureSasCredential.Update(SECOND_SIGNATURE_VALUE);
```
annelo-msft, line 112: Would you be willing to describe why this fixes the issue?

christothes: Yes - in certain timing conditions, `message.ProcessingContext.StartTime.Ticks` is equal to `AzureSasCredential.SignatureUpdated`. The placement of this delay ensures that there is some space between the two time values. That is what I intended to do initially, but I put the delay in the wrong place.

annelo-msft (approved):

sdk/core/Azure.Core/tests/AzureSasCredentialSynchronousPolicyTests.cs
annelo-msft: Thanks!

PR #33087, opened 2022-12-15
Remove new() constraint from TableClient methods

christothes:
After adding the `TypeBinder` implementation for serialization, the `new()` constraint was only used for one code path. That path didn't actually need the constraint because the type would be known in that case.


sdk/tables/Azure.Data.Tables/src/Extensions/DictionaryTableExtensions.cs
```diff
@@ -132,20 +132,19 @@ internal static List<T> ToTableEntityList<T>(this IReadOnlyList<IDictionary<stri
...
                 entity.CastAndRemoveAnnotations();
 
                 foreach (var entProperty in entity.Keys)
                 {
-                    dictionary[entProperty] = entity[entProperty];
+                    tableEntity[entProperty] = entity[entProperty];
                 }
 
-                return result;
+                return (T)(object)tableEntity;
```
jsquire, line 147: Will this always be a safe cast?   I may not be following fully, but it looks like we're casting from `TableEntity` to a derived type, which I wouldn't expect to work.

jsquire (approved):

sdk/tables/Azure.Data.Tables/src/Extensions/DictionaryTableExtensions.cs
christothes: Actually, you are right. Initially I was thinking that these would be either standard custom models or `TableEntity`. We need to preserve the type we got in for cases where a custom model type also implements `IDictionary<string, object>`

PR #33106, opened 2022-12-16
Only log timeout retries as verbose

JoshLove-msft:


sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/ServiceBusRetryPolicy.cs
```diff
@@ -168,7 +168,7 @@ public abstract class ServiceBusRetryPolicy
                     TimeSpan? retryDelay = CalculateRetryDelay(activeEx, failedAttemptCount);
                     if (retryDelay.HasValue && !scope.IsDisposed && !cancellationToken.IsCancellationRequested)
                     {
-                        if (logRetriesAsVerbose)
+                        if (logTimeoutRetriesAsVerbose && activeEx is ServiceBusException { Reason: ServiceBusFailureReason.ServiceTimeout })
```
christothes, line 171: Do we need a test covering this behavior?

christothes (approved):

sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/ServiceBusRetryPolicy.cs
JoshLove-msft: We already have a test covering the Timeout failure case. It's hard to test the other way as there are no other retriable errors that we can deterministically cause.

PR #33130, opened 2022-12-20
[WebPubSub]Move StringEnumConverter to explicit type converter

JialinXin:
Remove generic converter `StringEnumConverter` and `CamelCasePropertyNamesContractResolver` in global settings to avoid breaking others. Change to explicit type converter for the enum types that read by customer.

Fix issue: https://github.com/Azure/azure-functions-durable-extension/issues/2338

### DIFF
--- a/sdk/webpubsub/Microsoft.Azure.WebJobs.Extensions.WebPubSub/src/Config/WebPubSubConfigProvider.cs
+++ b/sdk/webpubsub/Microsoft.Azure.WebJobs.Extensions.WebPubSub/src/Config/WebPubSubConfigProvider.cs
@@ -158,11 +158,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.WebPubSub
             {
                 Converters = new List<JsonConverter>
                 {
-                    new StringEnumConverter(),
                     new BinaryDataJsonConverter(),
                     new ConnectionStatesNewtonsoftConverter(),
+                    new WebPubSubDataTypeJsonConverter(),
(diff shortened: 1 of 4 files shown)


sdk/webpubsub/Microsoft.Azure.WebJobs.Extensions.WebPubSub/src/Config/WebPubSubConfigProvider.cs
```diff
@@ -158,11 +158,11 @@ internal static void RegisterJsonConverter()
             {
```
vicancy, line 158: I think the better way is to avoid using the default converter but use our own thread static JsonSerializer instance.

JialinXin: Agree, however we're not able to control all the serialization points and some are in azure function core. And to provide correct serialize/deserialize behavior like readable string value instead of default behavior enum int, we have to add our converter in the default settings. Limit to our customized type would lower the risk to affect others. This only works for script language like Javascript and Python, and customer doesn't have chance to change these settings. In C# language, these converters are not in use.

Possible overwrite place by WebJobs.Host SDK like this: https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Bindings/DefaultAttributeInvokerDescriptor.cs#L66

vicancy (approved):

PR #33149, opened 2022-12-21
DOCS: Fix `TableClient.GetEntityIfExistsAsync` thrown exception description

### LINKED ISSUE #33144: DOCS: `TableClient.GetEntityIfExists(-Async)` has misleading documentation about thrown exceptions
### Library name and version

Azure.Data.Tables 12.7.1

### Query/Question

(Since this is not strictly a bug or vulnerability, I file it as a question/query).

I believe that the documentation (xml comments) associated with the following methods
- [Azure.Data.Tables.TableClient.GetEntityIfExists](https://github.com/Azure/azure-sdk-for-net/blob/e8664098abd0fdc00edcee78867b1245d2f4529f/sdk/tables/Azure.Data.Tables/src/TableClient.cs#L710)
- [Azure.Data.Tables.TableClient.GetEntityIfExistsAsync](https://github.com/Azure/azure-sdk-for-net/blob/e8664098abd0fdc00edcee78867b1245d2f4529f/sdk/tables/Azure.Data.Tables/src/TableClient.cs#L727)

incorrectly inform consumers that these methods throw `RequestFailedException` if the entity is missing. Reading through https://github.com/Azure/azure-sdk-for-net/pull/30888 and related issues, I found that the purpose of these methods is to not raise an exception when a query to Azure table returns HTTP 404. We checked this against test Azure environment and got no exception when querying for a missing entity.

This information also gets reflected on the [Microsoft Learn](https://learn.microsoft.com/en-us/dotnet/api/azure.data.tables.tableclient.getentityifexistsasync?view=azure-dotnet#exceptions) page, and it is from there I first learned about inconsistent method documentation.

I am happy to submit a tiny PR if this is indeed an issue and it has not yet been fixed. Or perhaps you can fix it as part of your internal workflow.

### Environment

N/A

Ilia-Kosenkov:
Closes #33144.
`TableClient.GetEntityIfExistsAsync` documentation states that it throws an exception if the requested entity does not exist. This statement is misleading. The non-async version of this method, `TableClient.GetEntityIfExists`, correctly notifies the user that the exception is thrown if an unexpected error occurs.

This PR corrects this statement in the documentation.

Pinging @christothes

### DIFF
--- a/sdk/tables/Azure.Data.Tables/src/TableClient.cs
+++ b/sdk/tables/Azure.Data.Tables/src/TableClient.cs
@@ -721,7 +721,7 @@ namespace Azure.Data.Tables
         /// <param name="select">Selects which set of entity properties to return in the result set. Pass <c>null</c> to retreive all properties.</param>
         /// <param name="cancellationToken">A <see cref="CancellationToken"/> controlling the request lifetime.</param>
         /// <returns> The <see cref="NullableResponse{T}"/> whose <c>HasValue</c> property will return <c>true</c> if the entity existed, otherwise <c>false</c>.</returns>
-        /// <exception cref="RequestFailedException">Exception thrown if the entity doesn't exist.</exception>
+        /// <exception cref="RequestFailedException">Exception thrown if an unexpected error occurs.</exception>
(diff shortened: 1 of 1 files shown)


christothes (approved):
Thanks @Ilia-Kosenkov !

PR #33156, opened 2022-12-22
Download the latest SDK as the input to existing code

pshao25:
Will merge after this https://github.com/Azure/autorest.csharp/pull/2934


eng/scripts/Cadl-Project-Sync.ps1
```diff
@@ -9,6 +9,19 @@ $ErrorActionPreference = "Stop"
 . $PSScriptRoot/../common/scripts/Helpers/PSModule-Helpers.ps1
 Install-ModuleIfNotInstalled "powershell-yaml" "0.4.1" | Import-Module
 $sparseCheckoutFile = ".git/info/sparse-checkout"
+$sdkGitRemoteAPI = "https://api.github.com/repos/Azure/azure-sdk-for-net/git/refs/heads/main"
+$sdkGitRemote = "https://github.com/Azure/azure-sdk-for-net.git"
+
+function GetLatestCommit() {
```
weshaggard, line 15: Why do we need to do this? If we are just picking the latest then the clone will do that for us.

pshao25: The scenario is: when we want to build a new version of SDK, the new API depends on how the previous one looks like. E.g., https://github.com/Azure/autorest.csharp/issues/2887

eng/scripts/Cadl-Project-Generate.ps1
```diff
@@ -52,16 +52,18 @@ if (Test-Path "$npmWorkingDir/cadl-project.yaml") {
...
         $csharpSdkFolder = Resolve-Path $csharpSdkFolder
-        Write-Host("npx cadl compile $mainCadlFile --emit `"`@azure-tools/cadl-csharp`" --arg `"csharp-sdk-folder=$csharpSdkFolder`"")
-        npx cadl compile $mainCadlFile --emit "@azure-tools/cadl-csharp" --arg "csharp-sdk-folder=$csharpSdkFolder"
+        $command = "npx cadl compile $mainCadlFile --emit @azure-tools/cadl-csharp --arg csharp-sdk-folder=$csharpSdkFolder$existingOption"
+        Invoke $command
     } else {
-        Write-Host("npx cadl compile $mainCadlFile --emit `"`@azure-tools/cadl-csharp`" --option `"@azure-tools/cadl-csharp.emitter-output-dir=$resolvedProjectDirectory`"")
-        npx cadl compile $mainCadlFile --emit "@azure-tools/cadl-csharp" --option @azure-tools/cadl-csharp.emitter-output-dir=$resolvedProjectDirectory
+        $command = "npx cadl compile $mainCadlFile --emit @azure-tools/cadl-csharp --option @azure-tools/cadl-csharp.emitter-output-dir=$resolvedProjectDirectory$existingOption"
+        Invoke $command
```
archerzz, line 66: What's the benefit of `Invoke`? If it throws exception, then we can remove the `if ($LASTEXITCODE) { exit $LASTEXITCODE }` below.

eng/scripts/Cadl-Project-Sync.ps1
```diff
@@ -128,11 +141,42 @@ if ( $configuration["repo"] -and $configuration["commit"]) {
...
+CopyToProjectIfNeeded `
+    -cloneRoot $specCloneDir `
+    -mainDir $specSubDirectory `
     -dest $tempCadlDir `
-    -specAdditionalSubDirectories $configuration["additionalDirectories"]
+    -additionalSubDirectories $configuration["additionalDirectories"]
+
+
+# Download the existing SDK
+$loadExisting = $configuration["load-existing"] ?? $true
```
archerzz, line 154: My preference is to turn it off by default for now. We have pipeline timeout problems, downloading existing SDK will make things worse.

eng/scripts/Cadl-Project-Generate.ps1
pshao25: No, it cannot. I just want to make the log and the command same.

.gitignore
```diff
@@ -174,5 +174,6 @@ BenchmarkDotNet.Artifacts
 artifacts
 .assets
 
-# Temporary cadl folders for cadl generation
-TempCadlFiles/
\ No newline at end of file
+# Temporary folders for cadl version SDK generation
+TempCadlFiles/
+TempSDKFiles/
```
weshaggard, line 179: Do we really need another temp directory? Can it go under an existing directory like artifacts?

eng/scripts/Cadl-Project-Sync.ps1
weshaggard: @m-nash I assume this is something you are helping design? I'm a little worried about us trying to pull source from another sha and download files from other commits in the middle of our generation processes.

.gitignore
pshao25: @m-nash What do you think?

pshao25: At least to me, it would be more appropriate to put under each SDK folders since they are input not output. Do you have any concern about this?

eng/scripts/Cadl-Project-Sync.ps1
m-nash: I think this code should not be part of the cadl scripts since its not unique to cadl.

@weshaggard in order to generate proper version overloads we need to know the previous contract.  We can use either the package managers (nuget, npm, etc) or we can sync to the release tag of the last GA version.  If we do the sync it can be reused by all languages, if we do the package managers its unique code for everyone.

We should definitely have this sync turned off by default and only enabled once a language generator needs it as not all will need this for back compatability right away.

I also think there is opportunity here for having these things pre-fetched to minimize download / network usage since that is already an issue in some of our pipelines today.

@pshao25  can we mark this as draft as we work through the prototype?

.gitignore
m-nash: @weshaggard for the prototype here lets ignore "how we get the code" and once we have everything else hooked up we will sync with you on the best way to get this information.

PR #33162, opened 2022-12-22
Allow updating prefetch count with UpdatePrefetchCount

### LINKED ISSUE #31624: [FEATURE REQ] ServiceBusProcessor allow to adjust the prefetch count forUpdateConcurrency

danielmarbach:
Follow up for https://github.com/Azure/azure-sdk-for-net/pull/32572 using a different implementation

Fixes https://github.com/Azure/azure-sdk-for-net/issues/31624

Missing:

- [x] Running ApiCompat
- [x] Slightly better tests

### DIFF
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Core/TransportReceiver.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Core/TransportReceiver.cs
@@ -33,6 +33,11 @@ namespace Azure.Messaging.ServiceBus.Core
         public abstract DateTimeOffset SessionLockedUntil { get; protected set; }
 
         /// <summary>
+        /// The prefetch count associated with the receiver.
+        /// </summary>
+        public abstract int PrefetchCount { get; set; }
+
+        /// <summary>
         /// Receives a set of <see cref="ServiceBusReceivedMessage" /> from the entity using <see cref="ServiceBusReceiveMode"/> mode.
         /// </summary>
         /// <param name="maximumMessageCount">The maximum number of messages that will be received.</param>
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessor.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessor.cs
@@ -699,10 +699,7 @@ namespace Azure.Messaging.ServiceBus
                 for (int i = 0; i < receiverManagers; i++)
                 {
                     var receiverManager = _receiverManagers[i];
-                    if (receiverManager.Receiver != null && receiverManager.Receiver.PrefetchCount != prefetchCount)
-                    {
-                        receiverManager.Receiver.PrefetchCount = prefetchCount;
-                    }
+                    receiverManager.UpdatePrefetchCount(prefetchCount);
                 }
             }
         }
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessor.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessor.cs
@@ -48,7 +48,7 @@ namespace Azure.Messaging.ServiceBus
         // we use int.MaxValue as the maxCount since the concurrency can be changed dynamically by the user
         private readonly SemaphoreSlim _messageHandlerSemaphore = new SemaphoreSlim(0, int.MaxValue);
 
-        private readonly object _maxConcurrencySyncLock = new();
+        private readonly object _optionsLock = new();
 
         /// <summary>
         /// The primitive for ensuring that the service is not overloaded with
@@ -78,7 +78,7 @@ namespace Azure.Messaging.ServiceBus
         /// <summary>
         /// Gets the ID used to identify this processor. This can be used to correlate logs and exceptions.
         /// </summary>
-        public virtual string Identifier { get; }
+        public virtual string Identifier => Options.Identifier;
 
         /// <summary>
         /// Gets the <see cref="ReceiveMode"/> used to specify how messages are received. Defaults to PeekLock mode.
@@ -87,7 +87,7 @@ namespace Azure.Messaging.ServiceBus
         /// The receive mode is specified using <see cref="ServiceBusProcessorOptions.ReceiveMode"/>
         /// and has a default mode of <see cref="ServiceBusReceiveMode.PeekLock"/>.
         /// </value>
-        public virtual ServiceBusReceiveMode ReceiveMode { get; }
+        public virtual ServiceBusReceiveMode ReceiveMode => Options.ReceiveMode;
 
         /// <summary>
         /// Gets whether the processor is configured to process session entities.
@@ -103,7 +103,7 @@ namespace Azure.Messaging.ServiceBus
         /// The prefetch count is specified using <see cref="ServiceBusProcessorOptions.PrefetchCount"/>
         /// and has a default value of 0.
         /// </value>
-        public virtual int PrefetchCount { get; }
+        public virtual int PrefetchCount => Options.PrefetchCount;
 
         /// <summary>
         /// Gets whether or not this processor is currently processing messages.
@@ -129,8 +129,7 @@ namespace Azure.Messaging.ServiceBus
         /// The number of maximum concurrent calls is specified using <see cref="ServiceBusProcessorOptions.MaxConcurrentCalls"/>
         /// and has a default value of 1.
         /// </value>
-        public virtual int MaxConcurrentCalls => _maxConcurrentCalls;
-        private volatile int _maxConcurrentCalls;
+        public virtual int MaxConcurrentCalls => Options.MaxConcurrentCalls;
         private int _currentConcurrentCalls;
 
         internal int MaxConcurrentSessions => _maxConcurrentSessions;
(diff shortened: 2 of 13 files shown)


sdk/servicebus/Azure.Messaging.ServiceBus/src/Core/TransportReceiver.cs
```diff
@@ -32,6 +32,11 @@ internal abstract class TransportReceiver
         /// </summary>
         public abstract DateTimeOffset SessionLockedUntil { get; protected set; }
 
+        /// <summary>
+        /// The prefetch count associated with the receiver.
+        /// </summary>
+        public abstract int PrefetchCount { get; set; }
```
danielmarbach, line 38: At first glance I was on the fence to expose it here, but then I realized even though the abstract class says it should not have protocol-specific stuff it seems there is already a lot of protocol-specific stuff passed while creating the types, so I exposed it here since that is the central place where links are managed.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessor.cs
```diff
@@ -103,7 +103,7 @@ public class ServiceBusProcessor : IAsyncDisposable
         /// The prefetch count is specified using <see cref="ServiceBusProcessorOptions.PrefetchCount"/>
         /// and has a default value of 0.
         /// </value>
-        public virtual int PrefetchCount { get; }
+        public virtual int PrefetchCount => Options.PrefetchCount;
```
danielmarbach, line 106: Forward delegating will keep them in sync as discussed on the other PR

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessor.cs
```diff
@@ -245,21 +244,16 @@ public virtual bool IsClosed
             Options = options?.Clone() ?? new ServiceBusProcessorOptions();
             Connection = connection;
             EntityPath = EntityNameFormatter.FormatEntityPath(entityPath, Options.SubQueue);
-            Identifier = string.IsNullOrEmpty(Options.Identifier) ? DiagnosticUtilities.GenerateIdentifier(EntityPath) : Options.Identifier;
+            Options.Identifier = string.IsNullOrEmpty(Options.Identifier) ? DiagnosticUtilities.GenerateIdentifier(EntityPath) : Options.Identifier;
```
danielmarbach, line 247: Only the clone is modified

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessor.cs
```diff
@@ -700,6 +694,16 @@ private void ReconcileReceiverManagers(int maxConcurrentSessions)
                         }
                     }
                 }
+
+                int receiverManagers = _receiverManagers.Count;
+                for (int i = 0; i < receiverManagers; i++)
+                {
+                    var receiverManager = _receiverManagers[i];
+                    if (receiverManager.Receiver != null && receiverManager.Receiver.PrefetchCount != prefetchCount)
```
danielmarbach, line 702: This check is a bit weird because the SessionReceiveManager only associates the receiver at a later stage. This is not a problem though because the session manager has already a reference to the adjusted processor options and will take it into account once opened.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessor.cs
```diff
@@ -1088,7 +1107,7 @@ private async Task ReconcileConcurrencyAsync()
             else if (diff < 0)
             {
                 var activeTasks = TaskTuples.Where(t => !t.Task.IsCompleted).ToList();
-                int excessTasks = activeTasks.Count - _maxConcurrentCalls;
+                int excessTasks = activeTasks.Count - maxConcurrentCalls;
```
danielmarbach, line 1110: I was wondering if this is an oversight because the majority of the code here actually intends to use the copied value which wasn't the case here

sdk/servicebus/Azure.Messaging.ServiceBus/src/Receiver/ServiceBusReceiver.cs
```diff
@@ -171,7 +181,6 @@ public virtual bool IsClosed
                 _connection = connection;
                 _retryPolicy = connection.RetryOptions.ToRetryPolicy();
                 ReceiveMode = options.ReceiveMode;
-                PrefetchCount = options.PrefetchCount;
```
danielmarbach, line 174: Maybe if you like the pattern of storing the options and then forward delegating then it would make sense to align this here too.

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Processor/ProcessorLiveTests.cs
```diff
@@ -1064,6 +1064,69 @@ async Task ProcessMessage(ProcessMessageEventArgs args)
             }
         }
 
+        [Test]
+        public async Task CanUpdatePrefetchCount()
```
danielmarbach, line 1068: I really struggled to come up with a good test

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessor.cs
```diff
@@ -700,6 +694,16 @@ private void ReconcileReceiverManagers(int maxConcurrentSessions)
...
                     }
                 }
+
+                int receiverManagers = _receiverManagers.Count;
+                for (int i = 0; i < receiverManagers; i++)
+                {
+                    var receiverManager = _receiverManagers[i];
+                    if (receiverManager.Receiver != null && receiverManager.Receiver.PrefetchCount != prefetchCount)
+                    {
+                        receiverManager.Receiver.PrefetchCount = prefetchCount;
```
JoshLove-msft, line 704: nit: can we encapsulate with an UpdatePrefetch method on ReceiverManager?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessor.cs
JoshLove-msft: Yep, it sure was 😄 
Thanks for fixing!

sdk/servicebus/Azure.Messaging.ServiceBus/src/Receiver/ServiceBusReceiver.cs
JoshLove-msft: I don't think it is necessary here since none of the properties are publicly mutable.

JoshLove-msft:
Looks like some existing unit test are now failing. Can you try running locally?

JoshLove-msft:
Will need to run `eng\scripts\Export-API.ps1 servicebus`

JoshLove-msft (commented):
Looks great! Once we get the failing tests fixed and API exported, I will do another pass and it should be good to go.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Core/TransportReceiver.cs
JoshLove-msft: Yes, this is also on the public type so it makes sense it would be on this exposed on this type as well.

danielmarbach:
@JoshLove-msft done

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Processor/ProcessorLiveTests.cs
JoshLove-msft: Outside of doing some some sort of perf comparison, I think this is the best we can do for a live test.

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Processor/ProcessorLiveTests.cs
```diff
@@ -1064,6 +1064,69 @@ async Task ProcessMessage(ProcessMessageEventArgs args)
...
+                        processor.UpdatePrefetch(1);
+                        Assert.AreEqual(20, processor.MaxConcurrentCalls);
+                        Assert.AreEqual(1, processor.PrefetchCount);
+                    }
+
+                    // increase prefetch
+                    if (count == 150)
+                    {
+                        Assert.LessOrEqual(processor.TaskTuples.Where(t => !t.Task.IsCompleted).Count(), 20);
+                        processor.UpdatePrefetch(10);
```
JoshLove-msft, line 1111: Would be nice to exercise setting prefetch to 0 as well.

danielmarbach: Feel free to push that. I might have time at some point but it would also ok if you beat me to it

danielmarbach:
I think it also requires a changelog update

JoshLove-msft (approved):

PR #33186, opened 2022-12-24
Add .NET cryptography primitives

### LINKED ISSUE #3545: Key Vault support for .NET RSA Crypto primitives
Many applications in .NET need an `RSA` derived object for their crypto functions. Today, there is no built-in way of doing this with keys/certificates stored in Key Vault. I have written an implementation for this, but I believe it belongs in this SDK. 

Key Vault does have its own crypto abstractions, but that does not address the need to use Key Vault in places that are not "key vault crypto" aware. 

I'm happy to discuss API names, shapes, parameters, but I have this working already here:
https://github.com/novotnyllc/RSAKeyVaultProvider

The `RSAKeyVault` enables you to use secrets and certificates stored in an 
Azure Key Vault for performing signing and decryption operations. (Encrypt and verify
can be done locally with the public key material.) The type derives from `RSA` so can
be used anywhere an `AsymmetricAlgorithm` can be, including with `SignedXml` types.

I would like to contribute this as a PR into the Key Vault SDK but wanted to start a discussion here before sending a PR.

heaths:
Partly resolves #3545 by adding an RSA implementation.


sdk/keyvault/Azure.Security.KeyVault.Keys/src/Cryptography/CryptographyClient.cs
```diff
@@ -1592,6 +1593,62 @@ public virtual VerifyResult VerifyData(SignatureAlgorithm algorithm, Stream data
...
+        /// <summary>
+        /// Creates an <see cref="RSA"/> implementation backed by this <see cref="CryptographyClient"/>.
+        /// </summary>
+        /// <param name="cancellationToken">A <see cref="CancellationToken"/> to cancel this operation.</param>
+        /// <returns>An <see cref="RSA"/> implementation backed by this <see cref="CryptographyClient"/>.</returns>
+        /// <remarks>
+        /// The <see cref="CryptographyClient"/> will attempt to download the public key synchronously.
+        /// </remarks>
+        /// <exception cref="RequestFailedException">The server returned an error. See <see cref="Exception.Message"/> for details returned from the server.</exception>
+        public virtual RSA CreateRSA(CancellationToken cancellationToken = default)
```
heaths, line 1605: @KrzysztofCwalina @tg-msft @schaabs there are separate sync and async calls for the initial fetch attempt of the public key; otherwise, waiting to fetch inside the crypto primitive will always be sync over network.

Also, is returning the abstract base `RSA` here fine? One normally should not interact with the derivatives.

heaths:
Replaces #25178

sdk/keyvault/Azure.Security.KeyVault.Keys/src/Cryptography/RSAKeyVault.cs
```diff
@@ -0,0 +1,140 @@
...
+using System;
+using System.IO;
+using System.Security.Cryptography;
+
+namespace Azure.Security.KeyVault.Keys.Cryptography
+{
+    /// <summary>
+    /// Provides an implementation of the RSA algorithm backed by Azure Key Vault.
+    /// </summary>
+    internal class RSAKeyVault : RSA
```
heaths, line 13: Talking with @clairenovotny, it'd be nice to expose the public key as well.

heaths: Talking with @schaabs it makes sense to return this class (make it public) so we don't later paint ourselves in a corner if we choose to add functionality like exposing the public key, which you could get from the certificate now anyway.

sdk/keyvault/Azure.Security.KeyVault.Keys/src/Cryptography/CryptographyClient.cs
```diff
@@ -1592,6 +1601,64 @@ public virtual VerifyResult VerifyData(SignatureAlgorithm algorithm, Stream data
...
+        /// <exception cref="RequestFailedException">The server returned an error. See <see cref="Exception.Message"/> for details returned from the server.</exception>
+        public virtual RSAKeyVault CreateRSA(CancellationToken cancellationToken = default)
+        {
+            using DiagnosticScope scope = _pipeline.CreateScope($"{nameof(CryptographyClient)}.{nameof(CreateRSA)}");
+            scope.AddAttribute("key", _keyId);
+            scope.Start();
+
+            try
+            {
+                Initialize(GetOperation, cancellationToken);
```
christothes, line 1622: Why not this, like the other methods?
```suggestion
                Initialize(nameof(CreateRSA), cancellationToken);
```

sdk/keyvault/Azure.Security.KeyVault.Keys/src/Cryptography/RSAKeyVault.cs
```diff
@@ -0,0 +1,189 @@
...
+        public override void ImportParameters(RSAParameters parameters) => throw new NotSupportedException("Cannot import a key");
+
+        /// <summary>
+        /// Encrypts the input data using the specified padding mode.
+        /// </summary>
+        /// <param name="data">The data to encrypt.</param>
+        /// <param name="padding">The padding mode.</param>
+        /// <returns>The encrypted data.</returns>
+        /// <exception cref="NotSupportedException">The <paramref name="padding"/> is not supported.</exception>
+        /// <exception cref="RequestFailedException"></exception>
```
christothes, line 70: empty value

christothes (approved):

sdk/keyvault/Azure.Security.KeyVault.Keys/src/Cryptography/CryptographyClient.cs
heaths: Because I want to telemeter the `get` operation. Everything else is strictly client-side. That they called `CreateRSA` is captured just above.

Soren-Knudsen:
@christothes: Any news for when [4.6.0-beta.1 (Unreleased)](https://github.com/heaths/azure-sdk-for-net/blob/111e47ed904bcb95dbaba033ad8f46a4dc34fec6/sdk/keyvault/Azure.Security.KeyVault.Keys/CHANGELOG.md#460-beta1-unreleased) will be released?

heaths:
Before the end of this month.

Issue #27819, opened 2022-03-29
[BUG] Get Properties Services failed 


dioum2touba:
### Library name and version

Azure.ResourceManager.ArmClient.GetGenericResource (1.0.0-beta.7)

### Describe the bug

We have two app plan services in our azure subscription

`var properties = await armClient.GetGenericResource(new ResourceIdentifier(app.ResourceIdentifier)).GetAsync();`

When we try to run this code in a loop (over a list containing data from our two services), the first run fails and the second runs successfully.

### Expected behavior

Normally this piece of code should return the properties of the azure services (App Plan) that are in our subscription

### Actual behavior

We have two app plan services in our azure subscription that we wanted to retrieve their properties with Azure ResourceManager SDK with version 1.0.0-beta.7.

`var properties = await armClient.GetGenericResource(new ResourceIdentifier(app.ResourceIdentifier)).GetAsync();`

When we try to run this code in a loop (over a list containing data from our two services), the first run fails and the second runs successfully.

To solve this problem, we used the GET API https://management.azure.com/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Web/serverfarms/{name}?api-version= 2021-02-01 which works fine.

### Reproduction Steps

`var properties = await armClient.GetGenericResource(new ResourceIdentifier(app.ResourceIdentifier)).GetAsync();`

### Environment

Information about our environment:

.Net SDK: 6.x.x

C language#

Framework: .Net

Operating System: Windows 10

Project Type: Azure Functions

Company: Expertime

jsquire:
Thank you for your feedback.  Tagging and routing to the team member best able to assist.

ArthurMa1978:
@ArcturusZhang  please help to check if our new release has upgraded to right version

ArcturusZhang:
Hi @dioum2touba thanks for this issue.
I can reproduce your issue on my side using the resource id of `serverfarms`.

Our "GenericResource" API has the ability to automatically fetch the latest api-version of the corresponding resource type and use it immediately in the request.
[...]

PR #33196, opened 2022-12-26
[GenericResources] Update API version lookup to be case-insensitive

abhiroop-goyal:
**Why**
While using the GenericResources API collection, the SDK caches the available API versions for the resource type by invoking the   equivalent of GET /providers/Microsoft.<Namespace>

The API version lookup for a specific resource type is case sensitive, which is throwing an exception if there is a case mismatch between user input and the dictionary contents.

Issue - #27819
Sample PR - #27944

**What**
- Resources API Version dictionary has been modified to become case insensitive.


Yao725 (approved):
LGTM, this should also fix https://github.com/Azure/azure-sdk-for-net/issues/33293.

m-nash (approved):

Yao725:
PR is merged, thanks for your contribution! @abhiroop-goyal

PR #33216, opened 2022-12-28
fix(cadl): 3rd party exec error should be returned

archerzz:
## Description

PowerShell does not stop and throw error if 3rd party exec returns error exit code. Some places in our CADL build scripts do not check the exit code of 3rd party exec, so that the build process will not stop on error.


eng/scripts/Cadl-Project-Generate.ps1
```diff
@@ -56,6 +56,7 @@ try {
     NpmInstallForProject $npmWorkingDir
     Write-Host("npx cadl compile $mainCadlFile --emit `"`@azure-tools/cadl-csharp`" --output-path `"$resolvedProjectDirectory`"")
     npx cadl compile $mainCadlFile --emit "@azure-tools/cadl-csharp" --output-path "$resolvedProjectDirectory"
+    if ($LASTEXITCODE) { exit $LASTEXITCODE }
```
chunyu3, line 59: Better to give out an error message.

chunyu3 (approved):

archerzz:
wait until the fix for `adp` is merged

eng/scripts/Cadl-Project-Generate.ps1
archerzz: The `npx` command will print error message, so here I just exit.

Issue #33261, opened 2023-01-02
[Bug] New-TestResources.ps1 Scripts fails when $BaseName > 24 characters


NiveditJain:
While working with Azure Loadtestservice SDK, script New-TestResource.ps1 was failing as $BaseName for me was, "niveditjainloadtestservice" which is greater than 24 characters. It was showing this behaviour because Azure KeyVault Resource should have a name between 3 and 24 Characters as mentioned [here](https://azure.github.io/PSRule.Rules.Azure/en/rules/Azure.KeyVault.Name/).

Screenshot of the error:
![too long issue SS](https://user-images.githubusercontent.com/40313233/210198333-e59450c5-9a3d-4c90-8b9f-eb1d1827f40b.png)

benbp:
Fixed by https://github.com/Azure/azure-sdk-tools/pull/5242

PR #33262, opened 2023-01-02
Fix $base name too long

NiveditJain:
Fixed the issue: https://github.com/Azure/azure-sdk-for-net/issues/33261 by truncating the $BaseName up to 24 characters if it is more than 24 characters. This fixes the problems mentioned in issue #33261

### DIFF
--- a/eng/common/TestResources/SubConfig-Helpers.ps1
+++ b/eng/common/TestResources/SubConfig-Helpers.ps1
@@ -20,7 +20,6 @@ function GetBaseName([string]$user, [string]$serviceDirectoryName) {
     # Handle service directories in nested directories, e.g. `data/aztables`
     $serviceDirectorySafeName = $serviceDirectoryName -replace '[\./\\]', ''
     $BaseName =  "$user$serviceDirectorySafeName".ToLowerInvariant()
-    $BaseName = "loadtestingSDK"
     # Key-Vault accounts have a limit of 24 characters
     if ($BaseName.Length -gt 24) {
         Log "Truncating BaseName to 24 characters."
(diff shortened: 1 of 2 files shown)


eng/common/TestResources/SubConfig-Helpers.ps1
```diff
@@ -19,7 +19,15 @@ function GetUserName() {
 function GetBaseName([string]$user, [string]$serviceDirectoryName) {
     # Handle service directories in nested directories, e.g. `data/aztables`
     $serviceDirectorySafeName = $serviceDirectoryName -replace '[\./\\]', ''
-    return "$user$serviceDirectorySafeName".ToLowerInvariant()
+    $BaseName =  "$user$serviceDirectorySafeName".ToLowerInvariant()
+    $BaseName = "loadtestingSDK"
+    # Key-Vault accounts have a limit of 24 characters
+    if ($BaseName.Length -gt 24) {
```
jsquire, line 25: It would be better here to throw.  Truncating may cause name collisions and result in failures that are less clear.

NiveditJain: @jsquire it does make sense truncating can cause the issues listed, however, I wanted to understand that if we throw it here, then how would users like me be doing it? because when we run tests this script is also invoked and it might be a problem. 

How about we ask for input for BaseName as a fallback instead of truncating?

jsquire: Fair point, I forgot about that scenario.  Probably better that we validate the parameter upfront in the test resources script, when provided, and fallback here such that if we exceeded the limit, we truncate.

weshaggard:
Changes like this under eng/common cannot be made independently in a specific repo they need to be made in the azure-sdk-tools repo, see https://github.com/Azure/azure-sdk-tools/blob/main/doc/common/common_engsys.md

benbp:
Closing as this is fixed in https://github.com/Azure/azure-sdk-tools/pull/5242. FYI @NiveditJain

PR #33281, opened 2023-01-03
Make retrypolicy internal

JoshLove-msft:
Also renamed ProcessingContext and updated change log.


annelo-msft (approved):

danielmarbach:
It would be awesome to have some context on this PR about the why of this change. Would you mind adding or linking to it?

jsquire (approved):

JoshLove-msft:
> It would be awesome to have some context on this PR about the why of this change. Would you mind adding or linking to it?

Sure, the `RetryPolicy` type was in-flight work that is not ready to actually be shipped as a public type. Hence we are reverting it to internal in order to unblock the Azure.Core release containing other updates.

PR #33284, opened 2023-01-04
Fix for Share DeleteIfExists/Exists Error Live tests

amnguye:
Fix for more reliable tests when running live, these error codes are more reliable.


amnguye:
Some live tests may fail due to other flakey tests. The live tests we're aiming to pass are ExistsAsync_Error and DeleteExistsAsync_Error.

seanmcc-msft (approved):

PR #33310, opened 2023-01-05
build(template): remove global suppression of azc0002

### LINKED ISSUE #4715: Update AZC0002 to look for RequestOptions as well
Since cancellation token can be passed through the RequestContext for protocol methods we need to update this check to account for that.

https://github.com/Azure/azure-sdk-for-net/pull/32466/files#diff-4af651fe17380560084e5cb68456401c225c9a3c3f511049bbe8144e8c4eae08R6

archerzz:
We don't need to suppress it anymore, since the rule AZC0002 has been updated to include `RequestContext` as a valid last parameter. See:
- Latest implemetation of `AZC0002`: https://github.com/Azure/azure-sdk-tools/blob/17876d209b476d853f17b093a8012bea552f8354/src/dotnet/Azure.ClientSdk.Analyzers/Azure.ClientSdk.Analyzers/ClientMethodsAnalyzer.cs#L28-L29
- Corresponding commit: https://github.com/Azure/azure-sdk-tools/commit/17876d209b476d853f17b093a8012bea552f8354#diff-9877e11c2950c55f3d6c2af6cd8ef36fff8c17c44173a54c356b06247c016985R29

resolve https://github.com/Azure/azure-sdk-tools/issues/4715

### DIFF
--- a/sdk/template/.content/src/GlobalSuppressions.cs
+++ b/sdk/template/.content/src/GlobalSuppressions.cs
@@ -1,7 +1,0 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License.
-
-using System.Diagnostics.CodeAnalysis;
-
-// add global suppression here
-// [assembly: SuppressMessage(category, checkId)]
--- a/sdk/template/.content/src/GlobalSuppressions.cs
+++ b/sdk/template/.content/src/GlobalSuppressions.cs
@@ -3,4 +3,5 @@
 
 using System.Diagnostics.CodeAnalysis;
 
-[assembly: SuppressMessage("Usage", "AZC0002:DO ensure all service methods, both asynchronous and synchronous, take an optional CancellationToken parameter called cancellationToken.", Justification = "CancellationToken can be passed through RequestOptions")]
+// add global suppression here
+// [assembly: SuppressMessage(category, checkId)]


sdk/template/.content/src/GlobalSuppressions.cs
```diff
@@ -3,4 +3,5 @@
 
 using System.Diagnostics.CodeAnalysis;
 
-[assembly: SuppressMessage("Usage", "AZC0002:DO ensure all service methods, both asynchronous and synchronous, take an optional CancellationToken parameter called cancellationToken.", Justification = "CancellationToken can be passed through RequestOptions")]
\ No newline at end of file
+// add global suppression here
+// [assembly: SuppressMessage(category, checkId)]
```
annelo-msft, line 7: Do we need this line?  Please remove these comments.  Ideally the template would indicate that we don't suppress warnings in our libraries.

annelo-msft: Actually, would it be possible to delete this file entirely, if there are no suppressions needed anymore?  It is a goal for our libraries that GA libraries don't ship with a GlobalSuppressions file - this should be reflected in the template as well.  Thanks!

archerzz: @annelo-msft Got it. Previously I thought we'd better to leave a place holder.

heaths (approved):

PR #33351, opened 2023-01-07
[Perf] Refactor tests.yml into one file per package per language

mikeharder:
- Depends on Azure/azure-sdk-tools#5083


sdk/keyvault/Azure.Security.KeyVault.Secrets/perf-tests.yml
```diff
@@ -0,0 +1,19 @@
+Service: keyvault-secrets
```
heaths, line 1: Why a separate file for these now? Why not other KV tests? And why is the `Class` name below actually the perf test method name?

mikeharder: The point of this PR (and related PR Azure/azure-sdk-tools#5083), is to refactor the one huge `tests.yml` in the tools repo to one file per service per language.  This allows you to add or modify the perf tests in a single commit to your language repo, rather than needing to coordinate the changes across the language and tools repo.

The reason we only have a file for Keyvault.Secrets but not the other KV SDKs, is because we haven't actually enabled perf pipelines for KV yet.  It requires adding test-proxy to our perf pipelines, which is a fairly large work item we haven't done yet.  We did have a list of tests for Keyvault.Secrets in the shared `tests.yml`, so I ported them to this file so they are not lost.

In the YML, the `Test` field is the generic name for the test, that is consistent across languages so we can compare results.  The `Class` field is the language-specific name of the class implementing the test.  I don't love the name `Class` but I couldn't think of anything better.

mikeharder: The code in our perf test projects does not provide all the metadata needed by the pipelines:

* Which package versions should be tested
* Which tests should be run vs ignored, and with what arguments

This metadata could potentially be added to the perf test projects themselves, say by custom attributes in .NET.  However, this would need to be implemented differently in each language, which is a lot more work.

The `perf-tests.yml` file provides a simple and language-agnostic way to provide this metadata to the perf automation.  It also allows you to modify this metadata without changing your perf test project or needing to recompile it.

heaths: > Which package versions should be tested

Following the guidelines you set forth a while back, I should think we only put hero scenarios in perf test projects already, therefore the perf framework should run all of them. The burden of this extra file isn't just a matter of cost, but maintaining changes across multiple files now. So if someone changes a test method they put in a perf project to actually measure perf for code therein, they also have to remember to update this other file?

Are these few files just to test out your plan then?

At least for .NET, long term I'd rather spend a couple hours building a solution to generate these files (perhaps like how we do snippets). /cc @JoshLove-msft and @christothes for their thoughts.

heaths: Conversely, every time I add a perf test I have to remember to update this file. It's not only overhead, but error prone.

christothes: To add to Heath's concerns, it's not clear to me when we'd need to modify a perf test per language. Do we have examples today? Intuitively, I would think that snapping to latest version and running all tests would be what we usually want.

mikeharder: These questions are about where to encode the metadata required by the perf pipelines (and the PerfAutomation app).  Since this impacts all languages, let's consolidate discussion in https://github.com/Azure/azure-sdk-tools/pull/5083.

heaths (approved):

PR #33352, opened 2023-01-07
[Text Analytics] Fix error parsing in TextAnalyticsFailedDetailsParser

### LINKED ISSUE #33103: [TextAnalytics] DetectLanguageWithErrorCountryHintTest failing in nightly runs
TextAnalytics nightly test runs are failing with:

> Error message
> Expected: <Azure.RequestFailedException>
> But was:  <System.ArgumentNullException: Value cannot be null.
> Parameter name: errorCode
> at Azure.Core.Argument.AssertNotNullOrEmpty(String value, String name) in D:\a_work\1\s\sdk\core\Azure.Core\src\Shared\Argument.cs:line 117
> at Azure.AI.TextAnalytics.TextAnalyticsErrorCode..ctor(String errorCode) in D:\a_work\1\s\sdk\textanalytics\Azure.AI.TextAnalytics\src\TextAnalyticsErrorCode.cs:line 89
> at Azure.AI.TextAnalytics.TextAnalyticsErrorCode.op_Implicit(String errorCode) in D:\a_work\1\s\sdk\textanalytics\Azure.AI.TextAnalytics\src\TextAnalyticsErrorCode.cs:line 98
> at Azure.AI.TextAnalytics.TextAnalyticsError..ctor(String code, String message, String target) in D:\a_work\1\s\sdk\textanalytics\Azure.AI.TextAnalytics\src\TextAnalyticsError.cs:line 13
> at Azure.AI.TextAnalytics.Transforms.ConvertToError(Error error) in D:\a_work\1\s\sdk\textanalytics\Azure.AI.TextAnalytics\src\Transforms.cs:line 24
> at Azure.Core.Pipeline.TextAnalyticsFailedDetailsParser.GetResponseError(List1 errors, ResponseError& responseError, IDictionary2& data) in D:\a_work\1\s\sdk\textanalytics\Azure.AI.TextAnalytics\src\TextAnalyticsFailedDetailsParser.cs:line 84
> at Azure.Core.Pipeline.TextAnalyticsFailedDetailsParser.TryParse(Response response, ResponseError& error, IDictionary2& data) in D:\a\_work\1\s\sdk\textanalytics\Azure.AI.TextAnalytics\src\TextAnalyticsFailedDetailsParser.cs:line 49 at Azure.RequestFailedException.GetRequestFailedExceptionContent(Response response) in D:\a\_work\1\s\sdk\core\Azure.Core\src\RequestFailedException.cs:line 131 at Azure.RequestFailedException..ctor(Response response, Exception innerException) in D:\a\_work\1\s\sdk\core\Azure.Core\src\RequestFailedException.cs:line 106 at Azure.RequestFailedException..ctor(Response response) in D:\a\_work\1\s\sdk\core\Azure.Core\src\RequestFailedException.cs:line 97 at Azure.AI.TextAnalytics.ServiceClients.LegacyServiceClient.DetectLanguage(String document, String countryHint, CancellationToken cancellationToken) in D:\a\_work\1\s\sdk\textanalytics\Azure.AI.TextAnalytics\src\ServiceClients\LegacyServiceClient.cs:line 129 at Azure.AI.TextAnalytics.TextAnalyticsClient.DetectLanguage(String document, String countryHint, CancellationToken cancellationToken) in D:\a\_work\1\s\sdk\textanalytics\Azure.AI.TextAnalytics\src\TextAnalyticsClient.cs:line 199 --- End of stack trace from previous location where exception was thrown --- at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw() at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task) at System.Runtime.CompilerServices.TaskAwaiter1.GetResult()
> at NUnit.Framework.Internal.TaskAwaitAdapter.GenericAdapter1.GetResult() at NUnit.Framework.Internal.AsyncToSyncAdapter.Await(Func1 invoke)
> at NUnit.Framework.Assert.ThrowsAsync(IResolveConstraint expression, AsyncTestDelegate code, String message, Object[] args)>
> 
> 
> Stack trace
>    at Azure.AI.TextAnalytics.Tests.DetectLanguageTests.DetectLanguageWithErrorCountryHintTest() in D:\a\_work\1\s\sdk\textanalytics\Azure.AI.TextAnalytics\tests\DetectLanguageTests.cs:line 96
> 
> 1)    at Azure.AI.TextAnalytics.Tests.DetectLanguageTests.DetectLanguageWithErrorCountryHintTest() in D:\a\_work\1\s\sdk\textanalytics\Azure.AI.TextAnalytics\tests\DetectLanguageTests.cs:line 96

For more details check here:

- https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2060244&view=results

@jsquire for notification.

joseharriaga:
Resolves: https://github.com/Azure/azure-sdk-for-net/issues/33103

Pull Request https://github.com/Azure/azure-sdk-for-net/pull/33001 added the `TextAnalyticsFailedDetailsParser` class but introduced a small regression that was caught by the `DetectLanguageWithErrorCountryHintTest` test when called using API versions 3.0 and 3.1. 

With newer API versions, the response returned by the service looks like this:
```json
{
    "kind": "LanguageDetectionResults",
    "results": {
        "documents": [],
        "errors": [
            {
                "id": "0",
                "error": {
                    "code": "InvalidArgument",
                    "message": "Invalid Country Hint.",
                    "innererror": {
                        "code": "InvalidCountryHint",
                        "message": "Country hint is not valid. Please specify an ISO 3166-1 alpha-2 two letter country code."
                    }
                }
            }
        ],
        "modelVersion": "2022-10-01"
    }
}
```

This case is appropriately handled by the current implementation of the `TryParse` method of the `TextAnalyticsFailedDetailsParser` class. 

However, in the case of API versions 3.0 and 3,1, the response looks like this instead:
```json
{
    "documents": [],
    "errors": [
        {
            "id": "0",
            "error": {
                "code": "InvalidArgument",
                "message": "Invalid Country Hint.",
                "innererror": {
                    "code": "InvalidCountryHint",
                    "message": "Country hint is not valid. Please specify an ISO 3166-1 alpha-2 two letter country code."
                }
            }
        }
    ],
    "modelVersion": "2022-10-01"
}
```
The current implementation of `TryParse` assumes that when it encounters a property called "errors" that is not wrapped by a property called "results", then these "errors" can be parsed into a collection of `Error` objects. While this is correct in some cases, in other cases they could actually be `InputError` objects or `DocumentError` objects depending on the context. Consequently, the deserialization is done incorrectly, and things fail soon after.

Now, since the `InputError` and `DocumentError` entities have a property called "error" that does in fact correspond to the `Error` object that we are looking for, the fix is simply to check for the existence of this property before choosing what to deserialize.

### DIFF
--- a/sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsFailedDetailsParser.cs
+++ b/sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsFailedDetailsParser.cs
@@ -22,30 +22,39 @@ namespace Azure.Core.Pipeline
 
             if (response.ContentStream is { CanSeek: true })
             {
-                var position = response.ContentStream.Position;
+                long position = response.ContentStream.Position;
+
                 try
                 {
-                    // Try to parse the failure content and use that as the
+                    // Try to extract the standard Azure Error object from the response so that we can use it as the
                     // default value for the message, error code, etc.
 
                     response.ContentStream.Position = 0;
                     using JsonDocument doc = JsonDocument.Parse(response.ContentStream);
                     if (doc.RootElement.TryGetProperty("error", out JsonElement errorElement))
                     {
-                        var textAnalyticsError = Transforms.ConvertToError(Error.DeserializeError(errorElement));
+                        TextAnalyticsError textAnalyticsError = Transforms.ConvertToError(Error.DeserializeError(errorElement));
                         error = new ResponseError(textAnalyticsError.ErrorCode.ToString(), textAnalyticsError.Message);
                         return true;
                     }
 
+                    // If the response does not straight up correspond to the standard Azure Error object that we are
+                    // looking for, the Error object must actually be nested somewhere in there instead. For example,
+                    // this can happen in the case of the convenience methods that receive a single input document as a
+                    // parameter instead of a list of input documents. Here, rather than returning the typical
+                    // successful response that includes a list of errors that the user needs to look through, we
+                    // want to grab the first error in that list (which inevitably corresponds to a problem with the
+                    // single input document), and use that error to throw a useful RequestFailedException. Now,
+                    // depending on the circumstances, that standard Azure Error could be inside an InputError
+                    // object, a DocumentError object, etc., so we need to look for it among a handful of well-known
+                    // cases like those.
+
                     if (doc.RootElement.TryGetProperty("errors", out JsonElement errorsElement))
                     {
-                        var errors = new List<Error>();
+                        List<Error> errors = new();
+
                         foreach (JsonElement item in errorsElement.EnumerateArray())
                         {
-                            // This element could be the Error object that we are looking for, but it could also be an
-                            // InputError or a DocumentError object that is wrapping it instead. Therefore, we must
-                            // first check if the element has a property called "error". If it does, the value would
-                            // correspond to the actual Error object.
                             if (item.TryGetProperty("error", out errorElement))
                             {
                                 errors.Add(Error.DeserializeError(errorElement));
@@ -62,13 +71,10 @@ namespace Azure.Core.Pipeline
 
                     if (doc.RootElement.TryGetProperty("results", out JsonElement results) && results.TryGetProperty("errors", out errorsElement))
                     {
-                        var errors = new List<Error>();
+                        List<Error> errors = new();
+
                         foreach (JsonElement item in errorsElement.EnumerateArray())
                         {
-                            // This element could be the Error object that we are looking for, but it could also be an
-                            // InputError or a DocumentError object that is wrapping it instead. Therefore, we must
-                            // first check if the element has a property called "error". If it does, the value would
-                            // correspond to the actual Error object.
                             if (item.TryGetProperty("error", out errorElement))
                             {
                                 errors.Add(Error.DeserializeError(errorElement));
--- a/sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsFailedDetailsParser.cs
+++ b/sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsFailedDetailsParser.cs
@@ -4,7 +4,6 @@
 using System.Collections.Generic;
 using System.Text.Json;
 using Azure.AI.TextAnalytics;
-using Azure.AI.TextAnalytics.Legacy;
 using Azure.AI.TextAnalytics.Models;
 
 #nullable enable
@@ -38,24 +37,46 @@ namespace Azure.Core.Pipeline
                         return true;
                     }
 
-                    if (doc.RootElement.TryGetProperty("errors", out JsonElement errorElements))
+                    if (doc.RootElement.TryGetProperty("errors", out JsonElement errorsElement))
                     {
                         var errors = new List<Error>();
-                        foreach (var item in errorElements.EnumerateArray())
+                        foreach (JsonElement item in errorsElement.EnumerateArray())
                         {
-                            errors.Add(Error.DeserializeError(item));
+                            // This element could be the Error object that we are looking for, but it could also be an
+                            // InputError or a DocumentError object that is wrapping it instead. Therefore, we must
+                            // first check if the element has a property called "error". If it does, the value would
+                            // correspond to the actual Error object.
+                            if (item.TryGetProperty("error", out errorElement))
+                            {
+                                errors.Add(Error.DeserializeError(errorElement));
+                            }
+                            else
+                            {
+                                errors.Add(Error.DeserializeError(item));
+                            }
                         }
 
                         GetResponseError(errors, out error, out data);
                         return true;
                     }
 
-                    if (doc.RootElement.TryGetProperty("results", out JsonElement results) && results.TryGetProperty("errors", out errorElements))
+                    if (doc.RootElement.TryGetProperty("results", out JsonElement results) && results.TryGetProperty("errors", out errorsElement))
                     {
                         var errors = new List<Error>();
-                        foreach (var item in errorElements.EnumerateArray())
+                        foreach (JsonElement item in errorsElement.EnumerateArray())
                         {
-                            errors.Add(InputError.DeserializeInputError(item).Error);
+                            // This element could be the Error object that we are looking for, but it could also be an
+                            // InputError or a DocumentError object that is wrapping it instead. Therefore, we must
+                            // first check if the element has a property called "error". If it does, the value would
+                            // correspond to the actual Error object.
+                            if (item.TryGetProperty("error", out errorElement))
+                            {
+                                errors.Add(Error.DeserializeError(errorElement));
+                            }
+                            else
+                            {
+                                errors.Add(Error.DeserializeError(item));
+                            }
                         }
 
                         GetResponseError(errors, out error, out data);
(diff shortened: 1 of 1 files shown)


sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsFailedDetailsParser.cs
```diff
@@ -38,24 +37,46 @@ public override bool TryParse(Response response, out ResponseError? error, out I
                         return true;
                     }
 
-                    if (doc.RootElement.TryGetProperty("errors", out JsonElement errorElements))
+                    if (doc.RootElement.TryGetProperty("errors", out JsonElement errorsElement))
                     {
                         var errors = new List<Error>();
-                        foreach (var item in errorElements.EnumerateArray())
+                        foreach (JsonElement item in errorsElement.EnumerateArray())
```
heaths, line 43: 1. Could this even be a heterogenous array? Swaggers generally shouldn't be, so checking each item is more expensive.
2. If the swaggers dictate the "errors" array is different for legacy vs. Language APIs, perhaps this parser should instead be bifurcated on the api-version to support strict deserialization, which is what our generated deserialization (and, generally, even hand-written deserialization) does. This is a rather loose deserialization that could result in unforeseen problems in client code.

heaths (changes requested):

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsFailedDetailsParser.cs
joseharriaga: 1. The array is homogeneous in today's implementation, so checking one single item would be sufficient. I did it like this because it was cleaner, and I assumed that the array would only have a few items (less than ten?) and that `TryGetProperty` was a relatively cheap operation. Please correct me if my assumptions are incorrect; I can change it to check only once.
2. My first idea was actually to create a separate class derived from `RequestFailedDetailsParser` only for the `LegacyServiceClient` to achieve something similar to what you're proposing. However, if I understand correctly from looking at the swaggers, I noticed that both the Legacy and Language APIs have cases where the items of the "errors" property are the "Error"/"TextAnalyticsError" that we're looking for, and cases where the items are actually "InputErrors"/"DocumentErrors", which themselves have an "error" property that corresponds to the "Error"/"TextAnalyticsError" that we're looking for. In other words, my understanding is that both the Legacy and Language APIs need this check.

Examples in the Legacy API:
- https://github.com/Azure/azure-rest-api-specs/blob/4b7373a29b0cc90e654ca66a8ca3742affca732d/specification/cognitiveservices/data-plane/TextAnalytics/stable/v3.1/TextAnalytics.json#L2089
- https://github.com/Azure/azure-rest-api-specs/blob/4b7373a29b0cc90e654ca66a8ca3742affca732d/specification/cognitiveservices/data-plane/TextAnalytics/stable/v3.1/TextAnalytics.json#L1462

Examples in the Language API:
- https://github.com/Azure/azure-rest-api-specs/blob/4b7373a29b0cc90e654ca66a8ca3742affca732d/specification/cognitiveservices/data-plane/Language/preview/2022-10-01-preview/common.json#L276
- https://github.com/Azure/azure-rest-api-specs/blob/4b7373a29b0cc90e654ca66a8ca3742affca732d/specification/cognitiveservices/data-plane/Language/preview/2022-10-01-preview/common.json#L380

cc @AlexanderSher, in case he has any insights from working on the original change.

heaths: 1. If it's only a few, sure, that's not a big deal but from what I've seen in past examples of large batches, it could be a lot.
2. More importantly - and this effectively solves 1 anyway - you pointed to different model properties with different types, so my concern here still stands: you're using a heterogeneous deserializer when the swaggers are quite specific in what models can be contained within the array. We don't do that in generated code so I'm not sure why we'd want to do that here as well. @tg-msft what do you think? Should we have a class that can deserialize several different types and could return unexpected results if the model definition ever changes?

joseharriaga: I'll let @AlexanderSher speak for this approach.

The way I understand it is that this method is trying to find the `Error` object in the response so that it can use it to appropriately format the `RequestFailedException` that it is attempting to throw. In the case of Text Analytics, the tricky part is that an error response can come in different formats and the `Error` object can be nested somewhere in there, which is why this code is doing different things to try to find it. 

Looking at the original PR, it seems like a lot of this used to be done "manually" in place, which differs from the new approach where it seems that the expectation is that you just new up an `RequestFailedException` with nothing but the raw response and then this `TryParse` method will figure it out for you:
https://github.com/Azure/azure-sdk-for-net/pull/33001/files#diff-9769495e498d6504a088de4876816df2266b5e5a831acf8d3c0b38eb3ceb6278L88

This means that the `TryParse` method needs to be very smart since it doesn't have the context of where the error response is coming from. I wonder if there's a way to give it more context.

heaths: Document errors shouldn't throw, though. We collect them and store as an array to be retrieved by the user. My concern, though, stems from what you said (and what the swagger documents),

> an error response can come in different formats and the Error object can be nested somewhere in there

Exactly, hence they are not the same model. Similar, yes, and if a generic error parser is capable of deserializing part of that model by all means use it! :smile: But none of our code - certainly not our generated code - is so loose as to potentially deserialize into the wrong type which 1) may through a runtime error like a `InvalidCastException` (or whatever might be thrown if trying to add the wrong type to a collection, for example), or 2) get added to a collection but cause downstream issues because the type / property set was not expected.

For discriminated types, for example, we have separate classes even if those derivative types add no new properties, for example. In a similar vein, I think we should have separate classes or methods parsing different expected models in these cases. This would also help ensure we don't deserialize any properties we don't know, which is part of the contract we uphold today in all our deserialization (by design).

joseharriaga: Document errors do throw in the case of the convenience methods that receive a single document as input (as opposed to the methods that receive an array of documents).

tg-msft: What status code does the service return in these cases?  I'd expect this code path to run through the generated deserialization logic, not the `TryParse` exception code flow.

I kind of worry we're trying to be _too_ smart in the `TryParse` method here - especially with different payload formats across versions.  Ideally we'd let the generated code handle this and deserialize exactly what the swagger includes for a particular operation.  If that's not possible for some reason, then yes, I think we'll have to give the parser a lot more info to make a deterministic choice per version/operation rather than probe the shape.

heaths (approved):
After an offline conversation with Jose, I better understand the intent here (and was likely misreading it before). An error through this code path either *is* or *has* a standard-looking Azure error response object, which is all this is parsing.

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsFailedDetailsParser.cs
joseharriaga: I added a comment to the code to try to clarify what this method is doing for posterity. In sum, this method is just trying to find the standard Azure `Error` object in the response so that it can throw a useful `RequestFailedException`.

To do this, it must be aware of a special case: The convenience methods that receive a single input document as a parameter instead of a list of input documents. Here, rather than returning the typical successful (200 OK) response that includes a list of errors that the user needs to look through, we want to grab the first error in that list (which inevitably corresponds to a problem with the single input document) and use that error to throw a useful `RequestFailedException`. Now, depending on the circumstances, that standard Azure `Error` could be inside an `InputError` object, a `DocumentError` object, etc., so we need to look for it among a handful of well-known cases like those (which are all documented in the swagger spec).

PR #33359, opened 2023-01-09
[Sql] Polish the API and fix inheritance bug

Yao725:


sdk/sqlmanagement/Azure.ResourceManager.Sql/src/Custom/VirtualClusterResource.cs
```diff
@@ -31,6 +31,7 @@ public partial class VirtualClusterResource
         /// Operation Id: VirtualClusters_UpdateDnsServers
         /// </summary>
         /// <param name="cancellationToken"> The cancellation token to use. </param>
+        [ForwardsClientCalls]
         public virtual async Task<Response<ManagedInstanceUpdateDnsServersOperationData>> UpdateDnsServersAsync(CancellationToken cancellationToken = default) =>
             await (await UpdateDnsServersAsync(WaitUntil.Started, cancellationToken).ConfigureAwait(false)).WaitForCompletionAsync(cancellationToken).ConfigureAwait(false);
```
fengzhou-msft, line 36: For line 36, why not use `WaitUntil.Completed` directly?

fengzhou-msft (approved):

ArthurMa1978 (approved):

sdk/sqlmanagement/Azure.ResourceManager.Sql/src/Custom/VirtualClusterResource.cs
Yao725: If we use WaitUntil.Completed, we need to construct the Response<T> by ourselves like this:
```csharp
public virtual async Task<Response<ManagedInstanceUpdateDnsServersOperationData>> UpdateDnsServersAsync(CancellationToken cancellationToken = default)
{
    var operation = await UpdateDnsServersAsync(WaitUntil.Completed, cancellationToken).ConfigureAwait(false);
    return Response.FromValue(operatoin.Value, operatoin.GetRawResponse());
}
```
Personally, I prefer the first implementation so I use `WaitUntil.Started` here.

PR #33383, opened 2023-01-10
Added missing testcases, created recordings

ostoliarova-ee:
This pull requests addresses the story [User Story 3093174](https://skype.visualstudio.com/SPOOL/_workitems/edit/3093174): [SDK][Identity][.NET] Capture missing key customer scenarios in test cases

The following scenarios were captured:
Create a new user & token in a single request for scope VOIP
Create a new user & token in a single request for multiple scopes [VOIP, CHAT]
Get token for an existing user for scope VOIP

### DIFF
--- a/sdk/communication/Azure.Communication.Common/tests/CommunicationTestEnvironment.cs
+++ b/sdk/communication/Azure.Communication.Common/tests/CommunicationTestEnvironment.cs
@@ -18,7 +18,7 @@ namespace Azure.Communication.Tests
 
         public string LiveTestDynamicConnectionString => GetRecordedVariable(
             LiveTestDynamicConnectionStringEnvironmentVariableName,
-            options => options.HasSecretConnectionStringParameter("accessKey", SanitizedValue.Base64));
+            options => options.IsSecret("endpoint=https://sanitized.communication.azure.com/;accesskey=Kg=="));
 
         public Uri LiveTestDynamicEndpoint => new(Core.ConnectionString.Parse(LiveTestDynamicConnectionString).GetRequired("endpoint"));
 
--- a/sdk/communication/Azure.Communication.Identity/tests/Azure.Communication.Identity.Tests.csproj
+++ b/sdk/communication/Azure.Communication.Identity/tests/Azure.Communication.Identity.Tests.csproj
@@ -27,4 +27,5 @@
     <None Include="..\samples\Sample1_CommunicationIdentityClient.md" Link="samples\Sample1_CommunicationIdentityClient.md" />
     <None Include="..\samples\Sample1_CommunicationIdentityClientAsync.md" Link="samples\Sample1_CommunicationIdentityClientAsync.md" />
   </ItemGroup>
+  <ProjectExtensions><VisualStudio><UserProperties sessionrecords_4communicationidentityclientlivetests_4createuserandtokenwithchatscope_1json__JsonSchema="https://download.stackhawk.com/hawk/jsonschema/hawkconfig.json" /></VisualStudio></ProjectExtensions>
 </Project>
--- a/sdk/communication/Azure.Communication.Identity/tests/Azure.Communication.Identity.Tests.csproj
+++ b/sdk/communication/Azure.Communication.Identity/tests/Azure.Communication.Identity.Tests.csproj
@@ -27,5 +27,4 @@
     <None Include="..\samples\Sample1_CommunicationIdentityClient.md" Link="samples\Sample1_CommunicationIdentityClient.md" />
     <None Include="..\samples\Sample1_CommunicationIdentityClientAsync.md" Link="samples\Sample1_CommunicationIdentityClientAsync.md" />
   </ItemGroup>
-  <ProjectExtensions><VisualStudio><UserProperties sessionrecords_4communicationidentityclientlivetests_4createuserandtokenwithchatscope_1json__JsonSchema="https://download.stackhawk.com/hawk/jsonschema/hawkconfig.json" /></VisualStudio></ProjectExtensions>
 </Project>
--- a/sdk/communication/Azure.Communication.Identity/tests/CommunicationIdentityClient/CommunicationIdentityClientLiveTests.cs
+++ b/sdk/communication/Azure.Communication.Identity/tests/CommunicationIdentityClient/CommunicationIdentityClientLiveTests.cs
@@ -57,11 +57,14 @@ namespace Azure.Communication.Identity.Tests
         }
 
         [Test]
-        [TestCase(AuthMethod.ConnectionString, "chat", TestName = "GettingTokenWithSingleScopeWithConnectionString")]
-        [TestCase(AuthMethod.KeyCredential, "chat", TestName = "GettingTokenWithSingleScopeWithKeyCredential")]
-        [TestCase(AuthMethod.TokenCredential, "chat", TestName = "GettingTokenWithSingleScopeWithTokenCredential")]
+        [TestCase(AuthMethod.ConnectionString, "chat", TestName = "GettingTokenWithSingleChatScopeWithConnectionString")]
+        [TestCase(AuthMethod.ConnectionString, "voip", TestName = "GettingTokenWithSingleVoIPScopeWithConnectionString")]
         [TestCase(AuthMethod.ConnectionString, "chat", "voip", TestName = "GettingTokenWithMultipleScopesWithConnectionString")]
+        [TestCase(AuthMethod.KeyCredential, "chat", TestName = "GettingTokenWithSingleChatScopeWithKeyCredential")]
+        [TestCase(AuthMethod.KeyCredential, "voip", TestName = "GettingTokenWithSingleVoIPScopeWithKeyCredential")]
         [TestCase(AuthMethod.KeyCredential, "chat", "voip", TestName = "GettingTokenWithMultipleScopesWithKeyCredential")]
+        [TestCase(AuthMethod.TokenCredential, "chat", TestName = "GettingTokenWithSingleChatScopeWithTokenCredential")]
+        [TestCase(AuthMethod.TokenCredential, "voip", TestName = "GettingTokenWithSingleVoIPScopeWithTokenCredential")]
         [TestCase(AuthMethod.TokenCredential, "chat", "voip", TestName = "GettingTokenWithMultipleScopesWithTokenCredential")]
         public async Task GetTokenGeneratesTokenAndIdentityWithScopes(AuthMethod authMethod, params string[] scopes)
         {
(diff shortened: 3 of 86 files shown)


mjafferi-msft (approved):

sdk/communication/Azure.Communication.Identity/tests/SessionRecords/CommunicationIdentityClientLiveTests/CreateUserAndTokenWithChatScope.json
```diff
@@ -0,0 +1,52 @@
...
+        },
+        "accessToken": {
+          "token": "Sanitized",
+          "expiresOn": "2023-01-10T17:56:00.7844833\u002B00:00"
+        }
+      }
+    }
+  ],
+  "Variables": {
+    "COMMUNICATION_LIVETEST_DYNAMIC_CONNECTION_STRING": "endpoint=https://smstestapp.communication.azure.com/;accesskey=Kg==",
```
AikoBB, line 49: should the URL be also sanitized?

ostoliarova-msft: Sanitizers only work for a JSON in requests and responses and not to the "Variables" field. Some of the older recordings have this value sanitized and some not. My suggestion is that at some point of time the behavior of sanitizers library changed.
I investigated how the information is processed for sanitization and discovered that a separate Kestrel server is run by a sanitizer library that performs the sanitization. The code for this server is not accessible. None of the additional sanitizers with different syntax that I've tried adding has any influence on the 'Variables' field.
However, considering that some of the old recordings already have this value and some SDKs do no sanitization at all,  this is not a secret value anymore (even if it used to be). So my suggestion is to go with the recordings that are created and discuss further if we need to have an additional story for fixing sanitizers (if this can somehow be done from the code that we own).

@petrsvihlik any ideas on this?

petrsvihlik: It's surprising to say the least. I'd expect everything in the JSON to be sanitizable. And it seems it was the case before...unless someone really has an ACS resource named sanitized.communication.azure.com :)

But it seems @AikoBB managed to get everything sanitized in this PR: https://github.com/Azure/azure-sdk-for-net/commit/fc83708c7ea92408e26d82c514e14111d261c283#diff-8c6c38ddd926f56bc5b531c52f146a83d83f1181d2dd08e7a626fe1c16d588a0 maybe she'll be able to remember and help you

sdk/communication/Azure.Communication.Identity/tests/CommunicationIdentityClient/CommunicationIdentityClientLiveTests.cs
```diff
@@ -57,11 +57,14 @@ public async Task InitVariables()
...
-        [TestCase(AuthMethod.TokenCredential, "chat", TestName = "GettingTokenWithSingleScopeWithTokenCredential")]
+        [TestCase(AuthMethod.ConnectionString, "chat", TestName = "GettingTokenWithSingleChatScopeWithConnectionString")]
+        [TestCase(AuthMethod.ConnectionString, "voip", TestName = "GettingTokenWithSingleVoIPScopeWithConnectionString")]
         [TestCase(AuthMethod.ConnectionString, "chat", "voip", TestName = "GettingTokenWithMultipleScopesWithConnectionString")]
+        [TestCase(AuthMethod.KeyCredential, "chat", TestName = "GettingTokenWithSingleChatScopeWithKeyCredential")]
+        [TestCase(AuthMethod.KeyCredential, "voip", TestName = "GettingTokenWithSingleVoIPScopeWithKeyCredential")]
         [TestCase(AuthMethod.KeyCredential, "chat", "voip", TestName = "GettingTokenWithMultipleScopesWithKeyCredential")]
+        [TestCase(AuthMethod.TokenCredential, "chat", TestName = "GettingTokenWithSingleChatScopeWithTokenCredential")]
+        [TestCase(AuthMethod.TokenCredential, "voip", TestName = "GettingTokenWithSingleVoIPScopeWithTokenCredential")]
```
petrsvihlik, line 67: I wonder what's the purpose of the `AuthMethod`. Aren't all the tests run against all auth methods automatically in the pipelines?

ostoliarova-msft: @petrsvihlik I didn't know that tests in pipelines run using all auth methods automatically. Do you have any links to docs or code to confirm this?
I've added a couple of new test cases but different auth methods were already being tested.

sdk/communication/Azure.Communication.Identity/tests/SessionRecords/CommunicationIdentityClientLiveTests/CreateUserAndTokenWithChatScope.json
ostoliarova-msft: @petrsvihlik yes, there were recent changes in the sanitizer code that I noticed when I was debugging but I couldn't find any connection to not sanitizing 'Variables'.

I can manually edit the files and replace that URI with a sanitized value but it'll break the concept of having autogenerated filed. Besides, what's the purpose of sanitizing? I thought that it's to hide sensitive information like tokens and passwords from external users. As the value in question was already exposed in our SDK and a couple of other SDKs, does it make sense to still sanitize it?

petrsvihlik: This very value might have been exposed, however, everyone uses different ACS resources to run the tests and it'd be better to have them sanitized. (We probably don't want to have store JSONs with `my-gummy-bear-service.communication.azure.com` or worse.) Can you pls quickly check with EngSys if the sanitization logic has intentionally changed?

AikoBB: I did not remember how I achieved it, but what I remember I did not do anything regarding to sanitizers,  maybe it manually updated. We can reach out to the core team on how we can include those values to be sanitized, I guess there should be a solution like in [java](https://github.com/Azure/azure-sdk-for-java/blob/11d31c5476bd61208de0f501485b38450e64d2dd/sdk/communication/azure-communication-identity/src/test/java/com/azure/communication/identity/CommunicationIdentityClientTestBase.java#L187). As we are sanitizing the resource urls I'd vote for sanitize them everywhere, whether we will do in this PR or create another US for that

sdk/communication/Azure.Communication.Identity/tests/CommunicationIdentityClient/CommunicationIdentityClientLiveTests.cs
ostoliarova-msft: @AikoBB do you know anything about automatic usage of all auth methods in pipelines? Shall we go with all the test cases or do I need to remove some of them (which ones in this case)?

ostoliarova-msft: @petrsvihlik we have different methods to create a client. However, from what I see, a one client is instantiated per each test case, and it's usually a client created with a connection string. Only one parametrized test used to be created with different Auth methods. My suggestion is that it was considered enough to test that clients created with different option, produce the same correct test result.
image

https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/communication/Azure.Communication.Identity/tests/Infrastructure/CommunicationIdentityClientLiveTestBase.cs#L35-L54

As for the purpose of `AuthMethod`, it's not possible to call a method to create a client in test parameters because of CS0120 error  (An object reference is required for the nonstatic field, method, or property 'member'), so the indication to which credential to use when creating a client is passed.

sdk/communication/Azure.Communication.Identity/tests/Azure.Communication.Identity.Tests.csproj
```diff
@@ -27,4 +27,5 @@
     <None Include="..\samples\Sample1_CommunicationIdentityClient.md" Link="samples\Sample1_CommunicationIdentityClient.md" />
     <None Include="..\samples\Sample1_CommunicationIdentityClientAsync.md" Link="samples\Sample1_CommunicationIdentityClientAsync.md" />
   </ItemGroup>
+  <ProjectExtensions><VisualStudio><UserProperties sessionrecords_4communicationidentityclientlivetests_4createuserandtokenwithchatscope_1json__JsonSchema="https://download.stackhawk.com/hawk/jsonschema/hawkconfig.json" /></VisualStudio></ProjectExtensions>
```
petrsvihlik, line 30: is this intentional?

sdk/communication/Azure.Communication.Common/tests/CommunicationTestEnvironment.cs
```diff
@@ -18,7 +18,7 @@ public class CommunicationTestEnvironment : TestEnvironment
 
         public string LiveTestDynamicConnectionString => GetRecordedVariable(
             LiveTestDynamicConnectionStringEnvironmentVariableName,
-            options => options.HasSecretConnectionStringParameter("accessKey", SanitizedValue.Base64));
+            options => options.IsSecret("endpoint=https://sanitized.communication.azure.com/;accesskey=Kg=="));
```
petrsvihlik, line 21: Since `CommunicationTestEnvironment` is shared across all modalities, can you please run a couple of sister pipelines to make sure we didn't break it for others?

sdk/communication/Azure.Communication.Identity/tests/CommunicationIdentityClient/CommunicationIdentityClientLiveTests.cs
petrsvihlik: > . However, from what I see, a one client is instantiated per each test case, and it's usually a client created with a connection string. Only one parametrized test used to be created with different Auth methods.

@mjafferi-msft have you captured this in your research? afaik, in JS, we test all methods using all types of auth. perhaps it'd be good to do that for .net as well. it'd probably be good to solve it in a more generic way than adding an attribute with AuthMethod to every test method. Perhaps we could somehow extend the text matrix? Perhaps using the Azure.Core.TestFramework.TestEnvironment? I don't know...

> As for the purpose of AuthMethod, it's not possible to call a method to create a client in test parameters because of CS0120 error (An object reference is required for the nonstatic field, method, or property 'member'), so the indication to which credential to use when creating a client is passed.

sure, this is alright.


then I suggest the following:

can you please extract this

```
CommunicationIdentityClient client = authMethod switch
            {
                AuthMethod.ConnectionString => CreateClientWithConnectionString(),
                AuthMethod.KeyCredential => CreateClientWithAzureKeyCredential(),
                AuthMethod.TokenCredential => CreateClientWithTokenCredential(),
                _ => throw new ArgumentOutOfRangeException(nameof(authMethod)),
            };
```
as a separate method `protected CommunicationIdentityClient CreateClient(AuthMethod authMethod = AuthMethod.ConnectionString)` into the CommunicationIdentityClientLiveTestBase and make the other methods private? + replace the usages properly?

petrsvihlik (changes requested):

sdk/communication/Azure.Communication.Identity/tests/Azure.Communication.Identity.Tests.csproj
ostoliarova-msft: Removed

sdk/communication/Azure.Communication.Common/tests/CommunicationTestEnvironment.cs
ostoliarova-msft: @petrsvihlik I run pipelines for NetworkTraversal  and Rooms. They're green

petrsvihlik (approved):

PR #33409, opened 2023-01-11
sync whole spec RP folder

chunyu3:


dw511214992 (approved):

eng/scripts/Cadl-Project-Sync.ps1
```diff
@@ -19,7 +19,11 @@ function AddSparseCheckoutPath([string]$subDirectory) {
 function CopySpecToProjectIfNeeded([string]$specCloneRoot, [string]$mainSpecDir, [string]$dest, [string[]]$specAdditionalSubDirectories) {
     $source = "$specCloneRoot/$mainSpecDir"
     Write-Host "Copying spec from $source"
-    Copy-Item -Path $source -Destination $dest -Recurse -Force
+    # $mainSpecDir is the PR folder, we just need to copy its subfolders which include the cadl project folder
+    Get-ChildItem –Path "$source" -Exclude @("data-plane", "resource-manager")|
```
weshaggard, line 23: Why do we need to exclude these? Does it hurt anything to have them? I think it might actually help potentially with debugging to be able to see the swagger files as well.

chunyu3: Hello @weshaggard  In spec repo, the RP folder structure is:
{RP}
  -CADL project folder
  -data-plane // data-plane swagger folder
 -resource-manager //resource-manager swagger folder

This script is for generating SDK from CADL directly, so we just need to sync down the cadl project folder.  And those swagger files are not used in debugger since it may not consistent with the cadl.

weshaggard: I understand we aren't using the swagger files directly I was only implying it might make some manual diagnostics easier. Not a big deal but it just seems simpler if we just copy the entire contents of the service folder from the specs repo, or better yet not even copy by just clone it directly where we need it to avoid any copying.

chunyu3: Thanks @weshaggard It is good suggestion to clone directly to the `TempCadlFiles` in sdk repo(locally cloned) to avoid copying. We will investigate the whole sync/spark-checkout, sdk generation process to verify if it is workable. If so, we will plan to simplify this. Thanks,

PR #33425, opened 2023-01-11
Fixing download 304 tests again.

amnguye:
Another attempt (hopefully the last one) to fix the download 304 tests.

This time around we're just going to use the LastModifiedTime from the Upload Request and set that as the Last-Modified time for the download request.

That way we absolutely can't be sending a Last-Modified time that is before the blob was uploaded.

Turns out the Recording.UtcNow grabs the time from when the test started, not the current time at the time of calling that API.


amnguye:
Looks like the 304 tests aren't failing, I'm going to run the live tests one more time after I merge main as for a sanity check.

jaschrep-msft (approved):

amnguye:
Unrelated: There's a flakey tests I saw that ran on this pipeline.
FindBlobsByTagAsync_AccountSas 

https://github.com/Azure/azure-sdk-for-net/issues/33552

Made a github issue to address this test later

PR #33429, opened 2023-01-11
Update analyzers package and remove suppressions file from ACR

### LINKED ISSUE #19285: Remove GlobalSuppressions.cs file from sdk\containerregistry\Azure.Containers.ContainerRegistry
These will be removed one-by-one as convenience layer and static analyzers are added/

annelo-msft:
Added Azure.Containers to list of accepted namespaces recognized by analyzers in https://github.com/Azure/azure-sdk-tools/pull/4932.  This PR updates the repo to use the analyzer package that includes this change.

Fixes https://github.com/Azure/azure-sdk-for-net/issues/19285


pallavit (approved):

eng/Packages.Data.props
```diff
@@ -161,7 +161,7 @@
   -->
   <ItemGroup>
     <PackageReference Update="Microsoft.Azure.AutoRest.CSharp" Version="3.0.0-beta.20230104.3" PrivateAssets="All" />
-    <PackageReference Update="Azure.ClientSdk.Analyzers" Version="0.1.1-dev.20221031.1" PrivateAssets="All" />
+    <PackageReference Update="Azure.ClientSdk.Analyzers" Version="0.1.1-dev.20221229.1" PrivateAssets="All" />
```
pallavit, line 164: Since this updates the central location everyone will from now on get the new package, is there any process we follow to ensure no-one else breaks with it? It is likely not to be the case and there should not be breaking change but curious.

annelo-msft: I would have expected this PR to kick off the same pipelines as Core for that reason, but I don't see that happening.  @weshaggard, do you know if there is a way to check?

weshaggard: This pipeline should kick of the core pipeline which is a proxy for all the other pipelines as it does build those other projects that depend on core. Right now looks like there is a merge conflict which is why I expect the pipelines aren't triggering.

annelo-msft: Got it - thanks, @weshaggard!  I'll resolve that, and we'll go from there.  Thanks!

Issue #23828, opened 2021-09-07
[ACR] [BlobClient] Add support for non-OCI Manifests


annelo-msft:
We don't currently support the Docker v1 and v2 manifest types in create/get manifest:

- v1: `application/vnd.docker.container.image.v1+json`
- v2: `application/vnd.docker.distribution.manifest.v2+json`

Support for these types will involve solving a polymorphic type deserialization problem, which will involve some swagger transforms.  Given that the [Azure REST API Guidelines](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#polymorphic-types) discourage this, it would be worth looking for another way to solve these problems, perhaps by modifying the API to allow the caller to specify which manifest type to download, e.g.:
- `DownloadOciManifestAsync`
- `DownloadDockerV1ManifestAsync`
- `DownloadDockerV2ManifestAsync`

annelo-msft:
Address this PR comment as part of this: https://github.com/Azure/azure-sdk-for-net/pull/33432#discussion_r1070196999

PR #33432, opened 2023-01-12
Initial implementation of chunked download for ACR library.

### LINKED ISSUE #32414: Enable support for large blob download in the BlobClient
Reimplement DownloadBlob to download chunks in parallel

- Suggested API implementation for DownloadBlobOptions here: [DownloadBlobOptions](https://apiview.dev/Assemblies/Review/abdd1f12d0384f1daed0563c01fd3b89?diffRevisionId=c062a4ee001f412d81c8eef6a39c16f1&diffOnly=False&revisionId=70d6f339459247a7919e11d09a5d33e0&doc=False#Azure.Containers.ContainerRegistry.ArtifactDownloadToOptions)
  - Specifically, provide options for caller to specify number of workers and chunk size.

Notes:
- [x] Move to ArrayPool implementation for buffers, per: https://github.com/Azure/azure-sdk-for-net/pull/32059#issuecomment-1312052875
- [x] Reimplement ValidateDigest to throw the exception in one place
- [x] Remove Debug.Assert from ComputeDigest, per: https://github.com/Azure/azure-sdk-for-net/pull/32059#discussion_r1020499353
- [ ] Per architect feedback, it would be good to look at consolidating parallel download functionality possibly to the Storage Common library, and using it from a single place.  Investigate this.  If we decide to do this, also look into whether we can re-implement chunked using a a common implementation.

annelo-msft:
Customers have reported OutOfMemoryExceptions when downloading large blobs using the `DownloadBlob()` API, which tries to download a blob in a single request.  This PR:
1. Updates the `DownloadBlob()` method to use the same semantics as the [Storage BlobClient DownloadContent()](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/storage/Azure.Storage.Blobs/src/BlobBaseClient.cs#L1799) method and return a result with a `BinaryData` Content property.  The XML comment indicates that this is intended for blobs of sizes that can fit in memory.
2. Implements a new `DownloadBlobTo()` method similar to the [Storage BlobClient DownloadTo()](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/storage/Azure.Storage.Blobs/src/BlobBaseClient.cs#L2324) method  It takes a destination Stream, and downloads the blob sequentially in chunks into the destination stream.

Fixes https://github.com/Azure/azure-sdk-for-net/issues/32414


sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/BlobHelper.cs
```diff
@@ -0,0 +1,51 @@
...
+            return $"sha256:{BytesToString(hash)}";
+        }
+
+        private static string BytesToString(byte[] bytes)
+        {
+            var builder = new StringBuilder();
+            for (int i = 0; i < bytes.Length; i++)
+            {
+#pragma warning disable CA1305 // Specify IFormatProvider
+                builder.AppendFormat("{0:X2}", bytes[i]);
```
lmolkova, line 43:
```suggestion
                builder.AppendFormat("{0:x2}", bytes[i]);
```

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/BlobHelper.cs
```diff
@@ -0,0 +1,51 @@
...
+#pragma warning restore CA1305 // Specify IFormatProvider
+            }
+#pragma warning disable CA1304 // Specify CultureInfo
+            return builder.ToString().ToLower();
```
lmolkova, line 47:
```suggestion
            return builder.ToString();
```

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/BlobHelper.cs
```diff
@@ -0,0 +1,51 @@
...
+            }
+            finally
+            {
+                stream.Position = position;
+            }
+        }
+
+        internal static string FormatDigest(byte[] hash)
+        {
+            return $"sha256:{BytesToString(hash)}";
```
lmolkova, line 34: would be a bit more efficient to prepend `sha256:` inside `BytesToString` in string builder

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs
```diff
@@ -405,44 +395,51 @@ private async Task<ChunkedUploadResult> UploadInChunksInternal(string location,
...
-                    uploadChunkResult = async ?
-                        await _blobRestClient.UploadChunkAsync(location, chunk, contentRange, bytesRead.ToString(CultureInfo.InvariantCulture), cancellationToken).ConfigureAwait(false) :
-                        _blobRestClient.UploadChunk(location, chunk, contentRange, bytesRead.ToString(CultureInfo.InvariantCulture), cancellationToken);
+                    var contentRange = GetContentRange(chunkCount * chunkSize, bytesRead);
+                    location = uploadChunkResult?.Headers.Location ?? location;
+
+                    // Incrementally compute hash for digest.
+                    sha256.TransformBlock(buffer, 0, bytesRead, buffer, 0);
+
+                    using (Stream chunk = new MemoryStream(buffer, 0, bytesRead))
```
lmolkova, line 418: nit: is there a way to pass buffer (without reading it into intermediary `chunk` stream)? Please ignore my comment if it's all generated code and it can't be easily done.

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs
```diff
@@ -535,17 +528,15 @@ public virtual async Task<Response<DownloadManifestResult>> DownloadManifestAsyn
...
+                ValidateDigest(contentDigest, digest);
 
-                using var document = JsonDocument.Parse(rawResponse.ContentStream);
+                using var document = JsonDocument.Parse(rawResponse.Content);
                 var manifest = OciManifest.DeserializeOciManifest(document.RootElement);
 
                 rawResponse.ContentStream.Position = 0;
 
-                return Response.FromValue(new DownloadManifestResult(digest, manifest, rawResponse.ContentStream), rawResponse);
+                return Response.FromValue(new DownloadManifestResult(digest, manifest, rawResponse.Content), rawResponse);
```
lmolkova, line 539: nit and future: we should probably deserialize content into manifest lazily inside `DownloadManifestResult` when user requests it.

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs
```diff
@@ -621,15 +598,80 @@ public virtual async Task<Response<DownloadBlobResult>> DownloadBlobAsync(string
...
+
+        /// <summary>
+        /// Download a blob to a passed-in destination stream.
+        /// </summary>
+        /// <param name="digest">The digest of the blob to download.</param>
+        /// <param name="destination">Destination for the downloaded blob.</param>
+        /// <param name="options">Options for the blob download.</param>
+        /// <param name="cancellationToken"> The cancellation token to use. </param>
+        /// <returns>The raw response corresponding to the final GET blob chunk request.</returns>
+        public virtual Response DownloadBlobTo(string digest, Stream destination, DownloadBlobToOptions options = default, CancellationToken cancellationToken = default)
```
lmolkova, line 634: something to consider: blob size and digest come from OciBlobDescriptor, it'd nice to do something like:
```
var manifest = client.DownloadManifest("foo").GetManifest();
foreach (var layer : manifest.Layers) {
 client.DownloadBlobTo(layer, stream, cancellationToken) ;
}
```

then `DownloadBlobTo` options are only needed for chunk size. But I'm not sure if chunk size makes sense on per-request basis.


So alternative proposal to consider is  
- have 2 overloads for `DownloadBlobInternalAsync`: 
  - `(string digest, Stream dest, CancellationToken cancellationToken)`
  - `(OciBlobDescriptor blobDescriptor, Stream dest,  CancellationToken cancellationToken)`
-  add chunk size as client-level property (or even hardcode one for now)
- If in future we need download options, we can always add them

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs
```diff
@@ -638,6 +680,100 @@ public virtual async Task<Response<DownloadBlobResult>> DownloadBlobAsync(string
...
+            byte[] buffer = ArrayPool<byte>.Shared.Rent(maxChunkSize);
+            long chunkCount = 0;
+            long bytesDownloaded = 0;
+            using SHA256 sha256 = SHA256.Create();
+
+            try
+            {
+                Response result = null;
+
+                while (bytesDownloaded < blobLength)
```
lmolkova, line 716: my understanding of [HTTP spec](https://www.rfc-editor.org/rfc/rfc7233#section-2.1) is that we don't need to know length ahead of time.

Service should respond to ranges if end is bigger than content-length

```
If a valid byte-range-set includes at least one byte-range-spec with
   a first-byte-pos that is less than the current length of the
   representation, or at least one suffix-byte-range-spec with a
   non-zero suffix-length, then the byte-range-set is satisfiable.
   Otherwise, the byte-range-set is unsatisfiable.

```
Service [should respond](https://www.rfc-editor.org/rfc/rfc7233#section-4.1) with `Content-Range` header that includes full blob size

e.g. `Content-Range: bytes 21010-47021/47022` (requested-range/size). 

I.m not sure what we do in other SDKs, but would be nice to avoid extra request (or getting size from user)

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs
lmolkova: if we end up not needing blobSize ahead of time (see my other comment) I still would like to consider chunk size on the client level (or hardcoded)

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/BlobHelper.cs
annelo-msft: 👍

sdk/containerregistry/Azure.Containers.ContainerRegistry/api/Azure.Containers.ContainerRegistry.netstandard2.0.cs
```diff
@@ -253,26 +255,30 @@ public partial class ContainerRegistryBlobClient
...
     {
         internal DownloadBlobResult() { }
-        public System.IO.Stream Content { get { throw null; } }
+        public System.BinaryData Content { get { throw null; } }
         public string Digest { get { throw null; } }
-        public void Dispose() { }
+    }
+    public partial class DownloadBlobToOptions
+    {
+        public DownloadBlobToOptions(int maxChunkSize) { }
```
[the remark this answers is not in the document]
annelo-msft: We use the [default chunk size](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs#L22) if options isn't passed ... options itself is optional.  If we don't ask the user for the blob size, this would be the only option someone could pass, so I think it makes sense to have a constructor that requires it ... the .NET pattern for options types wants us to set required values in the constructor.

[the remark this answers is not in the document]
annelo-msft: Hm, I like the idea to move chunk size to a client-level option.  Let me look into this - thanks!

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs
annelo-msft: I like the idea!  It is calling into generated code, so it would not be easy to accomplish, unfortunately.

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs
annelo-msft: I'll add this to the Manifest API work - good suggestion - thank you!

annelo-msft: Tracked here: https://github.com/Azure/azure-sdk-for-net/issues/23828#issuecomment-1396205254

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs
annelo-msft: One advantage to the expected blob size ahead of time is that is avoids a potentially large buffer allocation if the blob size is small.  But I agree it would be nice to avoid the extra request.

I guess if someone knows their blob will fit in memory, they call DownloadBlob and not DownloadBlobTo?

sdk/containerregistry/Azure.Containers.ContainerRegistry/api/Azure.Containers.ContainerRegistry.netstandard2.0.cs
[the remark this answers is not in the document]
annelo-msft: Added to ClientOptions.

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs
annelo-msft: Addressed.

lmolkova (approved):
LGMT!

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs
lmolkova: > one advantage to the expected blob size ahead of time is that is avoids a potentially large buffer allocation if the blob size is small

i think we don't really need to allocate the buffer until we know what the blob size is, but this is micro-optimization


> I guess if someone knows their blob will fit in memory, they call DownloadBlob and not DownloadBlobTo?

 great point! I haven't even thought that we can explain the difference on the API level.

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/BlobHelper.cs
```diff
@@ -0,0 +1,50 @@
...
+                stream.Position = position;
+            }
+        }
+
+        internal static string FormatDigest(byte[] hash)
+        {
+            return BytesToString(hash);
+        }
+
+        private static string BytesToString(byte[] bytes)
```
KrzysztofCwalina, line 37: this routine is pretty slow in comparison to what is possible if we write our own formatting routine. Do we care about perf here?

annelo-msft: I'm not sure this is our primary bottleneck, but tracking this with https://github.com/Azure/azure-sdk-for-net/issues/33614 and we can prioritize per customer needs.  Thanks @KrzysztofCwalina!

PR #33445, opened 2023-01-12
[DigitalTwins] Adds Support for API version 2023-02-27-preview

sjiherzig:


sdk/digitaltwins/Azure.DigitalTwins.Core/api/Azure.DigitalTwins.Core.netstandard2.0.cs
```diff
@@ -111,13 +111,12 @@ public partial class DigitalTwinsClient
...
     public partial class DigitalTwinsClientOptions : Azure.Core.ClientOptions
     {
-        public DigitalTwinsClientOptions(Azure.DigitalTwins.Core.DigitalTwinsClientOptions.ServiceVersion version = Azure.DigitalTwins.Core.DigitalTwinsClientOptions.ServiceVersion.V2022_05_31) { }
+        public DigitalTwinsClientOptions(Azure.DigitalTwins.Core.DigitalTwinsClientOptions.ServiceVersion version = Azure.DigitalTwins.Core.DigitalTwinsClientOptions.ServiceVersion.V2023_02_27_Preview) { }
         public Azure.Core.Serialization.ObjectSerializer Serializer { get { throw null; } set { } }
         public Azure.DigitalTwins.Core.DigitalTwinsClientOptions.ServiceVersion Version { get { throw null; } }
         public enum ServiceVersion
         {
-            V2020_10_31 = 1,
-            V2022_05_31 = 2,
```
rachelsophia, line 120: Why delete this as opposed to just adding the new version (which is what it seems like we did for V2022_05_31 at least)

sdk/digitaltwins/Azure.DigitalTwins.Core/src/DigitalTwinsClientOptions.cs
```diff
@@ -63,8 +58,7 @@ internal string GetVersionString()
         {
             return Version switch
             {
-                ServiceVersion.V2020_10_31 => "2020-10-31",
-                ServiceVersion.V2022_05_31 => "2022-05-31",
```
rachelsophia, line 67: Why delete these as opposed to just adding the new version (which is what it seems like we did for V2022_05_31 at least)

sdk/digitaltwins/Azure.DigitalTwins.Core/src/Customized/DigitalTwinsRestClient.cs
```diff
@@ -19,6 +19,234 @@ internal partial class DigitalTwinsRestClient
     {
         private const string DateTimeOffsetFormat = "MM/dd/yy H:mm:ss zzz";
 
+        /// <summary>
```
rachelsophia, line 22: A little confused why these need to be added, seems like its related to other parts of the SDK not bulk right?

rachelsophia: In other words - still a little confused about what changed in bulk to cause this - or is this something we should / could do as a separate PR? would it need more reviewing than bulk?

sjiherzig: I _think_ this was caused by the additionalProperties that were added, which are generic dictionaries. I.e., the response type really changed from a generic object to a dictionary. In our SDKs we previously already accounted for this case to some extent, but additional changes are necessary with the updated response model. I think in the solutions client we have something similar.

sdk/digitaltwins/Azure.DigitalTwins.Core/src/DigitalTwinsClientOptions.cs
sjiherzig: There was parity in terms of features between the two versions. If we only add it, then we would need to add additional logic to hide / avoid calling e.g., bulk with the previous API versions. This is the generally recommended pattern.

PR #33465, opened 2023-01-13
[Text Analytics] Edit README section about Azure CLI

### LINKED ISSUE #33464: [Text Analytics] Edit README about creating a resource using the Azure CLI
Creating a Language service resource requires the user to agree to the responsible AI terms, which is currently only possible through the Azure portal. To avoid confusion, we should edit the README to point to the following document, which describes the necessary actions: 
🔗 https://learn.microsoft.com/en-us/azure/cognitive-services/cognitive-services-apis-create-account-cli?tabs=windows#prerequisites

joseharriaga:
Resolves: https://github.com/Azure/azure-sdk-for-net/issues/33464

### DIFF
--- a/sdk/textanalytics/Azure.AI.TextAnalytics/README.md
+++ b/sdk/textanalytics/Azure.AI.TextAnalytics/README.md
@@ -1,23 +1,26 @@
 # Azure Cognitive Services Text Analytics client library for .NET
+
 Text Analytics is part of the Azure Cognitive Service for Language, a cloud-based service that provides Natural Language Processing (NLP) features for understanding and analyzing text. This client library offers the following features:
 
-  * Language detection
-  * Sentiment analysis
-  * Key phrase extraction
-  * Named entity recognition (NER)
-  * Personally identifiable information (PII) entity recognition
-  * Entity linking
-  * Text analytics for health
-  * Custom named entity recognition (Custom NER)
-  * Custom text classification
-  * Dynamic text classification
-  * Extractive text summarization
-  * Abstractive text summarization
+* Language detection
+* Sentiment analysis
+* Key phrase extraction
+* Named entity recognition (NER)
+* Personally identifiable information (PII) entity recognition
+* Entity linking
+* Text analytics for health
+* Custom named entity recognition (Custom NER)
+* Custom text classification
+* Dynamic text classification
+* Extractive text summarization
+* Abstractive text summarization
 
 [Source code][textanalytics_client_src] | [Package (NuGet)][textanalytics_nuget_package] | [API reference documentation][textanalytics_refdocs] | [Product documentation][language_service_docs] | [Samples][textanalytics_samples]
 
 ## Getting started
+
 ### Install the package
+
 Install the Azure Text Analytics client library for .NET with [NuGet][nuget]:
 
 ```dotnetcli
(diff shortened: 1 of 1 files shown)


kristapratico (approved):

sdk/textanalytics/Azure.AI.TextAnalytics/README.md
```diff
@@ -845,6 +784,8 @@ This project has adopted the [Microsoft Open Source Code of Conduct][code_of_con
 
 [azure_cli]: https://docs.microsoft.com/cli/azure
 [azure_sub]: https://azure.microsoft.com/free/dotnet/
+[azure_portal_create_ta_resource]: https://portal.azure.com/#create/Microsoft.CognitiveServicesTextAnalytics
```
mssfang, line 787: should we use this link reference: https://learn.microsoft.com/en-us/azure/cognitive-services/cognitive-services-apis-create-account?tabs=multiservice%2Canomaly-detector%2Clanguage-service%2Ccomputer-vision%2Cwindows

joseharriaga: Yeah, it's probably better to link the document than link to the Azure portal directly. Fixed.

sdk/textanalytics/Azure.AI.TextAnalytics/README.md
```diff
@@ -17,9 +17,7 @@ Text Analytics is part of the Azure Cognitive Service for Language, a cloud-base
 [Source code][textanalytics_client_src] | [Package (NuGet)][textanalytics_nuget_package] | [API reference documentation][textanalytics_refdocs] | [Product documentation][language_service_docs] | [Samples][textanalytics_samples]
 
 ## Getting started
-
```
heaths, line 20: I highly recommend installing the Microsoft Docs Pack (might be the Microsoft Learn Pack) now in VSCode. These are the same rules that learn.microsoft.com use and helps identify issues in the markdowns.

While this change isn't a problem rendering, it is a linting rule to keep the plaintext version more readable and ends up getting flagged, thus making actual issues harder to spot.

I try to keep all my markdowns clean with those rulesets for that reason: fix all the warnings - even seemingly lame ones like this - and the good warnings are easier to spot e.g., like not including locales e.g., "en-us", in links.

@weshaggard should we just add that VSCode extension pack into _.github/extensions.json_? I know I'm not the only one using them.

heaths (approved):
Approved, but see my content. I *highly* recommend putting back all these blank lines, which the Microsoft Docs Pack (markdownlint, specifically) will flag - likely because it makes the plaintext file harder to read.

mssfang (approved):

sdk/textanalytics/Azure.AI.TextAnalytics/README.md
weshaggard: I'm not opposed to you guys adding it. We have issue https://github.com/Azure/azure-sdk-tools/issues/1896 about MD linting in general but it hasn't been prioritized yet.

heaths (approved):

sdk/textanalytics/Azure.AI.TextAnalytics/README.md
joseharriaga: It appears that it is now called the "Learn Authoring Pack". Ironically, I removed the blank lines because I personally thought that making it less "airy" made it easier to read in plain text. 😂 I undid that change and also fixed a couple of other pre-existing linting issues. Thanks!

PR #33483, opened 2023-01-13
Dev/digitaltwins/rateitel/update to new api version

rachelsophia:

Generated Code, Fixed bugs, and added/updated tests. 
image


sdk/digitaltwins/Azure.DigitalTwins.Core/tests/ImportTests.cs
```diff
@@ -0,0 +1,152 @@
...
+    /// </summary>
+    public class ImportTests : E2eTestBase
+    {
+        public ImportTests(bool isAsync)
+            : base(isAsync)
+        {
+        }
+
+        [Test]
+        public async Task Import_Lifecycle()
```
rachelsophia, line 26: @abhinav-ghai any other tests we want to add here? Kind of went off what was in the ModelTests.cs file, but we can add some more as well.

sdk/digitaltwins/Azure.DigitalTwins.Core/src/Customized/Models/BulkImportJob.cs
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Azure.Core;
+
+namespace Azure.DigitalTwins.Core
+{
+    [CodeGenModel("BulkImportJob")]
+    public partial class BulkImportJob
```
rachelsophia, line 9: Do we want to rename this and all other files (e.g BulkImportJobCollection, Serialization, etc )to not include "Bulk" ? Could change this to just be "ImportJob"

cc @abhinav-ghai

sdk/digitaltwins/test-resources.bicep
```diff
@@ -87,3 +92,76 @@ resource digitaltwinEndpoints 'Microsoft.DigitalTwins/digitalTwinsInstances/endp
 }
 
 output DIGITALTWINS_URL string = 'https://${digitaltwin.properties.hostName}'
+
+
+var containerName = baseName
+
+resource storageAccount 'Microsoft.Storage/storageAccounts@2022-05-01' = {
```
rachelsophia, line 99: this section sets up the new test resources required by Import API. The input blob name and storage container endpoint are used in TestEnvironment.cs.

sdk/digitaltwins/Azure.DigitalTwins.Core/tests/ImportTests.cs
```diff
@@ -0,0 +1,152 @@
...
+                // DELETE job
+                var deleteResponse = await client.DeleteImportJobsAsync(jobId).ConfigureAwait(false);
+                Assert.IsNotNull(deleteResponse);
+                Assert.AreEqual((int)HttpStatusCode.NoContent, deleteResponse.Status);
+
+                // GET job after deletion - should fail
+                Func<Task> act = async () => await client.GetImportJobsByIdAsync(jobId).ConfigureAwait(false);
+                act.Should().Throw<RequestFailedException>()
+                    .And.Status.Should().Be((int)HttpStatusCode.NotFound);
+            }
```
rachelsophia, line 67: TODO - Add a test for the cancel operation.

sdk/core/Azure.Core.TestFramework/src/TestEnvironment.cs
```diff
@@ -187,6 +187,16 @@ static TestEnvironment()
         /// </summary>
         public string ClientId => GetRecordedVariable("CLIENT_ID");
 
+        /// <summary>
+        ///   The storage container endpoint to use during Import Live tests. Recorded.
+        /// </summary>
+        public string StorageContainerEndpoint => GetRecordedVariable("STORAGE_CONTAINER_URI");
```
JoshLove-msft, line 193: These seem like they are specific to your service, so they should go in the service level TestEnvironment class.

sdk/digitaltwins/Azure.DigitalTwins.Core/src/Customized/Models/BulkImportJob.cs
abhinav-ghai: Yes, let's rename BulkImportJob to ImportJob in all places.

sdk/core/Azure.Core.TestFramework/src/TestEnvironment.cs
rachelsophia: Good point - moved to sdk/digitaltwins/Azure.DigitalTwins.Core/tests/DigitalTwinsTestEnvironment.cs. Thanks!

abhinav-ghai (approved):

Aashish93-stack (approved):

sdk/digitaltwins/Azure.DigitalTwins.Core/src/Customized/Models/Error.cs
```diff
@@ -6,7 +6,7 @@
 namespace Azure.DigitalTwins.Core
 {
     [CodeGenModel("Error")]
-    internal partial class Error
+    public partial class ErrorInformation
```
johngallardo, line 9: Maybe I am just missing some context - why did this type name change? 

Should the file name change also to ErrorInformation.cs?

sdk/digitaltwins/Azure.DigitalTwins.Core/src/Customized/Models/InnerError.cs
```diff
@@ -6,7 +6,7 @@
 namespace Azure.DigitalTwins.Core
 {
     [CodeGenModel("InnerError")]
-    internal partial class InnerError
+    public partial class InnerError
```
johngallardo, line 9: Why did this change to public?

sdk/digitaltwins/Azure.DigitalTwins.Core/src/Customized/Models/ImportJob.Serialization.cs
```diff
@@ -0,0 +1,106 @@
...
+
+using System;
+using System.Text.Json;
+using Azure.Core;
+
+namespace Azure.DigitalTwins.Core
+{
+    public partial class ImportJob : IUtf8JsonSerializable
+    {
+        // This class definition overrides deserialization implementation in order to allow finishedDateTime to be null in the response.
```
johngallardo, line 12: Can you please clarify?

Why doesn't justting the member in the C# object to DateTimeOffset? not allow the response to contain a null value?

sdk/digitaltwins/Azure.DigitalTwins.Core/src/Customized/Models/Error.cs
johngallardo: In particular, does this have any impact on the existing API error message types?

sdk/digitaltwins/Azure.DigitalTwins.Core/src/DigitalTwinsClientOptions.cs
```diff
@@ -12,7 +12,7 @@ namespace Azure.DigitalTwins.Core
     /// </summary>
     public class DigitalTwinsClientOptions : ClientOptions
     {
-        internal const ServiceVersion LatestVersion = ServiceVersion.V2022_05_31;
+        internal const ServiceVersion LatestVersion = ServiceVersion.V2023_02_27_Preview;
```
johngallardo, line 15: Is it expected for the APIs in main/master to be referencing a Preview version of the service?

sdk/digitaltwins/Azure.DigitalTwins.Core/tests/ImportTests.cs
```diff
@@ -0,0 +1,163 @@
...
+                Assert.AreEqual((int)HttpStatusCode.NoContent, deleteResponse.Status);
+
+                // validate GET job after deletion - should fail
+                Func<Task> act = async () => await client.GetImportJobsByIdAsync(jobId).ConfigureAwait(false);
+                act.Should().Throw<RequestFailedException>()
+                    .And.Status.Should().Be((int)HttpStatusCode.NotFound);
+            }
+            catch (Exception ex)
+            {
+                Assert.Fail($"Failure in executing a step in the test case: {ex.Message}.");
```
johngallardo, line 76: `ex.Message` can lose important context like call stack. Can we just serialize the entire exception or otherwise ensure it is available to allow failures to be easily triaged and/or root caused.

sdk/digitaltwins/Azure.DigitalTwins.Core/tests/ImportTests.cs
rachelsophia: discussed offline and added more

sdk/digitaltwins/Azure.DigitalTwins.Core/src/Customized/Models/ImportJob.Serialization.cs
rachelsophia: @sjiherzig  and @abhinav-ghai can help elaborate here, since we talked extensively on this change - but essentially this was a bug we found where we are returning null in the api response but in the swagger this is defined as a non-null property. So the auto-generated code has a check (which I am replacing) that this property should not be null when it will be.

As it turns out, we should never give a null property to customers in an api response - this is what we were told by someone in the api review board. But that will require a change in the bulk code itself. So instead of marking this value as nullable in the swagger, which will require a swagger review and is ultimately not the behavior we even want, we are making this change here to buy us time. Then we will fix the real issue which is that this should never be returned as null, and then this will be removed from the SDK.

sdk/digitaltwins/Azure.DigitalTwins.Core/src/DigitalTwinsClientOptions.cs
rachelsophia: I guess same response as Santhosh here - @sjiherzig any insights on a separate branch for preview apis?

sdk/digitaltwins/Azure.DigitalTwins.Core/src/Customized/Models/Error.cs
rachelsophia: This discussion was taken offline to teams

sdk/digitaltwins/Azure.DigitalTwins.Core/src/Customized/Models/InnerError.cs
rachelsophia: Because InnerError in ErrorInformation class defines this as public. I guess another route would be to override the accessibility in ErrorInformation.cs so that InnerError is internal there as well - would that be preferable over this? 

Either way the accessibility somewhere has to change since ErrorInformation is now a public class (per our discussion offline of bulk api using this error model in the response)

sdk/digitaltwins/Azure.DigitalTwins.Core/tests/ImportTests.cs
rachelsophia: Updated

sdk/digitaltwins/Azure.DigitalTwins.Core/src/DigitalTwinsClientOptions.cs
rachelsophia: This PR will be updated with the GA version when that is available in the swagger

sjiherzig: You cannot take an SDK to GA if it references a preview version. We ran into this in the past. We'll need to add a new stable API version in this case, and reference that in the SDKs.

sdk/digitaltwins/Azure.DigitalTwins.Core/samples/DigitalTwinsClientSample/JobLifecycleSamples.cs
```diff
@@ -0,0 +1,100 @@
...
+        public async Task RunSamplesAsync(DigitalTwinsClient client, Options options)
+        {
+            PrintHeader("JOB LIFECYCLE SAMPLE");
+
+            // For the purpose of this example we will create an import job to import models, twins, and relationships.
+            // We have to make sure the job Id is unique within the DT instance.
+
+            string sampleImportJobId = await GetUniqueJobIdAsync(SamplesConstants.TemporaryJobPrefix, client);
+
+            // upload a sample file to their storage blob? or should we require that they already have one uploaded and provide the name of the file as input?
```
rachelsophia, line 29: I'm still not sure how we should do this. Options are:

**1. We upload input blob to customer storage account**
Pros: Less work for the customer, seems in line with the rest of the samples for not making the customer come up with any data (e.g model names, twin names, etc).
Cons: More work, more permissions needed to get the sample running.

**2. Customer uploads input blob to storage account**
Pros: Less work for us, keeps the sample scoped to modifying adt instance only (not customer storage blob)
Cons: More work for customer to get sample up and running, could be malformed.

sdk/digitaltwins/Azure.DigitalTwins.Core/samples/DigitalTwinsClientSample/JobLifecycleSamples.cs
```diff
@@ -0,0 +1,100 @@
...
+
+            ImportJob sampleImportJob = new ImportJob(options.StorageAccountEndpoint + "/sampleInputBlob.ndjson", options.StorageAccountEndpoint + "/sampleOutputBlob.ndjson");
```
rachelsophia, line 31: in other words - this part might change based on what we decide above

sdk/digitaltwins/Azure.DigitalTwins.Core/samples/DigitalTwinsClientSample/JobLifecycleSamples.cs
rachelsophia: Discussed offline with @sjiherzig and we will go with option 1, plus we will add some setup code to create the storage account for the customer if possible.

rachelsophia:
TODO: 
1. link quickstart readme to docs regarding how to set up storage container for import job api
2. update latest version with new GA version when available and re-record tests

sdk/digitaltwins/Azure.DigitalTwins.Core/samples/DigitalTwinsClientSample/JobLifecycleSamples.cs
rachelsophia: Opting not to create the storage account since it will require extra setup of permissions etc that may be complicated. Instead, will upload the input blob to the container as part of the sample which will be used in the import api sample.

PR #33484, opened 2023-01-13
Replace HttpMessage Type keyed property bag with array implementation

christothes:
Benchmarks comparing the old impl to the new one:
Note: the 5 items benchmark exercises a scenario where an item is updated frequently and then a value stored later is accessed frequently.

|                     Method |              Runtime | items |        Mean |     Error |    StdDev | Ratio | RatioSD |  Gen 0 |  Gen 1 | Allocated |
|--------------------------- |--------------------- |------ |------------:|----------:|----------:|------:|--------:|-------:|-------:|----------:|
|             DictionaryImpl |             .NET 6.0 |     2 |   137.54 ns |  1.940 ns |  1.515 ns |  1.00 |    0.00 | 0.0014 |      - |     440 B |
| ArrayBackedPropertyBagPool |             .NET 6.0 |     2 |    64.68 ns |  0.595 ns |  0.497 ns |  0.47 |    0.01 | 0.0007 |      - |     224 B |
|             DictionaryImpl | .NET Framework 4.6.2 |     2 |   156.12 ns |  2.093 ns |  1.958 ns |  1.14 |    0.02 | 0.0701 |      - |     441 B |
| ArrayBackedPropertyBagPool | .NET Framework 4.6.2 |     2 |    83.83 ns |  1.160 ns |  1.085 ns |  0.61 |    0.01 | 0.0356 |      - |     225 B |
|                            |                      |       |             |           |           |       |         |        |        |           |
|             DictionaryImpl |             .NET 6.0 |     5 | 1,144.65 ns | 22.355 ns | 21.956 ns |  1.00 |    0.00 | 0.0038 |      - |   1,480 B |
| ArrayBackedPropertyBagPool |             .NET 6.0 |     5 |   526.33 ns |  4.860 ns |  4.546 ns |  0.46 |    0.01 | 0.0029 |      - |   1,016 B |
|             DictionaryImpl | .NET Framework 4.6.2 |     5 | 1,249.27 ns | 24.333 ns | 29.883 ns |  1.09 |    0.02 | 0.2346 |      - |   1,484 B |
| ArrayBackedPropertyBagPool | .NET Framework 4.6.2 |     5 |   661.13 ns |  9.566 ns |  7.988 ns |  0.58 |    0.01 | 0.1612 |      - |   1,019 B |
|                            |                      |       |             |           |           |       |         |        |        |           |
|             DictionaryImpl |             .NET 6.0 |    30 | 1,511.50 ns | 17.713 ns | 15.702 ns |  1.00 |    0.00 | 0.0095 |      - |   2,976 B |
| ArrayBackedPropertyBagPool |             .NET 6.0 |    30 | 1,035.63 ns | 16.165 ns | 14.330 ns |  0.69 |    0.01 | 0.0019 |      - |     896 B |
|             DictionaryImpl | .NET Framework 4.6.2 |    30 | 1,545.29 ns | 30.831 ns | 54.802 ns |  1.03 |    0.04 | 0.4730 | 0.0019 |   2,985 B |
| ArrayBackedPropertyBagPool | .NET Framework 4.6.2 |    30 | 1,461.55 ns | 27.251 ns | 26.764 ns |  0.97 |    0.02 | 0.1411 |      - |     899 B |


sdk/core/Azure.Core/src/HttpMessage.cs
```diff
@@ -143,30 +138,25 @@ public bool TryGetProperty(string name, out object? value)
...
-                properties = new Dictionary<string, object>();
-                _typeProperties[typeof(MessagePropertyKey)] = properties;
-            }
-            else
-            {
-                properties = (Dictionary<string, object>)rawValue!;
-            }
-            properties[name] = value;
+            _properties ??= new Dictionary<string, object>();
+            _properties[name] = value;
```
JoshLove-msft, line 142: Why do we still use the dictionary here?

JoshLove-msft: Both APIs are public now so I don't really see the point of optimizing only one of them.

christothes: Good point - we could use it for both.

sdk/core/Azure.Core/src/HttpMessage.cs
```diff
@@ -14,12 +14,8 @@ namespace Azure.Core
     /// </summary>
     public sealed class HttpMessage : IDisposable
     {
-        /// <summary>
-        /// This dictionary is keyed with <c>Type</c> for a couple of reasons. Primarily, it allows values to be stored such that even if the accessor methods
-        /// become public, storing values keyed by internal types make them inaccessible to other assemblies. This protects internal values from being overwritten
-        /// by external code. See the <see cref="TelemetryDetails"/> and <see cref="UserAgentValueKey"/> types for an example of this usage.
-        /// </summary>
-        private Dictionary<Type, object>? _typeProperties;
+        private Dictionary<string, object>? _properties;
```
annelo-msft, line 17: Related to @JoshLove-msft's comment - is the goal to remove this entirely?

sdk/core/Azure.Core/src/Internal/ArrayBackedPropertyBag.cs
```diff
@@ -0,0 +1,140 @@
...
+                        _value1 = value;
+                        break;
+                    }
+                    if (EqualityComparer<TKey>.Default.Equals(key, _key2!))
+                    {
+                        duplicateKeyDetected = true;
+                        _value2 = value;
+                        break;
+                    }
+                    _array ??= new (TKey Key, TValue Value)[8];
```
annelo-msft, line 62: Is array always re-allocated on assignment?

christothes: This statement constitutes a new array only if `_array` is null.

sdk/core/Azure.Core/src/HttpMessage.cs
christothes: Not remove, just improve the performance of the implementation.

JoshLove-msft (approved):

sdk/core/Azure.Core/src/Internal/ArrayBackedPropertyBag.cs
KrzysztofCwalina, line 62: Should we rent arrays from the pool?

sdk/core/Azure.Core/src/Internal/ArrayBackedPropertyBag.cs
```diff
@@ -0,0 +1,140 @@
...
+            bool duplicateKeyDetected = false;
+            switch (_totalCount)
+            {
+                case 0:
+                    _key1 = key;
+                    _value1 = value;
+                    break;
+
+                case 1:
+                    if (EqualityComparer<TKey>.Default.Equals(key, _key1!))
```
KrzysztofCwalina, line 39: It might be worth measuring how much faster it would be to make the key long (not generic) and so use the simple == long comparison

sdk/core/Azure.Core/src/Internal/ArrayBackedPropertyBag.cs
```diff
@@ -0,0 +1,140 @@
...
+                        return true;
+                    }
+
+                    if (_array == null)
+                    {
+                        value = default;
+                        return false;
+                    }
+
+                    // search the array in reverse to favor the last item added in the case of duplicate keys.
```
KrzysztofCwalina, line 125: OH, then why do you even track if keys are duplicated?

sdk/core/Azure.Core/src/Internal/ArrayBackedPropertyBag.cs
```diff
@@ -0,0 +1,140 @@
...
+using System;
+using System.Collections.Generic;
+
+namespace Azure.Core
+{
+    /// <summary>
+    /// A property bag which is optimized for storage of a small number of items.
+    /// If the item count is less than 2, there are no allocations. Any additional items are stored in an array which will grow as needed.
+    /// </summary>
+    internal class ArrayBackedPropertyBag<TKey, TValue> where TKey : IEquatable<TKey>
```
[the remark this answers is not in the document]
christothes: Since it is only used as a field on a class, does making this a struct actually avoid any allocations?

sdk/core/Azure.Core/src/Internal/ArrayBackedPropertyBag.cs
```diff
@@ -0,0 +1,140 @@
...
+        /// <summary>
+        /// Adds a value to the property bag.
+        /// </summary>
+        /// <param name="key">The key of the value to add.</param>
+        /// <param name="value">The value to add.</param>
+        public void Add(TKey key, TValue value)
+        {
+            Argument.AssertNotNull(key, nameof(key));
+
+            bool duplicateKeyDetected = false;
```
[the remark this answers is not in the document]
christothes: We already do this for the most part, but the first two properties are an edge case that requires us to track this.

sdk/core/Azure.Core/src/Internal/ArrayBackedPropertyBag.cs
christothes: Good idea

sdk/core/Azure.Core/src/Internal/ArrayBackedPropertyBag.cs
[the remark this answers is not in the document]
KrzysztofCwalina: yes. only struct will be inlined in message. class will be a pointer inside message pointing to a different object on the heap

sdk/core/Azure.Core/src/Internal/ArrayBackedPropertyBag.cs
christothes: for the case where the dupe is in the first two items (not in the array)

sdk/core/Azure.Core/src/Internal/ArrayBackedPropertyBag.cs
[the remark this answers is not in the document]
KrzysztofCwalina: Why do first two make any difference?

sdk/core/Azure.Core/src/HttpMessage.cs
```diff
@@ -14,12 +14,8 @@ namespace Azure.Core
...
     public sealed class HttpMessage : IDisposable
     {
-        /// <summary>
-        /// This dictionary is keyed with <c>Type</c> for a couple of reasons. Primarily, it allows values to be stored such that even if the accessor methods
-        /// become public, storing values keyed by internal types make them inaccessible to other assemblies. This protects internal values from being overwritten
-        /// by external code. See the <see cref="TelemetryDetails"/> and <see cref="UserAgentValueKey"/> types for an example of this usage.
-        /// </summary>
-        private Dictionary<Type, object>? _typeProperties;
+        private ArrayBackedPropertyBag<string, object>? _properties;
+        private readonly ArrayBackedPropertyBag<long, object> _propertyBag = new();
```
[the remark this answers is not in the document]
christothes: The tradeoff is between simplifying the logic and optimizing for the type based scenario rather than saving the size of the field reference in the allocation of `HttpMessage`. I opted to go for the simplification, but I'm open to going the other way if you feel strongly that it's better.

sdk/core/Azure.Core/src/Internal/ArrayBackedPropertyBag.cs
```diff
@@ -0,0 +1,140 @@
...
+namespace Azure.Core
+{
+    /// <summary>
+    /// A property bag which is optimized for storage of a small number of items.
+    /// If the item count is less than 2, there are no allocations. Any additional items are stored in an array which will grow as needed.
+    /// </summary>
+    internal class ArrayBackedPropertyBag<TKey, TValue> where TKey : IEquatable<TKey>
+    {
+        private int _totalCount;
+        private int _arrayCount;
```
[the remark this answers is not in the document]
christothes: mainly simplification of the code when dealing with incrementing each, especially when dealing with duplicates.

sdk/core/Azure.Core/src/Internal/ArrayBackedPropertyBag.cs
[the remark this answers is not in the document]
christothes: because duplicates don't affect the total count of items.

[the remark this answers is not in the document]
KrzysztofCwalina: Why do we even need the total count? My mental model is that the only thing we need is the number of KVPs in the logical data structure, i.e. number of items stored in array + number of items stored in fields.

sdk/core/Azure.Core/src/HttpMessage.cs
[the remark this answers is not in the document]
AlexanderSher: Maybe I miss something, but can't we use something like this as a key:
```cs
public readonly struct Key : IEquatable<Key>
{
    private readonly object _key;

    public Key(Type type)
    {
        _key = type;
    }

    public Key(string s)
    {
        _key = s;
    }

    public override bool Equals(object obj) => obj is Key other && Equals(other);

    public bool Equals(Key other) =>
        _key is Type type && other._key is Type otherType 
            ? type == otherType 
            : _key is string s && other._key is string otherS && s == otherS;

    public override int GetHashCode() => _key.GetHashCode();
}
```

sdk/core/Azure.Core/src/Internal/ArrayBackedPropertyBag.cs
AlexanderSher: If we call directly `key.Equals(_key1!)`, it should we one `callvirt` difference.

sdk/core/Azure.Core/src/HttpMessage.cs
```diff
@@ -143,30 +138,25 @@ public bool TryGetProperty(string name, out object? value)
...
+        /// are public, storing values keyed by internal types make them inaccessible to other assemblies. This protects internal values from being overwritten
+        /// by external code. See the <see cref="TelemetryDetails"/> and <see cref="UserAgentValueKey"/> types for an example of this usage. Secondly, <c>Type</c>
+        /// comparisons are faster than string comparisons.
+        /// </remarks>
         /// <returns><c>true</c> if property exists, otherwise. <c>false</c>.</returns>
         public bool TryGetProperty(Type type, out object? value)
         {
-            value = null;
-            return _typeProperties?.TryGetValue(type, out value) == true;
+            return _propertyBag.TryGetValue((long)type.TypeHandle.Value, out value);
```
AlexanderSher, line 159: `type.TypeHandle.Value.ToInt64()`?

sdk/core/Azure.Core/src/HttpMessage.cs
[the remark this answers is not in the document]
AlexanderSher: If `ArrayBackedPropertyBag` becomes a struct, then accessing it as one of the values will cause boxing/unboxing.

sdk/core/Azure.Core/src/Internal/ArrayBackedPropertyBag.cs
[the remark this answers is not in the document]
AlexanderSher: > The only problem with such mutable struct would be that we would need to be super careful in our code

We can make it immutable (`Add` would return a new instance) or make `Add` static and pass instance of the `ArrayBackedPropertyBag` as `ref`.

sdk/core/Azure.Core/src/Internal/ArrayBackedPropertyBag.cs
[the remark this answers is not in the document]
AlexanderSher: I may be wrong, but to me it looks like `arrayCount` can be declared as local variable `var arrayCount = _totalCount - 2`

sdk/core/Azure.Core/src/Internal/ArrayBackedPropertyBag.cs
```diff
@@ -0,0 +1,140 @@
...
+                    _key1 = key;
+                    _value1 = value;
+                    break;
+
+                case 1:
+                    if (EqualityComparer<TKey>.Default.Equals(key, _key1!))
+                    {
+                        duplicateKeyDetected = true;
+                        _value1 = value;
+                        break;
```
AlexanderSher, line 43: Can we use `return` here instead of `break`?

sdk/core/Azure.Core/src/Internal/ArrayBackedPropertyBag.cs
AlexanderSher: I'm not sure I understand why we optimize for writing rather than for reading. Is it a common case that we store values for the keys that are never used?

sdk/core/Azure.Core/src/HttpMessage.cs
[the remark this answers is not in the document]
KrzysztofCwalina: @AlexanderSher, yes, but I am not sure we care about this code path. It's basically legacy. Also, if we make the bag be a struct, I would keep the long-object one as a Dictionary. The new bag is very large (8 longs), even when empty, and so we don't want it to be inlined in the message. Dictionary when empty is just one null long.

sdk/core/Azure.Core/src/Internal/ArrayBackedPropertyBag.cs
```diff
@@ -0,0 +1,149 @@
...
+                return false;
+            }
+            else
+            {
+                value = GetAt(index);
+                return true;
+            }
+        }
+
+        public void Add(ulong key, object value)
```
KrzysztofCwalina, line 50: I would call it Set. These semantics are typically called "set" in .NET.

sdk/core/Azure.Core/src/Internal/ArrayBackedPropertyBag.cs
```diff
@@ -0,0 +1,149 @@
...
+                return _second.Value;
+            return _rest[index - 2].Value;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private int GetIndex(ulong key)
+        {
+            if (_count == 0)
+                return -1;
+            if (_first.Key == key)
```
KrzysztofCwalina, line 127: this might accidentally succeed for key == 0

christothes: Good catch, I removed the previous check by accident.

sdk/core/Azure.Core/src/Internal/ArrayBackedPropertyBag.cs
```diff
@@ -0,0 +1,149 @@
...
+                return _second.Value;
+            return _rest[index - 2].Value;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private int GetIndex(ulong key)
+        {
+            if (_count == 0)
+                return -1;
+            if (_count > 0 &&_first.Key == key)
```
KrzysztofCwalina, line 127: nit: space needed

KrzysztofCwalina (approved):

PR #33490, opened 2023-01-14
[Azure.Monitor.Ingestion] EventHandler

nisha-bhatia:


sdk/monitor/Azure.Monitor.Ingestion/src/Models/UploadLogsOptions.cs
```diff
@@ -21,5 +28,18 @@ public class UploadLogsOptions
         /// <remarks> In the Upload method, this parameter is not used as the batches are uploaded in sequence. For parallel uploads, if this value is not set the default concurrency will be 5. </remarks>
         /// </summary>
         public int MaxConcurrency { get; set; }
+
+        /// <summary>
+        /// test
+        /// </summary>
+        public event SyncAsyncEventHandler<UploadFailedArgs> UploadFailed;
+
+        private ClientDiagnostics _clientDiagnostics = new ClientDiagnostics(new LogsIngestionClientOptions());
```
JoshLove-msft, line 37: This should be the same clientDiagnostics isntance used by the client.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -360,6 +375,56 @@ public virtual async Task<Response> UploadAsync<T>(string ruleId, string streamN
...
+        }
+
+        /// <summary>
+        /// test
+        /// </summary>
+        /// <param name="logsCount"></param>
+        /// <param name="options"></param>
+        /// <param name="response"></param>
+        /// <param name="cancellationToken"></param>
+        protected internal virtual async Task OnExceptionAsync(int logsCount, UploadLogsOptions options, Response response, CancellationToken cancellationToken = default)
```
JoshLove-msft, line 410: Instead of taking the response and logs count, should this take the UploadFailedArgs?

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -79,18 +81,17 @@ internal HttpMessage CreateUploadRequest(string ruleId, string streamName, Reque
         /// <summary>
         /// Hidden method for batching data - serializing into arrays of JSON no more than SingleUploadThreshold each
         /// </summary>
-        /// <typeparam name="T"></typeparam>
         /// <param name="logEntries"></param>
         /// <param name="options"></param>
         /// <returns></returns>
-        internal static IEnumerable<BatchedLogs<T>> Batch<T>(IEnumerable<T> logEntries, UploadLogsOptions options = null)
+        internal static IEnumerable<BatchedLogs> Batch(IEnumerable<object> logEntries, UploadLogsOptions options = null)
```
JoshLove-msft, line 87: Can we verify that the change from T to object is necessary?

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -4,6 +4,8 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Net.NetworkInformation;
+using System.Runtime.InteropServices;
```
JoshLove-msft, line 8: Why were these usings needed?

sdk/monitor/Azure.Monitor.Ingestion/api/Azure.Monitor.Ingestion.netstandard2.0.cs
```diff
@@ -6,10 +6,12 @@ public partial class LogsIngestionClient
         public LogsIngestionClient(System.Uri endpoint, Azure.Core.TokenCredential credential) { }
         public LogsIngestionClient(System.Uri endpoint, Azure.Core.TokenCredential credential, Azure.Monitor.Ingestion.LogsIngestionClientOptions options) { }
         public virtual Azure.Core.Pipeline.HttpPipeline Pipeline { get { throw null; } }
+        protected internal virtual void OnException(int logsCount, Azure.Monitor.Ingestion.UploadLogsOptions options, Azure.Response response, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { }
```
JoshLove-msft, line 9: I'm not sure that we will need the sync version. It seems that with the SyncAsyncEventHandler even for the sync case, users will need to define a method that returns a Task.

sdk/monitor/Azure.Monitor.Ingestion/api/Azure.Monitor.Ingestion.netstandard2.0.cs
```diff
@@ -19,11 +21,18 @@ public enum ServiceVersion
             V2021_11_01_Preview = 1,
         }
     }
+    public partial class UploadFailedArgs : Azure.SyncAsyncEventArgs
+    {
+        internal UploadFailedArgs() : base (default(bool), default(System.Threading.CancellationToken)) { }
```
JoshLove-msft, line 26: This should be public and take the properties as parameters.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -229,7 +229,7 @@ public virtual Response Upload<T>(string ruleId, string streamName, IEnumerable<
 
                     if (response.Status != 204)
                     {
-                        throw new RequestFailedException(response);
+                        OnException(batch.LogsCount, options, response, cancellationToken);
```
JoshLove-msft, line 232: Do we need to clone the options so that users don't change the event handler underneath us?

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -360,6 +375,56 @@ public virtual async Task<Response> UploadAsync<T>(string ruleId, string streamN
...
         }
 
+        /// <summary>
+        /// test
+        /// </summary>
+        /// <param name="logsCount"></param>
+        /// <param name="options"></param>
+        /// <param name="response"></param>
+        /// <param name="cancellationToken"></param>
+        protected internal virtual async void OnException(int logsCount, UploadLogsOptions options, Response response, CancellationToken cancellationToken = default)
```
JoshLove-msft, line 385: This probably belongs on the Options type rather than the client to go along with where the event is defined.

sdk/monitor/Azure.Monitor.Ingestion/src/Models/UploadLogsOptions.cs
nisha-bhatia: I followed this example where the ClientDiagnostics are also made in the same class. I don't think I can pass in the same ClientDiagnostics between classes.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
nisha-bhatia: I think this is because of the use of SyncAsyncEventArgs and RaiseAsync? @tg-msft

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
nisha-bhatia: I'm not sure how they got added - I'll remove them!

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
nisha-bhatia: In the RaiseAsync code, a reference to the EventHandler is used so it can't be modified then I believe

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
nisha-bhatia: I made the Args in the InvokeEvent method on the UploadLogsOptions class to follow the sample: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/tests/samples/FakeClients/AlarmClient.cs#L35. The Args use the isRunningSynchronously param

sdk/monitor/Azure.Monitor.Ingestion/src/Models/UploadLogsOptions.cs
JoshLove-msft: Why can't we pass it? The idea is that we want to use the diagnostics based on the user specified options. Instead we are creating a new diagnostics instance that ignores their settings.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -328,7 +328,14 @@ public virtual async Task<Response> UploadAsync<T>(string ruleId, string streamN
...
                             }
                             // Check completed task for Exception/RequestFailedException and increase logsFailed count
-                            ProcessCompletedTask(runningTasks[i], ref exceptions, ref logsFailed);
+                            if (options == null)
+                            {
+                                ProcessCompletedTask(runningTasks[i], ref exceptions, ref logsFailed);
+                            }
+                            else
+                            {
+                                await OnExceptionAsync(runningTasks[i].LogsCount, options, runningTasks[i].CurrentTask.Result, cancellationToken).ConfigureAwait(false);
```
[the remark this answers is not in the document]
JoshLove-msft: but how do we make the jump from options not being null, to there being an error?

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
JoshLove-msft: We wouldn't observe any modifications after we call RaiseAsync, but I thought we wanted to avoid having the user modify it out from under us after they call the Upload method.

sdk/monitor/Azure.Monitor.Ingestion/api/Azure.Monitor.Ingestion.net461.cs
```diff
@@ -7,9 +7,9 @@ public partial class LogsIngestionClient
...
         public virtual Azure.Core.Pipeline.HttpPipeline Pipeline { get { throw null; } }
         public virtual Azure.Response Upload(string ruleId, string streamName, Azure.Core.RequestContent content, string contentEncoding = null, Azure.RequestContext context = null) { throw null; }
+        public virtual Azure.Response Upload(string ruleId, string streamName, System.Collections.Generic.IEnumerable<object> logs, Azure.Monitor.Ingestion.UploadLogsOptions options = null, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
         public virtual System.Threading.Tasks.Task<Azure.Response> UploadAsync(string ruleId, string streamName, Azure.Core.RequestContent content, string contentEncoding = null, Azure.RequestContext context = null) { throw null; }
-        public virtual System.Threading.Tasks.Task<Azure.Response> UploadAsync<T>(string ruleId, string streamName, System.Collections.Generic.IEnumerable<T> logs, Azure.Monitor.Ingestion.UploadLogsOptions options = null, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
```
tg-msft, line 11: Where are the overloads with `T` going?  We shouldn't be getting rid of those.

sdk/monitor/Azure.Monitor.Ingestion/api/Azure.Monitor.Ingestion.net461.cs
```diff
@@ -19,11 +19,20 @@ public enum ServiceVersion
...
+        public System.Exception Exception { get { throw null; } }
+        public int FailedLogs { get { throw null; } }
+    }
     public partial class UploadLogsOptions
     {
         public UploadLogsOptions() { }
         public int MaxConcurrency { get { throw null; } set { } }
         public Azure.Core.Serialization.ObjectSerializer Serializer { get { throw null; } set { } }
+        public event Azure.Core.SyncAsyncEventHandler<Azure.Monitor.Ingestion.UploadFailedArgs> UploadFailed { add { } remove { } }
+        protected internal static void OnException(int logsCount, Azure.Monitor.Ingestion.UploadLogsOptions options, Azure.Response response, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { }
```
tg-msft, line 34: This should just be `internal`.  There's no need to make it visible to customers with `protected` too.  I'd also give it a more specific name than `OnException` because it's just for raising `UploadFailed`.

sdk/monitor/Azure.Monitor.Ingestion/api/Azure.Monitor.Ingestion.netstandard2.0.cs
tg-msft: Do we have a `ModelFactory`?  I think it makes more sense to have an entry there rather than make this constructible.

sdk/monitor/Azure.Monitor.Ingestion/api/Azure.Monitor.Ingestion.netstandard2.0.cs
```diff
@@ -19,11 +19,20 @@ public enum ServiceVersion
...
+        public System.Exception Exception { get { throw null; } }
+        public int FailedLogs { get { throw null; } }
```
tg-msft, line 26: This property name reads like a collection of the logs.  I think we should use `FailedLogCount` or similar to better describe what it's communication.

sdk/monitor/Azure.Monitor.Ingestion/api/Azure.Monitor.Ingestion.netstandard2.0.cs
```diff
@@ -19,11 +19,20 @@ public enum ServiceVersion
             V2021_11_01_Preview = 1,
         }
     }
+    public partial class UploadFailedArgs : Azure.SyncAsyncEventArgs
```
tg-msft, line 22: You need `EventArgs` in the name.  See the derived classes at https://learn.microsoft.com/en-us/dotnet/api/system.eventargs?view=net-7.0 for comparison.  `UploadFailedEventArgs` would be better.

sdk/monitor/Azure.Monitor.Ingestion/api/Azure.Monitor.Ingestion.netstandard2.0.cs
```diff
@@ -19,11 +19,20 @@ public enum ServiceVersion
...
+    {
+        internal UploadFailedArgs() : base (default(bool), default(System.Threading.CancellationToken)) { }
+        public System.Exception Exception { get { throw null; } }
+        public int FailedLogs { get { throw null; } }
+    }
```
tg-msft, line 27: I thought we were going to add `IReadOnlyList<object> FailedLogs { get; }` to this as well so people could do things like attempt to resubmit if they wanted.  If we had that, we could also drop the count because it'd just be `FailedLogs.Count`.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
tg-msft: No, we shouldn't do this here.  We should leave this alone but cast `T` to `object` when passing entries into the event args/helpers so we don't need to make all of them strongly typed (i.e., we don't want to make it `UploadLogsOptions<T>` to have a `SyncAsyncEventHandler<UploadFailedEventArgs<T>>`, etc., so we'll just use `object` there and people can cast back to `T` if they want).

sdk/monitor/Azure.Monitor.Ingestion/src/Models/UploadLogsOptions.cs
tg-msft: Yes, we should pass the client's into the `InvokeEvent` method.

tg-msft (commented):
Leaving a few quick comments but I haven't had a chance to look through completely yet.

sdk/monitor/Azure.Monitor.Ingestion/api/Azure.Monitor.Ingestion.netstandard2.0.cs
JoshLove-msft: FWIW, in Search the event args has a public ctor. Same thing for Service Bus. Model Factory is typically used for output model types.

sdk/monitor/Azure.Monitor.Ingestion/api/Azure.Monitor.Ingestion.net461.cs
JoshLove-msft: I thought it needs to be protected for extensibility/testing. At least these are exposed as protected in Search/Service Bus/Event Hubs.

sdk/monitor/Azure.Monitor.Ingestion/api/Azure.Monitor.Ingestion.net461.cs
```diff
@@ -7,9 +7,9 @@ public partial class LogsIngestionClient
         public LogsIngestionClient(System.Uri endpoint, Azure.Core.TokenCredential credential, Azure.Monitor.Ingestion.LogsIngestionClientOptions options) { }
         public virtual Azure.Core.Pipeline.HttpPipeline Pipeline { get { throw null; } }
         public virtual Azure.Response Upload(string ruleId, string streamName, Azure.Core.RequestContent content, string contentEncoding = null, Azure.RequestContext context = null) { throw null; }
+        public virtual Azure.Response Upload(string ruleId, string streamName, System.Collections.Generic.IEnumerable<object> logs, Azure.Monitor.Ingestion.UploadLogsOptions options = null, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
```
pallavit, line 10: Curious why do we have different APIVersions for ney461 and .netstandard2.0 Are they different in anyway? Or is this just because the SDK multi-targets?

sdk/monitor/Azure.Monitor.Ingestion/api/Azure.Monitor.Ingestion.net461.cs
tg-msft: Sure, there's a guideline in section 5.4 that says:

> **DO** use a protected virtual method to raise each event. This is applicable only to nonstatic events on unsealed classes—not to structs, sealed classes, or static events.
> 
> For each event, include a corresponding protected virtual method that raises the event. The purpose of the method is to provide a way for a derived class to handle the event using an override. Overriding is a more flexible, faster, and more natural way to handle base class events in derived classes. By convention, the name of the method should start with “On” and be followed with the name of the event.

Feel free to follow that, but it would look a lot different than this `OnException` method.

JoshLove-msft: Yeah, agreed that the API would need to change. My suggestion was to have it take a parameter for the UploadFailedEventArgs rather than the current parameters that it uses now.

sdk/monitor/Azure.Monitor.Ingestion/src/Models/UploadLogsOptions.cs
```diff
@@ -21,5 +28,62 @@ public class UploadLogsOptions
...
+        /// <param name="eventArgs"></param>
+        /// <param name="options"></param>
+        internal virtual async Task OnExceptionAsync(bool async, UploadFailedEventArgs eventArgs, UploadLogsOptions options)
+        {
+            try
+            {
+                if (!async)
+                {
+#pragma warning disable AZC0106 // Non-public asynchronous method needs 'async' parameter.
+                    options.InvokeEvent(eventArgs, _clientDiagnostics).EnsureCompleted();
```
nisha-bhatia, line 63: I suppressed the warning because InvokeEvent does not take in an async param as the boolean parameter is part of the UploadFailedEventArgs eventArgs that is passed in.

sdk/monitor/Azure.Monitor.Ingestion/src/Models/UploadLogsOptions.cs
```diff
@@ -21,5 +28,62 @@ public class UploadLogsOptions
...
+            await UploadFailedEventHandler.RaiseAsync(uploadFailedArgs, nameof(LogsIngestionClient), "Upload", clientDiagnostics).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// test
+        /// </summary>
+        /// <param name="async"></param>
+        /// <param name="eventArgs"></param>
+        /// <param name="options"></param>
+        internal virtual async Task OnExceptionAsync(bool async, UploadFailedEventArgs eventArgs, UploadLogsOptions options)
```
nisha-bhatia, line 56: I'm not sure if I can make OnExceptionAsync and InvokeEvent protected virtual methods because InvokeEvent takes in the ClientDiagnostics which is an internal type. I can make the OnExceptionAsync method protected and keep InvokeEvent as internal?

JoshLove-msft: InvokeEvent should not be protected. In fact, you can probably just delete that method. In terms of getting the ClientDiagnostics into the options, we can use an internal property to set it.

JoshLove-msft: Alternatively, I would be fine with punting this protected method for now for the sake of time.

sdk/monitor/Azure.Monitor.Ingestion/api/Azure.Monitor.Ingestion.netstandard2.0.cs
tg-msft: `EventArgs` are effectively output model types.  You never create them as a user.  You just handle an event and they're given to you.

I think it was probably a mistake to leave the .ctors public in Search.  If we already have a `ModelFactory`, I'd consider moving these there.  If we don't, I wouldn't create a `ModelFactory` just for them.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
tg-msft: I could go either way on this one.  Always referencing a fresh copy means people can do things like remove their handler if they decide they want to stop listening after the first time the event is raised.  The race condition situation where we assign it to a local is more about avoiding a very easy to make NullRefEx.

sdk/monitor/Azure.Monitor.Ingestion/api/Azure.Monitor.Ingestion.net461.cs
tg-msft: We pull in shared source at https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/core/Azure.Core/src/Shared which multitargets at https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/Shared/GZipUtf8JsonRequestContent.cs#L44 and we end up with separate API compat listings.

We should probably file a work item to have the CI automatically diff the files under the `/api/*` folders to ensure they're identical so we never diverge the public API across runtimes.

sdk/monitor/Azure.Monitor.Ingestion/src/Models/UploadLogsOptions.cs
tg-msft: Is `EnsureCompleted` the right move instead of `GetAwaiter().GetResult()`?  `EnsureCompleted` is more of a debugging aid for us than something where customers might implement this correctly causing sync-over-async.

sdk/monitor/Azure.Monitor.Ingestion/src/Models/UploadLogsOptions.cs
tg-msft: Alternatively, you can add an internal `ClientDiagnostics` to the `EventArgs` if you want to the `Foo` event to have an `OnFoo` virtual.

JoshLove-msft: Yeah that is probably cleaner.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -31,7 +31,7 @@ protected LogsIngestionClient()
         internal static string Compression;
 
         // If no concurrency count is provided for a parallel upload, default to 5 workers.
-        private const int DefaultParallelWorkerCount = 5;
+        //private const int DefaultParallelWorkerCount = 5;
```
JoshLove-msft, line 34: Is this still needed? If not, let's delete.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -342,10 +368,25 @@ public virtual async Task<Response> UploadAsync<T>(string ruleId, string streamN
                             }
                             else
                             {
-                                options._clientDiagnostics = ClientDiagnostics;
-                                var clonedOptions = options.Clone();
-                                var eventArgs = new UploadFailedEventArgs(batch.Logs, new RequestFailedException(runningTask.Result), isRunningSynchronously: true, cancellationToken);
-                                await options.OnUploadFailedAsync(async: true, eventArgs, clonedOptions).ConfigureAwait(false);
+                                UploadFailedEventArgs eventArgs;
```
JoshLove-msft, line 371: Can this be extracted into a method so that it can be reused in L417?

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
JoshLove-msft: Based on the Teams clarification with Ted, I think we can remove the clone method. We can instead just do something like:
```c#
options ??= new UploadOptions();
```
in the top of Upload.

sdk/monitor/Azure.Monitor.Ingestion/tests/ErrorTest.cs
```diff
@@ -107,5 +110,80 @@ public void TwoFailures()
...
+                Assert.AreEqual("ContentLengthLimitExceeded", ((RequestFailedException)(e.Exception)).ErrorCode);
+                Assert.IsNull(((RequestFailedException)(e.Exception)).InnerException);
+                Assert.AreEqual(413, ((RequestFailedException)(e.Exception)).Status);
+                //cts.Cancel();
+                return Task.CompletedTask;
+            }
+        }
+
+        [Test]
+        public async Task TwoFailuresWithEventHandler()
```
JoshLove-msft, line 150: It would be good to also have a test that covers the "error handler throwing" scenario.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
nisha-bhatia: If options is null and we instantiate new Options(), I don't think I would know when the EventHandler is enabled. The code currently uses null checks on Options to determine if there is an EventHandler. Could we not modify the existing options and use it as is?

JoshLove-msft: That's fair. I guess we could only do a null check for handler within the options class itself, so maybe not worth it.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -30,19 +30,16 @@ protected LogsIngestionClient()
         // If Compression wants to be turned off (hard to generate 1 Mb data gzipped) set Compression to gzip
         internal static string Compression;
 
-        // If no concurrency count is provided for a parallel upload, default to 5 workers.
-        private const int DefaultParallelWorkerCount = 5;
```
nisha-bhatia, line 34: I think we need this value

sdk/monitor/Azure.Monitor.Ingestion/tests/ErrorTest.cs
nisha-bhatia: created issue to be addressed before GA: https://github.com/Azure/azure-sdk-for-net/issues/33699

sdk/monitor/Azure.Monitor.Ingestion/tests/MonitorIngestionLiveTest.cs
```diff
@@ -220,5 +235,25 @@ public async Task ValidInputAlreadyGzipped()
...
+
+            // Make the request
+            UploadLogsOptions options = new UploadLogsOptions();
+            bool isTriggered = false;
+            options.UploadFailedEventHandler += Options_UploadFailed;
+            await client.UploadAsync(TestEnvironment.DCRImmutableId, TestEnvironment.StreamName, entries, options).ConfigureAwait(false);
+            Assert.IsFalse(isTriggered);
+            Task Options_UploadFailed(UploadFailedEventArgs e)
+            {
+                TestContext.Progress.WriteLine(e.Exception);
```
JoshLove-msft, line 253: nit: remove

sdk/monitor/Azure.Monitor.Ingestion/src/Models/UploadLogsOptions.cs
```diff
@@ -20,6 +27,61 @@ public class UploadLogsOptions
...
+                return null;
+            }
+            catch (Exception ex)
+            {
+                // return exception to caller and caller should check exception to abort processing and rethrow this exception
+                return ex;
+            }
+        }
+
+        internal static int AssertNotNegative(int argumentValue, string argumentName)
```
JoshLove-msft, line 75: We don't seem to be using this.

sdk/monitor/Azure.Monitor.Ingestion/src/Models/UploadFailedEventArgs.cs
```diff
@@ -0,0 +1,55 @@
...
+        /// <param name="isRunningSynchronously"></param>
+        /// <param name="cancellationToken"></param>
+        public UploadFailedEventArgs(List<object> failedLogs, Exception exception, bool isRunningSynchronously, CancellationToken cancellationToken) : base(isRunningSynchronously, cancellationToken)
+        {
+            FailedLogs = failedLogs;
+            Exception = exception;
+        }
+
+        /// <summary>
+        /// test
```
JoshLove-msft, line 45: Need to add ref docs as this is a public property.

sdk/monitor/Azure.Monitor.Ingestion/src/Models/UploadFailedEventArgs.cs
```diff
@@ -0,0 +1,55 @@
...
+        /// </summary>
+        public IReadOnlyList<object> FailedLogs { get; }
+        /// <summary>
+        /// test
+        /// </summary>
+        public Exception Exception { get; }
+
+        internal ClientDiagnostics _clientDiagnostics;
```
JoshLove-msft, line 53:
```suggestion
        internal ClientDiagnostics ClientDiagnostics;
```

sdk/monitor/Azure.Monitor.Ingestion/src/Models/UploadLogsOptions.cs
```diff
@@ -20,6 +27,61 @@ public class UploadLogsOptions
         /// The max concurrent requests to send to the Azure Monitor service when uploading logs.
         /// <remarks> In the Upload method, this parameter is not used as the batches are uploaded in sequence. For parallel uploads, if this value is not set the default concurrency will be 5. </remarks>
         /// </summary>
-        public int MaxConcurrency { get; set; }
+        public int MaxConcurrency { get; set; } = 5;
+
+        /// <summary>
+        /// test
```
JoshLove-msft, line 33: ref docs

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -33,18 +34,30 @@ protected LogsIngestionClient()
...
+        internal readonly struct ProcessEventHandler
+        {
+            public ProcessEventHandler(bool shouldAbort, List<Exception> exceptions)
+            {
+                ShouldAbort = shouldAbort;
+                Exceptions = exceptions;
+            }
+
+            public bool ShouldAbort { get; }
+            public List<Exception> Exceptions { get; }
```
JoshLove-msft, line 58: Why is this a list - wouldn't there only be at most one exception?

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -360,27 +423,52 @@ public virtual async Task<Response> UploadAsync<T>(string ruleId, string streamN
             return runningTasks.Select(_ => _.CurrentTask).Last().Result; //204 - response of last batch with header
         }
 
-        private static void ProcessCompletedTask((Task<Response> CurrentTask, int LogsCount) runningTask, ref List<Exception> exceptions, ref int logsFailed)
+        private static void ProcessCompletedTask((Task<Response> CurrentTask, List<object> Logs) runningTask, ref List<Exception> exceptions, ref int logsFailed)
```
JoshLove-msft, line 426: At this point, the task is completed, right? Might be better to name the parameter completedTask.

sdk/monitor/Azure.Monitor.Ingestion/src/Models/UploadFailedEventArgs.cs
```diff
@@ -0,0 +1,55 @@
...
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using Azure.Core.Pipeline;
+using Azure.Core.Serialization;
+
+namespace Azure.Monitor.Ingestion
+{
+    /// <summary>
+    /// The options model to configure the request to upload logs to Azure Monitor.
```
JoshLove-msft, line 13: This seems to be the description for the options rather than the args.

sdk/monitor/Azure.Monitor.Ingestion/src/Models/UploadFailedEventArgs.cs
```diff
@@ -0,0 +1,55 @@
...
+    /// </summary>
+    public class UploadFailedEventArgs : SyncAsyncEventArgs
+    {
+        //Add internal one with ClientDiagnostics
```
JoshLove-msft, line 17:
```suggestion
```

sdk/monitor/Azure.Monitor.Ingestion/tests/SessionRecords/MonitorIngestionLiveTest/ValidInputWithEventHandler.json
```diff
@@ -0,0 +1,5212 @@
...
+        "Authorization": "Sanitized",
+        "Content-Encoding": "gzip",
+        "Content-Length": "87",
+        "Content-Type": "application/json",
+        "traceparent": "00-04db2f94ab50f3498dd087714fe0b613-5b0745b54103ea4e-00",
+        "User-Agent": "azsdk-net-Monitor.Ingestion/1.0.0-alpha.20230127.1 (.NET Framework 4.8.9105.0; Microsoft Windows 10.0.22621 )",
+        "x-ms-client-request-id": "504441fcfc636177f273d6244cbf2fc3",
+        "x-ms-return-client-request-id": "true"
+      },
+      "RequestBody": "[[{\u0022Time\u0022:\u00222023-01-27T14:39:39.1324105\u0022,\u0022Computer\u0022:\u0022Computer0\u0022,\u0022AdditionalContext\u0022:0}]]",
```
JoshLove-msft, line 17: Seems like there is an issue here where the SingleUploadThreshold is not correct. The request sizes should match the async version.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -33,18 +34,30 @@ protected LogsIngestionClient()
...
+                Logs = logs;
                 LogsData = logsData;
             }
 
-            public int LogsCount { get; }
+            public List<object> Logs { get; }
             public BinaryData LogsData { get; }
         }
 
+        internal readonly struct ProcessEventHandler
```
JoshLove-msft, line 49: Isn't ShouldAbort based on whether the user throws an exception from their handler? If that's the case, all we need to return is the exception from our method that processes the completed task, rather than returning this type.

sdk/monitor/Azure.Monitor.Ingestion/src/Models/UploadLogsOptions.cs
nisha-bhatia: used here: https://github.com/Azure/azure-sdk-for-net/pull/33490/files#diff-6767d7871ae91cf5eb10fa3e4f24a9d5bbd9f9dbcc22bcbd72be0eabe315c103R330 Line 330 in LogsIngestionClient

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
nisha-bhatia: I think there can be an exception from the the Batch and then another one from the EventHandler? Maybe this is not a scenario though

sdk/monitor/Azure.Monitor.Ingestion/src/Models/UploadLogsOptions.cs
JoshLove-msft: Ah - any reason not to call this in the property setter?

pallavit:
>         /// Response response = client.Upload("<ruleId>", "<streamName>", data);

Did you add these samples manually? Or do we have a tool that does that for us?

---
Refers to: sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs:209 in 5f2767d. [](commit_id = 5f2767de8a6fada1d8468b01f7e5c5edba34b51a, deletion_comment = False)

nisha-bhatia:
> > ```
> >     /// Response response = client.Upload("<ruleId>", "<streamName>", data);
> > ```
> 
> Did you add these samples manually? Or do we have a tool that does that for us?
> 
> Refers to: sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs:209 in [5f2767d](https://github.com/Azure/azure-sdk-for-net/commit/5f2767de8a6fada1d8468b01f7e5c5edba34b51a). [](commit_id = [5f2767d](https://github.com/Azure/azure-sdk-for-net/commit/5f2767de8a6fada1d8468b01f7e5c5edba34b51a), deletion_comment = False)

I think these are added automatically when Update-Snippets is ran. I will try to figure out how to update this (this sample's comment is for the DPG method)

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -285,22 +312,24 @@ public virtual async Task<Response> UploadAsync<T>(string ruleId, string streamN
             Argument.AssertNotNullOrEmpty(logs, nameof(logs));
 
             // Calculate the number of threads to use.
-            // If there are 0 workers or an UploadLogsOptions object was not provided, method will run with 5 workers. Otherwise will run in parallel with number of workers given.
-            int _maxWorkerCount = (options == null || options.MaxConcurrency <= 0) ? DefaultParallelWorkerCount : options.MaxConcurrency;
+            int _maxWorkerCount = (options == null) ? new UploadLogsOptions().MaxConcurrency : options.MaxConcurrency;
```
JoshLove-msft, line 315: Why not just store the newly created options so we don't have to use check for nulls later on?

sdk/monitor/Azure.Monitor.Ingestion/src/Models/UploadLogsOptions.cs
```diff
@@ -20,6 +27,58 @@ public class UploadLogsOptions
         /// The max concurrent requests to send to the Azure Monitor service when uploading logs.
         /// <remarks> In the Upload method, this parameter is not used as the batches are uploaded in sequence. For parallel uploads, if this value is not set the default concurrency will be 5. </remarks>
         /// </summary>
-        public int MaxConcurrency { get; set; }
+        public int MaxConcurrency
+        {
+            get { return _maxConcurrency; }
+            set { AssertNotNegative(_maxConcurrency, "MaxConcurrency"); }
```
JoshLove-msft, line 33: We don't seem to be setting _maxConcurrency.

sdk/monitor/Azure.Monitor.Ingestion/src/Models/UploadLogsOptions.cs
```diff
@@ -20,6 +27,58 @@ public class UploadLogsOptions
...
+                // return exception to caller and caller should check exception to abort processing and rethrow this exception
+                return ex;
+            }
+        }
+
+        internal static int AssertNotNegative(int argumentValue, string argumentName)
+        {
+            if (argumentValue <= 0)
+            {
+                throw new ArgumentOutOfRangeException(argumentName, $"Argument {argumentName} must be a non-negative timespan value. The provided value was {argumentValue}.");
```
JoshLove-msft, line 76: This is not for a timespan, but rather an int.

eng/Packages.Data.props
```diff
@@ -282,7 +282,7 @@
   </ItemGroup>
 
   <PropertyGroup>
-    <TestProxyVersion>1.0.0-dev.20221111.1</TestProxyVersion>
+    <TestProxyVersion>1.0.0-dev.20230131.4</TestProxyVersion>
```
weshaggard, line 285: @scbedd I thought we controlled this version in eng/common. Is that not true for .NET?

scbedd: @weshaggard we do. Trying to figure wtf is going on with updating the recordings.

90% of the time _one_ or more of the recordings will randomly fail with a 400. We know it's proxy related, but at this point I have NO IDEA where it's coming from.

sdk/monitor/Azure.Monitor.Ingestion/api/Azure.Monitor.Ingestion.net461.cs
nisha-bhatia: Created: https://github.com/Azure/azure-sdk-for-net/issues/33830

sdk/monitor/Azure.Monitor.Ingestion/tests/ErrorTest.cs
```diff
@@ -107,5 +110,80 @@ public void TwoFailures()
...
+            await client.UploadAsync(TestEnvironment.DCRImmutableId, TestEnvironment.StreamName, entries, options).ConfigureAwait(false);
+            Assert.IsTrue(isTriggered);
+            Task Options_UploadFailed(UploadFailedEventArgs e)
+            {
+                isTriggered = true;
+                Assert.IsInstanceOf<RequestFailedException>(e.Exception);
+                Assert.AreEqual("ContentLengthLimitExceeded", ((RequestFailedException)(e.Exception)).ErrorCode);
+                Assert.IsNull(((RequestFailedException)(e.Exception)).InnerException);
+                Assert.AreEqual(413, ((RequestFailedException)(e.Exception)).Status);
+                //cts.Cancel();
```
JoshLove-msft, line 184:
```suggestion
```

sdk/monitor/Azure.Monitor.Ingestion/tests/ErrorTest.cs
```diff
@@ -107,5 +110,80 @@ public void TwoFailures()
...
```
JoshLove-msft, line 144:
```suggestion
```

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -283,24 +310,25 @@ public virtual async Task<Response> UploadAsync<T>(string ruleId, string streamN
             Argument.AssertNotNullOrEmpty(ruleId, nameof(ruleId));
             Argument.AssertNotNullOrEmpty(streamName, nameof(streamName));
             Argument.AssertNotNullOrEmpty(logs, nameof(logs));
+            options = (options == null) ? new UploadLogsOptions() : options;
```
JoshLove-msft, line 313:
```suggestion
            options ??= new UploadLogsOptions();
```

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -206,17 +203,19 @@ public virtual Response Upload<T>(string ruleId, string streamName, IEnumerable<
...
-
             scope.Start();
+
             // Keep track of the number of failed logs across batches
             int logsFailed = 0;
-
+            var shouldAbort = false;
             // Partition the stream into individual blocks
-            foreach (BatchedLogs<T> batch in Batch(logs, options))
+            foreach (BatchedLogs batch in Batch((List<object>)logs, options))
```
JoshLove-msft, line 214: Why are we casting to a list?

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -229,16 +228,44 @@ public virtual Response Upload<T>(string ruleId, string streamName, IEnumerable<
 
                     if (response.Status != 204)
                     {
-                        throw new RequestFailedException(response);
+                        // if there is no Handler on options, throw exception otherwise raise Handler
+                        if (options == null || !options.HasHandler)
```
JoshLove-msft, line 232: Can we use the same options approach as in the async method, i.e.
```c#
options ??= new Options();
```

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -328,7 +356,17 @@ public virtual async Task<Response> UploadAsync<T>(string ruleId, string streamN
...
                             }
                             // Check completed task for Exception/RequestFailedException and increase logsFailed count
-                            ProcessCompletedTask(runningTasks[i], ref exceptions, ref logsFailed);
+                            if (!options.HasHandler)
+                            {
+                                ProcessCompletedTask(runningTasks[i], ref exceptions, ref logsFailed);
+                            }
+                            else
+                            {
+                                Exception exceptionEventHandler = ProcessCompletedTaskEventHandlerAsync(runningTask, batch.Logs, options, cancellationToken).Result;
```
JoshLove-msft, line 365: Don't we need to await this?

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
nisha-bhatia: we could, we would just not use the Concurrency param. I'll add it so it's the same in both cases

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -347,7 +385,18 @@ public virtual async Task<Response> UploadAsync<T>(string ruleId, string streamN
...
             {
-                ProcessCompletedTask(task, ref exceptions, ref logsFailed);
+                // Check completed task for Exception/RequestFailedException and increase logsFailed count
+                if (!options.HasHandler)
+                {
+                    ProcessCompletedTask(task, ref exceptions, ref logsFailed);
+                }
+                else
+                {
+                    Exception exceptionEventHandler = ProcessCompletedTaskEventHandlerAsync(task.CurrentTask, task.Logs, options, cancellationToken).Result;
```
JoshLove-msft, line 395: Don't we need to await this?

sdk/monitor/Azure.Monitor.Ingestion/tests/Samples/LogDataAndQuery.cs
```diff
@@ -75,6 +77,41 @@ public void LogData()
             #endregion
         }
 
+        public void LogDataIEnumerable()
+        {
+            #region Snippet:UploadCustomLogsIEnumerable
```
JoshLove-msft, line 82: Where are these snippets being used?

nisha-bhatia: I think these just become samples a user can look at. I don't believe they are used anywhere

JoshLove-msft: I don't think so - they need to be included in an md file. @heaths don't we have validation to fail the Analyze step for unused snippets?

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
nisha-bhatia: The helper method Batch no longer has type <T>, let me figure out why I removed it

sdk/monitor/Azure.Monitor.Ingestion/tests/Samples/LogDataAndQuery.cs
heaths: We do, and it works because I ran into it for a couple service directories when I recently updated the snippet generator for unrelated reasons. Seems the "Build Snippets" step of the Analyze job was disabled:

image

Someone could look at the pipeline templates and figure out where one of those `False` settings come from.

JoshLove-msft: Looks like we are setting it to False by default - https://github.com/Azure/azure-sdk-for-net/pull/27551/files#diff-23849bb529451be6bf226bc895f60373a9181f673e6273d84474158a912d795bR22

nisha-bhatia: I always run the script manually - .\eng\scripts\Update-Snippets.ps1 monitor. I've never had the pipeline run the script automatically - (is this how it is intended to be run)?

JoshLove-msft: So @nisha-bhatia, do you get a failure when you run Update-Snippets?

nisha-bhatia: image

nisha-bhatia: oh I think I might have to reference every snippet in the ReadMe?

heaths: I wouldn't spam the README. Stick to hero scenarios. If you don't use a sample in the README or another sample markdown file under the project's "samples/" directory, don't put a `#region` around it. Then again, there's not much point in sample sources if you don't extract from them into sample markdowns.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
JoshLove-msft: It doesn't look like this was updated.

nisha-bhatia: missed two! sorry

JoshLove-msft: I don't see any null coalescing statement though.

heaths: Ah, yes. It's coming back to me. Pavel initially defaulted this to `false` because so many would've failed, but once we got clean we - the client library owners - need to set it `true` e.g., https://github.com/Azure/azure-sdk-for-net/pull/27551/files#diff-d6d7fdd20f8d32d2438f28c46f37f8bb028a8873ddfa90a12eb6bad3765159bbR29. IIRC, there was an issue on getting clean and enabling. I'll see if I can find and resurrect it; otherwise, I'll create a new one. We should drive to make this `true` by default.

sdk/monitor/Azure.Monitor.Ingestion/tests/Samples/LogDataAndQuery.cs
heaths: Ah, yes. It's coming back to me. Pavel initially defaulted this to `false` because so many would've failed, but once we got clean we - the client library owners - need to set it `true` e.g., https://github.com/Azure/azure-sdk-for-net/pull/27551/files#diff-d6d7fdd20f8d32d2438f28c46f37f8bb028a8873ddfa90a12eb6bad3765159bbR29. IIRC, there was an issue on getting clean and enabling. I'll see if I can find and resurrect it; otherwise, I'll create a new one. We should drive to make this `true` by default.

JoshLove-msft (approved):

sdk/monitor/Azure.Monitor.Ingestion/tests/Samples/LogDataAndQuery.cs
nisha-bhatia: Reverted samples

PR #33520, opened 2023-01-16
Fix double dots in template file name extensions

### LINKED ISSUE #33515: [BUG] "dotnet new azsdkdpg" generate unexpected file after PR 32814
### Library name and version

dotnet new azsdkdpg

### Describe the bug

This [PR](https://github.com/Azure/azure-sdk-for-net/pull/32814) breaks the command `dotnet new azsdkdpg --name Azure.XXX.Contosoz --clientName Contosoz --groupName XXX --serviceDirectory contoso --force` to generate a valid template. For example, there is a genereated file named `Azure.XXX..csproj` with two dots. It blocks all sdk automation pipeline for cadl.

I think the tool `dotnet new azsdkdpg` need to be fixed.

### Expected behavior

Generate valid file

### Actual behavior

Generate invalid file

### Reproduction Steps

See description

### Environment

_No response_

heaths:
Fixes #33515 and may be solution or work around to dotnet/templating#5955

### DIFF
--- a/sdk/template/.template.config/template.json
+++ b/sdk/template/.template.config/template.json
@@ -1,6 +1,9 @@
 {
   "$schema": "http://json.schemastore.org/template",
-  "classifications": [ "Azure", "ClassLibrary" ],
+  "classifications": [
+    "Azure",
+    "ClassLibrary"
+  ],
   "identity": "Azure.Template",
   "name": "Azure DataPlane SDK template: client project",
   "shortName": "azsdkdpg",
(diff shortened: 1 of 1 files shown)


heaths:
This may fix or at least work around the problem. In testing with 	ree /f, I don't see any double dots after this. File names appear right, but it may be good to pull this branch and test it with the automation process. If you have the [GitHub CLI](https://github.com/cli/cli) you can run gh pr clone #33520, or you can add my remote to your local clone and checkout my branch e.g., git remote add heaths https://github.com/heaths/azure-sdk-for-net.git && git fetch heaths && git checkout heaths/issue33515.

heaths:
Note: I also formatted the document since editors like VSCode are going to at least reformat code one modifies. In general, it's best to keep code formatted so frequent re-formatting doesn't happen.

If you look at the PR and hide whitespace changes, you'll find it's very small.

sdk/template/.template.config/template.json
```diff
@@ -328,43 +332,44 @@
...
+          {
+            "type": "const",
+            "value": "Microsoft"
+          },
+          {
+            "type": "ref",
+            "value": "ServiceNameCapitalCase"
+          }
+        ],
+        "separator": "."
```
chunyu3, line 372: Shall we add `"removeEmptyValues": true` here also?

chunyu3 (approved):

chunyu3:
Hello @heaths  I tried this solution in test env, it works.  After the PR is merged, I will try it in prod env. Thanks.

sdk/template/.template.config/template.json
```diff
@@ -254,7 +257,8 @@
             "value": "ClientNameCapitalCase"
           }
         ],
-        "separator": "."
+        "separator": ".",
+        "removeEmptyValues": true
```
jsquire, line 261: These entries seem to be the fix, no?

jsquire (approved):

sdk/template/.template.config/template.json
heaths: That's correct.

sdk/template/.template.config/template.json
heaths: I suppose it couldn't hurt, but may not be needed if you found everything else worked. You should test again with my update just to make sure.

heaths:
@chunyu3 can you test this again? I generated a new version and checked a few files' contents, but a more thorough test like you did last night would be great.

chunyu3:
> @chunyu3 can you test this again? I generated a new version and checked a few files' contents, but a more thorough test like you did last night would be great.

Hello @heaths  I have tried in test env, it works. Thanks

PR #33541, opened 2023-01-17
Adapt the script to the input of docker task engine

dw511214992:
Different from the input of spec unified pipeline, the `specFolder` and `relatedCadlProjectFolder`  of docker task engine's input looks like:
```
{
  "specFolder": "/spec-repo/specification",

  "relatedCadlProjectFolder": "contoso/Contoso"
}
```
We need to change their values before generating correct `cadl-location.yaml` file. 
The reason of specifing `relatedCadlProjectFolder`  to the relative path of specification is to adapt to shift-left pipeline, which doesn't have the `specification` folder in the spec repo. 
Add @chunyu3 for awareness.


dw511214992:
close this PR as it's not a good fix and @chunyu3 will fix it.

PR #33544, opened 2023-01-17
Fix regional endpoint validation error

### LINKED ISSUE #32498: Azure.Identity Question

christothes:
fixes #32498


sdk/identity/Azure.Identity/src/MsalConfidentialClient.cs
```diff
@@ -12,7 +12,7 @@ namespace Azure.Identity
...
+        private const string s_instanceMetadata = "{\"tenant_discovery_endpoint\":\"https://login.microsoftonline.com/common/v2.0/.well-known/openid-configuration\",\"api-version\":\"1.1\",\"metadata\":[{\"preferred_network\":\"login.microsoftonline.com\",\"preferred_cache\":\"login.windows.net\",\"aliases\":[\"login.microsoftonline.com\",\"login.windows.net\",\"login.microsoft.com\",\"sts.windows.net\"]},{\"preferred_network\":\"login.partner.microsoftonline.cn\",\"preferred_cache\":\"login.partner.microsoftonline.cn\",\"aliases\":[\"login.partner.microsoftonline.cn\",\"login.chinacloudapi.cn\"]},{\"preferred_network\":\"login.microsoftonline.de\",\"preferred_cache\":\"login.microsoftonline.de\",\"aliases\":[\"login.microsoftonline.de\"]},{\"preferred_network\":\"login.microsoftonline.us\",\"preferred_cache\":\"login.microsoftonline.us\",\"aliases\":[\"login.microsoftonline.us\",\"login.usgovcloudapi.net\"]},{\"preferred_network\":\"login-us.microsoftonline.com\",\"preferred_cache\":\"login-us.microsoftonline.com\",\"aliases\":[\"login-us.microsoftonline.com\"]}]}";
```
schaabs, line 15: As we discussed offline, it might be better to use some hardcoded invalid AuthorityHost here rather than trying to replicate the metadata for all clouds. That way we can provide a minimal constant set of metadata for that invalid host, and at the same time guarantee that the MSAL client would not be able to contact the AuthorityHost if for some reason it usurps our callback and tries to get a token from thee Authority.

schaabs (changes requested):

sdk/identity/Azure.Identity/src/MsalConfidentialClient.cs
```diff
@@ -12,7 +12,8 @@ namespace Azure.Identity
...
+        // The instance metadata our callback returns only contains a known aka.ms link to ensure it will match the authority sent when a _appTokenProviderCallback is configured.
+        private const string s_instanceMetadata ="{\"tenant_discovery_endpoint\":\"https://aka.ms/azsdk/net/identity/managedidentitycredential/troubleshoot\",\"api-version\":\"1.1\",\"metadata\":[{\"preferred_network\":\"aka.ms\",\"preferred_cache\":\"aka.ms\",\"aliases\":[\"aka.ms\"]}]}";
```
schaabs, line 16: I like this approach more than trying to get a static set of metadata for actual cloud authorities. I just want to check with @bgavrilMS to see if he sees any possible problems with this approach.

sdk/identity/Azure.Identity/src/MsalConfidentialClient.cs
```diff
@@ -75,7 +76,7 @@ protected override async ValueTask<IConfidentialClientApplication> CreateClientA
             if (_appTokenProviderCallback != null)
             {
                 confClientBuilder.WithAppTokenProvider(_appTokenProviderCallback)
-                    .WithAuthority(_authority.AbsoluteUri, TenantId, false)
+                    .WithAuthority("https://aka.ms/azsdk/net/identity/managedidentitycredential/troubleshoot", TenantId, false)
```
bgavrilMS, line 79: Have you tried using the new option to completely disable instance metadata? We tested that it plays well with regional. `WithInstanceDiscovery(false)`

bgavrilMS, line 79: Not sure about this. MSAL tries to parse the authority to determine if it's AAD or ADFS or B2C. It might fallback to AAD here, but it seems brittle.

eng/Packages.Data.props
```diff
@@ -104,7 +104,7 @@
     <!-- Other approved packages -->
     <PackageReference Update="Microsoft.Azure.Amqp" Version="2.6.1" />
     <PackageReference Update="Microsoft.Azure.WebPubSub.Common" Version="1.2.0" />
-    <PackageReference Update="Microsoft.Identity.Client" Version="4.46.0" />
+    <PackageReference Update="Microsoft.Identity.Client" Version="4.49.1" />
```
schaabs, line 107: You might consider rev'ing the Extensions and Broker packages at the same time.

schaabs (approved):

PR #33582, opened 2023-01-19
Fix an edge case around calculating resourcetype from resourceid

m-nash:

### DIFF
--- a/sdk/core/Azure.Core/src/ResourceIdentifier.cs
+++ b/sdk/core/Azure.Core/src/ResourceIdentifier.cs
@@ -197,7 +197,8 @@ namespace Azure.Core
 
         private static ResourceType ChooseResourceType(ReadOnlySpan<char> resourceTypeName, ResourceIdentifier parent, out SpecialType specialType)
         {
-            if (resourceTypeName.Equals(ResourceGroupKey.AsSpan(), StringComparison.OrdinalIgnoreCase))
+            //resourceGroups type is Microsoft.Resources/resourceGroups only when its parent is Subscription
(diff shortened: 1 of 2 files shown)


sdk/core/Azure.Core/src/ResourceIdentifier.cs
```diff
@@ -197,7 +197,8 @@ private T GetValue<T>(ref T value)
 
         private static ResourceType ChooseResourceType(ReadOnlySpan<char> resourceTypeName, ResourceIdentifier parent, out SpecialType specialType)
         {
-            if (resourceTypeName.Equals(ResourceGroupKey.AsSpan(), StringComparison.OrdinalIgnoreCase))
+            //resourceGroups type is Microsoft.Resources/resourceGroups only when its parent is Subscription
+            if (resourceTypeName.Equals(ResourceGroupKey.AsSpan(), StringComparison.OrdinalIgnoreCase) && parent.ResourceType == ResourceType.Subscription)
```
JoshLove-msft, line 201: Would it be faster to do the ResourceType check first?

JoshLove-msft (approved):

ArcturusZhang (approved):

sdk/core/Azure.Core/src/ResourceIdentifier.cs
m-nash: the typename is smaller than resourcetype of parent so it will quickly eliminate most and not have to do the second check.

Yao725 (approved):

christothes (approved):

PR #33593, opened 2023-01-19
[AzureMonitorExporter] Add support for AADAudience

TimothyMothra:
Add support for parsing the AADAudience value from the Connection String.
This is necessary to support AAD in Sovereign clouds.

## Changes
- add static class `AadHelper` to build the scope value
- update `ConnectionStringParser.cs` to parse AADAudience

## Note
As of today, AAD is not available for Azure Monitor in Sovereign clouds.
To test this feature, I set a value in the connection string for a resource in Public Azure, and I verified that it was correctly parsed and that telemetry was successfully ingested.

### DIFF
--- a/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/CHANGELOG.md
+++ b/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Features Added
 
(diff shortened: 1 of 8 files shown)


rajkumar-rangaraj:
Did you validate if the data flows to application insights component with AAD enabled connectionstring?

TimothyMothra:
> Did you validate if the data flows to application insights component with AAD enabled connectionstring?

Yes. I validated it in Public Azure. Will explain more in our standup.

rajkumar-rangaraj (approved):
LGTM

PR #33595, opened 2023-01-19
Remove CODEOWNERS rules `/**/ci.yml` and `/**/tests.yml`

konrad-jamrozik:
Per these comments:
- https://github.com/Azure/azure-sdk-tools/pull/5088#issuecomment-1397330147
- https://github.com/Azure/azure-sdk-tools/pull/5088#issuecomment-1397427097

Related work:
- #33584
- https://github.com/Azure/azure-sdk-tools/issues/4859
- https://github.com/Azure/azure-sdk-tools/pull/5088
- https://github.com/Azure/azure-sdk-tools/issues/2770

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -723,8 +723,6 @@
 # ######## Eng Sys ########
 /eng/           @hallipr @weshaggard @benbp
 /eng/mgmt/      @ArthurMa1978 @m-nash @markcowl
-/**/tests.yml   @hallipr @benbp
-/**/ci.yml      @hallipr @benbp
 
 # Add owners for notifications for specific pipelines
 /eng/pipelines/aggregate-reports.yml            @jsquire @maririos @pallavit


weshaggard (approved):
FYI @benbp

jsquire (approved):

benbp:
Sorry I guess had missed some of the context from https://github.com/Azure/azure-sdk-tools/pull/5088#issuecomment-1397330147, does this mean we won't get notified on ci/tests yml file changes, or that we are replacing that with a new codeowner or other config entry?

konrad-jamrozik:
> Sorry I guess had missed some of the context from [Azure/azure-sdk-tools#5088 (comment)](https://github.com/Azure/azure-sdk-tools/pull/5088#issuecomment-1397330147), does this mean we won't get notified on ci/tests yml file changes, or that we are replacing that with a new codeowner or other config entry?

@benbp see the Excel diff I linked to in this comment:
- https://github.com/Azure/azure-sdk-tools/pull/5088#issuecomment-1397427097
The tab with `_2` suffix.

`PATH` column is the file path considered. `LEFT PATH EXPRESSION` is the rule that matched **before this change**, and corresponding owners are in `LEFT OWNERS`. These people are currently getting build failure notification for given path. After this PR is merged, instead the `RIGHT OWNERS` will be notified.

benbp:
@konrad-jamrozik I see. I haven't been involved in these discussions so I don't want to block anything, but I did rely on being able to monitor pipeline yaml changes across all azure sdk pull requests. Is the plan to get rid of Wes and I as codeowners for those files permanently?

konrad-jamrozik:
> @konrad-jamrozik I see. I haven't been involved in these discussions so I don't want to block anything, but I did rely on being able to monitor pipeline yaml changes across all azure sdk pull requests. Is the plan to get rid of Wes and I as codeowners for those files permanently?

@benbp I believe so, but I am not sure if Wes was aware you rely on these rules. @weshaggard can you chip-in?

weshaggard:
@benbp we have a conflict of interest here when it comes to supporting wildcards and using the ci/test.yml file to identify pipeline notification owners. Once we add the wildcard support these patterns will make us the owners for all pipelines and get notified as opposed to the area owners. I agree it is somewhat nice to be notified when folks change those files but I also think at this point most folks simply copy/paste those and not really a lot for us to review and give it gives us lots of noise. We will of course still get notified for anything under /eng/ but not necessarily for every ci/test.yml file. If you feel strongly about still getting notified we can try to come up with another mechanism to make that work.

benbp:
@weshaggard I don't think my minority case is important enough to get in the way of the larger goal here. I'll figure out some automation for my own review workflows just wanted to clarify whether this functionality was going away.

weshaggard:
@benbp we can talk about other workflow options next week in our 1:1.

konrad-jamrozik:
@weshaggard @benbp we do own the `CODEOWNERS` interpreter we use for routing build failure notifications, so I would be happy to adjust it to make Ben's life easier. It could be as simple as hardcoding some rules.

benbp:
@konrad-jamrozik is my understanding correct that codeowners determines both who GitHub assigns as reviewers and that we use it for setting up build notifications? In this case it's the former functionality that I relied on, which I don't think we get in the middle of.

Maybe there's some sort of comment syntax we could parse to exclude certain codeowners directives from the notification parser? I don't know how strict the syntax is.

konrad-jamrozik:
> @konrad-jamrozik is my understanding correct that codeowners determines both who GitHub assigns as reviewers and that we use it for setting up build notifications? In this case it's the former functionality that I relied on, which I don't think we get in the middle of.

@benbp Yes.

It works like that:

`CODEOWNERS -> GitHub interpreter -> reviewers`
`CODEOWNERS -> our interpreter -> build failure notifications`

> Maybe there's some sort of comment syntax we could parse to exclude certain codeowners directives from the notification parser? I don't know how strict the syntax is.

We could do that; we already have support for interpreting comments, which are ignored by GitHub. We were using it for FabricBot labelling logic, albeit right now it is unused.

But in such case we will have to keep `/**/ci.yml` and `/**/tests.yml` for GitHub/ PR reviewers, meaning we would have only one set of owners (including you) for each of these. If we are OK with that, then :thumbsup:.

benbp:
@konrad-jamrozik I'm happy to be marked as owner for those files. I wouldn't say block the current work on that but if we could add a notification parser ignore comment in the future (and it doesn't complicate things) it would be helpful!

konrad-jamrozik:
@benbp @weshaggard 
- https://github.com/Azure/azure-sdk-tools/issues/5181

weshaggard:
Another option I was thinking is we could fake this by asking for the owner of a non-existent file like `\sdk\service\PipelineOwner` where PipelineOwner would be in the same path as ci.yml but not be a real file but we could use it to find the owner of files in that path.

weshaggard:
@konrad-jamrozik I wasn't suggesting we would add `/sdk/service/foo/PipelineOwner` to the codeowners file. I'm suggesting in our pipeline notification tool we would find the owner of that non-existing file and it would then map to owner2 in your example instead of owner1. Today our pipeline notification looks for the owner of `/sdk/service/foo/ci.yml` but with my proposal it would look for `/sdk/service/foo/PipelineOwner` instead.

konrad-jamrozik:
@weshaggard ah, I see, thx. So basically in [this line](https://github.com/Azure/azure-sdk-tools/blob/2325b3c1348bbe51e6a36819e26af951604b8278/tools/notification-configuration/notification-creator/NotificationConfigurator.cs#L212), i.e.:

`var codeOwnerEntry = CodeownersFile.GetMatchingCodeownersEntry(process.YamlFilename, codeOwnerEntries);`

replace this:

`process.YamlFilename`

with something like:

`GetSiblingFile(target: process.YamlFilename, siblingName: "PipelineOwner")`.

With this, we can keep the `/**/ci.yml` paths and there is no need to add "ignore this path for build failure notifications" comment directive logic. I like this approach more, so I will update the work item to reflect that:
- https://github.com/Azure/azure-sdk-tools/issues/5181

weshaggard:
Yes that was the proposal. However @benbp and I just got done chatting and for now we don't want to do anything special to support this we will just remove `/**/ci.yml` patterns in the codeowners files like you did with this PR.

PR #33597, opened 2023-01-19
Add missing `/sdk/<dir>` rules

konrad-jamrozik:
This PR adds missing `/sdk/<dir>/` rules to CODEOWNERS. These dirs need assignment of appropriate owners. @pallavit  @jsquire FYI.

This change was proposed here:
- https://github.com/Azure/azure-sdk-tools/pull/5088#issuecomment-1397332193

Related work:
- #33584
- #33595
- https://github.com/Azure/azure-sdk-tools/issues/4859
- https://github.com/Azure/azure-sdk-tools/pull/5088
- https://github.com/Azure/azure-sdk-tools/issues/2770

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -13,7 +13,24 @@
 #########
 
 # Catch all
-/sdk/                                           @pallavit @jsquire
+/sdk/                          @pallavit @jsquire
+
+# Rules for directories that fell under catch-all as of 1/19/2023
+/sdk/agrifood/                 @pallavit @jsquire
+# Note there are some sub-rules for this directory, defined below
+/sdk/applicationinsights/      @pallavit @jsquire
+# Note there are some sub-rules for this directory, defined below
+/sdk/cognitiveservices/        @pallavit @jsquire
+/sdk/entra/                    @pallavit @jsquire
+/sdk/graphrbac/                @pallavit @jsquire
+/sdk/machinelearningservices/  @pallavit @jsquire
+/sdk/maps/                     @pallavit @jsquire
+# Note there are some sub-rules for this directory, defined below
+/sdk/operationalinsights/      @pallavit @jsquire
+/sdk/purview/                  @pallavit @jsquire
+# Note there are some sub-rules for this directory, defined below
+/sdk/quantum/                  @pallavit @jsquire
+/sdk/remoterendering/          @pallavit @jsquire
 
 # ######## Core Libraries ########
 


.github/CODEOWNERS
```diff
@@ -13,7 +13,24 @@
...
 # Catch all
-/sdk/                                           @pallavit @jsquire
+/sdk/                          @pallavit @jsquire
+
+# Rules for directories that fell under catch-all as of 1/19/2023
+/sdk/agrifood/                 @pallavit @jsquire
+# Note there are some sub-rules for this directory, defined below
+/sdk/applicationinsights/      @pallavit @jsquire
+# Note there are some sub-rules for this directory, defined below
+/sdk/cognitiveservices/        @pallavit @jsquire
```
jsquire, line 23: Things with sub-rules will supersede this, right?   For example, I won't be pinged every time there's something from Text Analytics or Form Recognizer that comes through?

weshaggard: I think these are a set of directories that currently don't have owners and were covered by the /sdk/ pattern before. We should find the correct owners for these paths. 

To answer your question exactly any patterns further down would override these but currently there aren't any for these paths.

jsquire: I'm not concerned with the paths themselves, but the sub-rules with owners below.   For my example:

```text
/sdk/cognitiveservices/textanalytics @someone @another
```

I just want to be sure that would override the top-level rule.

weshaggard: Those would definitely override these rules however there currently isn't an /sdk/cognitiveservices/ rule so things like /sdk/cognitiveservices/ci.yml would match against this pattern and that is the path used for pipeline notifications. So we should try to add the correct owners for that so they get notified of pipeline failures.

pallavit: What is the guidance if we have multiple partners in the same directory? Do we separate the ci pipelines? Do the service team work with the engsys team to separate them?

weshaggard: I think it depends on the teams. If we have multiple teams that we should consider the following:
- Should they be in the same service directory?
- Can they all be owners of a shared pipeline for the service directory if they should be together?
- If they should be in the same service directory but don't want to share ownership then we would need to split up the pipelines further perhaps to the package level. I would say this is the last resort, but it is possible.

jsquire (approved):

.github/CODEOWNERS
pallavit: For a lot of these directories we only have track1 SDKs with minimal to 0 activity. So, I am not sure we have clear answers to these questions.

Also, last I heard we were trying to use a central db to keep ownerships current - is that still the plan? Is it happening along with these updates?

weshaggard: Yes we are planning to try and move to a central db that would sync owners but we are still a long way from getting that all figured out.

PR #33600, opened 2023-01-19
callautomation private review internal revision

wangrui-msft:
1. Add text source constructor
2. Add polymorphism to recognizecomplete event


sdk/communication/Azure.Communication.CallAutomation/src/Models/Events/RecognizeCompleted.cs
```diff
@@ -52,7 +81,30 @@ public static RecognizeCompleted Deserialize(string content)
             using var document = JsonDocument.Parse(content);
             JsonElement element = document.RootElement;
 
-            return DeserializeRecognizeCompleted(element);
+            RecognizeCompleted parsedRecognizeCompleted = DeserializeRecognizeCompleted(element);
```
cochi2, line 84: why do we need to create the object twice? is not possible to add the below logic within DeserializerRecognizeCompleted?

wangrui-msft: yes, the reviewer require the recognizeComplete event only populate one result. Right now, the default recognizecomplete event populate two results as default behavior: 1. dtmf result 2. choice result. However, at the same time, there should only be one result be populated. That's why we add one more processing on this part.

cochi2: Is the default on an autogenerated class that we cannot modify?

cochi2 (approved):

sdk/communication/Azure.Communication.CallAutomation/src/Models/Events/RecognizeCompleted.cs
```diff
@@ -32,7 +32,36 @@ internal RecognizeCompleted(string callConnectionId, string serverCallId, string
...
+        public RecognizeCompleted(string callConnectionId, string serverCallId, string correlationId, string operationContext, ResultInformation resultInformation, CallMediaRecognitionType recognitionType, RecognizeResult recognizeResult)
+        {
+            CallConnectionId = callConnectionId;
+            ServerCallId = serverCallId;
+            CorrelationId = correlationId;
+            OperationContext = operationContext;
+            ResultInformation = resultInformation;
+            RecognitionType = recognitionType;
             ReasonCode = new ReasonCode(resultInformation.SubCode.ToString());
+            if (recognizeResult is CollectTonesResult collecToneResult)
```
Paresh-Arvind-Patil, line 57: Nit: seems like a typo collectToneResult ?

sdk/communication/Azure.Communication.CallAutomation/src/Models/Events/RecognizeCompleted.cs
wangrui-msft: > Is the default on an autogenerated class that we cannot modify?

Yes, correct, the event schema make the RecognizeComplete not allowed to modify on serialize/deserialize

sdk/communication/Azure.Communication.CallAutomation/src/Models/Events/RecognizeCompleted.cs
wangrui-msft: good catch, will fix it

Issue #33488, opened 2023-01-13
Add multi-targeting for Key Vault SDKs


heaths:
Before we add multi-targeting for all track 2 SDKs in #32597, do it first in Key Vault beta SDKs.

heaths:
After some [initial work](https://github.com/heaths/azure-sdk-for-net/tree/feature/multi-target-keyvault) and seeing how many permutations of TFMs we'd need to eliminate all reflection - at least 5, since `netstandard2.0` is lacking many APIs that are even declared in `net47`, some in `net472`, and some starting in `netcoreapp3.0`, even though some are defined in `netstandard2.0` that aren't actually in `net461` - @pallavit and I have decided not to take advantage of multi-targeting for Key Vault SDKs. The reflection-based light-up code is written to be as efficient as possible, though this may be a blocker for #24238. In that case, if/when we decide to support trimming (perhaps enough customer interest), perhaps we just multi-target with the current LTS then (`net6.0` is the current LTS now) and call it good. Most likely, anyone trying to trim is likely on a recent TFM anyway because of all the works that's gone in to make trimming worth the effort, like support for WASM Blazor.

PR #33619, opened 2023-01-20
Update Key Vault to better support multi-targeting

heaths:
Related to #33488. Though we decided not to take advantage of multi-targeting in KV, there may be some in the near future for all track 2 assemblies. This PR should prepare KV better, and updates some comments to better explain things w.r.t. multi-targeting.

### DIFF
--- a/sdk/keyvault/Azure.Security.KeyVault.Certificates/src/PemReader.cs
+++ b/sdk/keyvault/Azure.Security.KeyVault.Certificates/src/PemReader.cs
@@ -29,6 +29,7 @@ namespace Azure.Core
 
             if (s_ecCopyWithPrivateKeyMethod is null)
             {
+                // CopyWithPrivateKey was added in .NET Framework 4.7.2. We must otherwise throw without it.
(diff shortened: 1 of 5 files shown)


christothes (approved):

heaths:
I had to take a different approach rather than use inclusive - and more obvious, IMO - preproc conditions. This is because the Analyze job of our build passes `-p:BuildSnippets=true` to _eng/service.proj_, which in turn defines the `SNIPPET` symbol for all TFMs including `net461` and `net47`. Since `netstandard2.0` - nor those specific TFMs - define `X509StorageFlags.EphemeralKeySet`, the build was failing. So because `SNIPPET` is always passed in those cases, I need to specifically exclude them. I added comments in appropriate places to that effect.

pallavit:
> I had to take a different approach rather than use inclusive - and more obvious, IMO - preproc conditions. This is because the Analyze job of our build passes `-p:BuildSnippets=true` to _eng/service.proj_, which in turn defines the `SNIPPET` symbol for all TFMs including `net461` and `net47`. Since `netstandard2.0` - nor those specific TFMs - define `X509StorageFlags.EphemeralKeySet`, the build was failing. So because `SNIPPET` is always passed in those cases, I need to specifically exclude them. I added comments in appropriate places to that effect.

> This likely will make multi-targeting complex for other SDKs as well. Is there a cleaner approach we can take? Perhaps a change to the snippet tool or how it is invoked?

PR #33621, opened 2023-01-20
[Storage] [DataMovement] Adding fix for awaiting completion on a DataTransfer to prevent delay

amnguye:


sdk/storage/Azure.Storage.DataMovement/src/Shared/DataTransfer.cs
```diff
@@ -77,37 +77,21 @@ public async ValueTask DisposeAsync()
...
         /// <param name="cancellationToken"></param>
         public Task AwaitCompletion(CancellationToken cancellationToken = default)
         {
+            SpinWait spinWait = new SpinWait();
             while (!HasCompleted)
             {
-#if DEBUG
-                VerifyTaskCompleted(HasCompleted);
-#endif
+                spinWait.SpinOnce();
```
amnguye, line 94: @schaabs Definitely looking for something better if you're open to a discussion about.

We cant await a task since multiple tasks can be used to complete one DataTransfer.

sdk/storage/Azure.Storage.DataMovement/src/Shared/DataTransfer.cs
```diff
@@ -62,56 +63,29 @@ internal DataTransfer(string id, long bytesTransferred)
...
+            if ( _state.HasCompleted ||
+                (cancellationToken != default && cancellationToken.IsCancellationRequested))
             {
-#if DEBUG
-                VerifyTaskCompleted(HasCompleted);
-#endif
-                CancellationHelper.ThrowIfCancellationRequested(cancellationToken);
+                return;
             }
-
```
schaabs, line 113: You shouldn't need to special case the task being completed or cancelled. You should be able to always await the completion sources task as below.

sdk/storage/Azure.Storage.DataMovement/src/Shared/DataTransferState.cs
```diff
@@ -111,17 +104,22 @@ public void SetId(string id)
...
+            lock (statusLock)
             {
-                await _statusSemaphore.WaitAsync().ConfigureAwait(false);
                 if (_status != status)
                 {
+                    if (StorageTransferStatus.Completed == status ||
+                        StorageTransferStatus.CompletedWithSkippedTransfers == status ||
+                        StorageTransferStatus.CompletedWithFailedTransfers == status)
+                    {
+                        _completionSource.SetResult(status);
```
schaabs, line 117: Should this do a TrySetResult instead? I think this will throw if the task has been cancelled or already completed, which may or may not be the behavior you want here.

schaabs (approved):
I left a couple of comments which should be considered before merging, changes might or might not be required.

PR #33622, opened 2023-01-20
Add comment above RequiredTargetFrameworks to update install-dotnet step

heaths:


eng/pipelines/templates/steps/install-dotnet.yml
```diff
@@ -17,6 +17,15 @@ steps:
     inputs:
       useGlobalJson: true
       performMultiLevelLookup: true
+  - task: UseDotNet@2
```
JoshLove-msft, line 20: Do we still need 6.0 SDK?

jsquire: I think we need it so that we can run against the `net6.0` target.   I guess we could just install the runtime, but either way...

heaths: Correct. We still target net6.0 for tests because it's the current/only LTS release. net8.0 will be an LTS (any even number) and overlap a year, then maybe we can upgrade again.

weshaggard: Isn't .NET 7 already on the agent? If so, do we need to explicitly install it?

I know in the past we have tried to avoid installing SDK or runtimes that currently are on the agent to avoid wasting time as well as reducing reliability issues. So, my suggestion would be to only install the minimal set of tools. 

Honestly it would be nice to have some sort of "Ensure tool is installed" as opposed to what these tasks currently do which is to always install.

eng/pipelines/templates/steps/install-dotnet.yml
```diff
@@ -17,6 +17,15 @@ steps:
...
+    displayName: 'Use .NET 7.0 SDK'
+    retryCountOnTaskFailure: 3
+    inputs:
+      # AspNetCore runtime pack can't be installed outside of SDK and we need it for integration tests
+      packageType: sdk
+      performMultiLevelLookup: true
+      version: "7.0.x"
   - task: UseDotNet@2
     condition: and(succeeded(), ne(variables['Agent.OS'], 'Windows_NT')) # Windows supports MultiLevelLookup and doesn't need explicit framework installation
     displayName: 'Use .NET 6.0 SDK'
```
weshaggard, line 31: I assume you can remove 3.1 now as well based on our teams conversation.

eng/pipelines/templates/steps/install-dotnet.yml
heaths: net6.0 was "installing" even though it's there. This still acts like "ensure it's there". It's effectively a no-op if it's already installed. Seems to me any track 1 / track 2 targets should be in alignment with install-dotnet.yml to improve reliability. I don't see how this hurts reliability.

eng/pipelines/templates/steps/install-dotnet.yml
heaths: Not sure which conversation you're talking about, but in one thread I mentioned that track 1 still needs netcoreapp3.1.

eng/pipelines/templates/steps/install-dotnet.yml
weshaggard: It isn't a noop unfortunately as it always installs given that the task doesn't install it to the same place that the agent image does. So the reliability issue comes from the fact that we sometimes hit networking issues downloading and installing the sdk/runtime.

eng/pipelines/templates/steps/install-dotnet.yml
weshaggard: Ok maybe I misread that thread. If track 1 still needs this, then we need to file and issue to fix that still as .NET Core 3.1 is no longer supported.

eng/pipelines/templates/steps/install-dotnet.yml
JoshLove-msft: Yeah, if NET 7 wasn't already installed on the agents, wouldn't the .NET pipelines be broken right now?

heaths: Right, but my assumption was that these tasks no-op if .NET _whatever_ was already installed. I was mistaken, per @weshaggard's comment in Teams.

heaths: Actually, the `multiLevelLookup` comment may be key here. I removed net6.0 and net7.0 so we'll see what happens in CI, but I suspect we may need these still.

eng/pipelines/templates/steps/install-dotnet.yml
heaths: I opened https://github.com/Azure/azure-sdk-for-net/issues/33644 to track.

eng/pipelines/templates/steps/install-dotnet.yml
heaths: See https://learn.microsoft.com/azure/devops/pipelines/tasks/reference/use-dotnet-v2?view=azure-pipelines#:~:text=performMultiLevelLookup%20%2D-,Perform%20Multi%20Level%20Lookup,-boolean.%20Default%20value for context.

heaths: Sure enough, on Ubuntu 22.04:

```
You must install or update .NET to run this application.
App: /mnt/vss/_work/1/.nuget/packages/azure.sdk.tools.testproxy/1.0.0-dev.20221111.1/tools/net6.0/any/Azure.Sdk.Tools.TestProxy.dll
Architecture: x64
Framework: 'Microsoft.NETCore.App', version '6.0.0' (x64)
.NET location: /opt/hostedtoolcache/dotnet/
The following frameworks were found:
  3.1.32 at [/opt/hostedtoolcache/dotnet/shared/Microsoft.NETCore.App]
  7.0.1 at [/opt/hostedtoolcache/dotnet/shared/Microsoft.NETCore.App]
Learn about framework resolution:
https://aka.ms/dotnet/app-launch-failed
To install missing framework, download:
https://aka.ms/dotnet-core-applaunch?framework=Microsoft.NETCore.App&framework_version=6.0.0&arch=x64&rid=ubuntu.20.04-x64
```

Note that these steps were configured to only run on non-Windows. Only Windows, it reads, supports multi-level lookup out of the box.

weshaggard: LOL I have some history of multiLevelLookup, see https://github.com/microsoft/azure-pipelines-tasks/issues/9608. So you might be right we do need to set that or the env variable for any commands we want to fallback to global if they don't exist. 

As for installing .NET 6 if it isn't on the agent which appears to be the case on Ubuntu 22 we will need this install step still.

heaths: I will avoid installing net7.0 for right now, at least. Looks like that completed fine.

eng/pipelines/templates/steps/install-dotnet.yml
```diff
@@ -25,9 +25,9 @@ steps:
       # AspNetCore runtime pack can't be installed outside of SDK and we need it for integration tests
       packageType: sdk
       performMultiLevelLookup: true
-      version: "6.0.x"
-  - task: UseDotNet@2
+      version: "6.0.x"  - task: UseDotNet@2
```
weshaggard, line 28: Looks like you are missing a newline.

weshaggard (approved):

danishgh (approved):

PR #33625, opened 2023-01-21
[Analyze] Upgrade from Node 16 to Node 18

mikeharder:
- Latest autorest requires Node 17 or higher

Verification build: https://dev.azure.com/azure-sdk/public/_build/results?buildId=2131987&view=results


weshaggard (approved):
I'm OK moving forward to node 18 as long as the generation works as expected which seems like it does from CI jobs.

weshaggard:
We should link to issue https://github.com/Azure/azure-sdk-tools/issues/5183 to outline why we needed to bump the version of node.

PR #33628, opened 2023-01-21
[LOADTESTSERVICE] DP GA SDK

NiveditJain:
DP GA SDK for Azure.Developer.LoadTesting


sdk/loadtestservice/Azure.Developer.LoadTesting/src/LoadTestRunClient.cs
```diff
@@ -90,5 +90,67 @@ public virtual async Task<TestRunOperation> BeginTestRunAsync(WaitUntil waitUnti
...
+        {
+            Argument.AssertNotNullOrEmpty(testRunId, nameof(testRunId));
+            Argument.AssertNotNull(metricname, nameof(metricname));
+            Argument.AssertNotNull(metricNamespace, nameof(metricNamespace));
+            Argument.AssertNotNull(timespan, nameof(timespan));
+
+            if (content == null)
+            {
+                content = RequestContent.Create(new { });
+            }
```
christothes, line 149: It seems like this is the only diff between what was generated - Does the `CreateGetMetricsRequest` method not do the right thing with a null content? I can't seem to find where it is implemented.

sdk/loadtestservice/Azure.Developer.LoadTesting/tests/LoadTestRunClientTest.cs
```diff
@@ -291,33 +291,42 @@ public async Task GetServerMetricsConfig()
...
+                    metricNamespacesJson.RootElement.GetProperty("value")[0].GetProperty("name").GetString(),
+                    testRunJson.RootElement.GetProperty("startDateTime").GetString()+"/"+testRunJson.RootElement.GetProperty("endDateTime")
                 );
-            Assert.NotNull(metrics);
+
+            await foreach (var page in metricsReponsePageable.AsPages())
+            {
+                foreach (var item in page.Values)
+                {
+                    Console.WriteLine(item.ToString());
```
christothes, line 327: We should assert something about the results, rather than writing them to the console. Generally, tests should only write to the console for troubleshooting or samples.

NiveditJain: Thanks for pointing it out, missed changing debugging code.

sdk/loadtestservice/Azure.Developer.LoadTesting/src/LoadTestRunClient.cs
NiveditJain: Generated code and this implementation have the following differences. 
(1) Generated code function, did not have the default value of `RequestContent content = null`, it was required to pass the value for `content` this has been changed in the custom code and made the default value null. 
(2) As you commented here, we are checking if `content` is `null` and if so then we are passing the object. The reason being our service requires it to be blank. 

You can find `CreateGetMetricsRequest` here: https://github.com/NiveditJain/azure-sdk-for-net/blob/f106f064086cc51b1c540229ac56d205ed3c1cc9/sdk/loadtestservice/Azure.Developer.LoadTesting/src/Generated/LoadTestRunClient.cs#L914-L942

christothes: Should the swagger be adjusted to make content an option parameter instead?

Harshan01: @christothes The body is an optional parameter. But the backend of the service is implemented in .NET 6 and due to the limitation of .NET 6, `Content-Type` header needs to be passed even when no body is passed. Hence, we resorted to passing an empty JSON as body so that the header is passed always. Similar changes are implemented in Java SDK as well due to the above-mentioned reason. This should be resolved by upgrading the service to .NET 7, which would be in the near future. After that, we can revert these changes in SDKs. cc:// @NiveditJain

christothes (approved):

PR #33639, opened 2023-01-23
Classifier for Expected Errors in `BlobBaseClient.Exists()`

### LINKED ISSUE #18592: [BUG] BlobBaseClient.Exists always logs exceptions to App Insights when the blob does not exist.

jaschrep-msft:
Resolves #18592

Classifies expected HTTP failures from GET BlobProperties as not errors when calling from `BlobBaseClient.Exists()`.


amnguye (approved):

benbp:
@jaschrep-msft there was an issue with `/check-enforcer reset` that caused the above behavior. It's fixed now for future reference.

janid1967:
In which release will this be available? I guess it is not released yet, ref: https://github.com/Azure/azure-sdk-for-net/blob/Azure.Storage.Blobs_12.14.1/sdk/storage/Azure.Storage.Blobs/CHANGELOG.md

ChristopherLenz:
I have upgraded to the latest version (12.15.1). From what I can see, there shouldn't be any errors appearing in the trace anymore. Unfortunately, this is still the case. We are using methods such as `CreateIfNotExists`. Technically, it shouldn't be classified as an error anymore, right?

azure_404

jaschrep-msft:
> I have upgraded to the latest version (12.15.1). From what I can see, there shouldn't be any errors appearing in the trace anymore. Unfortunately, this is still the case. We are using methods such as `CreateIfNotExists`. Technically, it shouldn't be classified as an error anymore, right?

This change does not affect the HTTP status code from the Storage service, which doesn't expose a separate "exists" mechanism. This change prevents the Azure SDK logging system from logging the response as an error.

PR #33641, opened 2023-01-23
regen with version 20230123.3

m-nash:


sdk/alertsmanagement/Azure.ResourceManager.AlertsManagement/src/Generated/Extensions/AlertsManagementExtensions.cs
```diff
@@ -239,22 +239,14 @@ public static Pageable<AlertProcessingRuleResource> GetAlertProcessingRules(this
...
+        /// <param name="options"> A property bag which contains all the parameters of this method except the LRO qualifier and request context parameter. </param>
         /// <param name="cancellationToken"> The cancellation token to use. </param>
-        public static async Task<Response<ServiceAlertSummary>> GetServiceAlertSummaryAsync(this SubscriptionResource subscriptionResource, AlertsSummaryGroupByField groupby, bool? includeSmartGroupsCount = null, string targetResource = null, string targetResourceType = null, string targetResourceGroup = null, MonitorServiceSourceForAlert? monitorService = null, MonitorCondition? monitorCondition = null, ServiceAlertSeverity? severity = null, ServiceAlertState? alertState = null, string alertRule = null, TimeRangeFilter? timeRange = null, string customTimeRange = null, CancellationToken cancellationToken = default)
+        /// <exception cref="ArgumentNullException"> <paramref name="options"/> is null. </exception>
```
heaths, line 244: 1. Isn't this a breaking change, or had the API not GA'd yet?
2. Why extension methods? Seems very track 1-ish. If we're generating the client code as well, why not just regular methods?

jsquire (approved):

sdk/alertsmanagement/Azure.ResourceManager.AlertsManagement/src/Generated/Extensions/AlertsManagementExtensions.cs
heaths: Granted, they clearly already existed. 2 is more of a question.

sdk/billingbenefits/Azure.ResourceManager.BillingBenefits/api/Azure.ResourceManager.BillingBenefits.netstandard2.0.cs
```diff
@@ -15,8 +15,8 @@ public static partial class BillingBenefitsExtensions
...
-        public static Azure.AsyncPageable<Azure.ResourceManager.BillingBenefits.BillingBenefitsSavingsPlanResource> GetBillingBenefitsSavingsPlansAsync(this Azure.ResourceManager.Resources.TenantResource tenantResource, string filter = null, string orderBy = null, string refreshSummary = null, float? skipToken = default(float?), string selectedState = null, float? take = default(float?), System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
+        public static Azure.Pageable<Azure.ResourceManager.BillingBenefits.BillingBenefitsSavingsPlanResource> GetBillingBenefitsSavingsPlans(this Azure.ResourceManager.Resources.TenantResource tenantResource, Azure.ResourceManager.BillingBenefits.Models.TenantGetBillingBenefitsSavingsPlansOptions options, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
```
heaths, line 18: What with the "Tenant" prefix? I remember in the autorest.csharp repo you did something that stripped a "Subscription" prefix from generated methods. Typically, we use `{MethodName}Options` for the property bag, so this has an extra prefix.

m-nash: [Client][Method][Options is the convention since many clients within the same package can have a `Delete` method.

sdk/alertsmanagement/Azure.ResourceManager.AlertsManagement/src/Generated/Extensions/AlertsManagementExtensions.cs
m-nash: 1. This is the new API, the old one still exists in custom code.
2. This is how we place methods on parents that don't live in the same package.

PR #33655, opened 2023-01-24
Handle SAS credential updates during message retries

### LINKED ISSUE #31442: AzureSasCredentialSynchronousPolicy does not properly handle Refresh between retry attempts

### LINKED ISSUE #32514: Azure SAS credential synchronous policy is producing a wrong query string

christothes:
fixes https://github.com/Azure/azure-sdk-for-net/issues/32514
fixes https://github.com/Azure/azure-sdk-for-net/issues/31442


sdk/core/Azure.Core/src/Internal/AzureSasSignatureHistory.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Azure.Core
+{
+    /// <summary>
+    /// A marker class used to represent AzureSasCredential signature history on the <see cref="HttpMessage"/> property bag.
+    /// </summary>
+    internal class AzureSasSignatureHistory
```
annelo-msft, line 9: Would it make sense to add this in the policy the way we did in [RedirectPolicy](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/Pipeline/RedirectPolicy.cs#L281)?  Would it ever be used outside the policy?

sdk/core/Azure.Core/src/Shared/AzureSasCredentialSynchronousPolicy.cs
```diff
@@ -29,13 +29,26 @@ public override void OnSendingRequest(HttpMessage message)
             {
                 signature = signature.Substring(1);
             }
+
             if (!query.Contains(signature))
             {
-                query = string.IsNullOrEmpty(query) ? '?' + signature : query + '&' + signature;
+                // Get the signature history so we know which string to replace in the query if it was previously set.
+                if (message.TryGetProperty(typeof(AzureSasSignatureHistory), out object? previousSignature) && previousSignature != null & query.Contains((string)previousSignature!))
```
annelo-msft, line 36: nit: this is a lot of logical statements all together, would it be possible move into multiple statements for readability?

annelo-msft (commented):
Would you be willing to update the description to say why the implementation fixes the linked issues?

sdk/core/Azure.Core/src/Internal/AzureSasSignatureHistory.cs
christothes: Good idea, that makes the CI changes easier for clients that rely on the new type.

sdk/core/Azure.Core/src/Shared/AzureSasCredentialSynchronousPolicy.cs
christothes: They need to be one statement, but I'll break them into different lines.

sdk/core/Azure.Core/src/Shared/AzureSasCredentialSynchronousPolicy.cs
```diff
@@ -29,13 +29,33 @@ public override void OnSendingRequest(HttpMessage message)
...
 
-            base.OnSendingRequest(message);
+                base.OnSendingRequest(message);
+            }
         }
+
+        /// <summary>
+        /// A marker class used to represent AzureSasCredential signature history on the <see cref="HttpMessage"/> property bag.
+        /// </summary>
+        internal class AzureSasSignatureHistory { }
```
annelo-msft, line 59: Does it need to be internal?  One of the goals of this design, I think, was that only the policy can get and set these values.  If the type is private, it is easier to achieve that.

sdk/core/Azure.Core/src/Shared/AzureSasCredentialSynchronousPolicy.cs
```diff
@@ -29,13 +29,33 @@ public override void OnSendingRequest(HttpMessage message)
             {
                 signature = signature.Substring(1);
             }
+
             if (!query.Contains(signature))
             {
-                query = string.IsNullOrEmpty(query) ? '?' + signature : query + '&' + signature;
+                // Get the signature history so we know which string to replace in the query if it was previously set.
+                if (message.TryGetProperty(typeof(AzureSasSignatureHistory), out object? previousSignature) &&
+                    previousSignature != null &&
```
annelo-msft, line 37: Do we need to check previous signature for null?  If only the policy can get/set the value, then it could do validation on setting that it isn't null.  (Unless null is valid!)

sdk/core/Azure.Core/src/Shared/AzureSasCredentialSynchronousPolicy.cs
```diff
@@ -29,13 +29,33 @@ public override void OnSendingRequest(HttpMessage message)
...
+                    query.Contains((string)previousSignature!))
+                {
+                    query = query.Replace((string)previousSignature!, signature);
+                }
+                else
```
annelo-msft, line 42: It always makes me nervous when there is an else case for an if-statement with a lot of clauses on it.  For example, if there was a previous signature stored on the message, but it was null, is this really the first time we're setting the signature?  Is it possible there would be a previous signature on the message but it isn't in the query?  Why would that happen?  Does this else case address all of the different reasons we could reach it?

sdk/core/Azure.Core/src/Shared/AzureSasCredentialSynchronousPolicy.cs
christothes: It's mostly for nullable semantics. The output from TryGetProperty is a `object?` and then we need to cast it to string. The compiler gets cranky if we don't ensure that the value isn't null before casting it.

JoshLove-msft: You should be able to just use "is string"

sdk/core/Azure.Core/src/Shared/AzureSasCredentialSynchronousPolicy.cs
christothes: > if there was a previous signature stored on the message, but it was null, is this really the first time we're setting the signature?

If it was null, it doesn't matter. We effectively don't have a signature or a valid previous signature and we'll need to add one.

> Is it possible there would be a previous signature on the message but it isn't in the query? Why would that happen?

Yes, this is the scenario that describes the bug being fixed. 

1. The message is sent to the pipeline
2. `AzureSasCredentialSynchronousPolicy` adds the signature the query and to the previous signature property
3. The message is sent to the transport
4. The SAS credential is updated
5. The transport returns the response but it is a retriable failure
6. The message traverses the pipeline policies and `AzureSasCredentialSynchronousPolicy` evaluates the signature but it no longer matches, but the query does match the previous signature.
7. `AzureSasCredentialSynchronousPolicy` knows that the stale signature needs to be replaced rather than adding an additional signature with the new SAS value.

> Does this else case address all of the different reasons we could reach it?

I think so, but if you can think of an edge case not handled, please do!

annelo-msft: I see, ok, thank you for that explanation, it is incredibly helpful!

If that is the case (and since now that `AzureSasSignatureHistory`, this value can only be set in the policy), then do you even need to check for null or `Contains`?  You know the value isn't null since you're the only one setting it.  And I believe `string.Replace()` semantics are such that if the string isn't contained, it won't modify the value, so you would avoid doing two linear string checks which would be better for perf.

Would it make sense to take those out then and just do `previousSignature! as string`?  (I haven't tried it, maybe it doesn't work)

christothes: We still need the `Contains` because if it is true, we need to do the replace. But if it is false, we need to add the query in the else block.

annelo-msft: That makes sense.  Curious: how would you end up with a value for the previous signature without it being appended to the query?

christothes: It shouldn't happen normally, but if some other policy were to mess with the query, this could happen.

sdk/core/Azure.Core/src/Shared/AzureSasCredentialSynchronousPolicy.cs
```diff
@@ -29,13 +31,33 @@ public override void OnSendingRequest(HttpMessage message)
...
+                if (message.TryGetProperty(typeof(AzureSasSignatureHistory), out object? previousSignature) &&
+                    previousSignature is string previousSignatureString &&
+                    query.Contains(previousSignatureString))
+                {
+                    query = query.Replace(previousSignatureString, signature);
+                }
+                else
+                {
+                    // This is the first time we're setting the signature, so we need to add it to the query.
+                    query = string.IsNullOrEmpty(query) ? '?' + signature : query + '&' + signature;
```
JoshLove-msft, line 47: nit: string interpolation

sdk/core/Azure.Core/src/Shared/AzureSasCredentialSynchronousPolicy.cs
```diff
@@ -29,13 +31,33 @@ public override void OnSendingRequest(HttpMessage message)
...
+
+                    // Store the signature in the message property bag so we can replace it if it changes.
+                    message.SetProperty(typeof(AzureSasSignatureHistory), signature);
```
JoshLove-msft, line 50: Don't we need to store the new signature in the "if" case as well?

sdk/core/Azure.Core/tests/AzureSasCredentialSynchronousPolicyTests.cs
```diff
@@ -132,5 +131,46 @@ public async Task VerifyRetryAfterSasCredentialUpdateOperation()
...
+            const string SECOND_SIGNATURE_VALUE = "?sig=second_signature_value";
+            const string THIRD_SIGNATURE_VALUE = "?sig=third_signature_value";
+            int callCount = 0;
+            var azureSasCredential = new AzureSasCredential(FIRST_SIGNATURE_VALUE);
+            var transport = new MockTransport((req) =>
+            {
+                if (callCount++ == 0)
+                {
+                    Assert.AreEqual("?foo=bar&sig=first_signature_value", req.Uri.Query);
+                    Thread.Sleep(100);
```
JoshLove-msft, line 150: I'd suggest moving this into the RetryPolicyTests and using the pattern there with the AsyncGates to have more control over when the retry occurs.

sdk/core/Azure.Core/src/Shared/AzureSasCredentialSynchronousPolicy.cs
christothes: With simple concatenations, I think interpolation is actually slightly slower. Or at least the difference is negligable.

JoshLove-msft: It shouldn't be slower - the compiler will optimize as needed.
https://devblogs.microsoft.com/dotnet/string-interpolation-in-c-10-and-net-6/

JoshLove-msft (approved):

sdk/core/Azure.Core/tests/AzureSasCredentialSynchronousPolicyTests.cs
christothes: Fixed the existing tests instead.

Issue #33646, opened 2023-01-24
Investigate digest comparison logic in DownloadManifest


annelo-msft:
In the DownloadManifest method, if the manifest is requested via `tag` rather than `digest`, there is no pre-request digest value to use to validate the digest returned from the client.  We instead [compare the digest in the response header to the digest computed by the payload](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs#L502).

We should understand whether these are the intended values to compare in this scenario.

FYI: @pallavit, @lmolkova

annelo-msft:
At a minimum, we could improve this case by comparing to the requested digest in the cases where the `reference` is the digest.

annelo-msft:
@timovv investigated and based on his findings, we decided the following:

>  I did some investigation into how the other clients we know of were handling the digest verification in downloadManifest. Docker CLI appears to be only verifying the digest when an identifier (digest minus the `sha256:`) is provided, and not verifying when it's just a tag, since it can't. ORAS and the Azure CLI don't seem to be doing any validation at all.

We decided we will validate the manifest digest on download similar to what the Docker CLI does.  This is the [recommendation in the REST API spec](https://docs.docker.com/registry/spec/api/#digest-header) and provides a higher level of security for SDK customers.  If the user passes a tag and not a digest (determined by checking if `sha256:` is present), we will still validate the digest from the response header against the content to help detect data corruption.

PR #33663, opened 2023-01-25
Support Non-OCI Manifests

### LINKED ISSUE #23828: [ACR] [BlobClient] Add support for non-OCI Manifests

### LINKED ISSUE #32351: [BUG] Deserialization throws exception
### Library name and version

Azure.Containers.ContainerRegistry

### Describe the bug

We have custom code which marks [this base class as abstract](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ArtifactManifest.cs#L12) because its supposed to be a base type for a discriminator tree, however none of the properties on the type are marked as the discriminator which means that the internal `Unknown` type will not be generated which means that our deserialization has no way to construct the object.

The [deserialize throw statement is here](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Generated/Models/ArtifactManifest.Serialization.cs#L29).

### Expected behavior

No throw

### Actual behavior

throws

### Reproduction Steps

Deserialize the ArtifactManifest type.

### Environment

_No response_

annelo-msft:
This PR makes modifications to the Manifest APIs to accomplish the following:
- Caller can specify the manifest media type in the UploadManifest() operation.
- Caller can request a specific media type for a manifest in the DownloadManifest() operation.
- Let caller lazily deserialize the manifest based on the known media type
- Remove base ArtifactManifest type (which didn't contribute much) to simplify types in the API

OCI formats [are compatible with](https://github.com/opencontainers/image-spec/blob/main/media-types.md#compatibility-matrix) Docker formats, so these types can be used with corresponding Docker manifests.

Fixes https://github.com/Azure/azure-sdk-for-net/issues/23828
Fixes https://github.com/Azure/azure-sdk-for-net/issues/32351


sdk/containerregistry/Azure.Containers.ContainerRegistry/api/Azure.Containers.ContainerRegistry.netstandard2.0.cs
```diff
@@ -163,7 +163,7 @@ public static partial class ContainerRegistryModelFactory
...
         public static Azure.Containers.ContainerRegistry.ContainerRepositoryProperties ContainerRepositoryProperties(string registryLoginServer = null, string name = null, System.DateTimeOffset createdOn = default(System.DateTimeOffset), System.DateTimeOffset lastUpdatedOn = default(System.DateTimeOffset), int manifestCount = 0, int tagCount = 0, bool? canDelete = default(bool?), bool? canWrite = default(bool?), bool? canList = default(bool?), bool? canRead = default(bool?)) { throw null; }
         public static Azure.Containers.ContainerRegistry.Specialized.DownloadBlobResult DownloadBlobResult(string digest = null, System.BinaryData content = null) { throw null; }
-        public static Azure.Containers.ContainerRegistry.Specialized.DownloadManifestResult DownloadManifestResult(string digest = null, Azure.Containers.ContainerRegistry.Specialized.OciManifest manifest = null, System.BinaryData content = null) { throw null; }
+        public static Azure.Containers.ContainerRegistry.Specialized.DownloadManifestResult DownloadManifestResult(string digest = null, string mediaType = null, System.BinaryData content = null) { throw null; }
```
lmolkova, line 166: internal?

sdk/containerregistry/Azure.Containers.ContainerRegistry/api/Azure.Containers.ContainerRegistry.netstandard2.0.cs
```diff
@@ -266,14 +261,36 @@ public partial class DownloadManifestOptions
     {
         public DownloadManifestOptions(string tag = null, string digest = null) { }
```
lmolkova, line 262: I wonder why we had this type. Service API says that it takes tagOrDigest, but we require user to allocate another object for options just to later on pass any of them.

I'd like to entertain the following idea:
1. Have simple and advanced download overloads
   - `DownloadManifest(string tagOrDigest)`
   - `DownloadManifest(string tagOrDigest, DownloadManifestOptions options)`
2. `DownloadManifestOptions` to only have `MediaType` for the time being

lmolkova: Maybe even hide serialization for custom types?
```csharp
DownloadManifestResult<MyManifestListImpl> result = client.downloadManifest(tagOrDigest, new DownloadManifestOptions<MyManifestListImpl>);
MyManifestListImpl manifestList = result.Value; // uses BinaryData.ToObject<T> to serialize
```

sdk/containerregistry/Azure.Containers.ContainerRegistry/api/Azure.Containers.ContainerRegistry.netstandard2.0.cs
```diff
@@ -266,14 +261,36 @@ public partial class DownloadManifestOptions
...
-        public Azure.Containers.ContainerRegistry.Specialized.ArtifactManifest Manifest { get { throw null; } }
+        public Azure.Containers.ContainerRegistry.Specialized.ManifestMediaType MediaType { get { throw null; } }
+        public Azure.Containers.ContainerRegistry.Specialized.OciManifest AsOciManifest() { throw null; }
+    }
+    [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
+    public readonly partial struct ManifestMediaType : System.IEquatable<Azure.Containers.ContainerRegistry.Specialized.ManifestMediaType>
+    {
+        private readonly object _dummy;
+        private readonly int _dummyPrimitive;
+        public static readonly Azure.Containers.ContainerRegistry.Specialized.ManifestMediaType DockerManifestList;
```
lmolkova, line 280: should we define it if we're not going to support it?

lmolkova (commented):
Left a few ideas for API improvements to consider.

sdk/containerregistry/Azure.Containers.ContainerRegistry/api/Azure.Containers.ContainerRegistry.netstandard2.0.cs
annelo-msft: In .NET our model factories are public, so we need to keep this here.

sdk/containerregistry/Azure.Containers.ContainerRegistry/api/Azure.Containers.ContainerRegistry.netstandard2.0.cs
annelo-msft: Yeah, great question :)

Trying to remember back to when we added this type, I think the idea was that we wanted to strongly-type whether we had a tag or a digest, so we could use them differently in the method call.  As well as give us a place to add options if we decided we needed to grow the input parameter list without adding overloads.

Since we're not using them differently, I agree it would make sense to remove the type.  However, do we want to resolve https://github.com/Azure/azure-sdk-for-net/issues/33646 before we decide?

annelo-msft: > Maybe even hide serialization for custom types?

Offhand, I'm not sure I see the value of adding a convenience overload to support this.  If the caller has a custom type, they're already a somewhat advanced user, and it's not so bad to add an extra line to deserialize to their type?  But feel free to provide more reasons if it feels important ... ?

sdk/containerregistry/Azure.Containers.ContainerRegistry/api/Azure.Containers.ContainerRegistry.netstandard2.0.cs
annelo-msft: Yeah, I went back and forth on this when I wrote it -- glad you thought to comment on it.

If we plan to add support for types later, I like having the extensible enum for convenience.  But it also feels weird to ship an enum with only one value in it, so I added the other known values to keep it company.

Since we could have both DockerV2 and OciManifest in the enum, though, and leave lists out until we properly support them, maybe we could move forward with that instead.  I agree with your point that it's strange to provide the option for something we don't currently support.

annelo-msft: Addressed.

sdk/containerregistry/Azure.Containers.ContainerRegistry/api/Azure.Containers.ContainerRegistry.netstandard2.0.cs
annelo-msft: I added the suggestion so we can look at the API.

lmolkova: the reason to have 

`MyManifestListImpl manifestList = result.Value;` is to have fewer APIs. Otherwise we'd have `AsOciManifest()`, `AsOciIndext()`, potentially `AsDockerManifest` and so on.

But thinking about it more, it seems users don't know what manifest they will get, so it's impossible to write the first line anyway:
`DownloadManifestResult<MyManifestListImpl> result = client.downloadManifest<MyManifestListImpl>`

Sorry for misleading you :)

annelo-msft: No worries, thanks @lmolkova!

ShivangiReja (approved):

ShivangiReja:
LGTM

PR #33665, opened 2023-01-25
Update snippet-generator

### LINKED ISSUE #1991: SnippetGenerator still outputs red, with a new red line that causes some confusion
My PR #1954 has not taken effect, so the success message that SnippetGenerator finished is still in red which causes confusion. Did you take the right version of the package that includes it?

In addition, however you are installing the package also produces a non-fatal error that causes confusion right at the beginning:

image

I'm not sure what is causing this (didn't investigate further), but it does not seem to be `dotnet tool install` (since "Tool" was mentioned) since re-installing a version that was already installed doesn't report an error.

Ideally, no messages looking like errors should be emitted. It causes confusion and, hopefully, causes people to investigate why the tool seemed to fail (which it didn't).

heaths:
Pulls in the fix for Azure/azure-sdk-tools#5207 and fixes Azure/azure-sdk-tools#1991 while making sure snippet-generator is always updated. It wasn't before because `dotnet tool install` will not upgrade an installed tool and the .NET SDK team decided to add `dotnet tool update` in favor of adding a `--force` to `dotnet tool install`, thus making installation and updating more complicated.

Instead, `dotnet tool restore` does both and is fast if the tool is already installed. Additionally, running most (or any?) `dotnet` commands restores the tool automatically.

### DIFF
--- a/eng/pipelines/templates/jobs/ci.yml
+++ b/eng/pipelines/templates/jobs/ci.yml
@@ -182,7 +182,7 @@ jobs:
         inputs:
           command: custom
           custom: 'tool'
-          arguments: 'install --global --add-source "https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-for-net/nuget/v3/index.json" --version "1.0.0-dev.20211119.1" "Azure.Sdk.Tools.SnippetGenerator"'
+          arguments: 'restore'
           workingDirectory: '$(Agent.BuildDirectory)'
 
       - task: PowerShell@2
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -2,6 +2,12 @@
   "version": 1,
   "isRoot": true,
   "tools": {
+    "Azure.Sdk.Tools.SnippetGenerator": {
+      "version": "1.0.0-dev.20230124.1",
+      "commands": [
+        "snippet-generator"
+      ]
+    },
     "dotnet-reportgenerator-globaltool": {
       "version": "4.8.0",
       "commands": [
--- a/eng/scripts/Update-Snippets.ps1
+++ b/eng/scripts/Update-Snippets.ps1
@@ -5,43 +5,23 @@ param (
     [string] $ServiceDirectory,
 
     [Parameter()]
-    [switch] $StrictMode,
-
-    [Parameter()]
-    [string] $SnippetToolPath=''
+    [switch] $StrictMode = !(Test-Path Env:TF_BUILD)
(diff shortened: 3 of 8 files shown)


heaths:
@chunyu3 I had to update the template 1) to make it valid, since it needed test methods, and 2) had some content so this new snippet validation wouldn't fail. But I installed and tested the template - including running tests and update snippets - and it seemed to work just fine. The output tests from the template wouldn't be empty, but if that's a problem we can also solve that with the trick I had before using `#if 0` or something like that to have the template engine ignore anything between. Personally, I like the idea of it having some content so it's obvious what to do next, but perhaps that conflicts with the scripts you have that make use of the _azsdkdpg_ template. If you would, please test this change and we can fix anything as needed.

christothes (approved):

eng/pipelines/templates/jobs/ci.yml
```diff
@@ -182,7 +182,7 @@ jobs:
         inputs:
           command: custom
           custom: 'tool'
-          arguments: 'install --global --add-source "https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-for-net/nuget/v3/index.json" --version "1.0.0-dev.20211119.1" "Azure.Sdk.Tools.SnippetGenerator"'
+          arguments: 'restore'
```
annelo-msft, line 185: Will this work on a system where it isn't previously installed?

sdk/keyvault/Azure.Security.KeyVault.Certificates/src/PemReader.cs
```diff
@@ -29,6 +29,7 @@ internal static partial class PemReader
 
             if (s_ecCopyWithPrivateKeyMethod is null)
             {
+                // CopyWithPrivateKey was added in .NET Framework 4.7.2. We must otherwise throw without it.
```
annelo-msft, line 32: Are the keyvault changes needed as part of this change?

eng/pipelines/templates/jobs/ci.yml
heaths: Yes. It both installs and upgrades, unlike `dotnet tool install` vs. `dotnet tool update`.

sdk/keyvault/Azure.Security.KeyVault.Certificates/src/PemReader.cs
heaths: Oops. No. I'll rebase on main since they went in yesterday.

PR #33675, opened 2023-01-25
Add new TargetUnprocessedEventThreshold parameter in EventHub Options + Target Based Scaling for Event Hubs

chiangvincent:
## Overview 
Azure Functions scales event hubs based on the number of unprocessed events distributed across all the partitions in the hub. By default it takes the configured maximum batch size and assumes that each allocated instance will process that many events, and allocates sufficient instances to process all unprocessed events, taking into account that each partition can only be processed by one instance.  You can adjust this default behavior by specifying the `TargetUnprocessedEventThreshold` - the total unprocessed event count will be divided by this value to determine the number of instances, and then we will round that up to an instance count that creates a balanced partition distribution.

This PR introduces the new attribute that users can set to fine-tune the default target based scaling behavior, and it also adds support for target based scaling for Event Hub-triggered functions.

## Checklist for Target Based Scaling Changes
- [x] create EventHubsMetricsProvider
- [x] remove duplicate logic between EventHubsSaleMonitor, EventHubsMetricsProvider
- [x] EventHubsListener implementITargetScalerProvider
- [x] fix unit tests
- [x] add EventHubsargetScaler
- [x] add unit tests for EventHubsTargetScaler
- [x] E2E tests with Event Hub trigger with locally running function app
- [ ] test with scale controller prototype (dependent on finalizing the scalerfactory implementation)

## Logs from E2E Verification With Local Function App
Steps: 
1. Generate a local nuget package with EventHub changes from this PR
2. Create function app referencing local EventHub nuget packages
3. Publish and run the application locally with `azure-functions-host` repo 
4. Enable runtime driven scaling, target based scaling, and add `ScmHostingConfigurations.txt` to enable EventHubs TBS
5. Send 200k messages to event hub with 5 partitions, in batches of 300 messages
6. Ping `admin/host/scale/status` API using Postman, observe target worker count and local `azure-functions-host` logs emitted

_Target Based Scaling Enabled_
image

_Target Based Scaling Disabled_
image

image

image

_target scaler using the targetUnprocessedEventThreshold as the concurrency value in target based scaling_
image
image

image

## Resources
- [Reference Target Based Scaling PR (for Service Bus)](https://github.com/Azure/azure-sdk-for-net/pull/31367/files#diff-fa639f441f0f5a55d5a76231d750b00594a4197a9f20fa58bc18825cd8109de2)


sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/EventHubOptions.cs
```diff
@@ -132,6 +132,27 @@ public int MaxEventBatchSize
...
         }
 
+        private int _targetUnprocessedEventThreshold;
+
+        /// <summary>
+        /// Get or sets the target unprocessed events per worker for eventhub functions. This is used in target based scaling to override the default behavior of using maxEventBatchSize as the target value.
+        /// If TargetUnprocessedEventThreshold is set, the total unprocessed event count will be divided by this value to determine the number of instances,
+        /// and then we will round that up to an instance count that creates a balanced partition distribution.
+        /// </summary>
+        public int TargetUnprocessedEventThreshold
```
jsquire, line 142: @JoshLove-msft :  This would also require updates in the [EventHubExtensionConfigProvider](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/EventHubExtensionConfigProvider.cs), right?     

@chiangvincent:  Please also include this in bindings tests where we exercise the other config.

JoshLove-msft: I don't believe so. The binding should work automatically based on this [line](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/EventHubWebJobsBuilderExtensions.cs#L83). We have special logic for some properties that we wanted to have backwards compat support for in EventHubWebJobsBuilderExtensions. Also properties that are not able to be bound automatically by the extensions framework, e.g. the WebProxy require special handling, but this one shouldn't need anything special. We should definitely add tests to confirm though.

JoshLove-msft: Having said that, where is this property actually used? I don't see any code consuming it.

pragnagopa:
Tagging @alrod to help with review

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/EventHubOptions.cs
jsquire: > Having said that, where is this property actually used? I don't see any code consuming it.

It'll be picked up by the service-side scale controller and not used within the bindings directly.   That said, I have no idea how that actually happens - @chiangvincent?

chiangvincent: Created a teams chat with scale controller PR -- will also create a corresponding PR in the sdk-for-net repo

chiangvincent: Added unit tests for configuration checks

JoshLove-msft: Tests are failing due to the validation that ensures the value is greater than 0. Is there a default value that should apply here other than 0? Otherwise, should this property be nullable?

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/EventHubOptions.cs
```diff
@@ -132,6 +132,27 @@ public int MaxEventBatchSize
             }
         }
 
+        private int _targetUnprocessedEventThreshold;
+
+        /// <summary>
+        /// Get or sets the target unprocessed events per worker for eventhub functions. This is used in target based scaling to override the default behavior of using maxEventBatchSize as the target value.
```
pragnagopa, line 138: nit: `eventhub triggered functions`

jsquire: Event Hubs is the product name; Event Hub is a single resource.    For branding purposes, we shouldn't be using "eventhub".   I believe we'd want to use "Event Hub-triggered functions." in this context.

jsquire, line 138:
```suggestion
        /// Get or sets the target number of unprocessed events per worker for Event Hub-triggered functions. This is used in target-based scaling to override the default scaling threshold inferred from the <see cref="MaxEventBatchSize" /> option.
```

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/EventHubOptions.cs
```diff
@@ -132,6 +132,27 @@ public int MaxEventBatchSize
...
+        /// If TargetUnprocessedEventThreshold is set, the total unprocessed event count will be divided by this value to determine the number of instances,
```
jsquire, line 139:
```suggestion
        /// If TargetUnprocessedEventThreshold is set, the total unprocessed event count will be divided by this value to determine the number of worker instances,
```

nit:  You refer to "workers" above.  i think we need to be consistent here.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/EventHubOptions.cs
```diff
@@ -132,6 +132,27 @@ public int MaxEventBatchSize
...
+        /// and then we will round that up to an instance count that creates a balanced partition distribution.
```
jsquire, line 140:
```suggestion
        /// which will then be rounded up to a worker instance count that creates a balanced partition distribution.
```

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/EventHubOptions.cs
jsquire: +1 for nullable over using `0` as a magic number to represent unset.

TsuyoshiUshio:
Hi @chiangvincent 
Somewhat I can't leave the pointer, however, you need to update here. SyncTriggerManager will  use this part. [Formatter](https://github.com/chiangvincent/azure-sdk-for-net/blob/user/vchiang/addTargetUnprocessedEventThreshold/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/EventHubOptions.cs#L210)

Could you double check "spec" part on this PR description? https://github.com/Azure/azure-functions-host/pull/8148

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/EventHubOptions.cs
chiangvincent: Updated

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/EventHubOptions.cs
chiangvincent: Updated

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/EventHubOptions.cs
chiangvincent: Updated

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/EventHubOptions.cs
chiangvincent: Done, made `TargetUnprocessedEventThreshold` nullable in the options class.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/EventHubOptions.cs
chiangvincent: Done, I added Event Hub-triggered functions to description

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Microsoft.Azure.WebJobs.Extensions.EventHubs.csproj
```diff
@@ -14,7 +14,7 @@
   <ItemGroup>
     <PackageReference Include="Azure.Messaging.EventHubs" />
     <PackageReference Include="Azure.Storage.Blobs" />
-    <PackageReference Include="Microsoft.Azure.WebJobs" />
+    <PackageReference Include="Microsoft.Azure.WebJobs" VersionOverride="3.0.36-11991" />
```
chiangvincent, line 17: Update this when webjobs-sdk version is released

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubsTargetScaler.cs
```diff
@@ -0,0 +1,186 @@
...
+        private readonly EventHubsMetricsProvider _metricsProvider;
+        private readonly EventHubOptions _options;
+
+        private DateTime _lastScaleUpTime;
+        private TargetScalerResult _lastTargetScalerResult;
+
+        public EventHubsTargetScaler(string functionId,
+            IEventHubConsumerClient client,
+            EventHubOptions options,
+            EventHubsMetricsProvider metricsProvider,
```
chiangvincent, line 33: Added metrics provider as parameter here for mocking

pragnagopa:
tagging @paulbatum for review as well!

NuGet.Config
```diff
@@ -10,6 +10,7 @@
       Used for azure-sdk-tools repo until issue https://github.com/Azure/azure-sdk-tools/issues/1329 is addressed
     -->
     <add key="azure-sdk-tools" value="https://azuresdkartifacts.blob.core.windows.net/azure-sdk-tools/index.json" />
+    <add key="myget" value="https://www.myget.org/F/azure-appservice/api/v3/index.json" />
```
weshaggard, line 13: This PR cannot merge with this new nuget feed included.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/EventHubOptions.cs
JoshLove-msft: There appears to be build errors now.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubsTargetScaler.cs
```diff
@@ -0,0 +1,186 @@
...
+
+        internal int GetDesiredConcurrencyInternal(TargetScalerContext context)
+        {
+            int concurrency = 0;
+
+            if (!context.InstanceConcurrency.HasValue)
+            {
+                if (_options.TargetUnprocessedEventThreshold > 0)
+                {
+                    concurrency = _options.TargetUnprocessedEventThreshold;
```
jsquire, line 109:
```suggestion
                    concurrency = (int)_options.TargetUnprocessedEventThreshold;
```

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Microsoft.Azure.WebJobs.Extensions.EventHubs.csproj
jsquire: I'm going to request changes until it does, that way we don't accidentally merge.

jsquire (changes requested):

NuGet.Config
jsquire: I blocked the merge until the necessary package has been released and this can be removed.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubsMetricsProvider.cs
```diff
@@ -0,0 +1,195 @@
...
+                return metrics;
+            }
+
+            return CreateTriggerMetrics(partitionPropertiesTasks.Select(t => t.Result).ToList(), checkpoints);
+        }
+
+        private EventHubsTriggerMetrics CreateTriggerMetrics(List<PartitionProperties> partitionRuntimeInfo, EventProcessorCheckpoint[] checkpoints, bool alwaysLog = false)
+        {
+            long totalUnprocessedEventCount = 0;
+            bool logPartitionInfo = alwaysLog ? true : DateTime.UtcNow >= _nextPartitionLogTime;
```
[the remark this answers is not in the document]
chiangvincent: Done, saved utcnow into a var

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubsTargetScaler.cs
```diff
@@ -0,0 +1,186 @@
...
+            ILogger logger)
+        {
+            _functionId = functionId;
+            _logger = logger;
+            _client = client;
+            _metricsProvider = metricsProvider;
+            _options = options;
+
+            _lastScaleUpTime = DateTime.MinValue;
+            _lastTargetScalerResult = new TargetScalerResult { };
```
jsquire, line 43:
```suggestion
            _lastTargetScalerResult = new TargetScalerResult();
```

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubsTargetScaler.cs
```diff
@@ -0,0 +1,186 @@
...
+        {
+            if (partitionCount == 0)
+            {
+                return Enumerable.Range(0, partitionCount).ToArray();
+            }
+            else
+            {
+                List<int> list = Enumerable.Range(1, partitionCount).
+                    Select(c => (double)partitionCount / (double)c).
+                    Select(d => (int)Math.Ceiling((double)d)).
```
jsquire, line 145: Why chain LINQ statements rather than just do this in the initial `Select`?

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubsTargetScaler.cs
```diff
@@ -0,0 +1,186 @@
...
+                    Distinct().
+                    OrderBy(e => e).
+                    ToList();
+                list.Insert(index: 0, item: 0);
```
jsquire, line 149: Seems odd to allocate a list just to do an insertion in the beginning and then to allocate an array after.   Maybe consider just creating the array upfront and then filling via `foreach` rather than LINQ?

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/tests/EventHubsMetricsProviderTests.cs
```diff
@@ -0,0 +1,231 @@
...
+using Azure;
+using Azure.Messaging.EventHubs;
+using Azure.Messaging.EventHubs.Primitives;
+using Microsoft.Azure.WebJobs.Extensions.EventHubs.Listeners;
+using Microsoft.Azure.WebJobs.Host.TestCommon;
+using Microsoft.Azure.WebJobs.Logging;
+using Microsoft.Extensions.Logging;
+using Moq;
+using NUnit.Framework;
+using static Moq.It;
```
jsquire, line 18: Do we need this?   It looks like you're explicitly using `It` inline (and it would real oddly without doing so...)

jsquire (changes requested):
I left a couple of minor nits/questions, but overall looks good to me.   I'm going to continue to block until we can remove the temp nuget feed.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubsTargetScaler.cs
chiangvincent: Resolving from discussion via offline conversation -- distinct can change length of array allocated in line 143.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubsTargetScaler.cs
chiangvincent: Resolving from discussion via offline conversation -- distinct can change length of array allocated in line 143.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/tests/EventHubsMetricsProviderTests.cs
chiangvincent: Looks like this is used for the original tests from EventHubScaleMonitorTests, specifically the `IsAny` method in line 50
` _consumerClientMock.Setup(client => client.GetPartitionPropertiesAsync(IsAny<string>()))`. Open to suggestions here if there's a better approach.

jsquire: I overlooked that instance.   Personally, I'd prefer to see us use that in the intended fluent form. `It.IsAny` so that it reads more easily for those not familiar with Moq.   That said, I don't feel strongly enough about it that I'd push back.

chiangvincent: Sounds good! Going to leave as is for now -- resolving conversation

eng/Packages.Data.props
```diff
@@ -134,8 +134,8 @@
     <PackageReference Update="Microsoft.Azure.SignalR.Management" Version="1.19.2" />
     <PackageReference Update="Microsoft.Azure.SignalR.Protocols" Version="1.19.2" />
     <PackageReference Update="Microsoft.Azure.SignalR.Serverless.Protocols" Version="1.9.0" />
-    <PackageReference Update="Microsoft.Azure.WebJobs" Version="3.0.33" />
-    <PackageReference Update="Microsoft.Azure.WebJobs.Sources" Version="3.0.30" />
+    <PackageReference Update="Microsoft.Azure.WebJobs" Version="3.0.36" />
+    <PackageReference Update="Microsoft.Azure.WebJobs.Sources" Version="3.0.36" />
```
jsquire, line 138: @JoshLove-msft:  If I recall there was an issue in .36 and we had to stick with .35.   Can you confirm?

chiangvincent: @alrod should have context on this, I believe the issue was addressed last week?

JoshLove-msft: We were going to avoid bumping to 36 until we absolutely needed to. My understanding is that 36 has the scale controller support so we need to upgrade for this feature.

jsquire (approved):

PR #33676, opened 2023-01-25
Clarify that Verbose logging level is needed to log content

ayeltsov:
Clarifying that content logging uses Verbose logging level will save time to developers

### DIFF
--- a/sdk/core/Azure.Core/samples/Diagnostics.md
+++ b/sdk/core/Azure.Core/samples/Diagnostics.md
@@ -42,6 +42,9 @@ using AzureEventSourceListener traceListener = AzureEventSourceListener.CreateTr
 By default only URI and headers are logged. To enable content logging, set the `Diagnostics.IsLoggingContentEnabled` client option:
 
 ```C# Snippet:LoggingContent
+// Content is logged as Verbose event level
+using AzureEventSourceListener listener = AzureEventSourceListener.CreateTraceLogger(EventLevel.Verbose);
+
 SecretClientOptions options = new SecretClientOptions()
 {
     Diagnostics =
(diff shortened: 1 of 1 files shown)


sdk/core/Azure.Core/samples/Diagnostics.md
```diff
@@ -42,6 +42,9 @@ using AzureEventSourceListener traceListener = AzureEventSourceListener.CreateTr
 By default only URI and headers are logged. To enable content logging, set the `Diagnostics.IsLoggingContentEnabled` client option:
```
JoshLove-msft, line 42:
```suggestion
By default only URI and headers are logged. To enable content logging, set the `Diagnostics.IsLoggingContentEnabled` client option and specify a minimum log level of `Verbose`.
```

JoshLove-msft (approved):

sdk/core/Azure.Core/samples/Diagnostics.md
```diff
@@ -39,9 +39,12 @@ using AzureEventSourceListener traceListener = AzureEventSourceListener.CreateTr
 
 ### Enabling content logging
 
-By default only URI and headers are logged. To enable content logging, set the `Diagnostics.IsLoggingContentEnabled` client option:
+By default only URI and headers are logged. To enable content logging, set the `Diagnostics.IsLoggingContentEnabled` client option and specify a minimum log level of `Verbose`.
 
 ```C# Snippet:LoggingContent
+// Content is logged as Verbose event level
```
JoshLove-msft, line 45: So we can't directly change the md file as this gets generated after running:
```
eng\scripts\Update-Snippets.ps1
```

Instead, you'll need to update the c# [file](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/tests/samples/LoggingSamples.cs#L58) that contains this sample, and then regenerate the md file.

PR #33682, opened 2023-01-25
Support serializing received message to/from AMQP bytes

JoshLove-msft:
This lays the groundwork for binding to ServiceBusReceivedMessage in the .NET isolated functions worker.


NuGet.Config
```diff
@@ -5,6 +5,7 @@
     <!-- Do not add any additional feeds if new packages are needed they need to come from nuget.org or our azure-sdk-for-net DevOps feed -->
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
     <add key="azure-sdk-for-net" value="https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-for-net/nuget/v3/index.json" />
+    <add key="azure_app_service_staging" value="https://www.myget.org/F/azure-appservice-staging/api/v2" />
```
JoshLove-msft, line 8: This is temporary until new version of WebJobs SDK is available in Nuget.

sdk/core/Azure.Core.TestFramework/src/TestEnvironment.cs
```diff
@@ -309,7 +309,7 @@ private async Task ExtendResourceGroupExpirationAsync()
             string clientSecret = GetOptionalVariable("CLIENT_SECRET");
             string authorityHost = GetOptionalVariable("AZURE_AUTHORITY_HOST");
 
-            if (tenantId == null || clientId == null || clientSecret == null || authorityHost == null)
+            if (tenantId == null || clientId == null || clientSecret == null || authorityHost == null || ResourceManagerUrl == null)
```
JoshLove-msft, line 312: I noticed this was missing while testing - if there is no ResourceManagerUrl env var we cannot extend RG expiration.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Microsoft.Azure.WebJobs.Extensions.ServiceBus.csproj
```diff
@@ -35,4 +35,8 @@
     <Compile Include="$(AzureCoreSharedSources)AppContextSwitchHelper.cs" LinkBase="Shared" />
   </ItemGroup>
 
+  <ItemGroup>
+    <ProjectReference Include="..\..\Azure.Messaging.ServiceBus\src\Azure.Messaging.ServiceBus.csproj" />
```
JoshLove-msft, line 39: This is temporary until new SB library is shipped.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/ServiceBusModelFactory.cs
```diff
@@ -131,7 +131,7 @@ public static class ServiceBusModelFactory
             amqpMessage.MessageAnnotations[AmqpMessageConstants.DeadLetterSourceName] = deadLetterSource;
             amqpMessage.MessageAnnotations[AmqpMessageConstants.EnqueueSequenceNumberName] = enqueuedSequenceNumber;
             amqpMessage.MessageAnnotations[AmqpMessageConstants.EnqueuedTimeUtcName] = enqueuedTime.UtcDateTime;
-            amqpMessage.MessageAnnotations[AmqpMessageConstants.MessageStateName] = serviceBusMessageState;
+            amqpMessage.MessageAnnotations[AmqpMessageConstants.MessageStateName] = (int)serviceBusMessageState;
```
JoshLove-msft, line 134: This was preventing the message from being serialized by the AMQP library. Technically a breaking change, but very unlikely to impact anyone.

NuGet.Config
weshaggard: I assume by temporary you don't need to merge this PR in with this new feed correct?

JoshLove-msft: Correct

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
```diff
@@ -540,22 +546,39 @@ public virtual ServiceBusReceivedMessage AmqpMessageToSBMessage(AmqpMessage amqp
...
+        }
+
+        public virtual Guid ParseGuidBytes(ReadOnlyMemory<byte> bytes)
+        {
+            if (bytes.Length == GuidSizeInBytes)
             {
                 Span<byte> guidBytes = stackalloc byte[GuidSizeInBytes];
-                amqpMessage.DeliveryTag.AsSpan().CopyTo(guidBytes);
+                bytes.Span.CopyTo(guidBytes);
                 if (!MemoryMarshal.TryRead<Guid>(guidBytes, out var lockTokenGuid))
```
KrzysztofCwalina, line 573: Why is this reading a copy and not the original bytes? i.e. the bytes.Span

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
```diff
@@ -540,22 +546,39 @@ public virtual ServiceBusReceivedMessage AmqpMessageToSBMessage(AmqpMessage amqp
...
                 {
                     lockTokenGuid = new Guid(guidBytes.ToArray());
```
KrzysztofCwalina, line 575: I don't understand this code: TryRead return false if the size of the byte array does not correspond to the size of the Guid struct. But the Guid ctor pretty much does the same. In fact, the Guid ctor on little-endian machines just calls MemoryMarsha.Read. And so, when would the ctor call succeed when the TryRead fails (returns false)?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Compatibility/ServiceBusAmqpExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+    public static class ServiceBusAmqpExtensions
+    {
+        /// <summary>
+        /// Converts the <see cref="ServiceBusReceivedMessage"/> to its serialized AMQP form.
+        /// </summary>
+        /// <returns></returns>
+        public static BinaryData ToAmqpBytes(this ServiceBusReceivedMessage message)
+        {
+            var stream = AmqpMessageConverter.Default.AmqpAnnotatedMessageToAmqpMessage(message.GetRawAmqpMessage()).ToStream();
+            return BinaryData.FromStream(stream);
```
KrzysztofCwalina, line 23: isn't this converting byte[] to stream, and then stream to bytes[]?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Compatibility/ServiceBusAmqpExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// </summary>
+        /// <param name="messageBytes">The AMQP message bytes.</param>
+        /// <param name="lockTokenBytes">The lock token bytes.</param>
+        /// <returns></returns>
+        public static ServiceBusReceivedMessage FromAmqpBytes(BinaryData messageBytes, BinaryData lockTokenBytes)
+        {
+            var bufferStream = BufferListStream.Create(messageBytes.ToStream(), 4096);
+            var amqpMessage = AmqpLib.AmqpMessage.CreateInputMessage(bufferStream);
+
+            var message = AmqpMessageConverter.Default.AmqpMessageToSBReceivedMessage(amqpMessage);
```
KrzysztofCwalina, line 37: This also looks like some unnecessary copies. Is it not possible to simply read a byte array as a SBReceivedMessage?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Compatibility/ServiceBusAmqpExtensions.cs
JoshLove-msft: Yes, but the format we need is BinaryData, and the AMQP library doesn't currently expose a way to get the bytes directly.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Compatibility/ServiceBusAmqpExtensions.cs
JoshLove-msft: No, it needs to be read into an AMQP message first as this is where all of the decoding logic lives.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
JoshLove-msft: Good question - this code was not actually changed other than doing a slight refactor. I will see if this can be simplified.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
JoshLove-msft: Good question - this code was not actually changed other than doing a slight refactor. I will see if this can be simplified.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Compatibility/ServiceBusAmqpExtensions.cs
KrzysztofCwalina: We might want to ask the AMQP team to add an API for it.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Compatibility/ServiceBusAmqpExtensions.cs
KrzysztofCwalina: We might want to ask the AMQP team to add an API for it.

JoshLove-msft: Well the SBReceivedMessage is a Service Bus concept. Our service agnostic AMQP model is AmqpAnnotatedMessage, which lives in Azure.Core.Amqp, but does not contain any serialization logic.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
JoshLove-msft: /cc @jsquire @danielmarbach in case they have additional context.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
JoshLove-msft: I think we do the copy because the AMQP message will be disposed before returning the SBReceivedMessage.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Compatibility/ServiceBusAmqpExtensions.cs
JoshLove-msft: Filed https://github.com/Azure/azure-amqp/issues/238

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
jsquire: That is my understanding; we need the copy so that we can dispose the AMQP message.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
jsquire: I have no additional insight on this one.   Since we've copied to a temporary buffer already, I'm not sure what the marshal call adds for us here.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
```diff
@@ -218,38 +223,41 @@ public virtual AmqpMessage SBMessageToAmqpMessage(ServiceBusMessage sbMessage)
...
 
-            foreach (KeyValuePair<string, object> kvp in sbMessage.AmqpMessage.MessageAnnotations)
+            foreach (KeyValuePair<string, object> kvp in annotatedMessage.MessageAnnotations)
             {
                 switch (kvp.Key)
                 {
                     case AmqpMessageConstants.ScheduledEnqueueTimeUtcName:
-                        if ((sbMessage.ScheduledEnqueueTime != null) && sbMessage.ScheduledEnqueueTime > DateTimeOffset.MinValue)
+                        DateTimeOffset scheduledEnqueueTime = annotatedMessage.GetScheduledEnqueueTime();
+                        if (scheduledEnqueueTime != default && scheduledEnqueueTime > DateTimeOffset.MinValue)
```
christothes, line 244: nit: I think the `scheduledEnqueueTime > DateTimeOffset.MinValue` is redundant here because `DateTimeOffset.MinValue == default(DateTimeOffset)`

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
KrzysztofCwalina: The stack allocated span is destroyed even sooner (when this method returns), and so copying to the span cannot possibly help. But, I don't think you actually need any help. TryRead copies the bytes into the Guid struct. It does not hold to the  bytes passed into TryRead. Same with the Guid ctor below.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
JoshLove-msft: Yeah good call.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusExtensionConfigProvider.cs
```diff
@@ -108,6 +110,23 @@ public void Initialize(ExtensionConfigContext context)
...
+
+            byte[] combinedBytes = new byte[messageBytes.Length + lockTokenLength];
+
+            // The 16 lock token bytes go in the beginning
+            lockTokenBytes.CopyTo(combinedBytes.AsSpan());
+
+            // The AMQP message bytes go after the lock token bytes
+            messageBytes.CopyTo(combinedBytes.AsMemory(lockTokenLength));
+
+            return new ParameterBindingData("1.0", "AzureServiceBus", BinaryData.FromBytes(combinedBytes), "application/octet-stream");
```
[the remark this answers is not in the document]
JoshLove-msft: I was following the pattern from https://github.com/Azure/azure-sdk-for-net/pull/31757/files#diff-e323d68130e56f27c5b978175439cd66ade4dba43a7bcaf0e64ffcf89d03b8a4R9 where the extension name is used. I believe the consumer code would only get this particular form of ParameterBindingData, so the source is really just metadata. If a schema change were possible, we'd probably key off of the version.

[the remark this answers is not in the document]
JoshLove-msft: That said, I don't think it would hurt to be more specific and it could potentially be helpful in the future to disambiguate. @liliankasem do you have thoughts on this?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
jsquire: Ah, I thought we were talking about the body, but I'm realizing now this is also about the GUID.   Based on what you pointed out with the [GUID constructor](https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/Guid.cs,b622ef5f6b76c10a), we can probably eliminate this entire block and just pass `bytes.Span` to the constructor.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
```diff
@@ -540,24 +546,34 @@ public virtual ServiceBusReceivedMessage AmqpMessageToSBMessage(AmqpMessage amqp
...
             amqpMessage.Dispose();
 
             return sbMessage;
         }
 
+        public virtual Guid ParseGuidBytes(ReadOnlyMemory<byte> bytes)
+        {
+            if (bytes.Length == GuidSizeInBytes)
+            {
+                return new Guid(bytes.Span.ToArray());
```
danielmarbach, line 571: The previous code was optimized to not do the heap allocation in the majority of the cases now it does it again all the time. Why was this changed? 

Essentially I tried to by compliant here with .NET Standard 2.0 as well as keeping the copying mechanics in place when I optimized this code before. Unfortunately, we do not have the [GUID constructor available](https://learn.microsoft.com/en-us/dotnet/api/system.guid.-ctor?view=net-7.0#system-guid-ctor(system-readonlyspan((system-byte)))) that allows passing in the span in .NET Standard 2.0 but in theory the previous code could be optimized to `#if def` the `MemoryMarshal` part against a specific TFM that supports that constructor. 

But I would definitely try to preserve the previous memory characteristics of this code unless we can see that the before and after is not worse, unless you are hoping for me to optimize it again ;)

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusExtensionConfigProvider.cs
```diff
@@ -108,6 +110,25 @@ public void Initialize(ExtensionConfigContext context)
...
 
+        internal static ParameterBindingData ConvertReceivedMessageToBindingData(ServiceBusReceivedMessage message)
+        {
+            ReadOnlyMemory<byte> messageBytes = message.ToAmqpBytes().ToMemory();
+            byte[] lockTokenBytes = Guid.Parse(message.LockToken).ToByteArray();
+
+            // The lock token is a 16 byte GUID
+            const int lockTokenLength = 16;
+
+            byte[] combinedBytes = new byte[messageBytes.Length + lockTokenLength];
```
danielmarbach, line 121: Is there no way to do buffer rental and returning after the parameter binding data is no longer used?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
jsquire: Damn.   I never remember to check whether or not the `Span` overload is available for `netstandard2.0` - that explains the pattern that was there.

JoshLove-msft: @KrzysztofCwalina mentioned that TryRead allocates anyway - https://github.com/Azure/azure-sdk-for-net/pull/33682#discussion_r1088077365

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusExtensionConfigProvider.cs
JoshLove-msft: I don't think so - the ownership of the ParameterBindingData is not controlled by the extension.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
KrzysztofCwalina: If we want to avoid this allocation in NS2.0, we could do:
```csharp
Int32 a = BinaryPrimitives.ReadInt32LittleEndian(span);
Int16 b = BinaryPrimitives.ReadInt16LittleEndian(span.Slice(sizeof(Int32)));
Int16 c = BinaryPrimitives.ReadInt16LittleEndian(span.Slice(sizeof(Int32) + sizeof(Int16)));
var result = new Guid(a, b, c, span[8], span[9], span[10], span[11], span[12], span[13], span[14], span[15]);
return result;
```

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusExtensionConfigProvider.cs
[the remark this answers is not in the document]
liliankasem: Currently, source is only used to determine which converter we need to use to handle the content provided (i.e. to make sure we use the service bus specific converter vs the blob or string converters) so based on that I don't think we need to be more specific. I would expect the content to have any information we need about the message type? 

If we were more specific, i.e. `AzureServiceBusMessage` - what else would be something that you would potentially send over via ParameterBindingData? Just trying to think of scenarios where we would ever need the source to disambiguate

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
KrzysztofCwalina: in addition we would need to have another code path of big endian: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Guid.cs#L49

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusExtensionConfigProvider.cs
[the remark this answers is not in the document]
liliankasem: Based on the `[Extension("ServiceBus")]` header used here, I would expect this value to just be `ServiceBus`

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
KrzysztofCwalina: And BTW, I did not say that the previous code allocates. I said that it copies the data :-)

danielmarbach: When I originally contributed the version that was changed here we discussed the copying and decided keep the copying semantics. All I did was contributing a better version than the original that was faster and less allocation heavy. Didn't go down the path of a non-copying version. I still have the original benchmark available and can easily compare it to the current if there is interest

JoshLove-msft: So essentially the TryRead appears to copy without allocating, whereas when we call the Guid constructor we need to allocate due to the NS2.0 API. But if TryRead copies, why the need for the stackalloc span?

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusExtensionConfigProvider.cs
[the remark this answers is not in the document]
JoshLove-msft: We will need a converter for going in the other direction for the output case and this converter acts on a different message type - ServiceBusMessage. The Content contains binary data - so it would be better to know up front what we are working on rather than attempting to parse it to determine.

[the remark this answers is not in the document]
liliankasem: I think with the current design we would have one converter for all scenarios (input/output/trigger), so source is only there to tell us that it is service bus, but you make an interesting point. We don't have plans for output binding support at the moment and we are hoping to encourage customers to use the SDK type they already have to handle any output rather than relying on output bindings - but that's not to say we won't have output binding support, just out of scope for now so we just haven't spent much time considering the output binding design really

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
danielmarbach: I extended the previous benchmark I used to compare the memory marshal version against the original buffer block copy version

```csharp
using System.Runtime.CompilerServices;

namespace MicroBenchmarks.ServiceBus
{
    using System;
    using System.Runtime.InteropServices;
    using BenchmarkDotNet.Attributes;
    using BenchmarkDotNet.Configs;
    using BenchmarkDotNet.Diagnosers;
    using BenchmarkDotNet.Exporters;
    using BenchmarkDotNet.Jobs;

    [Config(typeof(Config))]
    public class GuidConvertion
    {
        private ArraySegment<byte> data;
        private ReadOnlyMemory<byte> dataAsMemory;

        private class Config : ManualConfig
        {
            public Config()
            {
                AddExporter(MarkdownExporter.GitHub);
                AddDiagnoser(MemoryDiagnoser.Default);
                AddJob(Job.Default.WithInvocationCount(15600000));
            }
        }
        [IterationSetup]
        public void Setup()
        {
            data = new ArraySegment<byte>(Guid.NewGuid().ToByteArray());
            dataAsMemory = new ReadOnlyMemory<byte>(data.Array, data.Offset, data.Count);
        }

        [Benchmark]
        public Guid BufferAndBlockCopy()
        {
            var guidBuffer = new byte[16];
            Buffer.BlockCopy(data.Array, data.Offset, guidBuffer, 0, 16);
            return new Guid(guidBuffer);
        }

        [Benchmark()]
        public Guid SpanAndMemoryMarshal()
        {
            Span<byte> guidBytes = stackalloc byte[16];
            data.AsSpan().CopyTo(guidBytes);
            if (!MemoryMarshal.TryRead<Guid>(guidBytes, out var lockTokenGuid))
            {
                lockTokenGuid = new Guid(guidBytes.ToArray());
            }
            return lockTokenGuid;
        }
        
        [Benchmark()]
        [SkipLocalsInit]
        public Guid SpanAndMemoryMarshalLocalsInit()
        {
            Span<byte> guidBytes = stackalloc byte[16];
            data.AsSpan().CopyTo(guidBytes);
            if (!MemoryMarshal.TryRead<Guid>(guidBytes, out var lockTokenGuid))
            {
                lockTokenGuid = new Guid(guidBytes.ToArray());
            }
            return lockTokenGuid;
        }
        
        [Benchmark(Baseline = true)]
        public Guid SpanToArray()
        {
            return new Guid(dataAsMemory.Span.ToArray());;
        }
    }
}
```

image

JoshLove-msft: @danielmarbach can you clarify why we need the stackalloc span, as opposed to just using TryRead with data.AsSpan?

danielmarbach: Let me reiterate what I mentioned before (probably worded really poorly). The original PR is this one

https://github.com/Azure/azure-sdk-for-net/pull/19857

In that my intent was to improve the existing version that did take a defensive copy. I assumed that the defensive copy was necessary due to the underlying buffer management of the Amqp Library. To keep the surface area of change small, I went ahead and replaced the block copy version with a more efficient version _that still creates a defensive_ copy. 

Whether those assumptions were correct I don't know, but it was approved by two awesome people ;)

By looking at the conversation started by @jsquire in https://github.com/Azure/azure-sdk-for-net/pull/19857#discussion_r602297286 I guess we have never settled further on the "defensive copy" part of the discussion. So in case the defensive copy is not needed, then you could definitely `TryRead` directly.

danielmarbach: And yeah that would be :speedboat: 

image

JoshLove-msft: Got it - thanks for the pointers. I think the point is that the defensive copy with a stackalloc span is unnecessary - if TryRead didn't copy then it would be a bug because the span would go away; if TryRead does copy (which is what happens) then it is unnecessary.

jsquire: Looking at that comment, at least I'm consistent in never remembering to look at which netstandard versions support the Span overload.

JoshLove-msft: Will update to add back the TryRead.

danielmarbach: Seen now @KrzysztofCwalina great remark in https://github.com/Azure/azure-sdk-for-net/pull/33682#discussion_r1088077365 and yeah I haven't considered that aspect. One never stops learning :sunglasses:

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
```diff
@@ -568,7 +568,14 @@ public virtual Guid ParseGuidBytes(ReadOnlyMemory<byte> bytes)
         {
             if (bytes.Length == GuidSizeInBytes)
             {
-                return new Guid(bytes.Span.ToArray());
+                // Use TryRead to avoid allocating an array
+                if (!MemoryMarshal.TryRead<Guid>(bytes.Span, out var lockTokenGuid))
```
KrzysztofCwalina, line 572: The direct use of span makes sense. Having said that, I started to wonder about endianness here. Guid struct has 1 Int32 and 2 Int16 fields. TryRead will work only if the passed in bytes are the right endianness. Is that the case here? i.e. what's the source of these bytes?

JoshLove-msft: The bytes are sent over the wire by the Service Bus service.

danielmarbach: I think it uses [this part ](https://github.com/Azure/azure-amqp/blob/0c4cf8a15b7df2cbe155e59c26ea23cfb6d3f652/src/Encoding/AmqpBitConverter.cs#L374)for the delivery tag but I'm not an expert

JoshLove-msft: > TryRead will work only if the passed in bytes are the right endianness

Why is this specific to TryRead? Wouldn't there be the same issue when using the ctor?

danielmarbach: Which would confirm your hunch @KrzysztofCwalina because that encodes big endian and reads big endian in the read methods. So we would always hit the ToArray case 😭

JoshLove-msft: Does TryRead not working mean it will return false? Or give the incorrect value?

KrzysztofCwalina: > Wouldn't there be the same issue when using the ctor?

@JoshLove-msft, the same issue does exist for the ctor, see the branches in https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Guid.cs#L49

danielmarbach: This should be verifiable by debugging one of the live tests right?

JoshLove-msft: So I assume it is working correctly as this code has been around for years. Even if it somehow reversed the bits it would still probably "work" as the lock token acts as an in-memory handle to the AMQP Delivery.

KrzysztofCwalina: It happens to "work" because in the normal course of operation, the client does not care about the interpretation of the bytes. They get transported from the server, TryRead into Guid, and then probably bit-blasted as-is when the delivery handle is sent back to the server. 

But, it might be a problem for distributed tracing, i.e. when the server logs the GUID as string, it might be different string value than what's logged on the server. 

BTW, I also noticed that the server code assumes the code runs on a little-endian machine: https://github.com/Azure/azure-amqp/blob/0c4cf8a15b7df2cbe155e59c26ea23cfb6d3f652/src/Encoding/AmqpBitConverter.cs#L374. For the code to be portable across little-big endian architectures, it should probably have a similar branch to what's inside the Guid's ctor in .NET.

JoshLove-msft: > Which would confirm your hunch @KrzysztofCwalina because that encodes big endian and reads big endian in the read methods. So we would always hit the ToArray case 😭

My understanding is that TryRead should only be used for little endian, not that it would return false on big endian. This means that the current code could end up reversing the lock token if the client machine is actually big endian. Using the Guid constructor handles both cases at the cost of us needing to allocate always. We can do an endian check ourselves to only use TryRead for little endian to avoid the extra allocation.

Re: the AMQP lib handling that could cause issues now that we no longer have full control over the machine where the ReceivedMessage is serialized. Since we are serializing the received message as part of this PR, we should probably file an issue so that the AMQP lib has handling for endianness.

JoshLove-msft: https://github.com/Azure/azure-amqp/issues/240

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
```diff
@@ -540,22 +546,39 @@ public virtual ServiceBusReceivedMessage AmqpMessageToSBMessage(AmqpMessage amqp
...
+
+        public virtual Guid ParseGuidBytes(ReadOnlyMemory<byte> bytes)
+        {
+            if (bytes.Length == GuidSizeInBytes)
             {
-                Span<byte> guidBytes = stackalloc byte[GuidSizeInBytes];
-                amqpMessage.DeliveryTag.AsSpan().CopyTo(guidBytes);
-                if (!MemoryMarshal.TryRead<Guid>(guidBytes, out var lockTokenGuid))
+                // Use TryRead to avoid allocating an array if we are on a little endian machine.
+                if (!BitConverter.IsLittleEndian || !MemoryMarshal.TryRead<Guid>(bytes.Span, out var lockTokenGuid))
```
KrzysztofCwalina, line 572: But the first bool expression will almost always be true (as most machines are little endian), and so it will execute the "slow path". If we want to make the mainline scenario fast (not allocating), you just need to copy the code from here: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Guid.cs#L63

JoshLove-msft: I don't follow - if most machines are little endian than the first expression would be false.

KrzysztofCwalina: I guess I missed the '!'. It was late in the day :-)

JoshLove-msft: All good 😄

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
danielmarbach: Sorry to chime into the party again. Please let me know when I'm getting annoying. I'm pretty much here just for my learning experience.

Could it be that we are making false assumptions?

By looking at the Transfer it looks like the DeliveryTag is encoded binary

https://github.com/danielmarbach/azure-amqp/blob/b26e5c701fd82142d29ff8d9588a387bc61c6db2/src/Framing/Transfer.cs#L123

it is also decoded binary.

JoshLove-msft: Your feedback is always appreciated @danielmarbach! 

> By looking at the Transfer it looks like the DeliveryTag is encoded binary

Interesting, from a spec perspective the [delivery tag](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-transport-v1.0-os.html#type-delivery-tag) is just up to 32 opaque bytes. 

I think the 16 bytes GUID is an implementation detail of Service Bus. But still the issue about the encoding of GUID not being endianness aware seems to be something that should be fixed.

danielmarbach: When you look at [this fiddle](https://dotnetfiddle.net/2QINUw) it shows the code that is already "in production" would horribly fail today since the Guid encoding would not be preserved.

The AmqpSender also uses the DeliveryTag to send an incrementing batch number. This shows the delivery tag is "arbitrary bytes" right? When the service sends us the lock token guid over the delivery tag it comes through as binary encoded and the AMQP lib will not do any sort of interpretation if my the code I pointed out is the right one.

JoshLove-msft: > When you look at [this fiddle](https://dotnetfiddle.net/2QINUw) it shows the code that is already "in production" would horribly fail today since the Guid encoding would not be preserved.

I think we are in agreement - the data is encoded as arbitrary bytes as per the spec. However, the service is constructing these bytes from a Guid, and that is why the library is parsing into a Guid.

danielmarbach: > But still the issue about the encoding of GUID not being endianness aware seems to be something that should be fixed.

Correct but I believe this is not a problem in this code because like you said and I hinted at above the DeliveryTag is opaque for the AMQP lib and it will just send or receive whether the client or service brings. 

That being said, the original version which was tweaked by me months ago used Buffer.BlockCopy which deals with endianness.

JoshLove-msft: Yep, we still need the endianness handling because otherwise our LockToken property could appear incorrect on big endian machines.

danielmarbach: What does the service send? When calling ToByteArray on GUID you essentially get "mixed-endian"

https://stackoverflow.com/questions/10190817/guid-byte-order-in-net?utm_source=pocket_saves

danielmarbach: By the way that bug then also exists in https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpReceiver.cs#L479-L482

and to properly address this you essentially have to reimplement the Guid constructor @KrzysztofCwalina mentioned and also TryWriteBytes  on the TFMs that don't support it or go back to use BufferBlockCopy or ToByteArray which would deal with endianness.

My changes have been merged almost two years ago. I wonder why we haven't heard more about this problem. Probably because big endian machines are so seldom used?

JoshLove-msft: > My changes have been merged almost two years ago. I wonder why we haven't heard more about this problem. Probably because big endian machines are so seldom used?

I suspect that is the reason. I'm going to file an issue to investigate/address all Guid related serialization/deserialization issues.

JoshLove-msft: No worries at all - we reviewed these PRs and missed it too.

JoshLove-msft: Created https://github.com/Azure/azure-sdk-for-net/issues/33825

NuGet.Config
liliankasem: WebJobs 3.0.36 has been officially released, you should be able to remove this reference and bump the WebJobs version:

https://www.nuget.org/packages/Microsoft.Azure.WebJobs/3.0.36

jsquire (approved):

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
danielmarbach: @jsquire we I did the hash code optimization in Event Hubs you somehow discovered the endianness problems on some test runs. Is event hubs using different agents than Service Bus? Could Service Bus also leverage a similar mechanism to test on more exotic operating systems / agents, like event hub does (assuming it does)?

jsquire: I noticed it running on macOS ARM (big endian) - which was coincidental because it happens to be the machine that MS issued me.   Unfortunately, the Azure DevOps machines do not have M1 specs available at this point so there's no way that I'm aware of to automate it.

I'm happy to help by running snippets as a test in the meantime, if we want to confirm behavior.

amnguye (approved):
LGTM for the Storage Webjobs Extension SDK

JoshLove-msft:
> LGTM for the Storage Webjobs Extension SDK

Thanks - I ended up reverting these as we are leaving the dependency at 3.0.35 which doesn't have the new constructor.

JohnCarmichael3000:
_"This lays the groundwork for binding to ServiceBusReceivedMessage in the .NET isolated functions worker."_

Newbie question: will this include ServiceBusMessageActions as well?

PR #33686, opened 2023-01-26
[Text Analytics] Make `FhirBundle` property in `AnalyzeHealthcareEntitiesResult` of type `BinaryData`

### LINKED ISSUE #31946: [Text Analytics] Reconsider the type of the FhirBundle property in AnalyzeHealthcareEntitiesResult
The FHIR bundle is just a JSON blob. Using what was produced by autorest as is, its type is a IReadOnlyDictionary<string, object>, which is not necessarily the best way of doing it. Consider other alternatives such as BinaryData.

joseharriaga:
Resolves: https://github.com/Azure/azure-sdk-for-net/issues/31946

The gist of this change is simple:
1. Tell AutoRest to leave the internal property as a raw `JsonElement`; no need to do work to deserialize it into a `Dictionary` that we won’t use. (see HealthcareEntitiesDocumentResultInternal.cs)
2. Edit the method that transforms the internal object to its external representation so that the `JsonElement` object is used to create a `BinaryData` object. (see Transforms.cs) 

The end result works really well in my opinion. I grabbed the [Firely SDK (the official, third-party .NET SDK for working with FHIR)](https://github.com/FirelyTeam/firely-net-sdk) to be able to get strongly-typed objects from the `BinaryData` for everything in the FHIR bundle such as patients, encounters, medications statements, etc. Here's a code sample of how a customer could use it:
```csharp
string document =
    "RECORD #333582770390100 | MH | 85986313 | | 054351 | 2/14/2001 12:00:00 AM |"
    + " CORONARY ARTERY DISEASE | Signed | DIS |"
    + Environment.NewLine
    + " Admission Date: 5/22/2001 Report Status: Signed Discharge Date: 4/24/2001"
    + " ADMISSION DIAGNOSIS: CORONARY ARTERY DISEASE."
    + Environment.NewLine
    + " HISTORY OF PRESENT ILLNESS: The patient is a 54-year-old gentleman with a history of progressive"
    + " angina over the past several months. The patient had a cardiac catheterization in July of this"
    + " year revealing total occlusion of the RCA and 50% left main disease, with a strong family history"
    + " of coronary artery disease with a brother dying at the age of 52 from a myocardial infarction and"
    + " another brother who is status post coronary artery bypass grafting. The patient had a stress"
    + " echocardiogram done on July, 2001, which showed no wall motion abnormalities, but this was a"
    + " difficult study due to body habitus. The patient went for six minutes with minimal ST depressions"
    + " in the anterior lateral leads, thought due to fatigue and wrist pain, his anginal equivalent. Due"
    + " to the patient'sincreased symptoms and family history and history left main disease with total"
    + " occasional of his RCA was referred for revascularization with open heart surgery.";

List<string> batchedDocuments = new() { document };

TextAnalyticsClient client = new(endpoint, credential);

AnalyzeHealthcareEntitiesOptions options = new()
{
    FhirVersion = WellKnownFhirVersion.V4_0_1,
    DocumentType = HealthcareDocumentType.DischargeSummary
};

AnalyzeHealthcareEntitiesOperation operation = await client.StartAnalyzeHealthcareEntitiesAsync(batchedDocuments, options: options);
await operation.WaitForCompletionAsync();

foreach (AnalyzeHealthcareEntitiesResultCollection pagedResults in operation.GetValues())
{
    foreach (AnalyzeHealthcareEntitiesResult result in pagedResults)
    {
        if (result.FhirBundle is null) continue;

        Bundle? bundle = result.FhirBundle.ToObjectFromJson<Bundle>(new JsonSerializerOptions().ForFhir(typeof(Bundle).Assembly));
        if (bundle is null) continue;

        Console.WriteLine($"Bundle ID = {bundle.Id}");
        Console.WriteLine();

       Patient? patient = (Patient?)bundle.Entry?.Where(entry => entry.Resource is Patient).FirstOrDefault()?.Resource;
        if (patient is not null)
        {
            Console.WriteLine($"Patient **********************");
            Console.WriteLine($"Gender     : {patient.Gender}");
            Console.WriteLine($"Birth Date : {patient.BirthDate}");
            Console.WriteLine();
        }

        Encounter? encounter = (Encounter?)bundle.Entry?.Where(entry => entry.Resource is Encounter).FirstOrDefault()?.Resource;
        if (encounter is not null)
        {
            Console.WriteLine($"Encounter ********************");
            Console.WriteLine($"Start Date : {encounter.Period.Start}");
            Console.WriteLine($"End Date   : {encounter.Period.End}");
            Console.WriteLine();
        }
    }
}
```
Output:
```code
Bundle ID = 46890646-ba4b-4980-b61e-e0aa8c27ed6c

Patient **********************
Gender     : Male
Birth Date : 1947-06

Encounter ********************
Start Date : 2001-05-22
End Date   : 2001-05-22
```


sdk/textanalytics/Azure.AI.TextAnalytics/src/Internal/HealthcareEntitiesDocumentResultInternal.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.Text.Json;
+using Azure.Core;
+
+namespace Azure.AI.TextAnalytics.Models
+{
+    [CodeGenModel("HealthcareEntitiesDocumentResult")]
+    internal partial class HealthcareEntitiesDocumentResultInternal
+    {
+        [CodeGenMember("FhirBundle")]
+        public JsonElement FhirBundle { get; }
```
heaths, line 13: You'll need to migrate the comment as well (or even a better one).

sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.cs
```diff
@@ -367,12 +367,16 @@ internal static AnalyzeHealthcareEntitiesResultCollection ConvertToAnalyzeHealth
             // Read entities.
             foreach (var document in results.Documents)
             {
+                BinaryData fhirBundle = document.FhirBundle.ValueKind is System.Text.Json.JsonValueKind.Object
+                    ? new BinaryData(document.FhirBundle)
+                    : null;
```
heaths, line 372: May be better to throw here if it's not what we expect? Any other STJ method during deserialization would.

heaths (approved):
A few correctness suggestions, but otherwise LGTM.

sdk/textanalytics/Azure.AI.TextAnalytics/src/Transforms.cs
joseharriaga: What type of exception would you recommend? If I'm reading this right:
1. The old code would call `property.ThrowNonNullablePropertyIsNull()` if the property was `null`, which throws a `JsonException`, but only in Debug mode. 
2. Then, the old code calls `property.Value.EnumerateObject()`, which would throw an `InvalidOperationException` when `property.Value.ValueKind` is not of `ValueKind.Object` according to this [doc](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsonelement.enumerateobject). This sounds very similar to what we want. Furthermore, I see that we have a few methods in this Transforms.cs file that also throw an `InvalidOperationException` in the case of similar discrepancies.

Based on this, I'm leaning towards throwing an `InvalidOperationException` . What do you think?

heaths: Yes, I would throw an `InvalidOperationException`.

heaths (approved):

PR #33691, opened 2023-01-26
Simplified completions method and test for same

glharper:


sdk/openai/Azure.AI.OpenAI/README.md
```diff
@@ -26,21 +23,67 @@ dotnet add package Azure.AI.OpenAI --prerelease
...
+You will also need to [register a new AAD application][aad_register_app] and [grant access][aad_grant_access] to Configuration Store by assigning the `"App Configuration Data Reader"` or `"App Configuration Data Owner"` role to your service principal.
+
+Set the values of the client ID, tenant ID, and client secret of the AAD application as environment variables: AZURE_CLIENT_ID, AZURE_TENANT_ID, AZURE_CLIENT_SECRET.
+
+```C# Snippet:CreateOpenAIClientTokenCredential
+string endpoint = "<endpoint>";
+var client = new OpenAIClient(new Uri(endpoint), new DefaultAzureCredential());
+```
 
 ## Key concepts
```
jpalvarezl, line 64: Maybe we should include a brief explanation of `embeddings` here too. I can prepare it.

sdk/openai/Azure.AI.OpenAI/README.md
```diff
@@ -106,41 +148,43 @@ foreach (var prompt in examplePrompts)
...
-    var response = completion.Choices[0].Text;
-    Console.WriteLine($"Chatbot: {response}");
-}
-```
+    Summary:
+";
 
-## Troubleshooting
+Console.Write($"Input: {summarizationPrompt}");
+var request = new CompletionsOptions();
```
m-nash, line 176: avoid `var` in samples

sdk/openai/Azure.AI.OpenAI/README.md
```diff
@@ -106,41 +148,43 @@ foreach (var prompt in examplePrompts)
...
+request.Prompt.Add(summarizationPrompt);
 
-Describe common errors and exceptions, how to "unpack" them if necessary, and include guidance for graceful handling and recovery.
+Completion completion = client.GetCompletions("myModelDeployment", request);
```
m-nash, line 179: use `Response<Completions>` also I don't think this would compile as is make sure its backed by a cs file correctly

sdk/openai/Azure.AI.OpenAI/README.md
```diff
@@ -106,41 +148,43 @@ foreach (var prompt in examplePrompts)
...
+var response = completion.Choices[0].Text;
```
m-nash, line 180: avoid `var` in samples

sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs
```diff
@@ -0,0 +1,86 @@
...
+        /// <summary> Initializes a new instance of OpenAIClient. </summary>
+        /// <param name="endpoint">
+        /// Supported Cognitive Services endpoints (protocol and hostname, for example:
+        /// https://westus.api.cognitive.microsoft.com).
+        /// </param>
+        /// <param name="credential"> A credential used to authenticate to an Azure Service. </param>
+        /// <param name="completionsDeploymentId"> default deployment id to use for completions </param>
+        /// <param name="embeddingsDeploymentId"> default deployment id to use for embeddings </param>
+        /// <exception cref="ArgumentNullException"> <paramref name="endpoint"/> or <paramref name="credential"/> is null. </exception>
+        public OpenAIClient(Uri endpoint, TokenCredential credential, string completionsDeploymentId, string embeddingsDeploymentId = null) : this(endpoint, credential, completionsDeploymentId, embeddingsDeploymentId, new OpenAIClientOptions())
```
m-nash, line 31: I think we wanted to add these into the options class vs ctor parameters?

@tg-msft ^

sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        /// <summary> Return the completion for a given prompt. </summary>
+        /// <param name="prompt"> Input string prompt to create a prompt completion from a deployment. </param>
+        /// <param name="cancellationToken"> The cancellation token to use. </param>
+        public virtual async Task<Response<Completion>> CompletionsAsync(string prompt, CancellationToken cancellationToken = default)
+        {
+            Argument.AssertNotNullOrEmpty(_completionsDeploymentId, nameof(_completionsDeploymentId));
+            Argument.AssertNotNullOrEmpty(prompt, nameof(prompt));
+
+            CompletionsRequest completionsRequest = new CompletionsRequest();
```
m-nash, line 68: I think these need to update since `CompletionsRequest` doesn't exist anymore?

sdk/openai/Azure.AI.OpenAI/README.md
glharper: This does in fact compile, see the "[Ignore("Only verifying that the sample builds")]" header in the sample file (and the library code for Response for why it works), but we can use the "completionResponse.Value.Choices[0].Text;" syntax if you feel that's more helpful. Your call.

sdk/openai/Azure.AI.OpenAI/README.md
```diff
@@ -59,33 +102,32 @@ We guarantee that all client instance methods are thread-safe and independent of
...
-Completions response = client.GetCompletions("myModelDeployment", completionsRequest);
+OpenAIClient client = new OpenAIClient("myDeploymentId", new Uri(endpoint), new DefaultAzureCredential());
 
-foreach (Choice choice in response.Choices)
-{
-    Console.WriteLine(choice.Text);
-}
+string prompt = "What is Azure OpenAI?",
+Console.Write($"Input: {prompt}");
+Completions completion = client.GetCompletions(prompt);
```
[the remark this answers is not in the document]
glharper: See comment below. I'd advocate using the Completions type, since it's syntactically cleaner.

sdk/openai/Azure.AI.OpenAI/README.md
m-nash: The issue with using `Completion` is that people copy and paste samples and sometimes they change to `var`.

If they ended up with `var completion = client.GetCompletions("myModelDeployment", request);` the next line `string response = completion.Choices[0].Text;` will not compile since the type of the completion variable is now a `Response<Completions>`.  The line as written is taking advantage of an implict conversion of `Response<Completions>` -> `Completions` which is handy but isn't very discoverable and can lead to some frustration when copy paste isn't working.

glharper: Thanks for the explanation, Michael. :-)

sdk/openai/Azure.AI.OpenAI/README.md
```diff
@@ -26,21 +21,68 @@ dotnet add package Azure.AI.OpenAI --prerelease
...
+    //...
+};
 
-Include the *Thread safety* and *Additional concepts* sections below at the end of your *Key concepts* section. You may remove or add links depending on what your library makes use of:
+foreach (string prompt in examplePrompts)
+{
+    CompletionsOptions request = new CompletionsOptions();
+    request.Prompt.Add(prompt);
+
+    Response<Completions> completion = client.GetCompletions("myModelDeployment", request);
```
m-nash, line 81: There was an archboard comment on this that I believe the api call here should be outside of the for loop.  The for loop is just adding the prompts to the request but that request should be sent once.

You can also use the initialization syntax and this example starts to look like this

```c#
OpenAIClient client = new OpenAIClient(new Uri(endpoint), new DefaultAzureCredential());
CompletionsOptions completionOptions = new CompletionsOptions()
{
    Prompt =
    {
        "What is Azure OpenAI?",
    }
};
Response<Completions> completion = client.GetCompletions("myModelDeployment", request);
string response = completion.Choices[0].Text;
Console.WriteLine($"Chatbot: {response}");
```

sdk/openai/Azure.AI.OpenAI/README.md
```diff
@@ -26,21 +21,68 @@ dotnet add package Azure.AI.OpenAI --prerelease
...
+    string response = completion.Choices[0].Text;
```
m-nash, line 82: avoid request and response as variable names unless you they http request / response types

sdk/openai/Azure.AI.OpenAI/README.md
```diff
@@ -94,53 +137,56 @@ List<string> examplePrompts = new(){
     "Describe in single words only the good things that come into your mind about your mother.",
 };
 
-foreach (var prompt in examplePrompts)
+foreach (string prompt in examplePrompts)
 {
     Console.Write($"Input: {prompt}");
-    var request = new CompletionsOptions();
+    CompletionsOptions request = new CompletionsOptions();
```
m-nash, line 143: avoid request / response variable names

sdk/openai/Azure.AI.OpenAI/tests/OpenAIInferenceTests.cs
```diff
@@ -54,5 +65,18 @@ public async Task EmbeddingTest()
...
+        /// <summary>
+        /// Test Exception throw.
+        /// </summary>
+        [RecordedTest]
+        public void RequestFailedExceptionTest()
+        {
+            var client = GetClient();
+            CompletionsOptions completionsRequest = new CompletionsOptions();
+            completionsRequest.Prompt.Add("Hello world");
+            Assert.That(async () => await client.GetCompletionsAsync("BAD_DEPLOYMENT_ID", completionsRequest),
```
m-nash, line 78: nit: should we assert the code that we expect to come back?

sdk/openai/Azure.AI.OpenAI/README.md
```diff
@@ -26,21 +21,68 @@ dotnet add package Azure.AI.OpenAI --prerelease
...
+You must have an [Azure subscription](https://azure.microsoft.com/free/dotnet/) and [OpenAI access](https://learn.microsoft.com/azure/cognitive-services/openai/overview#how-do-i-get-access-to-azure-openai).
 
 ### Authenticate the client
 
-If your library requires authentication for use, such as for Azure services, include instructions and example code needed for initializing and authenticating.
+In order to interact with the App Configuration service, you'll need to create an instance of the [OpenAIClient][openai_client_class] class. To make this possible, you'll need the endpoint string for your OpenAI resource, as well as your Azure Subscription Key.
```
trrwilson, line 28: > App Configuration service
This seems confusing; can we call it the "Azure OpenAI service" or even just "the service?"

> your OpenAI resource
Suggest "Azure OpenAI" -- given the business relationship, it's likely a good idea to keep snapped to the resource type and keep branding very crisp in public docs. If/when we explicitly want to document interoperability with non-Azure OpenAI endpoints, we can cover that explicitly and not rely on a subtle "is a" relationship of AOAI/OAI.

> your Azure Subscrition Key
Agreed with Michael; would suggest this could overall be "the endpoint URI for your Azure OpenAI resource and an API key to access that resource."

sdk/openai/Azure.AI.OpenAI/README.md
```diff
@@ -26,21 +21,68 @@ dotnet add package Azure.AI.OpenAI --prerelease
...
+You can obtain the endpoint string and subscription key from the Azure OpenAI Portal.
+
+#### Create OpenAIClient
+
+Once you have the value of the endpoint string and subscription key, you can create the OpenAIClient:
 
-For example, include details on obtaining an account key and endpoint URI, setting environment variables for each, and initializing the client object.
+```C# Snippet:CreateOpenAIClient
+// Replace with your Azure subscription key
+string key = "YOUR_AZURE_KEY";
```
trrwilson, line 40: Since we're going more in depth, open question with @m-nash likely knowing best of contemporary participants: is there general guidance on when docs/sample collateral such as this should use implicit typing (`var`) versus explicit types? My developer intuition would be to use `var` for string literals and direct constructor use (all LHS in this block) and this would feel "stiff" to me to read; but if we gravitate towards "stiff" to accommodate things like that discussed `Response<T>` magic and its pitfalls, then "when in Rome."

sdk/openai/Azure.AI.OpenAI/README.md
trrwilson: If this is our lead-in ramp, it'd probably be good for us to use the simplified call (single prompt) in any event; multi-prompt requests already feel functionally marginal to me, and can definitely be construed as more advanced usage. At the very least, we could reserve it for a more detailed example that demonstrates the setting of other request options at the same time.

sdk/openai/Azure.AI.OpenAI/README.md
m-nash: In an IDE setting var is fine because its easy to hover and see exactly what type the variable actually is.  In docs that's a lot harder to do so as a general rule we use explicit variable types to make it as clear as possible what is happening.

Part of this is also consistency with the other API documentation in the repo.  Although there might be outliers that slipped through we try to keep a consistent style across all libraries in this repo.

sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs
tg-msft: No, I think we want them on the .ctor because they're required to use the Azure service correctly.  I'd also move them before the `credential` because they're part of the "addressing the resource" set of params.  My bigger concern is adding one per operation though.  OpenAI has a decent number of operations today and this will get messy quickly.

Instead, I'd suggest something like:

```C#
public OpenAIClient(Uri endpoint, string deploymentId, TokenCredential credential)
  : this(endpoint, deploymentId, credential, null)
{
}

public OpenAIClient(Uri endpoint, string deploymentId, TokenCredential credential, OpenAIClientOptions options)
{
    // ...
    options ??= new OpenAIClientOptions();
    // ...
    _completionsDeploymentId = options.CompletionsDeploymentId ?? deploymentId;
    _embeddingsDeploymentId = options.EmbeddingsDeploymentId ?? deploymentId;
}
```

Which lets people pass a single global `deploymentId` for all operations easily but still customize the `deploymentId` globally per operation type via the `OpenAIClientOptions` or per invocation via `*Options` parameters.  The  `OpenAIClientOptions` changes could be skipped for preview if we're low on time because the default .ctor value could still be overridden per invocation.

We can also talk in the future about creating a `OpenAIClientOptions.DefaultDeployments` model type/property if we want to group all the deployments in a single place so we don't have them sprinkled throughout the client options.

sdk/openai/Azure.AI.OpenAI/README.md
tg-msft: Does this sample code compile?  `Response<Completions> completion` shouldn't have a `.Choices` property.  Are we using snippets for our samples?

glharper: This fell through the cracks, as it's not tied to a snippet. Good catch.

sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs
glharper: Updated the constructor to match this comment (no changes to OpenAIClientOptions until after preview).

m-nash: @tg-msft I assume in there we wanted the ctor to be mirroed for both the TokenCredential set and the AzureKeyCredential set?  Also did we want to remove the `deploymentId` from the method signatures?  If so should we remove the ctor that doesn't have the `deploymentId` argument?

sdk/openai/Azure.AI.OpenAI/README.md
jpalvarezl: https://github.com/Azure/azure-sdk-for-net/issues/33789

sdk/openai/Azure.AI.OpenAI/README.md
```diff
@@ -26,21 +21,68 @@ dotnet add package Azure.AI.OpenAI --prerelease
...
+
+You can obtain the endpoint string and subscription key from the Azure OpenAI Portal.
+
+#### Create OpenAIClient
+
+Once you have the value of the endpoint string and subscription key, you can create the OpenAIClient:
 
-For example, include details on obtaining an account key and endpoint URI, setting environment variables for each, and initializing the client object.
+```C# Snippet:CreateOpenAIClient
+// Replace with your Azure subscription key
```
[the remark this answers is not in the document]
jpalvarezl: https://github.com/Azure/azure-sdk-for-net/issues/33789

sdk/openai/Azure.AI.OpenAI/README.md
```diff
@@ -26,21 +21,68 @@ dotnet add package Azure.AI.OpenAI --prerelease
...
+Client subscription key authentication is used in most of the examples in this getting started guide, but you can also authenticate with Azure Active Directory using the [Azure Identity library][azure_identity]. To use the [DefaultAzureCredential][azure_identity_dac] provider shown below,
+or other credential providers provided with the Azure SDK, please install the Azure.Identity package:
+
+```dotnetcli
+dotnet add package Azure.Identity
+```
+
+You will also need to [register a new AAD application][aad_register_app] and [grant access][aad_grant_access] to Configuration Store by assigning the `"App Configuration Data Reader"` or `"App Configuration Data Owner"` role to your service principal.
+
+Set the values of the client ID, tenant ID, and client secret of the AAD application as environment variables: AZURE_CLIENT_ID, AZURE_TENANT_ID, AZURE_CLIENT_SECRET.
```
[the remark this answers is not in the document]
jpalvarezl: https://github.com/Azure/azure-sdk-for-net/issues/33789

sdk/openai/Azure.AI.OpenAI/README.md
jpalvarezl: https://github.com/Azure/azure-sdk-for-net/issues/33789

sdk/openai/Azure.AI.OpenAI/README.md
jpalvarezl: https://github.com/Azure/azure-sdk-for-net/issues/33789

sdk/openai/Azure.AI.OpenAI/README.md
jpalvarezl: https://github.com/Azure/azure-sdk-for-net/issues/33789

sdk/openai/Azure.AI.OpenAI/tests/OpenAIInferenceTests.cs
jpalvarezl: https://github.com/Azure/azure-sdk-for-net/issues/33789

sdk/openai/Azure.AI.OpenAI/README.md
jpalvarezl: https://github.com/Azure/azure-sdk-for-net/issues/33789

sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs
m-nash: https://github.com/Azure/azure-sdk-for-net/issues/33789

m-nash (approved):

sdk/openai/Azure.AI.OpenAI/README.md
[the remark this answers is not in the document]
jpalvarezl: Remove line `56` and link to @m-nash documentation

sdk/openai/Azure.AI.OpenAI/README.md
jpalvarezl: Go for @trrwilson suggestion.

sdk/openai/Azure.AI.OpenAI/README.md
jpalvarezl: Verify that's tied to a snippet. Rename `response` and `request`. Scan full doc for other occurrences.

sdk/openai/Azure.AI.OpenAI/README.md
jpalvarezl: Bad naming.

sdk/openai/Azure.AI.OpenAI/README.md
jpalvarezl: `App Configuration service` should be `Azure OpenAI service`. Rename according to @trrwilson 's suggestions.

sdk/openai/Azure.AI.OpenAI/README.md
jpalvarezl: We keep they explicit type for consistency's sake.

jpalvarezl: Double check that we don't use `var` in any of the `*.md` files

sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs
jpalvarezl: https://github.com/Azure/azure-sdk-for-net/issues/33815

PR #33695, opened 2023-01-26
BlobLeaseClient Avoid Log 409 response

jaschrep-msft:
When attempting to acquire a lease on a blob or container where one already exists, the service returns a 409. An application may wish to consider this attempt as part of the normal flow of operation and schedule a retry at a later time, rather than consider it an error upfront and log it as such.

Error suppression on this operation is necessary to address #21511.

This introduces an overload to `AcquireLease()` that accepts a `ResponseClassifier` that can use a predicate to determine whether a response should be let through as a non-error. The mentioned issue can then construct the appropriate predicate for its scenario.


sdk/storage/Azure.Storage.Blobs/api/Azure.Storage.Blobs.netstandard2.0.cs
```diff
@@ -1637,8 +1637,12 @@ public partial class BlobLeaseClient
...
+        public virtual Azure.Response<Azure.Storage.Blobs.Models.BlobLease> Acquire(System.TimeSpan duration, Azure.RequestConditions conditions, System.Threading.CancellationToken cancellationToken) { throw null; }
+        public virtual System.Threading.Tasks.Task<Azure.Response<Azure.Storage.Blobs.Models.BlobLease>> AcquireAsync(System.TimeSpan duration, Azure.RequestConditions conditions = null, Azure.RequestContext requestContext = null) { throw null; }
+        [System.ComponentModel.EditorBrowsableAttribute(System.ComponentModel.EditorBrowsableState.Never)]
+        public virtual System.Threading.Tasks.Task<Azure.Response<Azure.Storage.Blobs.Models.BlobLease>> AcquireAsync(System.TimeSpan duration, Azure.RequestConditions conditions, System.Threading.CancellationToken cancellationToken) { throw null; }
         public virtual Azure.Response<Azure.Storage.Blobs.Models.BlobLease> Break(System.TimeSpan? breakPeriod = default(System.TimeSpan?), Azure.RequestConditions conditions = null, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
```
tg-msft, line 1646: We don't want to hide the old methods - those should still be the default for most customers.  I think we'd want this to look more like:
```C#
// Existing HLC conveniences
public virtual Response<BlobLease> Acquire(TimeSpan duration, RequestConditions conditions = null, CancellationToken cancellationToken = default);
public virtual Task<Response<BlobLease>> AcquireAsync(TimeSpan duration, RequestConditions conditions = null, CancellationToken cancellationToken = default);

// New protocol additions
public virtual Response Acquire(TimeSpan duration, RequestConditions conditions, RequestContext requestContext);
public virtual Task<Response> AcquireAsync(TimeSpan duration, RequestConditions conditions, RequestContext requestContext);
```

amnguye (approved):

sdk/storage/Azure.Storage.Blobs/src/BlobLeaseClient.cs
```diff
@@ -307,37 +410,31 @@ private void EnsureClient()
...
+                                requestConditions: conditions,
                                 ifTags: tagCondition,
-                                cancellationToken: cancellationToken);
+                                context: context);
                         }
 
                         response = Response.FromValue(
-                            blobClientResponse.ToBlobLease(),
-                            blobClientResponse.GetRawResponse());
+                            ResponseWithHeaders.FromValue(new BlobAcquireLeaseHeaders(rawResponse), rawResponse).ToBlobLease(),
```
tg-msft, line 436: What happens here on a failed request (that I told not to throw via `RequestContext`'s classifier)?  Will creating the new headers object throw any exceptions?  If so may be we should only instantiate the object if `rawResponse.Status == 201` and use `null` otherwise?

tg-msft (approved):
Looks great - I'm really happy we're making this change.

sdk/storage/Azure.Storage.Blobs/src/BlobLeaseClient.cs
tg-msft: Alternately, we could also create two versions of `AcquireInternal` where one simply wraps the other but does the CancellationToken => RequestContext on input and Raw Response => Strong Type on output.  Then the protocol code path doesn't need to be smart enough to worry about those edge cases.

jaschrep-msft: No all these models have a lot of null safety built in and use try-gets on headers.

However, it's probably best if the internal method returns only raw response and the methods that need the response<T> do the construction instead.

jaschrep-msft: haha I did not see the second comment when I posted my response. I think we are aligned here and the changes have been made, though not in that exact structure you describe.

amnguye (approved):

PR #33716, opened 2023-01-27
Set parent Id when starting ActivitySource

### LINKED ISSUE #33338: [BUG] Correlation between Producer and Consumer Traces is broken in Azure.Message.ServiceBus OpenTelemetry Support
### Library name and version

Azure.Message.ServiceBus 7.11.1

### Describe the bug

I am posting a message to a service bus topic in one of my web api endpoint as below

```cs
[HttpPost]
public async Task<IActionResult> PostServiceBusMessage()
{
    _logger.LogTrace("{functionName} requested", nameof(PostServiceBusMessage));

   string connectionString = configuration.GetValue<string>("ServiceBusConnectionStringSend") ;
    string topicName = configuration.GetValue<string>("topicName");

    await using var client = new ServiceBusClient(connectionString);

    ServiceBusSender sender = client.CreateSender(topicName);

    var message = new ServiceBusMessage("Hello world!")
    {
        ScheduledEnqueueTime = DateTime.UtcNow,
        MessageId = Guid.NewGuid().ToString("D")
    };
    await sender.SendMessageAsync(message);
    return Ok();
}
```

A consumer class is defined in the same project inheriting the BackgroundService like 

```cs
using System;
using System.Threading;
using System.Threading.Tasks;
using Azure.Messaging.ServiceBus;
using Microsoft.Extensions.Configuration;
using Microsoft.Extensions.Hosting;
using Microsoft.Extensions.Logging;

namespace Tracing
{
    public class AzureServiceBusConsumer : BackgroundService
    {
        private readonly ServiceBusProcessor _processor;
        private readonly ILogger _logger;

        public AzureServiceBusConsumer(
            ILogger<AzureServiceBusConsumer> logger,
            IConfiguration configuration)
        {
            _logger = logger;
            _processor = new ServiceBusClient(configuration.GetValue<string>("ServiceBusConnectionStringListen"))
                .CreateProcessor(configuration.GetValue<string>("topicName"), configuration.GetValue<string>("subscriptionName"));
        }

        private async Task ProcessMessagesAsync(ProcessMessageEventArgs args)
        {
            var message = args.Message;
            if (message.ApplicationProperties.TryGetValue("Diagnostic-Id", out var objectId) && objectId is string diagnosticId)
            {
                Console.WriteLine($"----- Diagnostic-Id = {diagnosticId}");
            }
            _logger.LogInformation("Message found to process");
            await args.CompleteMessageAsync(message);
            return;
        }

        public override async Task StopAsync(CancellationToken cancellationToken)
        {
            await _processor.StopProcessingAsync(cancellationToken);
        }

        protected override async Task ExecuteAsync(CancellationToken stoppingToken)
        {
            _processor.ProcessMessageAsync += ProcessMessagesAsync;
            _processor.ProcessErrorAsync += ProcessErrorAsync;
            await _processor.StartProcessingAsync(stoppingToken);
        }

        private Task ProcessErrorAsync(ProcessErrorEventArgs args)
        {
            var message = args.Exception.Message;
            _logger.LogInformation(message);
            return Task.CompletedTask;
        }
    }
}
```

It is set up in Startup ConfigureServices(IServiceCollection services) as
```cs
services.AddHostedService<AzureServiceBusConsumer>();
```

When I send a message to service bus via invoking my web api endpoint, It is received by the consumer **but totally with a different trace Id** and the link between sender and consumer traces is totally broken.

image

TraceIds highlighted in yellow are within the webapi method where a message is sent to this service bus topic,  ApplicationProperty "Diagnostic-Id" contains this information, **BUT the TraceId in the consumer side is totally different (Underlined in red color)**

Jaeger is not happy with this and does not show anything about the consumer side because of this totally different traceId.
image

End to end Observability is broken in this scenario.

### Expected behavior

Service bus Consumer should expose Activity same TraceId as of Producer.

### Actual behavior

TraceIds are totally different between Producer and Consumer and this OpenTelemetry Instrumentation seems broken. 

### Reproduction Steps

It is reproducible every time with steps and sample code mentioned in description.

### Environment

Service bus is running in Azure.

Windows 11

Microsoft Visual Studio Enterprise 2022 (64-bit) - Current
Version 17.4.1

JoshLove-msft:
Fixes https://github.com/Azure/azure-sdk-for-net/issues/33338

Will add Service Bus E2E tests in a separate PR.

### DIFF
--- a/sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
+++ b/sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
@@ -340,7 +340,8 @@ namespace Azure.Core.Pipeline
                     (int)_kind,
                     startTime: _startTime,
                     tags: _tagCollection,
-                    links: GetActivitySourceLinkCollection());
+                    links: GetActivitySourceLinkCollection(),
+                    parentId: _traceparent);
             }
 
             public void SetStartTime(DateTime startTime)
@@ -404,7 +405,7 @@ namespace Azure.Core.Pipeline
         private static Action<Activity, string?>? SetTraceStateStringMethod;
         private static Func<Activity, int>? GetIdFormatMethod;
         private static Action<Activity, string, object?>? ActivityAddTagMethod;
-        private static Func<object, string, int, ICollection<KeyValuePair<string, object>>?, IList?, DateTimeOffset, Activity?>? ActivitySourceStartActivityMethod;
+        private static Func<object, string, int, string?, ICollection<KeyValuePair<string, object>>?, IList?, DateTimeOffset, Activity?>? ActivitySourceStartActivityMethod;
         private static Func<object, bool>? ActivitySourceHasListenersMethod;
         private static Func<string, string?, ICollection<KeyValuePair<string, object>>?, object?>? CreateActivityLinkMethod;
         private static Func<ICollection<KeyValuePair<string,object>>?>? CreateTagsCollectionMethod;
@@ -667,7 +668,7 @@ namespace Azure.Core.Pipeline
             return ActivitySourceHasListenersMethod.Invoke(activitySource);
         }
 
-        public static Activity? ActivitySourceStartActivity(object? activitySource, string activityName, int kind, DateTimeOffset startTime, ICollection<KeyValuePair<string, object>>? tags, IList? links)
+        public static Activity? ActivitySourceStartActivity(object? activitySource, string activityName, int kind, DateTimeOffset startTime, ICollection<KeyValuePair<string, object>>? tags, IList? links, string? parentId)
         {
             if (activitySource == null)
             {
@@ -681,7 +682,7 @@ namespace Azure.Core.Pipeline
                     ActivityContextType == null ||
                     ActivityKindType == null)
                 {
-                    ActivitySourceStartActivityMethod = (_, _, _, _, _, _) => null;
+                    ActivitySourceStartActivityMethod = (_, _, _, _, _, _, _) => null;
                 }
                 else
                 {
@@ -689,7 +690,7 @@ namespace Azure.Core.Pipeline
                     {
                         typeof(string),
                         ActivityKindType,
-                        ActivityContextType,
+                        typeof(string),
                         typeof(IEnumerable<KeyValuePair<string, object>>),
                         typeof(IEnumerable<>).MakeGenericType(ActivityLinkType),
                         typeof(DateTimeOffset)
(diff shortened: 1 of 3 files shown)


sdk/core/Azure.Core/tests/ClientDiagnosticsTests.Net50.cs
```diff
@@ -369,6 +369,54 @@ public void StartActivitySourceActivityIgnoresInvalidLinkParent()
 
             Assert.AreEqual(0, activityListener.Activities.Single().Links.Count());
         }
+
+        [Test]
+        [NonParallelizable]
+        public void ParentIdCanBeSetActivitySource()
+        {
+            using var _ = SetAppConfigSwitch();
+            string parentId = "00-6e76af18746bae4eadc3581338bbe8b1-2899ebfdbdce904b-00";
```
JoshLove-msft, line 378: Interestingly, the ID must be W3C format when calling ActivitySource.StartActivity, but not when calling SetParentId on an already started activity that was created via ActivitySource. I don't think this is an issue as I believe the ID would always be W3C format in the activity source path.

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
```diff
@@ -681,49 +682,50 @@ public static bool ActivitySourceHasListeners(object? activitySource)
...
                         var linksParameter = Expression.Parameter(typeof(IList));
                         var methodParameter = method.GetParameters();
-                        ActivitySourceStartActivityMethod = Expression.Lambda<Func<object, string, int, ICollection<KeyValuePair<string, object>>?, IList?, DateTimeOffset, Activity?>>(
+                        ActivitySourceStartActivityMethod = Expression.Lambda<Func<object, string, int, string?, ICollection<KeyValuePair<string, object>>?, IList?, DateTimeOffset, Activity?>>(
                             Expression.Call(
                                 Expression.Convert(sourceParameter, method.DeclaringType!),
                                 method,
                                 nameParameter,
                                 Expression.Convert(kindParameter,  methodParameter[1].ParameterType),
-                                Expression.Default(ActivityContextType),
```
JoshLove-msft, line 717: We were previously using the default ActivityContextType, so the change to use the overload that does not include this parameter appears safe.

jsquire (approved):
I can't speak to the validity of the change in the context of the OTel spec, but the implementation looks reasonable.

sdk/core/Azure.Core/tests/ClientDiagnosticsTests.Net50.cs
JoshLove-msft: We discovered Activity does set the parent, but we will prevent this in DiagnosticScope.
Yes, the DiagSource path has a test that sets the parent not using W3C format.

lmolkova (approved):
Thank you!

PR #33767, opened 2023-01-31
Add SecretClient.GetSecret with RequestContext

### LINKED ISSUE #25125: Delete throws error if secret doesn't exist
It is not clear how to delete a secret that might not exist. In most cases, I would definitely expect the default behaviour not to throw an exception if the secret does not exist but at worst, there should be an overload that allows us to ignore the error. Throwing an error means that for every delete, I either need to call GetSecret first or catch and ignore the exception which is nasty.

Also, the link to log a problem about "This Product" is broken.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 2b378d7c-6191-9f15-6c34-30c4bfbcb5a5
* Version Independent ID: 3fafa4d7-15a9-82dc-669a-f9e3b61a9065
* Content: [SecretClient.StartDeleteSecretAsync(String, CancellationToken) Method (Azure.Security.KeyVault.Secrets) - Azure for .NET Developers](https://docs.microsoft.com/en-us/dotnet/api/azure.security.keyvault.secrets.secretclient.startdeletesecretasync?view=azure-dotnet#Azure_Security_KeyVault_Secrets_SecretClient_StartDeleteSecretAsync_System_String_System_Threading_CancellationToken_)
* Content Source: [xml/Azure.Security.KeyVault.Secrets/SecretClient.xml](https://github.com/Azure/azure-docs-sdk-dotnet/blob/master/xml/Azure.Security.KeyVault.Secrets/SecretClient.xml)
* Service: **azure**
* GitHub Login: @rloutlaw
* Microsoft Alias: **routlaw**

heaths:
Fixes #25125


heaths:
@KrzysztofCwalina @tg-msft @pallavit I'm reluctant to smash this one into our upcoming release in a couple weeks. Initially when I spoke with @annelo-msft it seemed pretty straightforward. It was, actually, but as you can see required a lot of small changes. I tried to minimize the impact to code, but honestly a good refactoring of the `KeyVaultPipeline` (created when the Azure.Core pipelining was still in development) may be best. That said, I don't see any unnecessary allocations and should have minimized / eliminated any unnecessary stack pushes e.g., returning a value and not using it.

I'm open to pushing this into our upcoming GA, but leaning toward punting it till after to get more usage on a beta release. The issue has been open for 3 years, so another few months (or so) probably won't hurt.

/cc @vcolin7

sdk/keyvault/Azure.Security.KeyVault.Secrets/tests/SecretClientLiveTests.cs
```diff
@@ -479,5 +480,27 @@ public async Task AuthenticateCrossTenant()
 
             Assert.AreEqual(200, response.GetRawResponse().Status);
         }
+
+        [RecordedTest]
+        public async Task GetNonExistentSecretNoThrow()
```
annelo-msft, line 485: Would it be worth adding a test to validate that errors aren't logged either?

sdk/keyvault/Azure.Security.KeyVault.Shared/src/KeyVaultPipeline.cs
```diff
@@ -88,17 +90,28 @@ public Request CreateRequest(RequestMethod method, params string[] path)
...
-            return request;
+            return message;
         }
 
-#pragma warning disable CA1822 // Member can be static
-        public Response<T> CreateResponse<T>(Response response, T result)
+        public static NullableResponse<T> CreateResponse<T>(HttpMessage message, Func<T> resultFactory)
+            where T : IJsonDeserializable
+        {
+            if (message.Response.Status == 200 || message.Response.Status == 201)
```
annelo-msft, line 99: Do we know this is the full set of possible status codes for all cases?  Is there another check (like whether the response.Content is non-null) that might be valid here and reach a broader set of cases?  I defer to you on service logic, so if not, you can disregard this comment.

sdk/keyvault/Azure.Security.KeyVault.Shared/src/KeyVaultPipeline.cs
```diff
@@ -55,9 +55,10 @@ public Uri CreateFirstPageUri(string path, params ValueTuple<string, string>[] q
             return firstPage.ToUri();
         }
 
-        public Request CreateRequest(RequestMethod method, Uri uri, bool appendApiVersion)
+        public HttpMessage CreateRequest(RequestMethod method, ResponseClassifier classifier, Uri uri, bool appendApiVersion)
         {
-            Request request = _pipeline.CreateRequest();
+            HttpMessage message = _pipeline.CreateMessage(null, classifier);
+            Request request = message.Request;
```
annelo-msft, line 61: This approach LGTM.  One comment is that if you ever wanted to use other functionality available on RequestContext, or expose those APIs to callers, using RequestContext in the pipeline.CreateMessage() call rather than passing the classifier directly would give you that ability in the future.  But I don't think it's a huge deal if it's considerably more work.

sdk/keyvault/Azure.Security.KeyVault.Secrets/src/SecretClient.cs
```diff
@@ -139,6 +139,76 @@ public virtual Response<KeyVaultSecret> GetSecret(string name, string version =
...
+            {
+                return _pipeline.SendRequest(RequestMethod.Get, ResponseClassifier200404, () => new KeyVaultSecret(), cancellationToken, SecretsPath, name, "/", version);
+            }
+            catch (Exception e)
+            {
+                scope.Failed(e);
+                throw;
+            }
+        }
+#pragma warning restore AZC0015 // Unexpected client method return type.
```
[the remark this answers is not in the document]
heaths: They only apply to the method return type and I prefer this style. It's prevalent throughout code I've written because I don't really like how `#pragma`s around a single line break up the visual flow of the code.

sdk/keyvault/Azure.Security.KeyVault.Secrets/tests/SecretClientLiveTests.cs
heaths: As opposed to distributed tracing you mean? If there's some easy way, sure, but I also feel that making sure DT and logging are logged or not should be in the purview of Azure.Core and not repeated throughout. Better to isolate the tests of such conditionals to whatever implements them, which is Core.

sdk/keyvault/Azure.Security.KeyVault.Shared/src/KeyVaultPipeline.cs
heaths: I agree this is a hard call, but I don't want to check `Content`. The idea is that `Content` was non-null but had only an `{"error":{}}` object such that our factory pattern for deserializing would instantiate an "empty" object, such that `Response<KeyVaultSecret>.Value` was non-null but otherwise useless. I would want either a null `Value` or throw, such as this implementation does consistent with its documentation for `Value`.

Going by the success codes further down we supported before - 200, 201, 202, and 204 - 200 and 201 should really be the only values that should return content. Well, maybe 202. Just to be sure, I'll add that in. We would not expect 204 to return content, given it's `204 No Content`.

sdk/keyvault/Azure.Security.KeyVault.Shared/src/KeyVaultPipeline.cs
heaths: That's a good idea, but for now I'll add a comment as a design idea. For now, it'd just be an unnecessary allocation.

pallavit:
> I'm open to pushing this into our upcoming GA, but leaning toward punting it till after to get more usage on a beta release. The issue has been open for 3 years, so another few months (or so) probably won't hurt.

Agreed with the approach.

heaths:
> > I'm open to pushing this into our upcoming GA, but leaning toward punting it till after to get more usage on a beta release. The issue has been open for 3 years, so another few months (or so) probably won't hurt.
> 
> Agree with the approach.

@pallavit could you clarify: you agree with the idea to punt this till after the 7.4-related GA, or you agree with the concept in this PR (as @annelo-msft said in her PR review)?

pallavit:
I agree with moving this out of the GA scope :)

heaths:
I'll leave it as a draft for now then and can rebase after we GA.

sdk/keyvault/Azure.Security.KeyVault.Secrets/src/SecretClient.cs
```diff
@@ -139,6 +139,76 @@ public virtual Response<KeyVaultSecret> GetSecret(string name, string version =
             }
         }
 
+#pragma warning disable AZC0015 // Unexpected client method return type.
```
pallavit, line 142: >AZC0015 [](http://example.com/codeflow?start=24&length=7)

Is this coming from our analyzers?

christothes: Yes, although it should be [fixed now ](https://github.com/Azure/azure-sdk-tools/pull/4512)- we just need to update the version reference to the latest analyzer

heaths: Updated. @christothes I also opened PR #33817 to update it in main since I'll hold on this PR until after we GA.

heaths:
@KrzysztofCwalina @tg-msft now that Storage has shipped with a `RequestContext` overload like I wanted here, can we revisit shipping with an overload and I can do a sample how to do this no-throw technique using the RC?

sdk/keyvault/Azure.Security.KeyVault.Secrets/tests/SecretClientLiveTests.cs
annelo-msft: Sounds good.

heaths: @annelo-msft do you, @JoshLove-msft, @christothes and whoever else is focused on core feel it has decent general test coverage for dtracing now?

sdk/keyvault/Azure.Security.KeyVault.Shared/src/KeyVaultPipeline.cs
```diff
@@ -55,9 +55,12 @@ public Uri CreateFirstPageUri(string path, params ValueTuple<string, string>[] q
             return firstPage.ToUri();
         }
 
-        public Request CreateRequest(RequestMethod method, Uri uri, bool appendApiVersion)
+        public HttpMessage CreateRequest(RequestMethod method, ResponseClassifier classifier, Uri uri, bool appendApiVersion)
```
christothes, line 58: Should we rename this to `CreateMessage` now?

heaths:
Per offline discussion with @KrzysztofCwalina, @tg-msft, and @m-nash I am working on changing this to an overload of `GetSecret(..., RequestContext)` but the desirable change may not be possible. Previously, @KrzysztofCwalina or @tg-msft indicated I should make the existing signature e.g., `Response<KeyVaultSecret> GetSecret(string name, string version = null, CancellationToken cancellationToken = default)` the all-required signature, similar to what we do for overloads any other time. This means making `version` required as well, which isn't a problem. However, this does mean that the typical caller e.g., `GetSecret("secret-name")` is now broken because the response type now changes to a `NullableResponse<KeyVaultSecret>` instread of the derived `Response<KeyVaultSecret>` i.e., `Response<T> : NullableResponse<T>`. This means that the new method can't return a `NoValueResponse<KeyVaultSecret>`.

I three possible outcomes and am torn on the first two:

1. Leave this PR as-is at the time this comment was posted i.e., define a singularly purposed `GetSecretIfExists`. It's obvious but it serves one purpose.
2. Keep the existing signature as is and make the new convenience method taking a `RequestContext` use all-required parameters i.e., `NullableResponse<KeyVaultSecret> GetSecret(string name, string version, RequestContect context)`. This provides more robust support for user scenarios without breaking existing code. Anyone who wants to use it has to pass in a `RequestContext` anyway. The downside is that this does not seem to fit the model @KrzysztofCwalina and @tg-msft want, but I wonder if it was considered that existing HLCs or hand-coded clients are likely returning a `Response<T>` so may have this same problem. `NullableResponse<T>` is newer.
3. Define a new method name - a "pseudo-overload" like `GetSecretIfExists` - that takes a `RequestContext` but has a different name e.g., `GetSecretWithContext`. This feels far afoul from what we want to accomplish long-term, though.

KrzysztofCwalina:
> @KrzysztofCwalina @tg-msft @pallavit I'm reluctant to smash this one into our upcoming release in a couple weeks. Initially when I spoke with @annelo-msft it seemed pretty straightforward. It was, actually, but as you can see required a lot of small changes. [...] I'm open to pushing this into our upcoming GA, but leaning toward punting it till after to get more usage on a beta release. The issue has been open for 3 years, so another few months (or so) probably won't hurt.

I am totally fine with that.

KrzysztofCwalina:
> Per offline discussion with @KrzysztofCwalina, @tg-msft, and @m-nash I am working on changing this to an overload of `GetSecret(..., RequestContext)` but the desirable change may not be possible. Previously, @KrzysztofCwalina or @tg-msft indicated I should make the existing signature e.g., `Response<KeyVaultSecret> GetSecret(string name, string version = null, CancellationToken cancellationToken = default)` the all-required signature, similar to what we do for overloads any other time. 

This would be a breaking change. I think we want these two signatures:
```csharp
public virtual Response<KeyVaultSecret> GetSecret(string name, string version = null, CancellationToken cancellationToken = default);
public virtual NullableResponse<KeyVaultSecret> GetSecret(string name, string version = null, RequestContext context);
```

heaths:
> > Per offline discussion with @KrzysztofCwalina, @tg-msft, and @m-nash I am working on changing this to an overload of `GetSecret(..., RequestContext)` but the desirable change may not be possible. Previously, @KrzysztofCwalina or @tg-msft indicated I should make the existing signature e.g., `Response<KeyVaultSecret> GetSecret(string name, string version = null, CancellationToken cancellationToken = default)` the all-required signature, similar to what we do for overloads any other time.
> 
> This would be a breaking change. I think we want these two signatures:
> 
> ```cs
> public virtual Response<KeyVaultSecret> GetSecret(string name, string version = null, CancellationToken cancellationToken = default);
> public virtual NullableResponse<KeyVaultSecret> GetSecret(string name, string version = null, RequestContext context);
> ```

The `context` would also have to optional, but I get what you're saying. Just that when we talked a month or two back, seems it was the `CancellationToken` signature that should've been optional but perhaps it wasn't known at the time it'd need a different return type.

That is something we'll have to keep in mind for any other bolt-on convenience methods.

sdk/keyvault/Azure.Security.KeyVault.Secrets/src/SecretClient.cs
```diff
@@ -140,7 +140,79 @@ public virtual Response<KeyVaultSecret> GetSecret(string name, string version =
...
+            Argument.AssertNotNullOrEmpty(name, nameof(name));
+
+            using DiagnosticScope scope = _pipeline.CreateScope($"{nameof(SecretClient)}.{nameof(GetSecret)}");
+            scope.AddAttribute("secret", name);
+            scope.AddAttribute("version", version);
+            scope.Start();
+
+            try
+            {
+                return await _pipeline.SendRequestAsync(RequestMethod.Get, context, () => new KeyVaultSecret(), SecretsPath, name, "/", version).ConfigureAwait(false);
```
annelo-msft, line 169: You will need to call the pipeline differently in order for RequestContext to function correctly.  I've added details here: https://github.com/Azure/azure-sdk-for-net/issues/35014#issuecomment-1476811424

sdk/keyvault/Azure.Security.KeyVault.Shared/src/KeyVaultPipeline.cs
```diff
@@ -69,12 +72,14 @@ public Request CreateRequest(RequestMethod method, Uri uri, bool appendApiVersio
...
-            return request;
+            return message;
         }
 
-        public Request CreateRequest(RequestMethod method, params string[] path)
+        public HttpMessage CreateMessage(RequestMethod method, RequestContext context, params string[] path)
         {
-            Request request = _pipeline.CreateRequest();
+            // See comment in CreateMessage overload above for future design consideration.
+            HttpMessage message = _pipeline.CreateMessage(context);
```
annelo-msft, line 81: The expectation is that the generated code will pass in a StatusCodeClassifier with a list of all the valid non-error status code, which it has obtained from the API spec.  Since you are writing this manually, you will need to do this by hand in order to satisfy the contract.  We can discuss whether this is a good design for hand-written code, but the APIs were not designed to work without a DPG client, thus the need to work within the spec on this.

```suggestion
            HttpMessage message = _pipeline.CreateMessage(context, new StatusCodeClassifier(stackalloc ushort[] { 200 }));
```

heaths: If the `ResponseClassifier` parameter is required as you suggest, why is the parameter optional via overload? Seems it shouldn't be an overload if I'm allowed to call it "wrong".

/cc @m-nash

annelo-msft: These APIs were designed to be called by generated code and are reasonably advanced.  I believe, though, that nothing should block you from making the same calls the generated code does manually.

m-nash:
Not saying what we have here isn't correct but I want to call out that this method type is a "new" thing that will either always remain hand written even after we migrate to tsp or we will need to add a new feature to the emitter / generator.

Right now we have 3 types of methods that get generated

1. Internal rest level operation responsible for creating and sending the httpmessage
2. public protocol method which returns Response and takes in RequestContent / RequestContext
3. public convenience method which returns Response<T> where T is a strongly typed model and takes in a strongly toped model for input as well and uses CancellationToken instead of RequestContext.

This is a 4th type which is a hybrid of 2 and 3.  If this is something we want to move past hand written only we will need to come up with a way to express this difference in client teaspoon and then have the generator create it.  Since client teaspoon is declarative this will be explicitly opted into and won't be generated otherwise.

pallavit:
> This is a 4th type which is a hybrid of 2 and 3. If this is something we want to move past hand written only we will need to come up with a way to express this difference in client teaspoon and then have the generator create it. Since client teaspoon is declarative this will be explicitly opted into and won't be generated otherwise.

Thanks for raising this. @tg-msft could you please weigh in here.

heaths:
Why not just treat this as an entirely hand-authored convenience method, if so desired? I think the main problem here - relating to #34988 and #35014, seemingly - is that the `CreateMessage(RequestContext)` method could copy both (and default the `ResponseClassifier` to the base class that seems to work for general cases e.g., 2xx means success). Can you see any problems relate to this approach? It's entirely within Azure.Core and doesn't depend on autorest.csharp, though it also shouldn't break autorest.csharp.

sdk/keyvault/Azure.Security.KeyVault.Secrets/src/SecretClient.cs
```diff
@@ -140,7 +140,79 @@ public virtual Response<KeyVaultSecret> GetSecret(string name, string version =
...
+        /// <param name="version">The version of the secret.</param>
+        /// <param name="context">A <see cref="RequestContext"/> controlling the request lifetime, error handling, and per-call pipeline policies.</param>
+        /// <returns>A response containing the <see cref="KeyVaultSecret"/> or <c>null</c> if not found.</returns>
+        /// <exception cref="ArgumentException"><paramref name="name"/> is an empty string.</exception>
+        /// <exception cref="ArgumentNullException"><paramref name="name"/> is null.</exception>
+        /// <exception cref="RequestFailedException">The server returned an error. See <see cref="Exception.Message"/> for details returned from the server.</exception>
+#pragma warning disable AZC0002 // DO ensure all service methods, both asynchronous and synchronous, take an optional CancellationToken parameter called cancellationToken.
+        public virtual async Task<NullableResponse<KeyVaultSecret>> GetSecretAsync(string name, string version, RequestContext context)
+#pragma warning restore AZC0002 // DO ensure all service methods, both asynchronous and synchronous, take an optional CancellationToken parameter called cancellationToken.
```
annelo-msft, line 158: Curious: is this warning suppression still needed?  I had thought we did some work to allow this type of method without an analyzer warning.

sdk/keyvault/Azure.Security.KeyVault.Shared/src/KeyVaultPipeline.cs
```diff
@@ -171,115 +190,135 @@ public Page<T> GetPage<T>(Uri firstPageUri, string nextLink, Func<T> itemFactory
             where TContent : IJsonSerializable
             where TResult : IJsonDeserializable
         {
-            using Request request = CreateRequest(method, path);
-            request.Content = RequestContent.Create(content.Serialize());
+            using HttpMessage message = CreateMessage(method, null, path);
```
annelo-msft, line 193: Does this call, passing in null for the RequestContext, have good test coverage?

sdk/keyvault/Azure.Security.KeyVault.Shared/src/KeyVaultPipeline.cs
```diff
@@ -171,115 +190,135 @@ public Page<T> GetPage<T>(Uri firstPageUri, string nextLink, Func<T> itemFactory
...
-                case 204:
-                    return response;
-                default:
-                    throw Diagnostics.CreateRequestFailedException(response);
+                throw Diagnostics.CreateRequestFailedException(message.Response);
             }
         }
+
+        private static ResponseClassifier s_defaultResponseClassifier;
+        private static ResponseClassifier DefaultResponseClassifier => s_defaultResponseClassifier ??= new StatusCodeClassifier(stackalloc ushort[]
```
annelo-msft, line 316: 👍

sdk/keyvault/Azure.Security.KeyVault.Secrets/CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 
 ### Bugs Fixed
 
+- Added `SecretClient.GetSecretIfExists` and `GetSecretIfExistsAsync` to avoid throwing or logging when a secret was not found. ([#25125](https://github.com/Azure/azure-sdk-for-net/issues/25125))
```
annelo-msft, line 11: Nit: is it a bug fix, or a new feature?

sdk/keyvault/Azure.Security.KeyVault.Secrets/samples/Sample4_GetSecretIfExists.md
```diff
@@ -0,0 +1,55 @@
...
+
+In the sample below, you can set `keyVaultUrl` based on an environment variable, configuration setting, or any way that works for your application.
+
+```C# Snippet:SecretsSample4SecretClient
+var client = new SecretClient(new Uri(keyVaultUrl), new DefaultAzureCredential());
+```
+
+## Get a secret
+
+To prevent throwing a `RequestFailedException` when a secret does not exist, or to alter other behaviors of a specific API call, you can create a `RequestContext` and pass that to the API call:
```
annelo-msft, line 21: Would it be worth linking to some of the docs on RequestContext here, since this may be a new concept for Azure SDK .NET customers?

annelo-msft (approved):

sdk/keyvault/Azure.Security.KeyVault.Secrets/src/SecretClient.cs
heaths: Yep. I wasn't able to compile successfully without it. Perhaps it's because `context` isn't optional and the validation error isn't that specific? Just a guess.

annelo-msft: Cool, thanks for the explanation!  I'll update my mental model.

sdk/keyvault/Azure.Security.KeyVault.Secrets/samples/Sample4_GetSecretIfExists.md
```diff
@@ -0,0 +1,58 @@
+# Get a secret without throwing if it is not defined
```
KrzysztofCwalina, line 1: is "defined" the right word here? should we say "not found"?

sdk/keyvault/Azure.Security.KeyVault.Secrets/samples/Sample4_GetSecretIfExists.md
```diff
@@ -0,0 +1,58 @@
...
+```
+
+## Get a secret
+
+To prevent throwing a `RequestFailedException` when a secret does not exist, or to alter other behaviors of a specific API call, you can create a `RequestContext` and pass that to the API call:
+
+```C# Snippet:SecretsSample4GetSecretIfExists
+// Do not treat HTTP 404 responses as errors.
+RequestContext context = new RequestContext();
+context.AddClassifier(404, false);
```
KrzysztofCwalina, line 26: Can we pass the parameters by name? It will be easier to understand the sample.

heaths:
@KrzysztofCwalina @tg-msft @m-nash where do we stand on this?

I still feel like making customers use System.Text.Json or DynamicData just to use the power of a `RequestContext` is overkill. If we make `Response<T>` a `NullableResponse<T>`, we can still provide convenience methods that take/return explicit types as C# developers would be more comfortable with. If customers need the raw response, they can still get it with `NullableResponse<T>.GetRawResponse().Content`, yes?

m-nash:
> @KrzysztofCwalina @tg-msft @m-nash where do we stand on this?
> 
> I still feel like making customers use System.Text.Json or DynamicData just to use the power of a `RequestContext` is overkill. If we make `Response<T>` a `NullableResponse<T>`, we can still provide convenience methods that take/return explicit types as C# developers would be more comfortable with. If customers need the raw response, they can still get it with `NullableResponse<T>.GetRawResponse().Content`, yes?
> 
> EDIT: We got another request today: #36942

I believe the [explicit casts](https://github.com/Azure/azure-sdk-for-net/blob/03d2c715f3954854e866f0aaa244039b6c0231dc/sdk/core/Azure.Core/samples/Serialization.md#using-explicit-cast) will be the answer here, but the exact public api for those isn't fully decided but hopefully by Azure.Core 34 we will be ready at which point we can follow that pattern here.

vcolin7:
In progress.

pallavit:
@heaths Has the story streamlined here? Can we tackle this as part of this or upcoming milestone?

heaths:
Asked our architects and DPG driver for clarification. It seemed they were open to some variation of what I have here currently.

PR #33782, opened 2023-01-31
Make sure backup/restore tests are non-parallel

heaths:
Get frequent conflicts when running live tests; though, I'm not sure how
much this will help because the RecordingTestBase itself is marked as
[NonParallelizable], which is an inheritted attribute. Won't hurt,
though. Could just be a service issue because, sometimes, an LRO may not
be completely done when it says it is and we're stressing the service
pretty hard (and rather impractically).

### DIFF
--- a/sdk/keyvault/Azure.Security.KeyVault.Administration/tests/BackupRestoreTestBase.cs
+++ b/sdk/keyvault/Azure.Security.KeyVault.Administration/tests/BackupRestoreTestBase.cs
@@ -11,6 +11,10 @@ using NUnit.Framework;
 
 namespace Azure.Security.KeyVault.Administration.Tests
 {
+    // Though RecordingTestBase is attributed [NonParallelizable] now, make sure these tests never run in parallel if that ever changes
+    // or an intermediate base class adds [Parallelizable] in the future.
+    //
+    // Note: CIs still build/test all assemblies in parallel, so MHSM Keys tests may still run simultaneously.
     [NonParallelizable]
     [IgnoreServiceError(404, "NotFound", Message = "The given jobId is not found", Reason = "Backup/restore tests have inherent concurrency issues")]
     [IgnoreServiceError(409, "Conflict", Message = "User triggered Restore operation is in progress", Reason = "Backup/restore tests have inherent concurrency issues")]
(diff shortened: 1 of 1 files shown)


sdk/keyvault/Azure.Security.KeyVault.Administration/tests/BackupRestoreTestBase.cs
```diff
@@ -11,6 +11,7 @@
 
 namespace Azure.Security.KeyVault.Administration.Tests
 {
+    [NonParallelizable]
     [IgnoreServiceError(404, "NotFound", Message = "The given jobId is not found", Reason = "Backup/restore tests have inherent concurrency issues")]
     [IgnoreServiceError(409, "Conflict", Message = "User triggered Restore operation is in progress", Reason = "Backup/restore tests have inherent concurrency issues")]
     public abstract class BackupRestoreTestBase : AdministrationTestBase
```
JoshLove-msft, line 17: This shouldn't have any impact for the reasons you mentioned in the description.

heaths: Likely not, but it does also make it more local in case something up the chain of base classes changes (including on my own). This is more of a correctness change. Yes, I doubt it'll actually fix the issue. Think of it more as "future proofing".

JoshLove-msft: Yeah, but it introduces ambiguity - others may be unsure whether they need a similar change.

heaths: If they want to make sure no future changes to base classes in or out of their control don't suddenly parallelize, why shouldn't they?

JoshLove-msft: I think the issue is that people looking at the code won't realize that this line does nothing. If they remove it and test they would expect it to run in parallel, but it still wouldn't because of the base class. I think a comment would at least help clear this up.

heaths: I'll add a comment. But there's always `git blame` where I also described it. ;)

JoshLove-msft: Thank for your humoring me.

JoshLove-msft (approved):

PR #33790, opened 2023-02-01
Workaround .NET on Windows Nano bug with Environment.GetFolderPath

NCarlsonMSFT:


sdk/identity/Azure.Identity/src/Credentials/VisualStudioCredential.cs
```diff
@@ -113,6 +108,26 @@ private async ValueTask<AccessToken> GetTokenImplAsync(TokenRequestContext reque
...
+        private static string GetTokenProviderPath()
+        {
+            string baseFolder;
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                baseFolder = Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData);
+                if (string.IsNullOrEmpty(baseFolder))
+                {
+                    // There is a known issue that Environment.GetFolderPath does not work on Windows Nano: https://github.com/dotnet/runtime/issues/21430
+                    baseFolder = Environment.GetEnvironmentVariable("LOCALAPPDATA");
```
christothes, line 120: Should we handle the case where this is also null?

NCarlsonMSFT: Great point! If this is null Path.Combine will throw. Would you prefer that this throw a specific exception about not being able to find the local app data folder, or continue the current behavior of using string.Empty so that it falls back to the working directory?

christothes: yeah, I think some specific error would be a bit better from a troubleshooting standpoint. Ideally we could link to our [troubleshooting doc](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/identity/Azure.Identity/TROUBLESHOOTING.md) as well so that we could add more content there if necessary.

NCarlsonMSFT: I've added an error message. I used the link https://aka.ms/azsdk/net/identity/vscredential/troubleshoot to match what I saw from the VSCode credential. But that aka.ms link does not exist yet. Do you need/want me to create it?

christothes: Sorry for the delay, We just added that aka link, so you should be able to use it now

NCarlsonMSFT:
@christothes I've added the error message, is there anything else missing?

christothes (approved):

PR #33820, opened 2023-02-01
[Schema Registry] Test Resources and Samples Updates

m-redding:
These changes:
- add additional namespaces and schema groups to the live test resources in order to resolve pipeline failures, in order to add these new namespaces to the environment, some of the tests needed to be re-recorded
- add samples for the new schema format additions, which are still in preview


sdk/schemaregistry/Azure.Data.SchemaRegistry/samples/Sample01_HelloWorld.md
```diff
@@ -0,0 +1,122 @@
+# Registering and retrieving Avro, Json, and custom schemas (preview)
```
m-redding, line 1: Should we note this as a preview?

sdk/schemaregistry/Azure.Data.SchemaRegistry/README.md
```diff
@@ -31,11 +31,11 @@ In order to interact with the Azure Schema Registry service, you'll need to crea
 
 #### Get credentials
 
-To acquire authenicated credentials and start interacting with Azure resources, please see the [quickstart guide here][quickstart_guide].
+To acquire authenticated credentials and start interacting with Azure resources, please see the [quickstart guide here][quickstart_guide].
 
 #### Get Event Hubs namespace hostname
 
-The simpliest way is to use the [Azure portal][azure_portal] and navigate to your Event Hubs namespace. From the Overview tab, you'll see `Host name`. Copy the value from this field.
+The simplest way is to use the [Azure portal][azure_portal] and navigate to your Event Hubs namespace. From the Overview tab, you'll see `Host name`. Copy the value from this field.
```
JoshLove-msft, line 38: 😆

sdk/schemaregistry/Azure.Data.SchemaRegistry/samples/Sample01_HelloWorld.md
JoshLove-msft: Yes, I think that is a good idea.

JoshLove-msft: Though Avro should not be listed as preview.

sdk/schemaregistry/test-resources.json
```diff
@@ -108,6 +129,159 @@
         "kafkaEnabled": true
       }
     },
+    {
+      "type": "Microsoft.EventHub/namespaces",
```
JoshLove-msft, line 133: Curious why we need a separate namespace for each schema type - I realize we already have this pattern, but wondering if it would be simpler to just have a single namespace containing all of the schema groups that we need, i.e. 2 for each type.

sdk/schemaregistry/Azure.Data.SchemaRegistry/tests/SchemaRegistryClientLiveTests.cs
```diff
@@ -23,22 +23,50 @@ public SchemaRegistryClientLiveTests(bool isAsync, SchemaRegistryClientOptions.S
...
+                    else
+                    {
+                        endpoint = (group == 1) ? TestEnvironment.SchemaRegistryEndpointAvro2021 : TestEnvironment.SchemaRegistryEndpointAvro2021_2;
+                    }
                     break;
                 case Custom:
-                    endpoint = TestEnvironment.SchemaRegistryEndpointCustom;
+                    if (_serviceVersion == SchemaRegistryClientOptions.ServiceVersion.V2022_10)
+                    {
+                        endpoint = (group == 1) ? TestEnvironment.SchemaRegistryEndpointCustom : TestEnvironment.SchemaRegistryEndpointCustom2;
```
JoshLove-msft, line 54: Do we actually need multiple endpoints per service version? I thought it was just that we would need 1 for each service version. But as mentioned in this [comment](https://github.com/Azure/azure-sdk-for-net/pull/33820/files?short_path=0b58de4#r1095243687), I think we can just use a single namespace and create 6 schema groups within it.

sdk/schemaregistry/test-resources.json
```diff
@@ -138,11 +312,101 @@
         "schemaType": "custom"
       }
     },
+    {
+      "type": "Microsoft.EventHub/namespaces/schemagroups",
```
JoshLove-msft, line 316: What do you think of using the [ARM library](https://www.nuget.org/packages/Azure.ResourceManager.EventHubs) within the live test class for the schema group creation/deletion rather than test-resources? I'm thinking it will be cumbersome to keep updating this template, the TestEnvironment file, _and_ the live test class whenever a new service version is introduced.

This would add a bit of overhead for running the live tests locally, but given the 25 schema group limit, it is not currently working well anyway so this should still be a net improvement.

sdk/schemaregistry/Azure.Data.SchemaRegistry/README.md
JoshLove-msft: Nice catch!

sdk/schemaregistry/Azure.Data.SchemaRegistry/samples/Sample01_HelloWorld.md
```diff
@@ -0,0 +1,124 @@
...
+In the current service version, the Schema Registry service only accepts and validates draft 3 Json schemas. Functionality to support additional drafts of Json will be added.
+
+- [Registering and retrieving Avro, Json, and custom schemas (preview)](#registering-and-retrieving-avro-json-and-custom-schemas-preview)
+    - [Register an Avro schema](#register-an-avro-schema)
+    - [Retrieve an Avro schema](#retrieve-an-avro-schema)
+    - [Register a Json schema](#register-a-json-schema)
+    - [Retrieve a Json schema](#retrieve-a-json-schema)
+    - [Register a custom schema](#register-a-custom-schema)
+    - [Retrieve a custom schema](#retrieve-a-custom-schema)
+  - [Contributing](#contributing)
```
JoshLove-msft, line 16: Does contributing belong in the samples? I think this goes in the readme.

sdk/schemaregistry/Azure.Data.SchemaRegistry/tests/Samples/Sample02_HelloWorld.cs
```diff
@@ -0,0 +1,166 @@
...
+using Azure.Core.TestFramework;
+using Azure.Identity;
+using NUnit.Framework;
+
+namespace Azure.Data.SchemaRegistry.Tests.Samples
+{
+    public class Sample02_HelloWorld : SamplesBase<SchemaRegistryClientTestEnvironment>
+    {
+#pragma warning disable IDE1006 // Naming Styles
+        private SchemaRegistryClient client;
```
JoshLove-msft, line 14: Interesting.. why do we need this field?

sdk/schemaregistry/test-resources.json
m-redding: I tried doing this initially but was running into issues when creating the resources, I'll try it again and try to dig into the error more.

sdk/schemaregistry/Azure.Data.SchemaRegistry/samples/Sample01_HelloWorld.md
```diff
@@ -0,0 +1,124 @@
+# Registering and retrieving Avro, Json, and custom schemas (preview)
+
+The following shows examples of how to use the basic functionality of the `SchemaRegistryClient` with each of the available schema format types. All of the client methods can be used with each format by passing in the proper 'SchemaFormat' value when needed.
+
+Each Event Hubs namespace can only accept one format of schemas.
```
JoshLove-msft, line 5: Hmm is this actually true? If so, it would explain why we needed separate endpoints for each format, but I find this to be a strange restriction.

sdk/schemaregistry/test-resources.json
m-redding: Well, I tried a lot of different things but keep getting the error:

"The deployment 'xxx' failed with error(s). Showing 1 out of 1 error(s). Status Message: SubCode=40300, Cannot create schema group azsdk_net_test_group_avro because no more than 1 groups is allowed. Currently there are 1 groups for the namespace"

So I think going with the ARM library as mentioned in the other comment, is probably the best option. I don't think using test resources is going to be the best developer experience. The one group per namespace and 25 schemas per group is definitely very limiting.

sdk/schemaregistry/Azure.Data.SchemaRegistry/samples/Sample01_HelloWorld.md
m-redding: I guess the real restriction is actually on the groups, but since only one group is allowed per namespace it's kind of the same thing, but I will change this to be groups to align better.

sdk/schemaregistry/Azure.Data.SchemaRegistry/tests/Samples/Sample02_HelloWorld.cs
m-redding: This is so that the different snippets can work off of the same client.

sdk/schemaregistry/Azure.Data.SchemaRegistry/samples/Sample01_HelloWorld.md
JoshLove-msft: Ah, that is unfortunate. I guess we will need to stick to creating the namespaces in test-resources.json then.

sdk/schemaregistry/Azure.Data.SchemaRegistry/tests/Properties/Resources.resx
```diff
@@ -0,0 +1,101 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<root>
```
JoshLove-msft, line 2: What is the deal with this file?

sdk/schemaregistry/Azure.Data.SchemaRegistry/tests/SchemaRegistryClientLiveTests.cs
```diff
@@ -23,27 +24,43 @@ public SchemaRegistryClientLiveTests(bool isAsync, SchemaRegistryClientOptions.S
...
-                    endpoint = TestEnvironment.SchemaRegistryEndpointJson;
-                    break;
-                case Avro:
-                    endpoint = TestEnvironment.SchemaRegistryEndpointAvro;
-                    break;
-                case Custom:
-                    endpoint = TestEnvironment.SchemaRegistryEndpointCustom;
-                    break;
+                case (Avro, 1):
+                    return TestEnvironment.SchemaRegistryEndpointAvro1;
```
JoshLove-msft, line 32: I don't quite understand how we went from having one Avro endpoint to needing 4 just from adding the new service version. Are we really creating more than 25 schemas per service version?

sdk/schemaregistry/Azure.Data.SchemaRegistry/tests/SchemaRegistryClientLiveTests.cs
```diff
@@ -23,27 +24,43 @@ public SchemaRegistryClientLiveTests(bool isAsync, SchemaRegistryClientOptions.S
             _serviceVersion = version;
         }
 
-        private SchemaRegistryClient CreateClient(string format)
+        private string GetEndpoint(string format, int group)
```
JoshLove-msft, line 27: Ideally we'd be able to remove the group param and condition this on service version. But that is dependent on us getting to the point where we need one endpoint per format per service version, rather than what we have currently which appears to be 2.

sdk/schemaregistry/Azure.Data.SchemaRegistry/tests/SchemaRegistryClientLiveTests.cs
m-redding: Yeah, I'm honestly not sure either, I tried just adding one but then was seeing local failures when running the tests in live mode. But also when I run them locally I have 279 tests, between the three .NET versions, the different API versions, and async/sync

JoshLove-msft: I would focus on one platform. Sure the local experience won't be great, but at least for the live test pipeline it will work.

m-redding: Do the live test pipelines only run one version of .NET? If so, then I think we could just go with one for each service version

m-redding: Yeah I can look into adding that, I have been deleting them in the portal in between runs (unfortunately, generating new test resources doesn't actually delete the schemas :/ )

sdk/schemaregistry/Azure.Data.SchemaRegistry/tests/Properties/Resources.resx
m-redding: I don't know, I think it got generated by something. I'll delete it

sdk/schemaregistry/Azure.Data.SchemaRegistry/tests/SchemaRegistryClientLiveTests.cs
m-redding: > Another thing we could do would be to automatically delete the schema groups at the end of each test run so that we can make things work locally.

I think for now it's fine, but I can add it in the future if it gets annoying

sdk/schemaregistry/Microsoft.Azure.Data.SchemaRegistry.ApacheAvro/src/Microsoft.Azure.Data.SchemaRegistry.ApacheAvro.csproj
```diff
@@ -11,11 +11,11 @@
 
   <ItemGroup>
     <PackageReference Include="Apache.Avro" />
-    <!--<PackageReference Include="Azure.Data.SchemaRegistry" /> -->
+    <!-- <PackageReference Include="Azure.Data.SchemaRegistry" /> -->
```
JoshLove-msft, line 14: A lot of whitespace moving around here 😆

JoshLove-msft (approved):
Looks great! Thanks for sticking with this one - the whole team learned a lot as a result. I just kicked off the live test run.

sdk/schemaregistry/Microsoft.Azure.Data.SchemaRegistry.ApacheAvro/src/Microsoft.Azure.Data.SchemaRegistry.ApacheAvro.csproj
m-redding: Yeah I tried to switch the package ref back but it didn't work so I made a different PR that I will circle back to after this

https://github.com/Azure/azure-sdk-for-net/pull/33872

There is some weird stuff happening over there

JoshLove-msft: Looks like the Avro encoder tests are failing - https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2173742&view=ms.vss-test-web.build-test-results-tab&runId=39270351&resultId=100122&paneView=debug

PR #33821, opened 2023-02-01
Azure OpenAI test management update

trrwilson:
The primary goal of this PR is to enact a small set of functional changes to the tests for Azure OpenAI to facilitate maintenance by my team. We now have an appropriately configured service principal and subscription for the test framework to interact with, and the code in this branch works seamlessly against it for recording -- all tests have fresh recording documents to take the first opportunity for that.

Structural/naming changes were made in the course of me learning how things were working and -- I hope -- help make it a bit more self-documenting for future readers as clueless as I was. Feedback greatly welcomed and appreciated!


sdk/openai/Azure.AI.OpenAI/tests/OpenAITestBase.cs
```diff
@@ -16,138 +15,179 @@ namespace Azure.AI.OpenAI.Tests
...
             }
-
-            if ((Mode == RecordedTestMode.Live || Mode == RecordedTestMode.Record) && (DeploymentId == null || EmbeddingsDeploymentId == null))
+            else if (Mode == RecordedTestMode.Live || Mode == RecordedTestMode.Record)
             {
+                TestEnvironment.ThrowIfCannotDeploy();
+
                 lock (_deploymentIdLock)
                 {
-                    if (DeploymentId == null || EmbeddingsDeploymentId == null)
```
m-nash, line 50: I think we still want a 2nd if here to skip if another thread beat us inside the lock and took care of set up.  It could be that running the code twice is idempotent so this is just a performance enhancement, but its a good practice in general.

m-nash (approved):

sdk/openai/Azure.AI.OpenAI/tests/OpenAITestBase.cs
trrwilson: Great point; I flipped this back around to move the check inside the lock.

trrwilson: I think this should be good to go; per discussion, the `else -> lock -> if (use service)` chain should be just about equivalent to the top-level `else if` in this case, given we're already handling Playback earlier and don't expect anything other than `Live` and `Record` to proceed.

jpalvarezl (approved):

PR #33829, opened 2023-02-02
Add docs on running live tests serially

heaths:
- General clean-up of markdownlint issues
- Add section on serial live tests

### DIFF
--- a/sdk/core/Azure.Core.TestFramework/README.md
+++ b/sdk/core/Azure.Core.TestFramework/README.md
@@ -1,4 +1,5 @@
 # Using the TestFramework
+
 To start using the Test Framework, add a project reference using the alias `AzureCoreTestFramework` into your test `.csproj`:
 
 ``` xml
@@ -9,8 +10,8 @@ To start using the Test Framework, add a project reference using the alias `Azur
 ...
 
 </Project>
-
 ```
+
 As an example, see the [Template](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/template/Azure.Template/tests/Azure.Template.Tests.csproj#L15) project.
 
 ## Sync-async tests
@@ -30,7 +31,6 @@ public class ConfigurationLiveTests: ClientTestBase
                 ...,
                 InstrumentClientOptions(
                     new ConfigurationClientClientOptions())));
-    }
 
     public async Task DeleteSettingNotFound()
     {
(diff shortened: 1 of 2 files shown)


heaths:
I included some general clean-up for markdownlint errors as recommended by the Microsoft Learn Extension Pack for Visual Studio Code (same rules as required for https://learn.microsoft.com, which makes spotting actual errors/warnings easier!).

JoshLove-msft (approved):

sdk/core/Azure.Core.TestFramework/README.md
```diff
@@ -147,31 +146,53 @@ public class AppConfigurationTestEnvironment : TestEnvironment
 }
 ```
 
+### Running live tests serially
+
+By default, NUnit does not run tests within each assembly in parallel, but this be [configured](https://docs.nunit.org/articles/nunit/technical-notes/usage/Framework-Parallel-Test-Execution.html).
+Especially for unit tests, this is often desirable; however, live and [recorded tests](#recorded-tests) may run into some issues. Thus, by default, the `RecordedTestBase` described below is attributed
+as `[NonParallelizable]`.
+
+However, when projects are built and tested in CIs, all projects are testing in parallel. This means, for example, you can have two or more assemblies running tests such as one backing up or restoring
```
jsquire, line 155:
```suggestion
However, when projects are built and tested in CI pipelines, all projects are testing in parallel. This means, for example, you can have two or more assemblies running tests such as one backing up or restoring
```

jsquire (approved):

christothes (approved):

sdk/core/Azure.Core.TestFramework/README.md
```diff
@@ -147,31 +146,53 @@ public class AppConfigurationTestEnvironment : TestEnvironment
 }
 ```
 
+### Running live tests serially
```
annelo-msft, line 149: Thanks for adding this - I've been wanting to know how to do this for the ACR library!

pallavit (approved):

sdk/core/Azure.Core.TestFramework/README.md
JoshLove-msft: +1 - I'm going to add this for Service Bus which will hopefully address some of the flakiness in the Functions extensions tests.

sdk/core/Azure.Core.TestFramework/README.md
heaths: CIs - Continuous Integration - already imply some form of pipeline, but I don't see harm in taking this if it helps clear it up.

PR #33833, opened 2023-02-02
Azure OpenAI: address custom deserialization issue for Choice.FinishReason

trrwilson:
This library uses a custom deserializer for this part of the models; for the `"finish_reason"` response field, that custom logic unfortunately checks for `"finishReason"` (camelCased vs. snake_cased) and thus never populates the result's `Choice` object appropriately.

Since it's custom, the fix is trivial. Also added a test to verify it; existing recordings already have the right data and an A/B comparison of the deserialization fix verifies that this makes it work.


jsquire:
Thank you for your contribution, @trrwilson!

@m-nash:  Would you please take point on reviewing?

trrwilson:
For the CI failures: test records are updated in !33821 and should address the failures seen here. I'll rebase this branch as soon as the earlier PR is in.

jpalvarezl (approved):

m-nash:
Change looks good but there are CI failures still @trrwilson

trrwilson:
Thanks, @m-nash -- that caught an honest, overly restrictive assumption! Appropriately updated the test and re-recorded the results for good measure. Everything passes locally; fingers crossed CI agrees.

trrwilson:
And CI did agree! #sign-off, ready to merge :) Thank you!

m-nash (approved):

PR #33855, opened 2023-02-02
Storage `PartitionedUploader` supports `BinaryData`

jaschrep-msft:
Client APIs that upload via `PartitionedUploader` accept either `Stream` or `BinaryData` as input. But the uploader itself only accepted streams, wrapping the `BinaryData` in one under the hood. This leads to a lot of extra buffering and copying on data that is already referenceable in memory.

This abstracts the main logic of PartitionedUploader to handle content without knowing the type. Public APIs on the type now accept either Stream or BinaryData, injecting the appropriate data partitioning and upload behaviors into the main logic.


jaschrep-msft:
> Still need to replicate existing Stream tests as BinaryData versions

Binary data tests already exist extensively in blob tests. Added some low-level tests in common to asserts delegates are called as expected.

amnguye (approved):

PR #33857, opened 2023-02-02
[ResourceManager] Bump resources API version to 2022-09-01

### LINKED ISSUE #33050: Generate SDK for new API version for resources/Azure.ResourceManager.Resources SDK
### Library name

Azure.ResourceManager.Resources

We require the SDK team's help in generating the SDK for the new [2022-09-01](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/resources/resource-manager/Microsoft.Resources/stable/2022-09-01/resources.json) API version for resources.json. When we attempted to generate the new SDK by updating the API version in the [autorest](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/resources/Azure.ResourceManager.Resources/src/autorest.md?plain=1#L333) config file, we ran into multiple DuplicateSchema errors and were unable to resolve them.

fengzhou-msft:
fixes https://github.com/Azure/azure-sdk-for-net/issues/33050


sdk/resourcemanager/Azure.ResourceManager/tests/Scenario/SubscriptionOperationsTests.cs
```diff
@@ -182,6 +182,7 @@ await foreach (var feature in subscription.GetFeaturesAsync())
         }
 
         [RecordedTest]
+        [Ignore("Re-record of this test will cause github not able to show file diffs and run CI")]
         public async Task ValidateResourceInRestApi()
```
fengzhou-msft, line 186: The test passed locally but checking in 2 re-recorded files with more tahn 400 thousands diffs each caused CI unable to be triggered. It might be a github glitch, but I prefer to add back this test in another PR. We probably need to remove the recordings in one PR, then reenable the test and add back the recordings in another PR if re-recording still cause CI not to run.

sdk/resourcemanager/Azure.ResourceManager/src/Resources/Custom/RestOperations/ResourcesRestOperations.cs
```diff
@@ -20,22 +20,22 @@ namespace Azure.ResourceManager.Resources
 {
```
m-nash, line 20: we can address in a future PR, but do we still need this restoperations in custom code?

m-nash (approved):

sdk/resourcemanager/Azure.ResourceManager/src/Resources/Custom/RestOperations/ResourcesRestOperations.cs
fengzhou-msft: The main blocker is the `apiVersion` parameter for generic resource operations not correctly generated as a method parameter. I filed an issue to track it: https://github.com/Azure/autorest.csharp/issues/3094.

Issue #33697, opened 2023-01-26
[FEATURE REQ] Azure Function EventHub allow passing of the PartitionKey in the EventData ctor


fuzzlebuck:
### Library name

Azure.Messaging.EventHubs

### Please describe the feature.

It would be great if could pass a PartitionKey while creating a new EventData object.

This would allow the use of user defined partition keys when adding message via the EventHub output binding's `IAsyncCollector.AddAsync` method.

This is desirable over the use of `EventHubProducerClient.SendAsync` due to the fact that `IAsyncCollector.AddAsync` will queue the message and batch behind the scenes in parallel, whereas `EventHubProducerClient.SendAsync` needs to be awaited on each send of a single message or a batch.

The two mechanisms behave quite differently from each other and we've seen large performance gains while using the `IAsyncCollector` and `IAsyncCollector.Flush` compared to `EventHubProducerClient.SendAsync`.

```cs
IAsyncCollector collector = from output binding
EventData message = new EventData(data);
await collector.AddAsync(message);      <-- returns immediately and is sent in parallel
...
more stuff
...
await collector.Flush();      <-- waits for remaining messages to send
```
vs
```cs
EventHubProducerClient client = from output binding
EventData message = new EventData(data);
await client.SendAsync(message);      <-- waits for the send operation before returning
```

This is similar to some of the comments in https://github.com/Azure/azure-sdk-for-net/issues/28245 but it's not the same as the issue itself.

We've done some testing using the `EventHubsModelFactory.EventData` testing method and verified the partition key when set via this method appears to be used correctly by `IAsyncCollector` and the events do in fact send and arrive downstream in the correct batches with the set partition.

Clearly we can't use this testing and mocking method for production though. 

Or maybe there's another way to use the `IAsyncCollector`?

Thanks :-)

jsquire:
Hi @fuzzlebuck.   Thank you for reaching out and we regret that you're experiencing difficulties.    The `PartitionKey` property of `EventData` is a broker-owned field that is populated only for events read from Event Hubs.   It is intentionally not possible to set a partition key on an event created otherwise.

The implementation of the async collector appears to be incorrect, relying on what was a bug in the previous generation of the Event Hubs client library.  Because of this, partition keys are unable to be used with the collector, despite it attempting to manage batches correctly to do so.   To support this will require a new `AddAsync` overload that accepts a partition key alongside the event.

[...]

Issue #28245, opened 2018-03-26
.NET Core v2 Azure Function EventHub PartitionKey doesn't seem to work


josh-endries:
I'm unable to effectively set the PartitionKey property when using an Azure Function v2 (.NET Core).

Firstly, I am unsure *how* I am correctly supposed to set the PartitionKey property. I initially used the new Microsoft.Azure.WebJobs.Extensions.EventHubs package and its Microsoft.Azure.EventHubs.EventData class, however there is no PartitionKey property on that class. There is a PartitionKey property in the SystemProperties member on EventData, however SystemProperties and its PartitionKey property are both read-only, so I can't figure out how to set it from that package. Using the "old" Microsoft.ServiceBus.Messaging.EventData class, I can set PartitionKey on the object, however it doesn't seem to have any effect.

I'm assuming that I am doing something wrong but I can't figure out what. I am expecting that when I set the same PartitionKey value, it gets hashed and modded or something and should cause the message to be assigned to the same partition every time (as described in the documentation).

This is the Function code (using the "old" EventData class):

    public static class Function1
    {
        [FunctionName("Function1")]
        public static async Task Run(
            [TimerTrigger("0 */1 * * * *")] TimerInfo myTimer,
            [EventHub("events", Connection = "ConnectionString")] IAsyncCollector<EventData> results,
            CancellationToken token,
            TraceWriter log
        )
        {
            var ed1 = new EventData(Encoding.UTF8.GetBytes("test"));
            ed1.PartitionKey = "0";
            await results.AddAsync(ed1);
        }
    }

I've tried various partition keys (non-numeric, long, short, ...) but that didn't make a difference. The function runs every minute, es expected, but this is what I see in my client, which just dumps out the metadata:

![untitled](https://user-images.githubusercontent.com/1500571/37915204-fc669ad8-30e6-11e8-9c32-c5c6ae8a0e67.png)

This shows two Function invocations a minute apart. Note that the first line of each list item includes the partition on which the message was received, and that it alternates from partition 1 to partition 0, even though the PartitionKey property is always "0". If I change PartitionKey to "1", it updates in the data and stays consistent, but the reader-partition reporting the event keeps alternating the same as it does in the image. I would expect these partition IDs to remain the same if the messages were received on the same partition.

This is the code that displays the messages:

    public async Task ProcessEventsAsync(PartitionContext context, IEnumerable<EventData> messages) {
        await processor.ProcessEventsAsync(context, messages);

[...]

PR #33858, opened 2023-02-02
[Event Hubs] Allow partition key use in Functions async collector

jsquire:
# Summary

The focus of these changes is to allow for partition keys to be used when binding to the async collector to publish events.  In the original v1.x version of Functions, which used the track-zero Event Hubs library, the implementation relied on a broker-owned property of `EventData` to be set, which was possible only due to the approach used for enabling unit testing.  Since the bindings moved to the track-one Event Hubs library which disallowed setting the broker-owned property, it is not been possible to use the collector with a partition key.

In addition to fixing a bug with batch creation, these changes expose the `EventHubAsyncCollector` and allow binding to it, exposing an `AddAsync` overload specific to Event Hubs which accepts a partition key.  This enables events to be logically grouped together, ensuring they are assigned to the same partition.  Previously, binding was restricted to the `IAsyncCollector<T>` interface, which only accepted an `EventData` instance leaving no way to specify a partition key.  As a result, all events were assigned partitions via round-robin.   

# References and Related

- [[FEATURE REQ] Azure Function EventHub allow passing of the PartitionKey in the EventData ctor (#33697)](https://github.com/Azure/azure-sdk-for-net/issues/33697)
- [.NET Core v2 Azure Function EventHub PartitionKey doesn't seem to work (#28245)](https://github.com/Azure/azure-sdk-for-net/issues/28245)


sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Triggers/EventHubAsyncCollector.cs
```diff
@@ -12,21 +13,21 @@
 namespace Microsoft.Azure.WebJobs.EventHubs
 {
     /// <summary>
-    /// Core object to send events to EventHub.
-    /// Any user parameter that sends EventHub events will eventually get bound to this object.
-    /// This will queue events and send in batches, also keeping under the 1024kb event hub limit per batch.
+    /// This collector allows events to be published to Event Hubs asynchronously in the background.
     /// </summary>
-    internal class EventHubAsyncCollector : IAsyncCollector<EventData>, IDisposable
+    public class EventHubAsyncCollector : IAsyncCollector<EventData>, IDisposable
```
JoshLove-msft, line 18: An alternative would be to define a new model type that contains an EventData and PartitionKey property. I think this is probably cleaner, but we may want to run it by the Functions folks as I believe this is a unique pattern we are establishing here.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/api/Microsoft.Azure.WebJobs.Extensions.EventHubs.netstandard2.0.cs
```diff
@@ -20,6 +20,15 @@ public sealed partial class EventHubTriggerAttribute : System.Attribute
 }
 namespace Microsoft.Azure.WebJobs.EventHubs
 {
+    public partial class EventHubAsyncCollector : Microsoft.Azure.WebJobs.IAsyncCollector<Azure.Messaging.EventHubs.EventData>, System.IDisposable
```
JoshLove-msft, line 23: Who is responsible for disposing this?

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Triggers/EventHubAsyncCollector.cs
JoshLove-msft: Well either way we are introducing a new type that users will need to somehow discover, so docs will be important here. But I agree that your approach is probably better for users.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Triggers/EventHubAsyncCollector.cs
```diff
@@ -38,19 +39,40 @@ public EventHubAsyncCollector(IEventHubProducerClient client)
...
+        public virtual async Task AddAsync(EventData item, string partitionKey, CancellationToken cancellationToken = default(CancellationToken))
         {
             if (item == null)
             {
                 throw new ArgumentNullException(nameof(item));
             }
 
-            string key = item.PartitionKey ?? string.Empty;
+            // If the partition key has no value, set it to null.
+            if (partitionKey is { Length: 0 })
```
JoshLove-msft, line 70: Interesting.. why do we need to set to null if we are just resetting to empty on L75?

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/api/Microsoft.Azure.WebJobs.Extensions.EventHubs.netstandard2.0.cs
jsquire: The same party that was responsible before we made it public.    We didn't alter the contract, just the visibility for the concrete type versus just the interface projection.   My assumption is that it's part of the infrastructure's DI management.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Triggers/EventHubAsyncCollector.cs
jsquire: I don't see a reason to prefer the interface over this, so I converted all samples over so that we are demonstrating a unified approach rather than positioning as "just in this specific scenario"

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Triggers/EventHubAsyncCollector.cs
jsquire: That is an excellent question to which I have no answer.   I assume that I was doing something with the null at one point - but it sure doesn't add any value now.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/api/Microsoft.Azure.WebJobs.Extensions.EventHubs.netstandard2.0.cs
JoshLove-msft: You are probably right, but I'm not certain that this wasn't an existing bug. My other concern is that users may think they need to dispose this type, which I assume would break their app.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Triggers/EventHubAsyncCollector.cs
jsquire: Spoke too soon.   We need to have `partitionKey` null (and not empty) for L86 when we create the batch, as an empty string is a valid partition key where `null` signals that no key is present.   The `key` is used as a dictionary key, which must be non-null.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/api/Microsoft.Azure.WebJobs.Extensions.EventHubs.netstandard2.0.cs
JoshLove-msft: Would a concurrent dictionary allow us to avoid making this Disposable?

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Triggers/EventHubAsyncCollector.cs
JoshLove-msft: Sounds good - I think we will also need to update these docs - https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-event-hubs-output?tabs=in-process%2Cfunctionsv2%2Cextensionv5&pivots=programming-language-csharp#example

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/api/Microsoft.Azure.WebJobs.Extensions.EventHubs.netstandard2.0.cs
jsquire: I don't believe so.   The batch needs to stay unchanged between multiple stages in order to not lose events.   Under the semaphore, the following takes place:

- Create if not exists
- Attempt to add event to it
- If at capacity, pop it for send
- Create a new batch
- Add the current item

That said, it looks like the last two steps are not done, and the `item` that was added when the batch moves across the threshold gets dropped on the floor.   I need to poke at that to be sure that I'm not mistaken - but I think there's a bug that's been here.

jsquire: Nope, no bug.   I missed the `while(true)` - I added a test to validate as well.

jsquire: @JoshLove-msft :  Would it help to implement `IDisposable` explicitly so that `Dispose` doesn't appear for most cases?

JoshLove-msft: That is probably a good idea.

jsquire: Moving the binding to an extension method off the interface makes this moot as the type is internal again.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/CHANGELOG.md
```diff
@@ -4,10 +4,14 @@
 
 ### Features Added
 
+- Added the ability to bind to the `EventHubAsyncCollector` for output, which supports an overload allowing a partition key to be specified.
```
jsquire, line 7: Note to self:  Reword this if we decide to move forward with the extension method.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/EventHubExtensionConfigProvider.cs
```diff
@@ -95,7 +96,7 @@ internal static void LogExceptionReceivedEvent(ExceptionReceivedEventArgs e, ILo
             Utility.LogException(e.Exception, message, logger);
         }
 
-        private IAsyncCollector<EventData> BuildFromAttribute(EventHubAttribute attribute)
+        private EventHubAsyncCollector BuildFromAttribute(EventHubAttribute attribute)
```
jsquire, line 99: Note to self:  Revert this if we decide to move forward with the extension method

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/api/Microsoft.Azure.WebJobs.Extensions.EventHubs.netstandard2.0.cs
```diff
@@ -17,6 +17,10 @@ public sealed partial class EventHubTriggerAttribute : System.Attribute
         public string ConsumerGroup { get { throw null; } set { } }
         public string EventHubName { get { throw null; } }
     }
+    public static partial class IAsyncCollectorExtensions
+    {
+        public static System.Threading.Tasks.Task AddAsync(this Microsoft.Azure.WebJobs.IAsyncCollector<Azure.Messaging.EventHubs.EventData> instance, Azure.Messaging.EventHubs.EventData item, string partitionKey, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
```
JoshLove-msft, line 22: nit: should we call the param eventData to match the EventDataBatch from the library?

JoshLove-msft (approved):

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/api/Microsoft.Azure.WebJobs.Extensions.EventHubs.netstandard2.0.cs
jsquire: I called it "item" to match the existing `IAsyncCollector<T>` nomenclature.

jsquire: Updated.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Triggers/EventHubAsyncCollector.cs
jsquire: I'm going to create an issue to follow-up on this one and get these docs updated once we release.

jsquire: Tracking via #33920.

PR #33862, opened 2023-02-02
Set LangVersion to 11

AlexanderSher:


weshaggard (commented):
I don't have any issue with this change but I'd like to get folks from the .NET team to agree and review this.

heaths (changes requested):
Given language versions are basically in lock-step with .NET versions now, this is effectively the same as "latest". (See <https://learn.microsoft.com/dotnet/csharp/language-reference/configure-language-version#defaults>). That said, there was pushback on changing this to anything other than a fixed version. This should probably just be "11.0".

AlexanderSher:
> Given language versions are basically in lock-step with .NET versions now, this is effectively the same as "latest". (See https://learn.microsoft.com/dotnet/csharp/language-reference/configure-language-version#defaults). That said, there was pushback on changing this to anything other than a fixed version. This should probably just be "11.0".

I'm fine with either one, but I agree that a fixed version is more solid (I don't expect any breaking changes in the language, but who knows...)

sdk/core/Azure.Core/src/RequestFailedException.cs
```diff
@@ -183,7 +183,8 @@ internal static (string FormattedError, string? ErrorCode, IDictionary<string, s
             foreach (HttpHeader responseHeader in response.Headers)
             {
                 string headerValue = response.Sanitizer.SanitizeHeader(responseHeader.Name, responseHeader.Value);
-                messageBuilder.AppendLine($"{responseHeader.Name}: {headerValue}");
```
christothes, line 186: Are these non- `LangVersion` change required for this PR?

AlexanderSher: No, those are fixes for breaking changes

AlexanderSher:
> Given language versions are basically in lock-step with .NET versions now, this is effectively the same as "latest". (See https://learn.microsoft.com/dotnet/csharp/language-reference/configure-language-version#defaults). That said, there was pushback on changing this to anything other than a fixed version. This should probably just be "11.0".

Changed to 11, updated the name of PR.

jsquire:
Thanks for opening this, @AlexanderSher.  

We had discussed this update as a team last week and agreed that we would like to move to latest.   As Heath mentioned, the consensus was that we wanted to be explicit and deliberate about rolling forward in the future, so an explicit version is preferred to "latest."   Beyond that, things look good to me.

jsquire (approved):

sdk/core/Azure.Core/src/ResourceIdentifier.cs
```diff
@@ -278,7 +278,7 @@ private static ResourceType ChooseResourceType(ReadOnlySpan<char> resourceTypeNa
             return "Invalid resource id.";
         }
 
-        private static ReadOnlySpan<char> PopNextWord(ref ReadOnlySpan<char> remaining)
+        private static ReadOnlySpan<char> PopNextWord(scoped ref ReadOnlySpan<char> remaining)
```
heaths, line 281: I must have missed the announcement on this. Having read about it now, seems a bit like rust lifetime parameters...but simple.

heaths (approved):

heaths:
@AlexanderSher rebase and main and force push to eliminate those build errors. I ignored a failing (never succeeded, actually) test and filed a bug to get it fixed.

Issue #20431, opened 2021-04-15
Update Snippet replacement syntax


pakrym:
We are switching from the `//@@` and `/*@@*/` syntax for line replacement in snippets to a more C#-native `#if SNIPPET .. #endif` preprocessor directive.

Please switch your samples to use the new syntax. Example migration: https://github.com/Azure/azure-sdk-for-net/pull/20391/files#diff-988626950b27d8a9a30cfe0d394aba056f460eeadfd350f5f8e7b894f63c6c0d

- [ ] #33874 
- [ ] #33873 
- [x] #33876 
- [ ] #33875 

Before:
``` C#
 /*@@*/string workspaceId = TestEnvironment.WorkspaceId;
 //@@string workspaceId = "<workspace_id>";
```

After:

``` C#
#if SNIPPET
            string workspaceId = "<workspace_id>";
#else
            string workspaceId = TestEnvironment.WorkspaceId;
#endif
```

## 🍏  appconfiguration
Owners:
@annelo-msft
@AlexanderSher

Files:

    sdk/appconfiguration/Azure.Data.AppConfiguration/tests/samples/Sample10_SecretReference.cs
    sdk/appconfiguration/Azure.Data.AppConfiguration/tests/samples/SampleSnippets.cs

## attestation
Owners:
@anilba06

Files:

    sdk/attestation/Azure.Security.Attestation/tests/Samples/AttestationServiceAttestationSamples.cs

## communication
Owners:
@acsdevx-msft

Files:

    sdk/communication/Azure.Communication.Chat/tests/samples/Sample2_MessagingOperations.cs
    sdk/communication/Azure.Communication.Chat/tests/samples/Sample4_KeyConcepts.cs
    sdk/communication/Azure.Communication.Common/tests/Identity/CommunicationTokenCredentialTest.cs
    sdk/communication/Azure.Communication.Identity/tests/samples/Sample1_CommunicationIdentityClient.cs
    sdk/communication/Azure.Communication.PhoneNumbers/tests/samples/Sample_PhoneNumbersClient.cs
    sdk/communication/Azure.Communication.Sms/tests/SmsClientLiveTestBase.cs
    sdk/communication/Azure.Communication.Sms/tests/SmsClientLiveTests.cs
    sdk/communication/Azure.Communication.Sms/tests/samples/Sample1_SmsClient.cs

## 🍏 core
Owners:
@pakrym
@KrzysztofCwalina

Files:

[...]

PR #33878, opened 2023-02-03
Build snippets during Analyze job by default

### LINKED ISSUE #33877: Change default BuildSnippets to check all snippets are used
The CI pipeline can check that all snippets are used to make sure people don't accidentally forget to use any they defined (and also makes sure any referenced are defined), but the default is turned off currently. Many SDKs have already opted in to do this check during the Analyze job, but I think we should enable that by default.

Running `eng/scripts/Update-Snippets.ps1 * -StrictMode`, I find the following SDKs still have some unreferenced snippets:

- [ ] communication
- [ ] dns
- [ ] formrecognizer
- [ ] iot
- [ ] servicebus

We should make sure that a new instance of a DPG template does not run afoul of this error as well, but given the command above didn't find any issues it should be fine. We should create an instance though and double check.

Client library authors should run `Update-Snippets.ps1 {ServiceDirectory}` to check compliance, and to make sure any README or sample markdowns are properly updated (which the Analyze step will also do or fail).

Also see related snippet clean-up issue #20431

heaths:
Resolves #33877

### DIFF
--- a/sdk/dns/ci.mgmt.yml
+++ b/sdk/dns/ci.mgmt.yml
@@ -18,6 +18,7 @@ extends:
   template: /eng/pipelines/templates/stages/archetype-sdk-client.yml
   parameters:
     ServiceDirectory: dns
+    BuildSnippets: true
     LimitForPullRequest: true
     Artifacts:
     - name: Azure.ResourceManager.Dns
--- a/sdk/dns/ci.mgmt.yml
+++ b/sdk/dns/ci.mgmt.yml
@@ -18,7 +18,7 @@ extends:
   template: /eng/pipelines/templates/stages/archetype-sdk-client.yml
   parameters:
     ServiceDirectory: dns
-    BuildSnippets: true
+    BuildSnippets: false
     LimitForPullRequest: true
     Artifacts:
     - name: Azure.ResourceManager.Dns
--- a/sdk/iot/ci.yml
+++ b/sdk/iot/ci.yml
@@ -20,7 +20,7 @@ extends:
   template: /eng/pipelines/templates/stages/archetype-sdk-client.yml
   parameters:
     ServiceDirectory: iot
-    BuildSnippets: true
+    BuildSnippets: false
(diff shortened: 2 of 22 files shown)


JoshLove-msft (approved):

sdk/iot/ci.yml
```diff
@@ -20,8 +20,9 @@ extends:
   template: /eng/pipelines/templates/stages/archetype-sdk-client.yml
   parameters:
     ServiceDirectory: iot
+    BuildSnippets: true
```
jsquire, line 23: If this is being done by default, do we need this here?   Maybe it was intended to be `false`?

sdk/dns/ci.mgmt.yml
```diff
@@ -18,6 +18,7 @@ extends:
   template: /eng/pipelines/templates/stages/archetype-sdk-client.yml
   parameters:
     ServiceDirectory: dns
+    BuildSnippets: true
```
jsquire, line 21: Since this is done by default, do we need this here?

jsquire (approved):

annelo-msft (approved):
ACR LGTM, Thanks @heaths!

weshaggard (approved):
The change looks reasonable to me.

sdk/iot/ci.yml
heaths: `true`, `false` - you say "toe-ma-toe" I say "toe-may-toe".

But, yes. :)

pallavit:
Do we want to create follow-up issues for the SDKs where we had to manually set this to false? This looks great, thank you for doing it.

pallavit (approved):

heaths:
> Do we want to create follow-up issues for the SDKs where we had to manually set this to false? This looks great, thank you for doing it.

@pallavit already did. :) See https://github.com/Azure/azure-sdk-for-net/issues/20431. I updated the OP with individual tracking issues as tasks.

PR #33881, opened 2023-02-03
Fix Guid handling

### LINKED ISSUE #33825: Issues with Guid serialization/deserialization
Historically, the Service Bus library has parsed the delivery tag as Guid when receiving a message, and serialized from a Guid when settling/sending. The AMQP spec states that the delivery tag is 32 opaque bytes, but apparently the service has implemented this using Guid.ToByteArray. When we serialize/deserialize, we do not currently have endianness handling for all code paths.
The Guid constructor handles this automatically, but we use MemoryMarshal.TryRead/TryWrite in various places which does not have any endianness handling.

Initial discussion - https://github.com/Azure/azure-sdk-for-net/pull/33682#discussion_r1091161237

Guid usages that need to be fixed:

https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs#L549

https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpReceiver.cs#L479

JoshLove-msft:
Fixes https://github.com/Azure/azure-sdk-for-net/issues/33825

### DIFF
--- a/sdk/core/Azure.Core/src/Shared/GuidUtilities.cs
+++ b/sdk/core/Azure.Core/src/Shared/GuidUtilities.cs
@@ -20,6 +20,7 @@ internal static class GuidUtilities
     {
         if (bytes.Length != GuidSizeInBytes)
         {
+            guid = default;
             return false;
         }
 
--- a/sdk/core/Azure.Core/src/Shared/GuidUtilities.cs
+++ b/sdk/core/Azure.Core/src/Shared/GuidUtilities.cs
@@ -50,10 +50,7 @@ internal static class GuidUtilities
 
     public static void WriteGuidToBuffer(Guid guid, Span<byte> buffer)
     {
-        if (buffer.Length != GuidSizeInBytes)
-        {
-            throw new ArgumentException($"The length of the supplied buffer must be {GuidSizeInBytes}.", nameof(buffer));
-        }
+        AssertBufferSize(buffer);
 
         // Based on https://github.com/dotnet/runtime/blob/9129083c2fc6ef32479168f0555875b54aee4dfb/src/libraries/System.Private.CoreLib/src/System/Guid.cs#L836
 
@@ -78,6 +75,14 @@ internal static class GuidUtilities
         buffer[14] = data.J;
     }
 
+    private static void AssertBufferSize(Span<byte> buffer)
+    {
+        if (buffer.Length != GuidSizeInBytes)
+        {
+            throw new ArgumentException($"The length of the supplied buffer must be {GuidSizeInBytes}.", nameof(buffer));
+        }
+    }
+
     // This struct has the fields layed out to be GUID-like in order to read the GUID fields
     // to efficiently write them into memory without having to deal with endianness
     // Do not rename or reorder the fields.
@@ -88,8 +93,8 @@ internal static class GuidUtilities
         public readonly short C;
         public readonly byte D;
         public readonly byte E;
-        public readonly byte F; 
-        public readonly byte G; 
+        public readonly byte F;
+        public readonly byte G;
         public readonly byte H;
         public readonly byte I;
         public readonly byte J;
--- a/sdk/core/Azure.Core/src/Shared/GuidUtilities.cs
+++ b/sdk/core/Azure.Core/src/Shared/GuidUtilities.cs
@@ -6,20 +6,21 @@ using System.Buffers.Binary;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 
-namespace Azure.Messaging.ServiceBus.Primitives;
+namespace Azure.Core.Shared;
 
+/// <summary>
+/// The purpose of this class is to provide an allocation free way to parse and write Guids on .NET Standard 2.0, taking into account
+/// endianness.
+/// </summary>
 internal static class GuidUtilities
 {
-    /// <summary>
-    /// The size, in bytes, to use for extracting the delivery tag bytes into <see cref="Guid"/>.
-    /// </summary>
     private const int GuidSizeInBytes = 16;
 
     public static Guid ParseGuidBytes(ReadOnlyMemory<byte> bytes)
     {
         if (bytes.Length != GuidSizeInBytes)
         {
-            ThrowArgumentException();
+            throw new ArgumentException("The length of the supplied bytes must be 16.", nameof(bytes));
         }
 
         ReadOnlySpan<byte> bytesSpan = bytes.Span;
@@ -28,6 +29,7 @@ internal static class GuidUtilities
             return MemoryMarshal.Read<Guid>(bytesSpan);
         }
 
+        // copied from https://github.com/dotnet/runtime/blob/9129083c2fc6ef32479168f0555875b54aee4dfb/src/libraries/System.Private.CoreLib/src/System/Guid.cs#L49
         // slower path for BigEndian:
         byte k = bytesSpan[15];  // hoist bounds checks
         int a = BinaryPrimitives.ReadInt32LittleEndian(bytesSpan);
@@ -42,19 +44,16 @@ internal static class GuidUtilities
         byte j = bytesSpan[14];
 
         return new Guid(a, b, c, d, e, f, g, h, i, j, k);
-
-        static void ThrowArgumentException()
-        {
-            throw new ArgumentException("TBD", nameof(bytes));
-        }
     }
 
     public static void WriteGuidBytes(Guid guid, byte[] buffer)
     {
         if (buffer.Length != GuidSizeInBytes)
         {
-            ThrowArgumentException();
+            throw new ArgumentException("The length of the supplied buffer must be 16.", nameof(buffer));
         }
+
+        // Copied from https://github.com/dotnet/runtime/blob/9129083c2fc6ef32479168f0555875b54aee4dfb/src/libraries/System.Private.CoreLib/src/System/Guid.cs#L836
 
         if (BitConverter.IsLittleEndian)
         {
@@ -76,11 +75,6 @@ internal static class GuidUtilities
         buffer[12] = data.H;
         buffer[13] = data.I;
         buffer[14] = data.J;
-
-        static void ThrowArgumentException()
-        {
-            throw new ArgumentException("TBD", nameof(buffer));
-        }
     }
 
     private readonly struct GuidData
--- a/sdk/core/Azure.Core/src/Shared/GuidUtilities.cs
+++ b/sdk/core/Azure.Core/src/Shared/GuidUtilities.cs
@@ -78,20 +78,25 @@ internal static class GuidUtilities
         buffer[14] = data.J;
     }
 
+    // This struct has the fields layed out to be GUID-like in order to read the GUID fields
+    // to efficiently write them into memory without having to deal with endianness
+    // Do not rename or reorder the fields.
     private readonly struct GuidData
     {
-        public readonly int A;   // Do not rename (binary serialization)
-        public readonly short B; // Do not rename (binary serialization)
-        public readonly short C; // Do not rename (binary serialization)
-        public readonly byte D;  // Do not rename (binary serialization)
-        public readonly byte E;  // Do not rename (binary serialization)
-        public readonly byte F;  // Do not rename (binary serialization)
-        public readonly byte G;  // Do not rename (binary serialization)
-        public readonly byte H;  // Do not rename (binary serialization)
-        public readonly byte I;  // Do not rename (binary serialization)
-        public readonly byte J;  // Do not rename (binary serialization)
-        public readonly byte K;  // Do not rename (binary serialization)
+        public readonly int A;
+        public readonly short B;
+        public readonly short C;
+        public readonly byte D;
+        public readonly byte E;
+        public readonly byte F; 
+        public readonly byte G; 
+        public readonly byte H;
+        public readonly byte I;
+        public readonly byte J;
+        public readonly byte K;
 
+        // Creates a new GUID like struct initialized to the value represented by the
+        // arguments.  The bytes are specified like this to avoid endianness issues.
         public GuidData(int a, short b, short c, byte d, byte e, byte f, byte g, byte h, byte i, byte j, byte k)
         {
             A = a;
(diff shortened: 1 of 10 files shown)


sdk/servicebus/Azure.Messaging.ServiceBus/tests/Amqp/AmqpConverterTests.cs
```diff
@@ -156,5 +156,24 @@ public void CanParseMaxAbsoluteExpiryTime()
...
+        public void CanParseGuidBytes()
+        {
+            var input = Guid.NewGuid();
+
+            Guid output = AmqpMessageConverter.ParseGuidBytes(input.ToByteArray());
+            Assert.AreEqual(input, output);
+        }
+
+        [Test]
+        public void CanWriteGuidBytes()
```
JoshLove-msft, line 170: @jsquire I might have to take you up on your offer to run these on a big endian machine 😄

danielmarbach:
What do you think about this one https://github.com/danielmarbach/azure-sdk-for-net/commit/f19230e9b9da43f929f904f61c662a3ca68973b3 ?

danielmarbach:
For some reason I wasn't able to make a PR against the PR

danielmarbach:
Pushed another commit that should get rid of the allocations for writing in big endian https://github.com/Azure/azure-sdk-for-net/commit/c4b010272ba7b7eb4739525722c646714b0a2c16

sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/GuidUtilities.cs
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+
+namespace Azure.Messaging.ServiceBus.Primitives;
+
+internal static class GuidUtilities
```
jsquire, line 10: Should we consider moving this to Core?   May be situational, but definitely general purpose.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/GuidUtilities.cs
```diff
@@ -0,0 +1,42 @@
...
+{
+    /// <summary>
+    /// The size, in bytes, to use for extracting the delivery tag bytes into <see cref="Guid"/>.
```
jsquire, line 13: nit:  Since this is general now, we probably want to remove the reference to delivery tag here.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/GuidUtilities.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// </summary>
+    private const int GuidSizeInBytes = 16;
+
+    public static Guid ParseGuidBytes(ReadOnlyMemory<byte> bytes)
+    {
+        if (bytes.Length == GuidSizeInBytes)
+        {
+            // Use TryRead to avoid allocating an array if we are on a little endian machine.
+            if (!BitConverter.IsLittleEndian || !MemoryMarshal.TryRead<Guid>(bytes.Span, out var lockTokenGuid))
```
jsquire, line 22: It may be helpful to offer a bit more context around the approach and how endianness plays into it.   Would be especially helpful if this does move to Core.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/GuidUtilities.cs
```diff
@@ -0,0 +1,42 @@
...
+                // Even if the bytes were not valid, use the Guid constructor to leverage the Guid validation rather than throwing ourselves.
+                lockTokenGuid = new Guid(bytes.ToArray());
+            }
+            return lockTokenGuid;
+        }
+
+        return default;
+    }
+
+    public static void WriteGuidBytes(Guid guid, byte[] buffer)
```
jsquire, line 34: The name confused me a bit here and I wasn't clear that we were writing the GUID into the buffer at first.   Maybe something like `WriteGuidToBuffer` or some comments that describe things would help?

jsquire (approved):

JoshLove-msft:
> Pushed another commit that should get rid of the allocations for writing in big endian [c4b0102](https://github.com/Azure/azure-sdk-for-net/commit/c4b010272ba7b7eb4739525722c646714b0a2c16)

Thanks, let me take a look and cherry-pick these onto my branch.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/GuidUtilities.cs
JoshLove-msft: There are some nuances that are somewhat idiomatic to the SB use case, e.g. returning default rather than throwing when trying to parse a Guid from a buffer that is not 16 bytes. That said, we can probably rethink whether we need this behavior. I will look into moving this to Core.

danielmarbach: My implementation is more aligned with the runtime code so be aware.

JoshLove-msft: Yep, I noticed that. I think it is probably okay as receiving a message with a malformed delivery tag would make it unusable in most scenarios anyway.

sdk/core/Azure.Core/src/Shared/GuidUtilities.cs
```diff
@@ -0,0 +1,109 @@
...
+        byte f = bytesSpan[10];
+        byte g = bytesSpan[11];
+        byte h = bytesSpan[12];
+        byte i = bytesSpan[13];
+        byte j = bytesSpan[14];
+
+        return new Guid(a, b, c, d, e, f, g, h, i, j, k);
+    }
+
+    public static void WriteGuidToBuffer(Guid guid, byte[] buffer)
```
KrzysztofCwalina, line 49: This should take Span so that I can write a GUID in a middle of an existing (longer) buffer.

sdk/core/Azure.Core/src/Shared/GuidUtilities.cs
```diff
@@ -0,0 +1,109 @@
...
+
+/// <summary>
+/// The purpose of this class is to provide an allocation free way to parse and write Guids on .NET Standard 2.0, taking into account
+/// endianness.
+/// </summary>
+internal static class GuidUtilities
+{
+    private const int GuidSizeInBytes = 16;
+
+    public static Guid ParseGuidBytes(ReadOnlyMemory<byte> bytes)
```
KrzysztofCwalina, line 19: This should take ROS, not ROM, so that I can parse a guid from a stack allocated buffer, for example.

sdk/core/Azure.Core/src/Shared/GuidUtilities.cs
```diff
@@ -0,0 +1,109 @@
...
+    {
+        if (bytes.Length != GuidSizeInBytes)
+        {
+            return default;
```
KrzysztofCwalina: I wonder if we should not simply throw. There is no good reason to handle this error condition

danielmarbach: I'm not disagreeing. Just sharing the previous comments

https://github.com/Azure/azure-sdk-for-net/pull/33881#discussion_r1096161178

and showing this change https://github.com/Azure/azure-sdk-for-net/pull/33881/commits/0ec7a80d125576f3b09c58d7da71cb20e2b029a8

sdk/core/Azure.Core/src/Shared/GuidUtilities.cs
danielmarbach: I was actually thinking the same when I introduced my implementation changes but left it as is since the method signature was suggested like that based on the service bus usage. Now that it is in the core I think there are even stronger reasons to make this ROS instead of ROM

sdk/core/Azure.Core/src/Shared/GuidUtilities.cs
danielmarbach: I think especially since it is now in the Core it should either throw or be changed into a Try pattern so that the consumer of the API can decide whether to throw or not

sdk/core/Azure.Core/src/Shared/GuidUtilities.cs
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    public static void WriteGuidToBuffer(Guid guid, byte[] buffer)
+    {
+        if (buffer.Length != GuidSizeInBytes)
+        {
+            throw new ArgumentException("The length of the supplied buffer must be 16.", nameof(buffer));
+        }
+
+        // Copied from https://github.com/dotnet/runtime/blob/9129083c2fc6ef32479168f0555875b54aee4dfb/src/libraries/System.Private.CoreLib/src/System/Guid.cs#L836
```
danielmarbach, line 56: Nit: One could say this is a partial copy since that code has access to the fields while here we need to do the struct trick with `Unsafe.As

sdk/core/Azure.Core/src/Shared/GuidUtilities.cs
```diff
@@ -0,0 +1,109 @@
...
+/// </summary>
+internal static class GuidUtilities
+{
+    private const int GuidSizeInBytes = 16;
+
+    public static Guid ParseGuidBytes(ReadOnlyMemory<byte> bytes)
+    {
+        if (bytes.Length != GuidSizeInBytes)
+        {
+            throw new ArgumentException("The length of the supplied bytes must be 16.", nameof(bytes));
```
danielmarbach, line 23: My version had sort of a throw helper. I haven't checked but in theory it could make the code more inlineable and we might even think about adding `[MethodImpl(MethodImplOptions.AggressiveInlining)]`

sdk/core/Azure.Core/src/Shared/GuidUtilities.cs
```diff
@@ -0,0 +1,109 @@
...
+        public readonly byte D;  // Do not rename (binary serialization)
+        public readonly byte E;  // Do not rename (binary serialization)
+        public readonly byte F;  // Do not rename (binary serialization)
+        public readonly byte G;  // Do not rename (binary serialization)
+        public readonly byte H;  // Do not rename (binary serialization)
+        public readonly byte I;  // Do not rename (binary serialization)
+        public readonly byte J;  // Do not rename (binary serialization)
+        public readonly byte K;  // Do not rename (binary serialization)
+
+        public GuidData(int a, short b, short c, byte d, byte e, byte f, byte g, byte h, byte i, byte j, byte k)
```
danielmarbach, line 94:
```suggestion
        // Creates a new GUID like struct initialized to the value represented by the
        // arguments.  The bytes are specified like this to avoid endianness issues.
        public GuidData(int a, short b, short c, byte d, byte e, byte f, byte g, byte h, byte i, byte j, byte k)
```

sdk/core/Azure.Core/src/Shared/GuidUtilities.cs
```diff
@@ -0,0 +1,109 @@
...
+        public readonly short B; // Do not rename (binary serialization)
+        public readonly short C; // Do not rename (binary serialization)
+        public readonly byte D;  // Do not rename (binary serialization)
+        public readonly byte E;  // Do not rename (binary serialization)
+        public readonly byte F;  // Do not rename (binary serialization)
+        public readonly byte G;  // Do not rename (binary serialization)
+        public readonly byte H;  // Do not rename (binary serialization)
+        public readonly byte I;  // Do not rename (binary serialization)
+        public readonly byte J;  // Do not rename (binary serialization)
+        public readonly byte K;  // Do not rename (binary serialization)
```
danielmarbach, line 92:
```suggestion
        public readonly int A;
        public readonly short B;
        public readonly short C;
        public readonly byte D;
        public readonly byte E;
        public readonly byte F; 
        public readonly byte G; 
        public readonly byte H;
        public readonly byte I;
        public readonly byte J;
        public readonly byte K;
```

sdk/core/Azure.Core/src/Shared/GuidUtilities.cs
JoshLove-msft: Yeah this was updated to throw.

sdk/core/Azure.Core/src/Shared/GuidUtilities.cs
```diff
@@ -0,0 +1,109 @@
...
+        buffer[8] = data.D;
+        buffer[9] = data.E;
+        buffer[10] = data.F;
+        buffer[11] = data.G;
+        buffer[12] = data.H;
+        buffer[13] = data.I;
+        buffer[14] = data.J;
+    }
+
+    private readonly struct GuidData
```
danielmarbach, line 80:
```suggestion
    // This struct has the fields layed out to be GUID-like in order to read the GUID fields
    // to efficiently write them into memory without having to deal with endianness
    // Do not rename or reorder the fields.
    private readonly struct GuidData
```

sdk/core/Azure.Core/src/Shared/GuidUtilities.cs
JoshLove-msft: How about "based on" 😆

sdk/core/Azure.Core/src/Shared/GuidUtilities.cs
JoshLove-msft: Sorry, but why would adding a method make it more inlineable?

JoshLove-msft: Ah I understand - these utility methods will be more inlineable by not throwing inline.

jsquire:
Confirming that the version merged has been run on a big endian system and all tests are passing.

danielmarbach:
Awesome!

_Files an official request to make @jsquire laptop a build agent for the SDK_

:smile_cat:

jsquire:
> Awesome!
> 
> _Files an official request to make @jsquire laptop a build agent for the SDK_
> 
> 😸

Well, funny story.  I went to look for something this weekend and it turns out that the article that I referenced was an early pre-production discussion.   Apple ended up going little endian for the final version.   https://developer.apple.com/documentation/apple-silicon/porting-your-macos-apps-to-apple-silicon

So, I basically provided no value at all.

Issue #33433, opened 2023-01-12
Migrate to .NET Package Validation


heaths:
We're using an old _Microsoft.DotNet.ApiCompat_ package that doesn't evaluate compatible targets, and our own logic in _eng/Directory.Build.Common.Targets_ below is currently causing issues like #33431:

```xml
  <Target Name="_ResolveResolvedMatchingContract" BeforeTargets="ValidateApiCompatForSrc" Condition="'$(ApiCompatVersion)' != ''">
    <ItemGroup>
      <_ReferencePathDirectories Include="@(ReferencePath -> '%(RootDir)%(Directory)')"  />
      <ResolvedMatchingContract Include="$(NuGetPackageRoot)\$(PackageId.ToLower())\$(ApiCompatVersion)\lib\$(TargetFramework)\$(TargetFileName)">
        <DependencyPaths>@(_ReferencePathDirectories->Distinct(), ',')</DependencyPaths>
      </ResolvedMatchingContract>
    </ItemGroup>
  </Target>
```

It was recommended to switch to [built-in package validation](https://learn.microsoft.com/dotnet/fundamentals/package-validation/overview) now that we're moving to net7.0.

This will require some changes to any exceptions we have, so something we probably want to do only when working on changing all packages for #32597.

- [x] Remove old _Microsoft.DotNet.ApiCompat_ package
- [ ] Migrate any exceptions to the new format
- [ ] Rename `<ApiCompatVersion>` and updating tooling, **OR**
- [ ] Add `<Target>` to use `<ApiCompatVersion>` to define whatever property the built-in validation needs.

heaths:
Be sure to take advantage of the cross-platform, cross-framework API checks as [described](https://learn.microsoft.com/dotnet/fundamentals/package-validation/overview#validator-types).

heaths:
There may be some problems:

1. Can't find anything that would effectively ignore certain attributes being added or removed from a definition like we use now with _Microsoft.DotNet.ApiCompat_: `<ApiCompatExcludeAttributeList>$(MSBuildThisFileDirectory)ApiListing.exclude-attributes.txt</ApiCompatExcludeAttributeList>`

heaths:
I got some answers back from the .NET team that are relevant here:

To just run validation, you can install a global tool:

```bash
dotnet tool install --global Microsoft.DotNet.ApiCompat.Tool --version --prerelease
```

However, they state,

[...]

PR #33890, opened 2023-02-03
Update LangVersion in individual projects

AlexanderSher:


sdk/eventhub/Directory.Build.props
```diff
@@ -24,6 +24,6 @@
     this needs to be set after the common properties import.
   -->
   <PropertyGroup Condition="$(MSBuildProjectName.StartsWith('Azure.')) and $(MSBuildProjectName.EndsWith('.Tests'))">
-    <LangVersion>9.0</LangVersion>
+    <LangVersion>11.0</LangVersion>
```
jsquire, line 27: If we don't need to explicitly pin anymore, then we can just remove this section (L18 - L28).    There was something that caused issues with C# 10 where we had to pin to 9.

jsquire (approved):

pallavit:
Can/should we plan to add an analyzer rule to check for this?

sdk/eventhub/Directory.Build.props
AlexanderSher: Done

AlexanderSher:
> Can/should we plan to add an analyzer rule to check for this?

@pallavit You mean check that internal projects don't define lang version? From time to time it can be useful, cause new versions of language can introduce breaking changes that we may have no capacity to fix immediately.

jsquire:
> Can/should we plan to add an analyzer rule to check for this?

I don't think so.  Individual packages should be able to deviate if needed.   Particularly this is useful when you need to back-pin a specific library to work around compiler issues.

common/SmokeTests/SmokeTest/SmokeTest.csproj
```diff
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFrameworks>net7.0;net6.0;net461</TargetFrameworks>
-    <LangVersion>latest</LangVersion>
+    <LangVersion>11.0</LangVersion>
```
weshaggard, line 5: Can't this override be removed?

AlexanderSher: It is in a sample project, so I want to leave it here in case someone just copy it as a whole

heaths:
> > Can/should we plan to add an analyzer rule to check for this?
> 
> @pallavit You mean check that internal projects don't define lang version? From time to time it can be useful, cause new versions of language can introduce breaking changes that we may have no capacity to fix immediately.

As long as new syntactic language features can be built by the SDK in the root _global.json_. Language features that require a newer SDK should already fail the build since we target `netstandard2.0` at least. We do require public APIs to be the same across TFMs for now, and I have #33433 on my plate to hook up to help ensure that since the .NET team already has a solution for this (as well as backcompat checks - a replacement for what we're using now).

PR #33900, opened 2023-02-05
Included a sample code to use checkpointing

Pramod2392:
In the available samples event hub checkpointing is not included. Added a sample code to start reading the events based on the previous checkpoint position and also updating the checkpoint after reading the events.

### DIFF
--- a/sdk/eventhub/Azure.Messaging.EventHubs/samples/Sample05_ReadingEvents.md
+++ b/sdk/eventhub/Azure.Messaging.EventHubs/samples/Sample05_ReadingEvents.md
@@ -303,6 +303,83 @@ finally
 }
 ```
 
+## Read events from a partition, starting from the previous checkpoint position and updating the checkpoint after the events are read
+
+The "GetCheckpointAsync" method of "BlobCheckpointStore" allows to fetch the checkpoint information for a specific partition.This enables to read the data from where previously we had stopped and helps in preventing duplicate processing of events.
+The "UpdateCheckpointAsync" method of "BlobCheckpointStore" allows to update the checkpoint information for a specific partition.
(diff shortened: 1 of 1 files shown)


Pramod2392:

jsquire (changes requested):
Hi @Pramod2392.   Thank you for your contribution, but I don't believe that we'll be moving forward with the proposed sample.

We intentionally try to avoid external dependencies in samples, focusing on the core scenarios for types in the `Azure.Messaging.EventHubs` package.   In this case, the `BlobCheckpointStore` is  defined in the `Azure.Messaging.EventHubs.Processor` package and also brings in a dependency on the Storage package.

The checkpoint store is explicitly intended for use with our event processors and while this stand-alone use would work, it's not a scenario that we want to officially promote as a recommended pattern.   There are quite a few edge cases and performance considerations that are difficult to take into account in a short sample like this without becoming overly complex.

I do think this could be helpful to Azure developers and that it may be a good fit for an article in the Azure SDK blog where the scenario could be expanded, the corner cases and performance considerations covered with discussion.   If you'd like to consider this as an article, the [Azure SDK blog site](https://devblogs.microsoft.com/azure-sdk/) describes the process as:

> ### Azure SDK Blog Contributions
>
>Thank you for reading this Azure SDK blog! We hope you learned something new, and we welcome you to share these posts. We're open to Azure SDK blog contributions. To get started, contact us at [azsdkblog@microsoft.com](mailto:azsdkblog@microsoft.com) with your idea, and we’ll set you up as a guest blogger.

jsquire:
Because we won't be moving forward with this sample, I'm going to close the PR out.   Thank you, again, for your contribution.   Your efforts are greatly appreciated.

PR #33907, opened 2023-02-06
[Automation] Fix the problem that stream is still used after being closed

### LINKED ISSUE #33275: [BUG] ArgumentException 'stream must be seekable' when calling ReplaceContentRunbookDraftAsync.
### Library name and version

Azure.ResourceManager.Automation 1.0.0

### Describe the bug

We are consistently observing an ArgumentException 'stream must be seekable' when calling ReplaceContentRunbookDraftAsync to upload a runbook to Azure Automation. The runbook content was converted to a buffer of UTF encoded bytes and which was then used to create a MemoryStream. We have confirmed that the stream that is being bassed to ReplaceContentRunbookDraftAsync  is indeed seekable.

### Expected behavior

The runbook content should be uploaded to Azure Automation without error.

### Actual behavior

Fails with ArgumentException 'stream must be seekable'.

### Reproduction Steps

public async Task UploadRunbookContent(ResourceIdentifier runbookIdentifier, string content)
{
    var options = new InteractiveBrowserCredentialOptions
    {
        TenantId = "<Your Tenant ID>"
        ClientId = "<Your Cliend ID>"		
        RedirectUri = new Uri("<Your Redirect URI>")
    };

    var credential = new InteractiveBrowserCredential(options);
    var armClient = new ArmClient(credential);	
    AutomationRunbookResource resource = armClient.GetAutomationRunbookResource(runbookIdentifier);

    using (var stream = new MemoryStream(Encoding.UTF8.GetBytes(content)))
    {
        await resource.ReplaceContentRunbookDraftAsync(WaitUntil.Completed, stream);
    }
}

### Environment

_No response_

Yao725:
Fix #33275 

Details:

The body parameter of the method ReplaceContentRunbookDraft is a stream, so it will be closed after we get the response. However, this parameter will be used again when constructing the ArmOperation and it leads to that we will try to create a RequestContent from stream with CanSeek is equal to false, which will throw the [exception](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/RequestContent.cs#L130).

To fix this issue, considering that the ArmOperation only needs the Uri and Method of the request, this PR customized the internal `CreateReplaceContentRequest` method with an additional parameter `isContentSkipped` to skip content when we construct the request.


m-nash (approved):

Yao725:
Merge this PR as a temporary workaround, will remove the customized code once [this](https://github.com/Azure/autorest.csharp/issues/3076) is fixed in generator.

PR #33915, opened 2023-02-06
[Text Analytics] Remove options parameter from `DynamicClassify` and `DynamicClassifyAsync` methods

### LINKED ISSUE #33914: [Text Analytics] Remove options parameter from `DynamicClassify` and `DynamicClassifyAsync` methods
The `DynamicClassify` and `DynamicClassifyAsync` methods of the `TextAnalyticsClient` class should not receive a `DynamicClassifyOptions` parameter. This is to align with the other convenience methods that receive a single document as a string parameter. 

A side effect of doing this is that customers would no longer be able to choose the `ClassificationType` when calling these methods because this is currently a property in the `DynamicClassifyOptions` class. The methods would always use the service default, i.e., multi-category classification. Alternatively, the property can be "moved out" of the `DynamicClassifyOptions` class and turned into a standalone parameter that can be passed into all these different client methods. This also means that we would no longer need the `DynamicClassifyOptions` sub-class, since `ClassificationType` is the only property that it adds on top of the `TextAnalyticsRequestOptions` base class instead.

Here's an example of what the `DynamicClassify` method could potentially look like:
```csharp
public virtual Response<ClassificationCategoryCollection> DynamicClassify(
    string document,
    IEnumerable<string> categories,
    ClassificationType? classificationType = default,
    string language = default,
    CancellationToken cancellationToken = default);
```

joseharriaga:
Resolves: https://github.com/Azure/azure-sdk-for-net/issues/33914

The `DynamicClassify` and `DynamicClassifyAsync` methods of the `TextAnalyticsClient` class should not receive a `DynamicClassifyOptions` parameter. This is to align with the other convenience methods that receive a single document as a string parameter (e.g., `RecognizeEntities`, `ExtractKeyPhrases`, etc.)

The `ClassificationType` property is also being "moved out" of the `DynamicClassifyOptions` class and turned into a standalone parameter that can be passed into all the dynamic classification methods. This means that customers can still choose the classification type (multi vs. single) when calling `DynamicClassify` or `DynamicClassifyAsync` without the options parameter. Additionally, we no longer need the `DynamicClassifyOptions` sub-class, since `ClassificationType` was the only property that it added on top of the `TextAnalyticsRequestOptions` base class.


sdk/textanalytics/Azure.AI.TextAnalytics/api/Azure.AI.TextAnalytics.netstandard2.0.cs
```diff
@@ -1593,12 +1588,12 @@ public partial class TextAnalyticsClient
...
+        public virtual Azure.Response<Azure.AI.TextAnalytics.DynamicClassifyDocumentResultCollection> DynamicClassifyBatch(System.Collections.Generic.IEnumerable<Azure.AI.TextAnalytics.TextDocumentInput> documents, System.Collections.Generic.IEnumerable<string> categories, Azure.AI.TextAnalytics.ClassificationType? classificationType = default(Azure.AI.TextAnalytics.ClassificationType?), Azure.AI.TextAnalytics.TextAnalyticsRequestOptions options = null, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
+        public virtual Azure.Response<Azure.AI.TextAnalytics.DynamicClassifyDocumentResultCollection> DynamicClassifyBatch(System.Collections.Generic.IEnumerable<string> documents, System.Collections.Generic.IEnumerable<string> categories, Azure.AI.TextAnalytics.ClassificationType? classificationType = default(Azure.AI.TextAnalytics.ClassificationType?), string language = null, Azure.AI.TextAnalytics.TextAnalyticsRequestOptions options = null, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
```
mssfang, line 1594: DynamicClassifyOptions can extend TextAnalyticsRequestOptions.
In java, we prefer to have a feature-based optional bag DynamicClassifyOptions, because it could add more properties that only apply to  Dynamic Classification but not TextAnalyticsRequestOptions which is the base optional bag that applies to most of TA features.

joseharriaga: @heaths Should we do this in .NET too? I'm looking at the other methods in the `TextAnalyticsClient` class and I see many of them receive a simple `TextAnalyticsRequestOptions` parameter (e.g. `DetectLanguageBatch`, `ExtractKeyPhrasesBatch`, `RecognizeEntitiesBatch`, etc.). Is it a better idea to create a `DynamicClassifyOptions` class that inherits from `TextAnalyticsRequestOptions` (even if it doesn't add anything to it) like Java does?

mssfang: For more info: https://apiview.dev/Assemblies/Review/bd92063fce2d4a8c933c61ada05c8ce9#com.azure.ai.textanalytics.models.DynamicClassificationOptions

heaths: WE typically do this in .NET, too, but IIF the derivative class shares _all_ properties of the parent. In TA, there's quite a few cases where that's not true. So, @joseharriaga, if this does share all the parent properties that's fine. You might've been looking at some cases where previous maintainers including myself didn't inherit because not all parent properties were legit. It should follow derivation based on swagger, really.

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
```diff
@@ -2237,10 +2237,10 @@ public virtual async Task<AnalyzeHealthcareEntitiesOperation> StartAnalyzeHealth
         public virtual Response<ClassificationCategoryCollection> DynamicClassify(
             string document,
             IEnumerable<string> categories,
+            ClassificationType? classificationType = default,
             string language = default,
```
heaths, line 2241: This hasn't GA'd yet, right? If it has, don't change the order and use an overload. If it hasn't (which I think is the case), I would still consider putting the language third for consistency. Or, is it consistent that it is last (before a cancellation token)? I don't recall.

heaths (changes requested):
LGTM, but I have an open question/concern I'd like to discuss first - to understand what prompted this change. It seems unnecessary and divergent from what we've done in the past.

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
joseharriaga: It hasn't GA'd so reordering is okay. Regarding consistency, the pattern that I think I'm seeing appears to determine the position of the `language` property by counting from the last parameter to the first with two possible options:
1. `cancellationToken` and `language` (does not receive an `options` parameter)
2. `cancellationToken`, `options`, and `language`.

For example, the `StartSingleLabelClassify` method looks like this: 
```csharp
public virtual ClassifyDocumentOperation StartSingleLabelClassify(
    IEnumerable<string> documents,
    string projectName,
    string deploymentName,
    string language = default,
    SingleLabelClassifyOptions options = default,
    CancellationToken cancellationToken = default)
```

sdk/textanalytics/Azure.AI.TextAnalytics/api/Azure.AI.TextAnalytics.netstandard2.0.cs
joseharriaga: In this case, all the properties in `TextAnalyticsRequestOptions` are valid. In other words, it's not a problem per se, but rather just wondering if we have a preference for using `TextAnalyticsRequestOptions` as the parameter type, or if we prefer to create individual option classes that derive from it for each different feature and use that instead. 

For example, in this case, I would end up with a `DynamicClassifyOptions` class that inherits from `TextAnalyticsRequestOptions` (but does not actually add anything on top of it), and I would change the parameter type of these DynamicClassify- methods to receive a `DynamicClassifyOptions` parameter specifically.

sdk/textanalytics/Azure.AI.TextAnalytics/src/TextAnalyticsClient.cs
heaths: Right, so seems `options` should go last. In fact, I thought we had a guideline about that: https://azure.github.io/azure-sdk/dotnet_introduction.html#dotnet-parameters

Turns out, though, we don't document position. But, IIRC, conversations with @KrzysztofCwalina and/or @tg-msft in the past said it should be next to last, where last is the `cancellationToken`.

joseharriaga: Yes, I believe this is correct! I have `options` (where applicable) next to last, and last is always `cancellationToken`:


```csharp
public virtual Azure.Response<Azure.AI.TextAnalytics.ClassificationCategoryCollection>
    DynamicClassify(
        string document,
        System.Collections.Generic.IEnumerable<string> categories,
        Azure.AI.TextAnalytics.ClassificationType? classificationType = default,
        string language = default,
        System.Threading.CancellationToken cancellationToken = default)
```
```csharp
public virtual System.Threading.Tasks.Task<Azure.Response<Azure.AI.TextAnalytics.ClassificationCategoryCollection>> 
    DynamicClassifyAsync(
        string document,
        System.Collections.Generic.IEnumerable<string> categories,
        Azure.AI.TextAnalytics.ClassificationType? classificationType = default, 
        string language = default,
        System.Threading.CancellationToken cancellationToken = default)
```
```csharp
public virtual Azure.Response<Azure.AI.TextAnalytics.DynamicClassifyDocumentResultCollection> 
    DynamicClassifyBatch(
        System.Collections.Generic.IEnumerable<Azure.AI.TextAnalytics.TextDocumentInput> documents,
        System.Collections.Generic.IEnumerable<string> categories,
        Azure.AI.TextAnalytics.ClassificationType? classificationType = default, 
        Azure.AI.TextAnalytics.TextAnalyticsRequestOptions options = default,
        System.Threading.CancellationToken cancellationToken = default)
```
```csharp
public virtual Azure.Response<Azure.AI.TextAnalytics.DynamicClassifyDocumentResultCollection> 
    DynamicClassifyBatch(
        System.Collections.Generic.IEnumerable<string> documents,
        System.Collections.Generic.IEnumerable<string> categories,
        Azure.AI.TextAnalytics.ClassificationType? classificationType = default, 
        string language = default,
        Azure.AI.TextAnalytics.TextAnalyticsRequestOptions options = default,
        System.Threading.CancellationToken cancellationToken = default) 
```
```csharp
public virtual System.Threading.Tasks.Task<Azure.Response<Azure.AI.TextAnalytics.DynamicClassifyDocumentResultCollection>>
    DynamicClassifyBatchAsync(
        System.Collections.Generic.IEnumerable<Azure.AI.TextAnalytics.TextDocumentInput> documents, 
        System.Collections.Generic.IEnumerable<string> categories,
        Azure.AI.TextAnalytics.ClassificationType? classificationType = default,
        Azure.AI.TextAnalytics.TextAnalyticsRequestOptions options = default,
        System.Threading.CancellationToken cancellationToken = default)
```
```csharp
public virtual System.Threading.Tasks.Task<Azure.Response<Azure.AI.TextAnalytics.DynamicClassifyDocumentResultCollection>> 
    DynamicClassifyBatchAsync(
        System.Collections.Generic.IEnumerable<string> documents,
        System.Collections.Generic.IEnumerable<string> categories,
        Azure.AI.TextAnalytics.ClassificationType? classificationType = default,
        string language = default,
        Azure.AI.TextAnalytics.TextAnalyticsRequestOptions options = default,
        System.Threading.CancellationToken cancellationToken = default)
```

heaths (approved):

PR #33925, opened 2023-02-06
[Smoke Tests] Ignore Known Incompatible Packages

jsquire:
# Summary

The focus of these changes is to define a set that contains the list of packages which should not be included as references to the smoke tests. This is necessary because some compatibility packages do not target `netstandard2.0`, and the package source is unable to filter upstream to restrict to the appropriate packages for the running target.

# References and Related

- [Example pipeline failure](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2165670&view=logs&j=3a441aab-9de4-504b-b9a2-4be57a5928c9&t=e498f9c1-7696-590c-95d2-e75523098f78)

### DIFF
--- a/common/SmokeTests/SmokeTest/Update-Dependencies.ps1
+++ b/common/SmokeTests/SmokeTest/Update-Dependencies.ps1
@@ -7,6 +7,15 @@ param(
 
 . (Join-Path $PSScriptRoot ../../../eng/scripts/smoketest/Get-Package-Version.ps1)
 
+# Define a set that contains the list of packages which should not be included
+# as references to the smoke tests.  This is necessary because some compatibility
+# packages do not target netstandard2.0, and the package source is unable to
+# filter upstream to restrict to the appropriate packages for the running target.
+
+$packageExcludeSet = [System.Collections.Generic.HashSet[string]]@(
+    "Microsoft.Azure.WebPubSub.AspNetCore"
+)
+
 function Log-Warning($message) {
     if ($CI) {
         Write-Host "##vso[task.logissue type=warning]$message"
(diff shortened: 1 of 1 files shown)


common/SmokeTests/SmokeTest/Update-Dependencies.ps1
```diff
@@ -7,6 +7,15 @@ param(
 
 . (Join-Path $PSScriptRoot ../../../eng/scripts/smoketest/Get-Package-Version.ps1)
 
+# Define a set that contains the list of packages which should not be included
+# as references to the smoke tests.  This is necessary because some compatibility
+# packages do not target netstandard2.0, and the package source is unable to
+# filter upstream to restrict to the appropriate packages for the running target.
+
+$packageExcludeSet = [System.Collections.Generic.HashSet[string]]@(
```
jsquire, line 15: Using a hash table would have allowed using pure PowerShell here, but it seemed wrong to use it when I only care about the keys.

heaths: I mean, given we probably need so few packages here anyway, even a simple array (just elide the type conversion) would be fine. It's unlikely someone would dup an item and, even if they did, we're talking about ms here, not even seconds.

common/SmokeTests/SmokeTest/Update-Dependencies.ps1
```diff
@@ -67,6 +76,7 @@ Log-Info "Querying package information."
 $referenceUpdateCount = 0
 
 Get-SmokeTestPkgProperties -ArtifactsPath $ArtifactsPath
+| Where-Object { -not $packageExcludeSet.Contains($_.Name) }
```
heaths, line 79: Calling a .NET API like this will use the default comparison, which is almost undoubtedly case-sensitive. Instead, use:

`where-object $packageExcludeSet -notcontains Name`

heaths (approved):

jsquire:
@ckairen:  We seem to be failing on the Azure.Core part of the package feed here and not able to hit my changes to test.   Any idea why?

common/SmokeTests/SmokeTest/Update-Dependencies.ps1
jsquire: Fair point.   Converted to an array.

jsquire:
Pipeline is bombing out on an unrelated area of the feed.   I'm going to go ahead and merge, as we're already seeing daily failures so these changes can't make it worse.

PR #33931, opened 2023-02-07
[Draft for discussion] Preliminary updates to Confidential Ledger samples to use DynamicJson

annelo-msft:
The `DynamicJson` type released in [Azure.Core.Experimental 0.1.0-preview.24](https://www.nuget.org/packages/Azure.Core.Experimental/0.1.0-preview.24).

This draft PR illustrates what it would look like at this point to move the Confidential Ledger samples to use DynamicJson instead of JsonDocument.


sdk/confidentialledger/Azure.Security.ConfidentialLedger/tests/samples/HelloWorldSamples.cs
```diff
@@ -44,21 +39,15 @@ public void HelloWorld()
 
             Response statusResponse = ledgerClient.GetTransactionStatus(transactionId);
 
-            string status = JsonDocument.Parse(statusResponse.Content)
-                .RootElement
-                .GetProperty("state")
-                .GetString();
+            string status = statusResponse.Content.ToDynamic().state;
```
annelo-msft, line 42: In an earlier version, we let callers access properties without case sensitivity.  This will need to be added back to `DynamicJson` if we want this feature in the new implementation.

sdk/confidentialledger/Azure.Security.ConfidentialLedger/tests/samples/HelloWorldSamples.cs
```diff
@@ -167,57 +148,58 @@ public void HelloWorld()
...
             contents = null;
             while (!loaded)
             {
-                loaded = JsonDocument.Parse(getResponse.Content)
-                    .RootElement
-                    .TryGetProperty("entry", out element);
+                dynamic json = getResponse.Content.ToDynamic();
+
+                // TODO: Not currently handled
+                //loaded = (LedgerEntry)json.entry != null;
```
annelo-msft, line 165: Previously, we returned a "null" JsonData if a property was missing, but this is not yet implemented in this new version.

sdk/confidentialledger/Azure.Security.ConfidentialLedger/README.md
```diff
@@ -72,21 +72,15 @@ Since Azure confidential ledger is a distributed system, rare transient failures
 ```C# Snippet:GetStatus
 Response statusResponse = ledgerClient.GetTransactionStatus(transactionId);
 
-string status = JsonDocument.Parse(statusResponse.Content)
-    .RootElement
-    .GetProperty("state")
-    .GetString();
+string status = statusResponse.Content.ToDynamic().state;
```
KrzysztofCwalina, line 75: I think out samples should always store the created DynamicJson in a variable, even if the sample uses the variable once. Otherwise, I worry that users will be reparsing the JSON many times as they evolve the code. i.e. instead of the line above, let's use:

```csharp
dynamic content = statusResponse.Content.ToDynamic();
string status = content.state;
```

annelo-msft: Cool, I was looking to minimize changes to the sample, but happy to set this as a best practice.  Thanks!

jsquire:
That's a pretty large set of boilerplate removed.   I think this makes a huge difference in readability.

sdk/confidentialledger/Azure.Security.ConfidentialLedger/README.md
christothes: I don't think we should standardize on naming the variable `content` as it is generic to the name of the response payload property. I'd suggest we pick a name related to the type being returned. e.g.

```c#
dynamic transactionStatus = statusResponse.Content.ToDynamic();
string status = transactionStatus.state;
```

sdk/confidentialledger/Azure.Security.ConfidentialLedger/tests/samples/HelloWorldSamples.cs
christothes: What is the current behavior?

sdk/confidentialledger/Azure.Security.ConfidentialLedger/tests/samples/HelloWorldSamples.cs
```diff
@@ -307,5 +296,10 @@ public void HelloWorld()
 
             #endregion
         }
+
+        private class LedgerEntry
```
christothes, line 300: Is this to support the ability to deal with missing properties?

sdk/confidentialledger/Azure.Security.ConfidentialLedger/tests/samples/HelloWorldSamples.cs
annelo-msft: Right now, if you access a property that isn't present in the JSON, it will throw, as shown in the few lines below this commented out line (which I added to implement the functionality using the current behavior).  Our goal is to enable checks for optional properties that look like strongly-typed models, so we want to be able to check for an absent property without an exception being thrown.  Further details here: https://github.com/Azure/azure-sdk-for-net/pull/33063#discussion_r1089532143

sdk/confidentialledger/Azure.Security.ConfidentialLedger/README.md
annelo-msft: I think the name of the dynamic variable is less important than the fact that we call ToDynamic() separately and then access members on it in separate lines of code.  Variable naming can be left to the author's discretion.

sdk/confidentialledger/Azure.Security.ConfidentialLedger/tests/samples/HelloWorldSamples.cs
annelo-msft: It is to enable the cast for the null-check, per https://github.com/Azure/azure-sdk-for-net/pull/33063#discussion_r1089532143

I'm not 100% sold on this as the best API for this, given the need to create a type just to do a null check.  I hope to work on that API this week, so watch this space for changes to this.  In an earlier version, we were able to just check for null:

https://github.com/Azure/azure-sdk-for-net/pull/32501/files#diff-d55b2105899a17c3fafc1a079b6bb5a777c542671abad5f96c12223977a26597R188

I'll look at whether we can do that here when I get in there.

sdk/confidentialledger/Azure.Security.ConfidentialLedger/README.md
annelo-msft: Sure, I'm happy to use the name you suggested for this variable when I loop back and make the suggested change.

sdk/core/Azure.Core.Experimental/src/DynamicJson.cs
```diff
@@ -38,7 +40,12 @@ internal override void WriteTo(Stream stream)
 
         private object GetProperty(string name)
         {
-            return new DynamicJson(_element.GetProperty(name));
+            if (_element.TryGetProperty(name, out MutableJsonElement element))
```
annelo-msft, line 43: The changes to DynamicJson will be merged as part of a separate PR.

sdk/core/Azure.Core.Experimental/tests/DynamicJsonTests.cs
```diff
@@ -278,6 +278,47 @@ public void CanMakeChangesAndAddNewProperty()
...
+            // Add property Baz
+            json.Baz = "baz";
+
+            // Remove property A
+            json.Bar = new { B = "b" };
+
+            // Properties are present
+            Assert.IsFalse((string)json.Foo == null);
+            Assert.IsFalse((string)json.Bar.B == null);
+            Assert.IsFalse((string)json.Baz == null);
```
christothes, line 316: What happens if we access a property on a non-existent property? e.g. `json.Baz.Fizz`?

sdk/confidentialledger/Azure.Security.ConfidentialLedger/tests/samples/HelloWorldSamples.cs
```diff
@@ -43,22 +38,17 @@ public void HelloWorld()
...
             // Wait for the entry to be committed
             while (status == "Pending")
             {
                 statusResponse = ledgerClient.GetTransactionStatus(transactionId);
-                status = JsonDocument.Parse(statusResponse.Content)
-                    .RootElement
-                    .GetProperty("state")
-                    .GetString();
+                transactionStatus = statusResponse.Content.ToDynamic();
+                status = transactionStatus.state;
```
KrzysztofCwalina, line 51: shouldn't we sleep at the end of loops like this?

sdk/confidentialledger/Azure.Security.ConfidentialLedger/README.md
```diff
@@ -71,22 +71,17 @@ Since Azure confidential ledger is a distributed system, rare transient failures
...
 // Wait for the entry to be committed
 while (status == "Pending")
 {
     statusResponse = ledgerClient.GetTransactionStatus(transactionId);
-    status = JsonDocument.Parse(statusResponse.Content)
-        .RootElement
-        .GetProperty("state")
-        .GetString();
+    transactionStatus = statusResponse.Content.ToDynamic();
+    status = transactionStatus.state;
```
KrzysztofCwalina, line 84: @annelo-msft, remind me what we decided about allowing pascal casing for dynamic property access.

annelo-msft: I am working on that in a separate branch now.

We can't implement it with full case-insensitivity as we did before, since we're not storing properties in a Dictionary anymore, where we could provide a case-insensitive Comparer for lookups.  JsonElement.GetProperty() is case sensitive (as far as I can tell), so we will need to proceed with a double-lookup on both PascalCase and camelCase.  I was working on the design for how to configure that.  It has implications for how we call Set on the properties as well.  If they are there, I think the logic is the same - you can access them either PascalCase or camelCase and set them without changing the casing of the property name.  When new properties get added is where it is tricky - ideally you would already know what the casing of the underlying JSON is and do that regardless of how the dynamic member is accessed, but I haven't come up with a good way of determining this yet either.  We will need to decide on defaults as well.

I'm hoping to share a proposal for this tomorrow or early next week, but happy to take suggestions.

sdk/confidentialledger/Azure.Security.ConfidentialLedger/README.md
```diff
@@ -116,37 +117,37 @@ ledgerClient.PostLedgerEntry(
...
-             getByCollectionResponse = ledgerClient.GetLedgerEntry(transactionId, collectionId);
-             ledgerEntry = getByCollectionResponse.Content.ToDynamic();
-             loaded = ledgerEntry.state != "Loading";
-         }
+Operation postResponse = ledgerClient.PostLedgerEntry(
+waitUntil: WaitUntil.Completed,
+    RequestContent.Create(
+        new { contents = "Hello world!" }));
+
+string content = postOperation.GetRawResponse().Content.ToString();
```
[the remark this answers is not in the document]
annelo-msft: I had done a bit of a rewrite of the sample in a different version of this PR, but wanted to keep this update clean to illustrate only the DynamicJson substitutions.  I'm happy to bring back a more opinionated rewrite if there's an appetite for it, but let me know.  I'm happy to go either way - there's value in both.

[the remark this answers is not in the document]
christothes: I'm fine with however we sequence the updates. However, in this case it appears that this is a totally unnecessary line of code. The others are minor renames for consistency.

sdk/confidentialledger/Azure.Security.ConfidentialLedger/README.md
KrzysztofCwalina: Oh, this sounds complicated. Maybe we don't want to do it after all. Yeah, let's talk about pros and cons tomorrow.

sdk/confidentialledger/Azure.Security.ConfidentialLedger/tests/samples/HelloWorldSamples.cs
annelo-msft: https://github.com/Azure/azure-sdk-for-net/issues/34085

sdk/confidentialledger/Azure.Security.ConfidentialLedger/README.md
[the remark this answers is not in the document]
annelo-msft: Created https://github.com/Azure/azure-sdk-for-net/issues/34085 for follow-ups.

PR #33932, opened 2023-02-07
Create call improvements

richardcho-msft:
- Add single target create call
- Update current create call to create group call
- add optional source identity in client
- remove client side validation as it's redundant with API


sdk/communication/Azure.Communication.CallAutomation/src/Models/CallInvite.cs
```diff
@@ -0,0 +1,58 @@
...
+        public CallInvite(CommunicationUserIdentifier targetIdentity)
+        {
+            Target = targetIdentity;
+        }
+
+        /// <summary>
+        /// Creates a new CallInvite object.
+        /// </summary>
+        /// <param name="targetIdentity"></param>
+        public CallInvite(MicrosoftTeamsUserIdentifier targetIdentity)
```
yyw-msft, line 35: nit: how come we use targetPhoneNumberIdentity as the name for PhoneNumberIdentifier but targetIdentity for both CommunicationUserIdentifier and MicrosoftTeamsUserIdentifier?

richardcho-msft: we can change the name. Do you prefer to have targetIdentity for all or something like targetTeamsIdentity

yyw-msft: hmmm, i think for consistency, we can name them all to targetIdentity?

PR #33940, opened 2023-02-07
use `GenerateTests` instead of `GenerateTest` in autorest

changlong-liu:
We currently have target `GenerateTests` in service.proj which leverage `GenerateTest` in Autorest.csharp.
And we got some feedbacks saying that inconsistent names make some confusion.

Since we also have target `GenerateCode` in service.proj which leverage `GenerateCode` in Autorest.csharp, I think we can also use singular name for test generation?

### DIFF
--- a/eng/scripts/CodeChecks.ps1
+++ b/eng/scripts/CodeChecks.ps1
@@ -90,7 +90,7 @@ try {
 
         Write-Host "Re-generating tests"
         Invoke-Block {
-            & dotnet msbuild $PSScriptRoot/../service.proj /restore /t:GenerateTest /p:SDKType=$SDKType /p:ServiceDirectory=$ServiceDirectory
+            & dotnet msbuild $PSScriptRoot/../service.proj /restore /t:GenerateTests /p:SDKType=$SDKType /p:ServiceDirectory=$ServiceDirectory
         }
     }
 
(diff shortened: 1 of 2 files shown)


jsquire:
> Since we also have target `GenerateCode` in service.proj which leverage `GenerateCode` in Autorest.csharp, I think we can also use singular name for test generation?

"GenerateCode" does not convey singular or plural.   "GenerateTest" implies that only a single test will be generated.    Unless I'm mistaken and this target only generates a single test, we should keep "GenerateTests", ideally updating AutoRest to match.

heaths:
The PR title should probably be changed. But I'm also confused. Isn't it already "GenerateTests"? So this PR should be a no-op, I would think.

archerzz:
> > Since we also have target `GenerateCode` in service.proj which leverage `GenerateCode` in Autorest.csharp, I think we can also use singular name for test generation?
> 
> "GenerateCode" does not convey singular or plural. "GenerateTest" implies that only a single test will be generated. Unless I'm mistaken and this target only generates a single test, we should keep "GenerateTests", ideally updating AutoRest to match.

We already got questions from a service team due to the inconsistency. The PR is to align the target name both in `service.proj` and `.csproj`. So maybe we should rename them all to `GenerateTests` since even for a single SDK, normally we generate more than one sample/test case.

changlong-liu:
> The PR title should probably be changed. But I'm also confused. Isn't it already "GenerateTests"? So this PR should be a no-op, I would think.

Thanks for catching @heaths . I plan to follow below steps to retire `GenerateTest`:
1. create target `dotnet build \t:GenerateTests` in autorest.csharp.
2. use GenerateTests  in azure-sdk-for-net
3. remove GenerateTest in autorest.csharp

I will update this PR as step 2 after the 1st step is done and merged.

PR #33955, opened 2023-02-07
Add standard metrics custom processor and request duration metric

vishweshbankwar:
This PR adds a custom span processor for extracting standard metrics. It also adds `request/duration` metric.

TODOs: 
1) Add remaining metrics.
2) Additional tests which includes updated sampler.
3) Update Changelog.

From end user perspective this will be enabled by default on configuring metric exporter.

Adding the complete design plan for extracting standard metrics below

**Proposal**: Standard metric collection in OpenTelemetry can be enabled by customizing following two parts in SDK:

**Sampler**: When the sampler decides to drop the data return SamplingResult.RecordOnly  instead of SamplingResult.None. This will achieve following two things:
Span will be created and will be populated with all the information (tags, events etc.)SampledFlag will still be set to false. So, this will not impact propagated sampling decision i.e. traceflags information.

**Custom Processor**:Add a custom processor which will extract standard metrics information as required. These metrics can then be reported as standard metrics using a metric exporter.


Badboilamba (approved):

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/StandardMetricsExtractionProcessor.cs
```diff
@@ -0,0 +1,102 @@
...
+                    break;
+                }
+            }
+
+            TagList tags = default;
+            tags.Add(new KeyValuePair<string, object>(StandardMetricConstants.RequestResultCodeKey, statusCodeAttributeValue));
+            tags.Add(new KeyValuePair<string, object>(StandardMetricConstants.MetricIdKey, StandardMetricConstants.RequestDurationMetricIdValue));
+            tags.Add(new KeyValuePair<string, object>(StandardMetricConstants.IsAutoCollectedKey, "True"));
+            tags.Add(new KeyValuePair<string, object>(StandardMetricConstants.IsSyntheticKey, "False"));
+            tags.Add(new KeyValuePair<string, object>(StandardMetricConstants.CloudRoleInstanceKey, StandardMetricResource.RoleInstance));
```
rajkumar-rangaraj, line 72: Most of these are common and does not change in every execution. Should we consider creating once?

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/AzureMonitorExporterTraceExtensions.cs
```diff
@@ -59,7 +64,14 @@ public static class AzureMonitorExporterTraceExtensions
                     configure(exporterOptions);
                 }
 
-                return new BatchActivityExportProcessor(new AzureMonitorTraceExporter(exporterOptions, credential));
+                var standardMetricReader = new PeriodicExportingMetricReader(new AzureMonitorMetricExporter(exporterOptions, credential))
```
rajkumar-rangaraj, line 65: What is the idea behind creating a new `PeriodicExportingMetricReader`? 
Is the plan to send data even if customer does not have metric exporter initialized?

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Customizations/Models/MetricDataPoint.cs
```diff
@@ -10,8 +10,15 @@ internal partial class MetricDataPoint
     {
         public MetricDataPoint(Metric metric, MetricPoint metricPoint)
         {
-            Name = metric.Name;
-            Namespace = metric.MeterName;
+            if (StandardMetricsExtractionProcessor.s_standardMetricNameMapping.TryGetValue(metric.Name, out var metricName))
```
rajkumar-rangaraj, line 13: What is the reason for this conditional check?

vishweshbankwar: OTel does not allow `/` in instrument names. So for `requests/duration` metric I have used a placeholder while creating instrument. That name is replaced back here.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/AzureMonitorExporterTraceExtensions.cs
vishweshbankwar: Yes - standard metric should work independent of customer's metric exporter

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/StandardMetricsExtractionProcessor.cs
vishweshbankwar: We could. However, only two of these will remain constant.

```csharp
tags.Add(new KeyValuePair<string, object>(StandardMetricConstants.MetricIdKey, StandardMetricConstants.RequestDurationMetricIdValue));
tags.Add(new KeyValuePair<string, object>(StandardMetricConstants.IsAutoCollectedKey, "True"));
```

cijothomas: for the constant one, model as resource.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/StandardMetricsExtractionProcessor.cs
```diff
@@ -0,0 +1,102 @@
...
+            tags.Add(new KeyValuePair<string, object>(StandardMetricConstants.CloudRoleNameKey, StandardMetricResource.RoleName));
+            tags.Add(new KeyValuePair<string, object>(StandardMetricConstants.RequestSuccessKey, RequestData.isSuccess(activity, statusCodeAttributeValue, OperationType.Http)));
+
+            // Report metric
+            _requestDuration?.Record(activity.Duration.TotalMilliseconds, tags);
```
cijothomas, line 77: why null check needed on requestDuration?

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/StandardMetricsExtractionProcessor.cs
```diff
@@ -0,0 +1,102 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#nullable disable // TODO: remove and fix errors
```
TimothyMothra, line 4: Since this is a new class, please remove this line.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/StandardMetricTests.cs
```diff
@@ -0,0 +1,68 @@
...
+    public class StandardMetricTests
+    {
+        [Fact]
+        public void ValidateRequestDurationMetric()
+        {
+            var activitySource = new ActivitySource(nameof(StandardMetricTests.ValidateRequestDurationMetric));
+            var traceTelemetryItems = new ConcurrentBag<TelemetryItem>();
+            var metricTelemetryItems = new ConcurrentBag<TelemetryItem>();
+            var exporterOptions = new AzureMonitorExporterOptions();
+            exporterOptions.ConnectionString = "InstrumentationKey=Ikey;IngestionEndpoint=https://westus2-0.in.applicationinsights.azure.com/;LiveEndpoint=https://westus2.livediagnostics.monitor.azure.com/";
```
TimothyMothra, line 26: nit: To keep this simple, I think you can remove the IngestionEndpoint and LiveEndpoint

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/StandardMetricTests.cs
```diff
@@ -0,0 +1,68 @@
...
+
+            using (var activity = activitySource.StartActivity("Test", ActivityKind.Server))
+            {
+                activity?.SetTag(SemanticConventions.AttributeHttpStatusCode, 200);
+            }
+
+            tracerProvider?.ForceFlush();
+            standardMetricCustomProcessor?.StandardMetricMeterProvider?.ForceFlush();
+
+            Assert.Single(traceTelemetryItems);
```
TimothyMothra, line 48: exporting Activity in a unit test can be flaky. you should use a SpinWait pattern here.
For example: https://github.com/Azure/azure-sdk-for-net/blob/4c575251241724be53423b43414842beed9695f7/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Integration.Tests/TestFramework/WebApplicationTestsBase.cs#L37-L48

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/StandardMetricsExtractionProcessor.cs
vishweshbankwar: typo

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/AzureMonitorExporterTraceExtensions.cs
vishweshbankwar: removed additional periodicExportingMetricReader

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/StandardMetricsExtractionProcessor.cs
rajkumar-rangaraj: rolename and roleNameInstance may not change.

rajkumar-rangaraj: As we discussed, this could be refactored in a different PR.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/StandardMetricsExtractionProcessor.cs
vishweshbankwar: will do a follow up on this, Still getting some errors.

rajkumar-rangaraj (approved):

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/StandardMetricsExtractionProcessor.cs
```diff
@@ -0,0 +1,95 @@
...
+            {
+                if (tag.Key == statusCodeAttribute)
+                {
+                    statusCodeAttributeValue = tag.Value.ToString();
+                    break;
+                }
+            }
+
+            TagList tags = default;
+            tags.Add(new KeyValuePair<string, object>(StandardMetricConstants.RequestResultCodeKey, statusCodeAttributeValue));
```
cijothomas, line 62: handle the case where `statusCodeAttributeValue` is null

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/StandardMetricsExtractionProcessor.cs
```diff
@@ -0,0 +1,95 @@
...
+            tags.Add(new KeyValuePair<string, object>(StandardMetricConstants.MetricIdKey, StandardMetricConstants.RequestDurationMetricIdValue));
+            tags.Add(new KeyValuePair<string, object>(StandardMetricConstants.IsAutoCollectedKey, "True"));
+            tags.Add(new KeyValuePair<string, object>(StandardMetricConstants.IsSyntheticKey, "False"));
+            tags.Add(new KeyValuePair<string, object>(StandardMetricConstants.CloudRoleInstanceKey, StandardMetricResource.RoleInstance));
+            tags.Add(new KeyValuePair<string, object>(StandardMetricConstants.CloudRoleNameKey, StandardMetricResource.RoleName));
```
cijothomas, line 67: this should be in resource.

cijothomas: more labels = more perf hit.
https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/test/Benchmarks/Metrics/HistogramBenchmarks.cs#L35-L38

vishweshbankwar: Need to look in to this. Resource attributes are not converted to telemetry custom dimensions right now. And standard metric requires these to be at the dimension level.

cijothomas: if resource attributes are not converted, then its already a bug in MetricExporter, that must be separately addressed. Atleast the rolename/roleinstance part from resource should be set as dimensions.

vishweshbankwar: > Atleast the rolename/roleinstance part from resource should be set as dimensions.

This gets converted to cloud role and role instance on telemetry level not as dimensions.

cijothomas: > > Atleast the rolename/roleinstance part from resource should be set as dimensions.
> 
> This gets converted to cloud role and role instance on telemetryItem level not as dimensions.

yes got it. that is incorrect and should be fixed.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Customizations/Models/MetricDataPoint.cs
rajkumar-rangaraj: We could remove few of the constant tags getting added to histogram and directly add it as a property. As there is a check here it makes it easier to understand for which items we need to add those constant properties.

Issue #31650, opened 2022-10-07
Distributed tracing [Zn] uber-task: GA OTel support in Azure.Core


lmolkova:
### Library name

Azure.Core

### Please describe the feature.

Here's a list of prereqs for GA on .NET

- [x] Support configurable OpenTelemetry [schema version](https://opentelemetry.io/docs/reference/specification/schemas/overview/#schema-version-number)
  - Stamp it on `az.schema_url` attribute on Activities reported via ActivitySource only
  - Support 1.17.0 initially only (TBD - will switch to latest by GA)
  - Map attribute names reported by SDKs to corresponding ones in configured schema version. Here's the current list https://github.com/Azure/azure-sdk/blob/main/docs/tracing/distributed-tracing-conventions.yml
  - We'll add new versions every once in a while and users will be able to configure them.
  - Don't change anything for the DiagnosticSource path

- [x] Cleanups and breaking changes
  - Propagate context using  `traceparent` in addition to `Diagnostic-Id` on AMQP SDKs 
  - https://github.com/Azure/azure-sdk-for-net/issues/31771
  - Enable nested scope suppression by default
  - http.status_code to int https://github.com/Azure/azure-sdk-for-net/issues/31438
  - investigate if "kind" and "otel.status_code" attributes  are needed on the ActivitySource path (use ActivityKind and Status enum if possible)
  - [x] https://github.com/Azure/azure-sdk-for-net/issues/33080

- [ ]  Testing
  - Retry tests: Activity per try and failed status
  - Perf testing: e.g. run storage perf tests with tracing enabled.
  - Update line-counter sample will latest and greatest OTel and AzMon exporter
  - [nice-to-have] Stress: e.g. enable tracing in stress tests

- [ ] Misc
  - Remove Experimental context switch

context: 
- https://github.com/microsoft/common-schema/pull/336/files
- https://gist.github.com/lmolkova/605bbf779d5599e8f01fd2c2e38f036e

JoshLove-msft:
Current version is 1.17.0

annelo-msft:
@lmolkova, is this work complete?  Can the issue be closed?

lmolkova:
I don't think we completed all the individual work items listed in the description. Let's keep it open for now and I'll check what's remaining.

lmolkova:
Closing in favor of https://github.com/Azure/azure-sdk-for-net/issues/46018

PR #33960, opened 2023-02-07
Set OTel schema url

JoshLove-msft:
Contributes to https://github.com/Azure/azure-sdk-for-net/issues/31650


sdk/core/Azure.Core/tests/ClientDiagnosticsTests.Net50.cs
```diff
@@ -329,6 +338,10 @@ public void CanSetActivitySourceAndDiagnosticSourceActivitiesTogether()
             CollectionAssert.Contains(diagnosticSourceActivity.Tags, new KeyValuePair<string, string>("Attribute2", "2"));
             CollectionAssert.Contains(diagnosticSourceActivity.Tags, new KeyValuePair<string, string>("Attribute3", "3"));
 
+            // Since both ActivitySource and DiagnosticSource listeners are used, we should see the az.schema_url tag set even in diagnostic source because they use the same
```
JoshLove-msft, line 341: I think this is expected.. @lmolkova please keep me honest

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
```diff
@@ -292,8 +292,11 @@ public void AddLink(string traceparent, string tracestate, IDictionary<string, s
             public Activity? Start()
             {
                 _currentActivity = StartActivitySourceActivity();
-
-                if (_currentActivity == null)
+                if (_currentActivity != null)
+                {
+                    _currentActivity.AddTag("az.schema_url", "https://opentelemetry.io/schemas/1.17.0");
```
lmolkova, line 297: I suggest bundling this change with mappings and actually complying to otel schema 1.17.0, (we're not compliant now)

Changes needed (https://github.com/Azure/azure-sdk/blob/main/docs/tracing/distributed-tracing-conventions.yml)


HTTP:
- `serviceRequestId` -> `az.service_request_id`
- `clientRequestId` -> `az.client_request_id`
- add `net.peer.name` (endpoint)
- add `net.peer.port` (if not 443), i.e. probably never, but it should be of int type
- http.status_code should be int

Messaging:
- add `messaging.system` - either 'eventhubs' or 'servicebus'
- add [`messaging.operation`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md#operation-names) (one of `publish`, `receive` or `process`)
- `message_bus.destination` -> `messaging.destination.name`
- `peer.address` -> `net.peer.name`

Or, if you prefer to keep PRs small and would be sending others with attribute changes before next core release, we can do this as well :)

sdk/core/Azure.Core/tests/ClientDiagnosticsTests.Net50.cs
lmolkova: yep, if both are used, there is nothing we can do to keep DiagSource path unchanged. But it's not a common (or valid) scenario to have both.

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
JoshLove-msft: I've added the HTTP changes in this PR other than enabling suppression by default (but perhaps that is already done?). I will do the messaging changes in a separate PR.

sdk/core/Azure.Core/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### Features Added
 
+- ActivitySource activities that are used when using the [experimental OpenTelemetry support](https://devblogs.microsoft.com/azure-sdk/introducing-experimental-opentelemetry-support-in-the-azure-sdk-for-net/) will include the `az.schema_url` tag indicating the OpenTelemetry schema version.
```
JoshLove-msft, line 7: Todo - document the breaking changes.

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
```diff
@@ -232,7 +237,7 @@ public void AddTag(string name, string value)
                 }
                 else
                 {
-                    _currentActivity?.AddTag(name, value!);
+                    _currentActivity?.AddObjectTag(name, value);
```
JoshLove-msft, line 240: This method already existed thanks to @pakrym

sdk/core/Azure.Core/tests/RequestActivityPolicyTests.cs
```diff
@@ -313,21 +313,31 @@ public async Task ActivitySourceActivityStartedOnRequest()
...
+
+                CollectionAssert.DoesNotContain(activity.TagObjects, new KeyValuePair<string, string>("requestId", clientRequestId));
+                CollectionAssert.Contains(activity.TagObjects, new KeyValuePair<string, string>("az.client_request_id", clientRequestId));
+
+                CollectionAssert.DoesNotContain(activity.TagObjects, new KeyValuePair<string, string>("serviceRequestId", "server request id"));
+                CollectionAssert.Contains(activity.TagObjects, new KeyValuePair<string, string>("az.service_request_id", "server request id"));
+
+                CollectionAssert.Contains(activity.TagObjects, new KeyValuePair<string, string>("net.peer.name", "example.com"));
+
+                CollectionAssert.DoesNotContain(activity.TagObjects, new KeyValuePair<string, int>("net.peer.port", 443));
```
JoshLove-msft, line 338: Add a test where the port is not 443.

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
```diff
@@ -292,8 +297,11 @@ public void AddLink(string traceparent, string tracestate, IDictionary<string, s
             public Activity? Start()
             {
                 _currentActivity = StartActivitySourceActivity();
-
-                if (_currentActivity == null)
+                if (_currentActivity != null)
+                {
+                    _currentActivity.AddTag("az.schema_url", "https://opentelemetry.io/schemas/1.17.0");
```
lmolkova, line 302: nit: my crystal ball tells me we'll need to support another version soon, so it'd be nice to move it to some constant and update all at once along with tests :)

lmolkova (approved):
Thank you!

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
JoshLove-msft: Looks like we also need messaging.batch.message_count for batch operations in messaging.

JoshLove-msft: Potentially also message.destination.name and message.source.name?
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md#producer-attributes

PR #33962, opened 2023-02-07
[AzureMonitorExporter] cleanup nullable in Exporter classes and Transmitter

TimothyMothra:
Towards #34013

## Changes
- remove `#nullable disable` in Exporter classes and resolve issues
- remove `#nullable disable` in Transmitter class and resolve issues

### Notes
The biggest change here is that the `AzureMonitorResource` in the `Exporter` could be null.
This is provided to the `TelemetryItem` constructor which is eventually used to set Tags.
The Tags were already checking for null so I think this change is safe.
https://github.com/Azure/azure-sdk-for-net/blob/b6eb01b26613b30808eddb5a26402613e03b714e/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Customizations/Models/TelemetryItem.cs#L99-L102

### DIFF
--- a/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/AzureMonitorTransmitter.cs
+++ b/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/AzureMonitorTransmitter.cs
@@ -1,8 +1,6 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-#nullable disable // TODO: remove and fix errors
-
 using System;
 using System.Collections.Generic;
 using System.Threading;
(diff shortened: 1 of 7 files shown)


rajkumar-rangaraj (approved):
LGTM.

@vishweshbankwar Could you please take a look and confirm.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/AzureMonitorTransmitter.cs
```diff
@@ -64,7 +62,7 @@ public AzureMonitorTransmitter(AzureMonitorExporterOptions options, TokenCredent
             {
                 try
                 {
-                    var storageDirectory = options.StorageDirectory ?? StorageHelper.GetDefaultStorageDirectory();
+                    var storageDirectory = options.StorageDirectory ?? StorageHelper.GetDefaultStorageDirectory() ?? throw new InvalidOperationException("Unable to determine offline storage directory.");
```
vishweshbankwar, line 65: behavior change?

TimothyMothra: it looks like that, but not really a change.

`StorageHelper.GetDefaultStorageDirectory` can fail and return null.
That then causes the `new FileBlobProvider` constructor to throw.

vishweshbankwar: but we were not throwing before right?

rajkumar-rangaraj: This throw does not crash the process. Catch block logs into event source. If this is acting as a blocker for this PR better to remove the changes and track it on an issue.

vishweshbankwar: My bad - I was not reading it properly.

vishweshbankwar (approved):

PR #33963, opened 2023-02-07
[FormRecognizer] Aggregate empty types in a single file

### LINKED ISSUE #29108: [FormRecognizer] Aggregate empty types in a single file
See https://github.com/Azure/azure-sdk-for-net/pull/29100#discussion_r889388997 for context.

kinelski:
Just moving types around to keep things clean. Following the pattern in `TextAnalytics`:
- Internal empty partial declarations were moved to the `GeneratorStubs/InternalTypes.cs` file.
- Public empty partial declarations were moved to the `GeneratorStubs/PublicTypes.cs` file.

These empty partial declarations are used to overwrite the behavior of the .NET code generator. For example, if a type is generated as `public` by default, we can manually add an `internal` partial declaration for that type to let the generator know it should have an `internal` access modifier.

Fixes https://github.com/Azure/azure-sdk-for-net/issues/29108.


sdk/formrecognizer/Azure.AI.FormRecognizer/src/autorest.md
```diff
@@ -159,13 +159,12 @@ input-file:
...
 ```
 
-## Make generated models internal by default
+## Move generated models to the DocumentAnalysis namespace
 ``` yaml
 directive:
   from: https://raw.githubusercontent.com/Azure/azure-rest-api-specs/2fd7dcd89afa70ff5ba7be88bee987da62099a28/specification/cognitiveservices/data-plane/FormRecognizer/stable/2022-08-31/FormRecognizer.json
   where: $.definitions.*
   transform: >
-    $["x-accessibility"] = "internal";
```
kinelski, line 168: We were forcibly making generated models internal by default. This is a pattern we inherited from the old `FormRecognizerClient` because the convenience layer we used at the time was reasonably different from the generated layer. For that reason, most generated models were made internal and had a public wrapper type manually written around them.

With the new `DocumentAnalysisClient` we represent models very similarly to what we get from the service. This autorest transform doesn't help much given that most models will be made public anyway, so we're removing it.

pallavit: Thanks for cleaning this. Any reason why we are using so many transformations? Should we not consider requesting the service team to update the swagger, if possible?
Also, is this only needed by the  .NET language or others as well. if everyone, there is further benefit in making the change up the stream.

pallavit:
> ### AutoRest Configuration

Also I see multiple versions of the input-file, curious how do you point dotnet build to a particular version (service 2.x) vs service (3.x). Does it pick up the most recent one by default?

---
Refers to: sdk/formrecognizer/Azure.AI.FormRecognizer/src/autorest.md:7 in deac315. [](commit_id = deac3150d8a693a53f4489da2c34bb6078e8b303, deletion_comment = False)

pallavit:
> 

Consider explaining in a comment why we have these files in a single place?

---
Refers to: sdk/formrecognizer/Azure.AI.FormRecognizer/src/GeneratorStubs/InternalTypes.cs:47 in deac315. [](commit_id = deac3150d8a693a53f4489da2c34bb6078e8b303, deletion_comment = False)

pallavit (approved):
:shipit:

kinelski:
> Also I see multiple versions of the input-file, curious how do you point dotnet build to a particular version (service 2.x) vs service (3.x). Does it pick up the most recent one by default?

Both versions are generated side by side.

We point to two different versions because there are two totally different swaggers. As you may have heard, last year the service drastically changed the structure of their response between versions 2.x and 3.x, introducing multiple breaking changes.

To avoid breaking changes in the SDK we need to continue supporting old service versions, so we generate both the old and the new swaggers and build two different convenience layers on top of them: the old `FormRecognizerClient` for 2.x; and the new `DocumentAnalysisClient` for 3.x.

pallavit:
> To avoid breaking changes in the SDK we need to continue supporting old service versions, so we generate both the old and the new swaggers and build two different convenience layers on top of them: the old `FormRecognizerClient` for 2.x; and the new `DocumentAnalysisClient` for 3.x.

Got it. Thank you for the explanation :). I have a couple of follow-up questions, will take to you about it offline.

sdk/formrecognizer/Azure.AI.FormRecognizer/src/autorest.md
kinelski: - Most of the transforms we had for v2.x (old service) were caused by minor inconsistencies between the swagger and the service response. In particular, they used to return some fields as `null` even though this was not allowed without setting the `x-nullable` property to `true` in the swagger specification. We've brought up this matter a couple of times but it has never been fixed on their side, so we had no choice but to keep the transforms. When running in debug mode, our generated code does an assertion and throws if the service sends back an unexpected `null`, which may not be true for other languages.
- For v3.x (new service), most transformations are used to avoid conflicts between the new and the old models since the service kept the same name for some of them. Also, we're moving models to the new `DocumentAnalysis` namespace.

Issue #33969, opened 2023-02-08
Build snippets for core


heaths:
Due to a number of issues below, `BuildSnippets: false` is passed in _sdk/core/ci.yml_. Eventually, we want to build snippets by default but these issues need to be fixed first:

- [ ] #33945
- [x] #33967
- [x] #33970
- [x] #33971
- [x] #33973
- [x] #33974
- [x] #33975
- [x] #33977
- [x] #33978
- [x] #33979
- [x] #33982
- [x] #33983
- [x] #33985
- [x] #33986
- [x] #33987
- [x] #33988
- [x] #33976
- [x] #34016

Separately - and this can be done at any time - we can consider:

- [x] #33968

To find outstanding issues, run:

```bash
dotnet build eng/service.proj -warnaserror /t:rebuild /p:DebugType=none /p:SDKType=all /p:ServiceDirectory=* /p:IncludePerf=false /p:IncludeStress=false /p:PublicSign=false /p:Configuration=Release /p:EnableSourceLink=false /p:BuildSnippets=true /bl
```

That command does not use _sdk/*/ci.yml_ or _sdk/*/ci.mgmt.yml, so do not consider any errors emitted to be breaking - just that the source issues haven't yet been resolved.

heaths:
#33964 is related, such that Storage disabled building snippets already but that will still fail the `net - core - ci` pipeline.

Amoon-123:
dotnet build eng/service.proj -warnaserror /t:rebuild /p:DebugType=none /p:SDKType=all /p:ServiceDirectory=* /p:IncludePerf=false /p:IncludeStress=false /p:PublicSign=false /p:Configuration=Release /p:EnableSourceLink=false /p:BuildSnippets=true /bl

PR #33989, opened 2023-02-08
Disable building snippets for more libraries

heaths:
Relates to #33969

### DIFF
--- a/sdk/tables/ci.yml
+++ b/sdk/tables/ci.yml
@@ -25,6 +25,7 @@ extends:
   template: /eng/pipelines/templates/stages/archetype-sdk-client.yml
   parameters:
     ServiceDirectory: tables
+    BuildSnippets: false
     ArtifactName: packages
     Artifacts:
     - name: Azure.Data.Tables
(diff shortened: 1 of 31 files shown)


jsquire (approved):

sdk/tables/ci.yml
```diff
@@ -25,6 +25,7 @@ extends:
   template: /eng/pipelines/templates/stages/archetype-sdk-client.yml
   parameters:
     ServiceDirectory: tables
+    BuildSnippets: false
```
christothes, line 28: You can revert this - I fixed the snippets in https://github.com/Azure/azure-sdk-for-net/pull/33998

m-nash (approved):

sdk/tables/ci.yml
heaths: As long as the validation on this PR takes, I'd rather wait till after so it just touches the _sdk/tables_ directory. I'll leave this unresolved for now so I don't forget while the 1 failed run on MacOS re-runs.

heaths:
@christothes discovered the source of the breaks on macOS (or maybe just because that was the only agent that exercised those tests) and is prepping a separate PR now. As such, overriding this one since every other test passed.

PR #33990, opened 2023-02-08
Azure OpenAI: Add support for faster, streaming Completions responses for ChatGPT-style use

trrwilson:
This change introduces support for *streaming Completions* to the Azure OpenAI C# client library. This allows Completions responses to provide incremental result data and facilitate latency-sensitive use of ChatGPT-style use cases where waiting many seconds for a full response to finish its entirety is prohibitive.

OpenAI's public documentation for REST/etc. covers this as a boolean `stream` parameter on the Completions request payload: https://platform.openai.com/docs/api-reference/completions/create

In the client library, this is projected via supplementary methods and objects:

- `OpenAIClient` gets a new pair of methods, `GetCompletionsStreaming` and `GetCompletionsStreamingAsync`, that parallel their non-streaming counterparts but provide a new `StreamingCompletions` type as the response payload.
- The `StreamingCompletions` response value will be created and usable as soon as the response stream is available and provides a `GetChoicesStreaming()` method to begin asynchronous enumeration of `StreamingChoice` objects -- these are populated as new server-sent event messages arrive with new choice data (as delineated by the 'index' field)
- `StreamingChoice`, in turn, has a `GetTextStreaming()` method that will asynchronously enumerate the streamed text components (typically each token in Completions.Choice) as they become available
- Common concepts between streaming and non-streaming Completions/Choices (ID, Created, Logprobs, etc.) are reimplemented/exposed from the base, deserialized SSE payloads while inapplicable/unavailable concepts (like Usage) are omitted

In implementation, this is accomplished by disabling HttpMessage buffering in the Completions request and then forking a fire-and-forget Task (owned by StreamingCompletions) to pump the response content stream and update collections/enumerators.


sdk/openai/Azure.AI.OpenAI/tests/OpenAIInferenceTests.cs
```diff
@@ -94,5 +110,32 @@ public void RequestFailedExceptionTest()
             var exception = Assert.ThrowsAsync<RequestFailedException>(async () => { await client.GetCompletionsAsync("BAD_DEPLOYMENT_ID", completionsRequest); });
             Assert.AreEqual(404, exception.Status);
         }
+
+        [RecordedTest]
+        public async Task StreamingCompletionsTest()
```
glharper, line 115: We need Asserts in tests, right?

sdk/openai/Azure.AI.OpenAI/tests/OpenAIInferenceTests.cs
```diff
@@ -94,5 +110,32 @@ public void RequestFailedExceptionTest()
...
+
+            fileWriter.WriteLine($"{Environment.TickCount} : START");
+            Response<StreamingCompletions> response = await client.GetStreamingCompletionsAsync(
+                CompletionsDeploymentId,
+                requestOptions);
+
+            await foreach (string message in response.Value.GetAsyncMessages())
+            {
+                fileWriter.WriteLine($"{Environment.TickCount} : {message}");
+            }
```
glharper, line 138: Do we want to model disposing the response properly (even if it doesn't matter for this test)?

sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs
```diff
@@ -42,5 +45,73 @@ public virtual Response<Completions> GetCompletions(string deploymentId, string
...
+            Argument.AssertNotNull(completionsOptions, nameof(completionsOptions));
+
+            using DiagnosticScope scope = ClientDiagnostics.CreateScope(
+                "OpenAIClient.GetStreamingCompletions");
+            scope.Start();
+
+            RequestContext context = FromCancellationToken(cancellationToken);
+
+            RequestContent nonStreamingContent = completionsOptions.ToRequestContent();
+            RequestContent streamingContent = GetStreamingEnabledRequestContent(nonStreamingContent);
```
KrzysztofCwalina, line 64: it's fine for the preview, but we should fix it at some point, i.e. we should generate the right content the first time (line above) as opposed to rewriting it.

sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs
```diff
@@ -42,5 +45,73 @@ public virtual Response<Completions> GetCompletions(string deploymentId, string
...
+            augmentedContent.JsonWriter.WriteStartObject();
+
+            // Copy the original JSON content back into the new copy
+            foreach (JsonProperty jsonThing in originalJsonRoot.EnumerateObject())
+            {
+                augmentedContent.JsonWriter.WritePropertyName(jsonThing.Name);
+                jsonThing.Value.WriteTo(augmentedContent.JsonWriter);
+            }
+
+            // ...Add the *one thing* we wanted to add
```
KrzysztofCwalina, line 108: @annelo-msft, here is an interesting use case for your mutable JSON APIs. For context, this routine is rewriting request content just to add one property to the JSON payload.

sdk/openai/Azure.AI.OpenAI/src/Custom/Utf8TStreamReader.cs
```diff
@@ -0,0 +1,65 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.IO;
+using System.Threading;
+
+namespace Azure.Core.Sse
+{
+    public sealed class Utf8TStreamReader : IDisposable
```
KrzysztofCwalina, line 10: You don't need this type in the preview. I just started to work on it to avoid unecessary UTF8-UTF16-UTF8 conversions, but it's not yet used by the reader.

sdk/openai/Azure.AI.OpenAI/tests/OpenAIInferenceTests.cs
KrzysztofCwalina: Yeah, we don't want incorrect code to make it to the internet these days. OpenAI will learn wrong patterns from these tests, suggest them to unsuspecting humans who will soon be developing using copy-paste-oriented-programming, which in turn will make OpenAI not work properly :-)

sdk/openai/Azure.AI.OpenAI/src/Custom/StreamingCompletions.cs
```diff
@@ -0,0 +1,81 @@
...
+        private bool _disposedValue;
+        private SseReader _reader;
+
+        internal StreamingCompletions(Response response)
+        {
+            _baseResponse = response;
+            _reader = new SseReader(response.ContentStream);
+        }
+
+        public async IAsyncEnumerable<string> GetAsyncMessages()
```
KrzysztofCwalina, line 23: Async should be a suffix. It's a .NET idiom.

sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs
annelo-msft: Love it!  Thank you, @KrzysztofCwalina!

sdk/openai/Azure.AI.OpenAI/tests/OpenAIInferenceTests.cs
trrwilson: Agreed on all counts, but now I'm curious: is there (or should there be) an `IDisposable` `Response<T>` equivalent wrapper? Without one, it seems like we're relegated to the extremely clunky, old-style try/finally resource management pattern, i.e.
```csharp
Response<SomeDisposableType> response = client.GetResponse();
try
{
    // ... use response
}
finally
{
    response.Value?.Dispose();
}
```
Using the implicit `<T>` operator gets around this (e.g. `using SomeDisposableType responseValue = client.GetResponse()` will work) but if you want the `Response<T>` itself (e.g. for the raw payload) then I don't see a way to use `using` and all that goodness.

sdk/openai/Azure.AI.OpenAI/tests/OpenAIInferenceTests.cs
trrwilson: Absolutely; I originally rushed this draft out right before my few days away and should have made that more clear in the description. The tests have asserts now (and quite a few of them!).

sdk/openai/Azure.AI.OpenAI/tests/OpenAIInferenceTests.cs
trrwilson: For the moment, I updated the test to use/demonstrate the declare -> try -> finally pattern mentioned. I'm now curious about whether or not OpenAI would end up being smart enough to recommend that properly!

sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs
```diff
@@ -42,5 +45,73 @@ public virtual Response<Completions> GetCompletions(string deploymentId, string
             completionsOptions.Prompt.Add(prompt);
             return GetCompletions(deploymentId, completionsOptions, cancellationToken);
         }
+
+        public virtual Response<StreamingCompletions> GetStreamingCompletions(
```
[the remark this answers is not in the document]
trrwilson: Thanks for the pointer! I updated these to `GetCompletionsStreaming` and `GetCompletionsStreamingAsync`.

sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs
trrwilson: Agreed, that would be super duper nice. We'll hopefully be able to address our instance issue with an internal generated field or the like, but having a generalized capability to mutate payload in this partial class setup with custom/generated interrelationship would be fantastic.

sdk/openai/Azure.AI.OpenAI/src/Custom/Utf8TStreamReader.cs
trrwilson: Thanks; should've looked more carefully during the port. Removed for now.

sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs
trrwilson: Agreed on both points -- I'd like to integrate the in-progress proper approach ASAP (maybe even before this PR is done) but won't hesitate to use this frowny-face-inducing workaround if it's the holdout.

sdk/openai/Azure.AI.OpenAI/tests/OpenAIInferenceTests.cs
KrzysztofCwalina: When I discussed it with @tg-msft, we thought that for the preview response.Value would be disposable, so the called would do:
```csharp
using Model result = client.GetResponse().Value;
```

sdk/openai/Azure.AI.OpenAI/src/Custom/StreamingCompletions.cs
trrwilson: Thanks -- this method isn't intended to actually exist/persist beyond the first data flow demonstration, but that's no excuse for even a code-sharing draft PR to flaunt proper patterns.

sdk/openai/Azure.AI.OpenAI/tests/OpenAIInferenceTests.cs
KrzysztofCwalina: After the preview, we will think about adding a disposable `Response<T>` variant/subtype. We don't want to make `Response<T>` disposable as most responses don't have to be disposed but if we implemented `IDisposable` on `Response<T>`, users would be compelled to dispose it.

trrwilson: Makes sense -- and I for sure agree that having a (fabricated name) `public class DisposableResponse<T> : Response<T> where T : IDisposable` would be the way to go vs. changing any existing `Respose` type.

Applying `using` to the field of the top-level `Response<T>` still has some rough edges (inappropriate scoping of the `using` statement could cause a still-valid `Response<T>` to have a disposed/invalid `.Value` field) but I certainly like that better than declare/try/finally. I'll apply that in the test code.

sdk/openai/Azure.AI.OpenAI/src/Custom/Choice.Serialization.cs
```diff
@@ -49,7 +49,7 @@ internal static Choice DeserializeChoice(JsonElement element)
                         continue;
                     }
                 }
-                if (property.NameEquals("finishReason"))
+                if (property.NameEquals("finish_reason"))
```
KrzysztofCwalina, line 52: This is probably not very important, but this comparison will be faster if you do:
```csharp
if (property.NameEquals("finish_reason"u8)) ...
```

JsonDocument stores the data as UTF8. If you call the UTF16 taking overload, the overload will convert the UTF16 string to UTF8 on the fly, before actually comparing the strings. But you can just append `u8` to the end of string literals to make them UTF8 at compile time.

trrwilson: That's a great observation. We can optimize it in our custom deserialization, but I suspect that having the C# emitter aware of this (for the generated code coming out of cadl) would be even better; I believe even these "custom" deserializes are just lightly modified generated code.

@m-nash, do you know if smart use of UTF8 string literals is already something on the radar for C# emission? Should I file an issue to track the request?

sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs
```diff
@@ -42,5 +45,99 @@ public virtual Response<Completions> GetCompletions(string deploymentId, string
...
+
+            RequestContext context = FromCancellationToken(cancellationToken);
+
+            RequestContent nonStreamingContent = completionsOptions.ToRequestContent();
+            RequestContent streamingContent = GetStreamingEnabledRequestContent(nonStreamingContent);
+
+            try
+            {
+                HttpMessage message = CreateGetCompletionsRequest(deploymentId, streamingContent, context);
+                message.BufferResponse = false;
```
tg-msft, line 69: You should be able to use `x-csharp-buffer-response` to have AutoRest generate this for you - if your version of the Cadl tooling allow a decorator like `@extension("x-csharp-buffer-response", false)`.

sdk/openai/Azure.AI.OpenAI/tests/OpenAIInferenceTests.cs
```diff
@@ -100,11 +104,93 @@ public async Task EmbeddingTest()
...
-            var exception = Assert.ThrowsAsync<RequestFailedException>(async () => { await client.GetCompletionsAsync("BAD_DEPLOYMENT_ID", completionsRequest); });
+            RequestFailedException exception = Assert.ThrowsAsync<RequestFailedException>(async () =>
+            {
+                await client.GetCompletionsAsync("BAD_DEPLOYMENT_ID", completionsRequest);
+            });
             Assert.AreEqual(404, exception.Status);
         }
+
+        [RecordedTest]
+        public async Task StreamingCompletionsTest()
```
scbedd, line 118: here.

sdk/openai/Azure.AI.OpenAI/src/Custom/Choice.Serialization.cs
trrwilson: Resolving for the purposes of this PR, in any event -- this change was ported from a different PR that's already merged (i.e. this file isn't being changed here now).

sdk/openai/Azure.AI.OpenAI/tests/OpenAIInferenceTests.cs
trrwilson: Yep, this is the place!
[posterity note: discussion about where the test is and code review bookmarking]

sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs
trrwilson: That'll be a *much* nicer way of achieving this and cut down on the custom code significantly. I believe we're not quite there yet with the cadl, but should follow up ASAP to converge. I'll move forward with it as-is here for expediency's sake.

sdk/openai/Azure.AI.OpenAI/src/Custom/AsyncAutoResetEvent.cs
```diff
@@ -0,0 +1,49 @@
...
+// Licensed under the MIT License.
+
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Azure.AI.OpenAI.Custom
+{
+    // Adapted from https://devblogs.microsoft.com/pfxteam/building-async-coordination-primitives-part-2-asyncautoresetevent/
+    internal class AsyncAutoResetEvent
```
m-nash, line 11: In general classes in the `Custom` folder are meant to mirror `Generated` classes.  This class can go in root or a sub folder named whatever we like.

sdk/openai/Azure.AI.OpenAI/src/Custom/StreamingCompletions.cs
```diff
@@ -0,0 +1,173 @@
...
+using System.Linq;
+using System.Runtime.CompilerServices;
+using System.Text.Json;
+using System.Threading;
+using System.Threading.Tasks;
+using Azure.Core.Sse;
+
+namespace Azure.AI.OpenAI.Custom
+{
+    public class StreamingCompletions : IDisposable
```
m-nash, line 16: Should we represent these models in the cadl?  If that is the plan for future lets double check the shape to ensure that when we add them they can be removed from custom code and generated

sdk/openai/Azure.AI.OpenAI/src/Custom/SseReader.cs
```diff
@@ -0,0 +1,147 @@
...
+            }
+        }
+        public void Dispose()
+        {
+            Dispose(disposing: true);
+            GC.SuppressFinalize(this);
+        }
+    }
+
+    public readonly struct SseLine
```
m-nash, line 123: should have its own file unless its private and only used by ssereader in which case lets make it private.

sdk/openai/Azure.AI.OpenAI/src/Custom/SseReader.cs
```diff
@@ -0,0 +1,147 @@
...
+
+        private static bool TryParseLine(string lineText, out SseLine line)
+        {
+            if (lineText.Length == 0)
+            {
+                line = default;
+                return false;
+            }
+
+            int colonIndex = lineText.IndexOf(':');
```
m-nash, line 91: I am pretty sure lineText.AsSpan().IndexOf is faster here.

sdk/openai/Azure.AI.OpenAI/src/Custom/SseReader.cs
```diff
@@ -0,0 +1,147 @@
...
+// Licensed under the MIT License.
+
+using System;
+using System.IO;
+using System.Threading.Tasks;
+
+namespace Azure.Core.Sse
+{
+    // SSE specification: https://html.spec.whatwg.org/multipage/server-sent-events.html#parsing-an-event-stream
+    public sealed class SseReader : IDisposable
```
m-nash, line 11: Does this need to be public?

sdk/openai/Azure.AI.OpenAI/src/Custom/AsyncAutoResetEvent.cs
trrwilson: Got it! I'll move the new, non-mirrored files into a folder named "Streaming." They may move back later when we can generate a lot of it, but better to be correct at every stage.

sdk/openai/Azure.AI.OpenAI/src/Custom/SseReader.cs
trrwilson: Agreed. `SseLine` is part of the outward-facing contract, so I moved it into a new `SseLine.cs` file.

sdk/openai/Azure.AI.OpenAI/src/Custom/SseReader.cs
trrwilson: OK, I have to confess that my initial reaction to that was "what? no way that can be true...". But I think you're completely right!
https://dotnetfiddle.net/PCc1ct

Using the span consistently looks to shave off at least 10-15% of the runtime. Pragmatically, the difference here is immaterial, but just out of principle and "that's cool" I'll make the update anyway :)

Courtesy of the OpenAI-empowered new Bing:
> String.AsSpan().IndexOf is faster than String.IndexOf in C# because it avoids memory allocation and garbage collection. Here is why:

> String.AsSpan() converts a string into a Span<char>[1](https://www.jitbit.com/alexblog/309-improving-c-performance-with-spant/)[2](https://programmingfire.com/what-is-span-in-csharp-and-how-it-improves-the-performance)[3](https://www.codeproject.com/Articles/5269747/Using-Span-T-to-improve-performance-of-Csharp-code), which is a ref struct[4](https://programmingfire.com/what-is-span-in-csharp-and-how-it-improves-the-performance)[2](https://programmingfire.com/what-is-span-in-csharp-and-how-it-improves-the-performance) that can point to a region of memory without copying it[2](https://programmingfire.com/what-is-span-in-csharp-and-how-it-improves-the-performance)[3](https://www.codeproject.com/Articles/5269747/Using-Span-T-to-improve-performance-of-Csharp-code)[5](https://learn.microsoft.com/en-us/archive/msdn-magazine/2018/january/csharp-all-about-span-exploring-a-new-net-mainstay)[6](https://learn.microsoft.com/en-us/dotnet/api/system.span-1?view=net-7.0). Span<char> can be allocated on the stack or inline in the containing type[2](https://programmingfire.com/what-is-span-in-csharp-and-how-it-improves-the-performance)[5](https://learn.microsoft.com/en-us/archive/msdn-magazine/2018/january/csharp-all-about-span-exploring-a-new-net-mainstay)[6](https://learn.microsoft.com/en-us/dotnet/api/system.span-1?view=net-7.0), which means it does not require garbage collection[4](https://programmingfire.com/what-is-span-in-csharp-and-how-it-improves-the-performance)[2](https://programmingfire.com/what-is-span-in-csharp-and-how-it-improves-the-performance).
Span<char>.IndexOf performs a non-copying slice operation on the span to find the index of a character or a substring[1](https://www.jitbit.com/alexblog/309-improving-c-performance-with-spant/)[7](https://www.dotnetperls.com/span)[5](https://learn.microsoft.com/en-us/archive/msdn-magazine/2018/january/csharp-all-about-span-exploring-a-new-net-mainstay). This means it does not create a new string object or allocate any memory on the heap[1](https://www.jitbit.com/alexblog/309-improving-c-performance-with-spant/)[7](https://www.dotnetperls.com/span)[5](https://learn.microsoft.com/en-us/archive/msdn-magazine/2018/january/csharp-all-about-span-exploring-a-new-net-mainstay).
String.IndexOf, on the other hand, performs a substring operation on the string to find the index of a character or a substring[8](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1846)[9](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1831)[10](https://stackoverflow.com/questions/51864673/c-sharp-readonlyspanchar-vs-substring-for-string-dissection). This means it creates a new string object and allocates memory on the heap[8](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1846)[9](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1831)[10](https://stackoverflow.com/questions/51864673/c-sharp-readonlyspanchar-vs-substring-for-string-dissection), which requires garbage collection later[8](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1846)[9](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1831)[10](https://stackoverflow.com/questions/51864673/c-sharp-readonlyspanchar-vs-substring-for-string-dissection).
Therefore, String.AsSpan().IndexOf is faster than String.IndexOf in C# because it avoids memory allocation and garbage collection by using Span<char>, which is a ref struct that can point to a region of memory without copying it.

sdk/openai/Azure.AI.OpenAI/src/Custom/SseReader.cs
trrwilson: Certainly not. I'll swap this (and the moved `SseLine`) to `internal`.

sdk/openai/Azure.AI.OpenAI/src/Custom/StreamingCompletions.cs
trrwilson: Absolutely--the goal is to do as much of this as possible from cadl as soon as possible. I'll follow up on the chat thread(s) about double-checking the surface; for preview, the scales likely tilt towards getting streaming support in even if we break it in a bit, but that doesn't mean we shouldn't avoid low-hanging fruit splatting down on us if we can.

trrwilson: Per quick response off-PR, this'll be an important consideration between now and exiting preview -- but not a blocker for this PR so much as an important note.

jpalvarezl (approved):

sdk/openai/Azure.AI.OpenAI/src/Custom/SseReader.cs
m-nash: I think if you added StringComparison.Ordinal it would be similar but I do think span still has a tiny edge of that if I recall correctly.

sdk/openai/Azure.AI.OpenAI/src/Streaming/StreamingChoice.cs
```diff
@@ -0,0 +1,115 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.CompilerServices;
+using System.Threading;
+
+namespace Azure.AI.OpenAI.Custom
+{
+    public class StreamingChoice
```
m-nash, line 12: Since this and `StreamingCompletions` are intended to eventually be in the cadl and therefore be generated these should go in the Custom folder.

trrwilson: Thanks--I was curious about doing that in advance or not. I moved `StreamingCompletions` and `StreamingChoice` into `Custom` and the remaining new files into a `Helpers` folder (as the former "Streaming" identity got weaker).

m-nash (approved):

PR #34001, opened 2023-02-08
Update Azure.Core docs for dynamic JSON

annelo-msft:
Add customer-facing documentation for dynamic JSON types.

Includes updates to existing docs to address warnings from markdown linting tools.


sdk/core/Azure.Core/README.md
```diff
@@ -1,20 +1,20 @@
 # Azure Core shared client library for .NET
 
-Azure.Core provides shared primitives, abstractions, and helpers for modern .NET Azure SDK client libraries.
```
annelo-msft, line 3: Changes to this file are just for linting.

sdk/core/Azure.Core/samples/ProtocolMethods.md
```diff
@@ -118,6 +118,14 @@ var doc = JsonDocument.Parse(response.Content.ToMemory());
 var name = doc.RootElement.GetProperty("name").GetString();
 ```
 
+JSON properties can also be accessed using a dynamic layer.
```
annelo-msft, line 121: I used a new fake WidgetClient instead of the test PetStoreClient because the other is used in a lot of tests and I wanted something specifically for samples. If we don't like this approach, I can look at either moving this article to the WidgetClient completely, or better understanding the needs of the PetStoreClient.  Both are generated from fake swagger files.

sdk/core/Azure.Core/samples/DynamicContent.md
```diff
@@ -0,0 +1,73 @@
...
+RequestContent update = RequestContent.Create(
+    new
+    {
+        Id = (string)widget.Id,
+        Name = "New Name"
+
+        // A forgotten field may be deleted!
+    }
+);
+await client.SetWidgetAsync((string)widget.Id, update);
```
KrzysztofCwalina, line 57: since ID is in the JSON payload, why does this API need the first parameter?

annelo-msft: I thought about this after I wrote it as well.  I believe a REST API definition to PUT a value would need to take a unique identifier as a parameter (although if you know of a counter-example, I'm happy to change this).  Since these are protocol methods, such a parameter would have to appear in the method signature.

It's the kind of thing we would add a convenience overload for in a convenience client (and do, for example, in the AppConfig [ConfigurationClient](https://apiview.dev/Assemblies/Review/8b7f5312697a458ab9e65c2fd9cdc2dd#Azure.Data.AppConfiguration.ConfigurationClient) for SetConfigurationSetting().

sdk/core/Azure.Core/samples/DynamicContent.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Collections can be enumerated
+
+Dynamic JSON arrays are `IEnumerable` and can be iterated over with the `foreach` keyword.
+
+[TODO: Add sample]
+
+## Setting Request Content
+
+Authoring JSON from scratch to pass as inputs to protocol methods is [done using anonymous types](ProtocolMethods.md#2-create-and-send-a-request).  When working with Azure services, however, it is common to retrieve a value from from the service, make some changes to it, and send the updated value back to the service.  This is called a "round-trip scenario."
```
jsquire, line 40: Does it need to be done via anonymous types, or is that just the recommendation?

sdk/core/Azure.Core/samples/ProtocolMethods.md
jsquire: Should we be using something like `CattleClient` and other bovine-related properties as examples so that we stay on-brand for our current initiatives?

_( yes, this is said in jest.  😄 )_

sdk/core/Azure.Core/samples/Response.md
```diff
@@ -55,6 +55,18 @@ using (StreamReader reader = new StreamReader(contentStream))
 }
 ```
 
+## Accessing HTTP response content with `dynamic`
```
jsquire, line 58: Since this is likely to be the more common case as DPG takes over the world, should we consider reordering so that it is the first section on accessing HTTP content?

sdk/core/Azure.Core/samples/ProtocolMethods.md
annelo-msft: 🐮

sdk/core/Azure.Core/samples/DynamicContent.md
```diff
@@ -0,0 +1,73 @@
+# Azure SDK Dynamic JSON samples
+
+Azure SDK client [Protocol methods](ProtocolMethods.md) do not take or return model types.  JSON response content can be accessed using BCL JSON types, but use of these APIs can result in code that is difficult to read and obscures the author's intent.  To improve the developer experience, Azure.Core provides a dynamic layer over JSON APIs.
```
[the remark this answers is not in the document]
annelo-msft: Yes, I like what you're doing there.  Let me take another iteration over that given your suggestions.

Also, good call out on links to .NET dynamic docs!  In the past, I've gotten feedback that I link out too much so I was trying to work against that instinct here, but I think this is a good place to do so for a basic and fundamental concept that may be unfamiliar.  Thanks!

sdk/core/Azure.Core/samples/DynamicContent.md
annelo-msft: It is the officially recommended guidance, but it's possible to author it any way that works with [RequestContent.Create()](https://learn.microsoft.com/en-us/dotnet/api/azure.core.requestcontent.create?view=azure-dotnet).  I can reword this to make that clear - thanks!

sdk/core/Azure.Core/README.md
```diff
@@ -1,20 +1,20 @@
 # Azure Core shared client library for .NET
 
-Azure.Core provides shared primitives, abstractions, and helpers for modern .NET Azure SDK client libraries. 
-These libraries follow the [Azure SDK Design Guidelines for .NET](https://azure.github.io/azure-sdk/dotnet_introduction.html) 
-and can be easily identified by package and namespaces names starting with 'Azure', e.g. ```Azure.Storage.Blobs```. 
-A more complete list of client libraries using Azure.Core can be found [here](https://github.com/Azure/azure-sdk-for-net#core-services). 
+Azure.Core provides shared primitives, abstractions, and helpers for modern .NET Azure SDK client libraries.
+These libraries follow the [Azure SDK Design Guidelines for .NET](https://azure.github.io/azure-sdk/dotnet_introduction.html)
+and can be easily identified by package and namespaces names starting with 'Azure', e.g. ```Azure.Storage.Blobs```.
```
heaths, line 5: Nit: This works with commonmark but is non-standard for inline code. While you're already changing it,
```suggestion
and can be easily identified by package and namespaces names starting with 'Azure', e.g. `Azure.Storage.Blobs`.
```

heaths: There's a few other places below, but up to you, of course.

sdk/core/Azure.Core/samples/DynamicContent.md
heaths, line 3: Not really a proper noun, is it? Otherwise, I'd think it'd be "Protocol Methods".
```suggestion
Azure SDK client [protocol methods](ProtocolMethods.md) do not take or return model types.  JSON response content can be accessed using BCL JSON types, but use of these APIs can result in code that is difficult to read and obscures the author's intent.  To improve the developer experience, Azure.Core provides a dynamic layer over JSON APIs.
```

heaths, line 3: We should spell out acronyms the first time we use them:
```suggestion
Azure SDK client [Protocol methods](ProtocolMethods.md) do not take or return model types.  JSON response content can be accessed using Base Class Library (BCL) JSON types, but use of these APIs can result in code that is difficult to read and obscures the author's intent.  To improve the developer experience, Azure.Core provides a dynamic layer over JSON APIs.
```

JSON is probably common enough not to do that, but "BCL" may not be.

sdk/core/Azure.Core/samples/DynamicContent.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Accessing Response Content
+
+### Get `dynamic` from response
+
+Dynamic content is obtained from the `Response` return value.
+
+```C# Snippet:GetDynamicJson
```
heaths, line 11: These have to be unique throughout the repo. Maybe add "AzureCore" to namespace these e.g., "AzureCoreGetDynamicJson"? I could see "GetDynamicJson" being vague enough to yield future collisions and using your package name (or would-be package name) is a common solution.

sdk/core/Azure.Core/samples/DynamicContent.md
```diff
@@ -0,0 +1,73 @@
...
+Response response = await client.GetWidgetAsync("123");
+dynamic widget = response.Content.ToDynamic();
+```
+
+### Get a JSON property
+
+JSON properties are read using dynamic member access.
```
heaths, line 18: Should we also document what the default binding behavior is e.g., originally-cased properties (I believe)?

sdk/core/Azure.Core/samples/DynamicContent.md
```diff
@@ -0,0 +1,73 @@
...
+
+```C# Snippet:GetDynamicJsonProperty
+Response response = await client.GetWidgetAsync("123");
+dynamic widget = response.Content.ToDynamic();
+string name = widget.Name;
```
heaths, line 23: At least in my work for CLU, the default seems to be originally-cased (should be camel, based on API guidelines) property accessors.

sdk/core/Azure.Core/samples/DynamicContent.md
```diff
@@ -0,0 +1,73 @@
...
+
+```C# Snippet:RoundTripAnonymousType
+Response response = client.GetWidget("123");
+dynamic widget = response.Content.ToDynamic();
+
+RequestContent update = RequestContent.Create(
+    new
+    {
+        Id = (string)widget.Id,
+        Name = "New Name"
```
heaths, line 52: I thought also that the default setter case is case-sensitive? Almost all REST APIs should be using camelCase per guidelines.

heaths: On a related note, we might want to show a sample using an array. As @KrzysztofCwalina was suggesting offline in another email thread, people should use `new object[]` instead of just `new[]` to avoid a compiler error if they elide any defaulted properties, e.g.:

```C#
var data = new
{
  myarray = new object[]
  {
    new
    {
      a = "foo"
    },
    new
    {
      a = "bar",
      b = true
    }
  }
};
```

sdk/core/Azure.Core/samples/ProtocolMethods.md
heaths: Now those jokes will be lost to the ether. 😥

sdk/core/Azure.Core/samples/Response.md
heaths: Is that necessarily true given we want to move everyone to TypeSpec and, ergo, generate models? Perhaps, for now, this will become the primary case but that hopefully one day changes. I keep hearing different things from different people, though.

KrzysztofCwalina:
I saw several references to DynamiJson being in Azure.Core. Is that what we decided? I though we were leaning toward a separate package, e.g. Azure.Core.Json (or something like that)

heaths:
> I saw several references to DynamiJson being in Azure.Core. Is that what we decided? I though we were leaning toward a separate package, e.g. Azure.Core.Json (or something like that)

@KrzysztofCwalina wouldn't that degrade the experience since a new package will need to be added? What I liked when migrating my samples - or even writing some new tests - was that, because I already had a `<ProjectReference>` added - I could just type `ToDynamic` and hit {Tab} and everything just worked e.g., a `using` to `Azure.Core.Dynamic` (or whatever - I forget) was added automatically.

annelo-msft:
> I saw several references to DynamiJson being in Azure.Core. Is that what we decided? I though we were leaning toward a separate package, e.g. Azure.Core.Json (or something like that)

I like that @heaths likes it in Azure.Core directly.

I've been thinking more about this, and two things have come up that make me think maybe the type is better in Azure.Core.
1. If we wanted to add a static RequestContent.Create(DynamicJson), I believe we can only do this if DynamicJson is in Azure.Core, since we aren't able to add a static extension method to RequestContent.
2. If we wanted to build a JSON Merge Patch story around DynamicJson/MutableJsonDocument, that would mean any generated library that wanted to use this convenience would need to include the separate library.

Neither of these is a deal breaker, but both feel like they reduce the convenience value of DynamicJson.

KrzysztofCwalina:
I see the problem with RequestContent.Create, though is it a common scenario to create requests from DJ?

As to JSON Path, I imagine we would use MutableDocument, not DJ if we were shipping smart/patchable models. In thins case, we could ship MutableJson on Azure.Core and then DJ in the other package.

annelo-msft:
> I see the problem with RequestContent.Create, though is it a common scenario to create requests from DJ?

It would be needed for any round-trip scenario.

sdk/core/Azure.Core/samples/DynamicContent.md
annelo-msft, line 3: Note: We should add a call out regarding best practices on testing per @heaths's discovery: https://github.com/Azure/azure-sdk-for-net/pull/35320#issuecomment-1495079875

heaths: Shouldn't we just require recorded tests already, though? Releasing an SDK without tests seems like a really bad idea. The reason I was impacted was because our Azure.Core.TestFramework doesn't check request bodies. In my case, I was switching from models to DynamicJson / DynamicData so I intentionally didn't re-record. If the records didn't match, then I did something wrong.

Using anonymous types for input was merely an implementation detail in that case.

If dev teams start with tests using dynamic, when recording them they would've found these issues. The only reason they'd miss them is if they didn't have any tests at all.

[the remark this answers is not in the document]
KrzysztofCwalina: Let's just make sure that the dynamic docs don't talk too much about how dynamic internals work and how IDynamicMetaObject is implemented. We don't need people to understand it.

sdk/core/Azure.Core/samples/DynamicContent.md
KrzysztofCwalina: It would be good to explain to users how they can know what properties they can access, i.e. REST API docs and debugger.

sdk/core/Azure.Core/samples/DynamicContent.md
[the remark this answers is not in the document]
heaths: Might be worth mentioning at least that we implement it, so that devs know what they can/can't do. Or perhaps that's implied enough that they can use `dynamic`?

sdk/core/Azure.Core/samples/DynamicContent.md
annelo-msft: > At least in my work for CLU, the default seems to be originally-cased (should be camel, based on API guidelines) property accessors.

Updated to show intended usage via defaults (and will update again once https://github.com/Azure/azure-sdk-for-net/pull/35777 lands)

sdk/core/Azure.Core/samples/DynamicContent.md
annelo-msft: Agreed, this addition is out of scope for these docs.

sdk/core/Azure.Core/samples/Response.md
annelo-msft: I'll move it to before the GetRawResponse().ContentStream sample and after the Response\<T\> sample.

sdk/core/Azure.Core/samples/DynamicContent.md
annelo-msft: Added a section on REST API docs and the VS Dynamic View.

sdk/core/Azure.Core/samples/DynamicContent.md
```diff
@@ -0,0 +1,152 @@
...
+
+## Accessing Response Content
+
+### Get `dynamic` from response
+
+Dynamic content is obtained from the `Response` return value.
+
+```C# Snippet:AzureCoreGetDynamicJson
+Response response = await client.GetWidgetAsync("123");
+dynamic widget = response.Content.ToDynamicFromJson(DynamicDataOptions.Default);
```
KrzysztofCwalina, line 13: I think it would be better to not pass the options here. We are still trying to explain the basics here and this parameter makes the sample more complex than it needs to be.

sdk/core/Azure.Core/samples/DynamicContent.md
```diff
@@ -0,0 +1,152 @@
+# Azure SDK Dynamic JSON samples
+
+Azure SDK client [protocol methods](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/samples/ProtocolMethods.md) take `RequestContent` as an input parameter and return `Response` as their return type.  These types hold raw JSON content that can be accessed using Base Class Library (BCL) types such as `JsonDocument`, but use of these APIs can result in code that is difficult to read and obscures the author's intent.  To improve the developer experience, Azure.Core provides a [dynamic](https://learn.microsoft.com/dotnet/csharp/advanced-topics/interop/using-type-dynamic) layer over JSON APIs.
```
KrzysztofCwalina, line 3: Instead of saying "Base Class Library (BCL) types" I would say "any JSON parser, e.g. JsonDocument"

KrzysztofCwalina: Also, it's a bit inconsistent that you are taking about RequestContent. It's an input JsonDocument does not help with that. 

Lastly, the APIs are more general than protocol methods, and so maybe this section should start with "Some Azure SDK APIs return values that hold raw JSON, e.g. protocol methods.

sdk/core/Azure.Core/samples/DynamicContent.md
```diff
@@ -0,0 +1,152 @@
...
+
+```C# Snippet:AzureCoreGetDynamicJsonProperty
+Response response = await client.GetWidgetAsync("123");
+dynamic widget = response.Content.ToDynamicFromJson(DynamicDataOptions.Default);
+string name = widget.Name;
+```
+
+You can learn what properties are available in the JSON response content from the REST API documentation for the service, examples in the protocol method documentation, or by expanding the [Dynamic View](https://learn.microsoft.com/visualstudio/debugger/watch-and-quickwatch-windows) in Visual Studio.
+
+If no parameter is passed to `ToDynamicFromJson()`, properties names must exactly match the member names in the JSON content.  Passing `DynamicDataOptions.Default` will enable properties to be accessed using PascalCase property names, and will write any added properties with camelCase names.
```
KrzysztofCwalina, line 28: I would move this whole section (and so discussion of the mapping feature) to the end, i.e. advanced concepts.

sdk/core/Azure.Core/samples/DynamicContent.md
```diff
@@ -0,0 +1,152 @@
...
+
+### Set a JSON property
+
+JSON members can be set on the dynamic object.
```
KrzysztofCwalina, line 32: I think this needs to be illustrated with a round-trip scenario, otherwise the read might wonder why they would ever do it.

sdk/core/Azure.Core/samples/DynamicContent.md
```diff
@@ -0,0 +1,152 @@
...
+### Check whether an optional property is present
+
+Optional properties will return null if not present in the JSON content.
+
+```C# Snippet:AzureCoreGetDynamicJsonOptionalProperty
+Response response = await client.GetWidgetAsync("123");
+dynamic widget = response.Content.ToDynamicFromJson(DynamicDataOptions.Default);
+
+// Check whether optional property is present
+if (widget.Properties != null)
```
KrzysztofCwalina, line 61: "property" is overloaded here. It's both a concept you are taking about here and a specific property. I would avoid this, i.e. rename "Properties" property to "Items" for example.

sdk/core/Azure.Core/samples/DynamicContent.md
```diff
@@ -0,0 +1,152 @@
...
+// JSON is "{ values = [1, 2, 3] }"
+int value = widget.Values[0];
+```
+
+### Check whether an optional property is present
+
+Optional properties will return null if not present in the JSON content.
+
+```C# Snippet:AzureCoreGetDynamicJsonOptionalProperty
+Response response = await client.GetWidgetAsync("123");
```
KrzysztofCwalina, line 57: It would be good if it was clear what the payload looks like here, e.g. 
```csharp
Response response = await client.GetWidgetAsync(); // returns { "properties" : { "color" : "blue" }  }
```

Also, I find the "123" distracting. I would remove it. Maybe even "Async". All these things are not relevant to this guide.

sdk/core/Azure.Core/samples/DynamicContent.md
```diff
@@ -0,0 +1,152 @@
...
+    string color = widget.Properties.Color;
+}
+```
+
+### Enumerate a collection
+
+Dynamic JSON objects and arrays are `IEnumerable` and can be iterated over with the `foreach` keyword.
+
+```C# Snippet:AzureCoreEnumerateDynamicJsonObject
+Response response = await client.GetWidgetAsync("123");
```
KrzysztofCwalina, line 72: t would be good if it was clear what the payload looks like here

sdk/core/Azure.Core/samples/DynamicContent.md
```diff
@@ -0,0 +1,152 @@
...
+dynamic widget = response.Content.ToDynamicFromJson(DynamicDataOptions.Default);
+
+foreach (dynamic property in widget.Properties)
+{
+    UpdateWidget(property.Name, property.Value);
+}
+
+void UpdateWidget(string name, string value)
```
KrzysztofCwalina, line 80: This does not update. Why is it called Update?

sdk/core/Azure.Core/samples/DynamicContent.md
```diff
@@ -0,0 +1,152 @@
...
+    Console.WriteLine($"Widget has property {name}='{value}'.");
+}
+```
+
+### Cast to a POCO type
+
+Dynamic JSON objects can be cast to CLR types using the cast operator.
+
+```C# Snippet:AzureCoreCastDynamicJsonToPOCO
+Response response = await client.GetWidgetAsync("123");
```
KrzysztofCwalina, line 91: t would be good if it was clear what the payload looks like here

sdk/core/Azure.Core/samples/DynamicContent.md
```diff
@@ -0,0 +1,152 @@
...
+    public string Name { get; set; }
+}
+```
+
+### Get a property with invalid C# characters in the name
+
+JSON members whose names have characters that are not valid for property names in C# can be accessed using property indexers.
+
+```C# Snippet:AzureCoreGetDynamicPropertyInvalidCharacters
+Response response = await client.GetWidgetAsync("123");
```
KrzysztofCwalina, line 108: It would be good if the JSON payload here actually had the required properties.

sdk/core/Azure.Core/samples/DynamicContent.md
annelo-msft: Love it, nice catch.

sdk/core/Azure.Core/samples/DynamicContent.md
annelo-msft: I believe I have addressed this, but please tell me if you had something different in mind.

KrzysztofCwalina (approved):

christothes (approved):

PR #34002, opened 2023-02-08
Fix binlog artifact upload

weshaggard:
@heaths I was trying to set System.Debug=true on a pipeline and keep hitting failures around the binlog stuff. I don't think this has ever worked I've fixed this so that the logs get generated and uploaded. I also noticed that the binlog only gets uploaded for the tests today if you want the build and packaging we still need to add a similar upload step in ci.yml.

### DIFF
--- a/eng/pipelines/templates/jobs/ci.tests.yml
+++ b/eng/pipelines/templates/jobs/ci.tests.yml
@@ -117,9 +117,9 @@ jobs:
           mergeTestResults: true
       - template: /eng/common/pipelines/templates/steps/publish-artifact.yml
         parameters:
-          ArtifactPath: $(Build.ArtifactStagingDirectory)/*.binlog
-          ArtifactName: logs
-          CustomCondition: eq(variables['System.Debug'], 'true')
+          ArtifactPath: $(Build.ArtifactStagingDirectory)/test.binlog
+          ArtifactName: binlog_$(Agent.JobName)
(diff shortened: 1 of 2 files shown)


eng/pipelines/templates/jobs/ci.tests.yml
```diff
@@ -117,9 +117,9 @@ jobs:
           mergeTestResults: true
       - template: /eng/common/pipelines/templates/steps/publish-artifact.yml
         parameters:
-          ArtifactPath: $(Build.ArtifactStagingDirectory)/*.binlog
-          ArtifactName: logs
-          CustomCondition: eq(variables['System.Debug'], 'true')
+          ArtifactPath: $(Build.ArtifactStagingDirectory)/test.binlog
+          ArtifactName: binlog_$(Agent.JobName)
```
heaths, line 121: Thanks. I swear I tested this, but maybe the conditional issue you fixed in the other file (`System.Debug` was documented to be available, but I didn't realize setting it queue time doesn't just update that globally scoped variable) and don't remember any conflicts. Besides, I incorrectly remembered AzDO can't just append to artifacts like GitHub Actions can...and I've been writing a heck of a lot more GitHub Actions the past few years.

heaths (approved):

weshaggard:
The mac failure is unrelated to my changes so overriding that check.

dw511214992:
aaa

dw511214992:
aaa

PR #34006, opened 2023-02-08
Core amqp converter

### LINKED ISSUE #33935: [Core.Amqp] Move ToBytes/FromBytes methods to Azure.Core.Amqp
Instead of defining these in Service Bus and Event Hubs, we should move these to Azure.Core.Amqp. This will mean we will need to add a dependency to Microsoft.Azure.Amqp. We should also try to factor out the AmqpAnnotatedMessage <-> AmqpMessage conversion to shared source in Azure.Core.Amqp.

JoshLove-msft:
Fixes https://github.com/Azure/azure-sdk-for-net/issues/33935


sdk/core/Azure.Core.Amqp/src/Shared/AmqpAnnotatedMessageConverter.cs
```diff
@@ -0,0 +1,956 @@
...
+        /// </remarks>
+        ///
+        public static AmqpMessage ToAmqpMessage(AmqpAnnotatedMessage sourceMessage)
+        {
+            var message = sourceMessage switch
+            {
+                _ when sourceMessage.Body.TryGetData(out var dataBody) => AmqpMessage.Create(TranslateDataBody(dataBody!)),
+                _ when sourceMessage.Body.TryGetSequence(out var sequenceBody) => AmqpMessage.Create(TranslateSequenceBody(sequenceBody!)),
+                _ when sourceMessage.Body.TryGetValue(out var valueBody) => AmqpMessage.Create(TranslateValueBody(valueBody!)),
+                _ => AmqpMessage.Create(new Data { Value = new ArraySegment<byte>(Array.Empty<byte>()) })
```
christothes, line 72: Is only one of these guaranteed to be true, or does it not matter which one is selected if so?

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Amqp/AmqpConverterTests.cs
```diff
@@ -127,20 +129,34 @@ public void ReceivedMessageShouldIncrementDeliveryCount()
             Assert.AreEqual(3, sbMessage.DeliveryCount);
         }
 
+        // [Test]
+        // public void CanParseDictionaryValueSection()
```
JoshLove-msft, line 133: This test started failing when adopting the change from EH to throw in TryGetNetPropertyFromAmqp for the default case statement. I don't think it is really an issue because we are constructing the AmqpMessage by hand and it is not representative of what it would come back as from Service Bus.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/ServiceBusReceivedMessage.cs
```diff
@@ -428,10 +445,6 @@ public ServiceBusMessageState State
 
                 return ServiceBusMessageState.Active;
             }
-            internal set
```
JoshLove-msft, line 431: This was not used.

sdk/core/Azure.Core.Amqp/api/Azure.Core.Amqp.netstandard2.0.cs
```diff
@@ -19,14 +19,16 @@ namespace Azure.Core.Amqp
     public partial class AmqpAnnotatedMessage
     {
         public AmqpAnnotatedMessage(Azure.Core.Amqp.AmqpMessageBody body) { }
-        public System.Collections.Generic.IDictionary<string, object> ApplicationProperties { get { throw null; } }
+        public System.Collections.Generic.IDictionary<string, object?> ApplicationProperties { get { throw null; } }
         public Azure.Core.Amqp.AmqpMessageBody Body { get { throw null; } set { } }
-        public System.Collections.Generic.IDictionary<string, object> DeliveryAnnotations { get { throw null; } }
-        public System.Collections.Generic.IDictionary<string, object> Footer { get { throw null; } }
+        public System.Collections.Generic.IDictionary<string, object?> DeliveryAnnotations { get { throw null; } }
```
JoshLove-msft, line 24: The nullability of all the dictionaries was not correct. The object is allowed to be null, per SB service behavior and AMQP spec.

sdk/core/Azure.Core.Amqp/src/AmqpAnnotatedMessage.cs
```diff
@@ -153,5 +155,27 @@ section switch
                 AmqpMessageSection.Footer => (_footer != null),
                 _ => throw new ArgumentException($"Unknown AMQP message section: { section }.", nameof(section))
             };
+
+        /// <summary>
+        /// Converts the AMQP message into its binary serialized form.
```
jsquire, line 160: Should we mention that this is the actual AMQP format described by the spec and not something proprietary?

jsquire, line 160:
```suggestion
        /// Serializes the message into its AMQP binary form.
```

sdk/core/Azure.Core.Amqp/src/AmqpAnnotatedMessage.cs
```diff
@@ -153,5 +155,27 @@ section switch
...
+        /// </summary>
+        /// <returns>A <see cref="BinaryData"/> instance representing the serialized message.</returns>
+        public virtual BinaryData ToBytes()
+        {
+            var stream = AmqpAnnotatedMessageConverter.ToAmqpMessage(this).ToStream();
+            return BinaryData.FromStream(stream);
+        }
+
+        /// <summary>
+        /// Constructs a message from its binary serialized form.
```
jsquire, line 170:
```suggestion
        /// Constructs a message from the AMQP binary format.
```

sdk/core/Azure.Core.Amqp/src/Shared/AmqpAnnotatedMessageConverter.cs
```diff
@@ -0,0 +1,956 @@
...
+                        //     };
+                        // }
+                        // else if (SystemPropertyLongKeys.Contains(pair.Key.ToString()))
+                        // {
+                        //     eventValue = eventValue switch
+                        //     {
+                        //         string stringValue when long.TryParse(stringValue, NumberStyles.Integer, CultureInfo.InvariantCulture, out var longValue) => longValue,
+                        //         _ => eventValue
+                        //     };
+                        // }
```
christothes, line 418: Is this intentional?

sdk/core/Azure.Core.Amqp/src/Shared/AmqpAnnotatedMessageConverter.cs
```diff
@@ -0,0 +1,956 @@
...
```
christothes, line 449: intentional?

sdk/core/Azure.Core.Amqp/src/Shared/AmqpAnnotatedMessageConverter.cs
```diff
@@ -0,0 +1,956 @@
...
+
+            // Message Annotations
+
+            if ((source.Sections & SectionFlag.MessageAnnotations) > 0)
+            {
+                foreach (var pair in source.MessageAnnotations.Map)
+                {
+                    if (TryCreateNetPropertyFromAmqpProperty(pair.Value, out var propertyValue))
+                    {
+                        // if (SystemPropertyDateTimeKeys.Contains(pair.Key.ToString()))
```
jsquire: Is this intended to be commented out?   Should we consider removing it?

sdk/core/Azure.Core.Amqp/src/Shared/AmqpAnnotatedMessageConverter.cs
```diff
@@ -0,0 +1,956 @@
...
+
+            // Delivery Annotations
+
+            if ((source.Sections & SectionFlag.DeliveryAnnotations) > 0)
+            {
+                foreach (var pair in source.DeliveryAnnotations.Map)
+                {
+                    if (TryCreateNetPropertyFromAmqpProperty(pair.Value, out var eventValue))
+                    {
+                        // if (SystemPropertyDateTimeKeys.Contains(pair.Key.ToString()))
```
jsquire: This is also left over from the normalization changes, no?

sdk/core/Azure.Core.Amqp/tests/AmqpAnnotatedMessageConverterTests.cs
```diff
@@ -0,0 +1,177 @@
...
+            new Dictionary<string, string> { { "key", "value" } },
+            new Dictionary<string, char> {{ "key", 'c' }},
+            new Dictionary<string, int> {{ "key", 5 }},
+            new Dictionary<string, byte> {{ "key", 1 } },
+            new Dictionary<string, sbyte> {{ "key", 1 } },
+            new Dictionary<string, short> {{ "key", 1 } },
+            new Dictionary<string, double> {{ "key", 3.1415926 } },
+            new Dictionary<string, decimal> {{ "key", new decimal(3.1415926) } },
+            new Dictionary<string, DateTime> {{ "key", DateTimeOffset.Parse("3/24/21").UtcDateTime } },
+            // for some reason dictionaries with DateTimeOffset, Timespan, or Uri values are not supported in AMQP lib
```
jsquire, line 50: They need to be set as described types, I believe.  You can't use them as just a plain value type.

https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Azure.Messaging.EventHubs/tests/Amqp/AmqpMessageConverterTests.cs#L49

sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/ServiceBusReceivedMessage.cs
```diff
@@ -42,6 +43,22 @@ internal ServiceBusReceivedMessage(AmqpAnnotatedMessage message)
         internal ServiceBusReceivedMessage(): this(body: default)
         { }
 
+        /// <summary>
+        /// Constructs a <see cref="ServiceBusReceivedMessage"/> from its serialized AMQP form.
+        /// </summary>
+        /// <param name="message">The AMQP message.</param>
+        /// <param name="lockTokenBytes">The lock token bytes.</param>
+        /// <returns></returns>
```
jsquire, line 51: nit:  We should either populate the return tag or remove it.

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Amqp/AmqpConverterTests.cs
```diff
@@ -127,20 +129,34 @@ public void ReceivedMessageShouldIncrementDeliveryCount()
...
+        //     Assert.IsTrue(body.TryGetValue(out object val));
+        //     Assert.AreEqual("value", ((Dictionary<string, string>)val)["key"]);
+        //
+        //     amqpMessage = AmqpMessage.Create(new AmqpValue { Value = new AmqpMap { { new MapKey("key"), "value" } } });
+        //     sbMessage = converter.AmqpMessageToSBReceivedMessage(amqpMessage);
+        //     body = sbMessage.GetRawAmqpMessage().Body;
+        //     Assert.IsTrue(body.TryGetValue(out val));
+        //     Assert.AreEqual("value", ((Dictionary<string, object>)val)["key"]);
+        // }
+
```
christothes, line 148: intentional?

sdk/core/Azure.Core.Amqp/src/Shared/AmqpAnnotatedMessageConverter.cs
JoshLove-msft: Yes, only 1 will be true.

sdk/servicebus/Azure.Messaging.ServiceBus/tests/Amqp/AmqpConverterTests.cs
jsquire: This is the "throw on unknown" part of the conversion that we were talking about, right?

JoshLove-msft: Yes, this test is failing because of that change.

sdk/core/Azure.Core.Amqp/tests/AmqpAnnotatedMessageConverterTests.cs
JoshLove-msft: Interesting.. Should we be handling this in the AmqpAnnoatedMessageConverter?

sdk/core/Azure.Core.Amqp/src/Shared/AmqpAnnotatedMessageConverter.cs
JoshLove-msft: These were removed already.

sdk/core/Azure.Core.Amqp/tests/AmqpAnnotatedMessageConverterTests.cs
JoshLove-msft: This would only impact value-body sections which are not widely used.

jsquire: > Should we be handling this in the AmqpAnnotatedMessageConverter? 

Maybe?  We convert for direct properties in maps.   As you noted, we do not crack open nested values nor do we do the Value body.    I don't think we want to start poking at nested values, but the value body may be nice.   

That said, it's such a specialized niche case that it's probably not worth the investment.

JoshLove-msft: Agreed.

jsquire (approved):

PR #34007, opened 2023-02-08
Fix tests that rely on reflection

christothes:
We recently changed an internal field name in Azure.Identity that broke these tests. This fix allows either the new or old property name so that project and package refs work until the change is published to NuGet.

### DIFF
--- a/sdk/extensions/Microsoft.Extensions.Azure/tests/ClientFactoryTests.cs
+++ b/sdk/extensions/Microsoft.Extensions.Azure/tests/ClientFactoryTests.cs
@@ -4,8 +4,10 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Numerics;
 using System.Reflection;
 using System.Security.Cryptography.X509Certificates;
+using Azure.Core.TestFramework;
 using Azure.Identity;
 using Microsoft.Extensions.Azure;
 using Microsoft.Extensions.Configuration;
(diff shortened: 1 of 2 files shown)


heaths (approved):

jsquire (commented):
It seems specific to this property, but just to confirm - this is intended to be specific to reflection against Identity?   I know that we use a similar reflection-based approach for other libraries and wanted to be sure there wasn't some issue post net7 changeover.

sdk/extensions/Microsoft.Extensions.Azure/tests/ClientFactoryTests.cs
```diff
@@ -204,7 +204,7 @@ public void CreatesCertificateCredentialsAdditionalTenants(string additionalTena
             Assert.AreEqual("ConfigurationTenantId", clientCertificateCredential.TenantId);
 
             var actualTenants = (string[]) typeof(ClientCertificateCredential)
-                .GetField("_additionallyAllowedTenantIds", BindingFlags.NonPublic | BindingFlags.Instance)
+                .GetFields(BindingFlags.NonPublic | BindingFlags.Instance).First(f => f.Name.EndsWith("dditionallyAllowedTenantIds"))
```
JoshLove-msft, line 207: Why not use the full word 😆

JoshLove-msft, line 207:
```suggestion
                .GetFields(BindingFlags.NonPublic | BindingFlags.Instance).First(f => f.Name.EndsWith("additionallyAllowedTenantIds"))
```

JoshLove-msft: The casing?

jsquire (approved):

sdk/extensions/Microsoft.Extensions.Azure/tests/ClientFactoryTests.cs
jsquire: ust ecause!

JoshLove-msft: 😆 
After I left the comment I realized this may be intentional in case we switch between casing.

christothes: > 😆 After I left the comment I realized this may be intentional in case we switch between casing.

Yeah - the name will be either `_additional...` or `Additional...`

heaths: I had the same thought initially, but realized this would solve the problem in the short window where a PackageReference or ProjectReference could make a difference.

JoshLove-msft (approved):

Issue #33976, opened 2023-02-08
Storage pipelines failing on Build Analyze step due to code snippets


seanmcc-msft:
Code snippets were recently enabled for the storage directory of the repo.  This has caused the storage pipelines to fail on the Build Analyze steps, like in this PR - https://github.com/Azure/azure-sdk-for-net/pull/33921

Azure Pipelines run example - https://github.com/Azure/azure-sdk-for-net/runs/11179015923

[Check failure on line 37 in sdk/storage/Azure.ResourceManager.Storage/tests/Samples/ChangeLog.cs](https://github.com/Azure/azure-sdk-for-net/commit/c908737527b16d8c3a5762f1733a990425225c15#annotation_8664390297) 

@azure-pipelinesazure-pipelines
/ net - storage - ci (Build Analyze)
sdk/storage/Azure.ResourceManager.Storage/tests/Samples/ChangeLog.cs#L37
sdk/storage/Azure.ResourceManager.Storage/tests/Samples/ChangeLog.cs(37,9): Error CS1022: Type or namespace definition, or end-of-file expected

heaths:
Duplicate of #33967 since both the _Changelog.cs_ and _Readme.cs_ files are affected, but you could fix it separately.

Issue #33967, opened 2023-02-08
Some ResourceManagement samples do not compile with `SNIPPET` symbol


heaths:
We're working on turning on building snippets for all projects by default to make sure all our source from which snippets are created actually build. This was one major reason we started this project a few years ago. When doing so, however, this means some code is invalid when compiled. Consider the following sample, which I see in a few mgmt projects (networking, storage, etc.):

```c#
// Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT License.

#region Snippet:Changelog_New
            using Azure.Identity;
            using Azure.ResourceManager.Compute.Models;
            using Azure.ResourceManager.Network;
            using Azure.ResourceManager.Network.Models;
            using Azure.ResourceManager.Resources;
            using Azure.ResourceManager.Resources.Models;
            using Azure.Core;
            using System;
            using System.Linq;

#if !SNIPPET
using System.Threading.Tasks;
using NUnit.Framework;

namespace Azure.ResourceManager.Compute.Tests.Samples
{
    public class Changelog
    {
        [Test]
        [Ignore("Only verifying that the sample builds")]
        public async Task NewCode()
        {
#endif
            ArmClient armClient = new ArmClient(new DefaultAzureCredential());

            AzureLocation location = AzureLocation.WestUS;
            // Create ResourceGroupResource
            SubscriptionResource subscription = await armClient.GetDefaultSubscriptionAsync();
            ArmOperation<ResourceGroupResource> rgOperation = await subscription.GetResourceGroups().CreateOrUpdateAsync(WaitUntil.Completed, "myResourceGroup", new ResourceGroupData(location));
            ResourceGroupResource resourceGroup = rgOperation.Value;

```

When we pass `/p:DefineConstants=SNIPPET`, part of that source is excluded and now invalid, so the build fails. What we've done in other projects is have separate regions for `using`s and for code. Even in the sample markdown files, we typically call this out as separate sections like you can see [here](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/cognitivelanguage/Azure.AI.Language.Conversations/README.md#namespaces).

I recommend following a similar approach so code is always compilable regardless of symbols passed.

To test this locally, you can run the following replacing `/p:ServiceDirectory=*` for the actual service directory e.g., "network":

```bash
[...]

PR #34009, opened 2023-02-08
Enable data plane BuildSnippets for Storage

heaths:
Relates to #33976, which is only needed for control plane snippets.


heaths:
@seanmcc-msft nevermind, I see now. While _ci.yml_ ignores MPG triggers and _ci.mgmt.yml_ only cares about MPG triggers, the pipeline is still going to run `dotnet build /p:ServiceDirectory=storage /p:BuildSnippets=true ...` for the whole directly regardless when either is triggered; thus, we need to disable the trigger for now. I believe this will be cleaned up with #33967, though.

/cc @m-nash @ArthurMa1978

PR #34021, opened 2023-02-08
[Azure.Monitor.Ingestion] Add tests and update Readme

nisha-bhatia:


sdk/monitor/Azure.Monitor.Ingestion/tests/Samples/LogDataAndQueryAsync.cs
```diff
@@ -95,5 +98,127 @@ public async Task QueryDataAsync()
...
+                    }
+                });
+            }
+
+            // Upload our logs
+            Response response = await client.UploadAsync(ruleId, streamName, entries).ConfigureAwait(false);
+            #endregion
+        }
+
+        public async Task LogDataIEnumerableCancellationTokenAsync()
```
JoshLove-msft, line 183: I thought we weren't going to show this one in a sample. I would move this to a test instead.

sdk/monitor/Azure.Monitor.Ingestion/tests/Samples/LogDataAndQueryAsync.cs
```diff
@@ -95,5 +98,127 @@ public async Task QueryDataAsync()
...
+            UploadLogsOptions options = new UploadLogsOptions();
+            options.MaxConcurrency = 10;
+            options.UploadFailedEventHandler += Options_UploadFailed;
+
+            // Upload our logs
+            Response response = await client.UploadAsync(ruleId, streamName, entries, options).ConfigureAwait(false);
+
+            Task Options_UploadFailed(UploadFailedEventArgs e)
+            {
+                // Throw exception from EventHandler to stop Upload if there is a failure
```
JoshLove-msft, line 141: We should add some logic around the exception, e.g. if it is a 403, then we want to propagate the exception because we expect everything to fail.

sdk/monitor/Azure.Monitor.Ingestion/README.md
```diff
@@ -147,6 +149,77 @@ Response response = await client.UploadAsync(
     RequestContent.Create(data)).ConfigureAwait(false);
 ```
 
+### Upload custom logs as IEnumerable
+
+You can upload logs using either the `LogsIngestionClient.Upload` or the `LogsIngestionClient.UploadAsync` method. In these two methods, logs are passed in a generic IEnumerable type. Additionally, there is an UploadLogsOptions options parameter in which a serializer, concurrency, and EventHandler can be set.
```
scottaddie, line 154:
```suggestion
You can upload logs using either the `LogsIngestionClient.Upload` or the `LogsIngestionClient.UploadAsync` method. In these two methods, logs are passed in a generic `IEnumerable` type. Additionally, there's an `UploadLogsOptions`-typed parameter in which a serializer, concurrency, and EventHandler can be set.
```

sdk/monitor/Azure.Monitor.Ingestion/README.md
```diff
@@ -147,6 +149,77 @@ Response response = await client.UploadAsync(
...
+    });
+}
+
+// Upload our logs
+Response response = await client.UploadAsync(ruleId, streamName, entries).ConfigureAwait(false);
+```
+
+### Upload custom logs as IEnumerable with EventHandler
+
+You can upload logs using either the `LogsIngestionClient.Upload` or the `LogsIngestionClient.UploadAsync` method. In these two methods, logs are passed in a generic IEnumerable type. Additionally, there is an UploadLogsOptions options parameter in which a serializer, concurrency, and EventHandler can be set. The default serializer is set to System.Text.Json, you can pass in the serializer you would like used. The MaxConcurrency sets the number of threads that will be used in UploadAsync method. The default value is 5 and this parameter is unused in the Upload method. The EventHandler is used for error handling. It gives the user the option to abort the Upload if a batch fails and access the failed logs and corresponding exception. Without the EventHandler, if an Upload fails an AggregateException will be thrown.
```
scottaddie, line 184:
```suggestion
You can upload logs using either the `LogsIngestionClient.Upload` or the `LogsIngestionClient.UploadAsync` method. In these two methods, logs are passed in a generic `IEnumerable` type. Additionally, there's an `UploadLogsOptions`-typed parameter in which a serializer, concurrency, and EventHandler can be set. The default serializer is set to `System.Text.Json`, but you can pass in the serializer you would like used. The `MaxConcurrency` property sets the number of threads that will be used in the `UploadAsync` method. The default value is 5, and this parameter is unused in the `Upload` method. The EventHandler is used for error handling. It gives the user the option to abort the upload if a batch fails and access the failed logs and corresponding exception. Without the EventHandler, if an upload fails, an `AggregateException` will be thrown.
```

pallavit:
> Custom logs can send data to any custom table that you create and to certain built-in tables in your Log Analytics workspace. The target table must exist before you can send data to it. The following built-in tables are currently supported:

Do we have tests for both in-built and customer table scenarios?

---
Refers to: sdk/monitor/Azure.Monitor.Ingestion/README.md:64 in d950b7d. [](commit_id = d950b7d8ab3215f348fe80fb7ce3d17826975082, deletion_comment = False)

sdk/monitor/Azure.Monitor.Ingestion/README.md
```diff
@@ -147,6 +149,77 @@ Response response = await client.UploadAsync(
     RequestContent.Create(data)).ConfigureAwait(false);
 ```
 
+### Upload custom logs as IEnumerable
+
+You can upload logs using either the `LogsIngestionClient.Upload` or the `LogsIngestionClient.UploadAsync` method. In these two methods, logs are passed in a generic `IEnumerable` type. Additionally, there's an `UploadLogsOptions`-typed parameter in which a serializer, concurrency, and EventHandler can be set.
```
pallavit, line 154: [nit] > Additionally, there's an `UploadLogsOptions`-typed parameter in which a 

-typed parameter should be typed parameter?

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -232,18 +232,23 @@ public virtual Response Upload<T>(string ruleId, string streamName, IEnumerable<
                         // if there is no Handler on options, throw exception otherwise raise Handler
                         if (!options.HasHandler)
```
pallavit, line 233: When options does not have a handler, we simply want to add exceptions to exception list.  then why throw and catch in the first place?

Am I missing something?

sdk/monitor/Azure.Monitor.Ingestion/tests/ErrorTest.cs
```diff
@@ -12,11 +12,11 @@
 
 namespace Azure.Monitor.Ingestion.Tests
 {
-    [LiveOnly]
+    //[LiveOnly]
```
pallavit: remove this now that it is commented out.

nisha-bhatia: this is only for testing the CancellationToken as the PR is still in draft. It will be uncommented to ensure these large tests only run in [LiveOnly] before merging :)

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
nisha-bhatia: An exception can come from two cases - one is if the status is not a 204 or if the Upload throws an Exception itself. Line 233 is the case where there is a failed status. Throwing a new Request Failed Exception is equivalent to adding it to the list here because it gets caught and added. In my opinion the code makes it clear that this case will throw in the future, but maybe it's not as clear as I'm intending to be.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -259,13 +264,24 @@ public virtual Response Upload<T>(string ruleId, string streamName, IEnumerable<
...
+                            shouldAbort = true;
                             AddException(ref exceptions, exceptionOnUpload);
+                        }
+                    }
+
+                    // Cancel all future Uploads if user triggers CancellationToken
+                    if (ex is OperationCanceledException && cancellationToken.IsCancellationRequested)
+                    {
+                        shouldAbort = true;
+                        AddException(ref exceptions, new OperationCanceledException());
```
JoshLove-msft, line 284: Why not use the existing exception?

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -377,9 +397,15 @@ public virtual async Task<Response> UploadAsync<T>(string ruleId, string streamN
...
                 }
-                catch (Exception)
+                catch (Exception ex)
                 {
                     // We do not want to log exceptions here as we will loop through all the tasks later
+                    // Cancel all future Uploads if user triggers CancellationToken
+                    if (ex is OperationCanceledException && cancellationToken.IsCancellationRequested)
+                    {
+                        shouldAbort = true;
+                        AddException(ref exceptions, new OperationCanceledException());
```
JoshLove-msft, line 407: Use existing exception?

sdk/monitor/Azure.Monitor.Ingestion/tests/ErrorTest.cs
```diff
@@ -183,5 +181,88 @@ Task Options_UploadFailed(UploadFailedEventArgs e)
...
+            bool isTriggered = false;
+            var cts = new CancellationTokenSource();
+            options.UploadFailedEventHandler += Options_UploadFailed;
+            AggregateException exceptions = Assert.ThrowsAsync<AggregateException>(async () => { await client.UploadAsync(TestEnvironment.DCRImmutableId, TestEnvironment.StreamName, entries, options, cts.Token).ConfigureAwait(false); });
+            Assert.IsTrue(isTriggered);
+            Assert.IsTrue(cts.IsCancellationRequested);
+            bool hasBeenCancelled = false;
+            // check if OperationCanceledException is in the Exception list
+            // may not be first one in async case
+            foreach (var exception in exceptions.InnerExceptions)
```
JoshLove-msft, line 218: Could we use a linq expression instead with [Any](https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.any?view=net-7.0)?

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -394,9 +420,20 @@ public virtual async Task<Response> UploadAsync<T>(string ruleId, string streamN
                 else
                 {
                     Exception exceptionEventHandler = await ProcessCompletedTaskEventHandlerAsync(task.CurrentTask, task.Logs, options, cancellationToken).ConfigureAwait(false);
-                    shouldAbort = exceptionEventHandler != null;
-                    if (shouldAbort)
+                    logsFailed += task.Logs.Count;
```
JoshLove-msft, line 423: This doesn't seem right - it looks like we call ProcessCompletedTask even for successful responses, but we are still adding the logs to the failed count.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -394,9 +420,20 @@ public virtual async Task<Response> UploadAsync<T>(string ruleId, string streamN
...
+                    // if exception is thrown stop processing future batches
+                    if (exceptionEventHandler != null)
+                    {
+                        shouldAbort = true;
                         AddException(ref exceptions, exceptionEventHandler);
+                    }
+                    // Cancel all future Uploads if user triggers CancellationToken
+                    // if exception was not thrown on line 400 - token could still be cancelled
+                    if (cancellationToken.IsCancellationRequested)
```
JoshLove-msft, line 432: If no OperationCanceledException is thrown, then it means no downstream work actually got canceled despite the user's attempt to cancel. So there is no need to throw an exception ourselves for this case.

sdk/monitor/Azure.Monitor.Ingestion/tests/Samples/LogDataAndQueryAsync.cs
```diff
@@ -95,5 +98,90 @@ public async Task QueryDataAsync()
...
+            credential = TestEnvironment.Credential;
+#endif
+            LogsIngestionClient client = new(endpoint, credential);
+
+            DateTimeOffset currentTime = DateTimeOffset.UtcNow;
+
+            var entries = new List<Object>();
+            for (int i = 0; i < 100; i++)
+            {
+                entries.Add(new Object[] {
```
JoshLove-msft, line 173: Do we need each entry to be an array?

sdk/monitor/Azure.Monitor.Ingestion/tests/Samples/LogDataAndQueryAsync.cs
```diff
@@ -95,5 +98,90 @@ public async Task QueryDataAsync()
...
+            var streamName = "<stream_name>";
+
+#if SNIPPET
+            var credential = new DefaultAzureCredential();
+#else
+            TokenCredential credential = new DefaultAzureCredential();
+            endpoint = new Uri(TestEnvironment.DCREndpoint);
+            credential = TestEnvironment.Credential;
+#endif
+            LogsIngestionClient client = new(endpoint, credential);
```
JoshLove-msft, line 166: nit: since we are using the `var foo = new bar()` pattern elsewhere in this sample, it would be better to use that here as well.

sdk/monitor/Azure.Monitor.Ingestion/tests/Samples/LogDataAndQueryAsync.cs
nisha-bhatia: I think each entry is a row of the table - that is why the grouping is done this way.

JoshLove-msft: What table do you mean? Isn't the data model completely custom?

nisha-bhatia: I see - I have an extra set of {}

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -457,22 +466,24 @@ private static void ProcessCompletedTask((Task<Response> CurrentTask, List<objec
             }
         }
 
-        internal async Task<Exception> ProcessCompletedTaskEventHandlerAsync(Task<Response> completedTask, List<object> logs, UploadLogsOptions options, CancellationToken cancellationToken)
+        internal async Task<Tuple<Exception, int>> ProcessCompletedTaskEventHandlerAsync(Task<Response> completedTask, List<object> logs, UploadLogsOptions options, CancellationToken cancellationToken)
```
JoshLove-msft, line 469: Use a ValueTuple by just doing `Task<(Exception, int)>` as the return.

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -415,13 +424,13 @@ public virtual async Task<Response> UploadAsync<T>(string ruleId, string streamN
...
-                    logsFailed += task.Logs.Count;
+                    var processTaskResult = await ProcessCompletedTaskEventHandlerAsync(task.CurrentTask, task.Logs, options, cancellationToken).ConfigureAwait(false);
+                    logsFailed += processTaskResult.Item2;
                     // if exception is thrown stop processing future batches
-                    if (exceptionEventHandler != null)
+                    if (processTaskResult.Item1 != null)
                     {
                         shouldAbort = true;
-                        AddException(ref exceptions, exceptionEventHandler);
+                        AddException(ref exceptions, processTaskResult.Item1);
```
JoshLove-msft, line 433: Name the tuple items.

JoshLove-msft (approved):

sdk/monitor/Azure.Monitor.Ingestion/src/LogsIngestionClient.cs
```diff
@@ -466,7 +466,7 @@ private static void ProcessCompletedTask((Task<Response> CurrentTask, List<objec
             }
         }
 
-        internal async Task<Exception> ProcessCompletedTaskEventHandlerAsync(Task<Response> completedTask, List<object> logs, LogsUploadOptions options, CancellationToken cancellationToken)
+        internal async Task<(Exception Ex, int LogsCount)> ProcessCompletedTaskEventHandlerAsync(Task<Response> completedTask, List<object> logs, LogsUploadOptions options, CancellationToken cancellationToken)
```
JoshLove-msft, line 469:
```suggestion
        internal async Task<(Exception Exception, int FailedLogsCount)> ProcessCompletedTaskEventHandlerAsync(Task<Response> completedTask, List<object> logs, LogsUploadOptions options, CancellationToken cancellationToken)
```

PR #34023, opened 2023-02-09
Only build snippets on latest TFM

### LINKED ISSUE #33975: Azure.Core.Tests has uncompilable snippets
When passing the `SNIPPET` symbol to build snippets and make sure they are valid, this project yields errors.

You can repro this issue by running:

```bash
dotnet build eng/service.proj -warnaserror /t:rebuild /p:DebugType=none /p:SDKType=all /p:ServiceDirectory=core /p:IncludePerf=false /p:IncludeStress=false /p:PublicSign=false /p:Configuration=Release /p:EnableSourceLink=false /p:BuildSnippets=true
```

When resolved, please remove the `BuildSnippets: false` line from _sdk/core/ci.yml_.

See #33969 for more context.

heaths:
Also fixes #33975 by reformatting snippets in sdk/core.

### DIFF
--- a/eng/Directory.Build.Common.props
+++ b/eng/Directory.Build.Common.props
@@ -127,6 +127,8 @@
     <RequiredTargetFrameworks>net7.0;net6.0</RequiredTargetFrameworks>
     <!-- Also test net461 on Windows. -->
     <RequiredTargetFrameworks Condition="'$(OS)' == 'Windows_NT'">$(RequiredTargetFrameworks);net461</RequiredTargetFrameworks>
+    <!-- But only build snippets for the latest. -->
+    <RequiredTargetFrameworks Condition="'$(BuildSnippets)' == 'true'">net7.0</RequiredTargetFrameworks>
   </PropertyGroup>
 
   <ItemGroup Condition="'$(IsMgmtSubLibrary)' == 'true' and !$(MSBuildProjectName.Equals('Azure.ResourceManager.Tests'))">
--- a/sdk/core/Azure.Core.TestFramework/tests/Azure.Core.TestFramework.Tests.csproj
+++ b/sdk/core/Azure.Core.TestFramework/tests/Azure.Core.TestFramework.Tests.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>$(RequiredTargetFrameworks);net6.0</TargetFrameworks>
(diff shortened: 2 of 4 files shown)


eng/Directory.Build.Common.props
```diff
@@ -127,6 +127,8 @@
     <RequiredTargetFrameworks>net7.0;net6.0</RequiredTargetFrameworks>
     <!-- Also test net461 on Windows. -->
     <RequiredTargetFrameworks Condition="'$(OS)' == 'Windows_NT'">$(RequiredTargetFrameworks);net461</RequiredTargetFrameworks>
+    <!-- But only build snippets for the latest. -->
+    <RequiredTargetFrameworks Condition="'$(BuildSnippets)' == 'true'">net7.0</RequiredTargetFrameworks>
```
heaths, line 131: This was actually an unexpected change I had to made once I fixed the first build error against `net7.0` - the main point of this commit. But I think this is what we want to do anyway. Snippets may use newer syntax, so why let older TFMs hobble us? It's also faster to run the Analyze step now.

annelo-msft: Interesting!  Do we have a precedent for providing samples for older supported versions of our libraries?  Curious.

jsquire: Would the older TFMs cause issues?    I'd have expected we'd always be compiling with `net7` and executing against the older platforms.    

I don't have a strong opinion, but it may be helpful to limit the new hotness features in snippets so that samples are more widely accessible.   Some of our larger customer segments are slower to adopt new syntax.

christothes (approved):

eng/Directory.Build.Common.props
heaths: This was necessary because Core samples used `ActivitySource` and friends, which isn't defined for net641. And because building snippets uses our existing `<TargetFrameworks>` to build (it's just the test projects) it does multi-targeting. That's what this change prevents.

As to @annelo-msft's question, I'd be surprised if we ever wrote any samples that weren't in newer targets. It's hypothetically possible (.NET "Core" can make breaking changes) but I think it not practically likely. Seems we could cross that bridge when we get to it.

jsquire (approved):

eng/Directory.Build.Common.props
weshaggard, line 131: With this change we definitely need to ensure we always keep BuildSnippets separate from the normal build and package commands. I somewhat fear we might end up dropping target frameworks for what we ship if someone starts passing this property to our normal builds.

Issue #33856, opened 2023-02-02
Add features to DynamicJson


annelo-msft:
A number of features remain to be added to `DynamicJson and related "JsonData" types:

P0 for GA:
- [x] Finalize decisions on what exceptions to throw for different operations, per https://github.com/Azure/azure-sdk-for-net/pull/33063#discussion_r1060848707
- [x] Finalize decisions on where to add `==` operators
  - Initial work done in https://github.com/annelo-msft/azure-sdk-for-net/tree/dynamicjson-optional-properties-backup
- [x] Add validation that property names don't include the `.` character
- [x] Add a custom implementation of RequestContent, and return it from a RequestContent.Create overload.
  - See [RoundTripDynamicJson sample](https://github.com/annelo-msft/azure-sdk-for-net/blob/core-dynamicjson-docs/sdk/core/Azure.Core/tests/samples/DynamicJsonSamples.cs#L63-L71) for details
- [x] Support WriteTo with `StandardFormat`
- [x] Finalize decision on what library to ship types in
- [x] Clean up: Make sure we're `using` all disposable types
- [x] Clean up: Make sure we've removed unnecessary constructor calls
- [x] Clean up/perf improvement: https://github.com/Azure/azure-sdk-for-net/pull/35204/files#r1164766524
- [x] Add test for WriteTo() validation

P1:
- [ ] ObjectDocument abstraction
- [x] Look into https://github.com/Azure/azure-sdk-for-net/pull/35930#discussion_r1183003618 re: perf
- [ ] Rough sketch of support for adding JSON Merge Patch functionality, per https://www.rfc-editor.org/rfc/rfc7386
  - When added, use this recommended style for refdoc formatting: Update doc comment per: https://github.com/Azure/azure-sdk-for-net/pull/35605#discussion_r1171851362
- [x] Debugger proxy
- [ ] Add `[RequiresUnreferencedCode]` attribute for AOT, per https://github.com/Azure/azure-sdk-for-net/pull/35777#discussion_r1175885680
- [ ] Implement indexer on MutableJsonElement
- [x] Implement equality (tests: `EqualsHandlesStringsSpecial`, `EqualsForObjectsAndArrays`)
- [ ] Enable optimizations, per:
      - https://github.com/Azure/azure-sdk-for-net/pull/33063#discussion_r1060841057
      - https://github.com/Azure/azure-sdk-for-net/pull/33063#discussion_r1060843717
[...]

PR #34026, opened 2023-02-09
Don't throw if a property is missing from DynamicJson, to allow for optional property checks

annelo-msft:
In order to write code against optional properties in DynamicJson in the same way we do in our strongly-typed models, we need to be able to check the properties for null to determine whether they are present or absent in the deserialized JSON.

An example of such a check can be seen in the [ConfidentialLedger HelloWorld sample](https://github.com/Azure/azure-sdk-for-net/blob/b378e4b570e015ce95115dda3d5b443256115e96/sdk/confidentialledger/Azure.Security.ConfidentialLedger/tests/samples/HelloWorldSamples.cs#L164-L166).

This PR implements the change to DynamicJson to enable this feature.


sdk/core/Azure.Core.Experimental/src/DynamicJson.cs
```diff
@@ -38,7 +40,12 @@ internal override void WriteTo(Stream stream)
 
         private object GetProperty(string name)
         {
-            return new DynamicJson(_element.GetProperty(name));
+            if (_element.TryGetProperty(name, out MutableJsonElement element))
+            {
+                return new DynamicJson(element);
+            }
+
+            return new DynamicJson(MutableJsonDocument.Parse(nullJsonData).RootElement);
```
jsquire, line 48: We have to parse this as a new document so that callers an update the value?

jsquire (approved):

annelo-msft:
> Do we have a way to ask "does this property exist"? Otherwise, I'm not sure how I'd be able to distinguish between "this property was in the JSON and the value was null for this call" versus "this property was not in the JSON and the value will always be null. I probably made a mistake and should fix it."

Yes, great question.

MutableJsonElement has a TryGetProperty() method (like JsonElement), which allows this.  The challenge with DynamicJson is that we don't want to add much to the public API.  This is because in earlier UX studies, people were confused about when to use the object in it's strongly-typed form vs. as a `dynamic` type.  Minimizing the public API surface is our way of steering away from this problem -- people should only use it as `dynamic`. 

Given this, I've been trying to use the guiding principle that the dynamic type should behave like one of our generated models (we have a fixed set of transforms from the REST API description to a model type built into the generator, so this is predictable).

You're right that the result is we can't tell the difference between the underlying JSON having the property and it being `null`, vs. the property being absent from the JSON.  I believe this is also a current limitation of our model types, so we're maintaining parity there.

We could say that to disambiguate this, you need to convert to MutableJsonDocument or JsonDocument.

sdk/core/Azure.Core.Experimental/tests/DynamicJsonTests.cs
```diff
@@ -278,6 +278,47 @@ public void CanMakeChangesAndAddNewProperty()
...
+            // Add property Baz
+            json.Baz = "baz";
+
+            // Remove property A
+            json.Bar = new { B = "b" };
+
+            // Properties are present
+            Assert.IsFalse((string)json.Foo == null);
+            Assert.IsFalse((string)json.Bar.B == null);
+            Assert.IsFalse((string)json.Baz == null);
```
christothes, line 316: What happens if we access a property on a non-existent property? e.g. `json.Baz.Fizz`?

annelo-msft: If you called `json.Baz.Fizz` on this example, it would throw because `json.Baz` is a string type in the JSON, and not an object that can have properties.

christothes: What would the exception be?

KrzysztofCwalina: @annelo-msft, why is the cast to string required? Can we overload == on DynamicJson to avoid it?

annelo-msft: > @annelo-msft, why is the cast to string required? Can we overload == on DynamicJson to avoid it?

Yes, apologies that got lost across versions.  I'll add it back.

annelo-msft: > What would the exception be?

`System.InvalidOperationException: 'Expected an 'Object' type but was String.'`

christothes: Do we have/need tests for these invalid property access scenarios?

jsquire:
> We could say that to disambiguate this, you need to convert to MutableJsonDocument or JsonDocument.

I've no issue with it being off the main path like this, so long as we have a way to do so.   I can see that being a frequently used troubleshooting step for incoming issues.

sdk/core/Azure.Core.Experimental/tests/DynamicJsonTests.cs
annelo-msft: Not yet, I'll add them.  Thanks!

sdk/core/Azure.Core.Experimental/src/DynamicJson.cs
annelo-msft: We do, and actually staring at this, I think there may be a bug here.  Will check and update.

sdk/core/Azure.Core.Experimental/tests/DynamicJsonTests.cs
annelo-msft: > Why is the cast to string required? Can we overload == on DynamicJson to avoid it?

@KrzysztofCwalina, with overloads for `==`, when comparing a dynamic to `null`, we don't know the type of either, so you get an ambiguous call site, e.g.:

```
    Microsoft.CSharp.RuntimeBinder.RuntimeBinderException : The call is ambiguous between the following methods or properties: 'Azure.Core.Dynamic.DynamicJson.operator ==(Azure.Core.Dynamic.DynamicJson, string)' and 'Azure.Core.Dynamic.DynamicJson.operator ==(string, Azure.Core.Dynamic.DynamicJson)'
```

Without the overload and without the cast, it compares null to the returned DynamicJson object, which has a value so it's not null.  

One option is to return an actual `null` from GetProperty() if the property is missing.  We would lose the ability to treat is as a DynamicJson, but I'm not sure that's a problem.

KrzysztofCwalina: > One option is to return an actual null from GetProperty() if the property is missing. We would lose the ability to treat it as a DynamicJson, but I'm not sure that's a problem.

This is an interesting idea to try

annelo-msft: I've added follow-up items to track both these requests to https://github.com/Azure/azure-sdk-for-net/issues/33856

sdk/core/Azure.Core.Experimental/src/DynamicJson.cs
annelo-msft: Changed the approach and now return a property `null` instead of DynamicJson wrapping a null JsonValueKind.

PR #34032, opened 2023-02-09
add MicrosoftBotIdentifier

ostoliarova-ee:
### Packages impacted by this PR
Communication - Common

### Issues associated with this PR
[User Story 3112539](https://skype.visualstudio.com/SPOOL/_workitems/edit/3112539): [SDK][Common][JS] As a developer I can identifer if a user is a Teams User or Bot

### Describe the problem that is addressed by this PR
Introducing MicrosoftBotIdentifier, which is a new type of CommunicationIdentifier.

### Are there test cases added in this PR? _(If not, why?)_
Added test cases for creating new MicrosoftTeamsBotIdentifier and converting from and to RawId.

### Provide a list of related PRs _(if any)_
https://github.com/Azure/azure-sdk-for-js/pull/24569
https://github.com/Azure/azure-sdk-for-net/pull/34032
https://github.com/Azure/azure-sdk-for-java/pull/33634
https://github.com/Azure/azure-sdk-for-python/pull/28910
https://github.com/Azure/azure-sdk-for-android/pull/1307
https://github.com/Azure/azure-sdk-for-ios/pull/1468

### Checklists
- [x] Added impacted package name to the issue description
- [ ] Does this PR needs any fixes in the SDK Generator?** _(If so, create an Issue in the [Autorest/typescript](https://github.com/Azure/autorest.typescript) repository and link it here)_
- [x] Added a changelog (if necessary)


sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
```diff
@@ -61,7 +61,13 @@ public static CommunicationIdentifier FromRawId(string rawId)
 
             var segments = rawId.Split(':');
             if (segments.Length < 3)
+            {
+                if (segments.Length == 2 && segments[0] == "28")
+                {
+                    return new MicrosoftBotIdentifier(segments[1], true, CommunicationCloudEnvironment.Public);
```
petrsvihlik, line 67: this section is starting to get a little messy...5 hardcoded numeric constants on 5 rows. we should keep an eye on this section in the future and possibly annotate it better, refactor it, or use some almighty regex to reduce the number of conditions.

petrsvihlik (approved):

sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
ostoliarova-msft: Let's keep this in mind for the future!

sdk/communication/Azure.Communication.Common/tests/CommunicationIdentifierTest.cs
```diff
@@ -26,6 +27,13 @@ public void RawIdTakesPrecendenceInEqualityCheck()
 
             Assert.AreEqual(new PhoneNumberIdentifier("+override", "4:14255550123"), new PhoneNumberIdentifier("14255550123"));
             Assert.AreEqual(new PhoneNumberIdentifier("14255550123"), new PhoneNumberIdentifier("+override", "4:14255550123"));
+
```
AikoBB, line 30: can we add tests for the following scenarios:

- if the bot id is null or empty
- if the isGlobal is not provided, it should be false
- the default cloud is public one

sdk/communication/Azure.Communication.Common/src/MicrosoftBotIdentifier.cs
```diff
@@ -0,0 +1,98 @@
...
+                        _rawId = $"28:orgid:{BotId}";
+                    }
+                }
+
+                return _rawId;
+            }
+        }
+
+        /// <summary>The id of the Microsoft bot.</summary>
+        public string BotId { get; }
```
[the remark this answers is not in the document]
AikoBB: let's keep the `botId` one and pls update the naming for js

sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
```diff
@@ -61,7 +61,13 @@ public static CommunicationIdentifier FromRawId(string rawId)
...
+                }
                 return new UnknownIdentifier(rawId);
+            }
```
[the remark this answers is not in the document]
ostoliarova-msft: The result of this operations could not be consistent if there is another `:` in the identifier. The current approach is consistent across all platforms and will use the remaining part of the string (no matter if it has `:` in it or not). The approach you're proposing will only return the third element of the initial list (no matter if there are more elements in it). this might fail in the future with other identifiers.

sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
```diff
@@ -73,6 +79,11 @@ public static CommunicationIdentifier FromRawId(string rawId)
                 "8:dod:" => new MicrosoftTeamsUserIdentifier(suffix, false, CommunicationCloudEnvironment.Dod),
                 "8:gcch:" => new MicrosoftTeamsUserIdentifier(suffix, false, CommunicationCloudEnvironment.Gcch),
                 "8:acs:" or "8:spool:" or "8:dod-acs:" or "8:gcch-acs:" => new CommunicationUserIdentifier(rawId),
+                "28:gcch-global:" => new MicrosoftBotIdentifier(suffix, true, CommunicationCloudEnvironment.Gcch),
```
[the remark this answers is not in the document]
ostoliarova-msft: Thanks! Good point!

sdk/communication/Azure.Communication.Common/tests/CommunicationIdentifierTest.cs
ostoliarova-msft: It is impossible to create a MicrosoftBotIdentifier with empty/null id. It will throw an exception.
I added tests for the second two proposals

sdk/communication/Azure.Communication.Common/src/MicrosoftBotIdentifier.cs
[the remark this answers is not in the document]
ostoliarova-msft: Done!

sdk/communication/Azure.Communication.Common/tests/CommunicationIdentifierTest.cs
AikoBB: so we have to test if the exception will be thrown for null or empty param, makes it sense?

sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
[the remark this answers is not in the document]
AikoBB: we have a fixed MRI format and so far there was no change, even if there will be another identifier it will fall under the unknown identifier and there are checks for the known one by prefixes, and we're pretty sure about their formats(as the third element is id) and for them should not be any changes.

sdk/communication/Azure.Communication.Common/tests/CommunicationIdentifierTest.cs
ostoliarova-msft: We're not testing it for any other identifier as we rely that the  Argument.AssertNotNullOrEmpty always performs the check.
@petrsvihlik do I need to add additional tests for all identifiers to check that input parameters are validated by  Argument.AssertNotNullOrEmpty?

sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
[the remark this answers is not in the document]
ostoliarova-msft: Ok

sdk/communication/Azure.Communication.Common/tests/CommunicationIdentifierTest.cs
AikoBB: my intention here was not about whether the Argument.AssertNotNullOrEmpty validates or not, it's more about the logic within the constructor, there should be a check for the botId param, and if the botId is invalid it should throw an exception and it will also be a guard for the future from removing the check from constructor.

ostoliarova-msft: As this is similar to other identifiers (they have the same constructor check), let's wait for the input from @petrsvihlik on whether we need to add and additional test. If we're to add one, ot should be consistent for other identifiers, too.

sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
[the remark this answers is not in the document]
petrsvihlik: I'm not quite convinced about this change.
> even if there will be another identifier it will fall under the unknown identifier

@AikoBB yes but we might be sending an incomplete payload to the UnknownIdentifier's constructor... e.g., what if a future identifier will have more than 3 segments?

sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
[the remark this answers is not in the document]
petrsvihlik: I was a bit reserved regarding this change. One one hand it reduces code duplication, on the other hand it reduces readability....it's about finding the balance between the two. But ok, since you've already went on with it, let's keep it this way.
And I think we were discussing it with @AikoBB the first time we introduced the From/ToRawID functionality and saying something along the lines that if we add more identifiers, we'll refactor it this way.

sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
```diff
@@ -54,25 +70,36 @@ public static CommunicationIdentifier FromRawId(string rawId)
...
+            if (rawId.StartsWith(Phone, StringComparison.OrdinalIgnoreCase))
             {
-                return new PhoneNumberIdentifier(rawId.Substring("4:".Length));
+                return new PhoneNumberIdentifier(rawId.Substring(Phone.Length));
             }
 
             var segments = rawId.Split(':');
             if (segments.Length < 3)
+            {
+                if (segments.Length == 2 && segments[0] == "28")
```
petrsvihlik, line 81: I would then use rawId.StartsWith(Bot) to reuse the constants as much as possible.

sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
```diff
@@ -54,25 +70,36 @@ public static CommunicationIdentifier FromRawId(string rawId)
...
+                TeamUserAnonymous => new MicrosoftTeamsUserIdentifier(suffix, true),
+                TeamUserPublicCloud => new MicrosoftTeamsUserIdentifier(suffix, false, CommunicationCloudEnvironment.Public),
+                TeamUserDodCloud => new MicrosoftTeamsUserIdentifier(suffix, false, CommunicationCloudEnvironment.Dod),
+                TeamUserGcchCloud => new MicrosoftTeamsUserIdentifier(suffix, false, CommunicationCloudEnvironment.Gcch),
+                AcsUser or SpoolUser or AcsUserDodCloud or AcsUserGcchCloud => new CommunicationUserIdentifier(rawId),
+                BotGcchCloudGlobal => new MicrosoftBotIdentifier(suffix, true, CommunicationCloudEnvironment.Gcch),
+                BotPublicCloud => new MicrosoftBotIdentifier(suffix, false, CommunicationCloudEnvironment.Public),
+                BotDodCloudGlobal => new MicrosoftBotIdentifier(suffix, true, CommunicationCloudEnvironment.Dod),
+                BotGcchCloud => new MicrosoftBotIdentifier(suffix, false, CommunicationCloudEnvironment.Gcch),
+                BotDodCloud => new MicrosoftBotIdentifier(suffix, false, CommunicationCloudEnvironment.Dod),
```
petrsvihlik, line 102: I would expect the constants to be replaced in the [`MicrosoftTeamsUserIdentifier`](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/communication/Azure.Communication.Common/src/MicrosoftTeamsUserIdentifier.cs) and other identifier classes as well.

sdk/communication/Azure.Communication.Common/src/MicrosoftBotIdentifier.cs
```diff
@@ -0,0 +1,98 @@
...
+        /// <param name="isGlobal">Set this to true if the bot is global.</param>
+        /// <param name="cloud">The cloud that the Microsoft bot belongs to. A null value translates to the Public cloud.</param>
+        /// <param name="rawId">Raw id of the Microsoft bot optional.</param>
+        /// <exception cref="System.ArgumentNullException">
+        /// Thrown when the <paramref name="botId"/> is null.
+        /// </exception>
+        /// <exception cref="System.ArgumentException">
+        /// Thrown when the <paramref name="botId"/> is empty.
+        /// </exception>
+        public MicrosoftBotIdentifier(string botId, bool isGlobal = false, CommunicationCloudEnvironment? cloud = null, string rawId = null)
```
petrsvihlik, line 81: I don't think we need the `string rawId = null` param here. For `MicrosoftTeamsUserIdentifier` it was some kind of legacy identifier support.

petrsvihlik: @AikoBB fyi

sdk/communication/Azure.Communication.Common/tests/CommunicationIdentifierTest.cs
petrsvihlik: the constructor's fine. @AikoBB just wants to add tests to check that if you provide the constructor with null or empty, it'll throw and that if you initialize it with default values, it'll be set to global+public. I'm fine with adding such tests.

petrsvihlik (changes requested):

sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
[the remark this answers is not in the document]
AikoBB: the unknown identififer is initiated by the rawid as follow:
`new UnknownIdentifier(rawId)`

I guess it was your concern and we not providing here the suffix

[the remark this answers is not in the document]
petrsvihlik: yeah, then it's fine I guess. I thought it was also being initialized using the suffix, sorry for missing that.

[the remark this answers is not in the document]
petrsvihlik: playing a devils advocate here - what about the case when a new `MicrosoftTeamsUserIdentifier` format will be introduced with more than 3 segments?

sdk/communication/Azure.Communication.Common/src/MicrosoftBotIdentifier.cs
AikoBB: so it means the raw id should be a mandatory parameter for initialization? should we also check for its validness(not null and empty)?

sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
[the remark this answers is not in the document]
petrsvihlik: and I honestly hope this will not happen :)

sdk/communication/Azure.Communication.Common/src/MicrosoftBotIdentifier.cs
petrsvihlik: no...the rawId shouldn't be there at all. you should use `CommunicationIdentifier.FromRawId()` if you want to initialize the identifier from rawId. we keep the rawId param in the constructors of MSTeamsUserIdentifier and PhoneNumber identifier for backward compatability reasons but we don't need to introduce it for new identifiers.

sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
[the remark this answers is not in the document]
AikoBB: good point, even for known identifiers if there will be introduced another segment we have to identify which part will the id part by releasing a new version, so makes sense to extend the check in this [line](https://github.com/Azure/azure-sdk-for-net/blob/b1a08d0b333e0a30edd12a46d4a22e3dd8c944d7/sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs#L79) maybe, wdyt?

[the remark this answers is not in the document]
petrsvihlik: > we have to identify which part will the id part by releasing a new version
exactly

>extend the check in this [line](https://github.com/Azure/azure-sdk-for-net/blob/b1a08d0b333e0a30edd12a46d4a22e3dd8c944d7/sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs#L79) maybe

yeah and add a couple more tests for the scenarios we've just described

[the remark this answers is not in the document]
ostoliarova-msft: OK

sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
ostoliarova-msft: OK

sdk/communication/Azure.Communication.Common/src/MicrosoftBotIdentifier.cs
ostoliarova-msft: OK, I'll remove this parameter

sdk/communication/Azure.Communication.Common/tests/CommunicationIdentifierTest.cs
ostoliarova-msft: Added new tests

sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
AikoBB: maybe `rawId.StartsWith(Bot, StringComparison.OrdinalIgnoreCase)` will help :)

sdk/communication/Azure.Communication.Common/src/MicrosoftBotIdentifier.cs
```diff
@@ -0,0 +1,96 @@
...
+                }
+
+                return _rawId;
+            }
+        }
+
+        /// <summary>The id of the Microsoft bot.</summary>
+        public string BotId { get; }
+
+        /// <summary>True if the bot is global and false (or missing) if the bot is tenantized.</summary>
```
AikoBB, line 62:
```suggestion
        /// <summary>True if the bot is global and false if the bot is tenantized.</summary>
```

sdk/communication/Azure.Communication.Common/src/MicrosoftBotIdentifier.cs
```diff
@@ -0,0 +1,96 @@
...
+        public bool IsGlobal { get; }
+
+        /// <summary> The cloud that the bot belongs to. </summary>
+        public CommunicationCloudEnvironment Cloud { get; }
+
+        /// <summary>
+        /// Initializes a new instance of <see cref="MicrosoftBotIdentifier"/>.
+        /// </summary>
+        /// <param name="botId">Id of the Microsoft bot. The unique Microsoft app ID for the bot as registered with the Bot Framework.</param>
+        /// <param name="isGlobal">Set this to true if the bot is global.</param>
```
AikoBB, line 72:
```suggestion
        /// <param name="isGlobal">Set this to true if the bot is global. If not provided, the default value will be false.</param>
```

sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
ostoliarova-msft: Updated

sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
```diff
@@ -54,25 +70,36 @@ public static CommunicationIdentifier FromRawId(string rawId)
...
-            if (rawId.StartsWith("4:", StringComparison.OrdinalIgnoreCase))
+            if (rawId.StartsWith(Phone, StringComparison.OrdinalIgnoreCase))
             {
-                return new PhoneNumberIdentifier(rawId.Substring("4:".Length));
+                return new PhoneNumberIdentifier(rawId.Substring(Phone.Length));
             }
 
             var segments = rawId.Split(':');
-            if (segments.Length < 3)
+            if (segments.Length > 3 || segments.Length < 3)
```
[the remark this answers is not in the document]
ostoliarova-msft: Updated

sdk/communication/Azure.Communication.Common/src/MicrosoftBotIdentifier.cs
ostoliarova-msft: Thanks, updated

sdk/communication/Azure.Communication.Common/src/MicrosoftBotIdentifier.cs
ostoliarova-msft: Updated

sdk/communication/Azure.Communication.Common/src/CommunicationIdentifier.cs
petrsvihlik: > maybe `rawId.StartsWith(Bot, StringComparison.OrdinalIgnoreCase)` will help :)

@ostoliarova-msft @AikoBB isn't that what I said... :DD

anyways, good that you resolved it 👍

petrsvihlik (approved):

AikoBB (approved):

sdk/communication/Azure.Communication.Common/CHANGELOG.md
```diff
@@ -1,10 +1,11 @@
 # Release History
 
-## 1.3.0-beta.1 (Unreleased)
+## 2.0.0-beta.1 (Unreleased)
 
 ### Features Added
 
 ### Breaking Changes
+- Added support for a new communication identifier `MicrosoftBotIdentifier`.
```
AikoBB, line 8: should be it under the added features?

ostoliarova-msft: It's by ARB request, please see my comment https://github.com/Azure/azure-sdk-for-js/pull/24569#discussion_r1133783718

ostoliarova-msft: Updated

petrsvihlik (approved):

AikoBB (approved):

PR #34036, opened 2023-02-09
Fix creat recognize request of call automation media

wangrui-msft:

The createRecognizeRequests needs to map the cognitionconfiguration field called InitialSilenceTimeout to InitialSilenceTimeoutInSeconds. Currently the recognizeRequestInternal for recognize choice miss this field. This fix adds the population of the field from cognitionconfiguration

### DIFF
--- a/sdk/communication/Azure.Communication.CallAutomation/src/CallMedia.cs
+++ b/sdk/communication/Azure.Communication.CallAutomation/src/CallMedia.cs
@@ -311,6 +311,7 @@ namespace Azure.Communication.CallAutomation
                 RecognizeOptionsInternal recognizeConfigurationsInternal = new RecognizeOptionsInternal(CommunicationIdentifierSerializer.Serialize(recognizeChoiceOptions.TargetParticipant))
                 {
                     InterruptPrompt = recognizeChoiceOptions.InterruptPrompt,
+                    InitialSilenceTimeoutInSeconds = (int)recognizeChoiceOptions.InitialSilenceTimeout.TotalSeconds
                 };
 
                 recognizeChoiceOptions.RecognizeChoices


cochi2 (commented):
Could you please provide a description of the issue, what is the solution, and also test cases that cover it?

wangrui-msft:
> Could you please provide a description of the issue, what is the solution, and also test cases that cover it?

Yes, basically it adds the missing  InitialSilenceTimeoutInSeconds field in  the RecongnizeRequest. The error is found at E2E validation, and should be working at the sample E2E application.

Paresh-Arvind-Patil:
> Could you please provide a description of the issue, what is the solution, and also test cases that cover it?

@cochi2 , The issue is that InitialSilenceTimeoutInSeconds is not passed to PMA from SDK, this fix should resolve it.
@wangrui-msft Just want to confirm did we test this ?

Paresh-Arvind-Patil (approved):

sdk/communication/Azure.Communication.CallAutomation/src/CallMedia.cs
```diff
@@ -311,6 +311,7 @@ private static RecognizeRequestInternal CreateRecognizeRequest(CallMediaRecogniz
                 RecognizeOptionsInternal recognizeConfigurationsInternal = new RecognizeOptionsInternal(CommunicationIdentifierSerializer.Serialize(recognizeChoiceOptions.TargetParticipant))
                 {
                     InterruptPrompt = recognizeChoiceOptions.InterruptPrompt,
+                    InitialSilenceTimeoutInSeconds = (int)recognizeChoiceOptions.InitialSilenceTimeout.TotalSeconds
```
cochi2, line 314: can't remember it clearly but, are we 100% sure that InitialSilenceTimout cannot be null?

bharat-kalyan-namburi (approved):

sdk/communication/Azure.Communication.CallAutomation/src/CallMedia.cs
Paresh-Arvind-Patil: @cochi2 , Yes InitialSilenceTimout can be null, The issue is even if we are setting the value on SDK the InitialSilenceTimout property is not getting passed to PMA
For example i have set the timeout to 40secs and on PMA i dont see the property getting passed 
PMA Logs: https://portal.microsoftgeneva.com/s/5C10F776

cochi2: If `InitialSilenceTimeout` can be null, this line will throw a NullPointerException.

wangrui-msft:
Basically it is consistent with dtmf tone setup. As dtmf serialize works, the choice recognize also work as expected as they inherit the same class.

PR #34044, opened 2023-02-09
[Event Hubs] Stress Tests Updates

m-redding:
These changes do the following:
- Upgrades stress test addons to the most recent version
- Upgrades to scenario matrix
- Updates Docker file to use newest version of .NET SDK
- Adds TestScenario interface that greatly simplifies current code and makes it easier to add new test scenarios


sdk/eventhub/Azure.Messaging.EventHubs/stress/src/Scenarios/TestScenario.cs
```diff
@@ -1,17 +1,97 @@
...
-public enum TestScenario
+public abstract class TestScenario
 {
-    BufferedProducerTest,
-    BurstBufferedProducerTest,
-    EventProducerTest,
-    ProcessorTest,
-    ConsumerTest
+    /// <summary>The <see cref="TestParameters"/> used to configure this test scenario.</summary>
+    internal readonly TestParameters _testParameters;
```
jsquire, line 19: If we're going to make these internal, I'd suggest making them properties.  The `_Name` pattern is generally used to imply "this is a class member used only by the class".

jsquire (approved):

PR #34045, opened 2023-02-09
Remove virtual from ToBytes method

JoshLove-msft:
No need to mock the method call as it is not a service call.

### DIFF
--- a/sdk/core/Azure.Core.Amqp/src/AmqpAnnotatedMessage.cs
+++ b/sdk/core/Azure.Core.Amqp/src/AmqpAnnotatedMessage.cs
@@ -161,7 +161,7 @@ namespace Azure.Core.Amqp
         /// <seealso href="http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#section-message-format"/>
         /// </summary>
         /// <returns>A <see cref="BinaryData"/> instance representing the serialized message.</returns>
-        public virtual BinaryData ToBytes()
+        public BinaryData ToBytes()
         {
             var stream = AmqpAnnotatedMessageConverter.ToAmqpMessage(this).ToStream();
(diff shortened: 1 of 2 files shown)


sdk/core/Azure.Core.Amqp/src/AmqpAnnotatedMessage.cs
```diff
@@ -161,7 +161,7 @@ section switch
         /// <seealso href="http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#section-message-format"/>
         /// </summary>
         /// <returns>A <see cref="BinaryData"/> instance representing the serialized message.</returns>
-        public virtual BinaryData ToBytes()
+        public BinaryData ToBytes()
```
jsquire, line 164: We're not concerned with testing/mocking here?   That seems odd.

JoshLove-msft: Well mocking is not needed as there are no service calls. Customers can just create an instance of the message and let the serialization do its thing. Is there some scenario where you expect customers to want to mock?

jsquire (approved):
Approving to unblock, probably good to get architect feedback on whether it's needed.

JoshLove-msft:
> Approving to unblock, probably good to get architect feedback on whether it's needed.

/cc @KrzysztofCwalina @tg-msft thoughts?

JoshLove-msft:
> > Approving to unblock, probably good to get architect feedback on whether it's needed.
> 
> /cc @KrzysztofCwalina @tg-msft thoughts?



> > Approving to unblock, probably good to get architect feedback on whether it's needed.
> 
> /cc @KrzysztofCwalina @tg-msft thoughts?

Confirmed that it is correct to leave off virtual here since there are public methods to create the type.

Issue #34077, opened 2023-02-10
[BUG] Can not get secret from KeyVault on account of failing ManagedIdentityCredential GetToken method in Gov and China clouds


JacobBovee:
### Library name and version

Azure.Identity 1.8.2, Azure.Security.KeyVault.Secrets 4.4.0

### Describe the bug

Hello, we are running into a scenario in the non-public clouds (reproduced in both UsGov and ChinaCloud) in which we are initializing a `ManagedIdentityCredential`, and using the resulting credential to initialize a secret client which we then use to obtain a secret from KeyVault. Unfortunately our call to `GetSecret` results in this error:
```
ManagedIdentityCredential authentication failed: 

 The application is configured for cloud login.microsoftonline.com and the request for a different cloud - login.microsoftonline.us. This is not supported - the app and the request must target the same cloud. 

See https://aka.ms/msal-net-authority-override for details
See the troubleshooting guide for more information. https://aka.ms/azsdk/net/identity/managedidentitycredential/troubleshoot 

 The application is configured for cloud login.microsoftonline.com and the request for a different cloud - login.microsoftonline.us. This is not supported - the app and the request must target the same cloud. 

See https://aka.ms/msal-net-authority-override for details 
```  
We have the AuthorityHost set correctly and the managed identity is created in the correct cloud. 

Through some manual testing I've found that if we initialize as a `DefaultAzureCredential` and pass in a managed identity client id the secrets library is able to successfully call `GetToken` on the `DefaultAzureCredential` which in turn successfully calls `ManagedIdentityCredential`'s get token method! Here is some telemetry of one of our successful calls vs a failing call, notice that they are both in the end simply calling `ManagedIdentityCredential`'s `GetToken` method. 

![09fc7a70-8c76-464f-ac36-54eb3bb2f1bf](https://user-images.githubusercontent.com/6301646/218162405-8a33dd78-e105-4837-a609-43d7b910407b.jpg)
![e493137a-8a8d-4197-bdf3-d0022e4ced58](https://user-images.githubusercontent.com/6301646/218162418-eb728cea-5646-4465-8887-0601fa152f4d.jpg)

Thanks for taking a look, really appreciate any insight on this. 

### Expected behavior

In public cloud we are simply able to call `GetSecret` and the key vault returns our secret. 

### Actual behavior

As mentioned above in non-public clouds we are unable to access key vault.

### Reproduction Steps

I've managed to reproduce it with just this:

```c#
[...]

PR #34078, opened 2023-02-10
Ensure Options are passed down to ManagedIdentityClient

pharring:
Specifically, this is to pass the AuthorityHost to the MSAL client

### DIFF
--- a/sdk/identity/Azure.Identity/src/Credentials/ManagedIdentityCredential.cs
+++ b/sdk/identity/Azure.Identity/src/Credentials/ManagedIdentityCredential.cs
@@ -45,9 +45,11 @@ namespace Azure.Identity
         /// </param>
         /// <param name="options">Options to configure the management of the requests sent to the Azure Active Directory service.</param>
         public ManagedIdentityCredential(string clientId = null, TokenCredentialOptions options = null)
-            : this(clientId, CredentialPipeline.GetInstance(options))
+            : this(
+                new ManagedIdentityClient(new ManagedIdentityClientOptions { ClientId = clientId, Pipeline = CredentialPipeline.GetInstance(options), Options = options }))
         {
             _logAccountDetails = options?.Diagnostics?.IsAccountIdentifierLoggingEnabled ?? false;
+            _clientId = clientId;
         }
 
         /// <summary>
@@ -60,7 +62,7 @@ namespace Azure.Identity
         /// <param name="options">Options to configure the management of the requests sent to the Azure Active Directory service.</param>
         public ManagedIdentityCredential(ResourceIdentifier resourceId, TokenCredentialOptions options = null)
             : this(
-                new ManagedIdentityClient(new ManagedIdentityClientOptions { ResourceIdentifier = resourceId, Pipeline = CredentialPipeline.GetInstance(options) }))
+                new ManagedIdentityClient(new ManagedIdentityClientOptions { ResourceIdentifier = resourceId, Pipeline = CredentialPipeline.GetInstance(options), Options = options }))
         {
             _logAccountDetails = options?.Diagnostics?.IsAccountIdentifierLoggingEnabled ?? false;
             _clientId = resourceId.ToString();
--- a/sdk/identity/Azure.Identity/src/Credentials/ManagedIdentityCredential.cs
+++ b/sdk/identity/Azure.Identity/src/Credentials/ManagedIdentityCredential.cs
@@ -45,11 +45,8 @@ namespace Azure.Identity
         /// </param>
         /// <param name="options">Options to configure the management of the requests sent to the Azure Active Directory service.</param>
         public ManagedIdentityCredential(string clientId = null, TokenCredentialOptions options = null)
-            : this(
-                new ManagedIdentityClient(new ManagedIdentityClientOptions { ClientId = clientId, Pipeline = CredentialPipeline.GetInstance(options), Options = options }))
+            : this(clientId, CredentialPipeline.GetInstance(options), options)
         {
-            _logAccountDetails = options?.Diagnostics?.IsAccountIdentifierLoggingEnabled ?? false;
-            _clientId = clientId;
         }
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


pharring:
Probably needs unit tests to show the problem. Note that DefaultAzureCredentialFactory.CreateManagedIdentityCredential already gets it right.

pharring:
I believe this is the fix for #34077

pharring:
@schaabs ping me when you get a chance because I want to discuss this with you. There are a 3 different fixes in here. I think any one of them would fix the bug. Do we want to include all three?

sdk/identity/Azure.Identity/src/MsalConfidentialClient.cs
```diff
@@ -146,7 +146,7 @@ protected override async ValueTask<IConfidentialClientApplication> CreateClientA
 
             if (!string.IsNullOrEmpty(tenantId))
             {
-                builder.WithAuthority(Pipeline.AuthorityHost.AbsoluteUri, tenantId);
+                builder.WithTenantId(tenantId);
```
pharring, line 149: This code path is hit by the ChallengeBasedAuthenticationPolicy in KeyVault. The `tenantId` comes from the challenge response. If `Pipeline.AuthorityHost` doesn't agree with the authority passed to the ConfidentialClientBuilder, then we hit the MSAL exception about mismatched authorities (https://aka.ms/msal-net-authority-override).

Switching to `.WithTenantId` sidesteps the issue of knowing what authority was passed to the ConfidentialClientBuilder.

sdk/identity/Azure.Identity/tests/ManagedIdentityCredentialImdsLiveTests.cs
```diff
@@ -74,9 +74,7 @@ private ManagedIdentityCredential CreateManagedIdentityCredential(string clientI
         {
             options = InstrumentClientOptions(options ?? new TokenCredentialOptions());
 
-            var pipeline = CredentialPipeline.GetInstance(options);
-
-            var cred = new ManagedIdentityCredential(new ManagedIdentityClient(pipeline, clientId));
+            var cred = new ManagedIdentityCredential(clientId, options);
```
pharring, line 77: Use the public constructor instead of the internal one, now that the public one passes through options correctly.

sdk/identity/Azure.Identity/tests/Mock/TestDefaultAzureCredentialFactory.cs
```diff
@@ -25,7 +25,7 @@ public override TokenCredential CreateEnvironmentCredential()
             => new EnvironmentCredential(Pipeline, Options);
 
         public override TokenCredential CreateManagedIdentityCredential()
-            => new ManagedIdentityCredential(new ManagedIdentityClient(Pipeline, Options.ManagedIdentityClientId));
+            => new ManagedIdentityCredential(Options.ManagedIdentityClientId, Pipeline, Options);
```
pharring, line 28: Again, using the public constructor instead of the internal one.

sdk/identity/Azure.Identity/src/MsalConfidentialClient.cs
```diff
@@ -58,7 +58,7 @@ public MsalConfidentialClient(CredentialPipeline pipeline, string tenantId, stri
             : base(pipeline, tenantId, clientId, options)
         {
             _appTokenProviderCallback = appTokenProviderCallback;
-            _authority = options?.AuthorityHost ?? AzureAuthorityHosts.AzurePublicCloud;
+            _authority = options?.AuthorityHost ?? pipeline.AuthorityHost ?? AzureAuthorityHosts.AzurePublicCloud;
```
pharring, line 61: This isn't strictly necessary, but it seems like a reasonable fallback. What do you think?

sdk/identity/Azure.Identity/src/Credentials/ManagedIdentityCredential.cs
```diff
@@ -45,9 +45,8 @@ protected ManagedIdentityCredential()
         /// </param>
         /// <param name="options">Options to configure the management of the requests sent to the Azure Active Directory service.</param>
         public ManagedIdentityCredential(string clientId = null, TokenCredentialOptions options = null)
-            : this(clientId, CredentialPipeline.GetInstance(options))
+            : this(clientId, CredentialPipeline.GetInstance(options), options)
```
pharring, line 48: This (and the related changes in constructors below) is really the key fix. Ensuring that `options` gets passed down to the inner ManagedIdentityClient and eventually into the MsalConfidentialClient.
This is the big difference between using the ManagedIdentityCredential stand-alone versus via DefaultAzureCredential. DefaultAzureCredential's factory calls one of the internal constructors, correctly passing in Options. So what I'm doing here is making sure the two approaches have the same behavior.

This fix alone would be enough to fix the bug.

sdk/identity/Azure.Identity/tests/ManagedIdentityCredentialTests.cs
```diff
@@ -137,9 +137,7 @@ public async Task VerifyImdsRequestWithClientIdAndNonPubCloudMockAsync(Uri autho
             var pipeline = new CredentialPipeline(authority, _pipeline, new ClientDiagnostics(options));
 
             ManagedIdentityCredential credential = InstrumentClient(
-                new ManagedIdentityCredential(
-                    new ManagedIdentityClient(
-                        new ManagedIdentityClientOptions { Pipeline = pipeline, ClientId = "mock-client-id", Options = options })));
+                new ManagedIdentityCredential("mock-client-id", pipeline, options));
```
pharring, line 140: Again, making use of the public constructor instead of the internal one.

christothes (commented):
Thanks so much for taking the time to submit this PR! We saw the bug and came to the same conclusion for the primary fix.

However, I've already started work on a fix that includes some additional refactoring to avoid this kind of confusion going forward.

sdk/identity/Azure.Identity/tests/ManagedIdentityCredentialTests.cs
```diff
@@ -137,11 +137,9 @@ public async Task VerifyImdsRequestWithClientIdAndNonPubCloudMockAsync(Uri autho
...
 
-            AccessToken actualToken = await credential.GetTokenAsync(new TokenRequestContext(MockScopes.Default));
+            AccessToken actualToken = await credential.GetTokenAsync(new TokenRequestContext(MockScopes.Default, tenantId: "mock-tenant-id"));
```
pharring, line 142: Adding the tenantId here exercises the code-path that hits the original bug.

pharring:
@christothes Great. Put me on the final PR when you have it.

PR #34082, opened 2023-02-10
Allow camelCase JSON properties to be accessed with PascalCase DynamicJson property accessors

annelo-msft:
The primary goal of `DynamicJson` is to help developers work with Azure SDK client protocol methods.  These methods essentially take and return raw JSON (`RequestContent` as an input parameter and `Response.Content` as output return value), instead of strongly-typed models.  `DynamicJson` lets users treat these values like C# POCOs using [the C# `dynamic` type](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/types/using-type-dynamic).

One challenge of this is that Azure APIs [use camelCased JSON](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#json), where C# style conventions use PascalCased property names.

This PR introduces a feature allowing callers to use PascalCased property names with DynamicJson while maintaining camelCased JSON needed for interaction with the service.  It also lets callers configure DynamicJson to use strict casing, and allows service libraries to provide a configuration that works with their service in the event they need to diverge from Azure API guidelines.

In addition, callers can always use DynamicJson indexers to get and set properties with exact string matches on property names.


sdk/core/Azure.Core.Experimental/src/DynamicJsonOptions.cs
```diff
@@ -0,0 +1,20 @@
...
+    /// Provides the ability for the user to define custom behavior when accessing JSON through a dynamic layer.
+    /// </summary>
+    public struct DynamicJsonOptions
+    {
+        /// <summary>
+        /// Specifies whether properties in the <see cref="DynamicJson"/> can be read
+        /// as with either "PascalCase" or "camelCase" property names. If set to false,
+        /// property reads are case sensitive.
+        /// </summary>
+        public bool AccessPropertyNamesPascalOrCamelCase { get; set; }
```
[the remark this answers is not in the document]
annelo-msft: I'm going to take another pass on this design to see if we can simplify the naming for this.

sdk/core/Azure.Core.Experimental/src/DynamicJson.cs
```diff
@@ -38,7 +40,39 @@ internal override void WriteTo(Stream stream)
...
+            }
+
+            // Either PascalCase or camelCase lookup failed. Try the other.
+            string otherCaseName = GetAsOtherCasing(name);
+            if (_element.TryGetProperty(otherCaseName, out property))
+            {
+                return new DynamicJson(property);
+            }
+
+            return new InvalidOperationException($"JSON does not contain property called {name}");
```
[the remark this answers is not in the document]
annelo-msft: I like this idea a lot.

sdk/core/Azure.Core.Experimental/src/BinaryDataExtensions.cs
```diff
@@ -16,7 +16,15 @@ public static class BinaryDataExtensions
         /// </summary>
         public static dynamic ToDynamic(this BinaryData data)
         {
-            return new DynamicJson(MutableJsonDocument.Parse(data).RootElement);
+            return new DynamicJson(MutableJsonDocument.Parse(data).RootElement, new DynamicJsonOptions());
```
annelo-msft, line 19: Default behavior is that you can get/set properties PascalCase (standard .NET style) and new properties are written in the JSON buffer camelCase (Azure API guidelines).

sdk/core/Azure.Core.Experimental/src/DynamicJsonOptions.cs
```diff
@@ -0,0 +1,81 @@
...
+        /// <summary>
+        /// Specifies how properties on <see cref="DynamicJson"/> will get accessed in the underlying JSON buffer.
+        /// </summary>
+        public DynamicJsonPropertyCasing PropertyCasing { get; set; } = DynamicJsonPropertyCasing.Default;
+    }
+
+    /// <summary>
+    /// Casing options for property access on DynamicJson.
+    /// </summary>
+    public struct DynamicJsonPropertyCasing
```
annelo-msft, line 27: These types will move to separate files if we go with this approach.

sdk/core/Azure.Core.Experimental/tests/DynamicJsonTests.cs
```diff
@@ -405,6 +405,93 @@ public void DisposingAChildDisposesTheParent()
...
+
+            DynamicJsonOptions options = new();
+            dynamic dynamicJson = new BinaryData(json).ToDynamic(options);
+
+            dynamicJson.foo = 2;
+
+            Assert.AreEqual(2, (int)dynamicJson.foo);
+            Assert.AreEqual(2, (int)dynamicJson.Foo);
+
+            dynamicJson.Foo = 3;
```
annelo-msft, line 489: Remains to be implemented if we move forward with this approach.

sdk/core/Azure.Core.Experimental/src/BinaryDataExtensions.cs
KrzysztofCwalina: I think we at least need an option parameter to JsonData ctor and to ToDynamic method to control this behavior. Something like:

```csharp
public enum JsonDataNameMapping {
    None,
    PascalCaseGetters,
    PascalCaseGettersCamelCaseSetters // default
}
```

annelo-msft: Does [this method](https://github.com/Azure/azure-sdk-for-net/pull/34082/files#diff-e99fb95729734f4c8b175ed7e726b01d65cf7a34efb5d2a500a2829c357fdab2R25) (just below this comment) address your suggestion?  

Or are you recommending replacing DynamicJsonOptions with the enum you've proposed?  Or a third `ToDynamic()` that takes the simplified `DynamicJsonNameMapping` enum?

annelo-msft: (`JsonData` has been replaced by `DynamicJson` in the new set of types.)

KrzysztofCwalina: Ah, sorry. I did not notice the method below. Thanks!

sdk/core/Azure.Core.Experimental/src/DynamicJsonOptions.cs
KrzysztofCwalina: Can this be simply an enum?

sdk/core/Azure.Core.Experimental/src/DynamicJsonOptions.cs
```diff
@@ -0,0 +1,81 @@
...
+    {
+        /// <summary>
+        /// Creates a new instance of DynamicJsonOptions.
+        /// </summary>
+        public DynamicJsonOptions() { }
+
+        /// <summary>
+        /// Specifies how properties on <see cref="DynamicJson"/> will get accessed in the underlying JSON buffer.
+        /// </summary>
+        public DynamicJsonPropertyCasing PropertyCasing { get; set; } = DynamicJsonPropertyCasing.Default;
```
KrzysztofCwalina, line 21: Are we going to have more options? If not, should the methods just take the enum as opposed to the options? If yes, then maybe we can have overlaods taking just the enum, or a cast from enum to options?

annelo-msft: I can imagine more options we might want - e.g. options to the JsonSerializer we use to do type conversions, possibly options to use a different delimiter than `.` if needed by an API, "StandardFormat" options, etc.  But an options type could be added later if we think casing is the most common thing people would want to configure, and taken as an overload (or provide a cast) per your suggestion.

sdk/core/Azure.Core.Experimental/src/DynamicJsonOptions.cs
annelo-msft: I like the names you suggested in your enum, and if we went with something like that we could.  Thanks for the suggestion!

sdk/core/Azure.Core.Experimental/src/DynamicJsonOptions.cs
annelo-msft: I kept the options type in this version because of the other options I think we'll want to add, but tell me if you hate it or feel strongly about the cast.  I'm leaning against the cast because I don't know if this would be the most common setting people would want to change from the options, and I didn't see casts like this on BCL types.

KrzysztofCwalina (approved):

sdk/core/Azure.Core.Experimental/tests/DynamicJsonTests.cs
```diff
@@ -405,6 +405,297 @@ public void DisposingAChildDisposesTheParent()
...
 
+        [Test]
+        public void CanGetCamelCasePropertyEitherCase()
+        {
+            string json = """{ "foo" : 1 }""";
+
+            dynamic dynamicJson = new BinaryData(json).ToDynamic();
+
+            Assert.AreEqual(1, (int)dynamicJson.foo);
+            Assert.AreEqual(1, (int)dynamicJson.Foo);
```
JoshLove-msft, line 416: Shouldn't this fail? I thought you would need to set PascalCaseGetters here.

sdk/core/Azure.Core.Experimental/tests/DynamicJsonTests.cs
```diff
@@ -405,6 +405,297 @@ public void DisposingAChildDisposesTheParent()
...
+        public void CanGetCamelCasePropertyNoMapping()
+        {
+            string json = """{ "foo" : 1 }""";
+
+            DynamicJsonOptions options = new()
+            {
+                PropertyNameCasing = DynamicJsonNameMapping.None
+            };
+
+            dynamic dynamicJson = new BinaryData(json).ToDynamic(options);
```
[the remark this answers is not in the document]
annelo-msft: Updated.

sdk/core/Azure.Core.Experimental/tests/DynamicJsonTests.cs
```diff
@@ -405,6 +405,308 @@ public void DisposingAChildDisposesTheParent()
...
+        }
+
+        [Test]
+        public void CanSetPascalCasePascalGettersCamelSetters()
+        {
+            string json = """{ "Foo": 1 }""";
+
+            dynamic dynamicJson = new BinaryData(json).ToDynamic();
+
+            // Existing property access does not add a camelCase property.
```
JoshLove-msft, line 687: Why wouldn't new properties just match whatever casing the user specifies via the dynamic?

sdk/core/Azure.Core.Experimental/api/Azure.Core.Experimental.net461.cs
```diff
@@ -151,6 +153,19 @@ public partial struct ArrayEnumerator : System.Collections.Generic.IEnumerable<A
...
         }
     }
+    public enum DynamicJsonNameMapping
+    {
+        None = 0,
+        PascalCaseGetters = 1,
+        PascalCaseGettersCamelCaseSetters = 2,
+    }
+    [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
+    public partial struct DynamicJsonOptions
```
JoshLove-msft, line 163: Is this type necessary yet given that we only have the mapping enum at this point?

sdk/core/Azure.Core.Experimental/tests/DynamicJsonTests.cs
JoshLove-msft: How would users be expected to know this? The default is currently called `None`.

JoshLove-msft: Ah I think I misunderstood - if no mapping is specified, then the behavior is PascalCaseGettersCamelCaseSetters.

JoshLove-msft: I have to say it is slightly unintuitive that the default value for the mapping changes the default behavior that is used when not passing anything.

sdk/core/Azure.Core.Experimental/tests/DynamicJsonTests.cs
annelo-msft: It is because Azure REST API guidance is to camelCase JSON, and .NET guidance is to PascalCase properties.  If services follow the stated guidance, everything should work correctly.  If services do not, users will need to override the default behavior.

sdk/core/Azure.Core.Experimental/tests/DynamicJsonTests.cs
JoshLove-msft: My point is just that the default behavior when the enum is not specified doesn't match the enum default of None. I believe this is unusual in our APIs. I like your suggestion of moving PascalCaseGettersCamelCaseSetters to the 0 value, but in that case it would be awkward to have None as something other than 0, so perhaps that would need to be renamed to something like RawJson.

JoshLove-msft: I suppose it is less about what is in the 0th position and more about the fact that `None` changes the default behavior. Another idea would be to rename `None` to something more descriptive.

sdk/core/Azure.Core.Experimental/src/DynamicJson.cs
```diff
@@ -41,39 +41,39 @@ internal override void WriteTo(Stream stream)
...
-            }
+        private bool CamelCaseSetters()
+        {
+            return _options.PropertyNameCasing == DynamicJsonNameMapping.PascalCaseGettersCamelCaseSetters;
+        }
 
-            return $"{char.ToUpperInvariant(value[0])}{value.Substring(1)}";
+        private static string GetAsCamelCase(string value)
+        {
+            return $"{char.ToLowerInvariant(value[0])}{value.Substring(1)}";
```
christothes, line 76: We should check that the string's length is at least 2.

sdk/core/Azure.Core.Experimental/src/DynamicJson.cs
```diff
@@ -96,6 +96,37 @@ private IEnumerable GetEnumerable()
 
         private object? SetProperty(string name, object value)
         {
+            Argument.AssertNotNullOrEmpty(name, nameof(name));
+
+            if (_options.PropertyNameCasing == DynamicJsonNameMapping.None)
```
christothes, line 101:
```suggestion
            if (_options.PropertyNameCasing == DynamicJsonNameMapping.None || name.Length  == 1)
```

sdk/core/Azure.Core.Experimental/tests/DynamicJsonTests.cs
annelo-msft: That makes sense.

We're trying to mirror the STJ APIs as much as possible, but there isn't an exact mapping for this one given the different requirements of this type.  The closest other types are things like:
- https://docs.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializeroptions.propertynamingpolicy
- https://learn.microsoft.com/en-us/dotnet/api/system.text.json.nodes.jsonnodeoptions.propertynamecaseinsensitive
- https://docs.microsoft.com/en-us/dotnet/api/system.text.json.serialization.jsonnumberhandling

Interestingly, there is this one:
https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializerdefaults

That lets you change the default settings of the JsonSerializer.  If we felt strongly enough, we could add this to client options with values like `General` and `Azure`, where the default was `Azure` for libraries unless they changed it.  Although we would need to find a way to plumb that through to the ToDynamic() call, which is currently independent of any library details.

@KrzysztofCwalina, do you have any strong feelings about the defaults here?  @JoshLove-msft is right that most enums are structured with `None` or `Strict` in the zeroth position and this typically maps to the default.  Is there a BCL precedent for doing otherwise?

KrzysztofCwalina: Enum value 0 should correspond to the default, i.e. no passing in enum argument should be equivalent to passing the 0 value. But I think this is exactly what these APIs do today, don't they? i.e. if I call response.ToDynamic() and response.ToDynamic((JsonMappingOptions)0) I will get the exact same behavior, won't i?

annelo-msft: No, I had configured the default to be `PascalCaseGettersCamelCaseSetters` because Azure REST API guidance is to camelCase JSON, and .NET guidance is to PascalCase properties, and given that it seems like, for Azure, we would want that behavior in the most common case, rather than asking callers to call `response.Content.ToDynamic(DynamicJsonNameMapping.PascalCaseGettersCamelCaseSetters)` everywhere.  But I can change that if we prefer that approach - it just looks rather awkward to me.

annelo-msft: After considering it overnight, I do think that if the default doesn't pass through the casing, that would be surprising behavior for a user, e.g. I do:

```
var json = response.Content.ToDynamic();
json.Value = "PascalCaseValue";
```

and then ToString prints:

```
{
  "value" : "PascalCaseValue";
}
```

I still don't love the 57 chars that `DynamicJsonNameMapping.PascalCaseGettersCamelCaseSetters` requires, though, so I'm going to look at adding a `DynamicJsonOptions.AzureDefault` (a savings of 25 chars!), to see if that feels a little better.

sdk/core/Azure.Core.Experimental/src/DynamicJson.cs
annelo-msft: I think I disagree with this - would you not lower-case an uppercase single character if your directive was to make it camelCase?  Do we know if this is defined?  It does seem like a degenerate case...

JoshLove-msft (approved):

sdk/core/Azure.Core.Experimental/src/DynamicJson.cs
annelo-msft: Merging for now, but happy to make changes later if you feel strongly about this.

Issue #34030, opened 2023-02-09
[BUG] [AzureMonitorExporter] Null Reference Exceptions at exporting logs


pocki:
### Library name and version

Azure.Monitor.OpenTelemetry.Exporter 1.0.0-beta.7

### Describe the bug

Null Reference Exceptions somewhere in exporting logs when running ASP.NET Core 7 web app.

```
OpenTelemetry-AzureMonitor-Exporter - EventId: [2], EventName: [WriteError], Message: [FailedToExport - System.NullReferenceException: Object reference not set to an instance of an object.
   at Azure.Monitor.OpenTelemetry.Exporter.Internals.LogsHelper.ExtractProperties(String& message, IDictionary`2 properties, IReadOnlyCollection`1 stateDictionary)
   at Azure.Monitor.OpenTelemetry.Exporter.Internals.LogsHelper.GetMessageAndSetProperties(LogRecord logRecord, IDictionary`2 properties)
   at Azure.Monitor.OpenTelemetry.Exporter.Models.MessageData..ctor(Int32 version, LogRecord logRecord)
   at Azure.Monitor.OpenTelemetry.Exporter.Internals.LogsHelper.OtelToAzureMonitorLogs(Batch`1 batchLogRecord, AzureMonitorResource resource, String instrumentationKey)
   at Azure.Monitor.OpenTelemetry.Exporter.AzureMonitorLogExporter.Export(Batch`1& batch)]
```

### Expected behavior

Running WebApp without errors in logs

### Actual behavior

Null Reference Exception during running ASP.NET Core v7 web apps.

```
OpenTelemetry-AzureMonitor-Exporter - EventId: [2], EventName: [WriteError], Message: [FailedToExport - System.NullReferenceException: Object reference not set to an instance of an object.
   at Azure.Monitor.OpenTelemetry.Exporter.Internals.LogsHelper.ExtractProperties(String& message, IDictionary`2 properties, IReadOnlyCollection`1 stateDictionary)
   at Azure.Monitor.OpenTelemetry.Exporter.Internals.LogsHelper.GetMessageAndSetProperties(LogRecord logRecord, IDictionary`2 properties)
   at Azure.Monitor.OpenTelemetry.Exporter.Models.MessageData..ctor(Int32 version, LogRecord logRecord)
[...]

Issue #34013, opened 2023-02-08
[AzureMonitorExporter] enable nullable reference types


TimothyMothra:
### Library name

Azure.Monitor.OpenTelemetry.Exporter

### Please describe the feature.

This issue is to turn on compile time checking of nullables in the Azure Monitor Exporter project.
I'm opening this issue to track and organize PRs.

For this effort, we're turning on `nullable enable` for the entire project. and then setting `nullable disable` for offending files. 
This will help ensure any new files are compliant while we work to resolve other the individual files.

### Related 
- https://learn.microsoft.com/en-us/dotnet/csharp/nullable-references
- https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/attributes/nullable-analysis

PR #34092, opened 2023-02-11
[AzureMonitorExporter] fix more nullables

TimothyMothra:
Towards #34013 

This should also address #34030 

## Changes
- remove `#nullable disable` and resolve issues
  - TraceHelper and LogHelper


sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Customizations/Models/TelemetryItem.cs
```diff
@@ -73,7 +73,7 @@ internal partial class TelemetryItem
             SampleRate = telemetryItem.SampleRate;
         }
 
-        public TelemetryItem (LogRecord logRecord, AzureMonitorResource resource, string instrumentationKey) :
+        public TelemetryItem (LogRecord logRecord, AzureMonitorResource? resource, string instrumentationKey) :
```
TimothyMothra, line 76: this change is safe. we handle the null further down at line 102

vishweshbankwar (approved):
LGTM

PR #34107, opened 2023-02-13
fix(build): cadl build error when repo url does not end with `.git`

archerzz:
`[.git]` is a char set which means the url should ends with either `.`, or `g`, or `i`, or `t`. So if the url doesn't end with `.git` then there is no match which results in error.

The fix is to remove `[.git]` char set. We don't need `(.git)*` either, because `\S+` will suppress it.

### DIFF
--- a/eng/common/scripts/Cadl-Project-Sync.ps1
+++ b/eng/common/scripts/Cadl-Project-Sync.ps1
@@ -44,10 +44,10 @@ function GetGitRemoteValue([string]$repo) {
         $gitRemotes = (git remote -v)
         foreach ($remote in $gitRemotes) {
             if ($remote.StartsWith("origin")) {
-                if ($remote -match 'https://github.com/\S+[\.git]') {
+                if ($remote -match 'https://github.com/\S+') {
                     $result = "https://github.com/$repo.git"
                     break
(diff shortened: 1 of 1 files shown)


archerzz:
## Current Behavior

image


## After Fix

image

dw511214992:
I don't think you can fix it in .Net sdk repo directly. Based on my understanding, it only can be fixed in https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/scripts/Cadl-Project-Sync.ps1. Then it will be deployed to every sdk repo.

archerzz:
> I don't think you can fix it in .Net sdk repo directly. Based on my understanding, it only can be fixed in [Azure/azure-sdk-tools@`main`/eng/common/scripts/Cadl-Project-Sync.ps1](https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/scripts/Cadl-Project-Sync.ps1?rgh-link-date=2023-02-13T09%3A40%3A55Z). Then it will be deployed to every sdk repo. cc: @m-nash

Thanks @dw511214992 I didn't realize it's already in `auzre-sdk-tools`.

archerzz:
fix in https://github.com/Azure/azure-sdk-tools/pull/5427, so close this pr

PR #34122, opened 2023-02-13
Properly pass options to ManagedIdentityCredential

### LINKED ISSUE #34077: [BUG] Can not get secret from KeyVault on account of failing ManagedIdentityCredential GetToken method in Gov and China clouds

christothes:
fixes https://github.com/Azure/azure-sdk-for-net/issues/34077

### DIFF
--- a/sdk/identity/Azure.Identity/src/MsalConfidentialClient.cs
+++ b/sdk/identity/Azure.Identity/src/MsalConfidentialClient.cs
@@ -18,7 +18,6 @@ namespace Azure.Identity
         private readonly Func<string> _assertionCallback;
         private readonly Func<CancellationToken, Task<string>> _asyncAssertionCallback;
         private readonly Func<AppTokenProviderParameters, Task<AppTokenProviderResult>> _appTokenProviderCallback;
-        private readonly Uri _authority;
 
         internal string RedirectUrl { get; }
 
@@ -58,7 +57,6 @@ namespace Azure.Identity
             : base(pipeline, tenantId, clientId, options)
         {
             _appTokenProviderCallback = appTokenProviderCallback;
-            _authority = options?.AuthorityHost ?? AzureAuthorityHosts.AzurePublicCloud;
         }
 
         internal string RegionalAuthority { get; } = EnvironmentVariables.AzureRegionalAuthorityName;
@@ -74,12 +72,12 @@ namespace Azure.Identity
             if (_appTokenProviderCallback != null)
             {
                 confClientBuilder.WithAppTokenProvider(_appTokenProviderCallback)
-                    .WithAuthority(_authority.AbsoluteUri, TenantId, false)
+                    .WithAuthority(AuthorityHost.AbsoluteUri, TenantId, false)
                     .WithInstanceDiscovery(false);
             }
             else
             {
-                confClientBuilder.WithAuthority(Pipeline.AuthorityHost.AbsoluteUri, TenantId);
+                confClientBuilder.WithAuthority(AuthorityHost.AbsoluteUri, TenantId);
                 if (DisableInstanceDiscovery)
                 {
                     confClientBuilder.WithInstanceDiscovery(false);
@@ -146,7 +144,7 @@ namespace Azure.Identity
 
             if (!string.IsNullOrEmpty(tenantId))
             {
-                builder.WithAuthority(Pipeline.AuthorityHost.AbsoluteUri, tenantId);
+                builder.WithAuthority(AuthorityHost.AbsoluteUri, tenantId);
             }
             return await builder
                 .ExecuteAsync(async, cancellationToken)
@@ -179,7 +177,7 @@ namespace Azure.Identity
             var builder = client.AcquireTokenSilent(scopes, account);
             if (!string.IsNullOrEmpty(tenantId))
             {
-                builder.WithAuthority(Pipeline.AuthorityHost.AbsoluteUri, tenantId);
+                builder.WithAuthority(AuthorityHost.AbsoluteUri, tenantId);
             }
             return await builder
                 .ExecuteAsync(async, cancellationToken)
@@ -213,7 +211,7 @@ namespace Azure.Identity
 
             if (!string.IsNullOrEmpty(tenantId))
             {
-                builder.WithAuthority(Pipeline.AuthorityHost.AbsoluteUri, tenantId);
+                builder.WithAuthority(AuthorityHost.AbsoluteUri, tenantId);
             }
             return await builder
                 .ExecuteAsync(async, cancellationToken)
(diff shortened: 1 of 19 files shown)


sdk/identity/Azure.Identity/src/MsalConfidentialClient.cs
```diff
@@ -146,7 +144,7 @@ protected override async ValueTask<IConfidentialClientApplication> CreateClientA
 
             if (!string.IsNullOrEmpty(tenantId))
             {
-                builder.WithAuthority(Pipeline.AuthorityHost.AbsoluteUri, tenantId);
+                builder.WithAuthority(AuthorityHost.AbsoluteUri, tenantId);
```
pharring, line 147: >WithAuthority [](http://example.com/codeflow?start=24&length=13)

Could we use WithTenant here instead of WithAuthority? 

We don't intend or need to change the authority host -- that was set up when the ConfidentialClientApp was built.

sdk/identity/Azure.Identity/src/MsalPublicClient.cs
```diff
@@ -119,7 +118,7 @@ protected virtual async ValueTask<AuthenticationResult> AcquireTokenSilentCoreAs
             // otherwise we should authenticate with the tenant specified by the authentication record since that's the tenant the
             // user authenticated to originally.
             return await client.AcquireTokenSilent(scopes, (AuthenticationAccount)record)
-                .WithAuthority(Pipeline.AuthorityHost.AbsoluteUri, TenantId ?? record.TenantId)
+                .WithAuthority(AuthorityHost.AbsoluteUri, TenantId ?? record.TenantId)
```
pharring, line 121: >WithAuthority [](http://example.com/codeflow?start=17&length=13)

Can we use WithTenantId here?

sdk/identity/Azure.Identity/tests/ManagedIdentityCredentialTests.cs
```diff
@@ -326,7 +317,7 @@ public async Task VerifyIMDSRequestWithPodIdentityEnvVarResourceIdMockAsync()
             var options = new TokenCredentialOptions() { Transport = mockTransport };
             var pipeline = CredentialPipeline.GetInstance(options);
```
pharring, line 318: >var pipeline = CredentialPipeline.GetInstance(options); [](http://example.com/codeflow?start=12&length=55)

Unused, delete?

sdk/identity/Azure.Identity/tests/ManagedIdentityCredentialTests.cs
```diff
@@ -169,7 +160,7 @@ public async Task VerifyImdsRequestWithResourceIdMockAsync()
             var options = new TokenCredentialOptions { Transport = mockTransport };
             var pipeline = CredentialPipeline.GetInstance(options);
```
pharring, line 161: >var pipeline = CredentialPipeline.GetInstance(options); [](http://example.com/codeflow?start=12&length=55)

Unused, delete?

sdk/identity/Azure.Identity/src/MsalConfidentialClient.cs
```diff
@@ -179,7 +177,7 @@ protected override async ValueTask<IConfidentialClientApplication> CreateClientA
             var builder = client.AcquireTokenSilent(scopes, account);
             if (!string.IsNullOrEmpty(tenantId))
             {
-                builder.WithAuthority(Pipeline.AuthorityHost.AbsoluteUri, tenantId);
+                builder.WithAuthority(AuthorityHost.AbsoluteUri, tenantId);
```
pharring, line 180: >WithAuthority [](http://example.com/codeflow?start=24&length=13)

Similarly, GetTenantId here? (and 2 further instances)

sdk/identity/Azure.Identity/src/MsalConfidentialClient.cs
christothes: I believe there was some edge case scenario where this indirect approach to setting tenantId was what we implemented. I'd rather not change it as part of this PR to avoid possible regressions.

schaabs (approved):

PR #34126, opened 2023-02-13
MutableJsonDocument namespace, IDisposable

annelo-msft:
Addresses items in https://github.com/Azure/azure-sdk-for-net/issues/33856

- Move MutableJsonDocument and related types out of Azure.Core.Dynamic into Azure.Core.Json
- Make MutableJsonDocument and DynamicJson disposable so they can dispose the inner JsonDocument

### DIFF
--- a/sdk/core/Azure.Core.Experimental/src/DynamicJson.cs
+++ b/sdk/core/Azure.Core.Experimental/src/DynamicJson.cs
@@ -15,7 +15,7 @@ namespace Azure.Core.Dynamic
     /// Dynamic layer over MutableJsonDocument.
     /// </summary>
     [JsonConverter(typeof(JsonConverter))]
-    public partial class DynamicJson : DynamicData
+    public sealed partial class DynamicJson : DynamicData, IDisposable
     {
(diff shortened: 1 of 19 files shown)


sdk/core/Azure.Core.Experimental/src/DynamicJson.cs
```diff
@@ -103,6 +104,12 @@ public override string ToString()
             return _element.ToString();
         }
 
+        /// <inheritdoc/>
+        public void Dispose()
+        {
+            _element.DisposeRoot();
```
annelo-msft, line 110: There is a challenge here that DynamicJson wraps a MutableJsonElement that may or may not represent the root of the document.  If disposed, it will dispose the full document and not just that element.

We can recommend that users only call Dispose() on a DynamicJson representing the top-level document (e.g. one returned from ToDynamic()), or just make sure it is well-documented that disposing any element in the document disposes the full document.

KrzysztofCwalina (approved):

sdk/core/Azure.Core.Experimental/src/DynamicJson.cs
christothes: Should we add a test that demonstrates this dispose behavior?

christothes (approved):

PR #34133, opened 2023-02-14
[AzureMonitorExporter] fix nullables in Tests

TimothyMothra:
Towards #34013

This PR fixes about half of our tests.

## Changes
- update `xunit` to v2.4.2.
  This version adds NullableAttributes to methods like `Assert.NotNull()`
- update `xunit.runner.visualstudio` to v2.4.2
- remove `#disable nullable` in tests and fix issues.

### DIFF
--- a/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/E2ETelemetryItemValidation/TracesTests.cs
+++ b/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/E2ETelemetryItemValidation/TracesTests.cs
@@ -250,12 +250,12 @@ namespace Azure.Monitor.OpenTelemetry.Exporter.Tests.E2ETelemetryItemValidation
                 expectedSpanId: spanId,
                 expectedProperties: null);
 
-            Assert.True(logTelemetryItems.Any(), "Unit test failed to collect telemetry.");
+            Assert.True(logTelemetryItems?.Any(), "Unit test failed to collect telemetry.");
             this.telemetryOutput.Write(logTelemetryItems);
-            var logTelemetryItem = logTelemetryItems.Single();
+            var logTelemetryItem = logTelemetryItems?.Single();
 
             TelemetryItemValidationHelper.AssertMessageTelemetry(
-                telemetryItem: logTelemetryItem,
+                telemetryItem: logTelemetryItem!,
                 expectedSeverityLevel: expectedSeverityLevel,
                 expectedMessage: "Hello {name}.",
                 expectedMessageProperties: new Dictionary<string, string> { { "name", "World" } },
--- a/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/E2ETelemetryItemValidation/TracesTests.cs
+++ b/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/E2ETelemetryItemValidation/TracesTests.cs
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+#nullable disable // TODO: remove and fix errors
+
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
@@ -49,7 +51,7 @@ namespace Azure.Monitor.OpenTelemetry.Exporter.Tests.E2ETelemetryItemValidation
 
             // ACT
             string spanId, traceId;
-            using (var activity = activitySource.StartActivity(name: "SayHello", kind: activityKind) ?? throw new Exception("Failed to create Activity"))
+            using (var activity = activitySource.StartActivity(name: "SayHello", kind: activityKind ))
             {
                 traceId = activity.TraceId.ToHexString();
                 spanId = activity.SpanId.ToHexString();
@@ -61,7 +63,7 @@ namespace Azure.Monitor.OpenTelemetry.Exporter.Tests.E2ETelemetryItemValidation
             }
 
             // CLEANUP
-            tracerProvider?.Dispose();
+            tracerProvider.Dispose();
 
             // ASSERT
             Assert.True(telemetryItems.Any(), "Unit test failed to collect telemetry.");
@@ -94,7 +96,7 @@ namespace Azure.Monitor.OpenTelemetry.Exporter.Tests.E2ETelemetryItemValidation
 
             // ACT
             string spanId, traceId;
-            using (var activity = activitySource.StartActivity(name: "SayHello", kind: activityKind) ?? throw new Exception("Failed to create Activity"))
+            using (var activity = activitySource.StartActivity(name: "SayHello", kind: activityKind))
             {
                 traceId = activity.TraceId.ToHexString();
                 spanId = activity.SpanId.ToHexString();
@@ -106,7 +108,7 @@ namespace Azure.Monitor.OpenTelemetry.Exporter.Tests.E2ETelemetryItemValidation
             }
 
             // CLEANUP
-            tracerProvider?.Dispose();
+            tracerProvider.Dispose();
 
             // ASSERT
             Assert.True(telemetryItems.Any(), "Unit test failed to collect telemetry.");
@@ -137,9 +139,9 @@ namespace Azure.Monitor.OpenTelemetry.Exporter.Tests.E2ETelemetryItemValidation
                 .Build();
 
             // ACT
-            string spanId, traceId;
+            string spanId = null, traceId = null;
 
-            using (var activity = activitySource.StartActivity(name: "ActivityWithException") ?? throw new Exception("Failed to create Activity"))
+            using (var activity = activitySource.StartActivity(name: "ActivityWithException"))
             {
                 traceId = activity.TraceId.ToHexString();
                 spanId = activity.SpanId.ToHexString();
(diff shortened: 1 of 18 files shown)


sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/E2ETelemetryItemValidation/TracesTests.cs
```diff
@@ -280,14 +278,14 @@ public void TestActivityEvents()
                 .Build();
 
             // ACT
-            string spanId = null, traceId = null;
+            string spanId, traceId;
 
-            using (var activity = activitySource.StartActivity(name: "ActivityWithException"))
+            using (var activity = activitySource.StartActivity(name: "ActivityWithException") ?? throw new Exception("Failed to create Activity"))
```
rajkumar-rangaraj, line 283: I see this in many places, why do we need to throw when activity is null. We don't have any tests to validate the scenario. In general activity could be null if it does not have an Listener. These tests might give a different understanding about activities for customer.

TimothyMothra: `StartActivity()` can return null if nothing is subscribed to the ActivitySource.
This was the easiest way to deal with the null and satisfy the compiler.
Here the exception is valid... if an activity isn't created, the test will fail.

The alternative is to decorate every interaction with the activity with null-conditional `?.` and null-forgiving `!` operators. 


These tests could possibly be rewritten to refactor this as boilerplate code.

vishweshbankwar: Not a strong preference but I would vote for `?`. IIRC visual studio will give suggestion to add `?` if someone missed that. Does it give suggestion to add throw as well?

TimothyMothra: Visual Studio did recommend these at first, but it was A LOT of `?.` and `!` decorations.
There was also some inconsistency which made the rest of the method look odd. 
Any direct member (property or method) required the null-conditional `?.`. But some extension methods do their own null check so the null-conditional wasn't needed.
And anywhere we pass the activity to a method required either the null-forgiving `!` operator or some additional null check.

I felt that this was the simplest option to address all cases.

TimothyMothra: Discussed offline. 
Will remove the "throw" from this PR and will review options in a follow up.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/TraceHelperTests.cs
```diff
@@ -226,8 +229,9 @@ public void ActivityWithExceptionEventCreatesExceptionTelemetry()
             var exceptionMessage = "Exception Message";
             using ActivitySource activitySource = new ActivitySource(ActivitySourceName);
             using var activity = activitySource.StartActivity(
-               ActivityName,
-               ActivityKind.Server);
+                ActivityName,
```
vishweshbankwar, line 232: leftover from clean up?

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/TraceHelperTests.cs
```diff
@@ -241,9 +245,11 @@ public void ActivityWithExceptionEventCreatesExceptionTelemetry()
             Assert.Equal(2, telemetryItems.Count());
             Assert.Equal("Exception", telemetryItems[0].Name);
             Assert.Equal("Request", telemetryItems[1].Name);
-            Assert.Equal(exceptionMessage, (telemetryItems[0].Data.BaseData as TelemetryExceptionData).Exceptions.First().Message);
-            Assert.Equal("System.Exception", (telemetryItems[0].Data.BaseData as TelemetryExceptionData).Exceptions.First().TypeName);
-            Assert.Equal("System.Exception: Exception Message", (telemetryItems[0].Data.BaseData as TelemetryExceptionData).Exceptions.First().Stack);
+
+            var telemetryExceptionData = (telemetryItems[0].Data.BaseData as TelemetryExceptionData) ?? throw new Exception("Invalid BaseData");
```
vishweshbankwar, line 249: is it possible to do Assert.NotNull here?

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/TraceHelperTests.cs
```diff
@@ -252,8 +258,9 @@ public void ActivityWithEventCreatesTraceTelemetry()
             var eventName = "Custom Event";
             using ActivitySource activitySource = new ActivitySource(ActivitySourceName);
             using var activity = activitySource.StartActivity(
-               ActivityName,
-               ActivityKind.Server);
+                ActivityName,
+                ActivityKind.Server)
+                ?? throw new Exception("Failed to create Activity");
```
vishweshbankwar, line 263: same as https://github.com/Azure/azure-sdk-for-net/pull/34133/files#r1109171054?

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/TraceHelperTests.cs
```diff
@@ -274,19 +281,22 @@ public void ActivityWithEventCreatesTraceTelemetry()
             Assert.Equal(2, telemetryItems.Count());
             Assert.Equal("Message", telemetryItems[0].Name);
             Assert.Equal("Request", telemetryItems[1].Name);
-            Assert.Equal(eventName, (telemetryItems[0].Data.BaseData as MessageData).Message);
-            Assert.True((telemetryItems[0].Data.BaseData as MessageData).Properties.TryGetValue("key1", out var value));
+
+            var messageData = (telemetryItems[0].Data.BaseData as MessageData) ?? throw new Exception("Invalid BaseData");
```
vishweshbankwar, line 285: same as https://github.com/Azure/azure-sdk-for-net/pull/34133/files#r1109171963

vishweshbankwar (approved):
LGTM with some minor comments.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/TraceHelperTests.cs
TimothyMothra: was just an odd whitespace. i fixed it here since i was already working in the class.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/TraceHelperTests.cs
TimothyMothra: THE LATEST VERSION OF XUNIT HAS THE CORRECT NULLABLE ATTRIBUTES!!! 🎉
I'll update the PR, this is going to fix so many issues!

TimothyMothra: changed to use Assert.NotNull

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/TraceHelperTests.cs
TimothyMothra: changed to use Assert.NotNull

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/TraceHelperTests.cs
TimothyMothra: changed to use Assert.NotNull

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/E2ETelemetryItemValidation/TracesTests.cs
TimothyMothra: After updating xunit, I was able to fix this with `Assert.NotNull`.
The newest version adds NullableAttributes to their implementation.

PR #34152, opened 2023-02-14
Link to additional resources

udidahan:


jsquire:
Hi Udi.  Thanks very much for your contribution and apologies for the delay in following up.   Unfortunately, we're unable to move forward with this pull request.

While we appreciate that the spirit of the proposed changes is helping developers find success using Azure Service Bus, it is difficult to include pointers to NServiceBus without it being perceived as an implicit endorsement over similar third-party offerings such as Mass Transit.   After internal discussion, a [guideline for including third-party dependencies in Azure SDK samples](https://github.com/Azure/azure-sdk-for-net/blob/main/CONTRIBUTING.md#samples) was established and limits doing so to scenarios focused on demonstrating how to use the Azure SDK library.  

Microsoft recognizes the value that NServiceBus provides to Azure customers and would be happy to consider an NServiceBus-based outbox pattern example on [learn.microsoft.com](https://review.learn.microsoft.com/en-us/help/contribute/contribute-get-started-channel-guidance?branch=main#non-microsoft-products) directly or as part of the existing [NServiceBus article](https://learn.microsoft.com/en-us/azure/service-bus-messaging/build-message-driven-apps-nservicebus?tabs=Sender).

Thank you, again, for your contribution.

udidahan:
@jsquire I'd be happy to include the link to MassTransit (as I have done many times in the past). Just for background, MassTransit has recently changed a bunch of things including their docs so I'm still getting up to speed on them.

Seeing as this PR is closed, I'll open a new one that includes the link to the comparable MassTransit feature.

Thanks.

Issue #33131, opened 2022-12-20
[BUG] Clients registered with `AddAzureClients` are not disposed


tom-wolfe:
### Library name and version

Microsoft.Extensions.Azure 1.6.0

### Describe the bug

Not 100% if bug or intended behaviour, but could do with a documentation clarification if not.

When using the Azure Client Builder/Factory, the registered clients aren't disposed along with the `IServiceProvider`.

### Expected behavior

When calling `provider.Dispose()` in the provided code snippet, I would expect the resolved client instance to be disposed, because the [Dependency Injection guidelines](https://learn.microsoft.com/en-us/dotnet/core/extensions/dependency-injection-guidelines#disposal-of-services) specify not to dispose of services resolved from the container manually.

### Actual behavior

The resolved client is left undisposed because the object that actually gets registered with the `IServiceCollection` is the `ClientRegistration` object, which doesn't implement `IDisposable`.

### Reproduction Steps

```cs
using Microsoft.Extensions.Azure;
using Microsoft.Extensions.DependencyInjection;

public class Program
{
    public static void Main()
    {
        var services = new ServiceCollection();
        services.AddAzureClients(builder => builder
            .AddClient<MyDisposableClient, object>(_ => new MyDisposableClient()).WithName("TestClient")
        );
        var provider = services.BuildServiceProvider();
        
        var factory = provider.GetRequiredService<IAzureClientFactory<MyDisposableClient>>();
        
        var client = factory.CreateClient("TestClient");
        client.Write("Client created successfully.");

        provider.Dispose(); // <-- Expected dispose message to be printed at this point.
        Console.WriteLine("Provider disposed");

        Console.ReadLine();
    }

    class MyDisposableClient : IDisposable
    {
        public void Write(string message) => Console.WriteLine(message);
        public void Dispose() => Write("Client disposed");
    }
}
```

### Environment

IDE: Visual Studio Professional 2022 (64-bit) (17.4.2)
```
.NET SDK:
 Version:   7.0.101
 Commit:    bb24aafa11

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.22621
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\7.0.101\

Host:
  Version:      7.0.1
  Architecture: x64
  Commit:       97203d38ba
```

jsquire:
[...]

PR #34154, opened 2023-02-14
[Extensions] AzureClientFactory disposable support

jsquire:
# Summary

The focus of these changes is to add support for clients to be disposed via `IDisposable` or `IAsyncDisposable` when the service factory is disposed.

# References and Related

- [[BUG] Clients registered with AddAzureClients are not disposed (#33131)](https://github.com/Azure/azure-sdk-for-net/issues/33131)

### DIFF
--- a/sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientRegistration.cs
+++ b/sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientRegistration.cs
@@ -20,6 +20,7 @@ namespace Microsoft.Extensions.Azure
         private readonly bool _asyncDisposable;
         private readonly bool _disposable;
 
+        private bool _clientInitialized;
         private TClient _cachedClient;
         private ExceptionDispatchInfo _cachedException;
 
@@ -36,7 +37,7 @@ namespace Microsoft.Extensions.Azure
         {
             _cachedException?.Throw();
 
-            if (_cachedClient != null)
+            if (_clientInitialized)
             {
                 return _cachedClient;
             }
@@ -45,7 +46,7 @@ namespace Microsoft.Extensions.Azure
             {
                 _cachedException?.Throw();
 
-                if (_cachedClient != null)
+                if (_clientInitialized)
                 {
                     return _cachedClient;
                 }
@@ -58,6 +59,7 @@ namespace Microsoft.Extensions.Azure
                 try
                 {
                     _cachedClient = _factory(serviceProvider, options, tokenCredential);
+                    _clientInitialized = true;
                 }
                 catch (Exception e)
                 {
@@ -71,7 +73,7 @@ namespace Microsoft.Extensions.Azure
 
         public async ValueTask DisposeAsync()
         {
-            if (_cachedClient != null)
+            if (_clientInitialized)
             {
                 if (_asyncDisposable)
                 {
@@ -79,13 +81,15 @@ namespace Microsoft.Extensions.Azure
 
                     lock (_cacheLock)
                     {
-                       if (_cachedClient == null)
+                       if (!_clientInitialized)
                        {
                            return;
                        }
 
                        disposableClient = (IAsyncDisposable)_cachedClient;
+
                        _cachedClient = default;
+                       _clientInitialized = false;
                     }
 
                     await disposableClient.DisposeAsync().ConfigureAwait(false);
@@ -99,7 +103,7 @@ namespace Microsoft.Extensions.Azure
 
         public void Dispose()
         {
-            if (_cachedClient != null)
+            if (_clientInitialized)
             {
                 if (_disposable)
                 {
@@ -107,13 +111,15 @@ namespace Microsoft.Extensions.Azure
 
                     lock (_cacheLock)
                     {
-                       if (_cachedClient == null)
+                       if (!_clientInitialized)
                        {
                            return;
                        }
 
                        disposableClient = (IDisposable)_cachedClient;
+
                        _cachedClient = default;
+                       _clientInitialized = false;
                     }
 
                     disposableClient.Dispose();
--- a/sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientRegistration.cs
+++ b/sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientRegistration.cs
@@ -3,11 +3,12 @@
 
 using System;
 using System.Runtime.ExceptionServices;
+using System.Threading.Tasks;
 using Azure.Core;
 
 namespace Microsoft.Extensions.Azure
 {
-    internal class ClientRegistration<TClient>
+    internal class ClientRegistration<TClient> : IDisposable, IAsyncDisposable
     {
         public string Name { get; set; }
         public object Version { get; set; }
@@ -16,15 +17,19 @@ namespace Microsoft.Extensions.Azure
         private readonly Func<IServiceProvider, object, TokenCredential, TClient> _factory;
 
         private readonly object _cacheLock = new object();
+        private readonly bool _asyncDisposable;
+        private readonly bool _disposable;
 
         private TClient _cachedClient;
-
         private ExceptionDispatchInfo _cachedException;
 
         public ClientRegistration(string name, Func<IServiceProvider, object, TokenCredential, TClient> factory)
         {
             Name = name;
             _factory = factory;
+
+            _asyncDisposable = typeof(IAsyncDisposable).IsAssignableFrom(typeof(TClient));
+            _disposable = typeof(IDisposable).IsAssignableFrom(typeof(TClient));
         }
 
         public TClient GetClient(IServiceProvider serviceProvider, object options, TokenCredential tokenCredential)
@@ -61,6 +66,62 @@ namespace Microsoft.Extensions.Azure
                 }
 
                 return _cachedClient;
+            }
+        }
+
+        public async ValueTask DisposeAsync()
+        {
+            if (_cachedClient != null)
+            {
+                if (_asyncDisposable)
+                {
+                    IAsyncDisposable disposableClient;
+
+                    lock (_cacheLock)
+                    {
+                       if (_cachedClient == null)
+                       {
+                           return;
+                       }
+
+                       disposableClient = (IAsyncDisposable)_cachedClient;
+                       _cachedClient = default;
+                    }
+
+                    await disposableClient.DisposeAsync().ConfigureAwait(false);
+                }
+                else if (_disposable)
+                {
+                    Dispose();
+                }
+            }
+        }
+
+        public void Dispose()
+        {
+            if (_cachedClient != null)
+            {
+                if (_disposable)
+                {
+                    IDisposable disposableClient;
+
+                    lock (_cacheLock)
+                    {
+                       if (_cachedClient == null)
+                       {
+                           return;
+                       }
+
+                       disposableClient = (IDisposable)_cachedClient;
+                       _cachedClient = default;
+                    }
+
+                    disposableClient.Dispose();
+                }
+                else if (_asyncDisposable)
+                {
+                    DisposeAsync().GetAwaiter().GetResult();
+                }
             }
         }
     }
--- a/sdk/extensions/Microsoft.Extensions.Azure/tests/AzureClientFactoryTests.cs
+++ b/sdk/extensions/Microsoft.Extensions.Azure/tests/AzureClientFactoryTests.cs
@@ -522,7 +522,7 @@ namespace Azure.Core.Extensions.Tests
         }
 
         [Test]
-        public void DisposableWorksWithMixedClients()
+        public void DisposableHandlesMixedClients()
         {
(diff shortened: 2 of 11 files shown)


sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientRegistration.cs
```diff
@@ -63,5 +68,61 @@ public TClient GetClient(IServiceProvider serviceProvider, object options, Token
...
+
+                    lock (_cacheLock)
+                    {
+                       if (_cachedClient == null)
+                       {
+                           return;
+                       }
+
+                       disposableClient = (IDisposable)_cachedClient;
+                       _cachedClient = default;
```
jsquire, line 116: This one is weird... the compiler doesn't like a `null` assignment here because there's no constraint ensuring `TClient` is a reference type.   But above, it does not like a comparison to `default` for the same reason.

sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientRegistration.cs
```diff
@@ -3,11 +3,12 @@
 
 using System;
 using System.Runtime.ExceptionServices;
+using System.Threading.Tasks;
 using Azure.Core;
 
 namespace Microsoft.Extensions.Azure
 {
-    internal class ClientRegistration<TClient>
+    internal class ClientRegistration<TClient> : IDisposable, IAsyncDisposable
```
jsquire, line 11: I'd really like to add a `class` constraint to the generic here to enforce a reference type, but I believe that would be a breaking change.    I'm afraid that the factory already doesn't work with value types, however, as L48 would never evaluate to `true`...  am I overlooking something?

jsquire: Confirming, this does not work with a `struct` or `int` due to the factory assuming `null` for no value.

m-redding: To me that seems like a justifiable breaking change since presumably no customers would be able to pass anything in besides a class - so this would already be indirectly enforced. Unless there's some kind of weird case with a different ref type that this could break? To me this seems like an improvement to usability since it will make it faster for customers to see errors.

sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientRegistration.cs
```diff
@@ -16,15 +17,19 @@ internal class ClientRegistration<TClient>
...
         private TClient _cachedClient;
-
         private ExceptionDispatchInfo _cachedException;
 
         public ClientRegistration(string name, Func<IServiceProvider, object, TokenCredential, TClient> factory)
         {
             Name = name;
             _factory = factory;
+
+            _asyncDisposable = typeof(IAsyncDisposable).IsAssignableFrom(typeof(TClient));
```
m-redding, line 31: I've never seen IsAssignableFrom before, cool method

sdk/extensions/Microsoft.Extensions.Azure/tests/AzureClientFactoryTests.cs
```diff
@@ -417,9 +420,152 @@ public void CanRegisterCustomClientWithOptionsAndCredentialProvider()
...
+
+            var provider = services.BuildServiceProvider();
+            var factory = provider.GetRequiredService<IAzureClientFactory<NonDisposableClient>>();
+            var client = factory.CreateClient(nameof(NonDisposableClient));
+
+            Assert.DoesNotThrow(provider.Dispose);
+        }
+
+        [Test]
+        public void DisposibleWorksWithMixedClients()
```
m-redding, line 495: nit
```suggestion
        public void DisposableWorksWithMixedClients()
```

sdk/extensions/Microsoft.Extensions.Azure/tests/AzureClientFactoryTests.cs
```diff
@@ -417,9 +420,152 @@ public void CanRegisterCustomClientWithOptionsAndCredentialProvider()
...
+            services.AddAzureClients(builder =>
+                builder.AddClient<DisposableClient, object>(_ => new DisposableClient(disposeCallback)).WithName(nameof(DisposableClient)));
+
+            var provider = services.BuildServiceProvider();
+
+            var factory = provider.GetRequiredService<IAzureClientFactory<NonDisposableClient>>();
+            var client = factory.CreateClient(nameof(NonDisposableClient));
+
+            var disposableFactory = provider.GetRequiredService<IAzureClientFactory<DisposableClient>>();
+            var dispsableClient = disposableFactory.CreateClient(nameof(DisposableClient));
```
m-redding, line 524: nit
```suggestion
            var disposableClient = disposableFactory.CreateClient(nameof(DisposableClient));
```

christothes (approved):

sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientRegistration.cs
jsquire: I think I'm going to treat that as a potential second effort and open discussion with the architects.   It doesn't directly impact the support for disposable, which was the main goal here.

sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientRegistration.cs
```diff
@@ -63,5 +70,65 @@ public TClient GetClient(IServiceProvider serviceProvider, object options, Token
...
+                    Dispose();
+                }
+            }
+        }
+
+        public void Dispose()
+        {
+            if (_clientInitialized)
+            {
+                if (_disposable)
```
JoshLove-msft, line 108: How could Dispose be called if the registration is not IDisposable?

sdk/extensions/Microsoft.Extensions.Azure/tests/AzureClientFactoryTests.cs
```diff
@@ -417,9 +450,158 @@ public void CanRegisterCustomClientWithOptionsAndCredentialProvider()
...
+            Assert.DoesNotThrowAsync(async () => await tcs.Task.AwaitWithCancellation(cancellationSource.Token));
+            Assert.AreEqual(disposeCount, 2);
+        }
+
         private IConfiguration GetConfiguration(params KeyValuePair<string, string>[] items)
         {
             return new ConfigurationBuilder().AddInMemoryCollection(items).Build();
         }
+
+        private class DisposableClient : IDisposable
```
JoshLove-msft, line 572: Since we have logic around types that are both IDisposable and IAsyncDisposable, should we have a test case that covers this?

sdk/extensions/Microsoft.Extensions.Azure/tests/AzureClientFactoryTests.cs
```diff
@@ -417,9 +450,158 @@ public void CanRegisterCustomClientWithOptionsAndCredentialProvider()
...
         }
 
+        [Test]
+        public void DisposeHandlesDisposableClient()
+        {
+            var tcs = new TaskCompletionSource<bool>(TaskCreationOptions.RunContinuationsAsynchronously);
+            var services = new ServiceCollection();
+            var disposed = false;
+
+            Action disposeCallback= () =>
```
JoshLove-msft, line 460:
```suggestion
            Action disposeCallback = () =>
```

sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientRegistration.cs
```diff
@@ -63,5 +70,65 @@ public TClient GetClient(IServiceProvider serviceProvider, object options, Token
...
+
+                       _cachedClient = default;
+                       _clientInitialized = false;
+                    }
+
+                    disposableClient.Dispose();
+                }
+                else if (_asyncDisposable)
+                {
+                    DisposeAsync().GetAwaiter().GetResult();
```
JoshLove-msft, line 129: I'm not sure if it is correct to call DisposeAsync from Dispose - https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync#implement-both-dispose-and-async-dispose-patterns

sdk/extensions/Microsoft.Extensions.Azure/tests/AzureClientFactoryTests.cs
jsquire: The version of the DI package that we reference has a service factory that only supports Dispose, while the newer versions support both.   I could artificially test by retrieving each factory that we registered and manually invoking DisposeAsync, but I didn't feel that it offered enough value to justify the effort.

sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientRegistration.cs
jsquire: The case that I'm shooting for here is "we only implement DisposeAsync and not Dispose, but we still want to be cleaned up."    This isn't a problem in newer versions of the .NET DI library, where the service factory implements both, but in earlier versions - like the one we reference - they only support Dispose.    This would leave clients like Event Hubs or Service Bus unable to be cleaned up since we don't implement Dispose.

sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientRegistration.cs
jsquire: The registration is always disposable, but the `TClient` that it holds may not be.

JoshLove-msft (approved):

sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientRegistration.cs
jsquire: discussed offline

sdk/extensions/Microsoft.Extensions.Azure/tests/AzureClientFactoryTests.cs
jsquire: Bumped the dependency to allow access to DisposeAsync and mirrored tests

sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientRegistration.cs
jsquire: Discussed offline; decided to keep the implementation.

sdk/extensions/Microsoft.Extensions.Azure/tests/AzureClientFactoryTests.cs
```diff
@@ -417,9 +450,302 @@ public void CanRegisterCustomClientWithOptionsAndCredentialProvider()
...
+            var provider = services.BuildServiceProvider();
+
+            var factory = provider.GetRequiredService<IAzureClientFactory<NonDisposableClient>>();
+            var client = factory.CreateClient(nameof(NonDisposableClient));
+
+            var disposableFactory = provider.GetRequiredService<IAzureClientFactory<DisposableClient>>();
+            var disposableClient = disposableFactory.CreateClient(nameof(DisposableClient));
+
+            var asyncDisposableFactory = provider.GetRequiredService<IAzureClientFactory<AsyncDisposableClient>>();
+            var asyncDispsableClient = asyncDisposableFactory.CreateClient(nameof(AsyncDisposableClient));
```
pallavit, line 560: typo 'asyncDispsableClient'  should be 'asyncDisposableClient' instead.

sdk/extensions/Microsoft.Extensions.Azure/tests/AzureClientFactoryTests.cs
```diff
@@ -417,9 +450,302 @@ public void CanRegisterCustomClientWithOptionsAndCredentialProvider()
...
+
+            var bothDisposableFactory = provider.GetRequiredService<IAzureClientFactory<BothDisposableClient>>();
+            var bothDispsableClient = bothDisposableFactory.CreateClient(nameof(BothDisposableClient));
```
pallavit, line 563: same below.

sdk/extensions/Microsoft.Extensions.Azure/tests/AzureClientFactoryTests.cs
pallavit: Curious why cspell did not check it.

sdk/extensions/Microsoft.Extensions.Azure/tests/AzureClientFactoryTests.cs
jsquire: at least my copy/paste is consistent.   Wrong, but consistently wrong!

sdk/extensions/Microsoft.Extensions.Azure/tests/AzureClientFactoryTests.cs
jsquire: I don't think cspell checks variable names, as even with the correct spelling it's still not an actual word.

PR #34194, opened 2023-02-15
Adding TokenCredentialOptions.Clone for properly flowing options

schaabs:


sdk/identity/Azure.Identity/src/Credentials/TokenCredentialOptions.cs
```diff
@@ -41,10 +41,83 @@ public Uri AuthorityHost
...
+        private static void CloneIfImplemented<T>(TokenCredentialOptions original, TokenCredentialOptions clone, Action<T, T> cloneOperation)
+            where T : class
+        {
+            T originalAsT = (original as T);
+            T cloneAsT = (clone as T);
+
+            if (originalAsT != null && cloneAsT != null)
+            {
+                cloneOperation(originalAsT, cloneAsT);
+            }
```
christothes, line 119:
```suggestion
            if (original is T originalAsT && clone is T cloneAsT)
            {
                cloneOperation(originalAsT, cloneAsT);
            }
```

sdk/identity/Azure.Identity/src/EnvironmentVariables.cs
```diff
@@ -10,32 +10,42 @@ namespace Azure.Identity
 {
     internal class EnvironmentVariables
     {
-        public static string Username => Environment.GetEnvironmentVariable("AZURE_USERNAME");
-        public static string Password => Environment.GetEnvironmentVariable("AZURE_PASSWORD");
-        public static string TenantId => Environment.GetEnvironmentVariable("AZURE_TENANT_ID");
+        public static string Username => GetNonEmptyStringOrNull(Environment.GetEnvironmentVariable("AZURE_USERNAME"));
+        public static string Password => GetNonEmptyStringOrNull(Environment.GetEnvironmentVariable("AZURE_PASSWORD"));
+        public static string TenantId => GetNonEmptyStringOrNull(Environment.GetEnvironmentVariable("AZURE_TENANT_ID"));
         public static List<string> AdditionallyAllowedTenants => (Environment.GetEnvironmentVariable("AZURE_ADDITIONALLY_ALLOWED_TENANTS") ?? string.Empty).Split(new char[] { ';' }, StringSplitOptions.RemoveEmptyEntries).ToList();
```
christothes, line 16: Should this also use `GetNonEmptyStringOrNull`

sdk/identity/Azure.Identity/tests/SharedTokenCacheCredentialTests.cs
```diff
@@ -33,7 +33,6 @@ public override TokenCredential GetTokenCredential(CommonCredentialTestConfig co
             var options = new SharedTokenCacheCredentialOptions(tokenCacheOptions)
             {
                 Transport = config.Transport,
-                AdditionallyAllowedTenantsCore = config.AdditionallyAllowedTenants,
```
christothes, line 36: Why was this removed?

sdk/identity/Azure.Identity/tests/TokenCredentialOptionsTests.cs
```diff
@@ -101,5 +110,171 @@ public void EnviornmentSpecifiedNonHttpsAuthorityHostFails()
                 Assert.Throws<ArgumentException>(() => new UsernamePasswordCredential(tenantId, clientId, username, password, new TokenCredentialOptions()));
             }
         }
+
+        [Test]
+        public void VerifyClone([Values]SupportsInterfaces originalSupportsInterfaces, [Values]SupportsInterfaces cloneSupportsInterfaces, [Values]bool setTransport)
```
christothes, line 115: I think we should consider implementing this test using all the real credential types and using reflection to evaluate all implemented interfaces. The reason being that, as is, this test would continue to pass if we were to add a new `ISupports*` interface and forget to add the logic for it to the Clone method.

sdk/identity/Azure.Identity/src/Credentials/DefaultAzureCredentialOptions.cs
```diff
@@ -38,11 +38,11 @@ public T Value
             public bool Updated => _updated;
         }
 
-        private UpdateTracker<string> _tenantId = new UpdateTracker<string>(GetNonEmptyStringOrNull(EnvironmentVariables.TenantId));
-        private UpdateTracker<string> _interactiveBrowserTenantId = new UpdateTracker<string>(GetNonEmptyStringOrNull(EnvironmentVariables.TenantId));
-        private UpdateTracker<string> _sharedTokenCacheTenantId = new UpdateTracker<string>(GetNonEmptyStringOrNull(EnvironmentVariables.TenantId));
-        private UpdateTracker<string> _visualStudioTenantId = new UpdateTracker<string>(GetNonEmptyStringOrNull(EnvironmentVariables.TenantId));
-        private UpdateTracker<string> _visualStudioCodeTenantId = new UpdateTracker<string>(GetNonEmptyStringOrNull(EnvironmentVariables.TenantId));
+        private UpdateTracker<string> _tenantId = new UpdateTracker<string>(EnvironmentVariables.TenantId);
```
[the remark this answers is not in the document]
schaabs: I moved `GetNonEmptyStringOrNull` to the `EnvironmentVariables` class so that it is always called when we read environment variables, so I removed these calls here as they are redundant.

sdk/identity/Azure.Identity/src/Credentials/SharedTokenCacheCredentialOptions.cs
```diff
@@ -44,9 +47,18 @@ public string TenantId
...
-        /// Specifies the <see cref="TokenCachePersistenceOptions"/> to be used by the credential.
+        /// Specifies the <see cref="TokenCachePersistenceOptions"/> to be used by the credential. Value cannot be null.
         /// </summary>
-        public TokenCachePersistenceOptions TokenCachePersistenceOptions { get; }
+        public TokenCachePersistenceOptions TokenCachePersistenceOptions
+        {
+            get { return _tokenCachePersistenceOptions; }
+            set
+            {
+                Argument.AssertNotNull(value, nameof(value));
```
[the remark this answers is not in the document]
schaabs: This property used to only be settable through a constructor overload which ensures it always has a value. There are other places in the code which assume this will always have a value. I figured the easiest way to ensure adding a setter didn't regress anything was to prevent the value from being set to null. If you disagree I can reevaluate the other changes which would be needed to account for this possibly being null.

sdk/identity/Azure.Identity/src/Credentials/TokenCredentialOptions.cs
```diff
@@ -41,10 +41,83 @@ public Uri AuthorityHost
...
         internal bool IsLoggingPIIEnabled { get; set; }
 
-        /// <summary>
-        /// For multi-tenant applications, specifies additional tenants for which the credential may acquire tokens. Add the wildcard value "*" to allow the credential to acquire tokens for any tenant in which the application is installed.
-        /// </summary>
-        internal List<string> AdditionallyAllowedTenantsCore { get; set; } = new List<string>();
+        internal T Clone<T>()
+            where T : TokenCredentialOptions, new()
+        {
+            T clone = new T();
```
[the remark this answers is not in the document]
schaabs: In `SharedTokenCacheCredentialOptoins` the property `TokenCachePersistenceOptions` was readonly previously with a constructor overload which allowed users to set it. This PR already fixes that.

sdk/identity/Azure.Identity/src/Credentials/TokenCredentialOptions.cs
schaabs: 🤯

sdk/identity/Azure.Identity/src/EnvironmentVariables.cs
schaabs: In this case I actually want the value to be empty string rather than null. As you can see if the value of the environment variable is null actually switch it to string .Empty:

```c#
(Environment.GetEnvironmentVariable("AZURE_ADDITIONALLY_ALLOWED_TENANTS") ?? string. Empty)
```
I'm doing this so split will return an empty list if the value of the env var is null or string. Empty.

sdk/identity/Azure.Identity/tests/SharedTokenCacheCredentialTests.cs
schaabs: `SharedTokenCacheCredential` doesn't support `AdditionallyAllowedTenants`. While in the previous implementation this test was able to set `AdditionallyAllowedTenantsCore` on the `TokenCredentialOptions` base class, setting this property had no effect. Since this refactoring removes this property from the options base class, there is no longer anything to set.

sdk/identity/Azure.Identity/src/Credentials/AuthorizationCodeCredentialOptions.cs
```diff
@@ -26,7 +26,7 @@ public class AuthorizationCodeCredentialOptions : TokenCredentialOptions, ISuppo
         /// <summary>
         /// For multi-tenant applications, specifies additional tenants for which the credential may acquire tokens. Add the wildcard value "*" to allow the credential to acquire tokens for any tenant in which the application is installed.
         /// </summary>
-        public IList<string> AdditionallyAllowedTenants => AdditionallyAllowedTenantsCore;
+        public IList<string> AdditionallyAllowedTenants { get; internal set; } = new List<string>();
```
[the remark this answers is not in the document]
schaabs: I originally made this property readonly everywhere, and found that the impact to the tests would be quite large. Previously the tests were setting `AdditionallyAllowedTenantsCore` which is arguably much worse than this since not all options types used the base property in the same way. I think this is approach is much lower risk than that, and I found that other options types which expose readonly collection types in Azure.Core do this as well, such as `DiagnosticOptions.LoggedQueryParameters` and others.

If you feel strongly about this I can change it but it does end up impacting a large percentage of the tests.

.gitignore
```diff
@@ -176,3 +176,4 @@ artifacts
 
 # Temporary cadl folders for cadl generation
 TempCadlFiles/
+/sdk/identity/Azure.Identity/tests/TokenCredentialOptionsTests.cs
```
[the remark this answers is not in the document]
schaabs: D'oh, I added test coverage and ignored it in the same commit 🤦

christothes (approved):

PR #34195, opened 2023-02-15
Use ArrayBackedPropertyBag for PipelineRequest

AlexanderSher:
Benchmarks:
```
|                    Categories | count |            Method |        Mean |    Error |   StdDev | Ratio |  Gen 0 | Allocated |
|------------------------------ |------ |------------------ |------------:|---------:|---------:|------:|-------:|----------:|
|      CreateHttpRequestMessage |     2 | HttpRequestHeader |    656.1 ns |  2.51 ns |  2.35 ns |  1.00 | 0.0048 |     496 B |
|      CreateHttpRequestMessage |     2 | ArrayBackedHeader |    470.6 ns |  0.74 ns |  0.58 ns |  0.72 | 0.0048 |     496 B |
|                               |       |                   |             |          |          |       |        |           |
|      CreateHttpRequestMessage |     3 | HttpRequestHeader |    727.0 ns |  2.69 ns |  2.52 ns |  1.00 | 0.0048 |     496 B |
|      CreateHttpRequestMessage |     3 | ArrayBackedHeader |    608.6 ns |  2.65 ns |  2.35 ns |  0.84 | 0.0076 |     776 B |
|                               |       |                   |             |          |          |       |        |           |
|      CreateHttpRequestMessage |     8 | HttpRequestHeader |  1,287.3 ns |  4.97 ns |  4.40 ns |  1.00 | 0.0114 |   1,104 B |
|      CreateHttpRequestMessage |     8 | ArrayBackedHeader |  1,180.9 ns | 14.22 ns | 13.30 ns |  0.92 | 0.0134 |   1,384 B |
|                               |       |                   |             |          |          |       |        |           |
|      CreateHttpRequestMessage |    16 | HttpRequestHeader |  2,490.9 ns |  6.50 ns |  5.43 ns |  1.00 | 0.0267 |   2,736 B |
|      CreateHttpRequestMessage |    16 | ArrayBackedHeader |  2,504.4 ns |  6.96 ns |  5.81 ns |  1.01 | 0.0305 |   3,016 B |
|                               |       |                   |             |          |          |       |        |           |
|      CreateHttpRequestMessage |    32 | HttpRequestHeader |  4,374.9 ns | 12.06 ns | 10.07 ns |  1.00 | 0.0381 |   4,120 B |
|      CreateHttpRequestMessage |    32 | ArrayBackedHeader |  5,410.5 ns | 21.57 ns | 19.12 ns |  1.24 | 0.0458 |   4,656 B |
|                               |       |                   |             |          |          |       |        |           |
| CreateHttpRequestMessageTwice |     2 | HttpRequestHeader |  1,324.0 ns |  5.69 ns |  5.04 ns |  1.00 | 0.0114 |   1,256 B |
| CreateHttpRequestMessageTwice |     2 | ArrayBackedHeader |    950.8 ns |  4.91 ns |  4.35 ns |  0.72 | 0.0095 |   1,032 B |
|                               |       |                   |             |          |          |       |        |           |
| CreateHttpRequestMessageTwice |     3 | HttpRequestHeader |  1,506.9 ns |  5.77 ns |  5.40 ns |  1.00 | 0.0134 |   1,336 B |
| CreateHttpRequestMessageTwice |     3 | ArrayBackedHeader |  1,137.6 ns |  0.91 ns |  0.80 ns |  0.76 | 0.0134 |   1,312 B |
|                               |       |                   |             |          |          |       |        |           |
| CreateHttpRequestMessageTwice |     8 | HttpRequestHeader |  2,855.4 ns | 15.61 ns | 14.60 ns |  1.00 | 0.0305 |   2,888 B |
| CreateHttpRequestMessageTwice |     8 | ArrayBackedHeader |  2,187.8 ns |  2.24 ns |  1.75 ns |  0.77 | 0.0267 |   2,528 B |
|                               |       |                   |             |          |          |       |        |           |
| CreateHttpRequestMessageTwice |    16 | HttpRequestHeader |  5,452.2 ns | 15.40 ns | 12.86 ns |  1.00 | 0.0687 |   6,792 B |
| CreateHttpRequestMessageTwice |    16 | ArrayBackedHeader |  4,419.1 ns | 29.19 ns | 27.31 ns |  0.81 | 0.0610 |   5,792 B |
|                               |       |                   |             |          |          |       |        |           |
| CreateHttpRequestMessageTwice |    32 | HttpRequestHeader |  9,906.1 ns | 11.67 ns |  9.11 ns |  1.00 | 0.1068 |  10,840 B |
| CreateHttpRequestMessageTwice |    32 | ArrayBackedHeader |  8,834.6 ns | 20.36 ns | 15.90 ns |  0.89 | 0.0916 |   8,816 B |
|                               |       |                   |             |          |          |       |        |           |
|                 MultipleReads |     8 | HttpRequestHeader |  2,976.5 ns | 11.14 ns |  9.30 ns |  1.00 | 0.0229 |   2,472 B |
|                 MultipleReads |     8 | ArrayBackedHeader |  1,447.9 ns |  2.72 ns |  2.55 ns |  0.49 | 0.0153 |   1,472 B |
|                               |       |                   |             |          |          |       |        |           |
|                 MultipleReads |    16 | HttpRequestHeader |  5,971.1 ns | 19.34 ns | 17.14 ns |  1.00 | 0.0534 |   5,448 B |
|                 MultipleReads |    16 | ArrayBackedHeader |  3,195.2 ns | 24.38 ns | 22.80 ns |  0.54 | 0.0305 |   3,168 B |
|                               |       |                   |             |          |          |       |        |           |
|                 MultipleReads |    32 | HttpRequestHeader | 11,225.7 ns | 40.46 ns | 35.87 ns |  1.00 | 0.0916 |   9,520 B |
|                 MultipleReads |    32 | ArrayBackedHeader |  7,953.4 ns | 37.65 ns | 35.22 ns |  0.71 | 0.0458 |   4,936 B |
```

First scenario - `CreateHttpRequestMessage` - is the base one, when we create a request and send it directly to the socket. In case of less than 16 headers, `ArrayBackedPropertyBag` is faster
Second scenario - `CreateHttpRequestMessageTwice` simulates the retry case. Here, even with 32 headers the benefit is about 10%
Third scenario - `MultipleReads` - simulates `Azure.Storage` case when headers are used to create a signature


sdk/core/Azure.Core/src/Internal/ArrayBackedPropertyBag.cs
```diff
@@ -47,59 +72,150 @@ public bool TryGetValue(ulong key, out object? value)
...
             if (index < 0)
                 AddInternal(key, value);
             else
-                SetAt(index, new KVP(key, value));
+                SetAt(index, new Kvp(key, value));
         }
 
-        [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        private void AddInternal(ulong key, object value)
+        public bool TryDelete(TKey key)
```
christothes, line 98: nit: should we call this `TryRemove`?

sdk/core/Azure.Core/src/Internal/ArrayBackedPropertyBag.cs
```diff
@@ -11,33 +13,56 @@ namespace Azure.Core
...
+                value = Value;
+            }
         }
 
+        public int Count => _count;
+
         public bool IsEmpty => _count == 0;
 
-        public bool TryGetValue(ulong key, out object? value)
+        public void GetAt(int index, out TKey key, out TValue value)
```
christothes, line 45: Now that we expose this, do you think it would make sense to have the Add/Set methods also return the index of the newly added/set item? Otherwise, is there another way users of this API could know the index of an item they want to get?

I realize that the addition of the `TryDelete` method potentially makes previously returned indexes untrustworthy, but perhaps that is an edge case or at least can be handled defensively.

KrzysztofCwalina:
It would be good if this PR description started with a note explaining about what this PR does, as opposed to starting with a benchmark table :-)

But I am not surprised this improves perf. The headers collection is pretty inefficient. We should send this benchmark and the scenarios we have (changing header values) that make the BCL headers collection suboptimal. Maybe this can be fixed in the BCL so that we don't have to write code like in this PR.

sdk/core/Azure.Core/src/Internal/ArrayBackedPropertyBag.cs
```diff
@@ -11,33 +13,56 @@ namespace Azure.Core
...
+        {
+            if (index >= _count)
+            {
+                throw new ArgumentOutOfRangeException(nameof(index));
```
JoshLove-msft, line 49: Should we avoid throwing directly to allow the method to be inlined?

sdk/core/Azure.Core/src/Internal/ArrayBackedPropertyBag.cs
```diff
@@ -11,33 +13,56 @@ namespace Azure.Core
...
-        private KVP _first;
-        private KVP _second;
-        private KVP[] _rest;
+        private Kvp _first;
+        private Kvp _second;
+        private Kvp[] _rest;
         private int _count;
 
-        private readonly struct KVP
+        private readonly struct Kvp
```
JoshLove-msft, line 23: Interesting, why not use KeyValuePair which is also a readonly struct.

sdk/core/Azure.Core/src/Internal/ArrayBackedPropertyBag.cs
```diff
@@ -47,59 +72,150 @@ public bool TryGetValue(ulong key, out object? value)
...
+                    {
+                        _second = new Kvp(key, value);
+                        _count = 2;
+                    }
+
+                    return;
+                default:
+                    if (_rest == null)
+                    {
+                        _rest = ArrayPool<Kvp>.Shared.Rent(8);
```
JoshLove-msft, line 199: Do we ever return this? Should this type implement IDisposable?

AlexanderSher: We do, and there is a `Dispose` method that is called in `PipelineRequest.Dispose`. @christothes , is there a reason why we don't implement `IDisposable` here?

sdk/core/Azure.Core/src/Internal/ArrayBackedPropertyBag.cs
AlexanderSher: Results with KeyValuePair are about 15% worse across all tests

sdk/core/Azure.Core/src/Internal/ArrayBackedPropertyBag.cs
christothes: I don't recall now exactly, but when I added the interface, it caused some problem.

sdk/core/Azure.Core/src/Internal/ArrayBackedPropertyBag.cs
AlexanderSher: yeah, and we don't really need this exception, cause `_rest` would throw the same exception

sdk/core/Azure.Core/src/Internal/ArrayBackedPropertyBag.cs
AlexanderSher: Ok, let's skip the interface then

JoshLove-msft: Was the issue that we needed to make another type IDisposable? 😆

sdk/core/Azure.Core/src/Internal/ArrayBackedPropertyBag.cs
AlexanderSher: Fixed

sdk/core/Azure.Core/src/Internal/ArrayBackedPropertyBag.cs
AlexanderSher: Since it is an internal type, we can always do it when needed. For now, I'd prefer to leave it as it is.

JoshLove-msft (approved):

seanmcc-msft:
This PR appears to have caused test flakiness in the .NET Live tests pipeline - Beginning on 2/17, we have seen intermittent test failures in our live batch tests - https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2203247&view=ms.vss-test-web.build-test-results-tab&runId=39465973&resultId=100295&paneView=debug

Based on the timing, it appears it was caused by this commit.

seanmcc-msft:
@AlexanderSher @amnguye

PR #34275, opened 2023-02-16
[AzureMonitorExporter] connection string via environment variable

TimothyMothra:
## Changes
- add support for Connection String via Environment Variable.
- refactor Transmitter ctor.
  this was getting large. I moved individual items into a static Initialize method.

### DIFF
--- a/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/AzureMonitorTransmitter.cs
+++ b/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/AzureMonitorTransmitter.cs
@@ -35,49 +35,12 @@ namespace Azure.Monitor.OpenTelemetry.Exporter.Internals
             }
 
             options.Retry.MaxRetries = 0;
-            _connectionVars = ConnectionStringParser.GetValues(options.ConnectionString);
 
-            HttpPipeline pipeline;
-            if (credential != null)
-            {
-                var scope = AadHelper.GetScope(_connectionVars.AadAudience);
-                var httpPipelinePolicy = new HttpPipelinePolicy[]
-                {
-                    new BearerTokenAuthenticationPolicy(credential, scope),
-                    new IngestionRedirectPolicy()
-                };
+            _connectionVars = InitializeConnectionVars(options);
 
-                pipeline = HttpPipelineBuilder.Build(options, httpPipelinePolicy);
-                AzureMonitorExporterEventSource.Log.WriteInformational("SetAADCredentialsToPipeline", $"HttpPipelineBuilder is built with AAD Credentials. TokenCredential: {credential.GetType().Name} Scope: {scope}");
-            }
-            else
-            {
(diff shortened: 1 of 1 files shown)


sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/AzureMonitorTransmitter.cs
```diff
@@ -180,6 +143,80 @@ private static ExportResult IsSuccess(HttpMessage httpMessage)
...
+                    return null;
+                }
+            }
+
+            return null;
+        }
+
+        private static ConnectionVars InitializeConnectionVars(AzureMonitorExporterOptions options)
+        {
+            if (options.ConnectionString == null)
```
rajkumar-rangaraj, line 203: Debatable topic :blush:

Which should take precedence - options set in the code or environment variable?

TimothyMothra: The decision we made for AI SDK was anything hard-coded would take precedence over the env var.

vishweshbankwar (approved):
LGTM - a todo note to discuss debatable topic would be great and a test :)

TimothyMothra:
> LGTM - a todo note to discuss debatable topic would be great and a test :)

Tests with Environment Variables is tricky and can have side effects. In Application Insights we solved this with an abstraction layer. 
Long term, I'd like to explore something similar here.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/AzureMonitorTransmitter.cs
TimothyMothra: Adding @vishweshbankwar to this thread.
I'm not certain that there's anything to debate here. :)
I feel like this was settled when we built Connection Strings a few years ago.

If we need to debate this, I suggest we bring this up in the next Contributors Meeting and let all languages weigh in.

rajkumar-rangaraj (approved):

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/AzureMonitorTransmitter.cs
vishweshbankwar: question: how does OTel SDK handles it? I cannot recall. I think we just need to be consistent with it.

vishweshbankwar: > If we need to debate this, I suggest we bring this up in the next Contributors Meeting and let all languages weigh in. 👍

TimothyMothra: Added this to the agenda for next week.

Issue #34262, opened 2023-02-16
ACR API Updates from ApiView


annelo-msft:
- [ ] Change name of blob in method names, e.g. to `DeleteContainerBlob` or `DeleteRegistryBlob`, per [architect feedback](https://apiview.dev/Assemblies/Conversation/7ebfcb36ff6149ba96170bd059094513)

> [KrzysztofCwalina](https://apiview.dev/Assemblies/Profile/KrzysztofCwalina) [11 weeks ago](https://apiview.dev/Assemblies/Conversation/7ebfcb36ff6149ba96170bd059094513#402c13a290ba47e0b4b9a10be0659443)
> One suggestion is to call it "ContainerBlob" where "Blob" appears in the API.
> 
> Clarification: I made this comment and I was referring to method names, e.g. DeleteContainerBlob instead of DeleteBlob. Though, now that I think about it, probably DeleteRegistryBlob would work better.

- [x] Remove UploadBlobOptions since `maxChunkSize` is now in client options.

annelo-msft:
Closing as "won't fix" per discussion with Krzyszotf.

PR #34328, opened 2023-02-17
Remove ACR UploadBlobOptions type

annelo-msft:
We moved `maxChunkSize` setting into ClientOptions in https://github.com/Azure/azure-sdk-for-net/pull/33432, but missed removing this options type.  This PR removes the type, which was not being used by the client.

Addresses https://github.com/Azure/azure-sdk-for-net/issues/34262


sdk/containerregistry/Azure.Containers.ContainerRegistry/api/Azure.Containers.ContainerRegistry.netstandard2.0.cs
```diff
@@ -317,11 +317,6 @@ public partial class OciManifest
         public System.Collections.Generic.IList<Azure.Containers.ContainerRegistry.Specialized.OciBlobDescriptor> Layers { get { throw null; } }
         public int? SchemaVersion { get { throw null; } set { } }
     }
-    public partial class UploadBlobOptions
```
pallavit, line 320: I think we use this type in changelog as well.

pallavit:
Small comment, other than that LGTM.

pallavit (approved):

sdk/containerregistry/Azure.Containers.ContainerRegistry/api/Azure.Containers.ContainerRegistry.netstandard2.0.cs
annelo-msft: Good call!  I will add it - thank you!

PR #34332, opened 2023-02-17
Implement updated digest validation for DownloadManifest

### LINKED ISSUE #33778: [ACR] Implement digest check for DownloadManifest
Per [this discussion](https://github.com/Azure/azure-sdk-for-net/issues/33646) the download manifest operation should implement the following logic for checking the digest:

1. If the caller passed a digest to fetch the manifest (reference string begins with `sha256:`), validate that this digest matches the digest computed on the response body content.
2. If the caller passed a tag to fetch the manifest, validate that the digest in the Docker-Content-Digest header received on the response matches the digest computed on the response body content.

annelo-msft:
Fixes https://github.com/Azure/azure-sdk-for-net/issues/33778

### DIFF
--- a/sdk/containerregistry/Azure.Containers.ContainerRegistry/tests/ContainerRegistryBlobClientTests.cs
+++ b/sdk/containerregistry/Azure.Containers.ContainerRegistry/tests/ContainerRegistryBlobClientTests.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Threading.Tasks;
 using Azure.Containers.ContainerRegistry.Specialized;
 using Azure.Core;
 using Azure.Core.TestFramework;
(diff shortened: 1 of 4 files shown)


sdk/containerregistry/Azure.Containers.ContainerRegistry/tests/ContainerRegistryBlobClientTests.cs
```diff
@@ -60,5 +61,54 @@ public void ServiceMethodsValidateArguments()
             Assert.That(async () => await client.UploadManifestAsync(manifestStream: null), Throws.InstanceOf<ArgumentNullException>(), "The method should validate that `manifest stream` is not null.");
             Assert.That(async () => await client.UploadBlobAsync(null), Throws.InstanceOf<ArgumentNullException>(), "The method should validate that `stream` is not null.");
         }
+
+        /// <summary>
+        /// Validates digest checks for DownloadManifest.
+        /// </summary>
+        [Test]
+        public async Task DownloadManifestValidatesDigest()
```
pallavit, line 69: Would we not need the test recording file as well?

pallavit: Never mind - just realized it is not a recorded test!

pallavit (approved):
:shipit:

PR #34338, opened 2023-02-17
Handle special case when HeaderStringValues.Count == 1

AlexanderSher:
When `HeaderStringValues.Count == 1`, `HttpHeaders` special cases it and creates `HeaderStringValues` instance from a single string: https://github.com/dotnet/runtime/blob/ef5e27eacecf34a36d72a8feb9082f408779675a/src/libraries/System.Net.Http/src/System/Net/Http/Headers/HttpHeadersNonValidated.cs#L150
https://github.com/dotnet/runtime/blob/ef5e27eacecf34a36d72a8feb9082f408779675a/src/libraries/System.Net.Http/src/System/Net/Http/Headers/HttpHeaders.cs#L1105
          
Then, `HeaderStringValues.ToString` simply returns that value:
https://github.com/dotnet/runtime/blob/729bf92e6e2f91aa337da9459bef079b14a0bf34/src/libraries/System.Net.Http/src/System/Net/Http/Headers/HeaderStringValues.cs#L47

While `HeaderStringValueToStringVsEnumerator` performance test shows that `HeaderStringValues.ToString` is always faster than `DefaultInterpolatedStringHandler`, we can't use it for `HeaderStringValues.Count > 1` because it uses `", "` as default separator and doesn't allow customization. Hence, custom enumeration logic is required.

Benchmark results:
```
|           Method | Count |      Mean |     Error |    StdDev |   Gen0 | Allocated |
|----------------- |------ |----------:|----------:|----------:|-------:|----------:|
|     CallToString |     1 |  2.282 ns | 0.0029 ns | 0.0024 ns |      - |         - |
| CreateEnumerator |     1 | 41.830 ns | 0.2883 ns | 0.2697 ns | 0.0004 |      40 B |
|     CallToString |     2 | 29.230 ns | 0.2876 ns | 0.2691 ns | 0.0005 |      56 B |
| CreateEnumerator |     2 | 52.589 ns | 0.1854 ns | 0.1643 ns | 0.0005 |      56 B |
|     CallToString |     5 | 57.748 ns | 0.1114 ns | 0.0870 ns | 0.0011 |     112 B |
| CreateEnumerator |     5 | 76.548 ns | 0.2156 ns | 0.1800 ns | 0.0011 |     104 B |
```

### DIFF
--- a/sdk/core/Azure.Core/src/Pipeline/HttpClientTransport.cs
+++ b/sdk/core/Azure.Core/src/Pipeline/HttpClientTransport.cs
@@ -214,7 +214,7 @@ namespace Azure.Core.Pipeline
             if (headers.NonValidated.TryGetValues(name, out HeaderStringValues values) ||
                 content is not null && content.Headers.NonValidated.TryGetValues(name, out values))
             {
-                value = JoinHeaderValues(values);
+                value = values.ToString();
                 return true;
             }
 #else
@@ -254,14 +254,14 @@ namespace Azure.Core.Pipeline
 #if NET6_0_OR_GREATER
             foreach (var (key, value) in headers.NonValidated)
             {
-                yield return new HttpHeader(key, JoinHeaderValues(value));
+                yield return new HttpHeader(key, value.ToString());
             }
 
             if (content is not null)
             {
                 foreach (var (key, value) in content.Headers.NonValidated)
                 {
-                    yield return new HttpHeader(key, JoinHeaderValues(value));
+                    yield return new HttpHeader(key, value.ToString());
                 }
             }
 #else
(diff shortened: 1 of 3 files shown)


sdk/core/Azure.Core/src/Pipeline/HttpClientTransport.cs
```diff
@@ -325,6 +325,16 @@ private static string JoinHeaderValues(HeaderStringValues values)
                 return string.Empty;
             }
 
+            // Special case when HeaderStringValues.Count == 1, because HttpHeaders also special cases it and creates HeaderStringValues instance from a single string
+            // https://github.com/dotnet/runtime/blob/ef5e27eacecf34a36d72a8feb9082f408779675a/src/libraries/System.Net.Http/src/System/Net/Http/Headers/HttpHeadersNonValidated.cs#L150
+            // https://github.com/dotnet/runtime/blob/ef5e27eacecf34a36d72a8feb9082f408779675a/src/libraries/System.Net.Http/src/System/Net/Http/Headers/HttpHeaders.cs#L1105
+            // Which is later used in HeaderStringValues.ToString:
+            // https://github.com/dotnet/runtime/blob/729bf92e6e2f91aa337da9459bef079b14a0bf34/src/libraries/System.Net.Http/src/System/Net/Http/Headers/HeaderStringValues.cs#L47
+            if (count == 1)
```
JoshLove-msft, line 333: Can we add a test?

AlexanderSher: What kind of test?

JoshLove-msft: A test where there is a single header?

JoshLove-msft: Perhaps this is already covered, but it would be good to verify.

AlexanderSher: It turns out that I have run benchmark over the wrong `string.Join` overload. `HeaderStringValues.ToString()` is better for most of the cases.

AlexanderSher: Ok, now I recall why I have added this in the first place. `ToString` has different separator value. I'll leave perf test here and add a comment.

annelo-msft: `SettingContentStreamPreservesHeaders` definitely covers the `count == 1` case, but I don't see it covering the zero or >1 cases.

annelo-msft (approved):

sdk/core/Azure.Core/src/Pipeline/HttpClientTransport.cs
AlexanderSher: >1 cases is covered in `CanGetAndSetMultiValueResponseHeaders` and `ResponseHeadersAreSplit`

PR #34344, opened 2023-02-17
[Event Hubs] Function Test Fix

jsquire:
# Summary

The focus of these changes is to fix an assumption in the Functions extension package that the same Storage suffix is valid in all clouds.  The implementation supports the Azure public cloud only, while the test was asserting that the scenario was valid for all clouds.  Going forward, the test will be ignored when the current execution environment is not a supported cloud.

Also riding along is a tweak to a processor test to extract additional error details for failure investigation.

### DIFF
--- a/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/tests/EventHubEndToEndTests.cs
+++ b/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/tests/EventHubEndToEndTests.cs
@@ -8,12 +8,14 @@ using System.Reflection;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
+using Azure.Core;
 using Azure.Core.TestFramework;
 using Azure.Messaging.EventHubs;
 using Azure.Messaging.EventHubs.Consumer;
 using Azure.Messaging.EventHubs.Producer;
 using Azure.Messaging.EventHubs.Tests;
 using Microsoft.Azure.WebJobs.EventHubs;
+using Microsoft.Azure.WebJobs.Extensions.Clients.Shared;
 using Microsoft.Azure.WebJobs.Host.TestCommon;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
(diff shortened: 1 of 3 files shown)


sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/tests/EventHubEndToEndTests.cs
```diff
@@ -252,6 +254,17 @@ public async Task CanSendAndReceive_BlobServiceUri_InConfiguration()
         [Test]
         public async Task CanSendAndReceive_AccountName_InConfiguration()
         {
+            // Use of an account name assumes an endpoint suffix that is only valid in some Azure
+            // cloud environments. If the current execution environment uses a different suffix,
+            // ignore the test.
+
+            var defaultSuffix = StorageClientProvider<object, ClientOptions>.DefaultStorageEndpointSuffix;
+
+            if (string.Equals(EventHubsTestEnvironment.Instance.StorageEndpointSuffix, defaultSuffix, StringComparison.OrdinalIgnoreCase))
```
JoshLove-msft, line 263:
```suggestion
            if (!string.Equals(EventHubsTestEnvironment.Instance.StorageEndpointSuffix, defaultSuffix, StringComparison.OrdinalIgnoreCase))
```

JoshLove-msft (approved):

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/tests/EventHubEndToEndTests.cs
jsquire: uh... yeah, that's an embarrassing one.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/tests/EventHubEndToEndTests.cs
```diff
@@ -252,6 +254,17 @@ public async Task CanSendAndReceive_BlobServiceUri_InConfiguration()
...
         {
+            // Use of an account name assumes an endpoint suffix that is only valid in some Azure
+            // cloud environments. If the current execution environment uses a different suffix,
+            // ignore the test.
+
+            var defaultSuffix = StorageClientProvider<object, ClientOptions>.DefaultStorageEndpointSuffix;
+
+            if (!string.Equals(EventHubsTestEnvironment.Instance.StorageEndpointSuffix, defaultSuffix, StringComparison.OrdinalIgnoreCase))
+            {
+                Assert.Ignore($"This test can only be run in the Azure cloud associated  with the suffix: `{defaultSuffix}`.");
```
JoshLove-msft, line 265:
```suggestion
                Assert.Ignore($"This test can only be run in the Azure cloud associated with the suffix: `{defaultSuffix}`.");
```

jsquire:

christothes (approved):

Issue #32948, opened 2022-12-07
Get the Container Registry test pipelines green.


pallavit:
This is a follow-up issue to get the ACR test pipelines green.

annelo-msft:
Since some of the failures are non-deterministic, tracking successful runs to evaluate success of interventions:

- 12/16: Green - https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2061934&view=results

annelo-msft:
Use this approach: https://github.com/Azure/azure-sdk-for-net/pull/33829#discussion_r1094748921

PR #34346, opened 2023-02-17
Make ACR live tests run serially

annelo-msft:
We have periodic tests failures in our live tests, and some are due to tests accessing the same resource at the same time.

In an effort to address https://github.com/Azure/azure-sdk-for-net/issues/32948, this PR implements the instructions in the TestFramework README regarding [Running live tests serially](https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/core/Azure.Core.TestFramework#running-live-tests-serially).


JoshLove-msft:
> We have periodic tests failures in our live tests, and some are due to tests accessing the same resource at the same time.

Is this happening across test assemblies? The fix here would prevent issues like this from occurring across assemblies.

JoshLove-msft (approved):

PR #34354, opened 2023-02-17
[AzureMonitorExporter] transmitter factory

TimothyMothra:
An ideal users will create a single exporter for each signal (Logs, Metrics, Traces).
This implements a factory to ensure that only one instance of the Transmitter is created for any unique connection string.

## Changes
- Add new class `TransmitterFactory`
- Add new test `TransmitterFactoryTests`
- Update Exporters to use factory

### DIFF
--- a/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/AzureMonitorLogExporter.cs
+++ b/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/AzureMonitorLogExporter.cs
@@ -19,7 +19,7 @@ namespace Azure.Monitor.OpenTelemetry.Exporter
         private readonly AzureMonitorPersistentStorage? _persistentStorage;
         private AzureMonitorResource? _resource;
 
-        public AzureMonitorLogExporter(AzureMonitorExporterOptions options, TokenCredential? credential = null) : this(new AzureMonitorTransmitter(options, credential))
+        public AzureMonitorLogExporter(AzureMonitorExporterOptions options, TokenCredential? credential = null) : this(TransmitterFactory.Instance.Get(options, credential))
         {
         }
 
--- a/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TransmitterFactory.cs
+++ b/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TransmitterFactory.cs
@@ -1,0 +1,42 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Collections.Generic;
+using Azure.Core;
+
+namespace Azure.Monitor.OpenTelemetry.Exporter.Internals
+{
+    /// <summary>
+    /// This Factory encapsulates the <see cref="AzureMonitorTransmitter"/>.
+    /// An ideal users will create a single exporter for each signal (Logs, Metrics, Traces).
+    /// This factory should ensure that only one instance of the Transmitter is created for
+    /// any unique connection string.
+    /// </summary>
+    internal class TransmitterFactory
+    {
+        public static TransmitterFactory Instance = new();
+
+        internal readonly Dictionary<string?, AzureMonitorTransmitter> _transmitters = new();
+        private readonly object _lockObj = new();
+
+        public AzureMonitorTransmitter Get(AzureMonitorExporterOptions azureMonitorExporterOptions, TokenCredential? tokenCredential = null)
+        {
+            var key = azureMonitorExporterOptions.ConnectionString ?? string.Empty;
+
+            if (!_transmitters.TryGetValue(key, out AzureMonitorTransmitter transmitter))
+            {
+                lock (_lockObj)
+                {
+                    if (!_transmitters.TryGetValue(key, out transmitter))
+                    {
+                        transmitter = new AzureMonitorTransmitter(azureMonitorExporterOptions, tokenCredential);
+
+                        _transmitters.Add(key, transmitter);
+                    }
+                }
+            }
+
+            return transmitter;
+        }
+    }
+}
--- a/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TransmitterFactory.cs
+++ b/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TransmitterFactory.cs
@@ -16,7 +16,7 @@ namespace Azure.Monitor.OpenTelemetry.Exporter.Internals
     {
         public static TransmitterFactory Instance = new();
 
-        internal readonly Dictionary<string?, AzureMonitorTransmitter> _transmitters = new();
+        internal readonly Dictionary<string, AzureMonitorTransmitter> _transmitters = new();
         private readonly object _lockObj = new();
 
         public AzureMonitorTransmitter Get(AzureMonitorExporterOptions azureMonitorExporterOptions, TokenCredential? tokenCredential = null)
--- a/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/AzureMonitorMetricExporter.cs
+++ b/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/AzureMonitorMetricExporter.cs
@@ -19,7 +19,7 @@ namespace Azure.Monitor.OpenTelemetry.Exporter
         private readonly AzureMonitorPersistentStorage? _persistentStorage;
         private AzureMonitorResource? _resource;
 
-        public AzureMonitorMetricExporter(AzureMonitorExporterOptions options, TokenCredential? credential = null) : this(new AzureMonitorTransmitter(options, credential))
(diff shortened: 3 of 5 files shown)


sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TransmitterFactory.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// This Factory encapsulates the <see cref="AzureMonitorTransmitter"/>.
+    /// An ideal users will create a single exporter for each signal (Logs, Metrics, Traces).
+    /// This factory should ensure that only one instance of the Transmitter is created for
+    /// any unique connection string.
+    /// </summary>
+    internal class TransmitterFactory
+    {
+        public static TransmitterFactory Instance = new();
+
+        internal readonly Dictionary<string, AzureMonitorTransmitter> _transmitters = new();
```
vishweshbankwar, line 19: can we use ConcurrentDictionary instead?

TimothyMothra: good question. 
this dictionary isn't going to see heavy use, so I'm not sure that we need the extra overhead.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/AzureMonitorLogExporter.cs
```diff
@@ -19,7 +19,7 @@ internal class AzureMonitorLogExporter : BaseExporter<LogRecord>
         private readonly AzureMonitorPersistentStorage? _persistentStorage;
         private AzureMonitorResource? _resource;
 
-        public AzureMonitorLogExporter(AzureMonitorExporterOptions options, TokenCredential? credential = null) : this(new AzureMonitorTransmitter(options, credential))
+        public AzureMonitorLogExporter(AzureMonitorExporterOptions options, TokenCredential? credential = null) : this(TransmitterFactory.Instance.Get(options, credential))
```
cijothomas, line 22: what is the benefit of doing this?

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TransmitterFactory.cs
```diff
@@ -0,0 +1,42 @@
...
+        private readonly object _lockObj = new();
+
+        public AzureMonitorTransmitter Get(AzureMonitorExporterOptions azureMonitorExporterOptions, TokenCredential? tokenCredential = null)
+        {
+            var key = azureMonitorExporterOptions.ConnectionString ?? string.Empty;
```
rajkumar-rangaraj, line 24: Not sure, how dictionary will look up with a huge string, shall we use ikey here?

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TransmitterFactory.cs
```diff
@@ -0,0 +1,42 @@
...
+
+            if (!_transmitters.TryGetValue(key, out AzureMonitorTransmitter transmitter))
+            {
+                lock (_lockObj)
```
rajkumar-rangaraj, line 28: If ConcurrentDictionary could avoid this lock, my vote will be for it.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/AzureMonitorLogExporter.cs
TimothyMothra: which _this_ are you referring to?


Today, if a user creates all three exporters, the SDK creates three Transmitters. 
Each transmitter would have it's own instance of RestClient, PersistantStorage, and Statsbeat.

We want to have only 1 instance of Transmitter per ConnectionString. 
For an ideal use case, 1 user = 1 connection string = 1 transmitter.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TransmitterFactory.cs
TimothyMothra: my understanding is that dictionary will hash the key, regardless the length. 

We can change this to ikey, but would require additional refactoring. currently we only parse the connection string inside the Transmitter.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TransmitterFactory.cs
TimothyMothra: see the other thread above.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TransmitterFactory.cs
TimothyMothra: adding @rajkumar-rangaraj here. i'm not sure that ConcurrentDictionary avoids the extra lock but I'll get more information.

At worse, ConcurrentDictionary would prevent the second Add, but if we're calling Add a second time that means we've already created a second Transmitter :(

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TransmitterFactory.cs
rajkumar-rangaraj: If we have tested and if it works with connectionstring, it fine to use it.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TransmitterFactory.cs
rajkumar-rangaraj: can't we use something like dispose pattern here instead of locks?

cijothomas: the current lock approach does not ever "waste" a Transmitter due to locks. It should be fine.

rajkumar-rangaraj: Locks are primary cause for thread contention. If there are other ways we should avoid its usage.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/AzureMonitorLogExporter.cs
cijothomas: Thanks.
we were duplicating statsbeat today, if i add logs, metrics and traces?

TimothyMothra: Not yet because statsbeat isn't in use 😂
We're waiting on this change

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TransmitterFactory.cs
TimothyMothra: ConcurrentDictionary with a LazyInitializer might work. 
Though i'm worried that would be overengineering with a lot of extra overhead.

cijothomas: this lock is not in the hot path, right? so should not be a concern

vishweshbankwar (approved):

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TransmitterFactory.cs
vishweshbankwar: I was referring to using this https://learn.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.getoradd?view=net-7.0.

TimothyMothra: GetOrAdd() is a good idea, but reading the description I don't think this meets our needs here.

> For modifications and write operations to the dictionary, [ConcurrentDictionary<TKey,TValue>](https://learn.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2?view=net-7.0) uses fine-grained locking to ensure thread safety. (Read operations on the dictionary are performed in a lock-free manner.) However, the valueFactory delegate is called outside the locks to avoid the problems that can arise from executing unknown code under a lock. Since a key/value can be inserted by another thread while valueFactory is generating a value, you cannot trust that just because valueFactory executed, its produced value will be inserted into the dictionary and returned.

rajkumar-rangaraj (approved):
We may not hit the lock issue as all the 3 signals gets initialized synchronously. Could merge the PR and refactor later if it causes an issue.

PR #34355, opened 2023-02-17
Ensure proper parethesis matching in UserAgent

### LINKED ISSUE #34185: [BUG] Invalid user-agent header generated when OSDescription has unmatched parentheses
### Library name and version

Azure.Core 1.28.0

### Describe the bug

When RuntimeInformation.OSDescription has a string containing "(" with no matching ")", TelemetryDetails constructs an invalid User-Agent string. 

For example, if the OSDescription is "MyOSDescription (2023-", the resulting User-Agent header will be "azsdk-net-Security.KeyVault.Certificates/4.3.0 (.NET 6.0.0; MyOSDescription (2023-)". The mismatched parentheses cause the header to fail to be parsed by a server that expects headers to conform to the HTTP specification.

### Expected behavior

User-Agent header should be valid according to the HTTP specification. Characters in OSDescription that could lead to an invalid header should be correctly escaped.

### Actual behavior

Parentheses in OSDescription are not escaped and can lead to invalid User-Agent headers

### Reproduction Steps

Run on an OS platform where OSDescription has an unmatched parentheses. Inspect the User-Agent header added to all REST API calls.

### Environment

.NET 6.0.13
OSDescription: Linux 6.1.10-1-liquorix-amd64 # 1 ZEN SMP PREEMPT_DYNAMIC liquorix 6.1-23ubuntu1~kinetic (2023-

christothes:
fixes #34185


sdk/core/Azure.Core/src/TelemetryDetails.cs
```diff
@@ -91,5 +98,72 @@ internal static string GenerateUserAgentString(Assembly clientAssembly, string?
         /// The properly formatted UserAgent string based on this <see cref="TelemetryDetails"/> instance.
         /// </summary>
         public override string ToString() => _userAgent;
+
+        private static string EnsureProperParenthesisMatching(string userAgent)
```
JoshLove-msft, line 102: I'm not seeing where this is called?

sdk/core/Azure.Core/tests/TelemetryDetailsTests.cs
```diff
@@ -62,5 +64,48 @@ public void AppliesToMessage()
...
+        [Test]
+        [TestCase("MyOSDescription (2023-", "MyOSDescription (2023-)")]
+        [TestCase("MyOSDescription (2023-)", "MyOSDescription (2023-)")]
+        [TestCase("MyOSDescription (()", "MyOSDescription ()")]
+        [TestCase("MyOSDescription (", "MyOSDescription ()")]
+        [TestCase("MyOSDescription ())", "MyOSDescription ()")]
+        [TestCase("MyOSDescription )", "MyOSDescription ")]
+        [TestCase("MyO)SDescription ", "MyOSDescription ")]
+        [TestCase("MyO(SDescription ", "MyOSDescription ")]
+        public void ValidatesProperParenthesisMatching(string input, string output)
```
JoshLove-msft, line 77: output is not used in the test

sdk/core/Azure.Core/src/TelemetryDetails.cs
```diff
@@ -91,5 +99,73 @@ internal static string GenerateUserAgentString(Assembly clientAssembly, string?
...
+        {
+            var stack = new Stack<char>();
+            bool needsFix = false;
+            StringBuilder? fixedUserAgent = null;
+            for (int i = 0; i < userAgent.Length; i++)
+            {
+                switch (userAgent[i])
+                {
+                    case '(':
+                        stack.Push('(');
```
JoshLove-msft, line 113: We seem to only ever append open parens - why do we even need a stack? Could we just keep track of the open - close parens count?

christothes: Yeah, this will just ensure that we have the correct number of matching or will remove closed parentheses that don't have a matching open. I don't think just the count is sufficient unless we only wanted to validate without fixing.

JoshLove-msft: But if we are only pushing one type of element, how I don't see why the stack is needed.

christothes: oops - yes, you are correct. Fixed.

sdk/core/Azure.Core/tests/TelemetryDetailsTests.cs
```diff
@@ -62,5 +64,48 @@ public void AppliesToMessage()
 
             Assert.AreEqual(target.ToString(), actualValue);
         }
+
+        [Test]
+        [TestCase("MyOSDescription (2023-", "MyOSDescription (2023-)")]
+        [TestCase("MyOSDescription (2023-)", "MyOSDescription (2023-)")]
+        [TestCase("MyOSDescription (()", "MyOSDescription (())")]
+        [TestCase("MyOSDescription (", "MyOSDescription ()")]
+        [TestCase("MyOSDescription ())", "MyOSDescription ()")]
```
JoshLove-msft, line 73: So the rule is:
extra open parens => add closing parens at end of string
extra close parens => delete extra closing parens

I wonder if it would not be more explainable to escape the parens if we determine the input to be invalid.

seanbamsft: [RFC 7230 section 3.2.6](https://www.rfc-editor.org/rfc/rfc7230#section-3.2.6) defines the rules for the "comment" definition that is used by the User-Agent header. I believe you'd use "\(" and "\)" to escape parentheses contained in the comment. The gaps in the ASCII range in the ctext definition correspond to "(", ")", and "\" and there's a note below the quoted-pair definition about not escaping anything other than "(", ")", and "\" in comments.

christothes: Thanks @seanbamsft - As long as you are confident that escaping them would address the telemetry issue described in https://github.com/Azure/azure-sdk-for-net/issues/34185 this makes sense.

sdk/core/Azure.Core/src/TelemetryDetails.cs
```diff
@@ -80,16 +87,80 @@ internal static string GenerateUserAgentString(Assembly clientAssembly, string?
...
+        /// <returns></returns>
+        private static string EscapeProductInformation(string productInfo)
+        {
+            // If the string is already valid, we don't need to escape anything
+            bool success = false;
+            try
+            {
+                success = ProductInfoHeaderValue.TryParse(productInfo, out var _);
+            }
+            catch (Exception)
```
annelo-msft, line 117: Would it make sense to scope this to specific exception types?

sdk/core/Azure.Core/src/TelemetryDetails.cs
```diff
@@ -80,16 +87,80 @@ internal static string GenerateUserAgentString(Assembly clientAssembly, string?
             {
                 version = version.Substring(0, hashSeparator);
             }
-            var platformInformation = $"({RuntimeInformation.FrameworkDescription}; {RuntimeInformation.OSDescription})";
+            runtimeInformation ??= new RuntimeInformationWrapper();
+            var platformInformation = EscapeProductInformation($"({runtimeInformation.FrameworkDescription}; {runtimeInformation.OSDescription})");
 
-            return applicationId != null
+            var ua = applicationId != null
```
annelo-msft, line 93: Just curious: why?

sdk/core/Azure.Core/tests/HttpPipelineBuilderTest.cs
```diff
@@ -137,7 +137,7 @@ public async Task UsesAssemblyNameAndInformationalVersionForTelemetryPolicySetti
             HttpPipeline pipeline = HttpPipelineBuilder.Build(options);
 
             var message = pipeline.CreateMessage();
-            var userAgent = new TelemetryDetails(typeof(string).Assembly);
+            var userAgent = new TelemetryDetails(typeof(string).Assembly, default);
```
annelo-msft, line 140: Is it required to pass default now? If so, why?

sdk/core/Azure.Core/src/TelemetryDetails.cs
christothes: Didn't mean to leave that in - thanks for catching!

sdk/core/Azure.Core/src/TelemetryDetails.cs
christothes: In this case, I don't think it does. In all cases we want to treat it as a failure to parse correctly.

sdk/core/Azure.Core/src/TelemetryDetails.cs
```diff
@@ -80,16 +87,80 @@ internal static string GenerateUserAgentString(Assembly clientAssembly, string?
...
+            sb.Append('(');
+            // exclude the first and last characters, which are the enclosing parentheses
+            for (int i = 1; i < productInfo.Length - 1; i++)
+            {
+                char c = productInfo[i];
+                if (c == ')' || c == '(')
+                {
+                    sb.Append('\\');
+                }
+                // If we see a \, we don't need to escape it if it's followed by a '\', '(', or ')'
```
annelo-msft, line 137:
```suggestion
                // If we see a \, we don't need to escape it if it's followed by a '\', '(', or ')', because it is already escaped.
```

sdk/core/Azure.Core/tests/HttpPipelineBuilderTest.cs
christothes: Only for internal code - because of the ambiguity of the optional arguments. Client or customer code will not need to do this.

sdk/core/Azure.Core/src/TelemetryDetails.cs
```diff
@@ -80,16 +87,80 @@ internal static string GenerateUserAgentString(Assembly clientAssembly, string?
...
+                }
+                // If we see a \, we don't need to escape it if it's followed by a '\', '(', or ')', because it is already escaped.
+                else if (c == '\\')
+                {
+                    if (i + 1 < (productInfo.Length - 1))
+                    {
+                        char next = productInfo[i + 1];
+                        if (next == '\\' || next == '(' || next == ')')
+                        {
+                            sb.Append(c);
```
annelo-msft, line 145: nit: for consistency.

```suggestion
                            sb.Append(`\\`);
```

christothes: Yes, this is intentional

sdk/core/Azure.Core/src/TelemetryDetails.cs
```diff
@@ -80,16 +87,80 @@ internal static string GenerateUserAgentString(Assembly clientAssembly, string?
...
+                            sb.Append(next);
+                            i++;
+                            continue;
+                        }
+                        else
```
annelo-msft, line 149: nit: totally a style thing so up to you --

you only ever get to this block before the final Append(c) if you want to escape c, so you could remove the two else cases (lines 150-158) and just do sb.Append('\\').  (I think, I haven't tested it.)

But if you like the elses for clarity, that's cool too.  It's my nerdy thing to delete LOC.

sdk/core/Azure.Core/tests/HttpPipelineBuilderTest.cs
annelo-msft: Is it worth changing the constructor definition to avoid this?  There are FDG guidelines that recommend a pattern here, e.g. in section 5.1, "AVOID having two overloads of the same method that both use default parameters" and "DO move all default parameters to the new, longer overload when adding optional parameters to an existing method."  The latter is complicated by the longer overload being internal, but if you have to pass a value anyway if you have access to internal code, why provide a default value for the parameter?

sdk/core/Azure.Core/src/TelemetryDetails.cs
christothes: I'll leave it for now.

sdk/core/Azure.Core/tests/HttpPipelineBuilderTest.cs
christothes: I don't think the guideline applies here because the new parameter should not be exposed in the existing ctor and the existing method should call the internal ctor with the default value to avoid duplicating the code.

sdk/core/Azure.Core/src/TelemetryDetails.cs
```diff
@@ -91,5 +99,67 @@ internal static string GenerateUserAgentString(Assembly clientAssembly, string?
...
+                // Treat this as a failure to parse.
+            }
+            if (success)
+            {
+                return productInfo;
+            }
+
+            var sb = new StringBuilder(productInfo.Length + 2);
+            sb.Append('(');
+            // exclude the first and last characters, which are the enclosing parentheses
```
JoshLove-msft, line 128: But do we know this for sure given that this is an invalid string?

JoshLove-msft: Ah nevermind - these are the parens that we have added ourselves.

christothes: Yes, because we put those enclosing parens there [when calling this](https://github.com/Azure/azure-sdk-for-net/pull/34355/files/1dcaa44b9305801ca1f31dfc2f0f87eb48195569#diff-2b50cdd937e40ec812320781c9d4a8319f24484dde3263c0cd57b4b9428e3525R91)

sdk/core/Azure.Core/tests/TelemetryDetailsTests.cs
```diff
@@ -62,5 +64,50 @@ public void AppliesToMessage()
...
+        [TestCase("ValidParens (2023-)", "ValidParens (2023-)")]
+        [TestCase("ProperlyEscapedParens \\(2023-\\)", "ProperlyEscapedParens \\(2023-\\)")]
+        [TestCase("UnescapedOnlyParens (2023-)", "UnescapedOnlyParens (2023-)")]
+        [TestCase("UnmatchedOpenParen (2023-", "UnmatchedOpenParen \\(2023-")]
+        [TestCase("UnEscapedParenWithValidParens (()", "UnEscapedParenWithValidParens \\(\\(\\)")]
+        [TestCase("UnEscapedInvalidParen (", "UnEscapedInvalidParen \\(")]
+        [TestCase("UnEscapedParenWithValidParens2 ())", "UnEscapedParenWithValidParens2 \\(\\)\\)")]
+        [TestCase("InvalidParen )", "InvalidParen \\)")]
+        [TestCase("UnescapedParenInText MyO)SDescription ", "UnescapedParenInText MyO\\)SDescription ")]
+        [TestCase("UnescapedParenInText MyO(SDescription ", "UnescapedParenInText MyO\\(SDescription ")]
```
JoshLove-msft, line 78: Can we add some test cases where the first char has parens?

JoshLove-msft (approved):

Issue #34064, opened 2023-02-10
[ServiceBus]CanSendLargeMessageBatch(True) failing in nightly runs


Menghua1:
ServiceBus nightly test runs are failing with:

> Error message
> 243
> Expected: True
> But was:  False
> 
> 
> Stack trace
>    at Azure.Messaging.ServiceBus.Tests.Sender.ServiceBusMessageBatchLiveTests.<>c__DisplayClass0_0.<<CanSendLargeMessageBatch>g__AddAndSendMessages|0>d.MoveNext() in /Users/runner/work/1/s/sdk/servicebus/Azure.Messaging.ServiceBus/tests/Sender/ServiceBusMessageBatchLiveTests.cs:line 58
> --- End of stack trace from previous location ---
>    at Azure.Messaging.ServiceBus.Tests.Sender.ServiceBusMessageBatchLiveTests.CanSendLargeMessageBatch(Boolean enableTracing) in /Users/runner/work/1/s/sdk/servicebus/Azure.Messaging.ServiceBus/tests/Sender/ServiceBusMessageBatchLiveTests.cs:line 32
>    at Azure.Messaging.ServiceBus.Tests.Sender.ServiceBusMessageBatchLiveTests.CanSendLargeMessageBatch(Boolean enableTracing) in /Users/runner/work/1/s/sdk/servicebus/Azure.Messaging.ServiceBus/tests/Sender/ServiceBusMessageBatchLiveTests.cs:line 38
>    at Azure.Messaging.ServiceBus.Tests.Sender.ServiceBusMessageBatchLiveTests.CanSendLargeMessageBatch(Boolean enableTracing) in /Users/runner/work/1/s/sdk/servicebus/Azure.Messaging.ServiceBus/tests/Sender/ServiceBusMessageBatchLiveTests.cs:line 68
>    at NUnit.Framework.Internal.TaskAwaitAdapter.GenericAdapter`1.BlockUntilCompleted()
>    at NUnit.Framework.Internal.MessagePumpStrategy.NoMessagePumpStrategy.WaitForCompletion(AwaitAdapter awaiter)
>    at NUnit.Framework.Internal.AsyncToSyncAdapter.Await(Func`1 invoke)
>    at NUnit.Framework.Internal.Commands.TestMethodCommand.RunTestMethod(TestExecutionContext context)
>    at NUnit.Framework.Internal.Commands.TestMethodCommand.Execute(TestExecutionContext context)
[...]

PR #34399, opened 2023-02-21
Don't log receive cancellation as an error

### LINKED ISSUE #34360: ServiceBusTrigger with a ReceiveBatchAsync throws TaskCanceledException when Function App is inactive
### Library name and version

Microsoft.Azure.WebJobs.Extensions.ServiceBus Version=5.8.1

### Describe the bug

We have an Azure Function that listens on a ServiceBusTrigger via a batched trigger. When there is no continuous flow of messages, looks like the Trigger shuts down. But when it does that, it throws a `TaskCanceledException`
The trigger looks like this
```
 [FunctionName("my_function")]
 public async Task Run([ServiceBusTrigger("q.topic.mytopic", "q.subscription.mysubscription", Connection = "ServiceBusConnection", AutoCompleteMessages = false)] ServiceBusReceivedMessage[] messages,            ServiceBusMessageActions messageActions)
```

Full Exception:
```
q.topic.mytopic/Subscriptions/q.subscription.mysubscription-d48eed0c-a31e-42ef-912d-a06edfdc118c: ReceiveBatchAsync Exception: System.Threading.Tasks.TaskCanceledException: A task was canceled.
   at Microsoft.Azure.Amqp.AsyncResult.End[TAsyncResult](IAsyncResult result)
   at Microsoft.Azure.Amqp.ReceivingAmqpLink.EndReceiveMessages(IAsyncResult result, IEnumerable`1& messages)
   at Microsoft.Azure.Amqp.ReceivingAmqpLink.<>c.<ReceiveMessagesAsync>b__27_1(IAsyncResult r)
   at System.Threading.Tasks.TaskFactory`1.FromAsyncCoreLogic(IAsyncResult iar, Func`2 endFunction, Action`1 endAction, Task`1 promise, Boolean requiresSynchronization)
--- End of stack trace from previous location ---
   at Azure.Messaging.ServiceBus.Amqp.AmqpReceiver.ReceiveMessagesAsyncInternal(Int32 maxMessages, Nullable`1 maxWaitTime, TimeSpan timeout, CancellationToken cancellationToken)
   at Azure.Messaging.ServiceBus.Amqp.AmqpReceiver.<>c.<<ReceiveMessagesAsync>b__41_0>d.MoveNext()
--- End of stack trace from previous location ---
   at Azure.Messaging.ServiceBus.ServiceBusRetryPolicy.RunOperation[T1,TResult](Func`4 operation, T1 t1, TransportConnectionScope scope, CancellationToken cancellationToken, Boolean logRetriesAsVerbose)
   at Azure.Messaging.ServiceBus.ServiceBusRetryPolicy.RunOperation[T1,TResult](Func`4 operation, T1 t1, TransportConnectionScope scope, CancellationToken cancellationToken, Boolean logRetriesAsVerbose)
   at Azure.Messaging.ServiceBus.Amqp.AmqpReceiver.ReceiveMessagesAsync(Int32 maxMessages, Nullable`1 maxWaitTime, CancellationToken cancellationToken)
   at Azure.Messaging.ServiceBus.ServiceBusReceiver.ReceiveMessagesAsync(Int32 maxMessages, Nullable`1 maxWaitTime, Boolean isProcessor, CancellationToken cancellationToken).
```
 I see that the same issue was reported by some time back here https://github.com/Azure/azure-sdk-for-net/issues/31508

### Expected behavior

The ServiceBusTriggers gets shutdown without throwing exceptions or ServiceBusTrigger remains active always without shutting down

### Actual behavior

The ServiceBusTriggers gets shutdown with throwing exceptions

### Reproduction Steps

Set up a ServiceBusTrigger with a Batched operation and let it remain idle for a while without any messages. After a while it shuts down along with throwing the above exception

### Environment

Azure Functions latest version
.Net 6

JoshLove-msft:
Fixes https://github.com/Azure/azure-sdk-for-net/issues/34360

Also fix failing test - https://github.com/Azure/azure-sdk-for-net/issues/34064

### DIFF
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Diagnostics/ServiceBusEventSource.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Diagnostics/ServiceBusEventSource.cs
@@ -195,6 +195,7 @@ namespace Azure.Messaging.ServiceBus.Diagnostics
         internal const int ProcessorStoppingCancellationWarningEvent = 113;
 
         internal const int RunOperationExceptionVerboseEvent = 114;
+        internal const int ReceiveMessageCanceledEvent = 115;
 
         #endregion
         // add new event numbers here incrementing from previous
@@ -294,6 +295,15 @@ namespace Azure.Messaging.ServiceBus.Diagnostics
             if (IsEnabled())
             {
                 WriteEvent(ReceiveMessageExceptionEvent, clientId, exception);
+            }
+        }
+
+        [Event(ReceiveMessageCanceledEvent, Level = EventLevel.Verbose, Message = "A receive operation was cancelled. (Identifier '{0}'). Error Message: '{1}'")]
+        public void ReceiveMessageCanceled(string identifier, string exception)
+        {
+            if (IsEnabled())
+            {
+                WriteEvent(ReceiveMessageCanceledEvent, identifier, exception);
             }
         }
 
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/tests/Sender/ServiceBusMessageBatchLiveTests.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/tests/Sender/ServiceBusMessageBatchLiveTests.cs
@@ -54,8 +54,9 @@ namespace Azure.Messaging.ServiceBus.Tests.Sender
                         if (batch.Count < 4500)
                         {
                             var diff = batch.MaxSizeInBytes - batch.SizeInBytes;
-                            // the difference in size from the max allowable size should be less than the size of 1 message
-                            Assert.IsTrue(diff < 220, diff.ToString());
+                            // the difference in size from the max allowable size should be less than the size of a single
+                            // instrumented message
+                            Assert.IsTrue(diff < 250, diff.ToString());
                         }
                         Assert.Greater(batch.Count, 0);
                         await sender.SendMessagesAsync(batch);
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Receiver/ServiceBusReceiver.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Receiver/ServiceBusReceiver.cs
@@ -326,10 +326,13 @@ namespace Azure.Messaging.ServiceBus
                     cancellationToken).ConfigureAwait(false);
             }
             catch (OperationCanceledException ex)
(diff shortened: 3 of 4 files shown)


sdk/servicebus/Azure.Messaging.ServiceBus/tests/Sender/ServiceBusMessageBatchLiveTests.cs
```diff
@@ -54,8 +54,9 @@ async Task AddAndSendMessages()
                         if (batch.Count < 4500)
                         {
                             var diff = batch.MaxSizeInBytes - batch.SizeInBytes;
-                            // the difference in size from the max allowable size should be less than the size of 1 message
-                            Assert.IsTrue(diff < 220, diff.ToString());
```
JoshLove-msft, line 58: This just happened to work for instrumented messages before because the batch very close to the limit when attempting to add the last message. 

With the changes to AMQP conversion, the message size is now smaller as we don't include empty sections. This test started failing because the instrumented message size is 250 bytes.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Diagnostics/ServiceBusEventSource.cs
```diff
@@ -297,6 +298,15 @@ public virtual void ReceiveMessageStart(string identifier, int messageCount)
             }
         }
 
+        [Event(ReceiveMessageCanceledEvent, Level = EventLevel.Verbose, Message = "A receive operation was cancelled. (Identifier '{0}'). Error Message: '{1}'")]
+        public void ReceiveMessageCanceled(string identifier, string exception)
```
jsquire, line 302: Should we consider doing something similar for sending messages too?

jsquire (approved):

m-redding (approved):

sdk/servicebus/Azure.Messaging.ServiceBus/src/Diagnostics/ServiceBusEventSource.cs
JoshLove-msft: I considered this, but being that receiving is the only polling scenario we have, and the only scenario where customers have run into this issue, I didn't want to open up the can of worms of applying this pattern to every other operation. The main reason we need it for receiving is that functions, and potentially other third party libraries, will cancel the long polling operations when shutting down, and this may be transparent to the actual end user. I don't think it is likely that the same issue would come up for other types of operations.

PR #34408, opened 2023-02-21
[HealthcareApis] Add tests

dvbb:


sdk/healthcareapis/Azure.ResourceManager.HealthcareApis/tests/Azure.ResourceManager.HealthcareApis.Tests.csproj
```diff
@@ -1,4 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
+  <ItemGroup>
+    <PackageReference Include="Azure.ResourceManager.EventHubs" VersionOverride ="1.1.0-beta.1"/>
```
ArcturusZhang, line 3: is there a reason that we have to add this version override?

sdk/healthcareapis/Azure.ResourceManager.HealthcareApis/tests/HealthcareApisManagementTestBase.cs
```diff
@@ -30,12 +35,69 @@ public void CreateCommonClient()
...
+            var lro = await workspace.GetFhirServices().CreateOrUpdateAsync(WaitUntil.Completed, fhirServiceName, data);
+            return lro.Value;
+        }
+
+        protected async Task<HealthcareApisIotConnectorResource> CreateHealthcareApisIotConnector(ResourceGroupResource resourceGroup, HealthcareApisWorkspaceResource workspace, string iotConnectorName)
+        {
+            // Create an Eventhub
+            var eventhubNamespace = await resourceGroup.GetEventHubsNamespaces().CreateOrUpdateAsync(WaitUntil.Completed, Recording.GenerateAssetName("azmedtechspaces"), new EventHubsNamespaceData(resourceGroup.Data.Location));
+            string eventHubName = Recording.GenerateAssetName("eventhub");
+            var eventhub = await eventhubNamespace.Value.GetEventHubs().CreateOrUpdateAsync(WaitUntil.Completed, eventHubName, new EventHubData());
```
ArcturusZhang, line 83: We need to exclude the API calls from other RPs from the recording to avoid the case in the future, api-version of event hub upgraded and we have to record the test case in this RP again.

sdk/healthcareapis/Azure.ResourceManager.HealthcareApis/tests/Scenario/IotFhirDestinationTests.cs
```diff
@@ -0,0 +1,91 @@
...
+            var resourceGroup = await CreateResourceGroup();
+            var workspace = await CreateHealthcareApisWorkspace(resourceGroup, Recording.GenerateAssetName("workspace"));
+            var fhirService = await CreateFhirService(workspace, Recording.GenerateAssetName("fhir"));
+            _fhirServiceResourceId = fhirService.Data.Id;
+            var iotConnector = await CreateHealthcareApisIotConnector(resourceGroup, workspace, Recording.GenerateAssetName("medtech"));
+            _fhirDestinationCollection = iotConnector.GetHealthcareApisIotFhirDestinations();
+        }
+
+        [RecordedTest]
+        [Ignore("Pipeline playback error")]
```
ArcturusZhang, line 39: We need to investigate why these tests cannot playback

sdk/healthcareapis/Azure.ResourceManager.HealthcareApis/tests/Azure.ResourceManager.HealthcareApis.Tests.csproj
dvbb: There is no Azure.ResourceManager.EventHubs reference version in the Packages.Data.props file and the reason why not choose 1.0.0 version is because that version will cause [Pipeline issue](https://dev.azure.com/azure-sdk/public/_build/results?buildId=2208424&view=logs&j=bb926d89-89e7-581f-c2a9-0b96e9ca8a72&t=89e0f785-6520-5689-44e6-c45b4430a324&l=2276)

m-nash: We should add it to the packages.data.props in the same location as the other arm packages used by test projects.

sdk/healthcareapis/Azure.ResourceManager.HealthcareApis/tests/HealthcareApisManagementTestBase.cs
dvbb: Added a new rule to UriRegexSanitizers in the constructor of TestBase to ignore event hubs api-version

sdk/healthcareapis/Azure.ResourceManager.HealthcareApis/tests/SessionRecords/DicomServiceTests/AddRemoveTag(False).json
```diff
@@ -0,0 +1,1489 @@
...
+        "state": "Enabled",
+        "subscriptionPolicies": {
+          "locationPlacementId": "Internal_2014-09-01",
+          "quotaId": "Internal_2014-09-01",
+          "spendingLimit": "Off"
+        }
+      }
+    },
+    {
+      "RequestUri": "https://management.azure.com/subscriptions/db1ab6f0-4769-4b27-930e-01e2ef9c123c/resourcegroups/HealthCareApisRG5247?api-version=**",
```
m-nash, line 59: @ArcturusZhang can we look into why some of these are not picking up the change to no longer record dependency setup?

ArthurMa1978 (approved):

PR #34424, opened 2023-02-21
[Event Hubs Extension] Extensions class rename

jsquire:
# Summary

The focus of these changes is to rename the class used for the async collector extensions to something more generic so that it does not resemble an interface itself and can be used to house additional extensions in the future.


m-redding (approved):

KrzysztofCwalina (approved):
I think you need to regenerate the api files, but otherwise looks good.

jsquire:
> I think you need to regenerate the api files, but otherwise looks good.

Yup.   I ran the wrong service directory.  😄

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/api/Microsoft.Azure.WebJobs.Extensions.EventHubs.netstandard2.0.cs
```diff
@@ -8,6 +8,10 @@ public sealed partial class EventHubAttribute : System.Attribute
         public string Connection { get { throw null; } set { } }
         public string EventHubName { get { throw null; } }
     }
+    public static partial class EventHubsWebJobsExtensions
```
JoshLove-msft, line 11: This should probably be EventHubWebJobsExtensions for consistency.

PR #34430, opened 2023-02-21
Fix issue with Max DateTime

JoshLove-msft:


JoshLove-msft:
azp run net - servicebus - tests

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
```diff
@@ -214,6 +222,15 @@ public virtual ServiceBusReceivedMessage AmqpMessageToSBReceivedMessage(AmqpMess
                 annotatedMessage.Header.DeliveryCount = isPeeked ? annotatedMessage.Header.DeliveryCount : annotatedMessage.Header.DeliveryCount + 1;
             }
 
+            if (annotatedMessage.HasSection(AmqpMessageSection.MessageAnnotations) &&
+                annotatedMessage.MessageAnnotations.TryGetValue(AmqpMessageConstants.LockedUntilName, out object lockedUntil))
```
jsquire, line 226: Just to be sure that I'm following:  

```text
IF we have a section 
    AND we have a value 
    AND that value is greater than the offset max 
THEN
    force the value to the offset max
```

So, if we have a value that isn't too large, then we just use it as-is?

jsquire (approved):

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs
JoshLove-msft: Yep

PR #34450, opened 2023-02-22
add parameter to set addtional cadl emitter options

### LINKED ISSUE #34357: [FEATURE REQ] Improve the build script for typesepc/cadl projects
### Library name

build scripts

In our current build scripts, running `dotnet build /t:GenerateCode` command in the typespec/cadl projects will invoke our emitter ([here](https://github.com/Azure/azure-sdk-for-net/blob/3a69051dfc1f70155565b319b675052afddb3be4/eng/common/scripts/Cadl-Project-Generate.ps1#L80)) and the emitter then invokes the generator `Autorest.CSharp.dll` ([here](https://github.com/Azure/autorest.csharp/blob/89baf8397c2c406aae1cadb6551761f1a910fc2b/src/CADL.Extension/Emitter.Csharp/src/emitter.ts#L174)), which might not be the issue. The issue is that this invocation chain is not picking up the version of `Autorest.CSharp.dll` honoring the version inside [`Package.Data.props`](https://github.com/Azure/azure-sdk-for-net/blob/3a69051dfc1f70155565b319b675052afddb3be4/eng/Packages.Data.props#L163) and making things very complicated when we want to regenerate the SDK using a local generator version for those typespec/cadl projects.
We need to improve either the script so that we could easily set the version of the path to the generator, or let the emitter honor the version inside `Package.Data.props`

chunyu3:
Fix https://github.com/Azure/azure-sdk-for-net/issues/34357

- Add parameter to set cadl emitter options


chunyu3:
create a PR in azure-sdk-tools to update this common script. close this one.

PR #34459, opened 2023-02-22
[CognitiveServices] Revert flatten KeyVaultProperties in the Encryption model

### LINKED ISSUE #30987: [CognitiveServices] Dont flatten the `keyVaultProperties` in the `Encryption` model
https://github.com/Azure/azure-sdk-for-net/blob/f83fe390915c8e5df7d63b2bd1cd3a90d9d6f39f/sdk/cognitiveservices/Azure.ResourceManager.CognitiveServices/src/autorest.md?plain=1#L126

Service don't support empty object, so maybe we need to revert the flatten, but need use customization code.

Yao725:
Fix #30987


ArcturusZhang (approved):

sdk/cognitiveservices/Azure.ResourceManager.CognitiveServices/CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 
 ### Bugs Fixed
 
+- Reverted to flattening `KeyVaultProperties` in `ServiceAccountEncryptionProperties` as service side does not support passing empty object for this property.
```
fengzhou-msft, line 11: `Reverted to flattening` is a bit misleading as if the result is to use flattening, `Reverted the falttening of` may be better?

Yao725: Good catch, fixed.

fengzhou-msft (approved):

PR #34481, opened 2023-02-22
Updates OpenAI .NET SDK to allow access to public OpenAI endpoint

glharper:


glharper:
@trrwilson FYI

sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs
```diff
@@ -15,6 +15,35 @@ namespace Azure.AI.OpenAI
...
+        /// <exception cref="ArgumentNullException"> <paramref name="token"/> is null. </exception>
+        public OpenAIClient(string token) : this(token, new OpenAIClientOptions())
+        {
+        }
+
+        /// <summary> Initializes a instance of OpenAIClient using the public OpenAI endpoint. </summary>
+        /// <param name="token"> String token to generate a token credential </param>
+        /// <param name="options"> The options for configuring the client. </param>
+        /// <exception cref="ArgumentNullException"> <paramref name="token"/> is null. </exception>
+        public OpenAIClient(string token, OpenAIClientOptions options)
```
m-nash, line 33: Left a [comment](https://apiview.dev/Assemblies/Review/02ef788afd7e4aa29609e65be5788a4a/9d71fc252c764f66a4e4ab86fcf2f2f3?diffRevisionId=c8effb098bbf46ccbb8fb6c2038e06b6&doc=False&diffOnly=True#dcaef876d76842d790968556caa1710b) about this on the PR apiview as well.  I am not sure this should be on the same client as the azure interface.

@tg-msft @KrzysztofCwalina do you guys have an opinion on splitting the direct openai calls on a separate client from the azure openai calls?  If so what would we call the two?  `AzureOpenAIClient` and `OpenAIClient`?

sdk/openai/Azure.AI.OpenAI/tests/OpenAIInferenceTests.cs
```diff
@@ -26,6 +26,20 @@ public void InstanceTest()
             Assert.That(tokenClient, Is.InstanceOf<OpenAIClient>());
         }
 
+        [Test]
```
m-nash, line 29: You should be able to record this test.  Add a sanitizer for the token and look for it env variable.  Add `PlaybackOnly` attribute to the test so that it won't run live.

sdk/openai/Azure.AI.OpenAI/tests/OpenAITestBase.cs
```diff
@@ -49,6 +49,13 @@ protected OpenAITestBase(bool isAsync, RecordedTestMode? mode = null) : base(isA
         protected OpenAIClient GetClientWithCredential() => InstrumentClient(
             new OpenAIClient(_endpoint, TestEnvironment.Credential, GetInstrumentedClientOptions()));
 
+        protected OpenAIClient GetPublicOpenAIClient() => InstrumentClient(
+            new OpenAIClient("YOUR_TOKEN_STRING_HERE", InstrumentClientOptions(new OpenAIClientOptions(OpenAIClientOptions.ServiceVersion.V2022_12_01)
```
m-nash, line 53: pull the token from Environment.

sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs
trrwilson: As an alternative, could we consider having separate `Options` types for AOAI and OAI while leaving the client and its methods the same?

I agree that pushing the public endpoint capabilities down to the method level doesn't seem ideal, but having entirely separate client objects seems to conflict with the goal of having AOAI be a specialization/value-add atop OAI rather than an entirely parallel product. The differences "feel" more like "how the client is configured" than "what the client is" or "what the client can do," and that maps best to the abstraction of the `Options` type in my mind.

glharper: > As an alternative, could we consider having separate `Options` types for AOAI and OAI while leaving the client and its methods the same?

Unfortunately, given our desire to have as much generated code as possible, and the reality of "branch once, branch all the way up the call stack" when using that code, bifurcating on the Options type accomplishes pretty much the same thing as the code in this PR, unless I'm missing something.

Right now Michael's suggestion seems best, pending further discussion.

m-nash: An alternative could also be that we find a way to not need separate methods.  My main concern is that I don't think a customer would understand they can't use certain methods because they didn't use the right constructor.

glharper: > An alternative could also be that we find a way to not need separate methods. My main concern is that I don't think a customer would understand they can't use certain methods because they didn't use the right constructor.

That's completely fair.

Assuming we were to not use separate constructors (perhaps setting an added Token member on OpenAIClient changes the client to use the public OpenAI endpoint), I'm still not seeing a way around the bifurcation issue, meaning:
- the GetCompletions/GetCompletionsAsync path depends on a few generated methods before reaching CreateGetCompletionsRequest, which is the lowest point of divergence between AzureOpenAI completions and PublicOpenAI completions.

If we could somehow just ignore that generated CreateGetCompletionsRequest method, and use a custom one instead, that would save a ton of work and unify the method signatures. Is there a way to do that?

trrwilson: An important thing to note is that "getting it right" comes first. Keeping as much in generated code as possible is an important consideration for our own maintainability and agility, but we should always start by asking "what's the right thing for people using this?" and never make big compromises on that for our own sakes. We'd never want the honest answer to a question like "this seems confusing, why are there different methods for the different endpoints?" to end up as something like "because it would've been inconvenient for us to make it less confusing."

We already have a few pieces of code where we've effectively removed/stubbed the generated portion and replaced it  (hopefully only as short-term stopgaps) with modified custom code that started as copy/paste -- that's not an unreasonable direction for us to begin with here if that's what's holding us back (particularly as long as we're clearly articulating the feature needs for generation/emission).

sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs
```diff
@@ -15,13 +15,37 @@ namespace Azure.AI.OpenAI
...
+        private const string publicOpenAIVersion = "1";
+        private const string publicOpenAIEndpoint = $"https://api.openai.com/v{publicOpenAIVersion}";
+
+        private TokenCredential _tokenCredential;
+        private HttpPipeline _pipeline;
+        private Uri _endpoint;
+
+        private string _publicOpenAIToken = string.Empty;
+
+        public string PublicOpenAIToken
```
m-nash, line 27: I don't think this needs to be public does it?

sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs
m-nash: Now that we don't have separate methods that work based on which ctor that is used we don't need multiple clients.

sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs
```diff
@@ -35,12 +59,37 @@ public virtual async Task<Response<Completions>> GetCompletionsAsync(string depl
...
         {
-            Argument.AssertNotNullOrEmpty(deploymentId, nameof(deploymentId));
             Argument.AssertNotNullOrEmpty(prompt, nameof(prompt));
-
             CompletionsOptions completionsOptions = new CompletionsOptions();
             completionsOptions.Prompt.Add(prompt);
             return GetCompletions(deploymentId, completionsOptions, cancellationToken);
         }
+
+        internal HttpMessage CreateGetCompletionsRequest(string deploymentId, RequestContent content, RequestContext context)
```
m-nash, line 68: Do we need to also make this customization for the embeddings and streaming APIs?

We should avoid having a client that have methods that don't work simply due to which constructor you used.

sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs
trrwilson: If I'm reading correctly, it does need to be public because it's how the customer using the API would configure the client into "non-AOAI public endpoint mode." It could be an init-only setter if we desired (C# version permitting) but this is effectively the light switch.

glharper: Travis is correct, this is the light switch for setting Client to use the public OpenAI endpoint instead of Azure.

sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs
glharper: Yes, I will add those as well.

sdk/openai/Azure.AI.OpenAI/api/Azure.AI.OpenAI.netstandard2.0.cs
```diff
@@ -82,6 +82,8 @@ public partial class EmbeddingsUsage
     public partial class OpenAIClient
     {
         protected OpenAIClient() { }
+        public OpenAIClient(string token) { }
```
trrwilson, line 85: (Note: this can come later, far from blocking!) along the same lines as "should we make a typed wrapper for this token to make it clear it's not *Azure* OpenAI," should we give the parameter name a bit more detail at this stage?

Using OpenAI's docs as inspiration:
https://platform.openai.com/docs/api-reference/authentication

It could be something like "nonAzureApiKey," "openAiApiKey," "externalApiKey," etc.; I think we for sure want to use "key" rather than "token" in there.

sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs
```diff
@@ -17,13 +17,33 @@ namespace Azure.AI.OpenAI
...
     public partial class OpenAIClient
     {
+        private const string publicOpenAIVersion = "1";
+        private const string publicOpenAIEndpoint = $"https://api.openai.com/v{publicOpenAIVersion}";
+
+        private string PublicOpenAIToken { get; } = "";
+
+        /// <summary> Initializes a instance of OpenAIClient using the public OpenAI endpoint. </summary>
+        /// <param name="token"> A credential used to authenticate to an Azure Service. </param>
+        /// <exception cref="ArgumentNullException"> <paramref name="token"/> is null. </exception>
```
trrwilson, line 27: Per the previous comment, I'm (perhaps excessively) paranoid about averting the potential confusion between AOAI AzureKeyCredential and this non-Azure OpenAI API key. Can we add a remark to help? Like:

```
        <remarks>
            This key is used to connect to a non-Azure OpenAI resource created directly with OpenAI. For Azure OpenAI resources as created and maintained in Azure Portal, use a constructor that provides an Azure resource endpoint and credential, instead.
        </remarks>
```

sdk/openai/Azure.AI.OpenAI/api/Azure.AI.OpenAI.netstandard2.0.cs
tg-msft: In addition to allowing this be rolled, should it also be automatically refreshable?  See https://apiview.dev/Assemblies/Review/6badd6f3e0194ff2a17a3da02ee45dcd#Azure.Communication.CommunicationTokenCredential and https://apiview.dev/Assemblies/Review/6badd6f3e0194ff2a17a3da02ee45dcd#Azure.Communication.CommunicationTokenRefreshOptions for prior art.

tg-msft: And maybe `openAIAuthToken` as a stop gap name-only improvement to make it clearer?

sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs
tg-msft: I would love to understand this better - maybe it's something we could bring to another Arch Board review sooner than later?

sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs
glharper: remark added

trrwilson (approved):

sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs
```diff
@@ -17,13 +17,36 @@ namespace Azure.AI.OpenAI
...
+            PublicOpenAIToken = openAIAuthToken;
+        }
+
         /// <summary> Return the completion for a given prompt. </summary>
         /// <param name="deploymentId"> Deployment id (also known as model name) to use for operations </param>
         /// <param name="prompt"> Input string prompt to create a prompt completion from a deployment. </param>
         /// <param name="cancellationToken"> The cancellation token to use. </param>
         public virtual async Task<Response<Completions>> GetCompletionsAsync(string deploymentId, string prompt, CancellationToken cancellationToken = default)
         {
-            Argument.AssertNotNullOrEmpty(deploymentId, nameof(deploymentId));
```
m-nash, line 26: I assume we are removing this because its not needed if you construct with an openai auth token?  This is odd to have a required param and force customers to pass null.  I wonder if we should pull in the change to move deploymentid to a client level parameter and require it on the non openai auth token ctors so that this is more straight forward?

glharper: I think pulling that change into main would be reasonable, as it would certainly make usage here more straightforward.

trrwilson: Agreed. It's not great that we have a "garbage" parameter. With a client-level default available, I think the overall decision flow would be something like:
`GetCompletions(string deploymentId, ...)` 
- Throws if the client was configured using non-Azure OpenAI information (API key)
- For clients properly configured to use Azure OpenAI, uses the method-provided deployment ID irrespective of what client-level default was provided
- Not null or empty stays enforced

`GetCompletions(string prompt / CompletionsOptions ...) // no deployment ID`
- Throws if the client was configured for Azure OpenAI *and* no client-level default deployment was provided
  - It'd be nice to someday have default deployment selection be an option, so maybe we could be flexible here
- Ignores the concept of deployment entirely for clients configured for non-Azure OpenAI (correctly)
- Uses the client-level default deployment ID for Azure OpenAI

I'm not sure we should wholesale *move* deployment ID specification to the client, though (vs. having a default setting) -- that could end up pretty clunky to use for scenarios where you want to use multiple models (or, if we don't provide affordances, multiple capabilities between completions/embeddings/chat/etc.)

sdk/openai/Azure.AI.OpenAI/README.md
```diff
@@ -39,7 +39,7 @@ Once you have the value of the endpoint string and subscription key, you can cre
 // Replace with your Azure OpenAI key
 string key = "YOUR_AZURE_OPENAI_KEY";
 string endpoint = "https://myaccount.openai.azure.com/";
-OpenAIClient client = new OpenAIClient(new Uri(endpoint), new AzureKeyCredential(key));
+OpenAIClient client = new OpenAIClient(new Uri(endpoint), "myModelDeploymentId", new AzureKeyCredential(key));
```
trrwilson, line 42: This definitely does not and should not hold up the PR in any way, it's a thing for followup consideration: given our scenarios (embeddings, completion, likely chat completions) have specific and mutually exclusive model/deployment requirements, this will effectively lock an instantiated client into use with a single scenario, correct?

If so, I think we need to (after this PR, still!) consider either per-scenario deployment specification (e.g. as distinct fields on the client) or per-scenario *clients* (CompletionsClient, EmbeddingsClient, etc.).

sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs
```diff
@@ -17,40 +17,256 @@ namespace Azure.AI.OpenAI
     /// <summary> Azure OpenAI APIs for completions and search. </summary>
     public partial class OpenAIClient
     {
+        private const string publicOpenAIVersion = "1";
+        private const string publicOpenAIEndpoint = $"https://api.openai.com/v{publicOpenAIVersion}";
```
trrwilson, line 21: For consideration later: I wonder if we should parallel the way you specify service version in OpenAIClientOptions (for AOAI REST revision) to deal with future version revs.

sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs
```diff
@@ -17,40 +17,256 @@ namespace Azure.AI.OpenAI
...
+            }
+        }
+
+        /// <summary> Return the completions for a given prompt. </summary>
+        /// <param name="content"> The content to send as the body of the request. Details of the request body schema are in the Remarks section below. </param>
+        /// <param name="context"> The request context, which can override default behaviors of the client pipeline on a per-call basis. </param>
+        /// <exception cref="RequestFailedException"> Service returned a non-success status code. </exception>
+        /// <returns> The response returned from the service. Details of the response body schema are in the Remarks section below. </returns>
+        /// <include file="../Generated/Docs/OpenAIClient.xml" path="doc/members/member[@name='GetCompletions(String,RequestContent,RequestContext)']/*" />
+        public virtual Response GetCompletions(RequestContent content, RequestContext context = null)
```
trrwilson, line 248: I don't believe these are new (and thus it's yet another "later" thing), but when/why would a customer use these overloads with manual Response, RequestContent and RequestContext? It's a question for the larger audience, I'd imagine, but it feel strange to me that we have this capability exposed as a peer/sibling to the typed ones.

glharper: One the underlying typespec gets updated to remove deploymentId from every completions/embeddings method, this method will be autogenerated again as a public virtual. @jpalvarezl Is this just the way the emitter works, making this middle layer method public?

sdk/openai/Azure.AI.OpenAI/tests/OpenAIInferenceTests.cs
```diff
@@ -29,6 +30,46 @@ public void InstanceTest()
...
 
+        [RecordedTest]
+        [PlaybackOnly("Used only to test against non-Azure OpenAI public endpoint")]
+        public async Task PublicOpenAIEmbeddingsTest()
+        {
+            var client = GetPublicOpenAIClient();
+            Assert.That(client, Is.InstanceOf<OpenAIClient>());
+            EmbeddingsOptions embeddingsRequest = new EmbeddingsOptions("Your text string goes here")
+            {
+                Model = "text-embedding-ada-002"
```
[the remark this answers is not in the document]
glharper: I will keep this in mind for my next PR.

sdk/openai/Azure.AI.OpenAI/README.md
glharper: Agreed, I'd like feedback from @m-nash, @tg-msft, and @johanste, as well as the Speech stakeholders (you, @jpalvarezl, gerardo), but I think a shim scenario-specific client could definitely make sense and ease the cognitive burden on users.

sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs
```diff
@@ -17,40 +17,256 @@ namespace Azure.AI.OpenAI
     /// <summary> Azure OpenAI APIs for completions and search. </summary>
     public partial class OpenAIClient
     {
+        private const string publicOpenAIVersion = "1";
+        private const string publicOpenAIEndpoint = $"https://api.openai.com/v{publicOpenAIVersion}";
+
+        /// <remarks> This is the deployment id for all Azure OpenAI operations (completions and embeddings) </remarks>
+        private string DeploymentId { get; } = "";
```
joseharriaga, line 24: In .NET, when a string does not have a value, it is generally preferred to return `null` rather than an empty string.

sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs
```diff
@@ -17,40 +17,256 @@ namespace Azure.AI.OpenAI
...
+        /// <remarks> This is the deployment id for all Azure OpenAI operations (completions and embeddings) </remarks>
+        private string DeploymentId { get; }
+
+        /// <remarks> This key is used to connect to a non-Azure OpenAI resource created directly with OpenAI.
+        /// For Azure OpenAI resources as created and maintained in Azure Portal, use a constructor that provides
+        /// an Azure resource endpoint and credential, instead. </remarks>
+        private string PublicOpenAIToken { get; }
+
+        /// <summary> Initializes a instance of OpenAIClient using the public OpenAI endpoint. </summary>
+        /// <param name="openAIAuthToken "> A credential used to authenticate to an Azure Service. </param>
```
joseharriaga, line 32: The `<param>` says "A credential used to authenticate to an Azure Service", but my understanding is that this constructor is actually for non-Azure scenarios. Does this need to be changed?

sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs
```diff
@@ -17,40 +17,256 @@ namespace Azure.AI.OpenAI
     /// <summary> Azure OpenAI APIs for completions and search. </summary>
     public partial class OpenAIClient
     {
+        private const string publicOpenAIVersion = "1";
+        private const string publicOpenAIEndpoint = $"https://api.openai.com/v{publicOpenAIVersion}";
+
+        /// <remarks> This is the deployment id for all Azure OpenAI operations (completions and embeddings) </remarks>
+        private string DeploymentId { get; }
+
+        /// <remarks> This key is used to connect to a non-Azure OpenAI resource created directly with OpenAI.
```
joseharriaga, line 26: nit: These `<remarks>` are on a private member, which means they won't be visible to users via the public documentation or Intellisense, but it sounds like it could be very helpful if they were. What do you think about moving these `<remarks>` to the two constructors that take an `openAIAuthToken`?

sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs
```diff
@@ -17,40 +17,256 @@ namespace Azure.AI.OpenAI
...
+        /// Supported Cognitive Services endpoints (protocol and hostname, for example:
+        /// https://westus.api.cognitive.microsoft.com).
+        /// </param>
+        /// <param name="deploymentId"> The deployment id for operations initiated using this client instance </param>
+        /// <param name="credential"> A credential used to authenticate to an Azure Service. </param>
+        /// <param name="options"> The options for configuring the client. </param>
+        /// <exception cref="ArgumentNullException"> <paramref name="endpoint"/> or <paramref name="credential"/> is null. </exception>
+        public OpenAIClient(Uri endpoint, string deploymentId, TokenCredential credential, OpenAIClientOptions options) : this(endpoint, credential, options)
+        {
+            DeploymentId = deploymentId;
```
joseharriaga, line 96: Would it be valid if a user passed a null `deploymentId`? If not, then I recommend adding a check like the ones that the generated OpenAIClient does for the `endpoint` and the `credential` (see Generated/OpenAIClient.cs, line 72). Something like this:
```csharp
Argument.AssertNotNull(deploymentId, nameof(deploymentId));
```

And then modify the comment to also include the `deploymentId`:
```csharp
/// <exception cref="ArgumentNullException"> <paramref name="endpoint"/>, <paramref name="deploymentId"/>, or <paramref name="credential"/> is null. </exception>
```

sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs
glharper: Good catch, will update.

sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs
glharper: Agreed, will update.

sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs
```diff
@@ -17,40 +17,256 @@ namespace Azure.AI.OpenAI
...
+        }
+
         /// <summary> Return the completion for a given prompt. </summary>
-        /// <param name="deploymentId"> Deployment id (also known as model name) to use for operations </param>
         /// <param name="prompt"> Input string prompt to create a prompt completion from a deployment. </param>
         /// <param name="cancellationToken"> The cancellation token to use. </param>
```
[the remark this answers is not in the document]
trrwilson: That's a great catch and it looks like we holistically need a csxml sweep of our custom code -- e.g. the streaming methods I added are entirely missing docs.

I'd say we could (and likely should) promote that audit/cleanup as its own PR vs. addressing it in this one (though "getting it right" for new methods is still good).

sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs
```diff
@@ -17,40 +17,264 @@ namespace Azure.AI.OpenAI
...
+        /// Supported Cognitive Services endpoints (protocol and hostname, for example:
+        /// https://westus.api.cognitive.microsoft.com).
+        /// </param>
+        /// <param name="deploymentId"> The deployment id for all operations (completions and embeddings) initiated using this client instance </param>
+        /// <param name="credential"> A credential used to authenticate to an Azure Service. </param>
+        /// <param name="options"> The options for configuring the client. </param>
+        /// <exception cref="ArgumentNullException"> <paramref name="endpoint"/> or <paramref name="credential"/> or <paramref name="deploymentId"/> is null. </exception>
+        public OpenAIClient(Uri endpoint, string deploymentId, TokenCredential credential, OpenAIClientOptions options) : this(endpoint, credential, options)
+        {
+            Argument.AssertNotNull(endpoint, nameof(endpoint));
```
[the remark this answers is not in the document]
glharper: Good catch, updated.

sdk/openai/Azure.AI.OpenAI/tests/OpenAITestBase.cs
```diff
@@ -37,17 +38,27 @@ private static class Constants
...
         private string _completionsDeploymentId;
         private string _embeddingsDeploymentId;
+        private string _openAIAuthToken;
 
         protected OpenAITestBase(bool isAsync, RecordedTestMode? mode = null) : base(isAsync, mode)
         {
             HeaderRegexSanitizers.Add(new Core.TestFramework.Models.HeaderRegexSanitizer("api-key", "***********"));
         }
 
         protected OpenAIClient GetClient() => InstrumentClient(
```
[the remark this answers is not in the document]
glharper: Updated.

sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs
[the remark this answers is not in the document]
glharper: Updated.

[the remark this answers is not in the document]
glharper: Exception line added to previously documented methods

trrwilson (approved):

sdk/openai/Azure.AI.OpenAI/src/Custom/OpenAIClient.cs
```diff
@@ -17,40 +17,264 @@ namespace Azure.AI.OpenAI
...
+        }
+
+        /// <summary> Initializes a instance of OpenAIClient using the public OpenAI endpoint. </summary>
+        /// <param name="openAIAuthToken "> String token to generate a token credential.
+        /// This string is used to connect to a non-Azure OpenAI resource created directly with OpenAI.
+        /// For Azure OpenAI resources as created and maintained in Azure Portal, use a constructor that provides
+        /// an Azure resource endpoint and credential, instead. </param>
+        /// <param name="options"> The options for configuring the client. </param>
+        /// <exception cref="ArgumentNullException"> <paramref name="openAIAuthToken"/> is null. </exception>
+        public OpenAIClient(string openAIAuthToken, OpenAIClientOptions options) : this(new Uri(publicOpenAIEndpoint), "", CreateDelegatedToken(openAIAuthToken), options)
```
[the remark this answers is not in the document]
joseharriaga: Ah, I see! Then, in that case, I would recommend passing `null` instead of an empty string. That being said, it's okay to keep it like this for now since the `DeploymentId` is currently private anyway, and we can wait until the TypeSpec is updated to revisit this constructor.

joseharriaga (approved):

PR #34485, opened 2023-02-22
Removing RunApiCompat conditions for net 6.0

### LINKED ISSUE #32508: Enable ApiCompat in Azure.Storage for net 6.0
After `Azure.Storage` is released for net 6.0, we need to remove `<RunApiCompat Condition="'$(TargetFramework)' == 'net6.0'">false</RunApiCompat>` from the following projects:
- Azure.Storage.Blobs.csproj
- Azure.Storage.Blobs.Batch.csproj
- Azure.Storage.Blobs.ChangeFeed.csproj
- Azure.Storage.Common.csproj
- Azure.Storage.Files.DataLake.csproj
- Azure.Storage.Files.Shares.csproj
- Azure.Storage.Internal.Avro.csproj
- Azure.Storage.Queues.csproj

amnguye:
Yesterday we GAed the storage packages which all now point to the Azure.Core which was released for net 6.0

Resolves https://github.com/Azure/azure-sdk-for-net/issues/32508

Also stops those unnecessary net 6.0 api files from being generated every time.


AlexanderSher (changes requested):
net 6.0 assemblies aren't unnecessary:
- In storage, there are code paths that are available only in .net 6.0+, for example:
https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/storage/Azure.Storage.Common/src/StorageSharedKeyCredential.cs#L70-L74
-  Azure.Core also has codepaths that are specific to net 6.0+

amnguye:
> net 6.0 assemblies aren't unnecessary:
> 
> * In storage, there are code paths that are available only in .net 6.0+, for example:
>   https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/storage/Azure.Storage.Common/src/StorageSharedKeyCredential.cs#L70-L74
> * Azure.Core also has codepaths that are specific to net 6.0+

Honestly this explains a lot. I was wondering why for months we suddenly started generating these APIs for NET 6.0, but there was never a PR to create the APIs for NET 6.0. Actually up until yesterday, we never made a PR to generate the API files for NET 6.0, because we didn't want to muck anything up, and they started generating under our feet.

I even made an internal thread to figure out why but I didn't gain any traction on the reasoning, but I'm glad I know now!

I'm going to close the PR in preference to the PR you put out.

PR #34486, opened 2023-02-22
Azure OpenAI: fix streaming completions deserialization

trrwilson:
Manual, live testing uncovered that `GetStreamingCompletions` has regressed. This change fixes two problems uncovered in the course of investigating:

1. Last week, streamed Completions responses did not include `usage` in their `data` messages and we thus did not attempt deserialization of `CompletionsUsage`. This week, service responses include `"usage":null` in the same payload; intended or not, this breaks deserialization (`EnumerateObject` on a null `.Value`). A code generation fix will come soon that addresses this via the same approach in this PR, but in the interim a quick move to `Custom` and insertion of the requisite null check/return unblocks streaming use of the SDK.

2. A timing issue with one of the concurrently accessed collections and the SSE message pump task could cause the deserialization to advance past the end index of the list before marked as "done," which in turn resulted in the loop continuing forever. Softening the index check (`>=` instead of `==`) appropriately hardens this with no significant side effects.

Included in the change: new test recordings to reflect the payload difference; release staging to publish the unblocking package update ASAP.


m-nash (approved):

m-nash:
@archerzz when you [regen your change](https://github.com/Azure/autorest.csharp/pull/3163) can you undo this custom code here.

PR #34489, opened 2023-02-23
[Azure.Monitor.Query] Fix pipeline failure

nisha-bhatia:


sdk/monitor/Azure.Monitor.Query/tests/LogsTestData.cs
```diff
@@ -42,7 +42,7 @@ public class LogsTestData
 
         private string TableANameSent => nameof(TableA) + DataVersion + "_" + RetentionWindowStart.DayOfYear;
         public string TableAName => TableANameSent + "_CL";
-        public QueryTimeRange DataTimeRange => new QueryTimeRange(RetentionWindowStart, TimeSpan.FromDays(7));
+        public QueryTimeRange DataTimeRange => new QueryTimeRange(RetentionWindowStart, TimeSpan.FromDays(14));
```
nisha-bhatia, line 45: This range is set to calculate 7 days from the last Monday which doesn't include today's date. Expanding this range to include today's date in the calculation

pallavit: Looking at the failure it seems you need to rerecord your tests.

nisha-bhatia: I think we do need to re-record some tests but the live test pipeline is failing due to another issue.

pallavit: Yeah but we will not get to test the live test pipelines with these changes until we either check these changes in or trigger the live test manually against your private branch.

pallavit: It might be best to do both actually. Re-record the tests so the CI pipelines get green and trigger the live test pipelines to test your changes against live pipelines.

sdk/monitor/Azure.Monitor.Query/tests/LogsTestData.cs
```diff
@@ -51,7 +51,8 @@ public LogsTestData(RecordedTestBase<MonitorQueryTestEnvironment> test)
             _testEnvironment = test.TestEnvironment;
 
             var recordingUtcNow = DateTime.SpecifyKind(test.Recording.UtcNow.Date, DateTimeKind.Utc);
-            RetentionWindowStart = recordingUtcNow.AddDays(DayOfWeek.Monday - recordingUtcNow.DayOfWeek - 7);
+            //DateTime dateTime = DateTime.UtcNow.AddDays(-14);
```
JoshLove-msft, line 54:
```suggestion
```

JoshLove-msft (approved):

PR #34492, opened 2023-02-23
Add OTel support for Service Bus

JoshLove-msft:
Contributes to https://github.com/Azure/azure-sdk-for-net/issues/31650


sdk/core/Azure.Core.TestFramework/src/TestActivitySourceListener.cs
```diff
@@ -16,11 +17,11 @@ public class TestActivitySourceListener: IDisposable
         public Queue<Activity> Activities { get; } =
             new Queue<Activity>();
 
-        public TestActivitySourceListener(string name) : this(source => source.Name == name)
+        public TestActivitySourceListener(string name, Action<Activity> activityStartedCallback = default) : this(source => source.Name.StartsWith(name), activityStartedCallback)
```
JoshLove-msft, line 20: StartsWith aligns better with the usage for activity source where wildcards are used so updating the test class, e.g.:
```c#
.AddSource("Azure.*") // Collect all traces from Azure SDK
```

sdk/core/Azure.Core/src/Shared/MessagingClientDiagnostics.cs
```diff
@@ -0,0 +1,136 @@
...
+
+namespace Azure.Core.Shared
+{
+    /// <summary>
+    /// Client Diagnostics support for messaging clients. Currently, this is only used for AMQP clients.
+    /// HTTP libraries should use the ClientDiagnostics type instead.
+    /// </summary>
+    internal class MessagingClientDiagnostics
+    {
+        private readonly string _fullyQualifiedNamespace;
```
jsquire, line 18: Is `fullyQualifiedNamespace` a bit too "EventHubs/ServiceBus" for a generic thinger?

sdk/core/Azure.Core/src/Shared/MessagingClientDiagnostics.cs
```diff
@@ -0,0 +1,136 @@
...
+            {
+                id = stringId;
+                return true;
+            }
+
+            return false;
+        }
+
+        /// <summary>
+        /// Instrument the message properties for tracing.
```
jsquire, line 114: Should we consider adding a note here that this does modify the message properties and increase the size?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Diagnostics/DiagnosticExtensions.cs
```diff
@@ -24,7 +25,7 @@ public static void SetMessageData(this DiagnosticScope scope, ServiceBusReceived
 
         public static void SetMessageAsParent(this DiagnosticScope scope, ServiceBusReceivedMessage message)
         {
-            if (EntityScopeFactory.TryExtractDiagnosticId(
+            if (MessagingClientDiagnostics.TryExtractDiagnosticId(
```
jsquire, line 28: This is a nice change; seems like a more natural fit for a class name.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Diagnostics/DiagnosticExtensions.cs
```diff
@@ -113,7 +129,7 @@ private static void AddLinkedDiagnostics(this DiagnosticScope scope, IReadOnlyDi
 
         private static void AddLinkedDiagnostics(this DiagnosticScope scope, PropertiesMap properties)
```
jsquire, line 130: What's the scenario where we'd be operating on a `Microsoft.Azure.Amqp.AmqpMessage` instance?   I thought we did all the diagnostic manipulation at the model level.

sdk/servicebus/Azure.Messaging.ServiceBus/src/Diagnostics/MessagingDiagnosticOperation.cs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Azure.Core.Shared;
+
+internal readonly partial struct MessagingDiagnosticOperation
```
jsquire, line 6: Clever use of the partial.   I like the pattern.

jsquire (approved):

sdk/core/Azure.Core.TestFramework/src/TestActivitySourceListener.cs
```diff
@@ -46,6 +48,13 @@ public TestActivitySourceListener(Func<ActivitySource, bool> sourceSelector)
             ActivitySource.AddActivityListener(_listener);
         }
 
+        public Activity AssertAndRemoveActivity(string name)
```
christothes, line 51: nit: 
```suggestion
        public Activity RemoveActivityAndAssert(string name)
```

sdk/core/Azure.Core/src/Shared/MessagingClientDiagnostics.cs
```diff
@@ -0,0 +1,136 @@
...
+
+        #region legacy compat attributes
+        public const string MessageBusDestination = "message_bus.destination";
+        public const string PeerAddress = "peer.address";
+        public const string Component = "component";
+        #endregion
+
+        public const string DiagnosticIdAttribute = "Diagnostic-Id";
+
+        public MessagingClientDiagnostics(string clientNamespace, string? resourceProviderNamespace, string messagingSystem, string fullyQualifiedNamespace, string entityPath)
```
christothes, line 40: Do any of these args need to be validated for null?

sdk/servicebus/Azure.Messaging.ServiceBus/src/Diagnostics/DiagnosticExtensions.cs
JoshLove-msft: We do, but with the changes to cache the AmqpMessage when using a batch, we needed a way to extract the diagnostic Id from the message in order to add the link to the sender activity before sending.

sdk/core/Azure.Core.TestFramework/src/TestActivitySourceListener.cs
JoshLove-msft: This name actually matches the AssertAndRemoveScope method in ClientDiagnosticListener.

sdk/core/Azure.Core/src/Shared/MessagingClientDiagnostics.cs
JoshLove-msft: I think since this is currently only for AMQP it is okay. If/when we extend this for REST-based libraries we can revisit.

sdk/core/Azure.Core/src/Shared/MessagingClientDiagnostics.cs
JoshLove-msft: Since this is internal it seems like it is okay to skip validation.

sdk/core/Azure.Core/src/Shared/MessagingClientDiagnostics.cs
```diff
@@ -0,0 +1,141 @@
...
+        /// <param name="properties">The dictionary of application message properties.</param>
+        /// <param name="activityName">The activity name to use for the diagnostic scope.</param>
+        public void InstrumentMessage(IDictionary<string, object> properties, string activityName)
+        {
+            if (!properties.ContainsKey(DiagnosticIdAttribute))
+            {
+                using DiagnosticScope messageScope = CreateScope(
+                    activityName,
+                    DiagnosticScope.ActivityKind.Producer,
+                    MessagingDiagnosticOperation.Publish);
```
lmolkova, line 130: let's not set operation name yet

sdk/servicebus/Azure.Messaging.ServiceBus/src/Diagnostics/MessagingDiagnosticOperation.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    public static MessagingDiagnosticOperation RenewMessageLock = new("renew_message_lock");
```
lmolkova, line 8: based on offline discussion, I'm suggesting to remove all but `settle` for now - we might return them later on. 

If you came up with a good scenario for them that can help users or us to debug problems, I'm totally on board to add them now.

lmolkova (approved):

sdk/servicebus/Azure.Messaging.ServiceBus/src/Diagnostics/MessagingDiagnosticOperation.cs
JoshLove-msft: I removed all of the other settlement methods, but things like renew message lock are not settlement. Should we just leave these blank?

sdk/core/Azure.Core/src/Shared/MessagingClientDiagnostics.cs
```diff
@@ -73,44 +79,72 @@ public MessagingClientDiagnostics(string clientNamespace, string? resourceProvid
...
         /// <returns><c>true</c> if the message properties contained the diagnostic id; otherwise, <c>false</c>.</returns>
-        public static bool TryExtractDiagnosticId(IReadOnlyDictionary<string, object> properties, out string? id)
+        public static bool TryExtractTraceContext(IReadOnlyDictionary<string, object> properties, out string? traceparent, out string? tracestate)
         {
-            id = null;
+            traceparent = null;
+            tracestate = null;
 
-            if (properties.TryGetValue(DiagnosticIdAttribute, out var objectId) && objectId is string stringId)
+            if (properties.TryGetValue(DiagnosticIdAttribute, out var diagnosticId) && diagnosticId is string diagnosticIdString)
```
lmolkova, line 94: suggestion:

```csharp
if (properties.TryGetValue(TraceParent, out var traceParent) && traceParent is string traceParentString)
{
    traceparent = traceParentString;
    if (properties.TryGetValue(TraceState, out object state) && state is string stateString)
    {
        tracestate = stateString;
    }
    return true;
}

if (properties.TryGetValue(DiagnosticIdAttribute, out var diagnosticId) && diagnosticId is string diagnosticIdString)
{
    traceparent = diagnosticIdString;
    return true;
}
```

there should never be a tracestate without traceparent and reading it with `Diagnostic-Id` is not necessary.

sdk/core/Azure.Core/src/Shared/MessagingClientDiagnostics.cs
```diff
@@ -73,44 +79,72 @@ public MessagingClientDiagnostics(string clientNamespace, string? resourceProvid
...
-        ///
+        /// <param name="properties">The properties holding the trace context.</param>
+        /// <param name="traceparent">The trace parent of the message.</param>
+        /// <param name="tracestate">The trace state of the message.</param>
         /// <returns><c>true</c> if the message properties contained the diagnostic id; otherwise, <c>false</c>.</returns>
-        public static bool TryExtractDiagnosticId(IDictionary<string, object> properties, out string? id)
+        public static bool TryExtractTraceContext(IDictionary<string, object> properties, out string? traceparent, out string? tracestate)
         {
-            id = null;
+            traceparent = null;
```
lmolkova, line 126: same here

sdk/core/Azure.Core/src/Shared/MessagingClientDiagnostics.cs
```diff
@@ -134,6 +168,8 @@ public void InstrumentMessage(IDictionary<string, object> properties, string act
                 if (activity != null)
                 {
                     properties[DiagnosticIdAttribute] = activity.Id;
+                    properties[TraceParent] = activity.Id;
+                    properties[TraceState] = activity.TraceStateString;
```
lmolkova, line 172: suggestion:
```csharp
if (activity.TraceStateString != null) {
  properties[TraceState] = activity.TraceStateString;
}
```

`activity.TraceStateString` is usually null

lmolkova (approved):

PR #34506, opened 2023-02-23
Linking to additional resources for the sample

udidahan:
Following up on [Azure#34152.](https://github.com/Azure/azure-sdk-for-net/pull/34152)

/cc @jsquire

### DIFF
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/samples/Sample10_ClaimCheck.md
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/samples/Sample10_ClaimCheck.md
@@ -52,3 +52,7 @@ if (receivedMessage.ApplicationProperties.TryGetValue("blob-name", out object bl
     await blobClient.DeleteAsync();
 }
 ```
+
+## Additional Resources
+
+Some 3rd party libraries provide support for this pattern out of the box, including NServiceBus via its [data bus](https://docs.particular.net/samples/azure/blob-storage-databus/) and MassTransit via its [claim check](https://masstransit.io/documentation/patterns/claim-check).


jsquire:
Hi @udidahan.   Apologies for the confusion.  

The discussion in #34152 was not intended to imply that we should include Mass Transit, I meant that as a well-known example of a similar product.  The Azure SDK has a [guideline](https://github.com/Azure/azure-sdk-for-net/blob/main/CONTRIBUTING.md#samples) governing the use of third-party dependencies in Azure SDK samples, which limits doing so to scenarios focused on demonstrating how to use the Azure SDK library:

> ### Third-party dependencies
> Third party libraries should only be included in samples when necessary to demonstrate usage of an Azure SDK package; they should not be suggested or endorsed as alternatives to the Azure SDK.

Unfortunately, that means we're unable to accept this contribution.  As mentioned, the MS Learn site has a looser [guideline](https://review.learn.microsoft.com/help/contribute/contribute-get-started-channel-guidance?branch=main#non-microsoft-products) where non-Microsoft products are concerned and would be better able to include this style of content.

udidahan:
@jsquire I'm not sure I see the conflict with the guideline.

This PR does not introduce any third party libraries into the sample itself.

Also, both NServiceBus and MassTransit use the Azure SDK, not just in general, but for this specific use case.

I could see how this guideline were to apply if the Azure SDK had specific Claim Check functionality itself. Then, the use of 3rd party libraries with the functionality would indeed be "alternatives to the Azure SDK". I just don't see it in this specific case.

This seems to be more similar to the case of the [Transactional Outbox pattern with Azure Cosmos DB article](https://learn.microsoft.com/en-us/azure/architecture/best-practices/transactional-outbox-cosmos) where the sample is showing how a pattern can be implemented with some custom code on top of the underlying Azure SDK. In that example, the maintainers accepted a contribution mentioning NServiceBus and MassTransit into their [Considerations section](https://learn.microsoft.com/en-us/azure/architecture/best-practices/transactional-outbox-cosmos#considerations).

I do hope these explanations allow for you to accept the PR.

Thanks!

jsquire:
Hi @udidahan.   To be transparent, the content of your initial PR was not something that we had planned for and sparked discussion that traveled up through the Azure SDK product owners.  The guideline that I shared was an outcome of that, as was the decision to decline your proposed changes.   I followed-up on that with your revisions and thoughts above, but the outcome remains the same.  We've been asked to decline this contribution (and similar content in the Azure SDK samples).  

Unfortunately, that's where we're at.

udidahan:
OK. I understand.
Thanks for letting me know.

Issue #34472, opened 2023-02-22
[QUERY] Late ThrowIfCancellationRequested check in SendMessagesAsync


BillVolz:
### Library name and version

Azure Service Bus 7.12.0

### Query/Question

I was browsing the code and noticed that this cancellation check (link below) is performed after the operation is complete.  Wouldn't this result in a successful queue of the message and a cancellation as the result?  Then may be cause someone to queue the message twice on a retry? 

https://github.com/Azure/azure-sdk-for-net/blob/Azure.Messaging.ServiceBus_7.12.0/sdk/servicebus/Azure.Messaging.ServiceBus/src/Sender/ServiceBusSender.cs#L255

### Environment

N/A

jsquire:
Hi @BillVolz.  Thanks for your feedback.  That does indeed look like a bug.   We'll get this fixed up.

PR #34515, opened 2023-02-23
[Service Bus] Remove post-op cancellation

jsquire:
# Summary

The focus of these changes is to remove some checks for cancellation that were occurring after a service operation had been completed.  Because the operation was complete, cancellation was not performed and it is more accurate to return the result of the operation.

# References and Related

- [[QUERY] Late ThrowIfCancellationRequested check in SendMessagesAsync (#34472)](https://github.com/Azure/azure-sdk-for-net/issues/34472)

### DIFF
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpSender.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpSender.cs
@@ -457,17 +457,17 @@ namespace Azure.Messaging.ServiceBus.Amqp
                     }
 
                     entries.Add(entry);
+                    cancellationToken.ThrowIfCancellationRequested<TaskCanceledException>();
                 }
 
                 request.Map[ManagementConstants.Properties.Messages] = entries;
 
+                cancellationToken.ThrowIfCancellationRequested<TaskCanceledException>();
                 AmqpResponseMessage amqpResponseMessage = await ManagementUtilities.ExecuteRequestResponseAsync(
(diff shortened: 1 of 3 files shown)


sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpSender.cs
```diff
@@ -457,18 +457,18 @@ public override async Task CloseAsync(CancellationToken cancellationToken)
                     }
 
                     entries.Add(entry);
+                    cancellationToken.ThrowIfCancellationRequested<TaskCanceledException>();
```
jsquire, line 460: This may be overkill; the intent was to save on the cost of allocations and serialization if cancellation occurred while processing messages.

JoshLove-msft: Yeah, I would vote to not include this.

JoshLove-msft (approved):

sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpSender.cs
jsquire: Sorry, missed this comment when closing out.   I'll throw a new PR out to remove it.

jsquire: #34537

PR #34538, opened 2023-02-24
Fix #33474 by changing preprocessor statement

### LINKED ISSUE #33474: [BUG] UserTokenProvider.LoginWithPromptAsync is not available on the desktop framework
### Library name and version

Microsoft.Rest.ClientRuntime.Azure.Authentication 2.3.3+

### Describe the bug

`UserTokenProvider.LoginWithPromptAsync` is not available when targeting .NET 4.6-4.8. It's available on 4.5.

This appears to have been been the case since the `net461` target framework was added in version 2.3.3 of the package back in 2018. The target framework was added, but there's [a preprocessor statement](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/mgmtcommon/Auth/Az.Auth/Az.Authentication/UserLoginProvider.cs#L18) in the code making `LoginWithPromptAsync` available only on `net452`. So when referencing the package from a recent version of the Framework, we get a build that does not include `LoginWithPromptAsync`.

Was this intentional? [The accompanying comment](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/mgmtcommon/Auth/Az.Auth/Az.Authentication/UserLoginProvider.cs#L17) seems to indicate that the intention was to exclude `LoginWithPromptAsync` from the .NET Core/Standard builds of the package, not from any build other than `net452`. The code doesn't appear to have changed substantially since before version 2.3.3, so it appears that perhaps the preprocessor statement was overlooked when the `net461` target was added. I _think_ that preprocessor statement should read `#if NETFRAMEWORK`.

### Expected behavior

`LoginWithPromptAsync` is available on all versions of the .NET Framework, but unavailable when targeting .NET Core/(5+) and .NET Standard.

### Actual behavior

`LoginWithPromptAsync` is available only on .NET 4.5.

### Reproduction Steps

Set up a project targeting .NET 4.6-4.8 (I'm on `net472`), add the `Microsoft.Rest.ClientRuntime.Azure.Authentication` package, and try to call `UserTokenProvider.LoginWithPromptAsync`.

### Environment

_No response_

benjamin-hodgson:
As discussed in the bug, LoginWithPromptAsync is only available on .NET Framework <4.6 right now because the preprocessor statement targets net452 explicitly. I believe that to be a mistake, so I've changed the preprocessor statement to target any version of the Framework.

I also took the liberty of deleting a little bit of commented code.

### DIFF
--- a/sdk/mgmtcommon/Auth/Az.Auth/Az.Authentication/Microsoft.Rest.ClientRuntime.Azure.Authentication.csproj
+++ b/sdk/mgmtcommon/Auth/Az.Auth/Az.Authentication/Microsoft.Rest.ClientRuntime.Azure.Authentication.csproj
@@ -5,12 +5,12 @@
     <Description>Provides ADAL based authentication for Azure management client libraries</Description>
     <AssemblyName>Microsoft.Rest.ClientRuntime.Azure.Authentication</AssemblyName>
     <AssemblyTitle>Authentication for Azure Management Clients</AssemblyTitle>
-    <Version>2.4.1</Version>    
+    <Version>2.4.2</Version>
     <PackageTags>Microsoft AutoRest ClientRuntime Authentication $(NugetCommonTags) $(NugetCommonProfileTags)</PackageTags>
 	<PackageReleaseNotes>
       <![CDATA[
         Fixes:
-        - Relaxed the upped bound for Microsoft.IdentityModel.* dependencies.
+        - Made UserTokenProvider.LoginWithPromptAsync available on all supported .NET Framework versions.
       ]]>
     </PackageReleaseNotes>
   </PropertyGroup>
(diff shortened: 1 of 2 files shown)


benjamin-hodgson:
Some whitespace changes in here that must've been auto-added by my editor. I can undo them if you prefer

jsquire:
Hi @benjamin-hodgson.   Thank you for your contribution.   Please be aware that this package is no longer under active development and is receiving critical bug and security fixes only.   As the code impacted by this change relies on the `Microsoft.IdentityModel.Clients.ActiveDirectory` which has reached end-of-life and is no longer maintained, it seems unlikely that we would be able to include this change in a release.

I've looped in @ArthurMa1978 to review and make the decision on a potential release.

PR #34554, opened 2023-02-25
Install dotnet-trace for perf profiling

AlexanderSher:


eng/pipelines/templates/jobs/perf.yml
```diff
@@ -35,6 +35,13 @@ extends:
     - template: /eng/pipelines/templates/steps/install-dotnet.yml
       parameters:
         EnableNuGetCache: false
+    - ${{ if parameters.Profile }}:
```
weshaggard, line 38: This doesn't look like a declared parameter so I wouldn't expect this to work as is.

AlexanderSher: It is declared here: https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/pipelines/templates/jobs/perf.yml#L59-L61

What do I need to do to pass it to this template?

weshaggard: You will need to specify it as a parameter at the top of this file as well and then make sure it is passed through.

AlexanderSher: Fixed

mikeharder (approved):
The change to the shared `perf.yml` looks good to me.  And we can test this using pipeline `tools - perf-automation - tests`.  However, to actually use this feature in the `net - * - perf` pipelines, you will also need to add the `Profile` parameter to all the `perf.yml` files in the service directories.  You can do this in a separate PR, or add it to this PR.

sdk/monitor/perf.yml
```diff
@@ -19,6 +19,10 @@ parameters:
   displayName: Iterations (times to run each test)
   type: number
   default: '5'
+- name: Profile
+  displayName: Enable profiling
```
mikeharder, line 23:
```suggestion
```

In other languages we omit `displayName` because the parameter name is good enough.  Example:

https://github.com/Azure/azure-sdk-for-java/blob/77036c28abc8715b61fdad3f8e806028a72a3a21/sdk/storage/azure-storage-blob/perf.yml#L26-L28

mikeharder (changes requested):

mikeharder (approved):

PR #34557, opened 2023-02-25
[WebJobs.Extensions.ServiceBus] Scaler APIs

alrod:


sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusScalerProvider.cs
```diff
@@ -0,0 +1,148 @@
...
+                new Lazy<ServiceBusAdministrationClient>(() => adminClient),
+                _options.Value,
+                serviceBusMetadata.IsSessionsEnabled,
+                !string.Equals(serviceBusMetadata.Cardinality, "many"),
+                _loggerFactory);
+        }
+
+        private (ServiceBusAdministrationClient AdminClient, ServiceBusReceiver Receiver) CreateParameters(IConfiguration configuration, ServiceBusMetadata serviceBusMetadata, AzureComponentFactory azureComponentFactory)
+        {
+            ServiceBusClientFactory factory = new ServiceBusClientFactory(configuration, azureComponentFactory, _messagingProvider, _logForwarder, _options);
```
mathewc, line 104: Yes, this shows the drawback of flowing the AzureComponentFactory override as an object property. We basically have to ignore the ServiceBusClientFactory we've registered via DI for normal flows, and recreate a new instance explicitly not via DI. That could get messy since other components are using the default factory - if those components get used for anything they're using the wrong factory.

An alternative we could consider would be to embrace the fact that we're saying the component factory to use may now vary based on context in some cases - i.e. not a single instance for all usages in a host. To make that explicit we could introduce an interface like the following:

```csharp
IAzureComponentFactoryProvider
  AzureComponentFactory GetFactory(AzureComponentFactoryProviderContext context)

AzureComponentFactoryProvoderContext
  string FunctionId
```

- this would live in  the WebJobs Host.Storage package where AddAzureStorageCoreServices lives
- we'd register a default implementation of this that returns the default AzureComponentFactory in all cases regardless of context
- ScaleHosts would provide a custom implementation that returns function specific component factory instances based on context.
- in ServiceBusExtension, ServiceBusClientFactory gets updated to take an IAzureComponentFactoryProvider on construction. I.e. all cases where we're taking AzureComponentFactory via DI we replace with this interface
- There would then be some cascading changes because callers creating a client will need to flow function information down

This has the benefits of avoiding the issues mentioned above, and also means we don't need the Properties object hack on functions metadata for flowing this.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusHostBuilderExtensions.cs
```diff
@@ -93,5 +95,29 @@ public static IWebJobsBuilder AddServiceBus(this IWebJobsBuilder builder, Action
             builder.Services.AddSingleton<ServiceBusClientFactory>();
             return builder;
         }
+
+        public static IWebJobsBuilder AddServiceBusScaleForTrigger(this IWebJobsBuilder builder, TriggerMetadata triggerMetadata)
+        {
+            ServiceBusScalerProvider provider = null; // lazy init
```
mathewc, line 101: A cleaner way to do this can be seen in an example [here](https://github.com/Azure/azure-functions-host/blob/d207a4380e18aa79a7545762bd2f5107f8338893/src/WebJobs.Script/ScriptHostBuilderExtensions.cs#L275) where the Functions Host has a ScriptHost singleton registered for multiple different interface types.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusScalerProvider.cs
```diff
@@ -0,0 +1,124 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Azure.Messaging.ServiceBus;
+using Azure.Messaging.ServiceBus.Administration;
+using Microsoft.Azure.WebJobs.Extensions.ServiceBus.Config;
+using Microsoft.Azure.WebJobs.Host.Scale;
+using Microsoft.Azure.WebJobs.ServiceBus;
```
mathewc, line 8: Fix the using orderings. Update your VS settings?

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusScalerProvider.cs
```diff
@@ -0,0 +1,124 @@
...
+            _messagingProvider = serviceProvider.GetService<MessagingProvider>();
+            _triggerMetadata = triggerMetadata;
+            _logForwarder = serviceProvider.GetService<AzureEventSourceLogForwarder>();
+            _options = serviceProvider.GetService<IOptions<ServiceBusOptions>>();
+            _loggerFactory = serviceProvider.GetService<ILoggerFactory>();
+        }
+
+        public IScaleMonitor GetMonitor()
+        {
+            ServiceBusMetadata serviceBusMetadata = JsonConvert.DeserializeObject<ServiceBusMetadata>(_triggerMetadata.Metadata.ToString());
```
mathewc, line 45: Can avoid the double serialization/deserialization using techniques like https://stackoverflow.com/questions/30220328/how-can-i-populate-an-existing-object-from-a-jtoken-using-newtonsoft-json

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusScalerProvider.cs
```diff
@@ -0,0 +1,124 @@
...
+            AzureComponentFactory azureComponentFactory = GetAzureComponentFactory(_triggerMetadata);
+
+            (ServiceBusAdministrationClient adminClient, ServiceBusReceiver receiver) = CreateParameters(_configuration, serviceBusMetadata, azureComponentFactory);
+
+            return new ServiceBusScaleMonitor(
+                _triggerMetadata.FunctionName,
```
mathewc, line 51: Is this the same ID value as in other places where we create a monitor, e.g. https://github.com/Azure/azure-sdk-for-net/blob/460129360094a5617743024ec3e369055ad624fe/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusListener.cs#L107? I think ID is the full name?

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusScalerProvider.cs
```diff
@@ -0,0 +1,124 @@
...
+                receiver.EntityPath,
+                !string.IsNullOrEmpty(serviceBusMetadata.QueueName) ? ServiceBusEntityType.Queue : ServiceBusEntityType.Topic,
+                new Lazy<ServiceBusReceiver>(() => receiver),
+                new Lazy<ServiceBusAdministrationClient>(() => adminClient),
+                _loggerFactory);
+        }
+
+        public ITargetScaler GetTargetScaler()
+        {
+            ServiceBusMetadata serviceBusMetadata = JsonConvert.DeserializeObject<ServiceBusMetadata>(_triggerMetadata.Metadata.ToString());
```
mathewc, line 61: Can we just parse this once in the ctor so it's not duplicated?

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusScalerProvider.cs
```diff
@@ -0,0 +1,124 @@
...
+
+            ServiceBusAdministrationClient adminClient = factory.CreateAdministrationClient(serviceBusMetadata.Connection);
+            ServiceBusClient client = factory.CreateClientFromSetting(serviceBusMetadata.Connection);
+            ServiceBusReceiver receiver = !string.IsNullOrEmpty(serviceBusMetadata.QueueName) ? client.CreateReceiver(serviceBusMetadata.QueueName)
+                : client.CreateReceiver(serviceBusMetadata.TopicName, serviceBusMetadata.SubscriptionName);
+
+            return (AdminClient: adminClient, Receiver: receiver);
+        }
+
+        private AzureComponentFactory GetAzureComponentFactory(TriggerMetadata triggerMetadata)
```
mathewc, line 90: Why do we need to call this method twice, rather than once in the ctor? Both the existing ctor code and this code is pulling the factory from Properties.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/tests/ScaleHostEndToEndTests.cs
```diff
@@ -0,0 +1,254 @@
...
+                services.AddAzureStorageScaleServices();
+
+                services.AddSingleton<IScaleMetricsRepository, TestScaleMetricsRepository>();
+            })
+            .ConfigureWebJobsScale((context, builder) =>
+            {
+                builder.AddServiceBus();
+                builder.UseHostId(hostId);
+
+                foreach (var jtocken in JObject.Parse(triggers)["triggers"])
```
mathewc, line 110: spelling

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/tests/ScaleHostEndToEndTests.cs
```diff
@@ -0,0 +1,254 @@
...
+                ""azureWebJobs"" : {
+                    ""extensions"": {
+                        ""serviceBus"": {
+                            ""maxConcurrentCalls"" :  1
+                        }
+                    }
+                }
+            }";
+
+            // Function1Name uses conection string
```
mathewc, line 51: spelling

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/tests/ScaleHostEndToEndTests.cs
```diff
@@ -0,0 +1,254 @@
...
+            // Function2Name uses AzureCompoentFactory - simulating managed identity scenario in ScaleController
```
mathewc, line 52: spelling, here and throughout

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusHostBuilderExtensions.cs
alrod: I am not sure how to implement this, note we need to create an instance as we are passing trigger data. Can you please give me an example?

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusScalerProvider.cs
alrod: The file was created before I changes the VS setting.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusScalerProvider.cs
alrod: The value will be different for RDS and SC scenarios:
RDS - Microsoft.Azure.WebJobs.Host.EndToEndTests.ServiceBusEndToEndTests+ServiceBusTestJobs.SBQueue2SBQueue (full name)
SC - SBQueue2SBQueue (from sync trigger content)

I think it's ok as we use the fucntion id as a key for storage base metric repository in `TableStorageScaleMetricsRepository`

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusScalerProvider.cs
alrod: Are you proposing to inherit `ServiceBusMetadata` from `TriggerMetadata` here? We need a new `ctor` for `TriggerMetadata(TriggerMetadata triggerMetadata)` what is also not very clean.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusHostBuilderExtensions.cs
mathewc: Something like this:

```csharp
services.AddSingleton<ServiceBusScalerProvider>(s => new ServiceBusScalerProvider(s, triggerMetadata));
services.AddSingleton<IScaleMonitorProvider>(s => s.GetService<ServiceBusScalerProvider>());
services.AddSingleton<ITargetScalerProvider>(s => s.GetService<ServiceBusScalerProvider>());
```

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusScalerProvider.cs
```diff
@@ -0,0 +1,122 @@
...
+using Microsoft.Extensions.Options;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.WebJobs.Extensions.ServiceBus.Listeners
+{
+    internal class ServiceBusScalerProvider : IScaleMonitorProvider, ITargetScalerProvider
+    {
+        private readonly IConfiguration _configuration;
+        private AzureComponentFactory _defaultAzureComponentFactory;
```
mathewc, line 25: move this down after all the readonly fields

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusScalerProvider.cs
```diff
@@ -0,0 +1,122 @@
...
+        private readonly AzureEventSourceLogForwarder _logForwarder;
+        private readonly TriggerMetadata _triggerMetadata;
+        private readonly ILoggerFactory _loggerFactory;
+        private readonly ServiceBusMetadata _serviceBusMetadata;
+        private readonly AzureComponentFactory _azureComponentFactory;
+
+        public ServiceBusScalerProvider(IServiceProvider serviceProvider, TriggerMetadata triggerMetadata)
+        {
+            _configuration = serviceProvider.GetService<IConfiguration>();
+            _defaultAzureComponentFactory = triggerMetadata.Properties[nameof(AzureComponentFactory)] as AzureComponentFactory;
```
mathewc, line 37: Here and below where we're using values from the user supplied triggerMetadata the various properties can be null I think. I don't think TriggerMetadata does null checks in its ctors - it should.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusScalerProvider.cs
mathewc: Also, I think we can just remove the field entirely. It's only used once in GetAzureComponentFactory and that is called only once in the ctor. So this can just be inlined in that function.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusScalerProvider.cs
mathewc: Also, if the factory isn't in the properties, won't this throw?

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/tests/ScaleHostEndToEndTests.cs
```diff
@@ -0,0 +1,257 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Azure.Core;
+using Azure.Messaging.ServiceBus.Tests;
+using Microsoft.Azure.WebJobs.Host.EndToEndTests;
+using Microsoft.Azure.WebJobs.Host.Scale;
+using Microsoft.Azure.WebJobs.Host.TestCommon;
+using Microsoft.Extensions.Azure;
+using Microsoft.Extensions.Configuration;
```
mathewc, line 10: order these usings properlyl!

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/tests/ScaleHostEndToEndTests.cs
```diff
@@ -0,0 +1,257 @@
...
+            });
+            hostBuilder.ConfigureAppConfiguration((hostBuilderContext, config) =>
+            {
+                // Adding host.json here
+                config.AddJsonStream(new MemoryStream(Encoding.UTF8.GetBytes(hostJson)));
+
+                var settings = new Dictionary<string, string>();
+                settings.Add("AzureWebJobsServiceBus", ServiceBusTestEnvironment.Instance.ServiceBusConnectionString);
+                settings.Add("ServiceBusConnection2:fullyQualifiedNamespace", $"{SecondQueueScope.NamespaceName}.servicebus.windows.net");
+                settings.Add("HostingConfig:Microsoft.Azure.WebJobs.Extensions.ServiceBus", "1");
```
chiangvincent, line 93: App settings is probably not the best place to park the hosting config, as the config is at the stamp level (not app)

alrod: The settings are set for a scale host (one scale host per function app) so it's ok have then in `IConfiguration`

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusHostBuilderExtensions.cs
alrod: This approach does work as `ServiceBusScalerProvider` will be the same for all the triggers - last added by `services.AddSingleton<ServiceBusScalerProvider>` will be used.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusScalerProvider.cs
alrod: yes, the method is no needed.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusHostBuilderExtensions.cs
```diff
@@ -93,5 +95,22 @@ public static IWebJobsBuilder AddServiceBus(this IWebJobsBuilder builder, Action
             builder.Services.AddSingleton<ServiceBusClientFactory>();
             return builder;
         }
+
+        public static IWebJobsBuilder AddServiceBusScaleForTrigger(this IWebJobsBuilder builder, TriggerMetadata triggerMetadata)
+        {
+            ServiceBusScalerProvider provider = null;
```
mathewc, line 101: Something still seems off with this pattern. For example, if the ITargetScalerProvider is requested first before the IScaleMonitorProvider, null would be returned. To avoid race conditions like this, I think you need to define a single factory and reuse in both places.

alrod: Proposed `Lazy<>` implementation

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusScalerProvider.cs
```diff
@@ -0,0 +1,104 @@
...
+            IOptions<ServiceBusOptions> options = serviceProvider.GetService<IOptions<ServiceBusOptions>>();
+            ILoggerFactory loggerFactory = serviceProvider.GetService<ILoggerFactory>();
+            ServiceBusMetadata serviceBusMetadata = JsonConvert.DeserializeObject<ServiceBusMetadata>(triggerMetadata.Metadata.ToString());
+            ServiceBusClientFactory factory = new ServiceBusClientFactory(configuration, azureComponentFactory, messagingProvider, logForwarder, options);
+            ServiceBusAdministrationClient serviceBusAdminClient = factory.CreateAdministrationClient(serviceBusMetadata.Connection);
+            ServiceBusClient serviceBusClient = factory.CreateClientFromSetting(serviceBusMetadata.Connection);
+            ServiceBusReceiver serviceBusReceiver = !string.IsNullOrEmpty(serviceBusMetadata.QueueName) ? serviceBusClient.CreateReceiver(serviceBusMetadata.QueueName)
+            : serviceBusClient.CreateReceiver(serviceBusMetadata.TopicName, serviceBusMetadata.SubscriptionName);
+
+            _scaleMonitor = new ServiceBusScaleMonitor(
```
pragnagopa, line 50: Should this removed - https://github.com/Azure/azure-sdk-for-net/blob/aa1ef2a332d38a95e7c7fd39a7c8659754b63f19/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusListener.cs#L105

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusScalerProvider.cs
```diff
@@ -0,0 +1,104 @@
...
+                triggerMetadata.FunctionName,
+                serviceBusReceiver.EntityPath,
+                !string.IsNullOrEmpty(serviceBusMetadata.QueueName) ? ServiceBusEntityType.Queue : ServiceBusEntityType.Topic,
+                new Lazy<ServiceBusReceiver>(() => serviceBusReceiver),
+                new Lazy<ServiceBusAdministrationClient>(() => serviceBusAdminClient),
+                loggerFactory);
+
+            _targetScaler = new ServiceBusTargetScaler(
```
pragnagopa, line 58: same here:  should this be removed - https://github.com/Azure/azure-sdk-for-net/blob/aa1ef2a332d38a95e7c7fd39a7c8659754b63f19/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusListener.cs#L115

alrod: No, there a re 2 different paths to create `ServiceBusTargetScaler`- one is regular trigger creation and another one is creation from SC.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusScalerProvider.cs
alrod: No, there a re 2 different paths to create `ServiceBusScaleMonitor `- one is regular trigger creation and another one is creation from SC.

eng/Packages.Data.props
```diff
@@ -134,8 +134,9 @@
     <PackageReference Update="Microsoft.Azure.SignalR.Management" Version="1.19.2" />
     <PackageReference Update="Microsoft.Azure.SignalR.Protocols" Version="1.19.2" />
     <PackageReference Update="Microsoft.Azure.SignalR.Serverless.Protocols" Version="1.9.0" />
-    <PackageReference Update="Microsoft.Azure.WebJobs" Version="3.0.36" />
-    <PackageReference Update="Microsoft.Azure.WebJobs.Sources" Version="3.0.36" />
+    <PackageReference Update="Microsoft.Azure.WebJobs" Version="3.0.37" />
+    <PackageReference Update="Microsoft.Azure.WebJobs.Sources" Version="3.0.37" />
+    <PackageReference Update="Microsoft.Azure.WebJobs.Host.Storage" Version="5.0.0" />
```
JoshLove-msft, line 139: Why is this reference needed?

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/tests/ScaleHostEndToEndTests.cs
```diff
@@ -0,0 +1,183 @@
...
+        ""direction"": ""in"",
+        ""queueName"": ""{SecondQueueScope.QueueName}"",
+        ""connection"": ""ServiceBusConnection2"",
+        ""functionName"": ""{Function2Name}""
+    }}
+ ]}}";
+
+            IHost host = new HostBuilder().ConfigureServices(services => services.AddAzureClientsCore()).Build();
+            AzureComponentFactory defaultAzureComponentFactory = host.Services.GetService<AzureComponentFactory>();
+            AzureComponentFactoryWrapper factoryWrapper = new AzureComponentFactoryWrapper(defaultAzureComponentFactory, ServiceBusTestEnvironment.Instance.Credential);
```
JoshLove-msft, line 72: Is the only reason we are creating the wrapper type to be able to pass in the environment credential? If so, you can instead follow the pattern here - https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/tests/WebJobsServiceBusTestBase.cs#L143-L148

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusHostBuilderExtensions.cs
```diff
@@ -93,5 +95,25 @@ public static IWebJobsBuilder AddServiceBus(this IWebJobsBuilder builder, Action
...
         }
+
+        public static IWebJobsBuilder AddServiceBusScaleForTrigger(this IWebJobsBuilder builder, TriggerMetadata triggerMetadata)
+        {
+            IServiceProvider serviceProvider = null;
+            Lazy<ServiceBusScalerProvider> scalerProvider = new Lazy<ServiceBusScalerProvider>(() => new ServiceBusScalerProvider(serviceProvider, triggerMetadata));
+
+            builder.Services.AddSingleton<IScaleMonitorProvider>(resolvedServiceProvider =>
+            {
+                serviceProvider = serviceProvider ?? resolvedServiceProvider;
```
JoshLove-msft, line 106: Can you clarify why we are doing this? Are we attempting to cache the first captured provider? Is this thread-safe? Why do we need to do this anyway - it doesn't seem to be used anywhere?

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusScalerProvider.cs
```diff
@@ -0,0 +1,116 @@
...
+{
+    internal class ServiceBusScalerProvider : IScaleMonitorProvider, ITargetScalerProvider
+    {
+        private readonly ServiceBusScaleMonitor _scaleMonitor;
+        private readonly ServiceBusTargetScaler _targetScaler;
+
+        public ServiceBusScalerProvider(IServiceProvider serviceProvider, TriggerMetadata triggerMetadata)
+        {
+            AzureComponentFactory azureComponentFactory;
+            if ((triggerMetadata.Properties != null) && (triggerMetadata.Properties.TryGetValue(nameof(AzureComponentFactory), out object value)))
```
JoshLove-msft, line 30: Can you clarify in comments when each path would be used and why?

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusHostBuilderExtensions.cs
```diff
@@ -93,5 +95,25 @@ public static IWebJobsBuilder AddServiceBus(this IWebJobsBuilder builder, Action
             builder.Services.AddSingleton<ServiceBusClientFactory>();
             return builder;
         }
+
+        public static IWebJobsBuilder AddServiceBusScaleForTrigger(this IWebJobsBuilder builder, TriggerMetadata triggerMetadata)
```
JoshLove-msft, line 99: Where is TriggerMetadata defined?
Does this method need to be public?

JoshLove-msft: Also, is the fact that this is for the trigger important to include in the method name? What about AddServiceBusScaler or AddServiceBusScaling?

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusScalerProvider.cs
```diff
@@ -0,0 +1,116 @@
...
+            public string QueueName { get; set; }
+
+            [JsonProperty]
+            public string TopicName { get; set; }
+
+            [JsonProperty]
+            public string SubscriptionName { get; set; }
+
+            [JsonProperty]
+            public bool IsSessionsEnabled { get; set; }
```
chiangvincent, line 100: We also need to add public bool [IsBatched](https://learn.microsoft.com/en-us/azure/azure-functions/functions-target-based-scaling?tabs=v5%2Ccsharp#:~:text=Execution%20Model-,IsBatched,-IsSessionsEnabled) for functions that are using batched processing

eng/Packages.Data.props
alrod: we need it to call [AddAzureStorageScaleServices](https://github.com/Azure/azure-webjobs-sdk/blob/55f3d9a7c8e0f38c9ad31b94585c9d395eeb7f57/src/Microsoft.Azure.WebJobs.Host.Storage/StorageServiceCollectionExtensions.cs#L39) in `ScaleHostEndToEndTest`.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusHostBuilderExtensions.cs
alrod: We are going to call the method per trigger in ScaleController. See an example in the E2E test:
https://github.com/Azure/azure-sdk-for-net/pull/34557/files#diff-2f37d5be90529dcbb846e2db09fd6d906bec85a72a6f637e8fd8d429d017c4aeR115

This is the only public interface change in the PR and we need it to call from ScaleController. The method adds to DI scale entities for specific trigger.  We do not expect function app customers to call this method.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusHostBuilderExtensions.cs
alrod: We need only one `ServiceBusScalerProvider` to add `IScaleMonitorProvider` and `ITargetScalerProvider` to DI container. That's why I am using `Lazy<ServiceBusScalerProvider>` as it's [thread safe.](https://learn.microsoft.com/en-us/dotnet/framework/performance/lazy-initialization#thread-safe-initialization)

We need to call [AddServiceBusScaleForTrigger ](https://github.com/Azure/azure-sdk-for-net/pull/34557/files#diff-2f37d5be90529dcbb846e2db09fd6d906bec85a72a6f637e8fd8d429d017c4aeR115 ) per trigger to use the scale entities associated with the trigger in `ScaleController`.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusScalerProvider.cs
```diff
@@ -0,0 +1,116 @@
...
+            }
+            else
+            {
+                azureComponentFactory = serviceProvider.GetService<AzureComponentFactory>();
+            }
+
+            IConfiguration configuration = serviceProvider.GetService<IConfiguration>();
+            MessagingProvider messagingProvider = serviceProvider.GetService<MessagingProvider>();
+            AzureEventSourceLogForwarder logForwarder = serviceProvider.GetService<AzureEventSourceLogForwarder>();
+            IOptions<ServiceBusOptions> options = serviceProvider.GetService<IOptions<ServiceBusOptions>>();
```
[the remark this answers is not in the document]
alrod: yes, the options are coming from `host.json`. `host.json` is coming from `IConfiguration `passed from `ScaleController`.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusScalerProvider.cs
alrod: There is already `Cardinality` property in the metadata, it looks like IsBatched is only for C# in-proc/isolated worker. I am confused why it was added in 5.x as we always can check if trigger input is singular or plural.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusHostBuilderExtensions.cs
JoshLove-msft: I think this is the type - https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Scale/TriggerMetadata.cs#L13

It is unusual for us to add public surface that references a non-framework type from another library that is not part of the Azure SDK. Is the TriggerMetadata type already GA?

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusHostBuilderExtensions.cs
JoshLove-msft: Sorry my question was about the caching of the serviceProvider and using that in both of the AddSingleton lamdas.

JoshLove-msft: https://github.com/Azure/azure-sdk-for-net/pull/34557/files#diff-3b4ac6f1953372c2cb8f8861cda043dc9fe76aee45653789a18c54c1c3c2bf85R106
and
https://github.com/Azure/azure-sdk-for-net/pull/34557/files#diff-3b4ac6f1953372c2cb8f8861cda043dc9fe76aee45653789a18c54c1c3c2bf85R112

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusScalerProvider.cs
JoshLove-msft: I don't think this is exposed anywhere in 5.x for in-proc.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/tests/ScaleHostEndToEndTests.cs
alrod: We do have all the properties (clientId, clientSecret, tenantId) as user does not specify them. ScaleController/Antaers can make a special call to get the managed identity token credentials. Then we can pass them to the extension.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusHostBuilderExtensions.cs
JoshLove-msft: Another comment - why not just take the AzureComponentFactory as a parameter rather than needing to fish it out of the properties dictionary in a non-typed way?

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/tests/ScaleHostEndToEndTests.cs
JoshLove-msft: Please add test to the name, e.g. TestComponentFactory.

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusHostBuilderExtensions.cs
alrod: We do not want to create all the instances twice so we are using one instance of `ServiceBusScalerProvider `
https://github.com/Azure/azure-sdk-for-net/pull/34557/files#diff-6e49cf486d47dbf5f4e52aa9911bbe8dba4a2c286bcfdfdebe10b84346fe3d11R27

sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusHostBuilderExtensions.cs
alrod: `TriggerMetadata` is already GA (we had released the web jobs sdk 3.0.37).
We did not want to introduce new public types for each trigger, all the extensions will have the same mew public method:
`Add{**EXTENSION_NAME**}ScaleForTrigger(this IWebJobsBuilder builder, TriggerMetadata triggerMetadata)`

We considered passing AzureComponentFactory  as parameter and decided to use the properties collection as it give us more flexibility if you want to pass some other data in future.

JoshLove-msft:
Will need to generate the API listing.

JoshLove-msft (approved):

PR #34563, opened 2023-02-27
Data share tests

mcgallan:


sdk/datashare/Azure.ResourceManager.DataShare/tests/SessionRecords/AccountTests/ShareAccountCollectionTests().json
```diff
@@ -0,0 +1,2626 @@
...
+        "state": "Enabled",
+        "subscriptionPolicies": {
+          "locationPlacementId": "Internal_2014-09-01",
+          "quotaId": "Internal_2014-09-01",
+          "spendingLimit": "Off"
+        }
+      }
+    },
+    {
+      "RequestUri": "https://management.azure.com/subscriptions/db1ab6f0-4769-4b27-930e-01e2ef9c123c/resourcegroups/testRG-5740?api-version=**",
```
m-nash, line 60: Recently we added something to the base class which should suppress recording dependency setup since every time those api versions change this test will begin to fail unless re-recorded.  We should look into why that isn't working in this case.

mcgallan: fixed

ArthurMa1978 (approved):

PR #34574, opened 2023-02-27
Potential wrong use of assertion

msspider117:
We encountered an error while using the SDK where GetAll would always throw. After looking through the code we saw that where an instance of this class is created (TrafficManagerProfileResource), the ctor with the 3rd parameter is never called. I'm guessing that's from the AutoRest codegen.

Maybe adding AssertNotNull rather than AssertNull would fix the problem in the codegen?

How we made it work was to use reflection to get _profileData and set it before calling GetAll()


msspider117:
Bump

ArcturusZhang (approved):

ArcturusZhang:
Well despite I approved this PR because the assert is really incorrect, this is not enough to get the methods added in this file working - the new GetAll and GetAllAsync methods require the profileData to be set, but the only ctor that could set it is `internal` and no other methods in the SDK are calling it. Therefore it will always be null...

PR #34575, opened 2023-02-27
Improve test coverage calculation

ostoliarova-ee:
This PR addresses the [User Story 2539748](https://skype.visualstudio.com/SPOOL/_workitems/edit/2539748): [Pipeline][Identity][.NET] Test coverage improvement

The issue of Common and Common.Tests being included into the pipeline code calculation was in these projects being included in the Identity.Tests csproj file. This caused test coverage tool to include them in the total number of lines covered by tests.

It turned out that Common could be removed from being included in the  Identity.Tests csproj without any other code change. Probably it was included by mistake or there used to be some legacy code using Common, which is not the case any more.

Common.Tests is included to the test projects of other ACS SDKs as well resulting in Common.Tests being added to all SDK test coverage calculation. The reason for this change was an attempt to have one generic test class that reads required environment variables. As it doesn't have additional logic and contained variables not needed for Identity.Tests, I replaced it with a `CommunicationTestEnvironment ` class contained in the Identity.Tests and removed inclusion of a Common.Tests project. 
The updated code calculation result could be found here: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2604496&view=results

This is an approach that other SDK owners might want to follow in order to decrease coupling, make the tests more flexible and avoid reading/instantiating many unused values.


sdk/communication/Azure.Communication.Identity/tests/Infrastructure/CommunicationTestEnvironment.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        public string LiveTestStaticConnectionString => GetRecordedVariable(
+                LiveTestStaticConnectionStringEnvironmentVariableName,
+                options => options.HasSecretConnectionStringParameter("accessKey", SanitizedValue.Base64));
+
+        public Uri LiveTestStaticEndpoint => new(Core.ConnectionString.Parse(LiveTestStaticConnectionString).GetRequired("endpoint"));
+
+        public string LiveTestStaticAccessKey => Core.ConnectionString.Parse(LiveTestStaticConnectionString).GetRequired("accesskey");
+    }
+}
```
petrsvihlik, line 30: we don't want to duplicate common code.

petrsvihlik (changes requested):
This approach duplicates code. I don't think we want to take this path to resolve the coverage issue. We'd only introduce the Python's problem of the `__shared` folders in .NET.

Please explore the following:
- https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.excludefromcodecoverageattribute?view=net-7.0
- adjusting the assemblyfilters, classfilters, filefilters params in `C:\hostedtoolcache\windows\dotnet\dotnet.exe D:\a\_work\_tasks\reportgenerator_be803a55-9253-4895-a525-be570d86f161\4.8.13\tools\netcoreapp2.1\ReportGenerator.dll -reports:D:\a\_work\1\s\sdk\communication\**\coverage.cobertura.xml -targetdir:D:\a\_work\1\a\coverage -reporttypes:Cobertura -sourcedirs: -historydir: -plugins: -assemblyfilters:+* -classfilters:+* -filefilters:+D:\a\_work\1\s\sdk\communication\** -verbosity:Verbose -title: -tag:20230303.4_#2604496`  in the [pipeline](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2604496&view=logs&j=ab7e6336-ef84-5ec5-cd26-bc908c4b1482&t=ee0553fc-2c2d-5355-6b13-00dc01c42161)
- cobertura configuration - it seems that's [the tool](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2604496&view=logs&j=ab7e6336-ef84-5ec5-cd26-bc908c4b1482&t=ee0553fc-2c2d-5355-6b13-00dc01c42161) we use for coverage generation.

sdk/communication/Azure.Communication.Identity/tests/Infrastructure/CommunicationTestEnvironment.cs
AikoBB: totally agree, it's better to take another approach that will not introduce code duplication, as the common & common tests are included for other modalities as well, which will lead to the same issue as in python one

PR #34583, opened 2023-02-27
[AppConfiguration] Snapshot API Updates

m-redding:
These changes are based on discussions within the App Configuration SDK team across languages. The changes are based around the following:

- Internalize Protocol methods
- Changing ConfigurationSettingsSnapshot.RetentionPeriod from a long to a TimeSpan
- Separate ETag use into overloads for the methods updating status (Archive and Recover)
- Implementing `CreateSnapshot` and `CreateSnapshotAsync` as LROs


sdk/appconfiguration/Azure.Data.AppConfiguration/src/ConfigurationClient.cs
```diff
@@ -931,6 +963,25 @@ public virtual async Task<Response<ConfigurationSettingsSnapshot>> CreateSnapsho
             }
         }
 
+        internal async Task<Response> CreateSnapshotAsync(string name, RequestContent content, ContentType contentType, RequestContext context = null)
```
pallavit, line 966: are these manually generated?

Looking at the autorest.md settings we have set public-clients: true however we are also creating a public client manually which seems like a conflict. Please remove that flag.

sdk/appconfiguration/Azure.Data.AppConfiguration/src/ConfigurationClient.cs
```diff
@@ -14,6 +14,46 @@ namespace Azure.Data.AppConfiguration
...
+    [CodeGenSuppress("GetRevisionsAsync", typeof(string), typeof(string), typeof(string), typeof(string), typeof(IEnumerable<string>), typeof(RequestContext))]
+    [CodeGenSuppress("SetConfigurationSetting", typeof(string), typeof(RequestContent), typeof(ContentType), typeof(string), typeof(MatchConditions), typeof(RequestContext))]
+    [CodeGenSuppress("SetConfigurationSettingAsync", typeof(string), typeof(RequestContent), typeof(ContentType), typeof(string), typeof(MatchConditions), typeof(RequestContext))]
+    [CodeGenSuppress("DeleteConfigurationSetting", typeof(string), typeof(string), typeof(ETag), typeof(RequestContext))]
+    [CodeGenSuppress("DeleteConfigurationSettingAsync", typeof(string), typeof(string), typeof(ETag), typeof(RequestContext))]
+    [CodeGenSuppress("GetConfigurationSetting", typeof(string), typeof(string), typeof(string), typeof(IEnumerable<string>), typeof(MatchConditions), typeof(RequestContext))]
+    [CodeGenSuppress("GetConfigurationSettingAsync", typeof(string), typeof(string), typeof(string), typeof(IEnumerable<string>), typeof(MatchConditions), typeof(RequestContext))]
```
pallavit, line 56: This should not be needed if you remove the flag from the autorest file.

sdk/appconfiguration/Azure.Data.AppConfiguration/src/ConfigurationClient.cs
m-redding: These changes are still in progress, but I will look into why the autorest.md file is written that way. I asked around trying to learn how to internalize the dozens of unnecessary protocol methods being generated and this is what was shared with me as the solution.

pallavit: Got it. I will be happy to help you if you have questions. @m-nash may be able to assist us more here and appconfig may be in an interim state due to the intern project work. Here is where I would ultimately want AppConfig to be

1. Use the same autorest settings as ingestion (so use DPG generator w/o models and retrofit the convenience layer with it with or w/o making the protocol methods public (depending on what guidance architects give) and whatever is most efficient in terms of cost. 
2. Use HLC generator and be consistent with rest of older data plane SDKs that our team owns. I would prefer the top one because this is closest to what we tell our partners and we are so close to getting AppConfig there.

sdk/appconfiguration/Azure.Data.AppConfiguration/src/ConfigurationClient.cs
m-redding: After setting the flag to false or removing the flag I'm still finding I need to include these suppressions, is that unexpected behavior?

sdk/appconfiguration/Azure.Data.AppConfiguration/src/ConfigurationClient.cs
pallavit: @m-redding given we don't intend to expose protocol methods - we will have to move us back to the older flag generation1-convenience-client: true (this will move us back to HLC generator), however this should be fixed in a couple of months once we move to single generator. @m-nash please keep me honest here.

m-redding: Ok sounds good I'll fix that!

sdk/appconfiguration/Azure.Data.AppConfiguration/src/autorest.md
```diff
@@ -45,3 +44,92 @@ directive:
...
+  transform: >
+    $["x-accessibility"] = "internal";
+- from: swagger-document
+  where: $..[?(@.operationId=='GetLabels')]
+  transform: >
+    $["x-accessibility"] = "internal";
+- from: swagger-document
+  where: $..[?(@.operationId=='GetRevisions')]
+  transform: >
+    $["x-accessibility"] = "internal";
```
ShivangiReja, line 134: You can simplify this configuration using below logic:

```yaml
directive:
  from: swagger-document
  where: $.paths.*.*
  transform: >
    $["x-accessibility"] = "internal"
```

sdk/appconfiguration/Azure.Data.AppConfiguration/src/CreateSnapshotOperation.cs
```diff
@@ -0,0 +1,93 @@
...
+        protected CreateSnapshotOperation() { }
+
+        /// <inheritdoc/>
+        public override ConfigurationSettingsSnapshot Value => _snapshot;
+
+        /// <inheritdoc/>
+        public override bool HasValue => true;
+
+        /// <inheritdoc/>
+        public override string Id => _snapshot.Etag.ToString(); // TODO
```
m-redding, line 52: Need to confirm what the value should be here.

sdk/appconfiguration/Azure.Data.AppConfiguration/src/ConfigurationClient.cs
```diff
@@ -916,13 +913,18 @@ public virtual async Task<Response<ConfigurationSettingsSnapshot>> CreateSnapsho
...
+                //ContentType contentType = new ContentType(HttpHeader.Common.JsonContentType.Value.ToString());
 
-                Response response = await CreateSnapshotAsync(name, content, contentType, context).ConfigureAwait(false);
-                ConfigurationSettingsSnapshot value = ConfigurationSettingsSnapshot.FromResponse(response);
-                return Response.FromValue(value, response);
+                //Response response = await CreateSnapshotAsync(name, content, contentType, context).ConfigureAwait(false);
+                //ConfigurationSettingsSnapshot value = ConfigurationSettingsSnapshot.FromResponse(response);
+                //return Response.FromValue(value, response);
+
+                // TODO: update to LRO pattern
```
pallavit, line 924: Do you have a bug for this? Or is this something you are planning to do as part of this PR itself?

Want to make sure this does not get missed.

sdk/appconfiguration/Azure.Data.AppConfiguration/src/Models/ConfigurationSettingsSnapshot.cs
```diff
@@ -14,7 +14,7 @@ public partial class ConfigurationSettingsSnapshot
         /// <summary> Initializes a new instance of Snapshot. </summary>
         /// <param name="filters"> A list of filters used to filter the key-values included in the snapshot. </param>
         /// <exception cref="ArgumentNullException"> <paramref name="filters"/> is null. </exception>
-        public ConfigurationSettingsSnapshot(IEnumerable<ConfigurationSettingFilter> filters)
+        public ConfigurationSettingsSnapshot(IEnumerable<SnapshotSettingFilter> filters)
```
pallavit, line 17: >SnapshotSettingFilter [](http://example.com/codeflow?start=57&length=21)

Just curious, what is the motivation of this name change? ConfigurationSettingFilter seemed more aligned in terms of naming to the class name than snapshot setting filter.

sdk/appconfiguration/Azure.Data.AppConfiguration/src/ConfigurationClient.cs
m-redding: I'm working on it now, but I don't have access to corpnet for the time being, so nothing can be tested until the service is released.

sdk/appconfiguration/Azure.Data.AppConfiguration/src/Models/ConfigurationSettingsSnapshot.cs
m-redding: The reasoning was that the setting filter is only used by the snapshot feature, and it seemed unclear that that was the case without having snapshot in the name. The other alternative we considered was `ConfigurationSettingsSnapshotSettingFilter` but decided instead that all of the supplementary types would just be `Snapshot...` and then only the snapshot object would include configuration in the name (to avoid a single name type)

sdk/appconfiguration/Azure.Data.AppConfiguration/src/CreateSnapshotOperation.cs
jsquire: There should be an operation Id header that gets emitted by the service.   As far as I'm aware, the generator should take care of this.... I think?

m-redding: I actually wrote this class by hand (I think because we're generating protocol methods only), so I updated it accordingly. Since I can't test it it's hard for me to know if this will work and how the operation response is deserialized and all of those things, so it may change once I can do that.

sdk/appconfiguration/Azure.Data.AppConfiguration/src/Models/SnapshotSettingFilter.cs
```diff
@@ -7,12 +7,13 @@
 namespace Azure.Data.AppConfiguration
 {
     /// <summary> Enables filtering of key-values. </summary>
-    public partial class ConfigurationSettingFilter
+    [CodeGenModel("ConfigurationSettingFilter")]
```
ShivangiReja, line 10: We are not generating models, so why do we need `[CodeGenModel("ConfigurationSettingFilter")]`?

sdk/appconfiguration/Azure.Data.AppConfiguration/src/Models/ConfigurationSettingsSnapshot.cs
```diff
@@ -58,15 +58,36 @@ internal ConfigurationSettingsSnapshot(string name, SnapshotStatus? status, int?
...
         /// <summary> The composition type describes how the key-values within the snapshot are composed. The &apos;all&apos; composition type includes all key-values. The &apos;group_by_key&apos; composition type ensures there are no two key-values containing the same key. </summary>
         public CompositionType? CompositionType { get; set; }
         /// <summary> The time that the snapshot was created. </summary>
         public DateTimeOffset? Created { get; }
         /// <summary> The time that the snapshot will expire. </summary>
         public DateTimeOffset? Expires { get; }
+        private long? _retentionPeriod;
         /// <summary> The amount of time, in seconds, that a snapshot will remain in the archived state before expiring. This property is only writable during the creation of a snapshot. If not specified, the default lifetime of key-value revisions will be used. </summary>
-        public long? RetentionPeriod { get; set; }
+        [CodeGenMember("RetentionPeriod")]
```
ShivangiReja, line 70: We are not generating models, so why do we need `[CodeGenMember("RetentionPeriod")]`?

sdk/appconfiguration/Azure.Data.AppConfiguration/src/Models/ConfigurationSettingsSnapshot.cs
```diff
@@ -58,15 +58,36 @@ internal ConfigurationSettingsSnapshot(string name, SnapshotStatus? status, int?
...
+        internal long? RetentionPeriodLong { get { return _retentionPeriod; } set { _retentionPeriod = value; } }
```
ShivangiReja, line 71: Now that we have `RetentionPeriod`, why do we need an internal `RetentionPeriodLong` property?

sdk/appconfiguration/Azure.Data.AppConfiguration/src/ConfigurationClient.cs
```diff
@@ -1057,10 +1169,8 @@ public virtual async Task<Response<ConfigurationSettingsSnapshot>> RecoverSnapsh
 
         /// <summary> Updates the state of a configuration setting snapshot to ready. </summary>
         /// <param name="name"> The name of the configuration setting snapshot to delete. </param>
-        /// <param name="matchConditions">"IfMatch" Used to perform an operation only if the targeted resource's etag matches the value provided and
-        /// "IfNoneMatch" Used to perform an operation only if the targeted resource's etag does not match the value provided. </param>
         /// <param name="cancellationToken"> The cancellation token to use. </param>
-        public virtual Response<ConfigurationSettingsSnapshot> RecoverSnapshot(string name, MatchConditions matchConditions = null, CancellationToken cancellationToken = default)
+        public virtual Response<ConfigurationSettingsSnapshot> RecoverSnapshot(string name, CancellationToken cancellationToken = default)
```
ShivangiReja, line 1173: For a moment I thought you missed async method for this. For clarity, Can we move this method right after it's async method?

sdk/appconfiguration/Azure.Data.AppConfiguration/src/ConfigurationClient.cs
```diff
@@ -993,10 +1037,8 @@ public virtual async Task<Response<ConfigurationSettingsSnapshot>> ArchiveSnapsh
 
         /// <summary> Updates the state of a configuration setting snapshot to archive. </summary>
         /// <param name="name"> The name of the configuration setting snapshot to delete. </param>
-        /// <param name="matchConditions">"IfMatch" Used to perform an operation only if the targeted resource's etag matches the value provided and
-        /// "IfNoneMatch" Used to perform an operation only if the targeted resource's etag does not match the value provided. </param>
         /// <param name="cancellationToken"> The cancellation token to use. </param>
-        public virtual Response<ConfigurationSettingsSnapshot> ArchiveSnapshot(string name, MatchConditions matchConditions = null, CancellationToken cancellationToken = default)
+        public virtual Response<ConfigurationSettingsSnapshot> ArchiveSnapshot(string name, CancellationToken cancellationToken = default)
```
ShivangiReja, line 1041: For clarity, can we move this method right after it's async method?

m-redding:
Will circle back to the LRO implementation once I can test again, I believe there are also updates that will be going into the swagger in order to be able to generate the Operation, which will also change the LRO implementation.

sdk/appconfiguration/Azure.Data.AppConfiguration/src/CreateSnapshotOperation.cs
```diff
@@ -0,0 +1,141 @@
...
+                    throw;
+                }
+            }
+
+            return GetRawResponse();
+        }
+
+        private static bool IsJobComplete(string status)
+        {
+            if (status == "Succeeded" || status == "Failed" || status == "Canceled")
```
ShivangiReja, line 134: Instead of hardcoding the status value here, can we defined an enum same as defined in the swagger [here](https://github.com/Azure/azure-rest-api-specs/blob/48125b4a8cd39b3034da70bc57d6ddc738b10de5/specification/appconfiguration/data-plane/Microsoft.AppConfiguration/preview/2022-11-01-preview/appconfiguration.json#L1780-L1786)?

sdk/appconfiguration/Azure.Data.AppConfiguration/src/ConfigurationClient.cs
```diff
@@ -904,10 +902,11 @@ public virtual Response GetSnapshot(string name, IEnumerable<string> select, Mat
         }
 
         /// <summary> Creates a configuration setting snapshot. </summary>
+        /// <param name="wait">TODO</param>
         /// <param name="name"> The name of the configuration setting snapshot to create. </param>
         /// <param name="snapshot"> The configuration setting snapshot to create. </param>
         /// <param name="cancellationToken"> The cancellation token to use. </param>
-        public virtual async Task<Response<ConfigurationSettingsSnapshot>> CreateSnapshotAsync(string name, ConfigurationSettingsSnapshot snapshot, CancellationToken cancellationToken = default)
+        public virtual async Task<CreateSnapshotOperation> CreateSnapshotAsync(WaitUntil wait, string name, ConfigurationSettingsSnapshot snapshot, CancellationToken cancellationToken = default)
```
ShivangiReja, line 909: Can we add the test cases to cover all the LRO scenarios for create operation?

m-redding:
Returned to the office today so I can access dogfood resources again. 😄

sdk/appconfiguration/Azure.Data.AppConfiguration/tests/ConfigurationMockTests.cs
```diff
@@ -63,7 +63,7 @@ public async Task Get()
             var mockTransport = new MockTransport(response);
             ConfigurationClient service = CreateTestService(mockTransport);
 
-            ConfigurationSetting setting = await service.GetConfigurationSettingAsync(s_testSetting.Key);
+            ConfigurationSetting setting = await service.GetConfigurationSettingAsync(key: s_testSetting.Key);
```
ShivangiReja, line 66: Why do we need to update this test?

sdk/appconfiguration/Azure.Data.AppConfiguration/tests/samples/Sample11_SettingsSnapshot.cs
```diff
@@ -100,27 +183,29 @@ public void GetSnapshots()
...
-            foreach (ConfigurationSettingsSnapshot item in client.GetSnapshots())
-            {
-                count++;
-                Console.WriteLine($"Name {item.Name} status {item.Status}");
-            }
+            //foreach (ConfigurationSettingsSnapshot item in client.GetSnapshots())
+            //{
+            //    count++;
+            //    Console.WriteLine($"Name {item.Name} status {item.Status}");
+            //}
```
ShivangiReja, line 208: Why did we comment out this loop?

ShivangiReja (approved):
PR LGTM!!

I'll takeover the AppConfig work and update the implementation if needed. Thank you so much @m-redding for all the work!

sdk/appconfiguration/Azure.Data.AppConfiguration/tests/samples/Sample11_SettingsSnapshot.cs
m-redding: Since this is in the project, the compiler could no longer distinguish between the public GetSnapshots(string,string,IEnumerable<SnapshotFields>,IEnumerable<SnapshotStatus, CancellationToken) and the public GetSnapshots(string, string, IEnumerable<string>,IEnumerable<string>,RequestContext)

sdk/appconfiguration/Azure.Data.AppConfiguration/tests/ConfigurationMockTests.cs
m-redding: Similar to the other comment, it was to distinguish between the internal and public method

sdk/appconfiguration/Azure.Data.AppConfiguration/tests/samples/Sample11_SettingsSnapshot.cs
m-redding: We may need to update this to pass in a specific parameter to help distinguish between them. FWIW this isn't a problem for the public API

sdk/appconfiguration/Azure.Data.AppConfiguration/tests/ConfigurationMockTests.cs
ShivangiReja: We should not update these test cases, instead we should fix the method ambiguity because users can also face the same issues.

 If there is a method ambiguity between two calls, then as per [this](https://github.com/Azure/autorest.csharp/issues/2204) we should disambiguate the methods signatures by making RequestContext and other parameters required for protocol method. Since this decision is not clear yet([see here](https://github.com/Azure/autorest.csharp/issues/2204#issuecomment-1146403059)) and protocol methods are internal, we can remove rename configuration for those protocol methods where we see method ambiguity.

sdk/appconfiguration/Azure.Data.AppConfiguration/tests/samples/Sample11_SettingsSnapshot.cs
ShivangiReja: See response [here](https://github.com/Azure/azure-sdk-for-net/pull/34583#discussion_r1156923716).

I'll uncomment this code and fix the method ambiguity issue

sdk/appconfiguration/Azure.Data.AppConfiguration/src/ConfigurationClient.cs
m-redding: We initially discussed offline that I would not be writing the test cases or samples for the LRO case. But I added them per your feedback here.

sdk/appconfiguration/Azure.Data.AppConfiguration/tests/ConfigurationMockTests.cs
m-redding: Some of these changes were made in an effort to transition the work back to you as quickly as possible. Feel free to make any changes to what I've done here as you see fit now that it has been merged in.

sdk/appconfiguration/Azure.Data.AppConfiguration/src/ConfigurationClient.cs
ShivangiReja: Yeah, that time we didn’t have dogfood resources working. Now that they are working, I thought since you have implemented the feature, you will have more context and if you add test cases, you will also be sure that your implementation is working.

PR #34589, opened 2023-02-28
Add ability to enumerate over objects to DynamicJson

annelo-msft:
This PR adds an ObjectEnumerator to DynamicJson and MutableJsonElement, similar to the one in JsonElement.

Implemented to unblock a CloudMachine requirement.

Contributes to https://github.com/Azure/azure-sdk-for-net/issues/33856.


KrzysztofCwalina (approved):

sdk/core/Azure.Core.Experimental/src/MutableJsonElement.ObjectEnumerator.cs
```diff
@@ -0,0 +1,100 @@
...
+    /// <summary>
+    /// A mutable representation of a JSON element.
+    /// </summary>
+    public readonly partial struct MutableJsonElement
+    {
+        /// <summary>
+        ///   An enumerable and enumerator for the properties of a JSON object.
+        /// </summary>
+        [DebuggerDisplay("{Current,nq}")]
+        public struct ObjectEnumerator : IEnumerable<(string Name, MutableJsonElement Value)>, IEnumerator<(string Name, MutableJsonElement Value)>
```
christothes, line 21: Just curious - why do we use a Tuple in this implementation and a full type in the DynamicJson.ObjectEnumerator?

sdk/core/Azure.Core.Experimental/src/MutableJsonElement.ObjectEnumerator.cs
```diff
@@ -0,0 +1,100 @@
...
+        {
+            private readonly MutableJsonElement _target;
+            private JsonElement.ObjectEnumerator _enumerator;
+
+            internal ObjectEnumerator(MutableJsonElement target)
+            {
+                Debug.Assert(target.ValueKind == JsonValueKind.Object);
```
christothes, line 28: Is this intentionally only validating in Debug mode rather than `EnsureObject()`?

sdk/core/Azure.Core.Experimental/src/MutableJsonElement.cs
```diff
@@ -272,6 +272,19 @@ public ArrayEnumerator EnumerateArray()
             return new ArrayEnumerator(this);
         }
 
+        /// <summary>
+        /// Gets an enumerator to enumerate the properties in the JSON object represented by this JsonElement.
+        /// </summary>
+        /// <returns></returns>
```
christothes, line 278: provide value or remove?

sdk/core/Azure.Core.Experimental/tests/MutableJsonElementTests.cs
```diff
@@ -170,5 +168,110 @@ public void CanEnumerateArrayWithChanges()
                 Assert.AreEqual(expected++, el.GetInt32());
             }
         }
+
+        [Test]
```
christothes, line 172: Could we add tests that attempt to enumerate properties of the wrong type?

sdk/core/Azure.Core.Experimental/tests/DynamicJsonEnumerableTests.cs
```diff
@@ -84,5 +84,56 @@ public void CanForEachOverIntArray()
                 Assert.AreEqual(expected++, i);
             }
         }
+
+        [Test]
```
christothes, line 88: Might be nice to also add a test where the object to enumerate is nested in another property.

sdk/core/Azure.Core.Experimental/src/MutableJsonElement.ObjectEnumerator.cs
annelo-msft: It's because the DLR doesn't handle the field names on Tuples correctly, because the names are just attributes associated with user code.  I have a separate item to make the property Value settable, and some of that work will need to be done in the DynamicJsonProperty Value setter.  If any needs to be added to the tuple setter, we may need to graduate this to a struct, but TBD.

sdk/core/Azure.Core.Experimental/src/MutableJsonElement.ObjectEnumerator.cs
annelo-msft: Oh good catch, thank you.

sdk/core/Azure.Core.Experimental/tests/DynamicJsonEnumerableTests.cs
annelo-msft: Sounds good - will do!

Issue #33902, opened 2023-02-06
[BUG] System.InvalidOperationException: '' is not a valid value for a partition key or row key thrown when using Extensions.Table package with Azure functions output binding


kshyju:
### Library name and version

Microsoft.Azure.WebJobs.Extensions.Table 1.0.0

### Describe the bug

When using this package in the azure functions context to save data to the table using function's output bindings, an invalid operation [exception is thrown ](https://github.com/Azure/azure-sdk-for-net/blob/f79888a2068d2abc7aa0997b0599250d0bf55db6/sdk/tables/Microsoft.Azure.WebJobs.Extensions.Tables/src/TableClientHelpers.cs#L82)with below stack trace (trimmed to keep it only the relevant stack from this package).

> System.InvalidOperationException: '' is not a valid value for a partition key or row key.
         at Microsoft.Azure.WebJobs.Extensions.Tables.TableClientHelpers.ValidateAzureTableKeyValue(String value) in C:\Dev\OSS\azure-sdk-for-net\sdk\tables\Microsoft.Azure.WebJobs.Extensions.Tables\src\TableClientHelpers.cs:line 86
         at Microsoft.Azure.WebJobs.Extensions.Tables.TableEntityWriter.AddAsync(TableEntity item, CancellationToken cancellationToken) in C:\Dev\OSS\azure-sdk-for-net\sdk\tables\Microsoft.Azure.WebJobs.Extensions.Tables\src\TableEntityWriter.cs:line 58
         at Microsoft.Azure.WebJobs.Host.Bindings.TypedAsyncCollectorAdapter`3.AddAsync(TSrc item, CancellationToken cancellationToken) in C:\projects\azure-webjobs-sdk-rqm4t\src\Microsoft.Azure.WebJobs.Host\Bindings\AsyncCollector\TypedAsyncCollectorAdapter.cs:line 40
         at Microsoft.Azure.WebJobs.Script.Binding.FunctionBinding.BindAsyncCollectorAsync[T](BindingContext context) in C:\Dev\OSS\azure-functions-host\src\WebJobs.Script\Binding\FunctionBinding.cs:line 199

I did a bit of debugging and found the following.

1. TableEntityWriter.AddAsync is called with an instance of TableEntity and the PartitiionKey property was read and passed to ` TableClientHelpers.ValidateAzureTableKeyValue`. The exception is thrown from that method since the value passed is `null`.

[...]

PR #34618, opened 2023-03-01
[WebPubSub]Add isolated-process extension for WebPubSub

JialinXin:


sdk/webpubsub/Microsoft.Azure.Functions.Worker.Extensions.WebPubSub/src/Converters/ConnectionStatesJsonConverter.cs
```diff
@@ -0,0 +1,68 @@
...
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Text.Json.Serialization;
+using System.Text.Json;
+
+namespace Microsoft.Azure.Functions.Worker
+{
+    /// Copied from: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/webpubsub/Microsoft.Azure.WebPubSub.Common/src/Shared/ConnectionStatesConverter.cs
```
jsquire, line 11: Why not link to the Common instance as shared code?

sdk/webpubsub/Microsoft.Azure.Functions.Worker.Extensions.WebPubSub/src/Models/Output/CloseGroupConnectionsAction.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        /// <summary>
+        /// Target group name.
+        /// </summary>
+        public string Group { get; set; }
+
+        /// <summary>
+        /// ConnectionIds to exclude.
+        /// </summary>
+        public IList<string> Excluded { get; set; } = new List<string>();
```
jsquire, line 21: This goes against the SDK convention of having collection-type members for models as get-only.  Please discuss with @KrzysztofCwalina and @tg-msft.

sdk/webpubsub/Microsoft.Azure.Functions.Worker.Extensions.WebPubSub/src/Models/Output/WebPubSubAction.cs
```diff
@@ -0,0 +1,19 @@
...
+    /// <summary>
+    /// Abstract class of operation to invoke service.
+    /// </summary>
+    public abstract class WebPubSubAction
+    {
+        internal string ActionName
+        {
+            get
+            {
+                return GetType().Name.Replace("Action", "");
```
jsquire, line 15: Any reason why this can't be cached after the first call?

jsquire:
@JoshLove-msft:  Would you please review?

JoshLove-msft:
Are we able to add tests?

sdk/webpubsub/Microsoft.Azure.Functions.Worker.Extensions.WebPubSub/src/Models/Output/WebPubSubAction.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Azure.Functions.Worker
+{
+    /// <summary>
+    /// Abstract class of operation to invoke service.
+    /// </summary>
+    public abstract class WebPubSubAction
```
JoshLove-msft, line 9: Do we not need all of these factory methods? https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/webpubsub/Microsoft.Azure.WebJobs.Extensions.WebPubSub/api/Microsoft.Azure.WebJobs.Extensions.WebPubSub.netstandard2.0.cs#L112

sdk/webpubsub/Microsoft.Azure.Functions.Worker.Extensions.WebPubSub/src/Models/Output/SendToConnectionAction.cs
```diff
@@ -0,0 +1,30 @@
...
+
+using System;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.Azure.Functions.Worker
+{
+    /// <summary>
+    /// Operation to send message to a connection.
+    /// </summary>
+    public class SendToConnectionAction : WebPubSubAction
```
JoshLove-msft, line 12: I suppose we don't need any converters to use camel case because this is handled in the extension itself?
https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/webpubsub/Microsoft.Azure.WebJobs.Extensions.WebPubSub/src/Bindings/Output/RemoveConnectionFromGroupAction.cs#L12

sdk/webpubsub/Microsoft.Azure.Functions.Worker.Extensions.WebPubSub/src/Models/Output/CloseGroupConnectionsAction.cs
JialinXin: The isolated process extension needs these types to do Json deserialize and requires a parameterless constructor that not what we want to have in the Common package. And these types are specifically used in this isolated process project.
The trigger types are essential to declare and rest types are simply easier for customer to use in the isolated functions. Otherwise, they have to define these themselves.

sdk/webpubsub/Microsoft.Azure.Functions.Worker.Extensions.WebPubSub/src/Models/Output/WebPubSubAction.cs
JialinXin: It can be cached though it's called single time by the in-process extension. Let me update.

sdk/webpubsub/Microsoft.Azure.Functions.Worker.Extensions.WebPubSub/src/Models/Output/SendToConnectionAction.cs
JialinXin: Suppose the common practice is not to mix use in-process and isolated-process extension packages. So either customers have to declare themselves or we do this for convenience.

See sample without built-in types and customer have to do in this way in early SignalR extension: https://github.com/Azure/azure-functions-dotnet-worker/blob/main/samples/Extensions/SignalR/SignalRFunction.cs

JialinXin:
> Are we able to add tests?

Not much needs as these are simply type definitions. Refers to a few isolated-process extensions are usually very simple.

sdk/webpubsub/Microsoft.Azure.Functions.Worker.Extensions.WebPubSub/README.md
```diff
@@ -0,0 +1,65 @@
...
+Install the client library from [NuGet](https://www.nuget.org/):
+
+```dotnetcli
+dotnet add package Microsoft.Azure.Functions.Worker.Extensions.WebPubSub
+```
+
+### Prerequisites
+
+- An [Azure subscription][azure_sub].
+- An existing Azure Web PubSub service instance.
```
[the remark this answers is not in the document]
JialinXin: Looks I mistook copying from a wrong sdk readme. Let me update with a suitable one.

sdk/webpubsub/Microsoft.Azure.Functions.Worker.Extensions.WebPubSub/src/Models/Output/WebPubSubAction.cs
JialinXin: I think the common practice is not to mix use in-process and isolated-process extension packages. So either customers have to declare themselves or we do this for convenience.

See sample without built-in types and customer have to do in this way in early SignalR extension: https://github.com/Azure/azure-functions-dotnet-worker/blob/main/samples/Extensions/SignalR/SignalRFunction.cs

sdk/webpubsub/Microsoft.Azure.Functions.Worker.Extensions.WebPubSub/samples/WebPubSubConnectionInputFunction.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Net;
```
jsquire, line 4: If these are intended just as snippets for the README, please move to the standard `tests/samples` path an include the [standard verbiage](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/appconfiguration/Azure.Data.AppConfiguration/tests/samples/README.md) in the README.

JoshLove-msft:
> > Are we able to add tests?
> 
> Not much needs as these are simply type definitions. Refers to a few isolated-process extensions are usually very simple.

End-to-end testing is important for these because a lot of the assumptions we could make for the in-process extensions no longer hold. For instance, any output bindings would be delivered to the host extension as a byte array rather than a JSON string or some other JSON library primitive - do we know that the host extension will properly handle a byte array? 
This has come up recently in the Tables isolated extension - https://github.com/Azure/azure-sdk-for-net/issues/33902#issuecomment-1455554032

JialinXin:
> > > Are we able to add tests?
> > 
> > 
> > Not much needs as these are simply type definitions. Refers to a few isolated-process extensions are usually very simple.
> 
> End-to-end testing is important for these because a lot of the assumptions we could make for the in-process extensions no longer hold. For instance, any output bindings would be delivered to the host extension as a byte array rather than a JSON string or some other JSON library primitive - do we know that the host extension will properly handle a byte array? This has come up recently in the Tables isolated extension - [#33902 (comment)](https://github.com/Azure/azure-sdk-for-net/issues/33902#issuecomment-1455554032)

Thanks for sharing the context. From my E2E testing, single output is good though multiple array output will throw Object Null exceptions. I'm not able to figure out root cause yet but it could be worked around with wrapped class of multiple items like my sample shows.

benbp (approved):

JoshLove-msft:
> Thanks for sharing the context. From my E2E testing, single output is good to work though multiple array output will throw Object Null exceptions.
> image

Is there more info in the stack trace that points to where this is happening?

JialinXin:
> > Thanks for sharing the context. From my E2E testing, single output is good to work though multiple array output will throw Object Null exceptions.
> > image
> 
> Is there more info in the stack trace that points to where this is happening?

Unfortunately no. Even `--verbose` doesn't show much details. And I'm not able to attach debug to it from neither host project nor extension project, which attaches the `func.exe`.

I have a sample repo if you'd like to have a try. https://github.com/JialinXin/FunctionSamples/tree/main/IsolatedProcess/SimpleOutput

```
[2023-03-13T05:32:29.147Z] Executing 'Functions.Notification' (Reason='This function was programmatically called via the host APIs.', Id=3c4c4ca8-9a51-4d46-932e-2f4e52b650a0)
[2023-03-13T05:32:29.160Z] C# HTTP trigger function processed a request.
[2023-03-13T05:32:29.194Z] Executed 'Functions.Notification' (Failed, Id=3c4c4ca8-9a51-4d46-932e-2f4e52b650a0, Duration=45ms)
[2023-03-13T05:32:29.195Z] System.Private.CoreLib: Exception while executing function: Functions.Notification. Azure.Messaging.WebPubSub: Value cannot be null. (Parameter 'hub').
[2023-03-13T05:32:29.198Z] Executed HTTP request: {
[2023-03-13T05:32:29.199Z]   requestId: "ba066216-6cdd-4a3d-88c4-8692fdf202b3",
[2023-03-13T05:32:29.201Z]   identities: "(WebJobsAuthLevel:Admin)",
[2023-03-13T05:32:29.202Z]   status: "500",
[2023-03-13T05:32:29.203Z]   duration: "62"
[2023-03-13T05:32:29.204Z] }
```

JialinXin:
@JoshLove-msft Do you have any comments to move on review the PR? We get another few asks for isolated-process support on Web PubSub service. I believe this is due to current fact that .NET7.0 only supports isolated-process now.

JoshLove-msft (approved):

sdk/webpubsub/Microsoft.Azure.Functions.Worker.Extensions.WebPubSub/src/Models/Output/WebPubSubAction.cs
JialinXin: After a few tests, the `ActionName` can't be cached that in the case single function call multiple actions, it will always resolve the init one which is not correct. So revert to get it each time.

JialinXin:
@JoshLove-msft Appreciate the review! ❤️

I just merge with latest code to resolve the conflict and did some tests again to ensure it's working with latest. I notice a few more things about array output failure. Would you share some insights if I can improve that? Though it could be worked around by the multiple output solution. So I think it's OK if not able to support in the initial preview version if I couldn't find a way right now.

I think the problem might be it's not able to resolve the multiple different derived types under `WebPubSubAction`,
for example, if the return is `SendToAllAction[]` is good, but mixed `WebPubSubAction[]`(`SendToAllAction`, `SendToGroupAction`...) is not. Is there some special handling in the isolated process I should take care?

JoshLove-msft (approved):

drguthrie:
Sending/receiving Binary data does not appear to work. If I send X bytes, ~X*1.33 bytes are received. I would guess there is a Base64 encoding at play….however I tried to decode the received data and it failed. I have worked around it by manually Base64 encoding before sending as text then manually decoding. Note this is not ideal as the sender is a resource restricted microcontroller.

vicancy:
Hi @drguthrie, could you share more about how you "send X bytes" and how you read the received bytes?

PR #34629, opened 2023-03-01
[NetApp] Update to 2022-09-01 add tests

audunn:


sdk/netapp/Azure.ResourceManager.NetApp/api/Azure.ResourceManager.NetApp.netstandard2.0.cs
```diff
@@ -1540,4 +1596,10 @@ public partial class SnapshotPolicyWeeklySchedule
         public int? SnapshotsToKeep { get { throw null; } set { } }
         public long? UsedBytes { get { throw null; } set { } }
     }
+    public partial class VolumeRelocationProperties
+    {
+        public VolumeRelocationProperties() { }
+        public bool? ReadyToBeFinalized { get { throw null; } }
```
ArcturusZhang, line 1602: Could we rename this property to `IsReadyToBeFinalized`?

sdk/netapp/Azure.ResourceManager.NetApp/api/Azure.ResourceManager.NetApp.netstandard2.0.cs
```diff
@@ -1540,4 +1596,10 @@ public partial class SnapshotPolicyWeeklySchedule
...
+        public bool? RelocationRequested { get { throw null; } set { } }
```
ArcturusZhang, line 1603: Could we rename this property to `IsRelocationRequested`?

sdk/netapp/Azure.ResourceManager.NetApp/src/Custom/NetAppAccountResource.cs
```diff
@@ -0,0 +1,33 @@
...
+        {
+            throw null;
+        }
+
+        [Obsolete("This method is obsolete and will be removed in a future release", false)]
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public virtual Pageable<NetAppVault> GetVaults(CancellationToken cancellationToken = default)
+        {
+            throw null;
+        }
```
ArcturusZhang, line 30: We are removing this method in the new api-version, but does it still work in an old api-version?
The SDK supports api-version replacement, therefore it the API is still supported in an older api-version, we cannot just delete its implementation.
Is it still supported in an old api-version?

sdk/netapp/Azure.ResourceManager.NetApp/src/Custom/Models/NetAppVault.cs
```diff
@@ -0,0 +1,25 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+using Azure.Core;
+using Azure.ResourceManager.Models;
+
+#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
+namespace Azure.ResourceManager.NetApp.Models
+{
+    public partial class NetAppVault : ResourceData
```
ArcturusZhang, line 13: I believe these three models are moved here because they are deleted in the new api-version along with the deleted operations.

Could we add `[EditorBrowsable(Never)]` to this class? Maybe also obsolete it because it is only used in an obsolete operation.

Also please copy its previous xml document here as well and we no longer need the warning disable clause

sdk/netapp/Azure.ResourceManager.NetApp/src/Custom/Models/NetAppVault.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        public string VaultName { get; }
+
+        internal NetAppVault() { }
+        internal NetAppVault(ResourceIdentifier id, string name, ResourceType resourceType, Azure.ResourceManager.Models.SystemData systemData, string vaultName) : base(id, name, resourceType, systemData)
+        {
+            VaultName = vaultName;
+            throw null;
```
ArcturusZhang, line 21: Let us keep the implementation the same as the old version of this class.

sdk/netapp/Azure.ResourceManager.NetApp/src/Custom/Models/NetAppVolumeBackupConfiguration.cs
```diff
@@ -0,0 +1,24 @@
...
+
+#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
+namespace Azure.ResourceManager.NetApp.Models
+{
+    public partial class NetAppVolumeBackupConfiguration
+    {
+        public ResourceIdentifier VaultId { get; set; }
+
+        [Obsolete("This method is obsolete and will be removed in a future release", false)]
+        [EditorBrowsable(EditorBrowsableState.Never)]
```
ArcturusZhang, line 18: Same as the `NetAppVault` class, we should put these attributes on the class.

sdk/netapp/Azure.ResourceManager.NetApp/src/Custom/Models/NetAppVolumeBackupConfiguration.cs
```diff
@@ -0,0 +1,24 @@
...
+        internal NetAppVolumeBackupConfiguration(ResourceIdentifier backupPolicyId, bool? isPolicyEnforced, ResourceIdentifier vaultId, bool? isBackupEnabled)
+        {
+        }
```
ArcturusZhang, line 21: And we need to keep the implementation the same

sdk/netapp/Azure.ResourceManager.NetApp/src/Custom/Models/NetAppVolumeBackupConfiguration.cs
```diff
@@ -0,0 +1,24 @@
...
+    }
+}
+#pragma warning restore CS1591 // Missing XML comment for publicly visible type or member
```
ArcturusZhang, line 24: And we need to copy its previous documentation back so that we could remove this warning suppression

sdk/netapp/Azure.ResourceManager.NetApp/src/Custom/Models/NetAppVolumeDataProtection.cs
```diff
@@ -0,0 +1,21 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using Azure.ResourceManager.NetApp.Models;
+using static System.Net.Mime.MediaTypeNames;
+
+namespace Azure.ResourceManager.NetApp.Models
+{
+    public partial class NetAppVolumeDataProtection
```
ArcturusZhang, line 12: Same, please add EBNever attribute here

sdk/netapp/Azure.ResourceManager.NetApp/src/Custom/NetAppAccountResource.cs
```diff
@@ -0,0 +1,33 @@
...
+using Azure.ResourceManager.NetApp.Models;
+
+#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
+namespace Azure.ResourceManager.NetApp
+{
+    public partial class NetAppAccountResource: ArmResource
+    {
+        [Obsolete("This method is obsolete and will be removed in a future release", false)]
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public virtual AsyncPageable<NetAppVault> GetVaultsAsync(CancellationToken cancellationToken = default)
```
ArcturusZhang, line 20: Please put back its document so that we could remove the warning suppression

sdk/netapp/Azure.ResourceManager.NetApp/tests/ResourceTests/ANFBackupPolicyTests.cs
```diff
@@ -176,7 +171,7 @@ public async Task CreateVolumeWithBackupPolicy()
             {
                 DefaultVirtualNetwork = await CreateVirtualNetwork();
             }
-            NetAppVolumeBackupConfiguration backupPolicyProperties = new(backupPolicyResource1.Id, false, vault.Id, true);
+            NetAppVolumeBackupConfiguration backupPolicyProperties = new(backupPolicyResource1.Id, false, true);
```
ArcturusZhang, line 174: I just noticed this, are we using the internal constructors here?
We should not using internal constructors in test cases despite we could.
This might make things different from it is used by our customers who cannot use internal members.
And using internal constructors makes you hard to maintain your test cases because every time you introduce a new property, you might have to change your implementation despite it is not a breaking change

sdk/netapp/Azure.ResourceManager.NetApp/tests/ResourceTests/ANFBackupPolicyTests.cs
```diff
@@ -176,7 +171,7 @@ public async Task CreateVolumeWithBackupPolicy()
...
             NetAppVolumeDataProtection dataProtectionProperties = new NetAppVolumeDataProtection(backup: backupPolicyProperties, null,null);
```
ArcturusZhang, line 175: Same here

ArcturusZhang (commented):
The PR generally LGTM
But we need to bring the implementation of the removed operation back if it still works in an older api-version.

And I noticed that we are widely using internal constructors in our test cases, this is not recommended which increases the maintenance cost of the test cases and makes things different from when those are used by our customers.
But this is not urgent, we could come back to that later

sdk/netapp/Azure.ResourceManager.NetApp/src/Custom/NetAppAccountResource.cs
audunn: yes, it is supported in older api-version.  How would we call using previous api-versoin ?

sdk/netapp/Azure.ResourceManager.NetApp/api/Azure.ResourceManager.NetApp.netstandard2.0.cs
audunn: We should probably be able to do that in autorest.md config right?

sdk/netapp/Azure.ResourceManager.NetApp/src/Custom/NetAppAccountResource.cs
audunn: Added back using older api-version

audunn:
@ArcturusZhang  Are we good to move on with this or is there more work to be done?

sdk/netapp/Azure.ResourceManager.NetApp/CHANGELOG.md
```diff
@@ -3,9 +3,19 @@
...
+- Added propperty PreferredServersForLdapClient to NetAppAccountActiveDirectory, a comma separated list of IPv4 addresses of preferred servers for LDAP client
+- VolumeRelocation to NetAppVolumeDataProtection
+- DataStoreResourceId to NetAppVolumeGroupVolume
+- Added Tags to NetAppVolumeQuotaRulePatch and AddTag to NetAppVolumeQuotaResource
+- Added RestoreFiles for NetAppVolumeBackupResource
 
-### Breaking Changes
 
+### Breaking Changes
+- NetAppVault is no longer needed scheduled for deprication
```
ArcturusZhang, line 17: here is a typo
```suggestion
- NetAppVault is no longer needed scheduled for deprecation
```

sdk/netapp/Azure.ResourceManager.NetApp/CHANGELOG.md
```diff
@@ -3,9 +3,19 @@
 ## 1.1.0-beta.1 (Unreleased)
 
 ### Features Added
+- Added BackupRestoreFiles to NetAppVolumeResource NetAppVolumeBackupResource to rstore the specified files from the specified backup to the active filesystem
```
ArcturusZhang, line 6: typos
```suggestion
- Added BackupRestoreFiles to NetAppVolumeResource NetAppVolumeBackupResource to restore the specified files from the specified backup to the active file system
```

sdk/netapp/Azure.ResourceManager.NetApp/CHANGELOG.md
```diff
@@ -3,9 +3,19 @@
...
+- Added BreakFileLocks to NetAppVolumeResource to allow clients to break file locks on a volume
+- Added FileAccessLog to NetAppVolumeData, a lag indicating whether file access logs are enabled for the volume, based on active diagnostic settings present on the volume.
+- Added propperty PreferredServersForLdapClient to NetAppAccountActiveDirectory, a comma separated list of IPv4 addresses of preferred servers for LDAP client
+- VolumeRelocation to NetAppVolumeDataProtection
+- DataStoreResourceId to NetAppVolumeGroupVolume
```
ArcturusZhang, line 11: what do these two lines mean? Changed VolumeRelocation to NetAppVolumeDataProtection?

sdk/netapp/Azure.ResourceManager.NetApp/CHANGELOG.md
```diff
@@ -3,9 +3,19 @@
...
+- Added Tags to NetAppVolumeQuotaRulePatch and AddTag to NetAppVolumeQuotaResource
+- Added RestoreFiles for NetAppVolumeBackupResource
 
-### Breaking Changes
 
+### Breaking Changes
+- NetAppVault is no longer needed scheduled for deprication
+- Plan to depricate vaultId from NetAppVolumeBackupConfiguration
```
ArcturusZhang, line 18:
```suggestion
- Plan to deprecate vaultId from NetAppVolumeBackupConfiguration
```

sdk/netapp/Azure.ResourceManager.NetApp/api/Azure.ResourceManager.NetApp.netstandard2.0.cs
```diff
@@ -330,18 +334,21 @@ public partial class NetAppVolumeData : Azure.ResourceManager.Models.TrackedReso
         public int? CoolnessPeriod { get { throw null; } set { } }
         public string CreationToken { get { throw null; } set { } }
         public Azure.ResourceManager.NetApp.Models.NetAppVolumeDataProtection DataProtection { get { throw null; } set { } }
+        public System.Collections.Generic.IReadOnlyList<string> DataStoreResourceId { get { throw null; } }
```
ArcturusZhang, line 337: Does the elements of this array all resource identifiers as the name suggested? If so, please change the format of this property to `arm-id` so that it has proper type of `IReadOnlyList<ResourceIdentifier>`

sdk/netapp/Azure.ResourceManager.NetApp/CHANGELOG.md
```diff
@@ -3,9 +3,19 @@
 ## 1.1.0-beta.1 (Unreleased)
 
 ### Features Added
+- Added BackupRestoreFiles to NetAppVolumeResource NetAppVolumeBackupResource to rstore the specified files from the specified backup to the active filesystem
+- Added BreakFileLocks to NetAppVolumeResource to allow clients to break file locks on a volume
+- Added FileAccessLog to NetAppVolumeData, a lag indicating whether file access logs are enabled for the volume, based on active diagnostic settings present on the volume.
+- Added propperty PreferredServersForLdapClient to NetAppAccountActiveDirectory, a comma separated list of IPv4 addresses of preferred servers for LDAP client
+- VolumeRelocation to NetAppVolumeDataProtection
+- DataStoreResourceId to NetAppVolumeGroupVolume
+- Added Tags to NetAppVolumeQuotaRulePatch and AddTag to NetAppVolumeQuotaResource
```
ArcturusZhang, line 12:
```suggestion
- Added `Tags` property to NetAppVolumeQuotaRulePatch and `AddTag`, `RemoveTag` and `SetTags` to NetAppVolumeQuotaResource
```

sdk/netapp/Azure.ResourceManager.NetApp/src/Custom/Models/NetAppVault.Serialization.cs
```diff
@@ -69,7 +67,7 @@ internal static NetAppVault DeserializeNetAppVault(JsonElement element)
                     continue;
                 }
             }
-            return new NetAppVault(id, name, type, systemData.Value, vaultName.Value);
+            return new NetAppVault(id, name, type, systemData.Value, name);
```
ArcturusZhang, line 70: why we change the vaultName to name here? I believe this will break the behaviour on lower api-versions.

sdk/netapp/Azure.ResourceManager.NetApp/src/Custom/Models/NetAppVault.cs
```diff
@@ -1,35 +1,35 @@
...
+        internal NetAppVault() { }
 
         /// <summary> Initializes a new instance of NetAppVault. </summary>
         /// <param name="id"> The id. </param>
         /// <param name="name"> The name. </param>
         /// <param name="resourceType"> The resourceType. </param>
         /// <param name="systemData"> The systemData. </param>
         /// <param name="vaultName"> Vault Name. </param>
-        internal NetAppVault(ResourceIdentifier id, string name, ResourceType resourceType, SystemData systemData, string vaultName) : base(id, name, resourceType, systemData)
+        [EditorBrowsable(EditorBrowsableState.Never)]
```
ArcturusZhang, line 29: we do not need this if the class itself already has this attribute

sdk/netapp/Azure.ResourceManager.NetApp/src/Custom/Models/NetAppVault.cs
```diff
@@ -1,35 +1,35 @@
...
+    [EditorBrowsable(EditorBrowsableState.Never)]
     public partial class NetAppVault : ResourceData
     {
-        /// <summary> Initializes a new instance of NetAppVault. </summary>
-        internal NetAppVault()
-        {
-        }
+        /// <summary> Vault Name. </summary>
+       // [EditorBrowsable(EditorBrowsableState.Never)]
+        public string VaultName { get; }
```
ArcturusZhang, line 19: Could we have a format, clean up and reordering of its elements to minimize the diff in this file?
We could have just two changes in this file: adding EBNever attribute on this class and removing the `auto-generated` comment label on line 4

sdk/netapp/Azure.ResourceManager.NetApp/src/Custom/Models/VaultList.cs
```diff
@@ -1,16 +1,18 @@
...
 
+using System;
 using System.Collections.Generic;
+using System.ComponentModel;
 using Azure.Core;
 
 namespace Azure.ResourceManager.NetApp.Models
 {
     /// <summary> List of Vaults. </summary>
+    [EditorBrowsable(EditorBrowsableState.Never)]
```
ArcturusZhang, line 14: Since this class is internal, our customer will never see it and we do not need this change

sdk/netapp/Azure.ResourceManager.NetApp/src/Custom/Models/NetAppVolumeBackupConfiguration.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Text;
+using Azure.Core;
+
+namespace Azure.ResourceManager.NetApp.Models
+{
+    /// <summary> Volume Backup Properties. </summary>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public partial class NetAppVolumeBackupConfiguration
```
ArcturusZhang, line 14: Do we have its generated version previously? Why I did not see it gets renamed to this here?

sdk/netapp/Azure.ResourceManager.NetApp/src/Custom/Models/NetAppVolumeDataProtection.cs
```diff
@@ -0,0 +1,29 @@
...
+using System.ComponentModel;
+using System.Text;
+using Azure.ResourceManager.NetApp.Models;
+using static System.Net.Mime.MediaTypeNames;
+
+namespace Azure.ResourceManager.NetApp.Models
+{
+    /// <summary> DataProtection type volumes include an object containing details of the replication. </summary>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public partial class NetAppVolumeDataProtection
```
ArcturusZhang, line 15: Do we have its generated version previously? Why I did not see it gets renamed to this here?

sdk/netapp/Azure.ResourceManager.NetApp/src/Custom/RestOperations/VaultsRestOperations.cs
```diff
@@ -16,6 +15,8 @@
 
 namespace Azure.ResourceManager.NetApp
 {
+    [EditorBrowsable(EditorBrowsableState.Never)]
+   // [Obsolete("This class is obsolete and will be removed in a future release", false)]
```
ArcturusZhang, line 19: Since this class is internal, it is hidden all the time, therefore we do not need these attribute, or Obsolete attribute

sdk/netapp/Azure.ResourceManager.NetApp/CHANGELOG.md
audunn: added

sdk/netapp/Azure.ResourceManager.NetApp/src/Custom/Models/NetAppVolumeBackupConfiguration.cs
audunn: Not sure, it has not been  renamed 

https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/netapp/Azure.ResourceManager.NetApp/src/Generated/Models/NetAppVolumeBackupConfiguration.cs

sdk/netapp/Azure.ResourceManager.NetApp/src/Custom/Models/NetAppVolumeDataProtection.cs
audunn: not being renamed
https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/netapp/Azure.ResourceManager.NetApp/src/Generated/Models/NetAppVolumeDataProtection.cs

sdk/netapp/Azure.ResourceManager.NetApp/src/Custom/Models/NetAppVolumeBackupConfiguration.cs
ArcturusZhang: I see, this class is not removed in new version. Only this property `public ResourceIdentifier VaultId { get; set; }` is removed.
Then, we cannot add `[EditorBrowsable(EditorBrowsableState.Never)]` attribute on the class (because it is not deprecated), instead, we need to add this attribute to the property.

One question: will the model still return `vaultId` on an older api-version?
If the answer is yes, we need to move its deserialization to customization code directory, and add the deserialization of this property `vaultId` back to that method.

sdk/netapp/Azure.ResourceManager.NetApp/src/Custom/Models/NetAppVolumeDataProtection.cs
ArcturusZhang: I withdraw my previous comment on this class. In the new version is just something new added to this property, we do not need to add anything "internal" back because customers could never use them
Therefore please just remove this file because it does not need any customization.

sdk/netapp/Azure.ResourceManager.NetApp/src/Custom/Models/VaultList.Serialization.cs
```diff
@@ -1,11 +1,10 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
ArcturusZhang, line 1: Since nothing public changed in this file, we could just remove this file from this directory and regenerate to take its generated version.

sdk/netapp/Azure.ResourceManager.NetApp/src/Custom/Models/VaultList.cs
```diff
@@ -1,11 +1,11 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
ArcturusZhang, line 1: Since nothing public changed in this file, we could just remove this file from this directory and regenerate to take its generated version.

sdk/netapp/Azure.ResourceManager.NetApp/src/Custom/NetAppAccountResource.cs
```diff
@@ -0,0 +1,92 @@
...
+        //            _netAppVolumeGroupVolumeGroupsRestClient = new VolumeGroupsRestOperations(Pipeline, Diagnostics.ApplicationId, Endpoint, netAppVolumeGroupVolumeGroupsApiVersion);
+        //#if DEBUG
+        //            ValidateResourceId(Id);
+        //#endif
+        //        }
+        private string GetApiVersionOrNull(ResourceType resourceType)
+        {
+            TryGetApiVersion(resourceType, out string apiVersion);
+            return apiVersion;
+        }
```
ArcturusZhang, line 46: since we already have this method on the generated file and you did not change anything on this method, we could just remove this here, and regenerate to let the generated code handle this.

sdk/netapp/Azure.ResourceManager.NetApp/src/Custom/NetAppAccountResource.cs
```diff
@@ -0,0 +1,92 @@
...
+        //            _netAppAccountAccountsRestClient = new AccountsRestOperations(Pipeline, Diagnostics.ApplicationId, Endpoint, netAppAccountAccountsApiVersion);
+        //            _vaultsClientDiagnostics = new ClientDiagnostics("Azure.ResourceManager.NetApp", ProviderConstants.DefaultProviderNamespace, Diagnostics);
+        //            _vaultsRestClient = new VaultsRestOperations(Pipeline, Diagnostics.ApplicationId, Endpoint);
+        //            _netAppVolumeGroupVolumeGroupsClientDiagnostics = new ClientDiagnostics("Azure.ResourceManager.NetApp", NetAppVolumeGroupResource.ResourceType.Namespace, Diagnostics);
+        //            TryGetApiVersion(NetAppVolumeGroupResource.ResourceType, out string netAppVolumeGroupVolumeGroupsApiVersion);
+        //            _netAppVolumeGroupVolumeGroupsRestClient = new VolumeGroupsRestOperations(Pipeline, Diagnostics.ApplicationId, Endpoint, netAppVolumeGroupVolumeGroupsApiVersion);
+        //#if DEBUG
+        //            ValidateResourceId(Id);
+        //#endif
+        //        }
```
ArcturusZhang, line 41: Please remove these comment here

sdk/netapp/Azure.ResourceManager.NetApp/src/Custom/RestOperations/VaultsRestOperations.cs
```diff
@@ -65,6 +64,7 @@ internal HttpMessage CreateListRequest(string subscriptionId, string resourceGro
         /// <param name="cancellationToken"> The cancellation token to use. </param>
         /// <exception cref="ArgumentNullException"> <paramref name="subscriptionId"/>, <paramref name="resourceGroupName"/> or <paramref name="accountName"/> is null. </exception>
         /// <exception cref="ArgumentException"> <paramref name="subscriptionId"/>, <paramref name="resourceGroupName"/> or <paramref name="accountName"/> is an empty string, and was expected to be non-empty. </exception>
+        [Obsolete("This method is obsolete and not needed anymore and will be removed in a future release", false)]
```
ArcturusZhang, line 67: since this class is internal, we do not need to mark it as Obsolete here.

sdk/netapp/Azure.ResourceManager.NetApp/src/Custom/Models/VaultList.Serialization.cs
audunn: This will not be generated since the API has been removed from the latest api-version swagger.

sdk/netapp/Azure.ResourceManager.NetApp/src/Custom/Models/VaultList.cs
```diff
@@ -1,11 +1,11 @@
...
 // Licensed under the MIT License.
 
-// <auto-generated/>
-
 #nullable disable
 
+using System;
 using System.Collections.Generic;
+using System.ComponentModel;
```
ArcturusZhang, line 8: Since these using are not used, please remove these

sdk/netapp/Azure.ResourceManager.NetApp/src/Custom/Models/NetAppVolumeBackupConfiguration.Serialization.cs
```diff
@@ -0,0 +1,90 @@
...
+            {
+                writer.WritePropertyName("policyEnforced");
+                writer.WriteBooleanValue(IsPolicyEnforced.Value);
+            }
+            if (Optional.IsDefined(IsBackupEnabled))
+            {
+                writer.WritePropertyName("backupEnabled");
+                writer.WriteBooleanValue(IsBackupEnabled.Value);
+            }
+            writer.WriteEndObject();
```
ArcturusZhang, line 31: we need to add a segment to write vaultId in this method otherwise it will not work for older api-versions

sdk/netapp/Azure.ResourceManager.NetApp/api/Azure.ResourceManager.NetApp.netstandard2.0.cs
```diff
@@ -601,6 +617,19 @@ public partial class AvailabilityZoneMapping
...
         public bool? IsAvailable { get { throw null; } }
     }
+    public partial class BackupRestoreFiles
+    {
+        public BackupRestoreFiles(System.Collections.Generic.IEnumerable<string> fileList, string destinationVolumeId) { }
+        public string DestinationVolumeId { get { throw null; } }
+        public System.Collections.Generic.IList<string> FileList { get { throw null; } }
+        public string RestoreFilePath { get { throw null; } set { } }
+    }
+    public partial class BreakFileLocksContent
```
ArcturusZhang, line 627: Could we prepend the name of the specific resource name that using this to specify it? It should be `NetAppVolumeBreakFileLocksContent`

sdk/netapp/Azure.ResourceManager.NetApp/api/Azure.ResourceManager.NetApp.netstandard2.0.cs
```diff
@@ -601,6 +617,19 @@ public partial class AvailabilityZoneMapping
         public string AvailabilityZone { get { throw null; } }
         public bool? IsAvailable { get { throw null; } }
     }
+    public partial class BackupRestoreFiles
```
ArcturusZhang, line 620: Could we rename this to `NetAppVolumeBackupBackupRestoreFilesContent`

sdk/netapp/Azure.ResourceManager.NetApp/api/Azure.ResourceManager.NetApp.netstandard2.0.cs
```diff
@@ -662,6 +691,24 @@ public partial class CapacityPoolPatch : Azure.ResourceManager.Models.TrackedRes
         public static bool operator !=(Azure.ResourceManager.NetApp.Models.EnableNetAppSubvolume left, Azure.ResourceManager.NetApp.Models.EnableNetAppSubvolume right) { throw null; }
         public override string ToString() { throw null; }
     }
+    [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
+    public readonly partial struct FileAccessLog : System.IEquatable<Azure.ResourceManager.NetApp.Models.FileAccessLog>
```
ArcturusZhang, line 695: This is kind of a generic name - could we rename this to something like `NetappFileAccessLog`

sdk/netapp/Azure.ResourceManager.NetApp/api/Azure.ResourceManager.NetApp.netstandard2.0.cs
```diff
@@ -601,6 +617,19 @@ public partial class AvailabilityZoneMapping
         public string AvailabilityZone { get { throw null; } }
         public bool? IsAvailable { get { throw null; } }
     }
+    public partial class BackupRestoreFiles
+    {
+        public BackupRestoreFiles(System.Collections.Generic.IEnumerable<string> fileList, string destinationVolumeId) { }
+        public string DestinationVolumeId { get { throw null; } }
```
ArcturusZhang, line 623: Is the value of this property an arm resource identifier?
If so, please change the format of this property to `arm-id`

sdk/netapp/Azure.ResourceManager.NetApp/api/Azure.ResourceManager.NetApp.netstandard2.0.cs
```diff
@@ -601,6 +617,19 @@ public partial class AvailabilityZoneMapping
...
+        public System.Collections.Generic.IList<string> FileList { get { throw null; } }
+        public string RestoreFilePath { get { throw null; } set { } }
+    }
+    public partial class BreakFileLocksContent
+    {
+        public BreakFileLocksContent() { }
+        public string ClientIP { get { throw null; } set { } }
```
ArcturusZhang, line 630: Is the value of this property in the format of ip address?
If so, please change the format of this property to `ip-address`

sdk/netapp/Azure.ResourceManager.NetApp/api/Azure.ResourceManager.NetApp.netstandard2.0.cs
```diff
@@ -1540,4 +1596,10 @@ public partial class SnapshotPolicyWeeklySchedule
         public int? SnapshotsToKeep { get { throw null; } set { } }
         public long? UsedBytes { get { throw null; } set { } }
     }
+    public partial class VolumeRelocationProperties
```
ArcturusZhang, line 1599: Since in this RP, Volume corresponds to `NetappVolume`, could we rename this accordingly?
```suggestion
    public partial class NetappVolumeRelocationProperties
```

sdk/netapp/Azure.ResourceManager.NetApp/src/Custom/Models/NetAppVolumeBackupConfiguration.Serialization.cs
```diff
@@ -0,0 +1,95 @@
...
+                    if (property.Value.ValueKind == JsonValueKind.Null)
+                    {
+                        property.ThrowNonNullablePropertyIsNull();
+                        continue;
+                    }
+                    vaultId = new ResourceIdentifier(property.Value.GetString());
+                    continue;
+                }
+            }
+            return new NetAppVolumeBackupConfiguration(backupPolicyId.Value, Optional.ToNullable(policyEnforced), Optional.ToNullable(backupEnabled)){ VaultId = vaultId };
```
ArcturusZhang, line 92: Could we add a constructor that takes the extra parameter of vaultId and use that here? It should look like:
```suggestion
return new NetAppVolumeBackupConfiguration(backupPolicyId.Value, Optional.ToNullable(policyEnforced), Optional.ToNullable(backupEnabled), vaultId.Value);
```

sdk/netapp/Azure.ResourceManager.NetApp/src/Custom/Models/NetAppVolumeBackupConfiguration.cs
```diff
@@ -0,0 +1,19 @@
...
+using Azure.Core;
+
+namespace Azure.ResourceManager.NetApp.Models
+{
+    /// <summary> Volume Backup Properties. </summary>
+    public partial class NetAppVolumeBackupConfiguration
+    {
+        /// <summary> Vault Resource ID. </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public ResourceIdentifier VaultId { get; set; }
```
ArcturusZhang, line 17: Can we add an internal constructor here by calling the existing internal constructor with an extra parameter of `vaultId`?

sdk/netapp/Azure.ResourceManager.NetApp/tests/ResourceTests/ANFBackupTests.cs
```diff
@@ -136,24 +125,27 @@ public async Task ClearVolumes()
...
                 await _netAppAccount.DeleteAsync(WaitUntil.Completed);
             }
+            watch.Stop();
+            TestContext.WriteLine($"TearDown elapsed time {watch.ElapsedMilliseconds} ms {watch.Elapsed}");
             _resourceGroup = null;
         }
 
         [Test]
-        [Ignore("Ignore for now due to service side issue, re-enable when service side issue is fixed")]
+        //[Ignore("Ignore for now due to service side issue, re-enable when service side issue is fixed")]
```
ArcturusZhang, line 134: maybe we could delete this line since this is fixed now

sdk/netapp/Azure.ResourceManager.NetApp/tests/NetAppManagementTestBase.cs
```diff
@@ -12,14 +12,19 @@ namespace Azure.ResourceManager.NetApp.Tests
 {
     public class NetAppManagementTestBase : ManagementRecordedTestBase<NetAppManagementTestEnvironment>
     {
-        protected ArmClient Client { get; private set; }
+        protected ArmClient Client { get; set; }
```
ArcturusZhang, line 15: I did not see a place that we are setting the Client
And actually we should not set the Client, because in our test framework, the ArmClient instance is instrumented to support various of validations and playback of the test cases, we should not replace the instance of ArmClient in any test cases.
Please add the private accessor back so that we could ensure only in this class, we will set the reference of `Client`.

sdk/netapp/Azure.ResourceManager.NetApp/api/Azure.ResourceManager.NetApp.netstandard2.0.cs
audunn: This is not a resource it is a body payload to an `POST` action, should we still rename it?

ArcturusZhang (approved):

ArcturusZhang (approved):

PR #34633, opened 2023-03-01
Clarify CertificateContentType comment

heaths:
Relates to Azure/azure-sdk-for-cpp#4394

### DIFF
--- a/sdk/keyvault/Azure.Security.KeyVault.Certificates/src/CertificateContentType.cs
+++ b/sdk/keyvault/Azure.Security.KeyVault.Certificates/src/CertificateContentType.cs
@@ -7,7 +7,7 @@ using System.ComponentModel;
 namespace Azure.Security.KeyVault.Certificates
 {
     /// <summary>
-    /// Content type of the certificate when downloaded from getSecret.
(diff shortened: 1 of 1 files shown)


sdk/keyvault/Azure.Security.KeyVault.Certificates/src/CertificateContentType.cs
```diff
@@ -7,7 +7,7 @@
 namespace Azure.Security.KeyVault.Certificates
 {
     /// <summary>
-    /// Content type of the certificate when downloaded from getSecret.
+    /// Content type of the certificate when the managed secret is downloaded using a <c>SecretClient</c>.
```
christothes, line 10: nit: Without the `cref`, is it worth wrapping this in a code tag?

christothes (approved):

sdk/keyvault/Azure.Security.KeyVault.Certificates/src/CertificateContentType.cs
heaths: IMO, makes it more obvious it's a bit o' code. Obviously don't want a package dependency - especially for a comment - so I felt this was more obvious and a compromise.

PR #34636, opened 2023-03-01
Add missing region for Statsbeat

vishweshbankwar:


cijothomas (approved):

rajkumar-rangaraj (approved):

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/Statsbeat/StatsbeatConstants.cs
```diff
@@ -15,6 +15,7 @@ internal static class StatsbeatConstants
         {
             "francecentral",
             "francesouth",
+            "germanywestcentral",
```
reyang, line 18: What's the expected behavior when the customer is using an older version of the SDK which doesn't have this line? (will the data be simply dropped?)

vishweshbankwar: Yes, We will not initialize statsbeat in such cases. The region has to exactly match either one of EU regions or NonEU regions.

PR #34641, opened 2023-03-01
Switch Key Vault canary region to Central US

heaths:


heaths:
Talking with @benbp offline, we'll need this change anyway since the linux App Service is only available in centraluseuap anyway.

benbp:
@heaths my onboarding ticket is still pending.

christothes (approved):

sdk/keyvault/tests.yml
```diff
@@ -12,7 +12,7 @@ extends:
         SubscriptionConfiguration: $(sub-config-azure-cloud-test-resources)
       Canary:
         SubscriptionConfiguration: $(sub-config-azure-cloud-test-resources)
-        Location: 'eastus2euap'
+        Location: 'centraluseuap'
```
pallavit, line 15: Do we have to be in canary? Was this mostly to test early? Canary tends to always be less stable than the production environment so was curious.  Also is this a .NET decision or something that we do across? I see different SDKs doing it differently.

heaths: Canary was an intentional decision, yes. We provide backstop testing so it helps spot any regressions early.

pallavit: got it, thanks.

heaths:
The problems with China are all known timeouts that started happening about a month ago. Tracking resolution in https://github.com/Azure/azure-sdk-tools/issues/6175. Will override.

PR #34645, opened 2023-03-01
Disable redirects by default and add options to configure redirects

christothes:
# Summary of changes
The `RedirectPolicy` no longer follows redirects by default when using the default Transport. This behavior is configurable via new options in two scenarios:
- Customers choosing to enable or disable redirects globally via `ClientOptions.ClientRedirects.IsClientRedirectEnabled`. This setting takes precedence over the following setting.
- Client authors choosing to enable or disable redirects globally via `HttpPipelineTransportOptions.IsAutoRedirectEnabled`

Because the `RedirectPolicy` is still added to the pipeline in all scenarios, the per-message ` RedirectPolicy.SetAllowAutoRedirect(...)` method's behavior is unchanged.


sdk/core/Azure.Core/api/Azure.Core.net6.0.cs
```diff
@@ -518,6 +519,11 @@ public static partial class MultipartResponse
         public static Azure.Response[] Parse(Azure.Response response, bool expectCrLf, System.Threading.CancellationToken cancellationToken) { throw null; }
         public static System.Threading.Tasks.Task<Azure.Response[]> ParseAsync(Azure.Response response, bool expectCrLf, System.Threading.CancellationToken cancellationToken) { throw null; }
     }
+    public partial class RedirectPolicyOptions
+    {
+        public RedirectPolicyOptions() { }
+        public bool IsAutoRedirectEnabled { get { throw null; } set { } }
```
KrzysztofCwalina, line 525: I would call it IsClientRedirectEnabled to hint that it does not control networking stack redirects but just the code of the client.

sdk/core/Azure.Core/api/Azure.Core.net6.0.cs
```diff
@@ -381,6 +381,7 @@ public abstract partial class ClientOptions
         protected ClientOptions(Azure.Core.DiagnosticsOptions? diagnostics) { }
         public static Azure.Core.ClientOptions Default { get { throw null; } }
         public Azure.Core.DiagnosticsOptions Diagnostics { get { throw null; } }
+        public Azure.Core.RedirectPolicyOptions? RedirectPolicyOptions { get { throw null; } set { } }
```
KrzysztofCwalina, line 384: The other properties don't end in "Options". What about calling it ClientRedirects?

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Generated/ContainerRegistryBlobRestClient.cs
```diff
@@ -35,23 +35,6 @@ public ContainerRegistryBlobRestClient(ClientDiagnostics clientDiagnostics, Http
             _url = url ?? throw new ArgumentNullException(nameof(url));
         }
 
-        internal HttpMessage CreateGetBlobRequest(string name, string digest)
```
annelo-msft, line 38: Is there a way to approach the change to generated methods that doesn't require re-implementing them manually?

christothes: That would be great, but this is typically the tradeoff when needing to change the shape of generated requests. Do you have any suggestions to accomplish this in a better way?

sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -518,6 +519,11 @@ public static partial class MultipartResponse
         public static Azure.Response[] Parse(Azure.Response response, bool expectCrLf, System.Threading.CancellationToken cancellationToken) { throw null; }
         public static System.Threading.Tasks.Task<Azure.Response[]> ParseAsync(Azure.Response response, bool expectCrLf, System.Threading.CancellationToken cancellationToken) { throw null; }
     }
+    public partial class RedirectPolicyOptions
```
annelo-msft, line 522: Is this analogous to RetryOptions?  If it is, would it make sense to drop `Policy` from the name?  Or is communicating that these only apply to the policy important?  It doesn't look like we have `PolicyOptions` elsewhere in the Core API, so this feels like introducing a new pattern we might want to adhere to in the future ... is that the goal?

sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -518,6 +519,11 @@ public static partial class MultipartResponse
...
+    {
+        public RedirectPolicyOptions() { }
+        public bool IsClientRedirectEnabled { get { throw null; } set { } }
```
annelo-msft, line 525: I'm curious about the inclusion of `Is` in this name ... what is the thinking there vs. something like `EnableClientRedirect`?  `IsXx` implies a query regarding the state of an object (like type.IsPrimitive, e.g.) vs something a caller could pass an an input to change the behavior (e.g. `AllowSomethingToHappen`).  Is the intended usage here to ask if it is enabled?

sdk/core/Azure.Core/api/Azure.Core.net461.cs
christothes: Initially this was named relevant to the transport, however since this is only affecting the behavior of the RedirectPolicy itself, @KrzysztofCwalina suggested we scope the name to that to be explicit about what is being affected.

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/ContainerRegistryBlobRestClient.cs
```diff
@@ -0,0 +1,80 @@
...
+        /// <summary>
+        /// Replace the default implementation to call RedirectPolicy.SetAllowAutoRedirect(message, true).
+        /// </summary>
+        /// <param name="name"></param>
+        /// <param name="digest"></param>
+        /// <returns></returns>
+        internal HttpMessage CreateGetBlobRequest(string name, string digest)
+        {
+            var message = _pipeline.CreateMessage();
+            var request = message.Request;
```
annelo-msft, line 22: Was `message.BufferResponse = false;` removed intentionally?

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Generated/ContainerRegistryBlobRestClient.cs
annelo-msft: Could there be a way to indicate to the generator that there's a need to follow redirects in an API?

sdk/core/Azure.Core/api/Azure.Core.net461.cs
christothes: I don't 😄

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/ContainerRegistryBlobRestClient.cs
christothes: 😮 no it wasn't. I'm not sure how that even happened, considering it was copy/pasted! Good catch.

annelo-msft: Cool, ok, I wasn't sure if something about the new behavior made that necessary.  But this kind of mistake that's super easy to make unintentionally is essentially why I'm concerned about manually overriding these methods.  We've done a lot of configuring of the generated layer via generator knobs, and moving to manual has the effect of taking away that as the approach for implementing the methods, now, and in the future.  It'd be easy for a new dev to see those customizations and think that's the right place to make changes, then head scratch for a long time regarding why those changes aren't getting picked up.  Or any updates to the generator to then not get picked up in a new build, etc.

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Generated/ContainerRegistryBlobRestClient.cs
annelo-msft: We sometimes need to work with the  generator team to get them to use new features we add to Core.  It might be worth initiating that conversation.  Is this something that might needed in a fully generated client?

So as not to block this on that functionality, I'd also be open to a customization we could make in the outer HCL client if we could find a way to do it using that approach

christothes: I think this is a bit of an edge case. For clients that want this configured globally (see below), we have an option now. This only pertains to clients which need this on a per-operation basis. ACR is the only example, I believe.

https://github.com/Azure/azure-sdk-for-net/blob/6ad5dbbb1b5240e1aa7d9bb5dfd912371d69ddac/sdk/communication/Azure.Communication.CallAutomation/src/CustomBuildHttpPieline.cs#L18-L19

annelo-msft: Ah, interesting about the global option.  Why did we decide not to enable it globally on ACR?

christothes: We certainly could do that if you feel better about that. I was just trying to minimize the need for redirects. I'll defer to your judgement here.

annelo-msft: Help me understand this better, though, if you don't mind, since I'm still not sure what the best path forward is.  If the goal is to minimize the need for redirects, this implementation addresses that by only enabling redirects on messages for operations that can turn a 3xx response (is that correct?  I'm assuming that's how you chose the operations to customize, but I could totally be wrong).  But if all other operations don't return 3xx (or would throw if a non-2xx response is returned), is there a problem with enabling it for them too?  The latter is a serious question -- not rhetorical -- if there's a problem, I would want to know it before saying I think it's the right thing to enable it globally.  What risks do we take on if we were to do that?

christothes: > Help me understand this better, though, if you don't mind, since I'm still not sure what the best path forward is. If the goal is to minimize the need for redirects, this implementation addresses that by only enabling redirects on messages for operations that can turn a 3xx response (is that correct? I'm assuming that's how you chose the operations to customize, but I could totally be wrong). 

Correct

> But if all other operations don't return 3xx (or would throw if a non-2xx response is returned), is there a problem with enabling it for them too? The latter is a serious question -- not rhetorical -- if there's a problem, I would want to know it before saying I think it's the right thing to enable it globally. What risks do we take on if we were to do that?

In theory, if we know there will never be a redirect, we don't care if redirects are enabled. However, the core issue is that for untrusted URIs, we can be fail safe that an unexpected redirect will not be followed by default, hence the desire to minimize exposure to this.

annelo-msft: > In theory, if we know there will never be a redirect, we don't care if redirects are enabled. However, the core issue is that for untrusted URIs, we can be fail safe that an unexpected redirect will not be followed by default, hence the desire to minimize exposure to this.

Yeah, I think it would be safe to assume that even if the swagger says the service won't return a 3xx, we could maybe get one.  But, it sounds like we feel safe that if we did get a 3xx for some operation where it's not allowed, the generated code would have us throw before it was followed, so we'd be safe there too?

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/ContainerRegistryBlobRestClient.cs
annelo-msft: That's fair, we may want to wait and see if other libraries need this functionality before adding it to the generator.

I'd still like to see if we can find a solution other than overriding the generated methods for this library.

christothes: The other option is to do this globally for the whole client at pipeline creation time. as discussed [here](https://github.com/Azure/azure-sdk-for-net/pull/34645#discussion_r1214804433).

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Generated/ContainerRegistryBlobRestClient.cs
annelo-msft: Is there a way we could intercept that?  Are there pipeline configurations we could make that might help us achieve the goal here?

sdk/core/Azure.Core/api/Azure.Core.net461.cs
KrzysztofCwalina: We could call it ClientRedirectsOptions

sdk/core/Azure.Core/api/Azure.Core.net461.cs
KrzysztofCwalina: See boolean property naming guidelines: https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members#names-of-properties

sdk/core/Azure.Core/src/Pipeline/RedirectPolicy.cs
```diff
@@ -15,12 +15,17 @@ namespace Azure.Core.Pipeline
     public sealed class RedirectPolicy : HttpPipelinePolicy
     {
         private readonly int _maxAutomaticRedirections;
-        private readonly bool _allowAutoRedirect = true;
+        internal bool AllowAutoRedirects { get; set; } = false;
```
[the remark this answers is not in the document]
christothes: initially I thought I'd need this for testability, but turns out it wasn't. Will revert.

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/ContainerRegistryBlobRestClient.cs
annelo-msft: I agree it makes sense to keep the complexity out of the redirect policy until more than one library needs the functionality.  I would still like to have a way to handle this without overwriting the generated methods for ACR, though, even if that means adding complexity to its pipeline.  Would the recommendation be for us to add a policy to the pipeline to throw on a 3xx response from an unexpected operation?

sdk/core/Azure.Core/api/Azure.Core.net461.cs
annelo-msft: That is helpful context, thank you!

I'm interested to understand, with this guidance, how types in STJ ended up with names like ["IgnoreNullValues"](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializeroptions.ignorenullvalues?view=net-7.0).  Are they just uncompliant?

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/ContainerRegistryBlobRestClient.cs
christothes: In my mind, the additional policy is equivalent to overriding generated code since this behavior could change if the swagger does.

I believe the easiest alternative is to enable redirects for the whole pipeline. Another alternative could be to construct 2 pipelines, one with redirects enabled and one without, and use them as appropriate for each operation.

christothes: We'd construct another pipeline and rest client with options set to enable redirects, like we do [here](https://github.com/Azure/azure-sdk-for-net/blob/6ad5dbbb1b5240e1aa7d9bb5dfd912371d69ddac/sdk/containerregistry/Azure.Containers.ContainerRegistry/src/ContainerRegistryClient.cs#L81-L91C13).

then use the 2nd rest client (`_restClientWithRedirects`) in the client operations that require redirect support.

annelo-msft: Oh yeah, sold!  Where do I sign?

christothes: done!

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Content/ContainerRegistryContentClient.cs
```diff
@@ -760,8 +764,8 @@ public virtual async Task<Response<DownloadRegistryBlobResult>> DownloadBlobCont
         private async Task<Response<DownloadRegistryBlobResult>> DownloadBlobContentInternalAsync(string digest, bool async, CancellationToken cancellationToken)
         {
             ResponseWithHeaders<Stream, ContainerRegistryBlobGetBlobHeaders> blobResult = async ?
-                await _blobRestClient.GetBlobAsync(_repositoryName, digest, cancellationToken).ConfigureAwait(false) :
-                _blobRestClient.GetBlob(_repositoryName, digest, cancellationToken);
+                await _blobRestClientWithRedirects.GetBlobAsync(_repositoryName, digest, cancellationToken).ConfigureAwait(false) :
```
annelo-msft, line 767: 🎉

annelo-msft: Maybe let's update the CHANGELOG to document?

JoshLove-msft:
> * Customers choosing to enable or disable redirects globally via `ClientOptions.ClientRedirects.IsClientRedirectEnabled`. This setting takes precedence over the following setting.

Why do we need this option? Shouldn't it always be turned off unless needed for a library? For something truly custom, customers can customize the pipeline via the second option being added here.

JoshLove-msft:
> * Client authors choosing to enable or disable redirects globally via `HttpPipelineTransportOptions.IsAutoRedirectEnabled`

Couldn't this one be configured using the RedirectPolicy.SetAllowAutoRedirect method?

sdk/core/Azure.Core/src/Pipeline/HttpPipelineBuilder.cs
```diff
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
```
JoshLove-msft, line 6: Is this needed?

sdk/communication/Azure.Communication.CallAutomation/src/CallAutomationClient.cs
```diff
@@ -95,9 +95,9 @@ private CallAutomationClient(string endpoint, TokenCredential tokenCredential, C
 
         private CallAutomationClient(Uri endpoint, CallAutomationClientOptions options, ConnectionString connectionString)
         : this(
-        endpoint: endpoint,
-        httpPipeline: options.CustomBuildHttpPipeline(connectionString),
-        options: options)
+            endpoint: endpoint,
```
JoshLove-msft, line 98: nit: undo these formatting changes.

christothes:
> > * Customers choosing to enable or disable redirects globally via `ClientOptions.ClientRedirects.IsClientRedirectEnabled`. This setting takes precedence over the following setting.
> 
> Why do we need this option? Shouldn't it always be turned off unless needed for a library? For something truly custom, customers can customize the pipeline via the second option being added here.

One of the explicit goals was to allow this to be configurable by customers easily. The HttpPipelineTransportOptions.IsAutoRedirectEnabled option is only accessible to client authors that are building the pipeline. Setting a custom transport is the only other way a customer could enable redirects, but that seems like overkill.

christothes:
> > * Client authors choosing to enable or disable redirects globally via `HttpPipelineTransportOptions.IsAutoRedirectEnabled`
> 
> Couldn't this one be configured using the RedirectPolicy.SetAllowAutoRedirect method?

Not really. `RedirectPolicy.SetAllowAutoRedirect` is only available if you have access to the HttpMessage and must be set per-request. So, for example, in an HLC, this would require customizing each generated method.

sdk/core/Azure.Core/src/Pipeline/HttpPipelineBuilder.cs
christothes: Nope! Added inadvertently.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -392,6 +393,11 @@ public abstract partial class ClientOptions
         [System.ComponentModel.EditorBrowsableAttribute(System.ComponentModel.EditorBrowsableState.Never)]
         public override string? ToString() { throw null; }
     }
+    public partial class ClientRedirectsOptions
+    {
+        public ClientRedirectsOptions() { }
```
annelo-msft, line 398: Super nit: `ClientRedirectOptions` ?

sdk/resourcemanager/Azure.ResourceManager/src/Resources/Custom/TenantResource.cs
```diff
@@ -16,8 +16,10 @@ namespace Azure.ResourceManager.Resources
     [CodeGenSuppress("GetAsync", typeof(CancellationToken))]
     [CodeGenSuppress("GetAvailableLocations", typeof(CancellationToken))]
     [CodeGenSuppress("GetAvailableLocationsAsync", typeof(CancellationToken))]
+#pragma warning disable CA1825 // Avoid zero-length array allocations
```
annelo-msft, line 19: Curious: didn't @heaths do something to address this before?  I'm surprised to see it's back, but don't remember what the action was.

sdk/core/Azure.Core/src/Pipeline/RedirectPolicy.cs
```diff
@@ -15,12 +15,17 @@ namespace Azure.Core.Pipeline
...
 
-        internal static RedirectPolicy Shared { get; } = new RedirectPolicy();
+        internal static RedirectPolicy Shared { get; } = new RedirectPolicy(false);
 
-        private RedirectPolicy()
+        /// <summary>
+        /// Creates a new instance of the <see cref="RedirectPolicy"/> class.
+        /// </summary>
+        /// <param name="allowAutoRedirects">Determinds whether redirects will be handled by this policy. Rather than passing false, consider using the static <see cref="Shared"/> instance instead which defaults to false.</param>
+        internal RedirectPolicy(bool allowAutoRedirects)
```
annelo-msft, line 26: Nit: it's internal, so it totally doesn't matter, but this was singular previously because of the corresponding property name in HttpClient: https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclienthandler.allowautoredirect?view=net-7.0

sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -1087,6 +1093,7 @@ public partial class HttpPipelineTransportOptions
     {
         public HttpPipelineTransportOptions() { }
         public System.Collections.Generic.IList<System.Security.Cryptography.X509Certificates.X509Certificate2> ClientCertificates { get { throw null; } }
+        public bool IsClientRedirectEnabled { get { throw null; } set { } }
```
annelo-msft, line 1096: I'm interested in [this comment] @JoshLove-msft made, and its implication that we might be able to reduce added API surface area.  Can you help me understand why we need this flag in both the outer container and separately on the transport options?

sdk/communication/Azure.Communication.CallAutomation/src/CustomBuildHttpPieline.cs
```diff
@@ -14,7 +14,9 @@ public static HttpPipeline CustomBuildHttpPipeline(this ClientOptions options, C
             string stringSign = new Uri(connectionString.GetRequired("endpoint")).Host;
 
             var authPolicy = new CustomHMACAuthenticationPolicy(new AzureKeyCredential(connectionString.GetRequired("accesskey")), stringSign);
-            return HttpPipelineBuilder.Build(options, authPolicy);
+            HttpPipelineOptions httpPipelineOptions = new(options) { PerRetryPolicies = { authPolicy } };
+            HttpPipelineTransportOptions httpPipelineTransportOptions = new() { IsClientRedirectEnabled = true };
```
annelo-msft, line 18: Could this use options.ClientRedirects.IsClientRedirectEnabled instead?

sdk/core/Azure.Core/api/Azure.Core.net461.cs
annelo-msft, line 398: I believe options types used on ClientOptions typically have an internal constructor, since customers don't need to create them.

sdk/core/Azure.Core/tests/HttpPipelineFunctionalTests.cs
```diff
@@ -628,7 +628,7 @@ public async Task SendMultipartformData()
         }
 
         [Test]
-        public async Task HandlesRedirects()
+        public async Task HandlesRedirects([Values(true, false)] bool allowRedirects)
```
annelo-msft, line 631: Do we have tests of precedence rules of options vs. setting values on the message?

sdk/core/Azure.Core/api/Azure.Core.net461.cs
christothes: I had the same inclination, but I think the plural is intentional.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
christothes: Containers is the pattern for ClientOptions , so I don't think we should add a top level property there. I suppose we could add a container in the transport options, but I don't know if that makes things better.

sdk/core/Azure.Core/CHANGELOG.md
```diff
@@ -12,6 +12,8 @@
 
 ### Other Changes
 
+- Client redirects are disabled by default and can be enabled by setting `ClientOptions.ClientRedirects.IsClientRedirectEnabled` to `true`.
```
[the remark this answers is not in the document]
christothes: I'd lean towards not mentioning it. I agree that "Client redirects are now disabled ..." is a good change.

sdk/core/Azure.Core/tests/HttpPipelineFunctionalTests.cs
christothes: added

JoshLove-msft, line 631: Please use TestCase to match other tests in this file.

JoshLove-msft:
> One of the explicit goals was to allow this to be configurable by customers easily.

Why is this one of the goals? I thought the goal was to turn off redirects unless it is explicitly required by the service. To me that seems like we would only need this to be configurable by library authors, not necessarily by end users.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
annelo-msft: Why is that?

sdk/core/Azure.Core/api/Azure.Core.net461.cs
annelo-msft: Why is it that we also need it on the transport options?  If we have the same option in two places, what happens if one is set to true and one is set to false?

sdk/communication/Azure.Communication.CallAutomation/src/CustomBuildHttpPieline.cs
annelo-msft: Ah, gotcha, yes that makes sense.
I think I was misunderstanding that HttpPipelineTransportOptions isn't also on ClientOptions.

Can you help me understand why we have HttpPipelineTransportOptions in the public API?  Sorry that I don't remember this well.  Is it only for library authors, and not end-user customers?

So then we have a precedence ordering of:
1. Per message: RedirectPolicy.SetAllowAutoRedirect(message, true);
2. Per client, set by end-user customer: ClientOptions.ClientRedirects.IsClientRedirectEnabled = true;
3. Per client, set by library author: HttpPipelineTransportOptions.IsClientRedirectEnabled = true;

Where a lower number will override a higher number on that list?

To [Josh's point](https://github.com/Azure/azure-sdk-for-net/pull/34645#issuecomment-1581396095), do we have a need for item 2 on the list?  It seems like we could always add that in response to customer feedback, but not providing it now improves security of our libraries.

sdk/resourcemanager/Azure.ResourceManager/src/Resources/Custom/TenantResource.cs
heaths: Yes, it shouldn't be necessary anymore. Make sure your feature branch is rebased on main if you're seeing it. See https://github.com/Azure/autorest.csharp/issues/3261 for context.

sdk/core/Azure.Core/tests/HttpPipelineFunctionalTests.cs
christothes: It's not the only one 😄

sdk/communication/Azure.Communication.CallAutomation/src/CustomBuildHttpPieline.cs
christothes: Removing the clientOptions API.

christothes:
> > One of the explicit goals was to allow this to be configurable by customers easily.
> 
> Why is this one of the goals? I thought the goal was to turn off redirects unless it is explicitly required by the service. To me that seems like we would only need this to be configurable by library authors, not necessarily by end users.

There are scenarios where customers would want to enable this, but I'm ok with removing this for now and adding it back if we receive feedback on it.

annelo-msft (approved):
Thanks for all your hard work on this!

sdk/communication/Azure.Communication.CallingServer/src/Azure.Communication.CallingServer.csproj
```diff
@@ -27,7 +27,8 @@
     <Compile Include="$(AzureCoreSharedSources)CancellationHelper.cs" LinkBase="Shared" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Azure.Core" />
+    <!-- <PackageReference Include="Azure.Core" /> -->
```
chrwhit, line 30: Same as comment for sdk/communication/Azure.Communication.CallAutomation/src/Azure.Communication.CallAutomation.csproj

sdk/communication/Azure.Communication.PhoneNumbers/src/Azure.Communication.PhoneNumbers.csproj
```diff
@@ -18,7 +18,8 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Azure.Core" />
+    <!-- <PackageReference Include="Azure.Core" /> -->
```
chrwhit, line 21: same as sdk/communication/Azure.Communication.CallAutomation/src/Azure.Communication.CallAutomation.csproj

chrwhit (approved):
Just some minor feedback

sdk/communication/Azure.Communication.CallAutomation/src/Azure.Communication.CallAutomation.csproj
```diff
@@ -28,7 +28,8 @@
     <Compile Include="$(AzureCoreSharedSources)CancellationHelper.cs" LinkBase="Shared" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Azure.Core" />
+    <!-- <PackageReference Include="Azure.Core" /> -->
```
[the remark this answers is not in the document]
christothes: It's intentional - this should get reverted once Azure.Core ships.

PR #34648, opened 2023-03-01
Release preparation for Azure.Core 1.29.0

annelo-msft:
Changelog and project version for March Azure.Core release.


sdk/core/Azure.Core/CHANGELOG.md
```diff
@@ -1,19 +1,15 @@
 # Release History
 
-## 1.29.0-beta.1 (Unreleased)
+## 1.29.0 (2023-03-02)
 
 ### Features Added
 
 - `ActivitySource` activities that are used when using the [experimental OpenTelemetry support](https://devblogs.microsoft.com/azure-sdk/introducing-experimental-opentelemetry-support-in-the-azure-sdk-for-net/) will include the `az.schema_url` tag indicating the OpenTelemetry schema version. They will also include the attribute names specified [here](https://github.com/Azure/azure-sdk/blob/main/docs/tracing/distributed-tracing-conventions.yml).
```
JoshLove-msft, line 7: I documented this here because it affects all libraries via shared source, but if this is the only change, then there is nothing to actually release as this does not get shipped as part of Azure.Core. I think @AlexanderSher had a bunch of perf improvements though.

annelo-msft: Sounds good.  @AlexanderSher, would you add entries to the changelog for [the changes you've made/approved this release cycle](https://github.com/Azure/azure-sdk-for-net/commits/main/sdk/core/Azure.Core)?

AlexanderSher: All the changes that I've made are internal performance tweaks. Do we want to mention those in the changelog?

annelo-msft: There is precedent for doing this in our libraries in the following CHANGELOGs:
- https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/keyvault/Azure.Security.KeyVault.Administration/CHANGELOG.md?plain=1#L117
- https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/servicebus/Microsoft.Azure.ServiceBus/CHANGELOG.md?plain=1#L76
- https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/storage/Azure.Storage.Blobs/CHANGELOG.md?plain=1#L274

AlexanderSher: There is one change however that I've approved: https://github.com/Azure/azure-sdk-for-net/commit/0056e28d5ea65be7eb0be5a76913b7734beefa18

This is a public change, we should add it to the changelog

annelo-msft: It looks like you also approved a PR that added some values to an extensible enum, which may be worth mentioning?

christothes (approved):

PR #34660, opened 2023-03-02
[AzureMonitorOpenTelemetryDistro] APPLICATIONINSIGHTS_CONNECTION_STRING to take precedence over IConfiguration

rajkumar-rangaraj:


sdk/monitor/Azure.Monitor.OpenTelemetry/src/DefaultAzureMonitorOpenTelemetryOptions.cs
```diff
@@ -26,6 +27,13 @@ public void Configure(AzureMonitorOpenTelemetryOptions options)
             {
                 _configuration.GetSection(AzureMonitorOpenTelemetrySectionFromConfig).Bind(options);
             }
+
+            string? connectionString = _configuration?[ConnectionStringEnvironmentVariable];
```
cijothomas, line 31: Please add test to confirm this works the way intended.

sdk/monitor/Azure.Monitor.OpenTelemetry/src/DefaultAzureMonitorOpenTelemetryOptions.cs
```diff
@@ -9,6 +9,7 @@ namespace Azure.Monitor.OpenTelemetry
     internal class DefaultAzureMonitorOpenTelemetryOptions : IConfigureOptions<AzureMonitorOpenTelemetryOptions>
     {
         private const string AzureMonitorOpenTelemetrySectionFromConfig = "AzureMonitorOpenTelemetry";
+        private const string ConnectionStringEnvironmentVariable = "APPLICATIONINSIGHTS_CONNECTION_STRING";
```
cijothomas, line 12: If this is to take highest preference, then do we need to get it from IConfiguration? Can we just read the env-var and overwrite anything provided via IConfiguration? Otherwise, how do we ensure it takes precedence over some other source, like appsettings.json file

sdk/monitor/Azure.Monitor.OpenTelemetry/src/DefaultAzureMonitorOpenTelemetryOptions.cs
rajkumar-rangaraj: We don't have a test framework yet. Planning to merge the test for both exporters and distro. I validated it with the manual test.

sdk/monitor/Azure.Monitor.OpenTelemetry/src/DefaultAzureMonitorOpenTelemetryOptions.cs
rajkumar-rangaraj: > Can we just read the env-var and overwrite anything provided via IConfiguration? 

We are exactly doing the same with the proposed change.

cijothomas: _configuration?[ConnectionStringEnvironmentVariable]

^ This can read from anything bound to IConfiguration, not just env variable..

rajkumar-rangaraj: IConfiguration reads environment variables by default. There are 2 ways we could read APPLICATIONINSIGHTS_CONNECTION_STRING
1. Using IConfiguraion
2. Environment.GetEnvironmentVariable()

I used IConfiguration as we are in that context.

rajkumar-rangaraj: Got it, will change to read from `Environment.GetEnvironmentVariable`

cijothomas: > IConfiguration reads environment variables by default

This is not a guarantee. Its done by the default webhost builder, but we can't assume so.

sdk/monitor/Azure.Monitor.OpenTelemetry/src/DefaultAzureMonitorOpenTelemetryOptions.cs
cijothomas: ok. this is a p1 thing to have sufficient coverage. reading connectionstring from unintended place/priority or not reading it at all are big support costs.

sdk/monitor/Azure.Monitor.OpenTelemetry/src/DefaultAzureMonitorOpenTelemetryOptions.cs
cijothomas: image
My new assistant is doing good job today ;)

rajkumar-rangaraj: It's a cool assistant, comes handy nowadays :blush:

vishweshbankwar (approved):
LGTM

PR #34663, opened 2023-03-02
exclude common from code coverage calculation

ostoliarova-ee:


petrsvihlik:
we don't want to go this way. it's semantically incorrect - we don't want to turn off the coverage calculation for classes like `CommunicationIdentifier`. we want to have them covered, we just don't want to include the calculation in the overall report for Identity. sorry if this was not clear from the beginning.

please explore how we can configure the coverage tool properly as suggested in https://github.com/Azure/azure-sdk-for-net/pull/34575#pullrequestreview-1323958514

PR #34665, opened 2023-03-02
Cross-linking to additional guidance

udidahan:


jsquire (commented):
Hi @udidahan.  Thanks very much for your contribution!  I left a couple of thoughts below and asked for some additional perspectives on this change.

udidahan:
Folks, if I might ask a slightly more philosophical question, does the SDK team believe it would be valuable to cross-link between SDK Docs and the other learn.microsoft.com content?

jsquire:
> Folks, if I might ask a slightly more philosophical question, does the SDK team believe it would be valuable to cross-link between SDK Docs and the other learn.microsoft.com content?

Hi Udi.   Apologies for the delay; generally speaking, it's situational and depends on the doc.  Conceptual docs with general information about an Azure service are something we try to cross-link, as they fill in valuable context at a higher level than SDK documentation.  

Where docs with code samples and quick-starts are concerned, we're usually more cautious, as those have the potential to not be updated regularly and use stale versions, not follow recommended patterns, and not be at the same level of code quality that we'd like.  We do our best to work with the content authors that own them to address issues as we see them, but the cross-team ownership can add some time to the process and makes awareness of issues more difficult to keep on top of.

jsquire:
Hi @udidahan.  Thanks for your patience as we talked this over and coordinated with the folks over at MS Docs.   Given that the linked document is full "getting started" sample which overlaps with the local content, we'd prefer not to link this from the README.  While the additional examples for manually creating resources would complement our ARM deployment, the additional code snippets from the article introduce a potential drift with the content and samples in the repository and which we cannot directly control the evolution of.   Further rationale can be found in my earlier response.

I'm going to close this out but am happy to continue the discussion.  Thank you, again, for your contribution and your efforts to improve the Azure SDK developer experience.

udidahan:
OK, thanks. I'll try to keep all that in mind when I'm considering other improvements I can make.

PR #34679, opened 2023-03-02
Update to latest Key Vault stable/7.4

heaths:


christothes (approved):
No regeneration required?

heaths:
> No regeneration required?

Nope. I already had a transform handle this rename from when their swagger and service were different and the service was deploying the change.

PR #34683, opened 2023-03-03
[Text Analytics] Log warn-text response header

### LINKED ISSUE #33421: [Text Analytics] Add support for warnings reported in response headers
The service returns a `warn-text` header to communicate to users when an API version is retired, e.g. "Microsoft Cognitive Service Language APIs version 2022-04-01-preview and will be removed." Design a way to expose this information in the client libraries.

joseharriaga:
Resolves: https://github.com/Azure/azure-sdk-for-net/issues/33421

The service returns a `warn-text` header to communicate to users when an API version is retired, e.g. "Microsoft Cognitive Service Language APIs version 2022-04-01-preview and will be removed." This change adds the header to the list of logged headers.

### DIFF
--- a/sdk/textanalytics/Azure.AI.TextAnalytics/CHANGELOG.md
+++ b/sdk/textanalytics/Azure.AI.TextAnalytics/CHANGELOG.md
@@ -4,8 +4,12 @@
 
 ### Features Added
 
+- Added support for logging the "warn-text" response header.
+
 ### Breaking Changes
+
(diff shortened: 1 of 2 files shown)


heaths (changes requested):
The change looks fine, but I have a concern about the service and will also bring it up with them, but could you pass this along: the API Stewardship Board already has a standard for this: https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#deprecating-behavior-notification

heaths (approved):

PR #34691, opened 2023-03-03
Fix serialization issue for Maintenance

### LINKED ISSUE #34604: Receiving error while using Azure.ResourceManager.Maintenance package - invalid maintenance window. error: startdatetime is not set for maintenance window
Hey,

I'm using the Azure.ResourceManager.Maintenance Nuget in order to create Maintenance configuration resource in Azure and assign it to VMSS in our cloud resources.

I've been working with this documentation:
https://learn.microsoft.com/en-us/dotnet/api/azure.resourcemanager.maintenance.maintenanceconfigurationdata?view=azure-dotnet

For some reason while trying to create the resource of Maintenance Configuration I receive an error which states that I'm missing the StartDateTime property even tough this parameter is not optional in latest version of the Nuget only in beta version.
In the latest Nuget version we need to use the StartOn property which we did but for some reason we are still receiving this error:
 invalid maintenance window. error: startdatetime is not set for maintenance window

This is our constructor definition:

        protected override async Task DeployInternalAsync(MaintenanceConfigurationConfiguration configuration, IReadOnlyCollection<IAzureResource> dependencies, CancellationToken cancellationToken)
        {
                var maintenanceConfigurationResource = await _maintenanceConfigurationCollection.CreateOrUpdateAsync(
                WaitUntil.Completed,
                Name.ToLowerInvariant(),
                new MaintenanceConfigurationData(AzureLocation.EastUS)
                {
                    MaintenanceScope = MaintenanceScope.OSImage,
                    TimeZone = "UTC",
                    RecurEvery = "1Day",
                    Duration = 6.Hours(),
                    StartOn = DateTimeOffset.UtcNow,
                });

There is no StartDateTime property for this constructor.

ArthurMa1978:
Fixes https://github.com/Azure/azure-sdk-for-net/issues/34604


sdk/maintenance/Azure.ResourceManager.Maintenance/src/Custom/RestOperations/MaintenanceConfigurationsRestOperations.cs
```diff
@@ -0,0 +1,95 @@
...
+using Azure;
+using Azure.Core;
+using Azure.Core.Pipeline;
+using Azure.ResourceManager.Maintenance.Models;
+
+namespace Azure.ResourceManager.Maintenance
+{
+    internal partial class MaintenanceConfigurationsRestOperations
+    {
+        internal HttpMessage CreateDeleteRequest(string subscriptionId, string resourceGroupName, string resourceName)
```
Yao725, line 19: Any reason that we need to customize this method? I don't see the difference between this and the generated one.

sdk/maintenance/Azure.ResourceManager.Maintenance/src/Custom/RestOperations/MaintenanceConfigurationsRestOperations.cs
```diff
@@ -0,0 +1,95 @@
...
+        /// <param name="cancellationToken"> The cancellation token to use. </param>
+        /// <exception cref="ArgumentNullException"> <paramref name="subscriptionId"/>, <paramref name="resourceGroupName"/> or <paramref name="resourceName"/> is null. </exception>
+        /// <exception cref="ArgumentException"> <paramref name="subscriptionId"/>, <paramref name="resourceGroupName"/> or <paramref name="resourceName"/> is an empty string, and was expected to be non-empty. </exception>
+        public async Task<Response<MaintenanceConfigurationData>> DeleteAsync(string subscriptionId, string resourceGroupName, string resourceName, CancellationToken cancellationToken = default)
+        {
+            Argument.AssertNotNullOrEmpty(subscriptionId, nameof(subscriptionId));
+            Argument.AssertNotNullOrEmpty(resourceGroupName, nameof(resourceGroupName));
+            Argument.AssertNotNullOrEmpty(resourceName, nameof(resourceName));
+
+            MaintenanceConfigurationData data = Get(subscriptionId, resourceGroupName, resourceName, cancellationToken).Value;
```
Yao725, line 52: I think we shouldn't add an extra API call here, can we directly return null instead of the data? Returning null is also consistent with the service actual behavior.

sdk/maintenance/Azure.ResourceManager.Maintenance/src/Custom/RestOperations/MaintenanceConfigurationsRestOperations.cs
ArthurMa1978: I added comments for why customize this operation:

// In the swagger this delete operation is defined to returen a `MaintenanceConfigurationData` model, but actually the service doesn't return, that cause exception when try to descerialize the response payload .
// As this lib has already GAed, so here using custom code to retrieve the data before running the delete operation that can keep backward compatibility.

sdk/maintenance/Azure.ResourceManager.Maintenance/src/Custom/RestOperations/MaintenanceConfigurationsRestOperations.cs
ArthurMa1978: I added comments for why customize this operation:

// In the swagger this delete operation is defined to returen a MaintenanceConfigurationData model, but actually the service doesn't return, that cause exception when try to descerialize the response payload .
// As this lib has already GAed, so here using custom code to retrieve the data before running the delete operation that can keep backward compatibility.

sdk/maintenance/Azure.ResourceManager.Maintenance/src/Custom/MaintenanceConfigurationData.cs
```diff
@@ -1,8 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-// <auto-generated/>
```
ArcturusZhang, line 4: I did not see anything changed in this file comparing with its generated version.
Do we have any?
If not, please remove this file and keep this file generated so that we could get new properties in future generation.

sdk/maintenance/Azure.ResourceManager.Maintenance/src/Custom/Models/MaintenanceConfigurationData.Serialization.cs
```diff
@@ -12,6 +10,11 @@
 using Azure.ResourceManager.Maintenance.Models;
 using Azure.ResourceManager.Models;
 
+// Custom code to support custom datetime format for the 2 properties startDateTime and expirationDateTime
+// Here is an issue https://github.com/Azure/autorest.csharp/issues/3184 to track if codegen can support custom datetime format in the furture.
+
+[assembly: CodeGenSuppressType("MaintenanceConfigurationData")]
```
ArcturusZhang, line 16: We should be able to only replace the deserilaization method instead of suppress the entrie class
In addition, the CodeGenSuppressType attribute will also remove the corresponding model factory for this, therefore we should minimize our changes.

sdk/maintenance/Azure.ResourceManager.Maintenance/src/Custom/RestOperations/MaintenanceConfigurationsRestOperations.cs
ArcturusZhang: This is making an extra REST API call?
We should not do this.
How about we just return `null` for the data, or construct one using our internal constructor by setting minimum properties of it so that we do not have to call the GET API

ArcturusZhang: I saw an extra REST API call in our recording for the delete. We should avoid it.

sdk/maintenance/Azure.ResourceManager.Maintenance/tests/Scenario/MaintenanceConfigurationTests.cs
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        [TearDown]
+        public async Task TearDown()
+        {
+            await _resourceGroup.DeleteAsync(WaitUntil.Completed);
+        }
+
+        [RecordedTest]
+        public async Task DeleteTest()
```
ArcturusZhang, line 44: We do not need to add the word `Test` on the method name
```suggestion
        public async Task Delete()
```
If you change this method name, you do not need to record the test cases again, just rename the recording file correspondingly.

sdk/maintenance/Azure.ResourceManager.Maintenance/src/Custom/MaintenanceConfigurationData.cs
ArthurMa1978: The custom code is in the `MaintenanceConfigurationData.Serialization.cs` file, which need to use `CodeGenSuppressType` to suppress the whole `MaintenanceConfigurationData` model, so have to copy this file to the `Custom` folder too.

sdk/maintenance/Azure.ResourceManager.Maintenance/src/Custom/RestOperations/MaintenanceConfigurationsRestOperations.cs
Yao725: I agree that we should customize the method `public async Task<Response<MaintenanceConfigurationData>> DeleteAsync(...`, but this method, aka `internal HttpMessage CreateDeleteRequest(...`, is same as the generated code. So I don't think we should remove it from the generated directory to the custom directory.

sdk/maintenance/Azure.ResourceManager.Maintenance/src/Custom/RestOperations/MaintenanceConfigurationsRestOperations.cs
Yao725: I still think we should avoid the extra call. Think about this scenario, what if the customer call Delete to delete a non-existed resource. We will get the exception first from the Get method, this might confuse the customer.

archerzz (approved):

sdk/maintenance/Azure.ResourceManager.Maintenance/src/Custom/RestOperations/MaintenanceConfigurationsRestOperations.cs
ArthurMa1978: Fixed, use `directive` to fix the swagger spec and customize the `Delete` operations in the  `MaintenanceConfigurationResource` .

sdk/maintenance/Azure.ResourceManager.Maintenance/src/Custom/RestOperations/MaintenanceConfigurationsRestOperations.cs
ArthurMa1978: Fixed, use directive to fix the swagger spec and customize the `Delete` operations in the `MaintenanceConfigurationResource`.

sdk/maintenance/Azure.ResourceManager.Maintenance/src/Custom/MaintenanceConfigurationData.cs
ArthurMa1978: Fixed, use `CodeGenSuppress` directly overwrite the serialization method.

sdk/maintenance/Azure.ResourceManager.Maintenance/src/Custom/Models/MaintenanceConfigurationData.Serialization.cs
ArthurMa1978: Fixed, use `CodeGenSuppress` directly overwrite the serialization method.

Issue #19518, opened 2021-03-15
Support a MinBatchSize property in web job event hub extensions


remoba:
**Library or service name.**
Microsoft.Azure.WebJobs.Extensions.EventHubs
 
**Is your feature request related to a problem? Please describe.**
I have a scenario where I would like to aggregate events over time, and then handle them as a batch. The size of this batch must be pretty large for my downstream service's optimizations, so for example let's say 10k events.

Using the Event Hub SDK directly (Azure.Messaging.EventHubs) I could write a processor that aggregates incoming data into buckets (bucket-per-partition), and then when a bucket hits a certain event count threshold (or when enough time has passed since it was updated), I would "flush" the bucket, and then update the checkpoint for that particular partition. This way I am never at risk of data loss.
This same approach cannot be taken when using the current webjob SDK, as it automatically updates the checkpoint after every X batches are "processed", so if my process crashes before it can flush a bucket, all the data in that bucket is lost.

The current SDK has the property `MaxBatchSize` which puts an upper limit on the amount of messages in a batch, but it isn't related to the actual amount of unprocessed messages in a given partition, so even if I have thousands of events waiting to be processed, and my `MaxBatchSize` is set to 10k, I can still receive batches of 5-7 messages per batch.

I am proposing a `MinBatchSize` property that will tell the web job SDK to aggregate data in in-memory buckets. If set, the internal processor will fill the relevant bucket instead of triggering, and will not update the checkpoint. When a bucket hits the `MinBatchSize`, or when the configured time span since the last event has expired, the processor will trigger the user-code and afterwards will update the checkpoint for the given partition.

As far as I understand, the "invoke after enough time has passed since the last event" behavior was removed in a recent PR 
(#19140), but it might be needed here

dmeytin:
+1 for this feature request

DorBraunstein:
+1

jsquire:
Hi @remoba.  Thank you for your feedback.  Tagging and routing to the team member best able to assist.

paulbatum:
Couple of quick thoughts on this:
- Setting the timeout should be mandatory - we need the user to understand that this setting can cause processing delays. Forcing the user to specify a timeout value should address this.
- We'll need to enforce per-SKU limits on that timeout configuration setting. For example, the consumption plan does not allow executions of duration greater than ten minutes - it therefore does not make sense to allow a 1hr timeout in the consumption SKU.
[...]

PR #34698, opened 2023-03-03
[Event Hubs Web Jobs Extensions] Adding MinBatchSize and MaxWaitTime Properties

m-redding:
These changes add a minimum batch size property and maximum wait time property to the Event Hubs extensions. 

These changes were validated on the following set of local test runs:
- Multiple console apps sending small numbers of events every 10-30 seconds to the same partition, validating that all batches are between min and max size, unless maxWaitTime seconds have passed since last invocation. Tested this with a wide range of values for MinBatchSize, MaxBatchSize, MaxWaitTime, sending intervals, numbers of events being sent at a time, and number of console apps sending.
- Tested large numbers of events being sent with long intervals in between, validating that all batches are between min and max size, unless maxWaitTime seconds have passed since last invocation
- Tested the above cases with console apps sending to different partitions, validating partition separation
- Validated sending events right around the end of the maxWaitTime interval, in order to attempt to validate concurrency guards 

Addresses #19518


sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/EventHubOptions.cs
```diff
@@ -132,6 +134,48 @@ public int MaxEventBatchSize
...
+        /// </summary>
+        public int MinEventBatchSize
+        {
+            get => _minEventBatchSize;
+
+            set
+            {
+                if (value < 0)
+                {
+                    throw new ArgumentException("Batch size must be larger than or equal to 0.");
```
m-redding, line 152: Do we want to validate min < max? Seems impossible unless we write a constructor

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubListener.cs
```diff
@@ -104,6 +111,98 @@ public IScaleMonitor GetMonitor()
...
+            // This is internal for mocking purposes only.
+            internal ConcurrentDictionary<string, List<EventData>> StoredEvents { get; set; }
+
+            public bool HasStoredEvents
+            {
+                get
+                {
+                    try
+                    {
+                        _storedEventsGuard.Wait();
```
jsquire, line 146: I don't believe that we need to take the semaphore here.   The concurrent dictionary is thread-safe.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubListener.cs
```diff
@@ -104,6 +111,98 @@ public IScaleMonitor GetMonitor()
...
         }
 
+        internal class EventProcessorStoredEventsManager : IDisposable
+        {
+            private SemaphoreSlim _storedEventsGuard;
+            private int _maxBatchSize;
+            private int _minBatchSize;
+
+            // This is internal for mocking purposes only.
+            internal ConcurrentDictionary<string, List<EventData>> StoredEvents { get; set; }
```
jsquire, line 138: I'm wondering if we should consider a `ConcurrentQueue<EventData>` here, which would allow us to avoid the semaphore entirely.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubListener.cs
```diff
@@ -104,6 +111,98 @@ public IScaleMonitor GetMonitor()
...
+                Argument.AssertNotNull(partitionContext, nameof(partitionContext));
+                try
+                {
+                    _storedEventsGuard.Wait(cancellationToken);
+                    events ??= new List<EventData>();
+
+                    var partitionId = partitionContext.PartitionId;
+
+                    var numEventsToProcess = events.Count;
+                    var storedEventsForPartition = StoredEvents.GetOrAdd(partitionId, new List<EventData>());
```
jsquire, line 188: If we do move to a `ConcurrentQueue<EventData>` in the stored events set, I think we can unify this codepath to just loop and pop elements until the queue is empty or until we read the max size.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubListener.cs
```diff
@@ -104,6 +111,98 @@ public IScaleMonitor GetMonitor()
...
+                _maxBatchSize = maxBatchSize;
+                _minBatchSize = minBatchSize;
+            }
+
+            public void PartitionClosing()
+            {
+                try
+                {
+                    _storedEventsGuard.Wait();
+                    StoredEvents.Clear();
```
jsquire, line 169: This would just be for a single partition.   Should we be clearing the entire set?

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubListener.cs
```diff
@@ -114,20 +213,32 @@ internal class EventProcessor : IEventProcessor, IDisposable
...
+                _minimumBatchesEnabled = options.MinEventBatchSize > 1; // 1 is the default
+                StoredEventsManager = new EventProcessorStoredEventsManager(maxBatchSize: options.MaxEventBatchSize, minBatchSize: options.MinEventBatchSize);
+                _maxWaitTime = TimeSpan.FromSeconds(options.MaxWaitTime);
             }
 
             public Task CloseAsync(EventProcessorHostPartition context, ProcessingStoppedReason reason)
             {
                 // signal cancellation for any in progress executions
                 _cts.Cancel();
+                StoredEventsManager.PartitionClosing();
```
jsquire, line 258: Shouldn't this know which partition is being closed?

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubListener.cs
```diff
@@ -153,22 +264,23 @@ public async Task ProcessEventsAsync(EventProcessorHostPartition context, IEnume
                 using (CancellationTokenSource linkedCts =
                         CancellationTokenSource.CreateLinkedTokenSource(_cts.Token, processingCancellationToken))
                 {
-                    var events = messages.ToArray();
+                    _mostRecentPartitionContext = context;
+                    var events = messages == null ? Array.Empty<EventData>() : messages.ToArray();
```
jsquire, line 285: An enumerable should never be null.   What happened that we need this?

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubListener.cs
```diff
@@ -104,6 +111,98 @@ public IScaleMonitor GetMonitor()
...
+                    _storedEventsGuard.Wait();
+                    StoredEvents.Clear();
+                }
+                finally
+                {
+                    _storedEventsGuard.Release();
+                }
+            }
+
+            public EventData[] ProcessWithStoredEvents(EventProcessorHostPartition partitionContext, List<EventData> events = null, bool timerTrigger = false, CancellationToken cancellationToken = default)
```
jsquire, line 177: We're not so much processing as getting.   Maybe we should consider a rename?

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubListener.cs
```diff
@@ -214,8 +364,72 @@ public async Task ProcessEventsAsync(EventProcessorHostPartition context, IEnume
...
+                TriggeredFunctionData input = new()
+                {
+                    TriggerValue = triggerInput,
+                    TriggerDetails = triggerInput.GetTriggerDetails(context)
+                };
+
+                await _executor.TryExecuteAsync(input, cancellationToken).ConfigureAwait(false);
+
+                // The timer should reset upon invocation of the function
+                _storedEventsBackgroundTask = null;
```
jsquire, line 402: Seems like we should signal a cancellation token and then await the task.  How does this stop the task from doing stuff?

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubListener.cs
```diff
@@ -190,14 +302,52 @@ public async Task ProcessEventsAsync(EventProcessorHostPartition context, IEnume
...
+
+                                if (_storedEventsBackgroundTask != null)
+                                {
+                                    // If there is a background timer task, cancel it and dispose of the cancellation token.
+                                    _storedEventsBackgroundTaskCts.Cancel();
+                                    _storedEventsBackgroundTaskCts.Dispose();
+                                    _storedEventsBackgroundTaskCts = null;
+                                }
+
+                                if (StoredEventsManager.HasStoredEvents)
```
jsquire, line 341: I think we want to treat each partition independently.   Trying to keep them all on the same flush schedule while reads are potentially inconsistent doesn't seem the right thing to do.   Am I overlooking something?

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubListener.cs
```diff
@@ -104,6 +111,98 @@ public IScaleMonitor GetMonitor()
             return _scaleMonitor.Value;
         }
 
+        internal class EventProcessorStoredEventsManager : IDisposable
```
jsquire, line 131: I wonder if this is fleshed out enough that it should have its own file.

jsquire, line 131: I think it would be easier to convert this to something specific to one partition and then keep a dictionary of them in the Listener.    That would let you have each manager thingy own a `System.Threading.Timer` which would call `TriggerExecute` as it's callback.   Managing this timer would let you eliminate the explicit monitoring of background tasks for each partition.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubListener.cs
```diff
@@ -115,11 +122,103 @@ public IScaleMonitor GetMonitor()
...
         }
+        
+        internal class EventProcessorStoredEventsManager : IDisposable
+        {
+            private SemaphoreSlim _storedEventsGuard;
+            private int _maxBatchSize;
+            private int _minBatchSize;
+
+            // This is internal for mocking purposes only.
+            internal ConcurrentDictionary<string, List<EventData>> StoredEvents { get; set; }
```
jsquire, line 138: Given that we only have one set of events and no longer have to worry about partitions... do we need the dictionary at all or can we JUST use a `ConcurrentQueue<EventData>`?

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubListener.cs
jsquire: Ignore me... wrong scope.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubListener.cs
jsquire: Ignore me... wrong scope.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubListener.cs
jsquire: Ignore me... wrong scope.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubListener.cs
m-redding: I was getting null exceptions from this line at one point, but was never able to track them down via the debugger (or figure out where they were coming from logically), I'll try testing it without it and see if I'm still getting that exception.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/EventHubOptions.cs
JoshLove-msft: The validation can be added [here](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/EventHubWebJobsBuilderExtensions.cs#L89) as a PostConfigure step.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubListener.PartitionProcessor.cs
```diff
@@ -0,0 +1,367 @@
...
+							eventToCheckpoint = events[i];
+						}
+					}
+					else
+					{
+						// Batch dispatch
+
+						if (_minimumBatchesEnabled)
+						{
+							var triggerEvents = StoredEventsManager.ProcessWithStoredEvents(context, messages.ToList(), false, linkedCts.Token);
```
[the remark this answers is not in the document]
m-redding: The reason why I named it `ProcessWithStoredEvents` was to mirror `EventHubListener.PartitionProcessor.ProcessEvents`, since I was trying to denote, "do the same thing but add in stored events"

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/PartitionProcessorEventsManager.cs
```diff
@@ -0,0 +1,101 @@
...
+                finally
+                {
+                    _storedEventsGuard.Release();
+                }
+            }
+        }
+
+        public PartitionProcessorEventsManager(int maxBatchSize, int minBatchSize)
+        {
+            StoredEvents = new ConcurrentQueue<EventData>();
```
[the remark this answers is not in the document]
m-redding: Yeah I think we definitely need the semaphore for the ordering reason (which is why I didn't want to take it out) so I'll switch to a regular queue

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/EventHubOptions.cs
```diff
@@ -132,6 +134,52 @@ public int MaxEventBatchSize
...
+        {
+            get => _maxWaitTime;
+
+            set
+            {
+                if (value < TimeSpan.Zero)
+                {
+                    throw new ArgumentException("Max Wait Time must be larger than or equal to 0.");
+                }
+                if (value > TimeSpan.FromMinutes(10))
```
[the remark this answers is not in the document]
jsquire: Far better than opening the door to the support requests for "I know that I have events but my Function never runs."

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubListener.PartitionProcessor.cs
```diff
@@ -0,0 +1,367 @@
...
+
+								if (_storedEventsBackgroundTask != null)
+								{
+									// If there is a background timer task, cancel it and dispose of the cancellation token.
+									_storedEventsBackgroundTaskCts.Cancel();
+									_storedEventsBackgroundTaskCts.Dispose();
+									_storedEventsBackgroundTaskCts = null;
+								}
+
+								if (StoredEventsManager.HasStoredEvents)
```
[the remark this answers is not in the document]
m-redding: I could be wrong, but in theory that shouldn't happen since I'm closing out all the timers in the lines above? The intention is that with each function invocation the background timer gets cleared and restarted.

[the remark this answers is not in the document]
m-redding: *restarted if there are leftover events in the cache

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/PartitionProcessorEventsManager.cs
```diff
@@ -0,0 +1,83 @@
...
+            _maxBatchSize = maxBatchSize;
+            _minBatchSize = minBatchSize;
+        }
+
+        public void PartitionClosing()
+        {
+            CachedEvents.Clear();
+        }
+
+        public EventData[] GetBatchofEventsWithCached(EventData[] events = null, bool timerTrigger = false)
```
jsquire, line 48:
```suggestion
        public EventData[] GetBatchofEventsWithCached(EventData[] events = null, bool allowPartialBatch= false)
```

The listener is managing the timer; there's no need for this class to be aware that one is used.  What we're really saying here is whether or not the caller would like a partial batch to be returned.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubListener.PartitionProcessor.cs
```diff
@@ -0,0 +1,344 @@
...
+			public Task ProcessErrorAsync(EventProcessorHostPartition context, Exception error)
+			{
+				string errorDetails = $"Processing error (Partition Id: '{context.PartitionId}', Owner: '{context.Owner}', EventHubPath: '{context.EventHubPath}').";
+
+				Utility.LogException(error, errorDetails, _logger);
+
+				return Task.CompletedTask;
+			}
+
+			public async Task ProcessEventsAsync(EventProcessorHostPartition context, IEnumerable<EventData> messages, CancellationToken processingCancellationToken)
```
jsquire, line 79: I believe we should consider extracting the `lock` out of the events manager and using it here instead.   This would allow the lock to be taken at the beginning of the `ProcessEventsAsync` call and avoid having to take/release it multiple times.   Likewise, the `MonitorCachedEvents` method could take the lock before L209, which would prevent the handler from running as the background task emits the batch.

I think this provides a cleaner and more intuitive set of guarantees, as the partition processor has a better understanding of where concurrency can occur and how coarse-grained the locking should be.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/EventHubOptions.cs
m-redding: Should be addressed now :)

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/EventHubOptions.cs
```diff
@@ -132,6 +134,52 @@ public int MaxEventBatchSize
...
+        /// </summary>
+        public int MinEventBatchSize
+        {
+            get => _minEventBatchSize;
+
+            set
+            {
+                if (value < 1)
+                {
+                    throw new ArgumentException("Batch size must be larger than or equal to 0.");
```
jsquire, line 152:
```suggestion
                    throw new ArgumentException("Batch size must be larger than or equal to 1.");
```

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/EventHubOptions.cs
```diff
@@ -132,6 +134,52 @@ public int MaxEventBatchSize
...
+                _minEventBatchSize = value;
+            }
+        }
+
+        private TimeSpan _maxWaitTime;
+
+        /// <summary>
+        /// Gets or sets the maximum time in seconds. This only applies when <see cref="MinEventBatchSize"/> is set. It denotes the
+        /// maximum number of seconds the processor will wait after receiving a batch of events less than <see cref="MinEventBatchSize"/>
+        /// before invoking the function. Default is 60 seconds when <see cref="MinEventBatchSize"/> has been set, otherwise, this is not used.
```
jsquire, line 163:
```suggestion
        /// Gets or sets the maximum time that the trigger should wait to fill a batch before invoking the function.  This is only considered when <see cref="MinEventBatchSize"/> is set to larger than 1 and is otherwise unused.  If less than <see cref=MinEventBatchSize" /> events were available before the wait time elapses, the function will be invoked with a partial batch.  Default is 60 seconds.  The maximum wait time is 10 minutes.
```

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubListener.PartitionProcessor.cs
```diff
@@ -0,0 +1,356 @@
...
+using Microsoft.Azure.WebJobs.EventHubs.Processor;
+using Microsoft.Azure.WebJobs.Host.Executors;
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+using Azure.Core.Diagnostics;
+
+namespace Microsoft.Azure.WebJobs.EventHubs.Listeners
+{
+    internal sealed partial class EventHubListener
+	{
```
jsquire, line 22: nit:  Indenting seems to be off across the class.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubListener.PartitionProcessor.cs
```diff
@@ -0,0 +1,356 @@
...
+
+					while (_currentCycle.GetElapsedTime() < _maxWaitTime && !cancellationToken.IsCancellationRequested)
+					{
+						var remainingTime = RemainingTime(_currentCycle.GetElapsedTime());
+						await Task.Delay(remainingTime, cancellationToken).ConfigureAwait(false);
+					}
+
+                    if (!_cachedEventsGuard.Wait(0, cancellationToken))
+                    {
+                        await _cachedEventsGuard.WaitAsync(cancellationToken).ConfigureAwait(false);
```
jsquire, line 216: We need to release this in a `finally` block - but only if we actually acquire it.   It is possible for us to hit an exception that would cause `finally` to run before the semaphore is required.   We typically create a boolean var that we set to `true` right after the semaphore block to indicate that we need to release it and check that in `finally`.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Microsoft.Azure.WebJobs.Extensions.EventHubs.csproj
```diff
@@ -21,6 +21,7 @@
   <ItemGroup>
     <Compile Include="$(AzureCoreSharedSources)Argument.cs" LinkBase="SharedSource" />
     <Compile Include="$(AzureCoreSharedSources)ConnectionString.cs" LinkBase="SharedSource" />
+		<Compile Include="$(AzureCoreSharedSources)ValueStopwatch.cs" LinkBase="SharedSource" />
```
jsquire, line 24: nit:  indenting

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubListener.PartitionProcessor.cs
m-redding: had to redo all the tabs in this file but I think I fixed it

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/EventHubOptions.cs
```diff
@@ -132,6 +134,53 @@ public int MaxEventBatchSize
...
+            }
+        }
+
+        private TimeSpan _maxWaitTime;
+
+        /// <summary>
+        /// Gets or sets the maximum time that the trigger should wait to fill a batch before invoking the function.
+        /// This is only considered when <see cref="MinEventBatchSize"/> is set to larger than 1 and is otherwise unused.
+        /// If less than <see cref="MinEventBatchSize" /> events were available before the wait time elapses, the function
+        /// will be invoked with a partial batch.  Default is 60 seconds.  The maximum wait time is 10 minutes.
```
m-redding, line 164:
```suggestion
        /// will be invoked with a partial batch.  Default is 60 seconds.  The longest allowed wait time is 10 minutes.
```

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/PartitionProcessorEventsManager.cs
```diff
@@ -0,0 +1,90 @@
...
+                for (int i = 0; i < sizeOfBatch; i++)
+                {
+                    var nextEvent = CachedEvents.Dequeue();
+                    eventsToReturn[i] = nextEvent;
+                }
+            }
+            return eventsToReturn;
+        }
+
+        public void Dispose()
```
jsquire, line 85: Do we need to be disposable?   Since we're not cleaning anything up, it would be great if we could drop the interface/implementation.

jsquire (approved):

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/tests/EventHubEndToEndTests.cs
```diff
@@ -329,6 +329,58 @@ public async Task EventHub_MultipleDispatch_BinaryData()
...
+        public async Task EventHub_MultipleDispatch_MinBatchSize()
+        {
+            var (jobHost, host) = BuildHost<EventHubTestMultipleDispatchMinBatchSizeJobs>(
+                builder =>
+                {
+                    builder.ConfigureServices(services =>
+                    {
+                        services.Configure<EventHubOptions>(options =>
+                        {
+                            options.MinEventBatchSize = 5; // Increase from 1 to 5
```
JoshLove-msft, line 342: nit: use a constant for the number of events.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/tests/EventHubEndToEndTests.cs
```diff
@@ -708,6 +760,50 @@ public static void SendEvents_TestHub(int numEvents, string input, [EventHub(Tes
...
+                {
+                    PartitionKey = "key1"
+                };
+
+                // send one event at a time with a short time gap in between
+                for (int i = 0; i < numEvents; i++)
+                {
+                    var evt = new EventData(Encoding.UTF8.GetBytes(input));
+                    await client.SendAsync(new[] { evt }, options).ConfigureAwait(false);
+                    await Task.Delay(1000);
```
JoshLove-msft, line 781: Why do we have the delay?

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/PartitionProcessorEventsManager.cs
```diff
@@ -0,0 +1,90 @@
...
+        /// <summary>
+        /// Try to create a batch from <paramref name="events"/> and any events stored in the cache. If <paramref name="allowPartialBatch"/>
+        /// is true, this method return all events available even if there aren't enough to hit the minimum batch size threshold. If
+        /// <paramref name="allowPartialBatch"/> is false, this method either returns a batch of at least the minimum batch size, or it returns
+        /// nothing and retains all available events in the cache.
+        /// </summary>
+        /// <param name="events">An array of events to either add to a batch or cache.</param>
+        /// <param name="allowPartialBatch">True if batches smaller than the minimum batch size can be returned.</param>
+        /// <returns></returns>
+        public EventData[] TryGetBatchofEventsWithCached(EventData[] events = null, bool allowPartialBatch= false)
```
JoshLove-msft, line 51:
```suggestion
        public EventData[] TryGetBatchofEventsWithCached(EventData[] events = null, bool allowPartialBatch = false)
```

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/PartitionProcessorEventsManager.cs
```diff
@@ -0,0 +1,90 @@
...
+        {
+            get
+            {
+                return CachedEvents.Count > 0;
+            }
+        }
+
+        public PartitionProcessorEventsManager(int maxBatchSize, int minBatchSize)
+        {
+            CachedEvents = new Queue<EventData>();
```
JoshLove-msft, line 32: nit: this might yield a slight perf benefit

```suggestion
            CachedEvents = new Queue<EventData>(maxBatchSize);
```

JoshLove-msft: Actually, maybe not - for large batch sizes this could be wasteful.

JoshLove-msft (approved):

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/PartitionProcessorEventsManager.cs
m-redding: It can be dropped, that was there because of the semaphore (which is now in the other class)

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/tests/EventHubEndToEndTests.cs
m-redding: The delay is to space out the events a little to make sure they get dispatched to the `EventHubListener.PartitionProcessor` in separate calls, if they're sent to close together the host just puts them all in one array and only calls `EventHubListener.PartitionProcessor.ProcessEventsAsync` once

PR #34700, opened 2023-03-03
[AzureMonitorExporter] Change connection string error message.

rajkumar-rangaraj:


vishweshbankwar (approved):

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/ConnectionString/ConnectionStringParser.cs
```diff
@@ -44,7 +44,7 @@ public static ConnectionVars GetValues(string? connectionString)
             catch (Exception ex)
             {
                 AzureMonitorExporterEventSource.Log.WriteError("ConnectionStringError", ex);
-                throw new InvalidOperationException("Connection String Error: " + ex.Message, ex);
+                throw new InvalidOperationException("Connection String Error: A valid connection string is required to send telemetry data to Azure Monitor. Please set your connection string.", ex);
```
cijothomas, line 47: is this the message end users see? Could we make it more specific? what was invalid in the connection string? or was the string unparseable completely? or it was not given at all?

rajkumar-rangaraj: Earlier, we had the below exception. 
```
Unhandled exception. System.InvalidOperationException: A connection string was not found. This MUST be provided via either AzureMonitorExporterOptions or set in the environment variable 'APPLICATIONINSIGHTS_CONNECTION_STRING'
```

This error fits very well for exporter. But, when used with distro customer will set ConnectionString on `AzureMonitorOptions`. For distro users error was misleading. We made error message generic for now, and did not remove any logic specific connectionstring message.

We are planning to fix it together for both exporter and distro.

rajkumar-rangaraj: Addressed issue in https://github.com/Azure/azure-sdk-for-net/pull/34707

PR #34737, opened 2023-03-07
[DigitalTwins] Fix operation return type

HarveyLink:


sdk/digitaltwins/Azure.ResourceManager.DigitalTwins/src/Customized/DigitalTwinsDescriptionResource.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// <param name="cancellationToken"> The cancellation token to use. </param>
+        public virtual async Task<ArmOperation<DigitalTwinsDescriptionResource>> DeleteAsync(WaitUntil waitUntil, CancellationToken cancellationToken = default)
+        {
+            using var scope = _digitalTwinsDescriptionDigitalTwinsClientDiagnostics.CreateScope("DigitalTwinsDescriptionResource.Delete");
+            scope.Start();
+            try
+            {
+                var response = await _digitalTwinsDescriptionDigitalTwinsRestClient.DeleteAsync(Id.SubscriptionId, Id.ResourceGroupName, Id.Name, cancellationToken).ConfigureAwait(false);
+                var operation = new DigitalTwinsArmOperation<DigitalTwinsDescriptionResource>(Response.FromValue(new DigitalTwinsDescriptionResource(Client, response), response.GetRawResponse()));
```
ArcturusZhang, line 39: We should not add a Get request in delete request.
Instead we could change like this:
```suggestion
                var operation = new DigitalTwinsArmOperation<DigitalTwinsDescriptionResource>(Response.FromValue(this, response.GetRawResponse()));
```

ArcturusZhang (changes requested):
We should not change the RestOperations class. In this case, we could fix everything by just customize the generated methods in the Resource class.
We need to:
1. add some directive to fix this operation by removing its response schema.
2. add customized Delete and DeleteAsync method with the following suggested changes.

PR #34750, opened 2023-03-07
Better debugger support for DynamicJson

annelo-msft:
Implementing `GetDynamicMemberNames()` on DynamicJson enables the "Dynamic View" expansion Visual Studio:

image

Adding the `[DebuggerDisplay]` attribute makes the debugger experience the [same as that of JsonElement](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Text.Json/src/System/Text/Json/Document/JsonElement.cs#L1470-L1471).


sdk/core/Azure.Core.Experimental/src/DynamicJson.DynamicMetaObject.cs
```diff
@@ -16,8 +19,21 @@ public partial class DynamicJson : IDynamicMetaObjectProvider
...
+            private DynamicJson _value;
+
             internal MetaObject(Expression parameter, IDynamicMetaObjectProvider value) : base(parameter, BindingRestrictions.Empty, value)
             {
+                _value = (DynamicJson)value;
+            }
+
+            public override IEnumerable<string> GetDynamicMemberNames()
+            {
+                if (_value._element.ValueKind == JsonValueKind.Object)
```
christothes, line 31: could we add the same support for Arrays also?

christothes (approved):

sdk/core/Azure.Core.Experimental/src/DynamicJson.DynamicMetaObject.cs
annelo-msft: Arrays elements don't have member names, so I don't think this approach works for expanding arrays.  I wasn't find a way that expanding an enumerable was supported natively by IDMO.

The approach used in other types is to implement a [DebuggerTypeProxy](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.debuggertypeproxyattribute?view=net-7.0), but this is a bit more involved, and I wanted to improve the experience with a small investment in this PR - we can do more to improve it later if needed.

Thanks!

PR #34753, opened 2023-03-07
[Discussion only] One approach to enabling Console.WriteLine() on DynamicJson out of the box

annelo-msft:
If the only implicit cast on DynamicJson was to string, Console.WriteLine() would work out of the box.

I did a quick look into what this would look like - it effectively re-maps the semantics of a string cast to call ToString() if the underlying JSON doesn't have a string type.

It has the easier-to-spot disadvantages that now we wouldn't implicitly cast to a primitive type, so something like this would no longer work:
```
dynamic json = BinaryData.FromString("""{ "Foo" : 1 }""").ToDynamic();
int i = json.Foo;
```

Among the less-clear (to me :)) disadvantages are that now you _can_ do:
```
dynamic json = BinaryData.FromString("""{ "Foo" : 1 }""").ToDynamic();
string s = json.Foo;
```

When the following isn't allowed by the C# compiler:
```
int i = 1;
string s = (string)i;
```

This feels like we're removing too much type safety that could lead to errors.

Callers still have the option to call ToString() on any DynamicJson if those are the semantics they intend.

### DIFF
--- a/sdk/core/Azure.Core.Experimental/src/DynamicJson.Operators.cs
+++ b/sdk/core/Azure.Core.Experimental/src/DynamicJson.Operators.cs
@@ -21,66 +21,79 @@ namespace Azure.Core.Dynamic
         /// Converts the value to a <see cref="bool"/>.
         /// </summary>
         /// <param name="value">The value to convert.</param>
-        public static implicit operator bool(DynamicJson value) => value._element.GetBoolean();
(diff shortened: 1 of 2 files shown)


annelo-msft:
Note: a more common scenario is that callers will want to use Console.WriteLine() with interpolated strings.  This works out of the box, because the interpolated string handler calls ToString() in its implementation.

PR #34756, opened 2023-03-07
[Event Hubs] March Release Prep

jsquire:
# Summary

The focus of these changes is to prepare the Event Hubs core package for the March release.

### DIFF
--- a/sdk/eventhub/Azure.Messaging.EventHubs/CHANGELOG.md
+++ b/sdk/eventhub/Azure.Messaging.EventHubs/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Release History
 
-## 5.7.6 (2023-03-07)
+## 5.8.0 (2023-03-07)
+
+### Features Added
+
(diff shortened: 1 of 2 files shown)


m-redding (approved):

sdk/eventhub/Azure.Messaging.EventHubs/CHANGELOG.md
```diff
@@ -1,10 +1,6 @@
 # Release History
 
-## 5.8.0-beta.1 (Unreleased)
-
-### Features Added
-
-### Breaking Changes
+## 5.7.6 (2023-03-07)
```
JoshLove-msft, line 3: Sorry for not doing this already, but we should add a note about the OpenTelemetry updates.

JoshLove-msft: Also, it might be worth bumping to a new minor version for that.

jsquire: Got some verbiage to throw me and I'll add it in?

JoshLove-msft (approved):

PR #34774, opened 2023-03-07
[For Discussion] Early attempt to move logic from DynamicJson into DynamicData

annelo-msft:


sdk/core/Azure.Core.Experimental/src/DynamicTableEntity.cs
```diff
@@ -0,0 +1,71 @@
...
+
+// TODO: remove
+#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
+
+namespace Azure.Core.Dynamic
+{
+    /// <summary>
+    /// Dynamic layer over Azure.Tables.TableEntity.
+    /// </summary>
+    public class DynamicTableEntity : DynamicData
```
jsquire, line 17: This seems quite specific to be in Azure.Core.   Just a temporary stop before moving to the Tables package or are we thinking it'll live here forever?

sdk/core/Azure.Core.Experimental/src/DynamicData.cs
```diff
@@ -1,8 +1,23 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System;
```
jsquire, line 4: I missed this one... is this an attempt to pull DynamicJson up into a higher level abstraction or did this already exist?

sdk/core/Azure.Core.Experimental/src/DynamicTableEntity.cs
annelo-msft: Yes, it's just a proof of concept for this prototype - to show that DynamicData can be generalized to non-JSON types.  If we went with this approach, Krzysztof might put something like this into the CloudMachine code base.

sdk/core/Azure.Core.Experimental/src/DynamicData.cs
annelo-msft: This has been there the whole time, and it was always the intention to unify DynamicJson under this abstraction.

We've prioritized this as being on the critical path to GA'ing the type and moving it into Azure.Core, since we want to make sure we have this abstraction right before we would GA either type (or the full set of related types).  I'm actually working on a slightly different variant right now, and hope to put up a new draft PR with that in the next couple of days.

PR #34780, opened 2023-03-08
build(analyzer): replace `FxCopAnalyzers` with `NetAnalyzers`

### LINKED ISSUE #34779: [FEATURE REQ] Replace `Microsoft.CodeAnalysis.FxCopAnalyzers` with `Microsoft.CodeAnalysis.NetAnalyzers`
### Library name

All

## Description

We should replace `Microsoft.CodeAnalysis.FxCopAnalyzers` with `Microsoft.CodeAnalysis.NetAnalyzers` because:
- On [Roslyn analyzer pages](https://github.com/dotnet/roslyn-analyzers#main-analyzers), `FxCopAnalyzers` are obsolete and it's recommended to use `NetAnalyzers`
- Recently we got random reports of `CA1825` error which cannot be reproduced on everyone's machine. Initial analysis indicates it's a false positive and the solution is to use `NetAnalyzers`. See [here](https://learn.microsoft.com/bg-bg/dotnet/fundamentals/code-analysis/quality-rules/ca1825?view=vs-2017#when-to-suppress-warnings):
image

archerzz:
`FxCopAnalyzers` is obsolete and it's reocmmended to use `NetAnalyzers`. See correponding issue for details.

resolve #34779


archerzz:
@dvbb helped to verify that this PR can fix the false positive of `CA1825`. But replacing analyzers triggers more errors, probably due to newly added rules. Need further work on this.

archerzz:
I randomly check some SDKs and the new analyzer will not fail the build.

sdk/core/Azure.Core/src/Pipeline/HttpPipelineTransportOptions.cs
```diff
@@ -17,7 +17,9 @@ public class HttpPipelineTransportOptions
         /// </summary>
         public HttpPipelineTransportOptions()
         {
+#pragma warning disable CA1416
```
jsquire, line 20: Please add a comment describing what the rule is that is being suppressed and, if needed, why.

jsquire:
@heaths:  Would you please review?

eng/Directory.Build.Common.targets
```diff
@@ -87,7 +87,7 @@
 
     <PackageReference Condition="'$(EnableClientSdkAnalyzers)' == 'true'" Include="Azure.ClientSdk.Analyzers" PrivateAssets="All" />
 
-    <PackageReference Condition="'$(EnableFxCopAnalyzers)' == 'true'" Include="Microsoft.CodeAnalysis.FxCopAnalyzers" PrivateAssets="All" />
+    <PackageReference Condition="'$(EnableNetAnalyzers)' == 'true'" Include="Microsoft.CodeAnalysis.NetAnalyzers" PrivateAssets="All" />
```
heaths, line 90: It's unclear what this could break. You should add to a property group setting EnableNetAnalyzers based on EnableFxCopAnalyzers, which no-ops if neither is set. Add a comment this exists for legacy reasons.

Alternatively, you make this change as is and we see what breaks. @weshaggard @hallipr?

heaths (changes requested):

eng/Directory.Build.Common.targets
pallavit: I think it would be pragmatic to analyze the differences between the two analyzers and then enable them (once we have an idea of the impact on other SDKs). Looking at the pipelines we only have run core and batch as part of this PR and should run this against other SDKs before merge (it has a potential to break a large set of pipelines and block the code merge \ release).

archerzz: @pallavit Is there a command/script which can run all tests?

archerzz: @pallavit Is there a command or script we can build all projects in this repo?

weshaggard: You can build `eng/service.proj` without any service directory to build all of them locally.

heaths:
Do any of those errors happen without your changes? If they are regressed by your change, they'll need to be fixed. If not - if they already exist without your changes - then we can go ahead and merge.

archerzz:
> Do any of those errors happen without your changes? If they are regressed by your change, they'll need to be fixed. If not - if they already exist without your changes - then we can go ahead and merge.

@heaths They are not regressions by my change. The latter 4 have already been fixed on `main` branch. The former 3 (e.g. `xunit` style errors) exists on `main` branch as well. They are not detected probably because those test codes have not been touched for quite a long time. Other more active projects have worked around it: https://github.com/Azure/azure-sdk-for-net/blob/ac4b1b99adf4f0e6d886c5476089737d52f4dcd0/sdk/search/Microsoft.Azure.Management.Search/tests/ModelComparer/ModelComparerTests.cs#L90-L94

heaths (approved):

PR #34790, opened 2023-03-08
Throw mismatch exception if no variables section

### LINKED ISSUE #34781: [Test Framework] RecordedTestAttribute is not working for brand new tests
For tests without a recording, it should automatically run in record mode, but this isn't happening. We are not getting the expected RecordMismatchException.

JoshLove-msft:
Fixes https://github.com/Azure/azure-sdk-for-net/issues/34781

### DIFF
--- a/sdk/core/Azure.Core.TestFramework/src/TestRecording.cs
+++ b/sdk/core/Azure.Core.TestFramework/src/TestRecording.cs
@@ -65,6 +65,8 @@ namespace Azure.Core.TestFramework
                     catch (RequestFailedException ex)
                         when (ex.Status == 404)
                     {
+                        // We don't throw the exception here because Playback only tests that are testing the
+                        // recording infrastructure itself will not have session records.
                         MismatchException = new TestRecordingMismatchException(ex.Message, ex);
                         return;
                     }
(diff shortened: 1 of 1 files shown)


sdk/core/Azure.Core.TestFramework/src/TestRecording.cs
```diff
@@ -325,17 +328,25 @@ public string GetVariable(string variableName, string defaultValue, Func<string,
...
                     throw new ArgumentOutOfRangeException();
             }
         }
 
+        private void ValidateVariables()
+        {
+            if (Variables.Count == 0)
+            {
+                throw new TestRecordingMismatchException(
+                    "The record session does not exist or is missing the Variables section. If the test is " +
```
annelo-msft, line 344: Is it possible to tell which of the two cases you are in?  Or will re-recording the test fix both cases?

annelo-msft (approved):
LGTM

sdk/core/Azure.Core.TestFramework/src/TestRecording.cs
JoshLove-msft: Yeah - rerecording would fix both.

Issue #34784, opened 2023-03-08
[BUG] Unexpected type for header


kevinharing:
### Library name and version

Azure.Core 1.29.0

### Describe the bug

After upgrading to `Azure.Core` 1.29.0 (required by `Microsoft.Extensions.Azure` 1.6.2) we repeatedly receive this error:

```
System.InvalidOperationException: Unexpected type for header : 
 at Azure.Core.Pipeline.HttpClientTransport.PipelineRequest.GetHttpHeaderValue(String headerName, Object value)
 at Azure.Core.Pipeline.HttpClientTransport.PipelineRequest.EnumerateHeaders()+MoveNext()
 at Azure.Core.Diagnostics.AzureCoreEventSource.FormatHeaders(IEnumerable`1 headers, HttpMessageSanitizer sanitizer)
 at Azure.Core.Diagnostics.AzureCoreEventSource.Request(Request request, String assemblyName, HttpMessageSanitizer sanitizer)\n at Azure.Core.Pipeline.LoggingPolicy.ProcessAsync(HttpMessage message, ReadOnlyMemory`1 pipeline, Boolean async)
 at Azure.Core.Pipeline.BearerTokenAuthenticationPolicy.ProcessAsync(HttpMessage message, ReadOnlyMemory`1 pipeline, Boolean async)
 at Azure.Core.Pipeline.RedirectPolicy.ProcessAsync(HttpMessage message, ReadOnlyMemory`1 pipeline, Boolean async)\n at Azure.Core.Pipeline.RetryPolicy.ProcessAsync(HttpMessage message, ReadOnlyMemory`1 pipeline, Boolean async)
 at Azure.Core.Pipeline.RetryPolicy.ProcessAsync(HttpMessage message, ReadOnlyMemory`1 pipeline, Boolean async)
 at Azure.Core.Pipeline.HttpPipeline.SendRequestAsync(Request request, CancellationToken cancellationToken)
 at Azure.Messaging.ServiceBus.Administration.HttpRequestAndResponse.SendHttpRequestAsync(Request request, CancellationToken cancellationToken)\n at Azure.Messaging.ServiceBus.Administration.HttpRequestAndResponse.GetEntityAsync(String entityPath, String query, Boolean enrich, CancellationToken cancellationToken)
 at Azure.Messaging.ServiceBus.Administration.ServiceBusAdministrationClient.GetQueueAsync(String name, CancellationToken cancellationToken)
```

[...]

PR #34800, opened 2023-03-09
Fix for #34784: [BUG] Unexpected type for header

AlexanderSher:
The issue has been caused by multiple dispose of `ArrayBackedPropertyBag`. 

`ArrayBackedPropertyBag` is used to store headers in `PipelineRequest`. It uses array from the array pool to store items, which is returned back to the pool when `ArrayBackedPropertyBag` is disposed and [content of array is cleared](https://github.com/Azure/azure-sdk-for-net/blob/b0d465ffa8541b86c4ef7ae499e86f905a1ec26c/sdk/core/Azure.Core/src/Internal/ArrayBackedPropertyBag.cs#L280). However, since array is still referenced, second call of the `Dispose` will clear array again in case it is used by another instance of `ArrayBackedPropertyBag`. 

Double dispose of `ArrayBackedPropertyBag` has been caused by double dispose of `PipelineRequest` in [ManagedIdentitySource](https://github.com/Azure/azure-sdk-for-net/commit/b0d465ffa8541b86c4ef7ae499e86f905a1ec26c#diff-72571e3cca761ecd73c5855b39621f8883c8ee115319a0ecbb629deb5b8c0513L30-L31). New version of `Azure.Identity` has not been released yet, so `Azure.Messaging.ServiceBus` still uses the version where request is disposed twice.

### DIFF
--- a/sdk/core/Azure.Core/src/HttpMessage.cs
+++ b/sdk/core/Azure.Core/src/HttpMessage.cs
@@ -198,9 +198,15 @@ namespace Azure.Core
         /// </summary>
         public void Dispose()
         {
-            Request?.Dispose();
-            _response?.Dispose();
+            Request.Dispose();
             _propertyBag.Dispose();
+
+            var response = _response;
+            if (response != null)
+            {
+                _response = null;
+                response.Dispose();
+            }
         }
 
         private class ResponseShouldNotBeUsedStream : Stream
--- a/sdk/core/Azure.Core/src/HttpMessage.cs
+++ b/sdk/core/Azure.Core/src/HttpMessage.cs
@@ -24,6 +24,7 @@ namespace Azure.Core
         /// <param name="responseClassifier">The response classifier.</param>
         public HttpMessage(Request request, ResponseClassifier responseClassifier)
         {
+            Argument.AssertNotNull(request, nameof(Request));
             Request = request;
             ResponseClassifier = responseClassifier;
             BufferResponse = true;
--- a/sdk/core/Azure.Core/src/Internal/ArrayBackedPropertyBag.cs
+++ b/sdk/core/Azure.Core/src/Internal/ArrayBackedPropertyBag.cs
@@ -18,7 +18,7 @@ namespace Azure.Core
     {
         private Kvp _first;
         private Kvp _second;
-        private Kvp[] _rest;
+        private Kvp[]? _rest;
         private int _count;
 #if DEBUG
         private bool _disposed;
@@ -68,7 +68,7 @@ namespace Azure.Core
             {
                 0 => _first,
                 1 => _second,
-                _ => _rest[index - 2]
+                _ => _rest![index - 2] // Must throw if _rest is null
             };
         }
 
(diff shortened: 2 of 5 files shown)


sdk/core/Azure.Core/tests/ArrayBackedPropertyBagTests.cs
```diff
@@ -143,5 +144,75 @@ public void DeleteMultiple(int total, int increment)
...
+            Assert.AreEqual(count, target.Count);
+            for (var key = 0; key < count; key++)
+            {
+                Assert.IsTrue(target.TryGetValue(key, out var value));
+                Assert.AreEqual(key, value);
+            }
+
+            target.Dispose();
+#if DEBUG
+            Assert.Throws<InvalidOperationException>(() => { _ = target.Count; });
```
christothes, line 171: why not `ObjectDisposedException`?

christothes (approved):

sdk/core/Azure.Core/src/HttpMessage.cs
```diff
@@ -198,9 +198,15 @@ public void SetProperty(string name, object value)
         /// </summary>
         public void Dispose()
         {
-            Request?.Dispose();
-            _response?.Dispose();
+            Request.Dispose();
```
annelo-msft, line 201: In the prior version, we checked this for null.  Could Request ever be null here, would it make sense to keep that check?

sdk/core/Azure.Core/src/Internal/ArrayBackedPropertyBag.cs
```diff
@@ -68,7 +68,7 @@ public void GetAt(int index, out TKey key, out TValue value)
             {
                 0 => _first,
                 1 => _second,
-                _ => _rest[index - 2]
+                _ => _rest![index - 2] // Must throw if _rest is null
```
annelo-msft, line 71: Will this throw the desired exception type with a message that helps the customer understand how to address the issue?

sdk/core/Azure.Core/src/HttpMessage.cs
AlexanderSher: No, we don't expect it to be null. Probably, we should add an explicit check in constructor

sdk/core/Azure.Core/src/HttpMessage.cs
```diff
@@ -198,9 +198,15 @@ public void SetProperty(string name, object value)
...
             _propertyBag.Dispose();
+
+            var response = _response;
```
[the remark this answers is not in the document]
AlexanderSher: I want to set `_response` to null before calling `Dispose()`

sdk/core/Azure.Core/src/Internal/ArrayBackedPropertyBag.cs
```diff
@@ -260,35 +258,51 @@ private int GetIndex(TKey key)
...
                     return i + 2;
             }
             return -1;
         }
 
         internal void Dispose()
         {
-            CheckDisposed();
 #if DEBUG
+            if (_disposed)
```
KrzysztofCwalina, line 277: do we expect this to be thread safe? If yes, I think we need to add some memory barrier otherwise we can still dispose multiple times.

AlexanderSher: No, it is not thread-safe, and neither `PipelineRequest` or `HttpMessage`.

sdk/core/Azure.Core/src/HttpMessage.cs
```diff
@@ -198,9 +199,15 @@ public void SetProperty(string name, object value)
         /// </summary>
         public void Dispose()
         {
-            Request?.Dispose();
-            _response?.Dispose();
+            Request.Dispose();
```
JoshLove-msft, line 202: Don't we need a similar check to avoid double-disposing Request? Why not just use a _disposed flag?

JoshLove-msft: Ah I guess the double-dispose fix is in ArrayBackedPropertyBag itself.

sdk/core/Azure.Core/src/Internal/ArrayBackedPropertyBag.cs
JoshLove-msft, line 277: Why is this Debug only?

AlexanderSher: This field is used to catch the cases when `ArrayBackedPropertyBag` is used after disposing, but I don't want to crash user app if it happens cause with current implementation it is safe to use instance after disposing.

AlexanderSher: Also, that is a field that we can omit to make the struct smaller.

PR #34803, opened 2023-03-09
Monitor_dependency

### LINKED ISSUE #34782: Remove the hard-coded dependency in Azure.ResourceManager.Monitor test cases
### Library name

Azure.ResourceManager.Monitor

Some test cases in monitor mgmt plane test cases have some hard-coded dependent resources, we need to let the test cases be able to create them as well so that the test cases are repeatable.
For instance:
https://github.com/Azure/azure-sdk-for-net/blob/e1f27d7d40063dcfe6584a76dc35c697508bf919/sdk/monitor/Azure.ResourceManager.Monitor/tests/ResourceDataHelper.cs#L121
https://github.com/Azure/azure-sdk-for-net/blob/e1f27d7d40063dcfe6584a76dc35c697508bf919/sdk/monitor/Azure.ResourceManager.Monitor/tests/ResourceDataHelper.cs#L160
We have a few different approaches of doing this:
1. We could add a step to create them using the corresponding SDK package but we need to ensure we do not write the creation request of those dependent resources into the recording, otherwise when the dependent SDK upgrades its api-version (for instance compute RP), we will have to record the test in this RP again. We should avoid this case.
2. We could use an ArmTemplate to create the dependent resources and save the arm template json inside the test project.

mcgallan:
Fixes https://github.com/Azure/azure-sdk-for-net/issues/34782


sdk/monitor/Azure.ResourceManager.Monitor/tests/TestCase/MetricAlertCollectionTests.cs
```diff
@@ -14,10 +15,22 @@ namespace Azure.ResourceManager.Monitor.Tests
...
+        public async Task<string> GetStorageAccountId()
+        {
+            var resourceGroup = await CreateResourceGroupAsync();
+            var storageCeollection = resourceGroup.GetStorageAccounts();
+            string accountName = Recording.GenerateAssetName("metrictests");
+            var storageContent = ResourceDataHelper.GetContent();
+            var storageAccount = await storageCeollection.CreateOrUpdateAsync(WaitUntil.Completed, accountName, storageContent);
+            return storageAccount.Value.Data.Id;
+        }
+        #endregion
```
ArcturusZhang, line 32: We kind of want to exclude these operations from the recording. Otherwise whenever storage PR has an api-version upgrade, we will have to record monitor's test cases again as well

ArcturusZhang: You could refer to this on how to do that: https://github.com/Azure/azure-sdk-for-net/blob/abf92e3b4d559ac1cee311a274b7a3e3f662248c/sdk/servicebus/Azure.ResourceManager.ServiceBus/tests/Tests/ServiceBusNamespaceTests.cs#L566
Wrap the content of this function into a `if {} else {}`, if it is in playback mode, we just construct an id of storage account and return, otherwise we stop recording and call the API to create the storage account and return the id

mcgallan: fixed

archerzz (approved):

sdk/monitor/Azure.ResourceManager.Monitor/tests/TestCase/MetricAlertCollectionTests.cs
```diff
@@ -14,10 +15,34 @@ namespace Azure.ResourceManager.Monitor.Tests
...
         }
 
+        #region storage account
+        public async Task<string> GetStorageAccountId()
+        {
+            var resourceGroup = await CreateResourceGroupAsync();
+            var storageCeollection = resourceGroup.GetStorageAccounts();
+            string accountName = Recording.GenerateAssetName("metrictests");
+            var storageContent = ResourceDataHelper.GetContent();
+            var storageAccount = await storageCeollection.CreateOrUpdateAsync(WaitUntil.Completed, accountName, storageContent);
```
ArcturusZhang, line 29: We need to put all these into the else clause so that this traffic will not be recorded.

sdk/monitor/Azure.ResourceManager.Monitor/tests/TestCase/MetricAlertCollectionTests.cs
```diff
@@ -14,10 +15,34 @@ namespace Azure.ResourceManager.Monitor.Tests
...
+            string storageAccountId;
+            if (Mode == RecordedTestMode.Playback)
+            {
+                storageAccountId = $"/subscriptions/db1ab6f0-4769-4b27-930e-01e2ef9c123c/resourceGroups/{resourceGroup.Id.Name}/providers/Microsoft.Storage/storageAccounts/{accountName}";
```
ArcturusZhang, line 33: Please use the `StorageAccountResource.CreateResourceIdentifier` method, and use `resourceGroup.Id.SubscriptionId` for susbcriptionId instead of hard coding it here.

ArcturusZhang (changes requested):

sdk/monitor/Azure.ResourceManager.Monitor/tests/TestCase/MetricAlertCollectionTests.cs
mcgallan: fixed

sdk/monitor/Azure.ResourceManager.Monitor/tests/TestCase/MetricAlertCollectionTests.cs
mcgallan: fixed

ArcturusZhang (approved):

PR #34810, opened 2023-03-09
[For Discussion] Introduce ObjectDocument abstraction

annelo-msft:
This implementation boxes the MutableJsonElement struct - but is provided to look at the ObjectDocument/ObjectElement abstraction as a proof of concept.


sdk/core/Azure.Core.Experimental/api/Azure.Core.Experimental.net461.cs
```diff
@@ -111,97 +111,152 @@ namespace Azure.Core.Dynamic
...
+    {
+        protected ObjectDocument() { }
+        public abstract T As<T>(object element);
+        public abstract void Dispose();
+        public abstract Azure.Core.Dynamic.ObjectElement GetIndexElement(object element, int index);
+        public abstract bool HasValue(object element);
+        public abstract void Set(object element, object value);
+        public abstract Azure.Core.Dynamic.ObjectElement SetProperty(object element, string name, object value);
+        public abstract string? ToString(object element);
+        public abstract bool TryGetArrayLength(object element, out int length);
```
annelo-msft, line 199: Instead of ValueKind, whether or not the element is a given type is achieved through TryGet methods.  These will be on ObjectElement as well, with more variety.

sdk/core/Azure.Core.Experimental/api/Azure.Core.Experimental.net461.cs
```diff
@@ -191,17 +192,77 @@ public partial struct DynamicJsonOptions
         public string Name { get { throw null; } }
         public Azure.Core.Dynamic.DynamicJson Value { get { throw null; } }
     }
+    public abstract partial class ObjectDocument : System.IDisposable
+    {
+        protected ObjectDocument() { }
+        public abstract T As<T>(object element);
```
KrzysztofCwalina, line 192: do these method need to be public or they could be protected internal?

sdk/core/Azure.Core.Experimental/api/Azure.Core.Experimental.net461.cs
```diff
@@ -191,17 +192,77 @@ public partial struct DynamicJsonOptions
...
+        public abstract void Dispose();
+        public abstract Azure.Core.Dynamic.ObjectElement GetIndexElement(object element, int index);
+        public abstract bool HasValue(object element);
+        public abstract void Set(object element, object value);
```
KrzysztofCwalina, line 196: I think we should make this be TrySet. The reason is that I can imagine some properties being not settable.

sdk/core/Azure.Core.Experimental/api/Azure.Core.Experimental.net461.cs
```diff
@@ -111,97 +111,152 @@ namespace Azure.Core.Dynamic
...
-        public Azure.Core.Dynamic.DynamicJson Value { get { throw null; } }
+        public Azure.Core.Dynamic.DynamicData Value { get { throw null; } }
+    }
+    public abstract partial class ObjectDocument : System.IDisposable
+    {
+        protected ObjectDocument() { }
+        public abstract T As<T>(object element);
+        public abstract void Dispose();
+        public abstract Azure.Core.Dynamic.ObjectElement GetIndexElement(object element, int index);
+        public abstract bool HasValue(object element);
```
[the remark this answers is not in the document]
KrzysztofCwalina: I have to say HasValue is not a self-explanatory name. Since I am passing an element, how can this document not have its value?

sdk/core/Azure.Core.Experimental/api/Azure.Core.Experimental.net461.cs
```diff
@@ -191,17 +192,77 @@ public partial struct DynamicJsonOptions
...
+        public abstract void Set(object element, object value);
+        public abstract Azure.Core.Dynamic.ObjectElement SetProperty(object element, string name, object value);
+        public abstract string? ToString(object element);
```
KrzysztofCwalina, line 198: Why is this ToString (returning nullable) as opposed to TryGetString?

sdk/core/Azure.Core.Experimental/api/Azure.Core.Experimental.net461.cs
KrzysztofCwalina: It might be the best approach, but there is a tradeoff. In my deserialization code, I will have to keep trying many Try calls and possibly only the last one will succeed.

sdk/core/Azure.Core.Experimental/api/Azure.Core.Experimental.net461.cs
```diff
@@ -111,97 +111,152 @@ namespace Azure.Core.Dynamic
...
+        public abstract bool TryGetArrayLength(object element, out int length);
+        public abstract bool TryGetBoolean(object element, out bool value);
+        public abstract bool TryGetDouble(object element, out double value);
+        public abstract bool TryGetInt64(object element, out long value);
+        public abstract bool TryGetProperty(object element, string name, out Azure.Core.Dynamic.ObjectElement value);
+        public abstract bool TryGetPropertyNames(object element, out System.Collections.Generic.IEnumerable<string> enumerator);
```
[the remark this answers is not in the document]
KrzysztofCwalina: Would it be better to return an empty collection when there are no properties? instead of Try.

sdk/core/Azure.Core.Experimental/api/Azure.Core.Experimental.net6.0.cs
```diff
@@ -191,17 +192,77 @@ public partial struct DynamicJsonOptions
...
+        public abstract bool TryGetBoolean(object element, out bool value);
+        public abstract bool TryGetDouble(object element, out double value);
+        public abstract bool TryGetInt64(object element, out long value);
+        public abstract bool TryGetProperty(object element, string name, out Azure.Core.Dynamic.ObjectElement value);
+        public abstract bool TryGetPropertyNames(object element, out System.Collections.Generic.IEnumerable<string> enumerator);
+        public abstract bool TryGetString(object element, out string? value);
+        public abstract void WriteTo(object element, System.IO.Stream stream);
+    }
+    [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
+    public readonly partial struct ObjectElement
```
KrzysztofCwalina, line 209: Since this will be boxed every single time a method on it is called, does it still make sense to make it a struct?

sdk/core/Azure.Core.Experimental/api/Azure.Core.Experimental.net461.cs
annelo-msft: I see, yes that makes sense.  I can add a ValueKind enum instead - we will need to decide what those are ahead of time.

sdk/core/Azure.Core.Experimental/api/Azure.Core.Experimental.net461.cs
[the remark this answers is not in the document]
annelo-msft: Moving to ValueKind will let us remove this, if we follow the JSON convention and use Null as a value kind.

sdk/core/Azure.Core.Experimental/api/Azure.Core.Experimental.net461.cs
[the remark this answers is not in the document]
annelo-msft: We can do that - if we have Object as a ValueKind, we wouldn't need this TryGet to determine if it is an object or not.

sdk/core/Azure.Core.Experimental/api/Azure.Core.Experimental.net461.cs
annelo-msft: I had not considered the latter, but yes, that should work.

sdk/core/Azure.Core.Experimental/api/Azure.Core.Experimental.net461.cs
annelo-msft: It's because [object.ToString()](https://learn.microsoft.com/en-us/dotnet/api/system.object.tostring?view=net-8.0) returns `string?`.  We also have TryGetString() to get the value as a string.

sdk/core/Azure.Core.Experimental/api/Azure.Core.Experimental.net6.0.cs
annelo-msft: Yeah, I was hoping we'd find a way to do it without boxing - if we can't, then I think it does make sense for it to be a reference type instead.  In that case, I think it might make sense to make ObjectElement the abstract class rather than ObjectDocument, since that's the one that DynamicData calls into.  I'll have a look at that in the next iteration.

sdk/core/Azure.Core.Experimental/api/Azure.Core.Experimental.net461.cs
KrzysztofCwalina: Sorry, I asked the question very poorly. What I meant is that TryGetString seems to be doing exactly what this method does, and so why do we need this method in addition to TryGetString?

annelo-msft: Ah, I see.  I think we want the different methods - ToString() will return the value as a JSON string in all cases, for example, if it is a Number type, it would return e.g. `"5.1"`, or if it is an Object type, it would return `{"foo": "hi"}`.  TryGetString() will return false if the value is not a string type.  But looking at the [JsonElement API](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsonelement?view=net-8.0), I see that they to not have a TryGetString() - I guess that is because you can check the ValueKind property and if it is a string, ToString() will give you the string value.  We can follow their pattern if we introduce ValueKind.

KrzysztofCwalina: Also, I am not sure how callers of the abstraction (which can not be JSON) should think about this API, i.e. how would anybody use it successfully without knowing what this will return when called on an object.

sdk/core/Azure.Core.Experimental/api/Azure.Core.Experimental.net461.cs
annelo-msft: One observation here is that we end up back at the "JSON has fewer types than .NET" issue.  If we introduce an `ObjectValueKind` enum, we could map values to the smallest subset of types (e.g. JSON's number instead of int, long, double, etc) and let callers TryGet() to determine which .NET types are valid.  We could use CLR types, and adopt a convention that we assign kind the largest valid type and assume callers are able to downcast if they need it to be a different type.

annelo-msft: Or we could do something like:

```csharp
    public enum ObjectValueKind
    {
        Undefined,
        Object,
        Array,
        String,
        Integer,
        FloatingPoint,
        Boolean,
        Null
    }
```

Where we split JSON's Number into [Integer](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/integral-numeric-types) and [FloatingPoint](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/floating-point-numeric-types) and let users downcast from those.

sdk/core/Azure.Core.Experimental/api/Azure.Core.Experimental.net461.cs
annelo-msft: This is intended to just be a pass through to calling ToString() on the element.  I guess since ToString() is already on object, this is not needed.  Ok, removing.

sdk/core/Azure.Core.Experimental/api/Azure.Core.Experimental.net461.cs
annelo-msft: Updated to reflect this.  I'm not sure how this works in the case where you have 1-1 fidelity with CLR types (would you want to know it's an int and not a long in the calling code?), or where the set of types you're mapping to is potentially more expressive.

PR #34817, opened 2023-03-09
Fixes Triggered webjob endpoint location

deduefren:
### The problem faced

WebSite**Slot**TriggeredWebJobs were inside of WebSiteResource and WebSiteTriggeredWebjobs inside of WebSite**Slot**Resource instead of the other way around. 

This situation is forcing users to get a WebSiteSlotTriggeredWebJobResource without passing in a slot just to access the run method which is acting over a normal triggered WebJob without slot. From WebSiteTriggeredWebJobResource is not possible to trigger the run of the webjob. 

* Triggered Webjob without slot
 Before:
 `new ArmClient(new DefaultAzureCredential()).GetWebSiteSlotTriggeredWebJobResource(id).Run()`
After:
`new ArmClient(new DefaultAzureCredential()).GetWebSiteTriggeredWebJobResource(id).RunTriggeredWebJob()`

* Triggered Webjob within slot
Before:
 `new ArmClient(new DefaultAzureCredential()).GetWebSiteTriggeredwebJobResource(id).RunTriggeredWebJob()`
After:
`new ArmClient(new DefaultAzureCredential()).GetWebSiteSlotTriggeredWebJobResource(id).RunTriggeredWebJob()`

### Proposed solution

Changes made are just changing the request-path-to-resource-name to be of "Slot" when endpoint requires slot variable and just WebSiteTriggeredWebJob otherwise. This change will be considered a breaking change, please advice and comment if more changes need to be done.

Any recomendation or feedback are welcomed.


deduefren:

archerzz:
@deduefren Thanks for your contribution!
Unfortunately, the change will cause breaking change in generated SDK, and we have no breaking change rule. The workaround is to add customization codes to provide backward compatibility, see `compute` for reference: https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/compute/Azure.ResourceManager.Compute/src/Customize

Would you mind if you could try to add the customization? If it's not available, that's fine. I've pined corresponding owners for their awareness.

archerzz (dismissed):
Need to handle breaking change in API interface. See https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/compute/Azure.ResourceManager.Compute/src/Customize for reference

ArthurMa1978:
Thank you @deduefren for your contribution!

deduefren:
> @deduefren Thanks for your contribution! Unfortunately, the change will cause breaking change in generated SDK, and we have no breaking change rule. The workaround is to add customization codes to provide backward compatibility, see `compute` for reference: https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/compute/Azure.ResourceManager.Compute/src/Customize
> 
> Would you mind if you could try to add the customization? If it's not available, that's fine. I've pined corresponding owners for their awareness.

Thanks for your quick feedback! I'll try to perform the necessary changes to avoid breacking change and let you know. Many thanks!

deduefren:
> Need to handle breaking change in API interface. See https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/compute/Azure.ResourceManager.Compute/src/Customize for reference

Breaking change has been eliminated, please let me know if there is something else that needs changes.

ArthurMa1978 (approved):

PR #34825, opened 2023-03-10
Add ResultType discriminator to RecognizeResult and ReasonCode casting

atazimsft:
This pull request:
- Creates a RecognizeResultType struct to be used as a discriminator for RecognizeResult and its subclasses, ChoiceResult and CollectTonesResult, in order to make the types more discoverable via intellisense.

- Adds a GetReasonCodeValue() function to the ReasonCode struct to allow users to easily access the integer subcode
For example:
`int reasonCodeValue = playFailedEvent.ReasonCode.GetReasonCodeValue(); // 8536`


wangrui-msft:
Looks good to me, just need to run a api export on local to unblock the build pipeline.

wangrui-msft (approved):

cochi2 (commented):
overall it looks good, I have a couple of comments:
* what is the use case to have `GetReasonCodeValue`? BTW, we could also get the same value from the actual `RecognizeFailed` event through `resultInformation.SubCode`, is the reason to have this new method to help discoverability of such variable?
* Tests are missing, at least unit tests should be here to test that this works.

sdk/communication/Azure.Communication.CallAutomation/src/Models/RecognizeResultType.cs
```diff
@@ -0,0 +1,49 @@
...
+
+using System;
+using System.ComponentModel;
+
+namespace Azure.Communication.CallAutomation.Models
+{
+    /// <summary>
+    /// Determines the sub-type of the recognize result.
+    /// </summary>
+    public readonly partial struct RecognizeResultType: IEquatable<RecognizeResultType>
```
arupdutta-msft, line 12: Any reason that we cannot use enum here?

cochi2: Do we already know all the possible values this RecognizeResultType could get? If there is a chance that in the future there will be more types, we have to use the struct.

cochi2 (approved):

PR #34826, opened 2023-03-10
[Communication] - PhoneNumbers - Simplify status code assertions

danielortega-msft:
Due to a small backend change, tests that validate operations on unauthorized or invalid phone numbers now return status code 403, as opposed to 400 or 404 previously. To simplify the code of the test, the assertion now checks that the request has been rejected due to a client error, which is validated by checking if the status code falls under the 4xx bucket.

While it should be enough to validate that the operation itself fails, the check for client error was added to ensure that the request is indeed being rejected by the server instead of it failing due to another unrelated reason, which would usually be represented by a 5XX error.

A similar approach was followed to validate successful requests, instead of checking against a specific status code.

### DIFF
--- a/sdk/communication/Azure.Communication.PhoneNumbers/tests/PhoneNumbersClient/PhoneNumbersClientLiveTests.cs
+++ b/sdk/communication/Azure.Communication.PhoneNumbers/tests/PhoneNumbersClient/PhoneNumbersClientLiveTests.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
 using System.Threading.Tasks;
 using Azure.Communication.Tests;
 using Azure.Core.TestFramework;
@@ -212,7 +211,6 @@ namespace Azure.Communication.PhoneNumbers.Tests
             }
             catch (RequestFailedException ex)
             {
-                Assert.AreEqual(400, ex.Status);
                 Assert.NotNull(ex.Message);
             }
         }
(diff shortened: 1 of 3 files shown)


whisper6284 (approved):

sdk/communication/Azure.Communication.PhoneNumbers/tests/PhoneNumbersClient/PhoneNumbersClientLiveTests.cs
```diff
@@ -199,7 +198,7 @@ public async Task UpdateCapabilitiesAsync()
             Assert.IsTrue(updateOperation.HasCompleted);
             Assert.IsNotNull(updateOperation.Value);
             Assert.AreEqual(number, updateOperation.Value.PhoneNumber);
-            Assert.AreEqual(200, updateOperation.GetRawResponse().Status);
+            Assert.IsTrue(IsSuccess(updateOperation.GetRawResponse().Status));
```
whisper6284, line 201: Is there a way to have a log as to what the status code is in the case that this assert fails? I don't recall whether the actual value is logged by default when we do assert booleans.

danielortega-msft: Solved as part of [3836019](https://github.com/Azure/azure-sdk-for-net/pull/34826/commits/3836019f65c293a92f044cb689e08933d1c2c6bb)

whisper6284 (approved):

PR #34847, opened 2023-03-10
[ACR] Don't require Audience to be passed in ClientOptions

### LINKED ISSUE #34785: [ACR] Do not require audience on ClientOptions
@lmolkova is working with the service team to understand whether we will need to continue requiring callers to pass `Audience` into ACR clients, or whether we can set a default of "https://containerregistry.azure.net".

Once this is decided, implement the decided outcome.

annelo-msft:
Fixes https://github.com/Azure/azure-sdk-for-net/issues/34785


sdk/containerregistry/Azure.Containers.ContainerRegistry/src/ContainerRegistryClient.cs
```diff
@@ -72,19 +74,14 @@ public ContainerRegistryClient(Uri endpoint, TokenCredential credential, Contain
             Argument.AssertNotNull(credential, nameof(credential));
             Argument.AssertNotNull(options, nameof(options));
 
-            if (options.Audience == null)
```
pallavit, line 75: Was this done for consistency across languages? Is this a breaking change then?

annelo-msft: It's not a breaking change to remove an exception.  No client in the wild could be functioning without passing the Audience right now, so setting a default will not change any existing behavior.

@lmolkova confirmed with the service team that this is the correct path forward.

pallavit (approved):

PR #34859, opened 2023-03-11
Change settings types based on architect feedback

heaths:
Also updates the CHANGELOG dates.


pallavit:
> - Service version "7.4-preview.1" is not supported.

Do you want to also add the change of the type names in breaking? They would be different between the previews'

---
Refers to: sdk/keyvault/Azure.Security.KeyVault.Administration/CHANGELOG.md:7 in 71ea86e. [](commit_id = 71ea86e342c61e74ee7ba85d2c733baa53330b57, deletion_comment = False)

pallavit (approved):
:shipit:

sdk/keyvault/Azure.Security.KeyVault.Administration/api/Azure.Security.KeyVault.Administration.netstandard2.0.cs
```diff
@@ -311,20 +311,26 @@ public partial class KeyVaultSettingsClient
...
         public override int GetHashCode() { throw null; }
-        public static bool operator ==(Azure.Security.KeyVault.Administration.SettingType left, Azure.Security.KeyVault.Administration.SettingType right) { throw null; }
-        public static implicit operator Azure.Security.KeyVault.Administration.SettingType (string value) { throw null; }
-        public static bool operator !=(Azure.Security.KeyVault.Administration.SettingType left, Azure.Security.KeyVault.Administration.SettingType right) { throw null; }
+        public static bool operator ==(Azure.Security.KeyVault.Administration.KeyVaultSettingType left, Azure.Security.KeyVault.Administration.KeyVaultSettingType right) { throw null; }
+        public static implicit operator Azure.Security.KeyVault.Administration.KeyVaultSettingType (string value) { throw null; }
+        public static bool operator !=(Azure.Security.KeyVault.Administration.KeyVaultSettingType left, Azure.Security.KeyVault.Administration.KeyVaultSettingType right) { throw null; }
+        public override string ToString() { throw null; }
+    }
+    public partial class KeyVaultSettingValue
```
KrzysztofCwalina, line 330: should this be a readonly struct?

sdk/keyvault/Azure.Security.KeyVault.Administration/api/Azure.Security.KeyVault.Administration.netstandard2.0.cs
```diff
@@ -291,11 +292,10 @@ public partial class KeyVaultSelectiveKeyRestoreResult
     public partial class KeyVaultSetting
     {
         public KeyVaultSetting(string name, bool value) { }
+        public string Content { get { throw null; } }
         public string Name { get { throw null; } }
-        public Azure.Security.KeyVault.Administration.SettingType? Type { get { throw null; } }
-        public bool AsBoolean() { throw null; }
-        public string AsString() { throw null; }
-        public override string ToString() { throw null; }
+        public Azure.Security.KeyVault.Administration.KeyVaultSettingType? SettingType { get { throw null; } }
```
KrzysztofCwalina, line 297: I find it inconsistent that this property is called SettingType but the one below is not called SettingValue.

heaths: IMO, "SettingType" is already duplicative given the type is "KeyVaultSetting", but I appreciate we don't want properties called just "Type". That said, it seems even more extraneous to call the value "SettingValue".

tg-msft: Yes, I agree, this is a concession to avoid a property called just `Type`.  I prefer `Name` and `Value` for the other two properties and would relent on using `Type` before making it `SettingValue`.

sdk/keyvault/Azure.Security.KeyVault.Administration/api/Azure.Security.KeyVault.Administration.netstandard2.0.cs
```diff
@@ -311,20 +311,26 @@ public partial class KeyVaultSettingsClient
...
-        public static implicit operator Azure.Security.KeyVault.Administration.SettingType (string value) { throw null; }
-        public static bool operator !=(Azure.Security.KeyVault.Administration.SettingType left, Azure.Security.KeyVault.Administration.SettingType right) { throw null; }
+        public static bool operator ==(Azure.Security.KeyVault.Administration.KeyVaultSettingType left, Azure.Security.KeyVault.Administration.KeyVaultSettingType right) { throw null; }
+        public static implicit operator Azure.Security.KeyVault.Administration.KeyVaultSettingType (string value) { throw null; }
+        public static bool operator !=(Azure.Security.KeyVault.Administration.KeyVaultSettingType left, Azure.Security.KeyVault.Administration.KeyVaultSettingType right) { throw null; }
+        public override string ToString() { throw null; }
+    }
+    public partial class KeyVaultSettingValue
+    {
+        internal KeyVaultSettingValue() { }
```
tg-msft, line 332: Should this be constructible?

sdk/keyvault/Azure.Security.KeyVault.Administration/src/Generated/KeyVaultAdministrationModelFactory.cs
```diff
@@ -52,6 +52,16 @@ public static KeyVaultRoleAssignmentProperties KeyVaultRoleAssignmentProperties(
             return new KeyVaultRoleAssignmentProperties(scope, roleDefinitionId, principalId);
         }
 
+        /// <summary> Initializes a new instance of KeyVaultSetting. </summary>
+        /// <param name="name"> The account setting to be updated. </param>
+        /// <param name="content"> The value of the pool setting. </param>
+        /// <param name="settingType"> The type specifier of the value. </param>
+        /// <returns> A new <see cref="Administration.KeyVaultSetting"/> instance for mocking. </returns>
+        public static KeyVaultSetting KeyVaultSetting(string name = null, string content = null, KeyVaultSettingType? settingType = null)
```
tg-msft, line 60: Is `content` always a `string`?

sdk/keyvault/Azure.Security.KeyVault.Administration/src/KeyVaultSetting.cs
```diff
@@ -21,59 +28,37 @@ public KeyVaultSetting(string name, bool value)
...
-        /// Gets the boolean value of this account setting if <see cref="Type"/> is <see cref="SettingType.Boolean"/>.
+        /// Creates a new instance of the <see cref="KeyVaultSetting"/> class.
         /// </summary>
-        /// <returns>A boolean value if <see cref="Type"/> is <see cref="SettingType.Boolean"/>.</returns>
-        /// <exception cref="InvalidOperationException">The <see cref="Type"/> is not <see cref="SettingType.Boolean"/>, or the value cannot be normalized as a Boolean.</exception>
-        public bool AsBoolean() => CheckType(SettingType.Boolean, () => (bool.TryParse(Value, out bool parsedValue), parsedValue));
+        /// <param name="name">The name of the account setting.</param>
+        /// <param name="value">The boolean value of the account setting.</param>
+        /// <param name="settingType">The type specifier of the value.</param>
+        internal KeyVaultSetting(string name, string value, KeyVaultSettingType? settingType)
```
tg-msft, line 40: The comment calls it a `boolean value` but its typed as `string`?

sdk/keyvault/Azure.Security.KeyVault.Administration/src/KeyVaultSetting.cs
```diff
@@ -21,59 +28,37 @@ public KeyVaultSetting(string name, bool value)
...
-        /// <returns>The string value of this account setting.</returns>
-        public string AsString() => Value;
+        public KeyVaultSettingType? SettingType => Value.SettingType;
 
         /// <summary>
-        /// Gets the raw string value of this account setting.
+        /// Gets the value of the account setting.
         /// </summary>
-        /// <returns></returns>
-        public override string ToString()
```
tg-msft, line 49: We could leave this `ToString` if we think it adds value.  I'd just drop the method part and do `{Name}: {Value}` relying on `Value`'s `ToString`.

sdk/keyvault/Azure.Security.KeyVault.Administration/src/KeyVaultSettingValue.cs
```diff
@@ -0,0 +1,102 @@
...
+        {
+            string conversionMethod = null;
+            if (SettingType == KeyVaultSettingType.Boolean)
+            {
+                conversionMethod = nameof(AsBoolean);
+            }
+
+            return conversionMethod == null
+                ? $"{nameof(KeyVaultSetting)}: {nameof(SettingType)}={SettingType}"
+                : $"{nameof(KeyVaultSetting)}: {nameof(SettingType)}={SettingType}, {conversionMethod}()=>Value";
```
tg-msft, line 67: What do these look like?  Do we want the result to start with "KeyValueSetting: " when this is `KeyVaultSettingValue`?  Was the `=>Value` missing the actual value?

tg-msft (approved):

sdk/keyvault/Azure.Security.KeyVault.Administration/api/Azure.Security.KeyVault.Administration.netstandard2.0.cs
heaths: No, because 1) the `Value` property is read-only, and 2) it's set and controlled via constructing a `KeyVaultSetting` with only support types. Per previous discussion with @KrzysztofCwalina about this and conceptually matching what we did for the REST APIs, you can only create (post) a value that's supported, but we want to support any value coming back from the service since they could add new types in the future.

sdk/keyvault/Azure.Security.KeyVault.Administration/src/Generated/KeyVaultAdministrationModelFactory.cs
heaths: Yes. Their swagger defines `value` as a string, but I renamed it to `content` to match FR and because it couldn't be `Value` if that was returning our `KeyVaultSettingValue` type.

sdk/keyvault/Azure.Security.KeyVault.Administration/src/KeyVaultSetting.cs
heaths: Oops

sdk/keyvault/Azure.Security.KeyVault.Administration/src/KeyVaultSettingValue.cs
heaths: I copied this from FR, which hardcodes `=>Value` hence why I called out I found this odd over Teams. Personally, I'd rather `KeyVaultSetting.ToString` return `{Name}: {Value}` and `KeyVaultSettingValue.ToString` return just `{Value}` and get rid of `content`; though, I'd still need to figure out what to call it in the model factory since I can't use `Value` with an automatic signature. This started out just copying FR since you asked me to.

heaths: Overall, though, `KeyVaultSettingValue` is an implementation detail. You wanted a `Value` property, but it's not otherwise constructable. Everything is done through `KeyVaultSetting` otherwise, so that we can control the inputs for what the service supports currently. In hindsight, this was probably a bad service design: it's (intrinsic) polymorphism via string serialization.

heaths:
Resolved most of the comments offline. Working on changes now.

PR #34863, opened 2023-03-11
[Event Hubs] Update Perf/Stress Version

jsquire:
# Summary

The focus of these changes is to update the version of Event Hubs referenced for performance/stress testing.

### DIFF
--- a/eng/Packages.Data.props
+++ b/eng/Packages.Data.props
@@ -186,7 +186,7 @@
     <PackageReference Update="ApprovalTests" Version="3.0.22" />
     <PackageReference Update="ApprovalUtilities" Version="3.0.22" />
     <PackageReference Update="Azure.Identity" Version="1.7.0" />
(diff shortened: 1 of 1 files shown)


AlexanderSher (approved):

eng/Packages.Data.props
```diff
@@ -186,7 +186,7 @@
     <PackageReference Update="ApprovalTests" Version="3.0.22" />
     <PackageReference Update="ApprovalUtilities" Version="3.0.22" />
     <PackageReference Update="Azure.Identity" Version="1.7.0" />
-    <PackageReference Update="Azure.Messaging.EventHubs" Version="5.2.0" />
+    <PackageReference Update="Azure.Messaging.EventHubs" Version="5.8.1" />
```
weshaggard, line 189: Do we know why we need these versions to be different from the others listed above?

jsquire: We do not, and the existence of these overrides is something that we're discussing currently.   I would like to see them go away.

PR #34867, opened 2023-03-12
Prepare for release Azure.AI.TextAnalytics 5.3.0-beta.3

AlexanderSher:


jsquire (approved):

heaths (commented):
@joseharriaga is this actually needed, though, since you're preparing to release another beta? Or did that already go out?

joseharriaga:
@heaths It went out Tuesday last week, so I think making this change and re-releasing makes sense.

PR #34890, opened 2023-03-13
Add interface to phone number client

sofiar-msft:

Add an interface to the phone numbers client, so that they can use the interface for mocking purposes.


sdk/communication/Azure.Communication.PhoneNumbers/src/IPhoneNumbersClient.cs
```diff
@@ -0,0 +1,166 @@
...
+using Azure.Communication.Pipeline;
+using Azure.Core;
+using Azure.Core.Pipeline;
+
+namespace Azure.Communication.PhoneNumbers
+{
+    /// <summary>
+    /// The Azure Communication Services phone numbers client.
+    /// </summary>
+    public interface IPhoneNumbersClient
```
jsquire, line 16: This would seem to be in violation of the [Azure SDK Guidelines](https://azure.github.io/azure-sdk/dotnet_introduction.html#dotnet-appendix-overlookedguidelines) which state:

>  DO NOT use interfaces if you can use abstract classes. The only reasons to use an interface are: a) you need to “multiple-inherit”, b) you want structs to implement an abstraction.

Please discuss with @KrzysztofCwalina and @tg-msft prior to merging.

jsquire (changes requested):
Flagging to allow discussion of interface use with the .NET architects.

sdk/communication/Azure.Communication.PhoneNumbers/src/IPhoneNumbersClient.cs
KrzysztofCwalina: We make all methods on our clients virtual. This allows for mocking, i.e. not need for an interface. The following guide shows how such clients can be mocked: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/samples/Mocking.md

KrzysztofCwalina (changes requested):
All our clients are mockable, yet they don't implement any interfaces.

PR #34914, opened 2023-03-14
Improve test coverage calculation

ostoliarova-ee:
This PR addresses the [User Story 2539748](https://skype.visualstudio.com/SPOOL/_workitems/edit/2539748): [Pipeline][Identity][.NET] Test coverage improvement

The issue of Common and Common.Tests being included into the pipeline code calculation was in these projects being included in the Identity.Tests csproj file. This caused test coverage tool to include them in the total number of lines covered by tests.

It turned out that Common could be removed from being included in the  Identity.Tests csproj without any other code change. Probably it was included by mistake or there used to be some legacy code using Common, which is not the case any more.

Common.Tests is included to the test projects of other ACS SDKs as well resulting in Common.Tests being added to all SDK test coverage calculation. The reason for this change was an attempt to have one generic test class that reads required environment variables. As it doesn't have additional logic and contained variables not needed for Identity.Tests, I moved two helper classes that are reused in SDK tests of different modalitiesto the Shared folder and included by link like we do with other shared code. 
The updated code calculation result could be found here: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2629035&view=codecoverage-tab

This is an approach that other SDK owners might want to follow to update their tests, too in order to decrease coupling, make the tests more flexible and avoid reading/instantiating many unused values.


sdk/communication/Azure.Communication.Identity/tests/Azure.Communication.Identity.Tests.csproj
```diff
@@ -12,14 +12,14 @@
...
-    <ProjectReference Include="..\..\Azure.Communication.Common\src\Azure.Communication.Common.csproj" />
-    <ProjectReference Include="..\..\Azure.Communication.Common\tests\Azure.Communication.Common.Tests.csproj" />
     <ProjectReference Include="..\src\Azure.Communication.Identity.csproj" />
   </ItemGroup>
 
   <!-- Shared source from Azure.Core -->
   <ItemGroup>
     <Compile Include="$(AzureCoreSharedSources)ConnectionString.cs" LinkBase="Shared" />
+    <Compile Include="..\..\Shared\src\AuthMethod.cs" LinkBase="Shared\Communication.Tests" />
+    <Compile Include="..\..\Shared\src\CommunicationTestEnvironment.cs" LinkBase="Shared\Communication.Tests" />
```
mjafferi-msft, line 22: We can simply refer those two classes in just Identity.Tests.csproj file and don't need to make any change in Common.Tests.csproj.
Also, I would recommend referencing those files in different <ItemGroup> so that it becomes more clear that we are referencing them from Common Tests Project.

  
MicrosoftTeams-image

sdk/communication/Shared/src/AuthMethod.cs
```diff
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
petrsvihlik, line 1: let's keep test code separate from production code.

I'd recommend moving it to `sdk/communication/Shared/tests/`

sdk/communication/Azure.Communication.Identity/tests/Azure.Communication.Identity.Tests.csproj
petrsvihlik: what we're trying to achieve here is to move the shared test code outside from Common.Tests.csproj and keep that project solely for the purpose of testing the Common.csproj

sdk/communication/Azure.Communication.Common/tests/Azure.Communication.Common.Tests.csproj
```diff
@@ -14,6 +14,8 @@
     <Compile Include="$(AzureCoreSharedSources)ConnectionString.cs" LinkBase="Shared" />
     <Compile Include="..\..\Shared\src\HMACAuthenticationPolicy.cs" LinkBase="Shared\Communication" />
     <Compile Include="..\..\Shared\src\CommunicationBearerTokenCredential.cs" LinkBase="Shared\Communication" />
+    <Compile Include="..\..\Shared\src\AuthMethod.cs" LinkBase="Shared\Communication.Tests" />
+    <Compile Include="..\..\Shared\src\CommunicationTestEnvironment.cs" LinkBase="Shared\Communication.Tests" />
```
petrsvihlik, line 18: these two can be left out of the Common.Tests.csproj - the project doesn't utilize them.

instead, add them to other modalities in the same manner as you did for Identity, so that they're able to build.

petrsvihlik (changes requested):

sdk/communication/Shared/src/AuthMethod.cs
ostoliarova-msft: Done

sdk/communication/Azure.Communication.Common/tests/Azure.Communication.Common.Tests.csproj
ostoliarova-msft: Updated

sdk/communication/Azure.Communication.AlphaIds/tests/Azure.Communication.AlphaIds.Tests.csproj
```diff
@@ -10,8 +10,10 @@
   <ItemGroup>
     <ProjectReference Include="$(AzureCoreTestFramework)" />
     <ProjectReference Include="..\..\Azure.Communication.Common\src\Azure.Communication.Common.csproj" />
-    <ProjectReference Include="..\..\Azure.Communication.Common\tests\Azure.Communication.Common.Tests.csproj" />
     <ProjectReference Include="..\src\Azure.Communication.AlphaIds.csproj" />
+    <Compile Include="$(AzureCoreSharedSources)ConnectionString.cs" LinkBase="Shared" />
+    <Compile Include="..\..\Shared\tests\AuthMethod.cs" LinkBase="Shared\Communication.Tests" />
+    <Compile Include="..\..\Shared\tests\CommunicationTestEnvironment.cs" LinkBase="Shared\Communication.Tests" />
```
mjafferi-msft, line 16: I would suggest adding <Compile> tag in the seperate <ItemGroup> as it's done for Chat & Identity libraries

mjafferi-msft:
> /azp run net - Azure.Communication.Identity - tests

Also execute live tests of other packages to make sure their pipelines also executes successfully.

sdk/communication/Azure.Communication.Chat/tests/Azure.Communication.Chat.Tests.csproj
```diff
@@ -21,6 +20,8 @@
   <!-- Shared source from Azure.Core -->
```
petrsvihlik, line 20: this comment is not accurate. only the first row is azure core

sdk/communication/Azure.Communication.AlphaIds/tests/Azure.Communication.AlphaIds.Tests.csproj
petrsvihlik: you mean a separate ItemGroup? yeah we can do it, but doesn't make much of a difference.

petrsvihlik: you mean a separate ItemGroup? yeah we can do it, but doesn't make much of a difference.

petrsvihlik (approved):
this looks reasonable now. did it help the test coverage?

sdk/communication/Azure.Communication.AlphaIds/tests/Azure.Communication.AlphaIds.Tests.csproj
mjafferi-msft: Yeah separate <ItemGroup>. (Somehow it got deleted in my comment). Yes, won't make any difference as such, but just to keep it clear that these are shared test classes and to be consistent across all other libraries.

mjafferi-msft (approved):
This looks good. Now only library specific coverage is shown in the code coverage tab.

sdk/communication/Azure.Communication.AlphaIds/tests/Azure.Communication.AlphaIds.Tests.csproj
ostoliarova-msft: I agree that it doesn't influence the functionality but it's easy to make this change and the code will look more readable. I'll update

PR #34926, opened 2023-03-14
Add transmission state manager with exponential back off

vishweshbankwar:
This is the first PR towards refactoring the current mechanism for saving telemetry to storage and transmitting it back. The current implementation does not consider backing off exponentially in case of consecutive errors.

The current PR is adding `TransmissionStateManager` which will be used by the transmitter to decide whether it is ok to transmit to ingestion or directly save telemetry to storage. The usage will look like below.

```csharp
if (_transmissionStateManager.State == TransmissionState.Closed)
{
    // Transmit telemetry to ingestion
    //  if the response code from ingestion falls under one of the retryable errors
    //  then save telemetry offline and call EnableBackOff to set the back off time for all transmissions
      _transmissionStateManager.EnableBackOff(response);
   
   // else for success code
       _transmissionStateManager.ResetConsecutiveErrors();
        _transmissionStateManager.CloseTransmission();
 

}
else
{
    // Save telemetry offline
}
```

The implementation is based on CircuitBreaker pattern with some modifications


sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TransmissionState.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Azure.Monitor.OpenTelemetry.Exporter.Internals
+{
+    internal enum TransmissionState
```
TimothyMothra, line 10: nit: I recommend including a short comment about what the States represent.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TransmissionStateManager.cs
```diff
@@ -0,0 +1,94 @@
...
+            double delayInMilliseconds = MinDelayInMilliseconds;
+            if (_consecutiveErrors > 1)
+            {
+                double backOffSlot = (Math.Pow(2, _consecutiveErrors) - 1) / 2;
+                var backOffDelay = _random.Next(1, (int)Math.Min(backOffSlot * MinDelayInMilliseconds, int.MaxValue));
+                delayInMilliseconds = Math.Max(Math.Min(backOffDelay, MaxDelayInMilliseconds), MinDelayInMilliseconds);
+            }
+
+            return TimeSpan.FromMilliseconds(delayInMilliseconds);
+        }
```
TimothyMothra, line 92: 1. it's a little difficult to interpret how this works.
can you write a short summary explaining the function?

2. consider taking `consecutiveErrors` as an input to the method.
This would allow you to write unit tests and verify the math works as expected.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TransmissionStateManager.cs
```diff
@@ -0,0 +1,94 @@
...
+
+namespace Azure.Monitor.OpenTelemetry.Exporter.Internals
+{
+    internal class TransmissionStateManager
+    {
+        private int _consecutiveErrors;
+
+        private const long MaxDelayInMilliseconds = 3600000;
+
+        internal const long MinDelayInMilliseconds = 10000;
```
rajkumar-rangaraj, line 15: nit: consider private.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TransmissionStateManager.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        private readonly Random _random = new();
```
rajkumar-rangaraj, line 17: Why these fields are readonly if not initialized inside constructor?

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TransmissionStateManager.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        private readonly TimeSpan _minIntervalToUpdateConsecutiveErrors = TimeSpan.FromMilliseconds(20000);
+
+        private DateTimeOffset _nextMinTimeToUpdateConsecutiveErrors = DateTimeOffset.MinValue;
```
rajkumar-rangaraj, line 21: nit: Consider leaving a comment, if not we have to refer to `DateTimeOffset.MinValue` doc.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TransmissionStateManager.cs
```diff
@@ -0,0 +1,94 @@
...
+            }
+        }
+
+        internal TimeSpan GetBackOffTimeInterval()
+        {
+            double delayInMilliseconds = MinDelayInMilliseconds;
+            if (_consecutiveErrors > 1)
+            {
+                double backOffSlot = (Math.Pow(2, _consecutiveErrors) - 1) / 2;
+                var backOffDelay = _random.Next(1, (int)Math.Min(backOffSlot * MinDelayInMilliseconds, int.MaxValue));
```
rajkumar-rangaraj, line 87: Why are we picking a random value?

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/HttpPipelineHelper.cs
```diff
@@ -51,6 +51,23 @@ internal static int GetRetryInterval(Response httpResponse)
             return MinimumRetryInterval;
         }
 
+        internal static TimeSpan GetRetryIntervalTimespan(Response httpResponse)
+        {
+            if (httpResponse != null && httpResponse.Headers.TryGetValue(RetryAfterHeaderName, out var retryAfterValue))
+            {
+                if (int.TryParse(retryAfterValue, out var delaySeconds))
```
rajkumar-rangaraj, line 58: If there are 10 transmissions, then the 10th item will wait for all 9 previous items to complete their execution, as we are using Interlocked.Exchange. For example, if the 10th item has a retry-after value of 10 seconds and it has already spent 10 seconds waiting on the lock, are you planning to handle this scenario?

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/HttpPipelineHelper.cs
```diff
@@ -51,6 +51,23 @@ internal static int GetRetryInterval(Response httpResponse)
...
+                    return TimeSpan.FromSeconds(delaySeconds);
+                }
+                if (DateTimeOffset.TryParse(retryAfterValue, out DateTimeOffset delayTime))
+                {
+                    return (delayTime - DateTimeOffset.Now);
+                }
+            }
+
+            return TimeSpan.MinValue;
+        }
```
TimothyMothra, line 69: Consider changing this to a TryGet method.
My reasoning is that the MinValue return value is never used, instead it's compared in a ternary. 
In my mind, you can return a bool and save the extra comparison.

```
var retryAfterInterval = HttpPipelineHelper.GetRetryIntervalTimespan(response);
var backOffTimeInterval = retryAfterInterval != TimeSpan.MinValue ? retryAfterInterval : GetBackOffTimeInterval();
```

could be rewritten as:

```
var backOffTimeInterval = HttpPipelineHelper.TryGetRetryIntervalTimespan(response, out var retryAfterInterval )
    ? retryAfterInterval 
    : GetBackOffTimeInterval()
```

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TransmissionStateManager.cs
reyang: I guess it's for traffic shaping (e.g. several w3wp.exe instances started at almost the same time).

TimothyMothra:
I would have more confidence if we had some unit tests for this.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/HttpPipelineHelper.cs
vishweshbankwar: I am sure I understand the wait part. There is no waiting here. But, I just figured there is another issue with this.

vishweshbankwar:
Need to look in to an issue with consecutive errors. Will re-open after fixing that.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TransmissionStateManager.cs
vishweshbankwar: https://en.wikipedia.org/wiki/Exponential_backoff#:~:text=etc.%5B2%5D-,Collision%20avoidance,-%5Bedit%5D
In our case this^ can happen when we have many number of application instances sending data to same endpoint.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TransmissionStateManager.cs
vishweshbankwar: Without that as well, we need it to be readonly

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TransmissionStateManager.cs
```diff
@@ -0,0 +1,144 @@
...
+        /// Minimum time interval between failures to increment consecutive error count.
+        /// </summary>
+        private TimeSpan _minIntervalToUpdateConsecutiveErrors = TimeSpan.FromSeconds(20);
+
+        /// <summary>
+        /// Time threshold after which consecutive error count can be incremented.
+        /// </summary>
+        private DateTimeOffset _nextMinTimeToUpdateConsecutiveErrors = DateTimeOffset.MinValue;
+
+        private readonly System.Timers.Timer _backOffIntervalTimer;
```
TimothyMothra, line 27: `Timer` is Disposable.
This class will need to clean it up.

Azure.Core has this check turned off, but I think we need to find a way to turn it back on for our projects :(

vishweshbankwar: Added

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TransmissionStateManager.cs
vishweshbankwar: Could not find a way around it. Regarding the consecutive errors as input. It is intentional, I don't want to keep the old copy of it if it is updated by another thread.

rajkumar-rangaraj (approved):
LGTM

TimothyMothra (approved):

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TransmissionStateManager.cs
```diff
@@ -0,0 +1,165 @@
...
+            _minIntervalToUpdateConsecutiveErrors = minIntervalToUpdateConsecutiveErrors;
+            _nextMinTimeToUpdateConsecutiveErrors = nextMinTimeToUpdateConsecutiveErrors;
+            _backOffIntervalTimer = backOffIntervalTimer;
+            State = state;
+        }
+
+        /// <summary>
+        /// Stops transmitting data to backend.
+        /// </summary>
+        private void OpenTransmission()
```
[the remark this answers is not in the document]
vishweshbankwar: will take it in follow up.

PR #34936, opened 2023-03-15
Fix WebAppKeyInfo model 

JDECOMBE:
WebAppKeyInfo model doesn't match what the Rest API expects as body for the routes where it is used. This bug has been fixed in the Go SDK([here](https://github.com/Azure/azure-rest-api-specs/issues/14557)) but not in the .NET SDK.


JDECOMBE:

JDECOMBE:
I'm pretty sure the right way to fix this is via the swagger ([here](https://github.com/Azure/azure-rest-api-specs/blob/35f8a4df47aedc1ce185c854595cba6b83fa6c71/specification/web/resource-manager/Microsoft.Web/stable/2021-02-01/WebApps.json#L23134)), but looking at the Go SDK issue got me a bit confused on what was going on. 
So I just wanted to point out the problem. Please tell me if I need to do anything to get this fixed.

Issue #33614, opened 2023-01-20
Performance improvements to ContainerRegistryBlobClient


annelo-msft:
- Improve the performance of digest computation per https://github.com/Azure/azure-sdk-for-net/pull/33432#discussion_r1081950449
- Don't allocate a buffer if blob size is smaller than max chunk size per https://github.com/Azure/azure-sdk-for-net/pull/33432#discussion_r1081925298

KrzysztofCwalina:
FYI, I was thinking about something like this (probably 100x faster than the current SB implementation):

```CSharp    
    static readonly string prefixUtf16 = "sha256:";
    static readonly string nibbleToHexMap = "0123456789abcdef";
    public static bool TryFormatAsHexUtf16(byte[] bytes, scoped Span<char> utf16, out int charsWritten)
    {
        charsWritten = 0;
        if (utf16.Length < prefixUtf16.Length + (bytes.Length << 1)) return false;

        prefixUtf16.CopyTo(utf16);
        charsWritten = prefixUtf16.Length;

        for (int i = 0; i < bytes.Length; i++)
        {
            var nextByte = bytes[i];
            var high = (byte)(nextByte >> 4);
            utf16[charsWritten++] = nibbleToHexMap[high];
            var low = (byte)(nextByte & 0x0F);
            utf16[charsWritten++] = nibbleToHexMap[low];
        }

        return true;
    }
```

if you really need a string then this method could be called as:
```CSharp
    public static string BytesToString(byte[] bytes)
    {
        var length = prefixUtf16.Length + (bytes.Length << 1);
        char[] chars = new char[length];
        if (!TryFormatAsHexUtf16(bytes, chars, out var written)) throw new Exception("should not happen");
        return new string(chars);
    }
```

annelo-msft:
Another suggestion to consider here: https://github.com/Azure/azure-sdk-for-net/pull/34940#discussion_r1140506122

PR #34940, opened 2023-03-15
Add DownloadStreaming operation to ACR client

### LINKED ISSUE #34722: ACR API updates for parity with .NET Storage library
- [x] Change name of `DownloadBlob` to `DownloadBlobContent`
- [x] Provide an overload of `DownloadBlobTo` that takes a string `path`
- [x] Remove chunk size from ClientOptions

### LINKED ISSUE #34919: [ACR] Add DownloadStreaming method
Moving out of https://github.com/Azure/azure-sdk-for-net/issues/34722

annelo-msft:
This PR:
- Removes MaxChunkSize from ClientOptions
- Renames DownloadBlob to DownloadBlobContent
- Adds DownloadBlobStreaming
- Adds DownloadBlobTo that takes a path overload

Fixes https://github.com/Azure/azure-sdk-for-net/issues/34919
Fixes https://github.com/Azure/azure-sdk-for-net/issues/34722


sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/DownloadBlobStreamingResult.cs
```diff
@@ -0,0 +1,29 @@
...
+// Licensed under the MIT License.
+
+using System.IO;
+
+namespace Azure.Containers.ContainerRegistry.Specialized
+{
+    /// <summary>
+    /// The result from downloading a blob from the registry.
+    /// </summary>
+    public class DownloadBlobStreamingResult
```
annelo-msft, line 11: This should be IDisposable

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs
```diff
@@ -740,22 +739,140 @@ public virtual async Task<Response<DownloadBlobResult>> DownloadBlobAsync(string
...
+                async offset => await _blobRestClient.GetChunkAsync(_repositoryName, digest, new HttpRange(offset).ToString(), cancellationToken).ConfigureAwait(false),
+                _pipeline.ResponseClassifier,
+                _maxRetries,
+                (buffer, offset, length) =>
+                {
+                    if (validated)
+                        return;
+
+                    received += length;
+                    sha256.TransformBlock(buffer, offset, length, buffer, 0);
```
christothes, line 828: Should the last parameter always be 0, or track with `offset`?

sdk/core/Azure.Core/tests/RetriableStreamTests.cs
```diff
@@ -344,6 +344,48 @@ public async Task FlushDoesntThrow()
             await reliableStream.FlushAsync();
         }
 
+        [Test]
+        public async Task CanThrowOnRead()
```
christothes, line 348: should we also have some tests that just assert that both the `onRead` and `onDispose` Actions get called as expected?

sdk/core/Azure.Core/src/Shared/RetriableStream.cs
```diff
@@ -198,6 +228,7 @@ public override void Flush()
 
             protected override void Dispose(bool disposing)
             {
+                _onDispose?.Invoke();
```
AlexanderSher, line 231: - If `_onDispose` throws an exception, should we continue disposing the stream?
- If `_onDispose` recursively calls stream dispose, what should happen in this case?
- Do we care about `disposing` parameter?

sdk/core/Azure.Core/src/Shared/RetriableStream.cs
```diff
@@ -95,7 +119,10 @@ public override async Task<int> ReadAsync(byte[] buffer, int offset, int count,
                 {
                     try
                     {
-                        var result = await _currentStream.ReadAsync(buffer, offset, count, cancellationToken).ConfigureAwait(false);
+                        int result = await _currentStream.ReadAsync(buffer, offset, count, cancellationToken).ConfigureAwait(false);
+
+                        _onRead?.Invoke(buffer, offset, result);
```
AlexanderSher, line 124: Is it expected that `buffer` is writable here?

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/BlobHelper.cs
```diff
@@ -29,6 +30,13 @@ internal static string ComputeDigest(Stream stream)
             }
         }
 
+        internal static string ComputeDigest(BinaryData data)
+        {
+            using SHA256 sha256 = SHA256.Create();
```
AlexanderSher, line 35: Please use static implementation for .Net 6.0+: 
```cs
#if NET6_0_OR_GREATER
     return FormatDigest(SHA256.HashData(data));
#else
    using SHA256 sha256 = SHA256.Create();
    var hashValue = sha256.ComputeHash(data.ToArray());
    return FormatDigest(hashValue);
#endif
```

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/BlobHelper.cs
```diff
@@ -29,6 +30,13 @@ internal static string ComputeDigest(Stream stream)
             }
         }
 
+        internal static string ComputeDigest(BinaryData data)
```
AlexanderSher, line 33: Returning string here is a perf overhead. Since method is used only to compare with server-side digest, I'd rather pass it here as a parameter and instead of calling `FormatDigest`, compare the bytes

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs
```diff
@@ -740,22 +739,140 @@ public virtual async Task<Response<DownloadBlobResult>> DownloadBlobAsync(string
...
+            int contentLength = (int)blobResult.Headers.ContentLength.Value;
+            int received = 0;
+            bool validated = false;
+
+            // Wrap the response Content in a RetriableStream so we
+            // can return it before it's finished downloading, but still
+            // allow retrying if it fails.
+            Stream stream = RetriableStream.Create(
+                blobResult.Value,
+                offset => _blobRestClient.GetChunk(_repositoryName, digest, new HttpRange(offset).ToString(), cancellationToken).Value,
```
AlexanderSher, line 818: If it isn't too late (i.e. we don't use `RetriableStream` in too many places), maybe we should replace two delegates with one that has `async` parameter.

sdk/core/Azure.Core/src/Shared/RetriableStream.cs
annelo-msft: We don't intend to modify it in the _onRead method, good call out.  Is there a way to make it readonly?

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs
annelo-msft: Tracking this separately with https://github.com/Azure/azure-sdk-for-net/issues/34992

sdk/core/Azure.Core/src/Shared/RetriableStream.cs
annelo-msft: @AlexanderSher, these are great questions.  I know you've recently done some work with the dispose pattern, would you be willing to make a recommendation for the preferred approach in this case?

sdk/core/Azure.Core/tests/RetriableStreamTests.cs
annelo-msft: Good call out, will do - thanks!

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/BlobHelper.cs
annelo-msft: Tracking with https://github.com/Azure/azure-sdk-for-net/issues/33614

sdk/core/Azure.Core/tests/RetriableStreamTests.cs
annelo-msft: Now moot, per new approach.

sdk/core/Azure.Core/src/Shared/RetriableStream.cs
annelo-msft: Per offline conversation, using a new approach suggested by @AlexanderSher.

stdpower (approved):

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ValidatingStream.cs
```diff
@@ -0,0 +1,91 @@
...
+        private readonly SHA256 _sha256;
+        private readonly int _contentLength;
+        private readonly string _digest;
+
+        private int _received;
+        private bool _validated;
+
+        public ValidatingStream(Stream stream, int contentLength, string digest)
+        {
+            _stream = stream;
```
AlexanderSher, line 24: Let's do the null-check for `stream` here.

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/DownloadBlobToOptions.cs
```diff
@@ -0,0 +1,27 @@
...
+        /// <param name="maxChunkSize">The maximum size of chunk to transfer in a single request.</param>
+        public DownloadBlobToOptions(int maxChunkSize)
+        {
+            _maxChunkSize= maxChunkSize;
+        }
+
+        /// <summary>
+        /// The maximum size of chunk to transfer in a single request.
+        /// </summary>
+        public int MaxChunkSize => _maxChunkSize;
```
lmolkova, line 25: nit: any reason for **max**ChunkSize ? Why not just `ChunkSize`?

annelo-msft: It was to be approximately consistent with storage, which has [MaximumTransferSize](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/storage/Azure.Storage.Common/src/StorageTransferOptions.cs#L29) - I assume this is because the final chunk will often be smaller than the maximum size.  Shall we go with MaximumTransferSize instead?  I kept `chunk` because it is closer to the Registry API, but since we're aligning on other names, perhaps we should go in at 100%.

sdk/containerregistry/Azure.Containers.ContainerRegistry/api/Azure.Containers.ContainerRegistry.netstandard2.0.cs
```diff
@@ -239,10 +238,18 @@ public partial class ContainerRegistryBlobClient
...
         public virtual Azure.Response DownloadBlobTo(string digest, System.IO.Stream destination, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
+        public virtual Azure.Response DownloadBlobTo(string digest, string path, Azure.Containers.ContainerRegistry.Specialized.DownloadBlobToOptions options, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
+        public virtual Azure.Response DownloadBlobTo(string digest, string path, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
+        public virtual System.Threading.Tasks.Task<Azure.Response> DownloadBlobToAsync(string digest, System.IO.Stream destination, Azure.Containers.ContainerRegistry.Specialized.DownloadBlobToOptions options, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
```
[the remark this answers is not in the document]
annelo-msft: The .NET download mechanism downloads chunks in parallel up to a maximum number of workers and then orders them before writing them to the destination stream.  I do think we can do it, but if you'd prefer I can remove these from this release.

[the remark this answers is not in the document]
annelo-msft: Removed the overloads that take options from the public API.

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/DownloadBlobToOptions.cs
annelo-msft: I've removed the options type from the public API.

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/BlobHelper.cs
```diff
@@ -32,9 +32,13 @@ internal static string ComputeDigest(Stream stream)
 
         internal static string ComputeDigest(BinaryData data)
         {
+#if NET6_0_OR_GREATER
```
christothes, line 35: I believe we'll need to add net6.0 and net7.0 targets to the .csproj for this to ever take effect.

lmolkova (approved):

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/BlobHelper.cs
```diff
@@ -54,5 +58,15 @@ private static string BytesToString(byte[] bytes)
             }
             return builder.ToString();
         }
+
+        internal static void ValidateDigest(string clientDigest, string serverDigest, string message = default)
+        {
+            message ??= "The server-computed digest does not match the client-computed digest.";
```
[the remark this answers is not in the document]
annelo-msft: Done.

AlexanderSher (approved):
Great work!

PR #34953, opened 2023-03-16
Fix potential NullReferenceException

### LINKED ISSUE #34950: [BUG] NullReferenceException when updating concurrency in ServiceBusProcessor
### Library name and version

Azure.Messaging.ServiceBus 7.12.0.0

### Describe the bug

Under load we're getting periodic NullRefenceExceptions thrown from the Receiver Manager in response to a call to UpdateConcurrency:

Looking at the code for UpdatePrefetchCount, it is referencing the property for Receiver multiple times but this property can be updated while the method is running, meaning it can be set to null after the null check but prior to referencing the PrefetchCount property.

https://github.com/Azure/azure-sdk-for-net/blob/7b03cd1b95e378ad327b94e36ade6aaaf041a988/sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ReceiverManager.cs#L126-L132

When this happens, the ServiceBusSessionProcessor stops receiving messages.

### Expected behavior

Updating the concurrency should not cause the session processor to fail internally.

### Actual behavior

A null reference exception is thrown with the following stack trace:

```
Azure.Messaging.ServiceBus.ReceiverManager.UpdatePrefetchCount(int prefetchCount)
Azure.Messaging.ServiceBus.ServiceBusProcessor.ReconcileReceiverManagers(int maxConcurrentSessions, int prefetchCount)
Azure.Messaging.ServiceBus.ServiceBusProcessor.ReconcileConcurrencyAsync()
Azure.Messaging.ServiceBus.ServiceBusProcessor.RunReceiveTaskAsync(System.Threading.CancellationToken cancellationToken)
Azure.Messaging.ServiceBus.ServiceBusProcessor.RunReceiveTaskAsync(System.Threading.CancellationToken cancellationToken)
```

Once the exception occurs, the processor fails to retrieve any more messages from service bus

### Reproduction Steps

We are calling:

ServiceBusSessionProcessor.UpdateConcurrency(newTaskCount, 1);

While under high load.  It happens about once every 8 hours.

### Environment

Windows Service in .NET Framework 4.7.2

JoshLove-msft:
Fixes https://github.com/Azure/azure-sdk-for-net/issues/34950

### DIFF
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ReceiverManager.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ReceiverManager.cs
@@ -128,7 +128,8 @@ namespace Azure.Messaging.ServiceBus
         {
             var capturedReceiver = Receiver;
 
-            // If the Receiver property is set to null after we have captured a non-null instance, the Prefetch setter will no-op.
+            // If the Receiver property is set to null after we have captured a non-null instance, the Prefetch setter will essentially
+            // no-op as the underlying link has been closed.
             if (capturedReceiver != null && capturedReceiver.PrefetchCount != prefetchCount)
             {
                 capturedReceiver.PrefetchCount = prefetchCount;
--- a/sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ReceiverManager.cs
+++ b/sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ReceiverManager.cs
@@ -126,9 +126,12 @@ namespace Azure.Messaging.ServiceBus
 
         public virtual void UpdatePrefetchCount(int prefetchCount)
         {
-            if (Receiver != null && Receiver.PrefetchCount != prefetchCount)
+            var capturedReceiver = Receiver;
+
+            // If the Receiver property is set to null after we have captured a non-null instance, the Prefetch setter will no-op.
+            if (capturedReceiver != null && capturedReceiver.PrefetchCount != prefetchCount)
             {
-                Receiver.PrefetchCount = prefetchCount;
+                capturedReceiver.PrefetchCount = prefetchCount;
             }
         }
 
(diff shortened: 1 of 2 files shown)


sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ReceiverManager.cs
```diff
@@ -126,9 +126,13 @@ public virtual async Task ReceiveAndProcessMessagesAsync(CancellationToken cance
 
         public virtual void UpdatePrefetchCount(int prefetchCount)
         {
-            if (Receiver != null && Receiver.PrefetchCount != prefetchCount)
+            var capturedReceiver = Receiver;
+
+            // If the Receiver property is set to null after we have captured a non-null instance, the Prefetch setter will essentially
+            // no-op as the underlying link has been closed.
+            if (capturedReceiver != null && capturedReceiver.PrefetchCount != prefetchCount)
```
jsquire, line 133: If the captured receiver is not the current `Receiver` and we change the prefetch, what happens?   Is this a case where we potentially think we've updated the value but the new `Receiver` isn't aware?

Should we consider doing a reference check and resetting captured receiver if we don't have a `ReferenceEquals` match between `capturedReceiver` and `Receiver`?

jsquire (approved):

sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ReceiverManager.cs
JoshLove-msft: > If the captured receiver is not the current `Receiver` and we change the prefetch, what happens? Is this a case where we potentially think we've updated the value but the new `Receiver` isn't aware?

It's possible, but benign. The prefetch count would be updated on the next iteration of the RunReceiveTask loop. There isn't a way to eliminate this possibility without locking.

> Should we consider doing a reference check and resetting captured receiver if we don't have a ReferenceEquals match between capturedReceiver and Receiver?

The only way the Receiver would be different from the captured receiver is if the captured receiver has now been closed. In this case, the PrefetchCount update is a no-op as the link is closed.

jsquire:
> The prefetch count would be updated on the next iteration of the RunReceiveTask loop. 

👍  That addresses my concern.  We still reflect the right value eventually, timing is just delayed.

PR #34963, opened 2023-03-16
Set test in parallel to false for Service Bus tests

JoshLove-msft:
This should help avoid flakiness in tests that depend on a singleton diagnostic listener.

### DIFF
--- a/sdk/servicebus/service.projects
+++ b/sdk/servicebus/service.projects
@@ -21,6 +21,9 @@
   <ItemGroup Condition="'$(SDKType)' == 'client'">
     <ProjectsToIncludeBySDKType Include="$(MSBuildThisFileDirectory)Azure.Messaging.*\**\*.csproj" />
     <ProjectsToIncludeBySDKType Include="$(MSBuildThisFileDirectory)Microsoft.Azure.WebJobs.Extensions.ServiceBus\**\*.csproj" />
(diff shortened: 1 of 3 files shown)


sdk/servicebus/service.projects
```diff
@@ -21,6 +21,9 @@
   <ItemGroup Condition="'$(SDKType)' == 'client'">
     <ProjectsToIncludeBySDKType Include="$(MSBuildThisFileDirectory)Azure.Messaging.*\**\*.csproj" />
     <ProjectsToIncludeBySDKType Include="$(MSBuildThisFileDirectory)Microsoft.Azure.WebJobs.Extensions.ServiceBus\**\*.csproj" />
+    <ProjectReference Update="$(MSBuildThisFileDirectory)Azure.Messaging.*\tests\*.csproj">
```
jsquire, line 24: Can you help me understand what the effect of this is?

JoshLove-msft: This is essentially saying that the Azure.Messaging.ServiceBus tests cannot be run in parallel with other test assemblies.

jsquire (approved):

sdk/servicebus/service.projects
JoshLove-msft: @heaths discovered this recently and added to the docs - https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core.TestFramework/README.md#running-live-tests-serially

heaths (approved):

Issue #33080, opened 2022-12-15
[FEATURE REQ]/[BUG] [Distributed tracing] Report failures properly 


lmolkova:
### Library name

Azure.Core

### Please describe the feature.

1. Apparently, [`Scope.Failed(ex)`](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs#L354) does not work on OTel path (for `ActivitySource`).  We need to use `Activity.SetStatus` method there.

2. Also, we should add something like  `Scope.Failed(string)` that would set error  with optional description when there is no exception to report, like in [`RequestActivityPolicy`](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/Pipeline/Internal/RequestActivityPolicy.cs#L90) - `otel.status_code` was an old workaround and is not expected to work in a generic case 

3. We need to set error conditionally in `RequestActivityPolicy` depending on `RequestOptions` and don't fail HTTP activity when configured so.

PR #34975, opened 2023-03-17
Set status for failed activity

JoshLove-msft:
Contributes to https://github.com/Azure/azure-sdk-for-net/issues/33080

### DIFF
--- a/sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
+++ b/sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
@@ -139,9 +139,13 @@ namespace Azure.Core.Pipeline
             _activityAdapter?.Dispose();
         }
 
-        public void Failed(Exception e)
+        /// <summary>
+        /// Marks the scope as failed.
+        /// </summary>
+        /// <param name="exception">The exception to associate with the failed scope.</param>
+        public void Failed(Exception? exception = default)
         {
-            _activityAdapter?.MarkFailed(e);
+            _activityAdapter?.MarkFailed(exception);
         }
 
         /// <summary>
@@ -368,12 +372,16 @@ namespace Azure.Core.Pipeline
                 _currentActivity?.SetStartTime(startTime);
             }
 
-            public void MarkFailed(Exception exception)
+            public void MarkFailed(Exception? exception)
             {
-                _diagnosticSource?.Write(_activityName + ".Exception", exception);
+                if (exception != null)
+                {
+                    _diagnosticSource?.Write(_activityName + ".Exception", exception);
+                }
+
                 if (ActivityExtensions.SupportsActivitySource())
                 {
-                    _currentActivity?.SetErrorStatus(exception.ToString());
+                    _currentActivity?.SetErrorStatus(exception?.ToString());
                 }
             }
 
@@ -430,7 +438,7 @@ namespace Azure.Core.Pipeline
         private static Action<Activity, int>? SetIdFormatMethod;
         private static Func<Activity, string?>? GetTraceStateStringMethod;
         private static Action<Activity, string?>? SetTraceStateStringMethod;
-        private static Action<Activity, int, string>? SetErrorStatusMethod;
+        private static Action<Activity, int, string?>? SetErrorStatusMethod;
         private static Func<Activity, int>? GetIdFormatMethod;
         private static Action<Activity, string, object?>? ActivityAddTagMethod;
         private static Func<object, string, int, object?, ICollection<KeyValuePair<string, object>>?, IList?, DateTimeOffset, Activity?>? ActivitySourceStartActivityMethod;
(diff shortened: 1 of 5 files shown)


JoshLove-msft:
@lmolkova - are there any changes needed here - https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/Pipeline/Internal/RequestActivityPolicy.cs#L112?

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
```diff
@@ -569,6 +575,40 @@ public static void SetTraceState(this Activity activity, string? tracestate)
...
+                        SetErrorStatusMethod = (_, _, _) => { };
+                    }
+                    else
+                    {
+                        var methodParameters = method.GetParameters();
+
+                        var statusParameter = Expression.Parameter(typeof(int));
+                        var descriptionParameter = Expression.Parameter(typeof(string));
+                        SetErrorStatusMethod = Expression.Lambda<Action<Activity, int, string>>(
+                            Expression.Call(ActivityParameter, method, Expression.Convert(statusParameter, methodParameters[0].ParameterType), descriptionParameter),
```
christothes, line 604: Would it make sense to cache these lambdas?

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
```diff
@@ -371,6 +371,10 @@ public void SetStartTime(DateTime startTime)
             public void MarkFailed(Exception exception)
             {
                 _diagnosticSource?.Write(_activityName + ".Exception", exception);
+                if (ActivityExtensions.SupportsActivitySource())
+                {
+                    _currentActivity?.SetErrorStatus(exception.ToString());
```
lmolkova, line 376: does it also affect HTTP requests which return 4xx, 5xx? I.e. when we retry those, we should see

- outer logical activity  - error (after all tries)
  - HTTP try 1  - (DNS exception) - error 
  - HTTP try 2  - (503)  - error 
  - HTTP try 3  - (429)  - error 
  - .etc

sdk/core/Azure.Core/tests/ClientDiagnosticsTests.Net50.cs
```diff
@@ -429,6 +429,45 @@ public void ParentIdCannotBeSetOnStartedScopeActivitySource()
             scope.Start();
             Assert.Throws<InvalidOperationException>(() => scope.SetTraceContext(parentId));
         }
+
+        [Test]
+        [NonParallelizable]
+        public void FailedStopsActivityAndWritesExceptionEventActivitySource()
```
lmolkova, line 435: do we have any tests for http activities where we can test it as well?

lmolkova (commented):
LGTM! I wonder if we have HTTP tests whe

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
JoshLove-msft: We do cache these.

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
```diff
@@ -569,6 +575,40 @@ public static void SetTraceState(this Activity activity, string? tracestate)
             SetTraceStateStringMethod(activity, tracestate);
         }
 
+        public static void SetErrorStatus(this Activity activity, string description)
+        {
+            if (SetErrorStatusMethod == null)
+            {
+                if (ActivityStatusCodeType == null)
+                {
+                    SetErrorStatusMethod = (_, _, _) => { };
```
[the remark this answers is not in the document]
JoshLove-msft: We set it to a no-op so to avoid using reflection whenever the method is called to attempt to look it up, i.e. we have 3 states - null (before reflection is used for the first time), no-op (reflection was used but we don't have the method on this version), filled out method (we have the method on this version).

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
christothes: Oh, I see. `SetErrorStatusMethod` persists for future invocations.

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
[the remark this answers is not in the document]
JoshLove-msft: That's fair - I suppose a nullable bool could accomplish the same thing. However, I'm hesitant to change this as we use the same pattern throughout, and these are static lambdas so perf impact should be minimal.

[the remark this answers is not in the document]
annelo-msft: Cool, yeah, I am fine with whatever you decide.

lmolkova (approved):
Looks great!

PR #34976, opened 2023-03-17
[Text Analytics] Mitigate transient test failures due to "InternalServerError" with 200 OK

### LINKED ISSUE #34970: [Text Analytics] Mitigate transient test failures due to "InternalServerError" with 200 OK
Occasionally during our test runs, the service responds back with a 200 OK response status but includes an "InternalServerError" in the response body for one or more text analysis tasks in that batch. In the case of LROs, the client cannot recover using its retry policy. Implement a solution to improve test reliability.

joseharriaga:
Resolves: https://github.com/Azure/azure-sdk-for-net/issues/34970

I'm adding a custom `NUnitAttribute` that I'm calling `RetryOnInternalServerErrorAttribute` and which is basically a simple modification of the [`RetryAttribute`](https://github.com/nunit/nunit/blob/30e1255feb2c7108e6b7b168126360ac3689d4bb/src/NUnitFramework/framework/Attributes/RetryAttribute.cs) in the NUnit framework. Its purpose is to specify that a test must be retried in the specific case of a known transient issue where the server returns a successful 200 OK response status but reports an internal server error in the response body for one or more text analysis tasks. If the test continues to fail due to this particular issue after reaching the max try count, the test result is marked as inconclusive.

### DIFF
--- a/sdk/core/Azure.Core.TestFramework/src/RetryOnErrorAttribute.cs
+++ b/sdk/core/Azure.Core.TestFramework/src/RetryOnErrorAttribute.cs
@@ -1,0 +1,117 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using NUnit.Framework;
+using NUnit.Framework.Interfaces;
+using NUnit.Framework.Internal;
+using NUnit.Framework.Internal.Commands;
+
+namespace Azure.Core.TestFramework
+{
+    /// <summary>
+    /// Attribute used to specify that a test must be retried in the specific case of a known error. If the test
+    /// continues to fail due to this particular issue after reaching the specified limit of tries, the test result is
+    /// marked as inconclusive.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = false, Inherited = false)]
+    public class RetryOnErrorAttribute : NUnitAttribute, IRepeatTest
+    {
+        private readonly int _tryCount;
+
+        private readonly Func<TestExecutionContext, bool> _shouldRetry;
+
+        /// <summary>
+        /// Initializes a <see cref="RetryOnErrorAttribute"/> instance.
+        /// </summary>
+        /// <param name="tryCount">The number of times that the test can be executed.</param>
+        /// <param name="shouldRetry">The function that determines whether the test should be retried.</param>
+        public RetryOnErrorAttribute(int tryCount, Func<TestExecutionContext, bool> shouldRetry)
+        {
+            _tryCount = tryCount;
+            _shouldRetry = shouldRetry;
+        }
+
+        #region IRepeatTest Members
+
+        /// <summary>
+        /// Wrap a command and return the result.
+        /// </summary>
+        /// <param name="command">The command to wrap.</param>
+        /// <returns>The wrapped command.</returns>
+        public TestCommand Wrap(TestCommand command)
+        {
+            return new RetryOnErrorCommand(command, _tryCount, _shouldRetry);
+        }
+
+        #endregion
+
+        /// <summary>
+        /// The test command for the <see cref="RetryOnErrorAttribute"/>.
+        /// </summary>
+        public class RetryOnErrorCommand : DelegatingTestCommand
+        {
+            private readonly int _tryCount;
+
+            private Func<TestExecutionContext, bool> _shouldRetry;
+
+            /// <summary>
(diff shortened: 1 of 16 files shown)


sdk/textanalytics/Azure.AI.TextAnalytics/tests/Infrastructure/RetryOnInternalServerErrorAttribute.cs
heaths, line 1: Any reason not to add this to Azure.Core.TestFramework as I did with the IgnoreServerErrorAttribute? Could make the matching message customizable.

sdk/textanalytics/Azure.AI.TextAnalytics/tests/Infrastructure/RetryOnInternalServerErrorAttribute.cs
```diff
@@ -0,0 +1,106 @@
...
+                        context.CurrentResult = innerCommand.Execute(context);
+                    }
+                    catch (Exception ex)
+                    {
+                        context.CurrentResult ??= context.CurrentTest.MakeTestResult();
+                        context.CurrentResult.RecordException(ex);
+                    }
+
+                    if (!ShouldRetry(context))
+                        break;
```
heaths, line 74: Stylistic: we prefer braces around everything. Was there not at least a warning on this? There used to be.

sdk/textanalytics/Azure.AI.TextAnalytics/tests/Infrastructure/RetryOnInternalServerErrorAttribute.cs
```diff
@@ -0,0 +1,106 @@
...
+                    }
+                }
+
+                return context.CurrentResult;
+            }
+
+            /// <summary>
+            /// Indicates whether the encountered exception corresponds to the issue in question.
+            /// </summary>
+            private bool ShouldRetry(TestExecutionContext context)
```
heaths, line 95: Nit: could be static unless we did want to make it more general purpose.

sdk/textanalytics/Azure.AI.TextAnalytics/tests/AbstractSummaryTests.cs
```diff
@@ -76,6 +77,7 @@ public AbstractSummaryTests(bool isAsync, TextAnalyticsClientOptions.ServiceVers
         private const int AbstractiveSummarizationSentenceCount = 3;
 
         [RecordedTest]
+        [RetryOnInternalServerError]
```
heaths, line 80: Can an `IRepeatTest` be on a class? Might be better to support either so it's usage isn't so verbose.

heaths (approved):

deyaaeldeen (approved):

mssfang (approved):

sdk/textanalytics/Azure.AI.TextAnalytics/tests/AbstractSummaryTests.cs
joseharriaga: I wanted this too, but unfortunately, it only works with test methods. 😢

sdk/textanalytics/Azure.AI.TextAnalytics/tests/Infrastructure/RetryOnInternalServerErrorAttribute.cs
joseharriaga: I moved it to Azure.Core.TestFramework and made it customizable. Let me know what you think.

sdk/core/Azure.Core.TestFramework/src/RetryOnErrorAttribute.cs
```diff
@@ -0,0 +1,117 @@
...
+        private readonly int _tryCount;
+
+        private readonly Func<TestExecutionContext, bool> _shouldRetry;
+
+        /// <summary>
+        /// Initializes a <see cref="RetryOnErrorAttribute"/> instance.
+        /// </summary>
+        /// <param name="tryCount">The number of times that the test can be executed.</param>
+        /// <param name="shouldRetry">The function that determines whether the test should be retried.</param>
+        public RetryOnErrorAttribute(int tryCount, Func<TestExecutionContext, bool> shouldRetry)
```
heaths, line 29: This couldn't possibly be serialized as an attribute parameter, could it? If this is intended to be derived, I'd make the class abstract and this constructor protected. Or would the default implementation be good enough with just a `tryCount` to put in a public constructor, and still make this protected but keep the class instantiable?

heaths (approved):

Issue #34958, opened 2023-03-16
[ACR] Follow-ups from Arch Board review


annelo-msft:
This issue tracks the work for .NET coming out of the [ACR arch board review](https://github.com/Azure/azure-sdk/issues/5643).

Per comments from the [.NET APIView](https://apiview.dev/Assemblies/Review/7ebfcb36ff6149ba96170bd059094513/f989159861d846b3abb4384d01a27da9?diffRevisionId=500d32ab2730498084dedadf61d740fc&doc=False&diffOnly=True):
- [x] Move client out of the `.Specialized` namespace
- [ ] Rename client
- [x] Change `RepositoryName` property to `Repository` (or be consistent with the name of the parameter passed to the constructor)
- [x] Update `DownloadBlobTo()` doc comments to clarify return type (use Storage as an example)
- [x] Update doc comment on `UploadBlob` to clarify that RawResponse returned is the last of the chunks
- [x] Rename parameters to `content` in `UploadBlob` methods
- [x] Rename parameters to `manifest` in `UploadManifest` methods
- [x] Remove `UploadManifest` overload that takes a Stream
- [ ] Rename `DownloadBlobResult` to something Registry-specific
- [ ] Rework how deserialization to a strong type for manifest works
- [x] Change type of `OciAnnotations.Url` to `Uri`
- [ ] Consider adding an extensible enum of known media types for `OciDescriptor.MediaType`
- [ ] Move `schemaVersion=2` into a swagger constant
- [x] Change `OciImageManifest.Config` to `Configuration`

PR #34987, opened 2023-03-17
ACR Updates from Arch Board review

annelo-msft:
Contributes to https://github.com/Azure/azure-sdk-for-net/issues/34958

This PR includes the following subset of changes recommended by architects:
- Move ContainerRegistryBlobClient into main namespace
- Change `repository` parameter in constructor to `repositoryName` to match main client
- Change parameter names to blob methods to `content` and to manifest methods to `manifest`
- Change URI properties on OciAnnotations to `Uri` type
- Change `Config` to `Configuration`
- Update doc comments to indicate which response is returned when service method makes multiple calls


lmolkova (commented):
I should have raised it during arch-board, but I'd prefer to keep `config` and not rename it to `configuration` - this is how it's called in the [oci spec](https://github.com/opencontainers/image-spec/blob/main/manifest.md). Divergence from the spec terminology seems to hurt user experience more than improve it.

Other that that changes look great to me!

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/DownloadManifestResult.cs
```diff
@@ -5,7 +5,7 @@
 using System.IO;
 using System.Text.Json;
 
-namespace Azure.Containers.ContainerRegistry.Specialized
+namespace Azure.Containers.ContainerRegistry
 {
     /// <summary>
     /// The result from downloading an OCI manifest from the registry.
```
lmolkova, line 11: nit and out of scope on this PR: wonder if we should rename `AsOciManifest` to `AsOciImageManifest` across the languages (so we can add `AsOciArtifactManifest` if/when artifact manifest comes)

annelo-msft: The feedback from Ted was that we should change the name of this method to make it clear it is deserializing the BinaryData Content property.

lmolkova (approved):
LGTM per offline discussion!

PR #34555, opened 2023-02-25
[AzureMonitorExporter] Add AAD support through property

rajkumar-rangaraj:
Enabling AAD support through property will help the Azure Monitor Distro to keep simpler Public API's.

PR #34994, opened 2023-03-17
[AzureMonitorExporter] misc cleanup

TimothyMothra:
### Changes
- cleanup AAD example in Demo project
- add AAD instructions to Readme
- remove unnecessary `NoWarn` from csproj

### DIFF
--- a/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/README.md
+++ b/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/README.md
@@ -62,7 +62,17 @@ The following examples demonstrate how to add the `AzureMonitorExporter` to your
 
 ### Authenticate the client
 
-Exporter does not use authentication. 
+Azure Active Directory (AAD) authentication is an optional feature that can be used with the Azure Monitor Exporter.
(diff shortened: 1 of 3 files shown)


sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/README.md
```diff
@@ -62,7 +62,17 @@ The following examples demonstrate how to add the `AzureMonitorExporter` to your
 
 ### Authenticate the client
 
-Exporter does not use authentication. 
+Azure Active Directory (AAD) authentication is an optional feature that can be used with the Azure Monitor Exporter.
```
vishweshbankwar, line 65: We added this on the options itself.
https://github.com/Azure/azure-sdk-for-net/pull/34555. Need to decide on which one would be preferred way. IMO we should only have one way (may required breaking change).

vishweshbankwar: @rajkumar-rangaraj

TimothyMothra: good catch, I missed this change.
lets use this PR to fix the API.

We're going to need to mark the current method as `Obsolete` to direct users to the new way.

TimothyMothra: I opened https://github.com/Azure/azure-sdk-for-net/pull/35011 to address this.
Had an offline discussion and we're not going to take the change.

rajkumar-rangaraj (approved):

vishweshbankwar (approved):

PR #35001, opened 2023-03-20
Add function to get latest GA tag of a project

pshao25:
This is .net part of https://github.com/Azure/azure-sdk-tools/pull/5587. It gives the latest GA tag of a project.


eng/scripts/Language-Settings.ps1
```diff
@@ -550,3 +550,15 @@ function Get-dotnet-EmitterName() {
 function Get-dotnet-EmitterAdditionalOptions([string]$projectDirectory) {
   return "--option @azure-tools/typespec-csharp.emitter-output-dir=$projectDirectory/src"
 }
+
+function Get-dotnet-LatestTag([string]$projectDirectory) {
```
weshaggard, line 554: ApiCompatVersion will not work for preview packages as it only contains GA versions. 

Why are you parsing the project for this information? Why not query the repo tags?

weshaggard:
@pshao25 what do you need the latest tag?

eng/scripts/Language-Settings.ps1
pshao25: Yes. We only want the GA version.

The whole story is at https://github.com/Azure/autorest.csharp/issues/2887#issuecomment-1477330404. To fulfill it, we have three parts:
1. We should have a common script `Download-Latest-Contract.ps1` to download latest GA code by tag. Workflow and PR at https://github.com/Azure/azure-sdk-tools/pull/5587
2. In the above workflow, we .net choose to get the tag from csproj, which is implemented in this PR.
3. .net generator should have the ability to call Download-Latest-Contract.ps1: https://github.com/Azure/autorest.csharp/pull/2934/files#diff-d19f87d37f23d2330e0880bc308af772252c3bfcf8ef143698555ab31db24728R214-R268

PR #35011, opened 2023-03-20
[AzureMonitorExporter] Remove TokenCredential from extension methods. Use Options instead.

TimothyMothra:
`TokenCredential` was added to the `AzureMonitorExporterOptions` class in #34555.
This PR removes this from the extension methods in favor of using the options class.
We don't want multiple ways to set this property.

## Changes
- remove `TokenCredential` from the extension methods `AddAzureMonitorTraceExporter()`, `AddAzureMonitorLogExporter()`, and `AddAzureMonitorMetricExporter()`.
- refactor


rajkumar-rangaraj (changes requested):
All other Azure SDK provides both options to enable AAD through property and method parameter. Did we receive feedback to change this feature?

vishweshbankwar:
> All other Azure SDK provides both options to enable AAD through property and method parameter. Did we receive feedback to change this feature?

@rajkumar-rangaraj Do we know which one takes precedence?

rajkumar-rangaraj:
> > All other Azure SDK provides both options to enable AAD through property and method parameter. Did we receive feedback to change this feature?
> 
> @rajkumar-rangaraj Do we know which one takes precedence?

Yes. Option takes precedence over method parameter.

PR #35017, opened 2023-03-20
Add transmit from storage handler

vishweshbankwar:
Adding `TransmitFromStorageHandler`. Handler will periodically check for any files in storage and try transmitting it if transmission state is Closed. With this change the current implementation is still not impacted. Will do a follow up to integrate this and `TransmissionStateManager` with `AzureMonitorTransmitter`

### DIFF
--- a/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/AzureMonitorTransmitter.cs
+++ b/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/AzureMonitorTransmitter.cs
@@ -24,7 +24,7 @@ namespace Azure.Monitor.OpenTelemetry.Exporter.Internals
     /// </summary>
     internal class AzureMonitorTransmitter : ITransmitter
     {
-        private readonly ApplicationInsightsRestClient _applicationInsightsRestClient;
+        internal readonly ApplicationInsightsRestClient _applicationInsightsRestClient;
         internal PersistentBlobProvider? _fileBlobProvider;
         private readonly AzureMonitorStatsbeat? _statsbeat;
         private readonly ConnectionVars _connectionVars;
--- a/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/HttpPipelineHelper.cs
+++ b/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/HttpPipelineHelper.cs
@@ -9,6 +9,10 @@ using System.Threading;
 
 using Azure.Core;
 using Azure.Monitor.OpenTelemetry.Exporter.Models;
+using OpenTelemetry;
+using OpenTelemetry.Extensions.PersistentStorage.Abstractions;
+using OpenTelemetry.Extensions.PersistentStorage;
+using Azure.Monitor.OpenTelemetry.Exporter.Internals.PersistentStorage;
 
 namespace Azure.Monitor.OpenTelemetry.Exporter.Internals
 {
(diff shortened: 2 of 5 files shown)


sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/HttpPipelineHelper.cs
```diff
@@ -126,5 +129,63 @@ internal static bool TryGetRetryIntervalTimespan(Response? httpResponse, out Tim
...
+                    case ResponseStatusCodes.PartialSuccess:
+                        // Parse retry-after header
+                        // Send Failed Messages To Storage
+                        // Delete existing file
+                        TrackResponse trackResponse = GetTrackResponse(httpMessage);
+                        var content = GetPartialContentForRetry(trackResponse, httpMessage.Request.Content);
+                        if (content != null)
+                        {
+                            blob.TryDelete();
+                            blobProvider?.SaveTelemetry(content);
```
TimothyMothra, line 162: blobProvider shouldn't be null here because it's not nullable in the method signature.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TransmitFromStorageHandler.cs
```diff
@@ -0,0 +1,88 @@
...
+                    if (result == ExportResult.Success)
+                    {
+                        _transmissionStateManager.ResetConsecutiveErrors();
+                        _transmissionStateManager.CloseTransmission();
+
+                        AzureMonitorExporterEventSource.Log.WriteInformational("TransmitFromStorageSuccess", "Successfully transmitted a blob from storage.");
+
+                        // In case if the delete fails, there is a possibility
+                        // that the current batch will be transmitted more than once resulting in duplicates.
+                        blob.TryDelete();
```
TimothyMothra, line 52: Is there any way to know why the delete failed?
Should we try once more before giving up?

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/OfflineStorageTests.cs
```diff
@@ -169,7 +169,108 @@ public void TransmitFromStorage()
...
+
+            // Assert
+            // Blob will not be deleted as the transmission state is open.
+            Assert.Equal(TransmissionState.Open, transmissionStateManager.State);
+            Assert.Single(transmitter._fileBlobProvider.GetBlobs());
+
+            transmitter.Dispose();
+        }
+
+        private static AzureMonitorTransmitter GetTransmitter(MockResponse[] mockResponse)
```
TimothyMothra, line 273: this is a good change 👍

TimothyMothra (approved):
LGTM! left some minor comments

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TransmitFromStorageHandler.cs
vishweshbankwar: There is a possibility that our lease expired before we could get to delete part. In this case the delete will fail (underlying file name will change after lease expiration). This needs to be solved from the persistent storage package.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/AzureMonitorTransmitter.cs
```diff
@@ -24,7 +24,7 @@ namespace Azure.Monitor.OpenTelemetry.Exporter.Internals
     /// </summary>
     internal class AzureMonitorTransmitter : ITransmitter
     {
-        private readonly ApplicationInsightsRestClient _applicationInsightsRestClient;
+        internal readonly ApplicationInsightsRestClient _applicationInsightsRestClient;
```
rajkumar-rangaraj, line 27: Expose this as internal get property, it's not a best practice to expose a field variable.

rajkumar-rangaraj: I see it is used only in tests, its fine to leave it this way.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TransmitFromStorageHandler.cs
```diff
@@ -0,0 +1,88 @@
...
+
+        internal TransmitFromStorageHandler(ApplicationInsightsRestClient applicationInsightsRestClient, PersistentBlobProvider blobProvider, TransmissionStateManager transmissionStateManager)
+        {
+            _applicationInsightsRestClient = applicationInsightsRestClient;
+            _blobProvider = blobProvider;
+            _transmissionStateManager = transmissionStateManager;
+            _transmitFromStorageTimer = new System.Timers.Timer();
+            _transmitFromStorageTimer.Elapsed += TransmitFromStorage;
+            _transmitFromStorageTimer.AutoReset = true;
+            _transmitFromStorageTimer.Interval = 120000;
```
rajkumar-rangaraj, line 28: Is there a reason to select this interval?

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TransmitFromStorageHandler.cs
```diff
@@ -0,0 +1,88 @@
...
+            _transmitFromStorageTimer.Start();
+        }
+
+        internal void TransmitFromStorage(object? sender, ElapsedEventArgs? e)
+        {
+            while (_transmissionStateManager.State == TransmissionState.Closed && _blobProvider.TryGetBlob(out var blob) && blob.TryLease(1000))
```
rajkumar-rangaraj, line 34: This will cause high CPU issue, if there are more files.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TransmitFromStorageHandler.cs
vishweshbankwar: This is arbitrary. Open for any better suggestions.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TransmitFromStorageHandler.cs
rajkumar-rangaraj: A delay after specific transmission could also help.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/HttpPipelineHelper.cs
```diff
@@ -126,5 +129,63 @@ internal static bool TryGetRetryIntervalTimespan(Response? httpResponse, out Tim
...
+            bool shouldRetry = true;
+
+            if (httpMessage.HasResponse)
+            {
+                statusCode = httpMessage.Response.Status;
+                switch (statusCode)
+                {
+                    case ResponseStatusCodes.PartialSuccess:
+                        // Parse retry-after header
+                        // Send Failed Messages To Storage
```
rajkumar-rangaraj, line 155: This will create a new timestamp on the file name. Instead should be just release the lease on the blob?

vishweshbankwar: This is a special case. A new blob will be generated due to partial transmission. (see line 159-163)

rajkumar-rangaraj (approved):

PR #35020, opened 2023-03-20
[DynamicJson] Change delimiter used by changelist

annelo-msft:
Before this PR, we used a `.` character as a delimiter, however it is a valid character to use in a JSON property name, and as a result, if properties were present that had dots in the name, [it could break serialization](https://github.com/Azure/azure-sdk-for-net/compare/main...annelo-msft:azure-sdk-for-net:dynamicdata-delimiter?expand=1#diff-404b45ffe3711cbf1d5b4229cb6c300ffc0c1088e55ace215b8c2befd902e0f6R121).

By moving to an unescaped unicode control character as a delimiter, we're now keying changes on property names that are [invalid JSON strings](https://www.rfc-editor.org/rfc/rfc8259#section-7) and therefore invalid JSON.  The benefit is that the [BCL JsonReader will validate the JSON for us](https://github.com/Azure/azure-sdk-for-net/compare/main...annelo-msft:azure-sdk-for-net:dynamicdata-delimiter?expand=1#diff-404b45ffe3711cbf1d5b4229cb6c300ffc0c1088e55ace215b8c2befd902e0f6R155) when we call JsonDocument.Parse() to ensure that we can't be working with JSON strings where our delimiter could appear.

Many thanks to @bterlson for the Unicode/JSON consultation!

Contributes to https://github.com/Azure/azure-sdk-for-net/issues/33856

### DIFF
--- a/sdk/core/Azure.Core.Experimental/src/MutableJsonDocument.ChangeTracker.cs
+++ b/sdk/core/Azure.Core.Experimental/src/MutableJsonDocument.ChangeTracker.cs
@@ -13,7 +13,7 @@ namespace Azure.Core.Json
         {
             private List<MutableJsonChange>? _changes;
 
-            //private const char Delimiter = (char)1;
+            internal const char Delimiter = (char)1;
 
             internal bool HasChanges => _changes != null && _changes.Count > 0;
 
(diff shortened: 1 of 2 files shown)


sdk/core/Azure.Core.Experimental/src/MutableJsonDocument.ChangeTracker.cs
```diff
@@ -55,39 +57,6 @@ internal bool DescendantChanged(string path, int highWaterMark)
                 return changed;
             }
 
-            internal bool TryGetChange(ReadOnlySpan<byte> path, out MutableJsonChange change)
```
annelo-msft, line 58: Removing this to tidy up; will reinstate it when it can be implemented [using a preferred approach](https://github.com/annelo-msft/azure-sdk-for-net/blob/core-json-utf8span/sdk/core/Azure.Core.Experimental/src/MutableJsonDocument.ChangeTracker.cs#L61).

sdk/core/Azure.Core.Experimental/tests/MutableJsonDocumentTests.cs
```diff
@@ -117,6 +117,64 @@ public void CanSetPropertyMultipleTimes()
                 jsonString);
         }
 
+        [Test]
+        public void CanSetProperty_JsonWithDotProperty()
```
annelo-msft, line 121: This is implemented for TDD; it failed before and is green with this change.

sdk/core/Azure.Core.Experimental/tests/MutableJsonDocumentTests.cs
```diff
@@ -117,6 +117,64 @@ public void CanSetPropertyMultipleTimes()
...
+                  },
+                  "Foo" : 3,
+                  "Bar" : "Hi!"
+                }
+                """),
+                jsonString);
+        }
+
+        [Test]
+        public void JsonWithDelimiterIsInvalidJson()
```
annelo-msft, line 155: This is here as protection that the delimiter isn't accidentally changed to something that could break serialization.

jsquire (approved):

PR #35021, opened 2023-03-20
[AzureMonitorExporter] update offline storage subdirectory

TimothyMothra:
## Background
In the event telemetry cannot successfully be ingested, our Exporter will save telemetry on the local disk to retry later.
We will support multiple Exporters in the same process and each Exporter can have unique connection strings (ikey + endpoint).
When telemetry is later retried, it must be sent to the correct endpoint.
Therefore, we need to use the ikey in the storage subdirectory to ensure we can match it to the correct endpoint.

Note that if a customer changes their ikey for any reason (rotating ikey or new resource), any persisted telemetry will be lost.
This is because the ikey is stored as a member of the telemetry and the Exporter is not able to map an old ikey to a new ikey.

## Goal
- store telemetry in a unique subdirectory per ikey.
- directory name must not exceed platform max lengths.

## Changes
- Exporter will create a subdirectory for locally stored telemetry.
  Subdirectory is a SHA256 hash composed of InstrumentationKey + UserName + ProcessName + ApplicationDirectory.
- Update platform abstractions to include UserName, ProcessName, ApplicationDirectory .

## Blockers
~This PR is on hold until changes in the PersistentStorage SDK are released: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1110~


sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/AzureMonitorTransmitter.cs
```diff
@@ -92,27 +92,20 @@ private static ApplicationInsightsRestClient InitializeRestClient(AzureMonitorEx
             return new ApplicationInsightsRestClient(new ClientDiagnostics(options), pipeline, host: connectionVars.IngestionEndpoint);
         }
 
-        private static PersistentBlobProvider? InitializeOfflineStorage(AzureMonitorExporterOptions options)
+        private static PersistentBlobProvider? InitializeOfflineStorage(AzureMonitorExporterOptions options, ConnectionVars connectionVars)
```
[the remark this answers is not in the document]
TimothyMothra: I would vote to leave it as is for now. Currently all the internal Initialize methods provide the full ConnectionVars object.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/PersistentStorage/StorageHelper.cs
```diff
@@ -12,6 +12,15 @@ internal static class StorageHelper
     {
         private static string? s_defaultStorageDirectory;
 
+        internal static string GetStorageDirectory(string? configuredStorageDirectory, string instrumentationKey)
+        {
+            var storageDirectory = configuredStorageDirectory
+                ?? GetDefaultStorageDirectory()
+                ?? throw new InvalidOperationException("Unable to determine offline storage directory.");
+
+            return Path.Combine(storageDirectory, instrumentationKey);
```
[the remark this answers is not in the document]
TimothyMothra: I spent some time reviewing this topic.

Guid is 36 characters. 
I haven't found a hash that will significantly reduce the string length (SHA256 is 64 characters, MD5 is 32 characters).
Best I found was using a Base64String which would reduce it to 22 characters.

We could take a substring of the ikey, but then we can't guarantee no collisions. 

--
OpenTelemetry.Extensions.PersistentStorage is creating yet another subdirectory and uses SHA256. ([link](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/b06a6f290fc59a292994e2ba9973eefe0691fa19/src/OpenTelemetry.PersistentStorage.FileSystem/PersistentStorageHelper.cs#L205))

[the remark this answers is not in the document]
TimothyMothra: For context, here's an example full path (206 characters):

```
"C:\TEMP\00000000-0000-0000-0000-000000000000\a58c82c531c9ee2e54763f35b4c81c1f9b5ebbef091a556cbbf254151475cedd\2023-03-20T222007.2578717Z-0985894a111a41cb8462d7b10bab7cc0.blob@2023-03-20T222107.2645312Z.lock"
```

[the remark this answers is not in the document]
rajkumar-rangaraj: It will be more than 215 characters in most cases.

```
C:\Users\username\AppData\Local\Microsoft\AzureMonitor\3a70e0d7934b71314c58dc3302cb965d0e34883cde72da6d9c279977e8790f1a\2023-03-20T222007.2578717Z-0985894a111a41cb8462d7b10bab7cc0.blob@2023-03-20T222107.2645312Z.lock
```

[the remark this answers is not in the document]
rajkumar-rangaraj: What is the length of the path increased by this change?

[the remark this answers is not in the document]
TimothyMothra: We can discuss this in our next standup.
In the image below, the gray rows are WITH Ikey.
In all of these examples, we are within the max allowed length.

image


However, to have an extra safety net I think we should still reduce the length.
The best I've found is to convert the Guid (36 chars) to a Base64String (22 chars).

```csharp
Convert.ToBase64String(Guid.Parse(inputString).ToByteArray())
```

reyang (dismissed):
Need to understand the design:
1. What if the user rotate ikeys?
2. Would it be considered anti-pattern (mostly from security perspective) to put ikey in the path?

TimothyMothra:
Good questions!

> 1. What if the user rotate ikeys?

Then telemetry saved to disk will be lost.
Regardless of where we store the telemetry, the ikey is a member of the telemetry item.
SDK won't be able to map old telemetry containing old ikey to new ikey.
(Same behavior as Application Insights SDK).

> 2. Would it be considered anti-pattern (mostly from security perspective) to put ikey in the path?

This subject has come up in the past and the consensus has been that the ikey is not a secret.
Also, the ikey is already available in plain-text within this directory, it's a member of the telemetry items.
We do disclose that telemetry in local storage is not encrypted. ([link](https://learn.microsoft.com/azure/azure-monitor/app/data-retention-privacy#does-the-sdk-create-temporary-local-storage))

I am exploring options to obfuscate the ikey out of a necessity to shorten the file path.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/PersistentStorage/StorageHelper.cs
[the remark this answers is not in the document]
TimothyMothra: Discussed this in our standup today.
We're going to drive some changes in the PersistentStorage library.
The idea is that we could provide the ikey value to be included in PersistentStorage's hash.

This PR will be on hold while we pursue those changes.

reyang:
> Good questions!
> 
> > 1. What if the user rotate ikeys?
> 
> Then telemetry saved to disk will be lost. Regardless of where we store the telemetry, the ikey is a member of the telemetry item. SDK won't be able to map old telemetry containing old ikey to new ikey. (Same behavior as Application Insights SDK).

Is this the goal (I don't see the goal in the PR description)?

> > 2. Would it be considered anti-pattern (mostly from security perspective) to put ikey in the path?
> 
> This subject has come up in the past and the consensus has been that the ikey is not a secret. Also, the ikey is already available in plain-text within this directory, it's a member of the telemetry items. We do disclose that telemetry in local storage is not encrypted. ([link](https://learn.microsoft.com/azure/azure-monitor/app/data-retention-privacy#does-the-sdk-create-temporary-local-storage))
> 
> I am exploring options to obfuscate the ikey out of a necessity to shorten the file path.

Could you add comment and explain the consensus or link to where the consensus is captured?

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/PersistentStorage/StorageHelper.cs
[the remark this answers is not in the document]
TimothyMothra: The PersistentStorage library has been changed to omit the extra subdirectory.
https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1110

I'll change this PR to handle it's own sub directories.

TimothyMothra:
> > Good questions!
> > > 1. What if the user rotate ikeys?
> > 
> > 
> > Then telemetry saved to disk will be lost. Regardless of where we store the telemetry, the ikey is a member of the telemetry item. SDK won't be able to map old telemetry containing old ikey to new ikey. (Same behavior as Application Insights SDK).
> 
> Is this the goal (I don't see the goal in the PR description)?

@reyang, I've updated the PR description. Please re-review.

> > > 2. Would it be considered anti-pattern (mostly from security perspective) to put ikey in the path?
> > 
> > 
> > This subject has come up in the past and the consensus has been that the ikey is not a secret. Also, the ikey is already available in plain-text within this directory, it's a member of the telemetry items. We do disclose that telemetry in local storage is not encrypted. ([link](https://learn.microsoft.com/azure/azure-monitor/app/data-retention-privacy#does-the-sdk-create-temporary-local-storage))
> > I am exploring options to obfuscate the ikey out of a necessity to shorten the file path.
> 
> Could you add comment and explain the consensus or link to where the consensus is captured?

It was decided to take a new direction. 
We made some changes to the PersistentStorage library to give an Exporter more control over the subdirectory.
The ikey is hashed (along with other values about the process) to create a unique subdirectory.
This obfuscates the ikey AND keeps a reasonable directory length.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/PersistentStorage/StorageHelper.cs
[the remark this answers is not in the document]
TimothyMothra: New PersistentStorage library is available and I've updated this PR.

rajkumar-rangaraj:
Is there a reason for omitting  `Environment.User`?

TimothyMothra:
> Is there a reason for omitting `Environment.User`?

There wasn't a conscious decision to omit it, it just wasn't scoped. 
Do you think it needs to be included?

rajkumar-rangaraj:
> > Is there a reason for omitting `Environment.User`?
> 
> There wasn't a conscious decision to omit it, it just wasn't scoped. Do you think it needs to be included?

Yes, so the scope the stored telemetry is tied to an user.

TimothyMothra:
> > > Is there a reason for omitting `Environment.User`?
> > 
> > 
> > There wasn't a conscious decision to omit it, it just wasn't scoped. Do you think it needs to be included?
> 
> Yes, so the scope the stored telemetry is tied to an user.

updated, please review

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/StorageHelperTests.cs
```diff
@@ -15,13 +15,41 @@ public class StorageHelperTests
...
             var directoryPath = StorageHelper.GetStorageDirectory(
-                platform: new MockPlatform(),
-                configuredStorageDirectory: $"C:{ds}Temp");
+                platform: new MockPlatform
+                {
+                    UserName = "ApplicationPoolIdentity",
+                    ProcessName ="w3wp",
+                    ApplicationBaseDirectory = "C:\\inetpub\\wwwroot",
+                },
+                configuredStorageDirectory: $"C:{ds}Temp",
```
reyang, line 27: Is this a common practice?

TimothyMothra: No, this was an ugly hack to allow these tests to run on both Windows and Linux.
Internally, `Path.Combine` will use either `\` or `/` depending on the OS running the unit tests.

I can investigate some alternatives.

TimothyMothra: Updated.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/StorageHelperTests.cs
```diff
@@ -11,17 +12,53 @@ namespace Azure.Monitor.OpenTelemetry.Exporter.Tests
...
         [Fact]
-        public void VerifyConfiguredDirectory()
+        public void VerifyGetStorageDirectory_Configured()
+        {
+            var directoryPath = StorageHelper.GetStorageDirectory(
+                platform: new MockPlatform
+                {
+                    UserName = "ApplicationPoolIdentity",
+                    ProcessName ="w3wp",
+                    ApplicationBaseDirectory = "C:\\inetpub\\wwwroot",
```
reyang, line 32: maybe use https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/verbatim

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/StorageHelperTests.cs
```diff
@@ -11,17 +12,53 @@ namespace Azure.Monitor.OpenTelemetry.Exporter.Tests
...
     {
-        // The directory separator is either '/' or '\' depending on the Platform, making unit testing tricky.
-        private static readonly char ds = Path.DirectorySeparatorChar;
+        private readonly string _testStorageDirectory;
+
+        public StorageHelperTests()
+        {
+            _testStorageDirectory = RuntimeInformation.IsOSPlatform(OSPlatform.Windows)
+                ? "C:\\UnitTest"
+                : "/var/UnitTest";
```
reyang, line 21: Who would be responsible for the clean up?

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/PersistentStorage/StorageHelper.cs
```diff
@@ -43,12 +44,11 @@ internal static string GetStorageDirectory(IPlatform platform, string? configure
...
                     || TryCreateTelemetryDirectory(platform: platform, path: "/tmp/", createdDirectoryPath: out dirPath))
                 {
-                    s_defaultStorageDirectory = dirPath;
-                    return s_defaultStorageDirectory;
+                    return dirPath;
                 }
             }
 
-            return s_defaultStorageDirectory;
+            throw new InvalidOperationException("Unable to determine default storage directory.");
```
reyang, line 51: Why is this InvalidOperation?

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/StorageHelperTests.cs
TimothyMothra: No assumptions (that drive exists) and no cleanup.
In this test class, the platform has been mocked and fully testing the business logic.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/PersistentStorage/StorageHelper.cs
TimothyMothra: InvalidOperation is the default exception that's being used throughout most of the SDK.
We have a work item to fully review all exceptions and logging before GA.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/StorageHelperTests.cs
TimothyMothra: done

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/AzureMonitorTransmitter.cs
```diff
@@ -100,17 +100,18 @@ private static ApplicationInsightsRestClient InitializeRestClient(AzureMonitorEx
             return new ApplicationInsightsRestClient(new ClientDiagnostics(options), pipeline, host: connectionVars.IngestionEndpoint);
         }
 
-        private static PersistentBlobProvider? InitializeOfflineStorage(AzureMonitorExporterOptions options, IPlatform platform)
+        private static PersistentBlobProvider? InitializeOfflineStorage(AzureMonitorExporterOptions options, IPlatform platform, string instrumentationKey)
```
rajkumar-rangaraj, line 103: This class has a public property `InstrumentationKey`, should we use the property instead of passing the parameter?

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/PersistentStorage/StorageHelper.cs
```diff
@@ -11,16 +11,18 @@ namespace Azure.Monitor.OpenTelemetry.Exporter.Internals.PersistentStorage
...
-            var rootDirectory = configuredStorageDirectory
-                ?? (s_defaultStorageDirectory ??= GetDefaultStorageDirectory(platform))
-                ?? throw new InvalidOperationException("Unable to determine offline storage directory.");
+            var rootDirectory = configuredStorageDirectory ?? GetDefaultStorageDirectory(platform);
+
+            // get unique sub directory
+            var userName = platform.GetEnvironmentUserName();
+            var processName = platform.GetCurrentProcessName();
+            var applicationDirectory = platform.GetApplicationBaseDirectory();
+            string subDirectory = HashHelper.GetSHA256Hash($"{instrumentationKey};{userName};{processName};{applicationDirectory}");
```
rajkumar-rangaraj, line 23: Do we need a separator `;` here?

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/PersistentStorage/StorageHelper.cs
rajkumar-rangaraj: Why are we changing the behavior of this method to throw?

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/StorageHelperTests.cs
```diff
@@ -11,17 +12,53 @@ namespace Azure.Monitor.OpenTelemetry.Exporter.Tests
...
+        [Theory]
+        [InlineData("WINDOWS", "LOCALAPPDATA")]
+        [InlineData("LINUX", "TMPDIR")]
+        public void VerifyGetStorageDirectory_Default(string osName, string envVarName)
         {
+            var platform = new MockPlatform
+            {
+                UserName = "ApplicationPoolIdentity",
+                ProcessName = "w3wp",
+                ApplicationBaseDirectory = "C:\\inetpub\\wwwroot",
```
rajkumar-rangaraj, line 50: Little difficult to follow here. I understand its a mock, but seeing a c:\ for Linux tests confuses.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/PersistentStorage/StorageHelper.cs
TimothyMothra: we need a delimiter between individual values to help avoid hash collisions.
I choose a semi-colon `;` here. The PersistentStorage library was using an at symbol `@`. 


As a conversational example consider `applejuice;box` vs `apple;juicebox`. These would have the same hash without a delimiter.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/PersistentStorage/StorageHelper.cs
TimothyMothra: > Why are we changing the behavior of this method to throw?

This method seemed the more appropriate location to throw. 
Previously, this method returned a null, and the calling method threw.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/StorageHelperTests.cs
TimothyMothra: This started as the Windows example, and all three values would be familiar to someone with IIS experience.
It's not immediately clear from reading the test, but these values are only used to calculate the hash used in the final storage directory.

Any changes here for a Linux based example would also require a new hash value. My gut feeling was that would over complicate the test.

But I can make that change to make the test easier to read and understand.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/AzureMonitorTransmitter.cs
TimothyMothra: That would require changing the method from `static`. 
I would prefer these Initialize methods remain static to enable better unit testing.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/tests/Azure.Monitor.OpenTelemetry.Exporter.Tests/CommonTestFramework/MockPlatform.cs
```diff
@@ -16,6 +16,10 @@ internal class MockPlatform : IPlatform
         public string OSPlatformName { get; set; } = "UnitTest";
         public Func<OSPlatform, bool> IsOsPlatformFunc { get; set; } = (OSPlatform) => false;
         public Func<string, bool> CreateDirectoryFunc { get; set; } = (path) => true;
+        public string UserName { get; set; } = "UnitTestUser";
+        public string ProcessName { get; set; } = "UnitTestProcess";
+        public string ApplicationBaseDirectory { get; set; } = $"UnitTestDirectory";
```
vishweshbankwar, line 21:
```suggestion
        public string ApplicationBaseDirectory { get; set; } = "UnitTestDirectory";
```

TimothyMothra: Good catch, thank you!

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/AzureMonitorTransmitter.cs
rajkumar-rangaraj: It does not make a difference as this is a private method and the initialization is called from constructor. In my opinion we should not create duplicates to help the tests, if that is the case the method or the class in not well-written.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/PersistentStorage/StorageHelper.cs
rajkumar-rangaraj: My recommendation is to restore the previous behavior. `InvalidOperationException` is not a better fit here and we don't need to throw if storage directory is null, rather we could log to event source.

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/AzureMonitorTransmitter.cs
TimothyMothra: These Initialize methods are critically important because they're responsible for building all of the internals in our SDK and I want these to have very high test coverage.
Of the 4 Initialize methods, the only one that has 100% coverage is the one that's `internal static` and in my opinion, all 4 should move in this direction.
If all of these methods are instance based, then we can't test one without also testing all the others.

TimothyMothra: Discussed offline. 
Will keep `static` to align with testing goals.
Will modify method parameters to eliminate duplication.

rajkumar-rangaraj (approved):

PR #35038, opened 2023-03-21
Bump version of policy

### LINKED ISSUE #34986: Upgrade the `Azure.ResourceManager` lib to support new PolicyAssignments
Upgrade to latest `2022-06-01` version

ArcturusZhang:
Fixes https://github.com/Azure/azure-sdk-for-net/issues/34986


sdk/resourcemanager/Azure.ResourceManager/api/Azure.ResourceManager.netstandard2.0.cs
```diff
@@ -770,7 +770,9 @@ public partial class PolicyAssignmentResource : Azure.ResourceManager.ArmResourc
         public virtual System.Threading.Tasks.Task<Azure.ResourceManager.ArmOperation<Azure.ResourceManager.Resources.PolicyAssignmentResource>> DeleteAsync(Azure.WaitUntil waitUntil, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
         public virtual Azure.Response<Azure.ResourceManager.Resources.PolicyAssignmentResource> Get(System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
         public virtual System.Threading.Tasks.Task<Azure.Response<Azure.ResourceManager.Resources.PolicyAssignmentResource>> GetAsync(System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
+        public virtual Azure.Response<Azure.ResourceManager.Resources.PolicyAssignmentResource> Update(Azure.ResourceManager.Resources.Models.PolicyAssignmentPatch patch, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
```
archerzz, line 773: Just a confirmation. Is it expected that we have 2 sets of `Update` methods (one is for `PolicyAssignmentData`, another is for `PolicyAssignmentPatch`)? Looks like the former one should be `CreateOrUpdate`.

archerzz (dismissed):
Need to re-record the tests to resolve `api-version` mis-match.

sdk/resourcemanager/Azure.ResourceManager/api/Azure.ResourceManager.netstandard2.0.cs
ArcturusZhang: Generator creates a `Update` from its put request if the resource does not have a patch.
Now since the swagger adds a patch for this resource, the put version is removed.
They serve different purpose, put is overriding and patch is not. I believe we do not need to hide the other or if we decide to hide it, we could do it later

ArthurMa1978 (approved):

PR #35046, opened 2023-03-21
[Azure.Monitor.Query]Update LogsQueryClientBuilderExtensions.cs

nisha-bhatia:

### DIFF
--- a/sdk/monitor/Azure.Monitor.Query/src/LogsQueryClientBuilderExtensions.cs
+++ b/sdk/monitor/Azure.Monitor.Query/src/LogsQueryClientBuilderExtensions.cs
@@ -12,6 +12,14 @@ namespace Microsoft.Extensions.Azure
     {
         /// <summary> Registers a <see cref="LogsQueryClient"/> instance. </summary>
         /// <param name="builder"> The builder to register with. </param>
+        public static IAzureClientBuilder<LogsQueryClient, LogsQueryClientOptions> AddLogsQueryClient<TBuilder>(this TBuilder builder)
+        where TBuilder : IAzureClientFactoryBuilderWithCredential
+        {
+            return builder.RegisterClientFactory<LogsQueryClient, LogsQueryClientOptions>((options, cred) => new LogsQueryClient(cred, options));
+        }
+
+        /// <summary> Registers a <see cref="LogsQueryClient"/> instance. </summary>
+        /// <param name="builder"> The builder to register with. </param>
         /// <param name="endpoint"> The service endpoint to use. Default is 'https://api.loganalytics.io'. </param>
         public static IAzureClientBuilder<LogsQueryClient, LogsQueryClientOptions> AddLogsQueryClient<TBuilder>(this TBuilder builder, Uri endpoint)
         where TBuilder : IAzureClientFactoryBuilderWithCredential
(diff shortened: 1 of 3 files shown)


sdk/monitor/Azure.Monitor.Query/src/LogsQueryClientBuilderExtensions.cs
```diff
@@ -12,7 +12,7 @@ public static class LogsQueryClientBuilderExtensions
     {
         /// <summary> Registers a <see cref="LogsQueryClient"/> instance. </summary>
         /// <param name="builder"> The builder to register with. </param>
-        /// <param name="endpoint"> The resource manager service endpoint to use. For example <c>https://management.azure.com/</c> for public cloud. </param>
+        /// <param name="endpoint"> The service endpoint to use. Default is 'https://api.loganalytics.io'. </param>
```
JoshLove-msft, line 15: This isn't really the default in this context because this is a required parameter, this raises the question - should this be optional?

nisha-bhatia: I think the endpoint can vary based on what cloud you are using. For example I think government clouds have a different endpoint

JoshLove-msft: But we expose a constructor in the client that doesn't take an endpoint, so should we do the same here?

sdk/monitor/Azure.Monitor.Query/src/LogsQueryClientBuilderExtensions.cs
```diff
@@ -12,7 +12,15 @@ public static class LogsQueryClientBuilderExtensions
...
-        /// <param name="endpoint"> The resource manager service endpoint to use. For example <c>https://management.azure.com/</c> for public cloud. </param>
+        public static IAzureClientBuilder<LogsQueryClient, LogsQueryClientOptions> AddLogsQueryClient<TBuilder>(this TBuilder builder)
+        where TBuilder : IAzureClientFactoryBuilderWithCredential
+        {
+            return builder.RegisterClientFactory<LogsQueryClient, LogsQueryClientOptions>((options, cred) => new LogsQueryClient(cred, options));
+        }
+
+        /// <summary> Registers a <see cref="LogsQueryClient"/> instance. </summary>
+        /// <param name="builder"> The builder to register with. </param>
+        /// <param name="endpoint"> The service endpoint to use. Default is 'https://api.loganalytics.io'. </param>
```
JoshLove-msft, line 23: We should mention the default in the new method that is being added. For this one there is no need to mention the default, as the user is required to pass the endpoint.

sdk/monitor/Azure.Monitor.Query/src/MetricsQueryClientBuilderExtensions.cs
```diff
@@ -10,6 +10,14 @@ namespace Microsoft.Extensions.Azure
     /// <summary> Extension methods to add <see cref="MetricsQueryClient"/> to client builder. </summary>
     public static class MetricsQueryClientBuilderExtensions
     {
+        /// <summary> Registers a <see cref="MetricsQueryClient"/> instance. </summary>
+        /// <param name="builder"> The builder to register with. </param>
+        public static IAzureClientBuilder<MetricsQueryClient, MetricsQueryClientOptions> AddMetricsQueryClient<TBuilder>(this TBuilder builder)
```
JoshLove-msft, line 15: We should mention the default endpoint that is being used.

JoshLove-msft:
Will need to export the API.

JoshLove-msft (approved):

PR #35057, opened 2023-03-22
Fix flaky test

JoshLove-msft:


sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/tests/ServiceBusEndToEndTests.cs
```diff
@@ -1022,7 +1031,6 @@ public class ServiceBusTestJobs : ServiceBusTestJobsBase
                 Assert.AreEqual("value", userProperties["key"]);
                 Assert.Greater(expiresAtUtc, DateTime.UtcNow);
                 Assert.AreEqual(expiresAt.DateTime, expiresAtUtc);
-                Assert.Less(enqueuedTimeUtc, DateTime.UtcNow);
```
JoshLove-msft, line 1025: This was not always true probably due to clock skew.

jsquire: Event Hubs had the same issue.   We used added 5-minutes to "now" as a buffer and that seems to have made things happy.   No objections to just removing it, either.

jsquire (approved):

PR #35066, opened 2023-03-22
Added Microsoft.GraphServices .NET SDK

amatukmolina:


amatukmolina:
The contents of this PR were reviewed in the private repo https://github.com/Azure/azure-sdk-for-net-pr/pull/1831 by @ArcturusZhang

sdk/graphservices/Azure.ResourceManager.GraphServices/tests/GraphServicesManagementTestBase.cs
```diff
@@ -0,0 +1,41 @@
...
+        : base(isAsync, mode)
+        {
+        }
+
+        protected GraphServicesManagementTestBase(bool isAsync)
+            : base(isAsync)
+        {
+        }
+
+        public void CreateCommonClient()
```
ArcturusZhang, line 27: we could just add a [SetUp] attribute on this method, and use the property `Client` in the test cases.

sdk/graphservices/Azure.ResourceManager.GraphServices/tests/Tests/AccountCRUDTests.cs
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        [SetUp]
+        public void ClearAndInitialize()
+        {
+            if (Mode == RecordedTestMode.Record || Mode == RecordedTestMode.Playback)
+            {
+                CreateCommonClient();
+            }
+        }
```
ArcturusZhang, line 29: We could remove this if we add `[SetUp]` on the `CreateCommonClient` method

sdk/graphservices/Azure.ResourceManager.GraphServices/tests/Tests/AccountCRUDTests.cs
```diff
@@ -0,0 +1,77 @@
...
+        {
+            // Get Default Subscription
+            SubscriptionResource subscription = await this.Client.GetDefaultSubscriptionAsync();
+            ResourceGroupCollection resourceGroups = subscription.GetResourceGroups();
+
+            // Create Resource Group
+            string resourceGroupName = "myRgName";
+            ResourceGroupData resourceGroupData = new ResourceGroupData("westus");
+            ArmOperation<ResourceGroupResource> operation = await resourceGroups.CreateOrUpdateAsync(WaitUntil.Completed, resourceGroupName, resourceGroupData);
+            ResourceGroupResource resourceGroup = operation.Value;
```
ArcturusZhang, line 42: We could just call the `CreateResourceGroup` method on the base class here

sdk/graphservices/Azure.ResourceManager.GraphServices/tests/Tests/AccountCRUDTests.cs
```diff
@@ -0,0 +1,77 @@
...
+            var resourceName = Recording.GenerateAssetName("SDKAccounts");
+            string appId = new AccountTestEnvironment().ApplicationIDClient;
+            AccountResourceProperties accountResourceProperties = new AccountResourceProperties(appId);
+            AccountResourceData accountResourceData = new AccountResourceData("global", accountResourceProperties);
+
+            // Create Account Resource
+            var createAccountOperation = await collection.CreateOrUpdateAsync(WaitUntil.Completed, resourceName, accountResourceData);
+
+            Assert.IsTrue(createAccountOperation.HasCompleted);
+            Assert.IsTrue(createAccountOperation.HasValue);
```
ArcturusZhang, line 56: Could we assert some properties in the returned data?

sdk/graphservices/Azure.ResourceManager.GraphServices/tests/Tests/AccountCRUDTests.cs
```diff
@@ -0,0 +1,77 @@
...
+
+            // Get
+            Response<AccountResource> getAccountResponse = await collection.GetAsync(resourceName);
+            AccountResource accountResource = getAccountResponse.Value;
+            Assert.IsNotNull(accountResource);
```
ArcturusZhang, line 61: Shall we add an assertion to assert the `Id` is the same as the Id above?

sdk/graphservices/Azure.ResourceManager.GraphServices/tests/Tests/AccountCRUDTests.cs
```diff
@@ -0,0 +1,77 @@
...
+        [SetUp]
+        public void ClearAndInitialize()
+        {
+            if (Mode == RecordedTestMode.Record || Mode == RecordedTestMode.Playback)
+            {
+                CreateCommonClient();
+            }
+        }
+
+        [TestCase]
```
ArcturusZhang, line 31: Could we use `RecordedTest` instead?

sdk/graphservices/Azure.ResourceManager.GraphServices/tests/Tests/AccountCRUDTests.cs
```diff
@@ -0,0 +1,77 @@
...
+            var getAccountResponse_Update = await collection.GetAsync(resourceName);
+            AccountResource accountResource_Get_Update = getAccountResponse_Update.Value;
+            Assert.IsNotNull(accountResource_Get_Update);
+
+            // Delete
+            var accountOperation = await accountResource.DeleteAsync(WaitUntil.Completed);
+            await accountOperation.WaitForCompletionResponseAsync();
+            Assert.IsTrue(accountOperation.HasCompleted);
+
+            await resourceGroup.DeleteAsync(WaitUntil.Completed);
```
ArcturusZhang, line 74: We do not need this because our test base could automatically do this.

sdk/graphservices/Azure.ResourceManager.GraphServices/tests/Tests/AccountCRUDTests.cs
```diff
@@ -0,0 +1,77 @@
...
+
+            // Get
+            Response<AccountResource> getAccountResponse = await collection.GetAsync(resourceName);
+            AccountResource accountResource = getAccountResponse.Value;
+            Assert.IsNotNull(accountResource);
+
+            // Update
+            var updateResource = await accountResource.AddTagAsync("Test","Value");
+            var getAccountResponse_Update = await collection.GetAsync(resourceName);
+            AccountResource accountResource_Get_Update = getAccountResponse_Update.Value;
```
ArcturusZhang, line 66: These are not following the convention of dotnet, could we fix?

sdk/graphservices/Azure.ResourceManager.GraphServices/tests/autorest.tests.md
```diff
@@ -0,0 +1,13 @@
...
+
+# Azure.ResourceManager.GraphServices.Tests
+
+> see https://aka.ms/autorest
+``` yaml
+require: ../src/autorest.md
+include-x-ms-examples-original-file: true
+testgen:
+  sample: true
+```
```
ArcturusZhang, line 13: Could we remove this file?

sdk/graphservices/Azure.ResourceManager.GraphServices/src/Generated/AccountResource.cs
```diff
@@ -0,0 +1,602 @@
...
+
+namespace Azure.ResourceManager.GraphServices
+{
+    /// <summary>
+    /// A Class representing an AccountResource along with the instance operations that can be performed on it.
+    /// If you have a <see cref="ResourceIdentifier" /> you can construct an <see cref="AccountResource" />
+    /// from an instance of <see cref="ArmClient" /> using the GetAccountResource method.
+    /// Otherwise you can get one from its parent resource <see cref="ResourceGroupResource" /> using the GetAccountResource method.
+    /// </summary>
+    public partial class AccountResource : ArmResource
```
archerzz, line 28: The name is too generic. Consider appending RP prefix or renaming it.
- use `prepend-rp-prefix` to automatically append PR name. See example [here](https://github.com/Azure/azure-sdk-for-net/blob/176fd476588e8e61ebce747d66714cb39947cae5/sdk/azurestackhci/Azure.ResourceManager.Hci/src/autorest.md?plain=1#L52-L57).
- use `rename-mapping` to rename a model. See example [here](https://github.com/Azure/azure-sdk-for-net/blob/176fd476588e8e61ebce747d66714cb39947cae5/sdk/azurestackhci/Azure.ResourceManager.Hci/src/autorest.md?plain=1#L58-L76)

archerzz (commented):
Consider adding prefix to `AccountXXX` to avoid potential duplication.

sdk/graphservices/Azure.ResourceManager.GraphServices/tests/Tests/AccountCRUDTests.cs
```diff
@@ -0,0 +1,62 @@
...
+using Azure.Core.TestFramework;
+
+namespace Azure.ResourceManager.GraphServices.Tests.Tests
+{
+    [TestFixture]
+    public class AccountCRUDTests : GraphServicesManagementTestBase
+    {
+        public AccountCRUDTests() : base(true)
+        {
+        }
```
ArcturusZhang, line 18: we should not hard code the `isAsync` to true here.
The test framework always ensures us to have both version of the test cases, sync and async.
```suggestion
        public AccountCRUDTests(bool isAsync) : base(isAsync)
        {
        }
```
This might need a re-recording

ArcturusZhang: Also please ensure that every API we are calling in our test cases is the ASYNC version.
The test framework we automatically replace those async alls with their corresponding sync version when `isAsync` is false.

ArcturusZhang (commented):
We should be missing a `ci.mgmt.yaml` file in this PR? Something like this: https://github.com/Azure/azure-sdk-for-net/blob/76ff14813aeb4248856c1df9e7032c45d2234f90/sdk/compute/ci.mgmt.yml
Could you add a file like this in the `sdk/graphservices` directory so that we could have validation pipelines on your new SDK?

sdk/graphservices/Azure.ResourceManager.GraphServices/src/autorest.md
```diff
@@ -0,0 +1,57 @@
+# Generated code configuration
+
+Run `dotnet build /t:GenerateCode` to generate code.
+
+``` yaml
+azure-arm: true
+csharp: true
+library-name: GraphServices
+namespace: Azure.ResourceManager.GraphServices
+require: https://github.com/Azure/azure-rest-api-specs/blob/ee157d6bbf6e4c4fdc8a47b380e00ddd56a0e852/specification/graphservicesprod/resource-manager/readme.md
```
ArcturusZhang, line 10: Please also make sure this commit is on main branch

sdk/graphservices/Azure.ResourceManager.GraphServices/tests/Tests/AccountCRUDTests.cs
```diff
@@ -0,0 +1,61 @@
...
+            ResourceGroupCollection resourceGroups = subscription.GetResourceGroups();
+
+            // Create Resource Group
+            string resourceGroupName = "myRgName";
+            ResourceGroupResource resourceGroup = await this.CreateResourceGroup(subscription, resourceGroupName, Core.AzureLocation.WestUS);
+            GraphServicesAccountResourceCollection collection = resourceGroup.GetGraphServicesAccountResources();
+
+            // Set Account Resource Values
+            var resourceName = Recording.GenerateAssetName("SDKAccounts");
+            string appId = new AccountTestEnvironment().ApplicationIDClient;
```
ArcturusZhang, line 33: we are having some issues in the playback of this test case.
I am not very familiar with this feature, but here we cannot just create a new test environment. We should already have an instance in the test base because it inherits the generic version `RecordedTestBase<TestEnvironment>`, something like:
```
var appId = TestEnvironment.ApplicationIDClient;
```

sdk/graphservices/Azure.ResourceManager.GraphServices/tests/Helpers/AccountTestEnvironment.cs
```diff
@@ -0,0 +1,15 @@
...
+using Azure.Core.TestFramework;
+
+namespace Azure.ResourceManager.GraphServices.Tests.Helpers
+{
+    /// <summary>
+    /// A helper class used to retrieve information to be used for tests.
+    /// </summary>
+    public class AccountTestEnvironment: TestEnvironment
+    {
+        public string ApplicationIDClient => GetRecordedVariable("AZURE_CLIENT_ID");
```
ArcturusZhang, line 13: This reads weird, should we use `ApplicationClientId` instead?

ArcturusZhang (commented):
Please also run the command to update the API.
It is always recommended that we run all the four commands here together: https://github.com/Azure/azure-sdk-for-net/pull/35066/checks?check_run_id=12353115415
You can append the directory name of your RP as an argument after the first two commands to only run in your RP which takes shorter in time and does not introduce unexpected changes.

ArcturusZhang (approved):

sdk/graphservices/Azure.ResourceManager.GraphServices/README.md
```diff
@@ -0,0 +1,80 @@
+# Microsoft Azure GraphServices management client library for .NET
+
+**[Describe the service briefly first.]**
```
ArcturusZhang, line 3: Could we open another PR to introduce some introduction of your RP here?

PR #35075, opened 2023-03-22
Fix recognize complete event construction through two new events

wangrui-msft:

Add two new events: 1. RecognizeChoiceCompleted 2. RecognizeDtmfCompleted for manual contructions of recognize completed event. Since the recognizecomplete event  only include the recognizeResult which was an abstract class, it is not allowed to serialize this manual constructed class, and only able to be deserialized from the incoming json.

The two new events would help customer to construct the recognize completed event with specific recognize result. 


sdk/communication/Azure.Communication.CallAutomation/src/Models/Events/RecognizeDtmfCompleted.cs
```diff
@@ -0,0 +1,47 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Text.Json;
+using Azure.Core;
+using System.Text.Json.Serialization;
+using System;
+
+namespace Azure.Communication.CallAutomation.Models.Events
```
minwoolee-msft, line 9: namespace to be Azure.Communication.CallAutomation

sdk/communication/Azure.Communication.CallAutomation/src/Models/Events/RecognizeDtmfCompleted.cs
```diff
@@ -0,0 +1,47 @@
...
+        /// <summary> Get the recognize Tone result. </summary>
+        public CollectTonesResult CollectTonesResult { get; }
+        /// <summary>
+        /// The recognition type.
+        /// </summary>
+        [CodeGenMember("RecognitionType")]
+        [JsonConverter(typeof(EquatableEnumJsonConverter<CallMediaRecognitionType>))]
+        public CallMediaRecognitionType RecognitionType { get; set; }
+
+        /// <summary> Initializes a new instance of RecognizeChoiceCompleted. </summary>
```
minwoolee-msft, line 25: Summary here needs an update

sdk/communication/Azure.Communication.CallAutomation/src/Models/Events/RecognizeChoiceCompleted.cs
```diff
@@ -0,0 +1,48 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Text.Json;
+using Azure.Core;
+using System.Text.Json.Serialization;
+using System;
+
+namespace Azure.Communication.CallAutomation.Models.Events
```
minwoolee-msft, line 9: namespace update Azure.Communication.CallAutomation

sdk/communication/Azure.Communication.CallAutomation/src/Models/Events/RecognizeDtmfCompleted.cs
```diff
@@ -0,0 +1,47 @@
...
+    public class RecognizeDtmfCompleted : RecognizeCompleted
+    {
+        /// <summary> Get the recognize Tone result. </summary>
+        public CollectTonesResult CollectTonesResult { get; }
+        /// <summary>
+        /// The recognition type.
+        /// </summary>
+        [CodeGenMember("RecognitionType")]
+        [JsonConverter(typeof(EquatableEnumJsonConverter<CallMediaRecognitionType>))]
+        public CallMediaRecognitionType RecognitionType { get; set; }
```
minwoolee-msft, line 23: Internal set

sdk/communication/Azure.Communication.CallAutomation/src/Models/Events/RecognizeChoiceCompleted.cs
```diff
@@ -0,0 +1,48 @@
...
+    {
+        /// <summary> The recognize choice result. </summary>
+        public ChoiceResult ChoiceResult { get; }
+
+        /// <summary>
+        /// The recognition type.
+        /// </summary>
+        [CodeGenMember("RecognitionType")]
+        [JsonConverter(typeof(EquatableEnumJsonConverter<CallMediaRecognitionType>))]
+        public CallMediaRecognitionType RecognitionType { get; set; }
```
minwoolee-msft, line 24: Internal set

sdk/communication/Azure.Communication.CallAutomation/src/Models/Events/RecognizeCompleted.cs
minwoolee-msft, line 1: There was suggestion from Azure Review board, that may be related to this:
> There's nothing for a customer to polymorph over here. This is an internal implementation detail we should remove from the public API.

Do you know if customer has usage for this class? Since we have 2 classes that now inherits from this, we can have this removed to avoid confusion?

wangrui-msft: Basically it is used by customer for recognizeComplete Event serialization and deserialization validation. These two events are only for validation purpose. On functional workflow, customer is expected to use recognizeComplete for receiving the event.

sdk/communication/Azure.Communication.CallAutomation/src/Models/Events/RecognizeDtmfCompleted.cs
wangrui-msft: do you mean set the recognitionType as private field?

sdk/communication/Azure.Communication.CallAutomation/src/Models/Events/RecognizeCompleted.cs
minwoolee-msft: I am curious how would these derived classes (RecognizeChoiceCompleted/RecognizeDtmfCompleted) would be used - because the events classes are read-only that is returned by either EventParser or EventProcessor (customer would not create event manually, unless in testing cases, thats why all events have internal constructor)

These derived classes doesn't seem to be returned by either of parser/processor, and have public constructors - how would customer use this classes to their benefits?

wangrui-msft: The RecognizeCompleted Event is used for parsing and valiation. These two classes are used for testing RecognizeCompleted Event only, it could be viewed as using the inherited classes RecognizeChoiceCompleted/RecognizeDtmfCompleted to build RecognizeCompleted  event. Since the RecognizeCompleted Event is not serializable to json , due to the abstract class recognizeResult, we have to find a way to let the recognizeComplete event properties written to json.

minwoolee-msft: > These two classes are used for testing RecognizeCompleted Event only, it could be viewed as using the inherited classes RecognizeChoiceCompleted/RecognizeDtmfCompleted to build RecognizeCompleted event

If i understand correctly - if these events are used for testing and building recognizeCompleted only, then should it be internal class? it would be internal concept that customer would not use unless they need to do testing. (and for testing purpose, they can use factory)
I can set up a meeting offline for further discussion

wangrui-msft: yes, I would set them as internal

minwoolee-msft (approved):

PR #35081, opened 2023-03-22
[Storage] [DataMovement] Add BlobType option to `BlobStorageResourceContainerOptions`

### LINKED ISSUE #32731: [DataMovement] [Blobs] Change Directory Options to take in more general options for all blob types
### Library name and version

Azure.Storage.DataMovement.Blobs

### Query/Question

These options only apply to block blobs. For now we can support block blobs.

We can also try to generalize the different blob types since most take in general BlobRequestConditions and Copy conditions.

Related to: https://github.com/Azure/azure-sdk-for-net/issues/32728

### Environment

_No response_

jalauzon-msft:
This change adds a new option, `BlobType`, to `BlobStorageResourceContainerOptions` which allows users to specify which type of blob they would like to upload when uploading to a container.

Resolves #32731

### DIFF
--- a/sdk/storage/Azure.Storage.DataMovement.Blobs/src/BlobStorageResourceContainerOptions.cs
+++ b/sdk/storage/Azure.Storage.DataMovement.Blobs/src/BlobStorageResourceContainerOptions.cs
@@ -1,9 +1,6 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-using System;
-using System.Collections.Generic;
-using Azure.Storage.DataMovement;
 using Azure.Storage.Blobs.Models;
 using Azure.Storage.DataMovement.Models;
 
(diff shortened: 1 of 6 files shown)


sdk/storage/Azure.Storage.DataMovement.Blobs/src/BlobStorageResourceContainerOptions.cs
```diff
@@ -34,5 +31,12 @@ public class BlobStorageResourceContainerOptions
         /// Only applies when calling <see cref="BlockBlobStorageResource.CopyBlockFromUriAsync(StorageResource, HttpRange, bool, long, StorageResourceCopyFromUriOptions, System.Threading.CancellationToken)"/>.
         /// </summary>
         public TransferCopyMethod CopyMethod { get; set; }
+
+        /// <summary>
+        /// The <see cref="BlobType"/> that will be used when uploading blobs to the destiantion.
```
amnguye, line 36: NIT Spelling destination

amnguye:
If the live test Mac pipeline tests fail, that's kinda expected cause there's a known issue for that.

amnguye (approved):

PR #35087, opened 2023-03-22
Test authentication when Key Vault switches tenants

### LINKED ISSUE #35086: New token not fetched when tenant ID changes
https://github.com/Azure/azure-sdk-for-net/blob/b4f609fc38b095de6b714ee5e6745cddb9e06559/sdk/core/Azure.Core/src/Pipeline/BearerTokenAuthenticationPolicy.cs#L413-L415

I'm working on a test for when a Key Vault changes tenants and everything works up until this method is called after the tenant changes, so a new token for the new tenant is never created. Seems this should also check if the tenant has changed.

heaths:
Fixes #35086

### DIFF
--- a/sdk/core/Azure.Core/tests/BearerTokenAuthenticationPolicyTests.cs
+++ b/sdk/core/Azure.Core/tests/BearerTokenAuthenticationPolicyTests.cs
@@ -789,6 +789,109 @@ namespace Azure.Core.Tests
             Assert.IsTrue(logged);
         }
 
+        [Test]
+        public async Task BearerTokenAuthenticationPolicy_SwitchedTenants()
+        {
+            var responses = new[]
+            {
+                new MockResponse(401)
+                    .WithHeader("WWW-Authenticate", @"Bearer authorization=""https://login.windows.net/de763a21-49f7-4b08-a8e1-52c8fbc103b4"", resource=""https://vault.azure.net"""),
+
+                new MockResponse(200),
+                new MockResponse(200),
+
+                // Moved tenants.
+                new MockResponse(401)
+                    .WithHeader("WWW-Authenticate", @"Bearer authorization=""https://login.windows.net/72f988bf-86f1-41af-91ab-2d7cd011db47"", resource=""https://vault.azure.net""")
+                    .WithJson("""
+                    {
+                        "error": {
+                            "code": "Unauthorized",
+                            "message": "AKV10032: Invalid issuer. Expected one of https://sts.windows.net/72f988bf-86f1-41af-91ab-2d7cd011db47/, https://sts.windows.net/f8cdef31-a31e-4b4a-93e4-5f571e91255a/, https://sts.windows.net/e2d54eb5-3869-4f70-8578-dee5fc7331f4/, https://sts.windows.net/33e01921-4d64-4f8c-a055-5bdaffd5e33d/, https://sts.windows.net/975f013f-7f24-47e8-a7d3-abc4752bf346/, found https://sts.windows.net/96be4b7a-defb-4dc2-a31f-49ee6145d5ab/."
+                        }
+                    }
+                    """),
+
+                new MockResponse(200),
+            };
+
+            var transport = CreateMockTransport(responses);
+
+            var credential = new TokenCredentialStub((r, c) => new(Guid.NewGuid().ToString(), DateTimeOffset.Now.AddHours(2)), IsAsync);
+            var policy = new ChallengeBasedAuthenticationPolicy(credential, "scope");
+
+            await SendGetRequest(transport, policy, uri: new("https://example.com/1/Original"));
+            Assert.AreEqual("de763a21-49f7-4b08-a8e1-52c8fbc103b4", policy.TenantId);
+
+            await SendGetRequest(transport, policy, uri: new("https://example.com/1/Original"));
+            Assert.AreEqual("de763a21-49f7-4b08-a8e1-52c8fbc103b4", policy.TenantId);
+
+            await SendGetRequest(transport, policy, uri: new("https://example.com/1/Original"));
+            Assert.AreEqual("72f988bf-86f1-41af-91ab-2d7cd011db47", policy.TenantId);
+        }
+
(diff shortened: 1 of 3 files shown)


sdk/resourcemanager/Azure.ResourceManager/src/Resources/Custom/TenantResource.cs
```diff
@@ -16,8 +16,10 @@ namespace Azure.ResourceManager.Resources
     [CodeGenSuppress("GetAsync", typeof(CancellationToken))]
     [CodeGenSuppress("GetAvailableLocations", typeof(CancellationToken))]
     [CodeGenSuppress("GetAvailableLocationsAsync", typeof(CancellationToken))]
+#pragma warning disable CA1825 // Avoid zero-length array allocations
```
heaths, line 19: @m-nash this was necessary with Visual Studio 17.6.0 Preview 2.0 and has been for a little while, IIRC. We may need a constructor overload because you can't pass `Array.Empty<Type>()` since it's not a constant expression. Or is this a known issue?

sdk/core/Azure.Core/tests/BearerTokenAuthenticationPolicyTests.cs
```diff
@@ -789,6 +789,109 @@ public async Task BearerTokenAuthenticationPolicy_BackgroundRefreshFailsAndLogs(
...
+            var policy = new ChallengeBasedAuthenticationPolicy(credential, "scope");
+
+            await SendGetRequest(transport, policy, uri: new("https://example.com/1/Original"));
+            Assert.AreEqual("de763a21-49f7-4b08-a8e1-52c8fbc103b4", policy.TenantId);
+
+            await SendGetRequest(transport, policy, uri: new("https://example.com/1/Original"));
+            Assert.AreEqual("de763a21-49f7-4b08-a8e1-52c8fbc103b4", policy.TenantId);
+
+            await SendGetRequest(transport, policy, uri: new("https://example.com/1/Original"));
+            Assert.AreEqual("72f988bf-86f1-41af-91ab-2d7cd011db47", policy.TenantId);
```
JoshLove-msft, line 830: Is there a way to validate that the new token is generated when the tenant changes?

sdk/core/Azure.Core/tests/BearerTokenAuthenticationPolicyTests.cs
```diff
@@ -789,6 +789,109 @@ public async Task BearerTokenAuthenticationPolicy_BackgroundRefreshFailsAndLogs(
...
+        }
+
+        private class ChallengeBasedAuthenticationPolicy : BearerTokenAuthenticationPolicy
```
JoshLove-msft, line 833: nit: consider using Test in the name - I initially thought this was an released policy when I saw it used in the test.

sdk/resourcemanager/Azure.ResourceManager/src/Resources/Custom/TenantResource.cs
annelo-msft: I've seen this issue locally - is it possible to fix the issue rather than disable the warning?

sdk/core/Azure.Core/tests/BearerTokenAuthenticationPolicyTests.cs
heaths: I can see, but at some point I'm just going to end up reimplementing Key Vault's `ChallengeBasedAuthenticationPolicy`, which already tests this scenario. Maybe I should just copy it and be done with it rather than try to come up with a minimal repro.

sdk/core/Azure.Core/tests/BearerTokenAuthenticationPolicyTests.cs
```diff
@@ -789,6 +789,109 @@ public async Task BearerTokenAuthenticationPolicy_BackgroundRefreshFailsAndLogs(
             Assert.IsTrue(logged);
         }
 
+        [Test]
+        public async Task BearerTokenAuthenticationPolicy_SwitchedTenants()
```
christothes, line 793: This feels like it should be a test for the real `ChallengeBasedAuthenticationPolicy`. The unit test here could be simplified to just validate that the cache recognizes a change in tenantId.

heaths: I already added one in #35474 but you said you wanted one local to Azure.Core.

christothes: Ah, right. But I think the local one here should only need to validate the new token cache invalidation behavior.

sdk/resourcemanager/Azure.ResourceManager/src/Resources/Custom/TenantResource.cs
annelo-msft: Would it make sense to file an issue to track the work?

sdk/core/Azure.Core/tests/BearerTokenAuthenticationPolicyTests.cs
heaths: It should now with the latest push. Please have another look.

sdk/core/Azure.Core/tests/BearerTokenAuthenticationPolicyTests.cs
```diff
@@ -789,6 +789,114 @@ public async Task BearerTokenAuthenticationPolicy_BackgroundRefreshFailsAndLogs(
...
+
+                new MockResponse(200),
+            };
+
+            var transport = CreateMockTransport(responses);
+
+            string tenantId = null;
+            var credential = new TokenCredentialStub((r, c) =>
+            {
+                tenantId = r.TenantId;
```
christothes, line 823: should we also validate that the stub only gets invoked twice to validate that the cache satisfies the request as expected?

christothes (approved):

heaths:
I opened #36059 to track the failing test; otherwise, @JoshLove-msft and @christothes are good with the merge.

PR #35113, opened 2023-03-23
[Event Hubs] Tweak processor startup

jsquire:
# Summary

The focus of these changes is to tweak the approach used to validate connection permissions when the processor is first starting up.  Previously, it would attempt a read using the same explicit consumer used when running. That would potentially cause the processor working on that partition to temporarily be pushed off the partition.  Though it would recover on the next load balancing cycle, this would cause a delay in processing.

### DIFF
--- a/sdk/eventhub/Azure.Messaging.EventHubs.Processor/CHANGELOG.md
+++ b/sdk/eventhub/Azure.Messaging.EventHubs.Processor/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Bugs Fixed
 
-- Changed the approach that the event processor uses to validate permissions on startup to ensure that it uses a non-exclusive consumer as to not interfere with another processor that is already running.  Additionally, an error denying the non-exclusive read is now expected and accepted as proof of a valid connection.
+- Changed the approach that the event processor uses to validate permissions on startup to ensure that it does not interrupt other processors already running by temporarily asserting ownership of a partition.
 
 ### Other Changes
 
--- a/sdk/eventhub/Azure.Messaging.EventHubs.Processor/CHANGELOG.md
+++ b/sdk/eventhub/Azure.Messaging.EventHubs.Processor/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Bugs Fixed
 
+- Changed the approach that the event processor uses to validate permissions on startup to ensure that it uses a non-exclusive consumer as to not interfere with another processor that is already running.  Additionally, an error denying the non-exclusive read is now expected and accepted as proof of a valid connection.
+
 ### Other Changes
 
 ## 5.8.1 (2023-03-09)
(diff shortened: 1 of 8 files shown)


sdk/eventhub/Azure.Messaging.EventHubs.Processor/CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 
 ### Bugs Fixed
 
+- Changed the approach that the event processor uses to validate permissions on startup to ensure that it uses a non-exclusive consumer as to not interfere with another processor that is already running.  Additionally, an error denying the non-exclusive read is now expected and accepted as proof of a valid connection.
```
christothes, line 11: This seems very implementation specific - do you think we could re-word to focus on the customer observed behavior difference after the fix?

Also, I think the mention of the error could be confusing unless it is observable by developers.

christothes (approved):

sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/EventProcessor.cs
```diff
@@ -2029,12 +2041,17 @@ private async Task ValidateEventHubsConnectionAsync(CancellationToken cancellati
...
-            var consumer = CreateConsumer(ConsumerGroup, properties.PartitionIds[0], "validate", EventPosition.Earliest, connection, Options);
+            var consumer = CreateConsumer(ConsumerGroup, properties.PartitionIds[0], $"SV-{ Identifier }", EventPosition.Earliest, connection, Options, false);
 
             try
             {
                 await consumer.ReceiveAsync(1, TimeSpan.FromMilliseconds(5), cancellationToken).ConfigureAwait(false);
             }
+            catch (EventHubsException ex) when (ex.Reason == EventHubsException.FailureReason.ConsumerDisconnected)
+            {
+                // This is expected when another processor is already running; no action is needed, as it
```
JoshLove-msft, line 2052: Do we still intend to call close or should we return?

sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/EventProcessor.cs
```diff
@@ -2029,12 +2041,17 @@ private async Task ValidateEventHubsConnectionAsync(CancellationToken cancellati
             // To ensure validity of the requested consumer group and that at least one partition exists,
             // attempt to read from a partition.
 
-            var consumer = CreateConsumer(ConsumerGroup, properties.PartitionIds[0], "validate", EventPosition.Earliest, connection, Options);
+            var consumer = CreateConsumer(ConsumerGroup, properties.PartitionIds[0], $"SV-{ Identifier }", EventPosition.Earliest, connection, Options, false);
```
JoshLove-msft, line 2044: What does "SV" stand for?

JoshLove-msft (approved):

sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/EventProcessor.cs
jsquire: We intend to close it.   When the processor starts, we're using the consumer to validate the endpoint, credentials, and permissions so that we can fail clearly rather than run in a state where we know we'll never recover.

This consumer is just used to perform that validation and then it is no longer needed.

sdk/eventhub/Azure.Messaging.EventHubs.Processor/CHANGELOG.md
christothes: That sounds better!

jsquire: Applying manually in a commit for all readmes.

Issue #34763, opened 2023-03-07
[Event Hubs] EventData Instrumentation should not be cleared after adding to batch


JoshLove-msft:
Currently, we clear the instrumentation details added after an event is successfully added to a batch. We don't do this when sending events without using a batch. The reasoning is that when using a batch, we are copying the event and not sending the instance that the user handed us, whereas when sending without a batch we don't make a separate copy.

In Service Bus, we never clear the instrumentation details.

After talking to @lmolkova it sounds like we may want to consider not clearing the data that we instrument as it is technically correct that we use the same instrumentation details for all sends involving that event, (whether or not the underlying event was copied).

PR #35124, opened 2023-03-24
Retain instrumentation after adding event to batch

JoshLove-msft:
Addresses https://github.com/Azure/azure-sdk-for-net/issues/34763

### DIFF
--- a/sdk/eventhub/Azure.Messaging.EventHubs/src/Producer/EventDataBatch.cs
+++ b/sdk/eventhub/Azure.Messaging.EventHubs/src/Producer/EventDataBatch.cs
@@ -180,34 +180,16 @@ namespace Azure.Messaging.EventHubs.Producer
             {
                 AssertNotLocked();
 
-                var messageScopeCreated = false;
-                var traceparent = default(string);
+                ClientDiagnostics.InstrumentMessage(eventData.Properties, DiagnosticProperty.EventActivityName, out var traceparent, out var tracestate);
 
-                try
+                var added = InnerBatch.TryAdd(eventData);
+
+                if ((added) && (traceparent != null))
                 {
-                    ClientDiagnostics.InstrumentMessage(eventData.Properties, DiagnosticProperty.EventActivityName, out messageScopeCreated, out traceparent, out var tracestate);
-
-                    var added = InnerBatch.TryAdd(eventData);
-
-                    if ((added) && (traceparent != null))
-                    {
-                        EventDiagnosticIdentifiers.Add((traceparent, tracestate));
-                    }
-
-                    return added;
+                    EventDiagnosticIdentifiers.Add((traceparent, tracestate));
                 }
-                finally
-                {
-                    // If a new message scope was added when instrumenting the instance, the identifier was
-                    // added during this call.  If so, remove it so that the source event is not modified; the
-                    // instrumentation will have been captured by the batch's copy of the event, if it was accepted
-                    // into the batch.
 
-                    if ((messageScopeCreated) && (traceparent != null))
-                    {
-                        MessagingClientDiagnostics.ResetEvent(eventData.Properties);
-                    }
-                }
+                return added;
             }
         }
 
--- a/sdk/core/Azure.Core/src/Shared/MessagingClientDiagnostics.cs
+++ b/sdk/core/Azure.Core/src/Shared/MessagingClientDiagnostics.cs
@@ -156,12 +156,10 @@ namespace Azure.Core.Shared
         /// </summary>
         /// <param name="properties">The dictionary of application message properties.</param>
         /// <param name="activityName">The activity name to use for the diagnostic scope.</param>
-        /// <param name="scopeCreated">Whether or not a new scope was created while instrumenting the message.</param>
         /// <param name="traceparent">The traceparent that was either added, or that already existed in the message properties.</param>
         /// <param name="tracestate">The tracestate that was either added, or that already existed in the message properties.</param>
-        public void InstrumentMessage(IDictionary<string, object> properties, string activityName, out bool scopeCreated, out string? traceparent, out string? tracestate)
+        public void InstrumentMessage(IDictionary<string, object> properties, string activityName, out string? traceparent, out string? tracestate)
         {
-            scopeCreated = false;
             traceparent = null;
             tracestate = null;
 
@@ -171,7 +169,6 @@ namespace Azure.Core.Shared
                     activityName,
                     DiagnosticScope.ActivityKind.Producer);
                 messageScope.Start();
-                scopeCreated = true;
 
                 Activity activity = Activity.Current;
                 if (activity != null)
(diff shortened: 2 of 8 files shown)


sdk/eventhub/Azure.Messaging.EventHubs/src/Producer/EventDataBatch.cs
```diff
@@ -180,34 +180,16 @@ public bool TryAdd(EventData eventData)
...
-                {
-                    ClientDiagnostics.InstrumentMessage(eventData.Properties, DiagnosticProperty.EventActivityName, out messageScopeCreated, out traceparent, out var tracestate);
-
-                    var added = InnerBatch.TryAdd(eventData);
-
-                    if ((added) && (traceparent != null))
-                    {
-                        EventDiagnosticIdentifiers.Add((traceparent, tracestate));
-                    }
+                var added = InnerBatch.TryAdd(eventData);
```
jsquire, line 185: If we do not accept the event into the batch, we should continue to remove the instrumentation, no?   Why would we keep it?

jsquire (changes requested):

JoshLove-msft:
> It just feels a bit wrong to me that we're modifying data that you give us for a `TryAdd` - even when we explicitly reject it.

Yeah I would agree here - it probably makes more sense to remove the identifier if it gets rejected.

m-redding:
I feel like since `SendAsync` doesn't remove the instrumentation on failure, the best parallel for batches is to also not remove instrumentation on failure? Even though the failures have different natures, it could potentially be useful to have tracing information for all failures. 
The alternative could be to instrument the whole batch when it is passed to send, rather than when events are added to the batch. But I'm not sure about the performance implications/implementation challenges of that.

lmolkova:
I agree that adding context for rejected events is not helpful. I wonder if it's possible to instrument the message only after it's successfully added to the batch?

We instrument individual events so they
- capture the context in which they were added to batch
- we can add link from send span to all messages being published. So we already have span per publish and span per each message. 

If message is added to batch, but never published, users would see it (there will be no publish span).
If message was never added, there is no immediate need to trace this fact.

JoshLove-msft:
> If message is added to batch, but never published, users would see it (there will be no publish span).
> If message was never added, there is no immediate need to trace this fact.

The way things are currently implemented, we create the Message/Event scope regardless of whether the message/event  is successfully added to the batch. I'm not actually sure if we can avoid doing this if we want to instrument the message/event up front so that the calculation is correct (we could maybe just add overhead for the diagnostic-id without instrumenting up front). Assuming that we stick with the current implementation of creating the message/event scope up-front, wouldn't it be more confusing for users if this scope is not reflected anywhere in the message/event properties?

jsquire:
> The way things are currently implemented, we create the Message/Event scope regardless of whether the message/event is successfully added to the batch. I'm not actually sure if we can avoid doing this if we want to instrument the message/event up front so that the calculation is correct (we could maybe just add overhead for the diagnostic-id without instrumenting up front). Assuming that we stick with the current implementation of creating the message/event scope up-front, wouldn't it be more confusing for users if this scope is not reflected anywhere in the message/event properties, but they still see that a span was created in the telemetry?

If we do that, I think we end up with a non-trivial performance impact as we would have to serialize twice - once to measure and once with the actual instrumentation.

What we do today is:
- Instrument the message
- Serialize to AMQP format and measure
- If we accept the message in the batch, we keep the serialized bytes

If we were to instrument with a dummy to measure, we would have to:
- Instrument the message with the dummy value
- Serialize to AMQP format and measure
- If we accept the message in the batch, then instrument it for real, serialize it to AMQP format again, keep the serialized bytes



What happens if we create the scope, instrument the event, we reject it, then send the batch... then we add that event to a new batch and send that.  What would someone see in telemetry?   Does that make it look like the event was in two batches when it was really only in one of them?

lmolkova:
Chatted with @JoshLove-msft offline and decided that injecting context regardless is a good feature.

Rationale:
The context we capture on the event should continue the one when the message was created (for correlation and causation). E.g. when I get incoming requests from the upstream service, I create an event and add it to a batch. The event's context should be a child of the incoming request.
If the event does not fit into a batch and is published later on in background thread or in scope of some other operation, it should not change the context.

So:
1. I created a message and tried to add it to a batch at 1:58 pm 
2. It did not fit and I  added it to a different one at 2:00 pm
3. it was published at 2:00:01 pm

 
It's cool to see that message was sitting and waiting to be added to batch for 2 minutes. Otherwise I'd see

1. something appeared out of blue at 2:00 pm, it has no parent (e.g. stayed in the background queue)
2. it was published at 2:00:01 pm

To provide enough observability into this we can add an attribute explaining event was not added right away:

```csharp
if (hasTraceContext() {
  using var scope = createMessageScope(msg)
  if (!tryAdd(msg)) {
     scope.SetAttribute("message.added", false); // need to think about naming a bit
 }
} else { tryAdd(msg) }
```

This way I can also write a query to check if any messages were dropped because adding was not successful.

JoshLove-msft:
> Chatted with @JoshLove-msft offline and decided that injecting context regardless is a good feature.
> 
> Rationale: The context we capture on the event should continue the one when the message was created (for correlation and causation). E.g. when I get incoming requests from the upstream service, I create an event and add it to a batch. The event's context should be a child of the incoming request. If the event does not fit into a batch and is published later on in background thread or in scope of some other operation, it should not change the context.
> 
> So:
> 
> 1. I created a message and tried to add it to a batch at 1:58 pm
> 2. It did not fit and I  added it to a different one at 2:00 pm
> 3. it was published at 2:00:01 pm
> 
>   It's cool to see that message was sitting and waiting to be added to batch for 2 minutes. Otherwise I'd see
> 
> 1. something appeared out of blue at 2:00 pm, it has no parent (e.g. stayed in the background queue)
> 2. it was published at 2:00:01 pm
> 
> To provide enough observability into this we can add an attribute explaining event was not added right away:
> 
> ```cs
> if (!hasTraceContext(msg)) {
>   using var scope = createMessageScope(msg)
>   if (!tryAdd(msg)) {
>      scope.SetAttribute("message.added", false); // need to think about naming a bit
>  }
> } else { tryAdd(msg) }
> ```
> 
> This way I can also write a query to check if any messages were dropped because adding was not successful and the application didn't check.

This also aligns with @m-redding's reasoning!

JoshLove-msft:
Should we proceed with the PR as-is and come back later to include "message.added" (maybe it needs to be talked over with Otel folks?), or do we think it is important to have this attribute up front since we are changing the behavior in Event Hubs?

lmolkova:
> Should we proceed with the PR as-is and come back later to include "message.added" (maybe it needs to be talked over with Otel folks?), or do we think it is important to have this attribute up front since we are changing the behavior in Event Hubs?

I'm fine with proceeding as is and adding attribute later.

jsquire (approved):
per discussion, it looks like we landed on "we should instrument even when the event is not added" with the rationale that we'll have traceability on that activity and, ultimately, when it was sent.

PR #35145, opened 2023-03-25
Fix scan logic

alrod:
The scan logic for blob trigger is broken - scan blob with the latest scan time is never created and scan logic try to scan the same blobs again and again.

Help for the `ContinuationToken` says:
```
        /// <summary>
        /// Gets the continuation token used to request the next
        /// <see cref="Page{T}"/>.  The continuation token may be null or
        /// empty when there are no more pages.
        /// </summary>
        public abstract string? ContinuationToken { get; }
```

Current code only checks for `null`

### DIFF
--- a/sdk/storage/Microsoft.Azure.WebJobs.Extensions.Storage.Blobs/src/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
+++ b/sdk/storage/Microsoft.Azure.WebJobs.Extensions.Storage.Blobs/src/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
@@ -282,7 +282,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Storage.Blobs.Listeners
             containerScanInfo.ContinuationToken = page.ContinuationToken;
 
             // if ending a cycle then copy currentSweepCycleStartTime to lastSweepCycleStartTime, if changed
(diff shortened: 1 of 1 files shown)


amnguye (approved):

sdk/storage/Microsoft.Azure.WebJobs.Extensions.Storage.Blobs/src/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
```diff
@@ -282,7 +282,7 @@ private void ThrowIfDisposed()
             containerScanInfo.ContinuationToken = page.ContinuationToken;
 
             // if ending a cycle then copy currentSweepCycleStartTime to lastSweepCycleStartTime, if changed
-            if (page.ContinuationToken == null &&
+            if (string.IsNullOrEmpty(page.ContinuationToken) &&
```
mathewc, line 285: Above we're using string.IsNullOrWhiteSpace - we should be consistent. Also, do we have proof from logs that this is the actual issue?

alrod: I was able to validate the fix locally.

PR #32814, opened 2022-12-01
Drop netcoreapp3.1, add net7.0 support

heaths:
Resolves #32596. With netcoreapp3.1 falling out of support soon, we're upgrading to net7.0, net6.0, and, on Windows, net461.

PR #35151, opened 2023-03-27
Install .NET Core 3.1 SDK on Windows to support Track 1 SDK

live1206:
.NET Core 3.1 SDK does not exist on Windows hosted machine anymore, we still need it for Track 1 SDK
https://github.com/Azure/azure-sdk-for-net/pull/35148/checks?check_run_id=12290486264


live1206:
@weshaggard Could you help to review this? Thanks!

jsquire:
@live1206:  Please help me understand why we need a runtime that has reached end-of-life and is no longer supported.   What libraries are failing and what target are they using?


//fyi: @heaths

jsquire (dismissed):
Blocking merge until the scenario is understood.

benbp:
If this is indeed not necessary, I have a PR to remove the step entirely as we don't need to be wasting time on windows jobs installing it: https://github.com/Azure/azure-sdk-for-net/pull/35174

live1206:
> @live1206 I thought we were rolling forward to the latest .NET when running the code generation so why would we still need this? Is there something else that needs it?
> 
> Looking at your link it appears as though there are still some test projects targeting .NET Core 3.1. We should instead update those test projects to .NET 7.0 as .NET 3.1 is out support.

Tried with a draft PR https://github.com/Azure/azure-sdk-for-net/pull/35178, it takes more effort than expected. 
Basically a follow-up of https://github.com/Azure/azure-sdk-for-net/pull/32814

jsquire:
> Tried with a draft PR #35178, it takes more effort than expected. Basically a follow-up of #32814

There's a difference between building with the .NET 7 SDK and changing the target frameworks for libraries.   We need to do the former, we do not want to do the latter.   @live1206 - would you please help us understand what problem you're trying to solve?  What behavior are you seeing that requires changes?

live1206:
> https://github.com/Azure/azure-sdk-for-net/pull/35148/checks?check_run_id=12290486264

@jsquire As I mentioned in the description, the CI is failing for Track 1 .NET SDK PR: https://github.com/Azure/azure-sdk-for-net/pull/35148/checks?check_run_id=12290486264, it is blocking Track 1 .NET SDK release.

jsquire:
> @jsquire As I mentioned in the description, the CI is failing for Track 1 .NET SDK

What should change here is the SDK in use; these should be compiled using the .NET 7 SDK.   We do not want to install/use the .NET Core 3.1 SDK, as it has reached end-of-life.

live1206:
> > @jsquire As I mentioned in the description, the CI is failing for Track 1 .NET SDK
> 
> What should change here is the SDK in use; these should be compiled using the .NET 7 SDK. We do not want to install/use the .NET Core 3.1 SDK, as it has reached end-of-life.

As I explained in https://github.com/Azure/azure-sdk-for-net/pull/35178#discussion_r1151655504, it's not an easy fix to upgrade the target .NET to latest.

eng/pipelines/templates/steps/install-dotnet.yml
```diff
@@ -27,7 +27,7 @@ steps:
       performMultiLevelLookup: true
       version: "6.0.x"
   - task: UseDotNet@2
-    condition: and(succeeded(), ne(variables['Agent.OS'], 'Windows_NT')) # Windows supports MultiLevelLookup and doesn't need explicit framework installation
+    condition: and(succeeded()) # .NET Core 3.1 does not exist on Windows hosted machine anymore, we still need it for Track 1 SDK
```
weshaggard, line 30: To help unblock the current release let's put the windows install of this 3.1 directly in the ci.mgmt.yml file to scope it to only track 1 mgmt. Once that is done we can continue to remove this entire step from install-dotnet.yml, happening in https://github.com/Azure/azure-sdk-for-net/pull/35151. Once that happens we can keep the 3.1 install for all OS's in the ci.mgmt.yml file temporarily. We should continue to work on https://github.com/Azure/azure-sdk-for-net/pull/35178 to break this dependency on 3.1 before we can no longer use this EoL framemwork.

live1206: Added a step in .NET SDK Track 1 CI to install .NET core 3.1 for all OS, so that we can continue with #35174 for the rest.

eng/pipelines/templates/jobs/ci.mgmt.yml
```diff
@@ -15,6 +15,15 @@ jobs:
     steps:
       #- script: "echo $(system.pullrequest.pullrequestnumber), https://github.com/$(build.repository.id), https://github.com/$(build.repository.ID)"
       - template: /eng/pipelines/templates/steps/install-dotnet.yml
+      - task: UseDotNet@2
+        condition: and(succeeded()) # We still need .NET Core 3.1 for .NET Track 1 SDK
```
weshaggard, line 19: condition succeeded() is redundant but doesn't hurt anything either.

weshaggard (approved):

eng/pipelines/templates/jobs/ci.mgmt.yml
```diff
@@ -15,6 +15,15 @@ jobs:
     steps:
       #- script: "echo $(system.pullrequest.pullrequestnumber), https://github.com/$(build.repository.id), https://github.com/$(build.repository.ID)"
       - template: /eng/pipelines/templates/steps/install-dotnet.yml
+      - task: UseDotNet@2
```
weshaggard, line 18: I suspect you likely need this in the Test job below as well.

weshaggard:
I'd like to make sure we get a test run of the mgmt pipeline as well to ensure this change works as expected.

eng/pipelines/templates/jobs/ci.mgmt.yml
weshaggard: Looks like this condition does need to be removed because the and operator requires 2 arguments https://dev.azure.com/azure-sdk/public/_build/results?buildId=2677828&view=results

live1206: Good catch, removed.

eng/pipelines/templates/jobs/ci.mgmt.yml
live1206: Yes, it is needed, wondering why the test run did not fail though.

weshaggard: CI failed to trigger because of the bad and() expression.

weshaggard (approved):

PR #35169, opened 2023-03-27
[Event Hubs] Enrich processor log

jsquire:
# Summary

The focus of these changes is to add the source of the event position chosen for a partition reader to the log, to help troubleshooting to identify whether a checkpoint was found or a default was used.

### DIFF
--- a/sdk/eventhub/Azure.Messaging.EventHubs/tests/Primitives/EventProcessorTests.PartitionProcessing.cs
+++ b/sdk/eventhub/Azure.Messaging.EventHubs/tests/Primitives/EventProcessorTests.PartitionProcessing.cs
@@ -1628,7 +1628,7 @@ namespace Azure.Messaging.EventHubs.Tests
             cancellationSource.CancelAfter(EventHubsTestEnvironment.Instance.TestExecutionTimeLimit);
 
             var partition = new EventProcessorPartition { PartitionId = "4" };
-            var expectedEevntPosition = EventPosition.FromSequenceNumber(332);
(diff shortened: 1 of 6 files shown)


JoshLove-msft (approved):

sdk/eventhub/Azure.Messaging.EventHubs/tests/Primitives/EventProcessorTests.PartitionProcessing.cs
```diff
@@ -1664,7 +1664,8 @@ public async Task CreatePartitionProcessorProcessingTaskLogsTheStartingPosition(
                     mockProcessor.Object.Identifier,
                     mockProcessor.Object.EventHubName,
                     mockProcessor.Object.ConsumerGroup,
-                    expectedEevntPosition.ToString()),
+                    expectedEevntPosition.ToString(),
```
m-redding, line 1667: nit (def in L1631)
```suggestion
                    expectedEventPosition.ToString(),
```

m-redding (approved):

sdk/eventhub/Azure.Messaging.EventHubs/tests/Primitives/EventProcessorTests.PartitionProcessing.cs
jsquire: Nice catch!   Applied by hand to get the definition and other uses.

christothes (approved):

PR #35170, opened 2023-03-27
Moving to the correct LRO pattern

pallavit:
A basic attempt on making synapse move to the correct LRO pattern with documented autorest settings. Still going through adding XMLDocs and test scenarios but wanted to get it out there and see if I am thinking this right.

Here are the steps I took
1. This was a DPG SDK that added both generation1-convenience-client: true and public-clients: true
2. Since it has models - I removed the public-clients:true flag
3. I then moved the existing non-rest client files into customized folder and updated the LRO pattern to use WaitUntil

Please let me know if I am thinking this correctly.


sdk/synapse/Azure.Analytics.Synapse.Spark/src/Customization/SparkBatchClient.cs
```diff
@@ -1,41 +1,243 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+// <auto-generated/>
```
m-nash, line 4: This marker turns off quite a few of our azure sdk analyzers so we should always remove this outside the `.../Generated/...` folder.

sdk/synapse/Azure.Analytics.Synapse.Spark/src/Customization/SparkBatchClient.cs
```diff
@@ -1,41 +1,243 @@
...
+        {
+        }
+
+        /// <summary> Initializes a new instance of SparkBatchClient. </summary>
+        /// <param name="endpoint"> The workspace development endpoint, for example https://myworkspace.dev.azuresynapse.net. </param>
+        /// <param name="sparkPoolName"> Name of the spark pool. </param>
+        /// <param name="credential"> A credential used to authenticate to an Azure Service. </param>
+        /// <param name="livyApiVersion"> Valid api-version for the request. </param>
+        /// <param name="options"> The options for configuring the client. </param>
+        public SparkBatchClient(Uri endpoint, string sparkPoolName, TokenCredential credential, string livyApiVersion = "2019-11-01-preview", SparkClientOptions options = null)
```
pallavit, line 36: Also I am assuming the service version will need to fold into the SparkClientOptions now.

sdk/synapse/Azure.Analytics.Synapse.Spark/src/Customization/SparkBatchOperation.cs
```diff
@@ -12,9 +12,7 @@
 namespace Azure.Analytics.Synapse.Spark
 {
     /// <summary>
-    /// A long-running operation for:
-    /// <see cref="SparkBatchClient.StartCreateSparkBatchJobAsync(SparkBatchJobOptions, bool?, CancellationToken)"/>,
-    /// <see cref="SparkBatchClient.StartCreateSparkBatchJob(SparkBatchJobOptions, bool?, CancellationToken)"/>,
+    /// A long-running operation for spark batch operation.
     /// </summary>
     public class SparkBatchOperation : Operation<SparkBatchJob>
```
pallavit, line 17: Also not sure why SparkBatchOperation does not derive from IOperation

sdk/synapse/Azure.Analytics.Synapse.Spark/src/Customization/SparkBatchClient.cs
m-nash: yeah this is strange to see it here, not sure how the generator did this.

annelo-msft (commented):
@pallavit, how are you thinking about "what is the right LRO pattern"?  We still have a lot of technical debt in this area, so I'm interested in your experience here and if you've learned that we have a single recommended pattern to follow.

pallavit:
> @pallavit, how are you thinking about "what is the right LRO pattern"? We still have a lot of technical debt in this area, so I'm interested in your experience here and if you've learned that we have a single recommended pattern to follow.

While working on unification of the autorest we realized some of the newer SDKs were using incorrect flag combination and so had to be turned to HLCs. This PR was mostly an experiment to see how can we convert this SDK to HLC and use the correct design guideline for exposing LRO (WaitUntil pattern). I may have used LRO term incorrectly here since it was not about how the service exposes LRO but rather how we are exposing it in our APIs.

PR #35175, opened 2023-03-27
[FormRecognizer][MetricsAdvisor] Moved recordings to the assets repo

kinelski:


pallavit:
Did you just have to run the test recording script or were any other steps required? What does the work-flow for changing the recordings look like - after this change?

kinelski:
> Did you just have to run the test recording script or were any other steps required?

I just needed to run the [transition script](https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/testproxy/transition-scripts/generate-assets-json.ps1) and it worked like magic.

> What does the work-flow for changing the recordings look like - after this change?

This is a one-time step, but firstly you need to have the `test-proxy` command line tool installed.
```powershell
dotnet tool update azure.sdk.tools.testproxy --global --add-source https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-for-net/nuget/v3/index.json --version "1.0.0-dev*"
```
Other installation options can be found in their [README](https://github.com/Azure/azure-sdk-tools/blob/main/tools/test-proxy/Azure.Sdk.Tools.TestProxy/README.md#installation).

Regarding the workflow, the only difference from what we do today is when we want to update recordings, since they are not stored in the SDK repo anymore. You need to manually run the command:
```powershell
test-proxy push <path-to-assets-json>
```

JoshLove-msft:
> Regarding the workflow, the only difference from what we do today is when we want to update recordings, since they are not stored in the SDK repo anymore. You need to manually run the command:
> 
> ```powershell
> test-proxy push <path-to-assets-json>
> ```

Is there a way to automate this using a pre/post commit hook? Or something else?

JoshLove-msft:
> > Regarding the workflow, the only difference from what we do today is when we want to update recordings, since they are not stored in the SDK repo anymore. You need to manually run the command:
> > ```powershell
> > test-proxy push <path-to-assets-json>
> > ```
> 
> Is there a way to automate this using a pre/post commit hook? Or something else?

/cc @scottb @weshaggard

pallavit:
> Regarding the workflow, the only difference from what we do today is when we want to update recordings, since they are not stored in the SDK repo anymore. You need to manually run the command

So every partner will need to install test-proxy tool locally and then run the push command. I think we need to have a way to make this easier for the partners.

JoshLove-msft (approved):
automation work still needs to be done but approving to unblock recording of new FR tests - should only impact FR.

pallavit:
> automation work still needs to be done but approving to unblock recording of new FR tests - should only impact FR.

Can we please create a follow-up bug here, so we don't forget.

JoshLove-msft:
> > automation work still needs to be done but approving to unblock recording of new FR tests - should only impact FR.
> 
> Can we please create a follow-up bug here, so we don't forget.

The context is in https://github.com/Azure/azure-sdk-for-net/issues/35208

PR #35178, opened 2023-03-28
Upgrade .NET core 3.1 to .NET 6

live1206:


sdk/advisor/Microsoft.Azure.Management.Advisor/tests/Microsoft.Azure.Management.Advisor.Tests.csproj
```diff
@@ -5,7 +5,7 @@
     <Version>1.0.0</Version>
     <Description>Advisor.Tests Class library</Description>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-    <TargetFrameworks>netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks>net6.0</TargetFrameworks>
```
jsquire, line 8: These should not be changed.  The `netcoreapp3.1` target will compile using the .NET 7 SDK that is standard for the repository and will run on `net6` and `net7`.

Bumping this will break customers who have not upgraded their runtime; while `netcoreapp3.1` is no longer supported, there is no benefit to upgrading the target framework here - especially with the T1 libraries being retired in September.

jsquire (changes requested):

jsquire (dismissed):
Please revert all of the target framework changes.

sdk/advisor/Microsoft.Azure.Management.Advisor/tests/Microsoft.Azure.Management.Advisor.Tests.csproj
weshaggard: @jsquire I would agree with you for libraries but this is the test project and I do think we should move that to a later target framework as it is just ends up being the runtime that we run our tests on. I don't think we care to run them on .NET 3.1 any longer. In fact for the track 2 test projects we have already moved to .NET 6 and 7 (https://github.com/Azure/azure-sdk-for-net/blob/main/eng/Directory.Build.Common.props#L127)

weshaggard: @live1206 we probably should figure out a good place to put a common property for these projects as well.

live1206: @weshaggard @jsquire 
The test run is [failing ](https://dev.azure.com/azure-sdk/public/_build/results?buildId=2663095&view=logs&j=3dc8fd7e-4368-5a92-293e-d53cefc8c4b3&t=45862c6e-83c5-515f-73e2-e7009eff9f9b&l=4041)
image
I had a look at the nuget package [Microsoft.NET.Test.SDK](https://www.nuget.org/packages/Microsoft.NET.Test.SDK), the target framework is .NET Core 3.1. So it seems we can't easily upgrade to .NET 6 here. 

I don't know how long we would keep the source code and tests of Track 1 since it's deprecated. An easy fix for now is to install .NET Core 3.1 on Windows machine in https://github.com/Azure/azure-sdk-for-net/pull/35151. 
Any other options to unblock the Track 1 release for https://github.com/Azure/azure-sdk-for-net/pull/35148?

cc @ArthurMa1978

weshaggard: @ArthurMa1978 how much longer are we going to need to support releases for these? If we need to support them for a while we need to update the target framework of the test projects. .NET Core 3.1 is no longer supported and we will soon be unable to use it at all.

PR #35179, opened 2023-03-28
ACR: Mark .NET-specific transformations

lmolkova:
Might be useful to know which transformations in swagger are .NET-specific

### DIFF
--- a/sdk/containerregistry/Azure.Containers.ContainerRegistry/src/autorest.md
+++ b/sdk/containerregistry/Azure.Containers.ContainerRegistry/src/autorest.md
@@ -82,26 +82,6 @@ directive:
         delete $.responses["201"].schema;
 ```
 
-# Add content-range and content-length parameters to upload chunk
-``` yaml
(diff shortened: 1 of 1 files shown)


annelo-msft (approved):
Will the non-.NET specific transforms move to a central location in a later step?

pallavit:
> Will the non-.NET specific transforms move to a central location in a later step?

If I remember correctly, we try to put common transforms in the swagger file itself correct? Or update the swagger? Not saying we need to do this for GA but something to consider for completeness?

annelo-msft:
> If I remember correctly, we try to put common transforms in the swagger file itself correct? Or update the swagger? Not saying we need to do this for GA but something to consider for completeness?

There is also a practice of putting common transforms in a README in the azure-rest-api-specs repo, sometimes associated with a release tag.  We can pull details if we want to go this route - the nice thing about this is it sits near the swagger without having to go through the review process to modify it.

PR #35188, opened 2023-03-28
Prepare release for AppContainers 1.0.3

ArthurMa1978:


ArcturusZhang (approved):
Since we obsolete a property in this version, should we bump the minor version instead of patch?

PR #35192, opened 2023-03-28
ContentHasher BinaryData support + tests

jaschrep-msft:
- Adds `BinaryData` overloads to `ContentHasher`
- Refactors `ContentHasher` internals to avoid code duplication between `Stream` vs `BinaryData` implementations
- Adds unit tests for `ContentHasher`


amnguye (approved):
LGTM

PR #35199, opened 2023-03-28
Sparse checkout required files for event processor

### LINKED ISSUE #5844: Actions need sparse checkout
The github-event-processor yml files all use actions/checkout@v3. While this does sync the repository with a depth of 1, it still takes around 1m 30seconds or longer to sync up. The rest of the steps have roughly the following times
1. Setup the Actions job ~ 6 seconds
2. Install github-event-processor ~ 10 seconds.
3. github-event-processor processing the event ~ 5 seconds

//CC @jsquire @weshaggard @benbp @kurtzeborn

benbp:
Right now the event processor only needs a couple files in `.github/` but we are doing a full checkout of the repository which can take over a minute. Using a sparse checkout here reduces this step to a couple seconds.

Fixes https://github.com/Azure/azure-sdk-tools/issues/5844

Also updating to version 1.0.0-dev.20230328.3

### DIFF
--- a/.github/workflows/event-processor.yml
+++ b/.github/workflows/event-processor.yml
@@ -29,7 +29,14 @@ jobs:
     name: Handle ${{ github.event_name }} ${{ github.event.action }} event
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - name: 'Sparse Checkout'
+        run: |
+          set -ex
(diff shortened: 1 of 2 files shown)


jsquire (approved):

JimSuplizio (approved):
Looks great @benbp thank you for doing this!

PR #35200, opened 2023-03-28
Remove Azure OperationType mapping

vishweshbankwar:
Having `az.namespace` attribute mapped to OperationType.Azure results in missing az.namespace custom property (This would be ok when we use this tag to set some field on TelemetryItem). For now, we should just pass the tag as it is.


TimothyMothra:
> Having az.namespace attribute mapped to OperationType.Azure results in missing az.namespace custom property

Is this a bug fix? Do we need an entry in the Changelog?

rajkumar-rangaraj (approved):

vishweshbankwar:
> > Having az.namespace attribute mapped to OperationType.Azure results in missing az.namespace custom property
> 
> Is this a bug fix? Do we need an entry in the Changelog?

It is not a bug. We do not know what the final behavior is going to be (it will be based on what we decide to do with az.namespace attribute)

TimothyMothra (approved):

PR #35202, opened 2023-03-28
Use DynamicJson for CLU samples

### LINKED ISSUE #35041: Use DynamicData in CLU samples
For simpler samples and to provide feedback about DynamicJson to @annelo-msft, rewrite at least the CLU samples using DynamicJson. As a stretch, might be good to rewrite all tests using DynamicJson (start with samples, though, in case there are any blockers) so we can stop generating models using the old LLC generator. That would decrease the overall amount of work to update and should make it easier for the code gen folks to iterate on changes to the generator stack (one less potential roadblock).

heaths:
Resolves #35041

### DIFF
--- a/sdk/cognitivelanguage/Azure.AI.Language.Conversations/README.md
+++ b/sdk/cognitivelanguage/Azure.AI.Language.Conversations/README.md
@@ -186,10 +186,9 @@ foreach (dynamic entity in conversationPrediction.entities)
     Console.WriteLine($"Confidence: {entity.confidenceScore}");
     Console.WriteLine();
 
-    dynamic resolutions = entity.resolutions;
-    if (resolutions is not null)
+    if (entity.resolutions is not null)
     {
-        foreach (dynamic resolution in resolutions)
+        foreach (dynamic resolution in entity.resolutions)
         {
             if (resolution.resolutionKind == "DateTimeResolution")
             {
--- a/sdk/cognitivelanguage/Azure.AI.Language.Conversations/README.md
+++ b/sdk/cognitivelanguage/Azure.AI.Language.Conversations/README.md
@@ -44,6 +44,7 @@ Start by importing the namespace for the [`ConversationAnalysisClient`][conversa
 
 ```C# Snippet:ConversationAnalysisClient_Namespaces
 using Azure.Core;
+using Azure.Core.Dynamic;
 using Azure.AI.Language.Conversations;
 ```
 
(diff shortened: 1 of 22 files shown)


sdk/cognitivelanguage/Azure.AI.Language.Conversations/README.md
```diff
@@ -162,39 +163,39 @@ var data = new
 
 Response response = client.AnalyzeConversation(RequestContent.Create(data));
 
-using JsonDocument result = JsonDocument.Parse(response.ContentStream);
-JsonElement conversationalTaskResult = result.RootElement;
-JsonElement conversationPrediction = conversationalTaskResult.GetProperty("result").GetProperty("prediction");
+dynamic conversationalTaskResult = response.Content.ToDynamic();
```
annelo-msft, line 166: I defer to you on what the best thing for CLU customers is, but if you used `response.Content.ToDynamic(DynamicJsonOptions.AzureDefault)` here, you could use PascalCase property accessors in subsequent calls to make the sample code look more like it is using strongly-typed models.  An advantage of this is that if models are provided later, there is less refactoring needed to use them.

sdk/cognitivelanguage/Azure.AI.Language.Conversations/README.md
```diff
@@ -162,39 +163,39 @@ var data = new
...
+    Console.WriteLine($"Category: {entity.category}");
+    Console.WriteLine($"Text: {entity.text}");
+    Console.WriteLine($"Offset: {entity.offset}");
+    Console.WriteLine($"Length: {entity.length}");
+    Console.WriteLine($"Confidence: {entity.confidenceScore}");
     Console.WriteLine();
 
-    if (entity.TryGetProperty("resolutions", out JsonElement resolutions))
+    dynamic resolutions = entity.resolutions;
+    if (resolutions is not null)
```
annelo-msft, line 190: nit: if you wanted, you could collapse the two lines above into a single line.

sdk/cognitivelanguage/Azure.AI.Language.Conversations/README.md
```diff
@@ -302,28 +303,27 @@ var data = new
...
     Console.WriteLine($"Top intent: {respondingProjectName}");
 
-    JsonElement questionAnsweringResponse = targetIntentResult.GetProperty("result");
+    dynamic questionAnsweringResponse = targetIntentResult.result;
     Console.WriteLine($"Question Answering Response:");
-    foreach (JsonElement answer in questionAnsweringResponse.GetProperty("answers").EnumerateArray())
+    foreach (dynamic answer in questionAnsweringResponse.answers)
     {
-        Console.WriteLine(answer.GetProperty("answer").GetString());
+        Console.WriteLine(answer.answer?.ToString());
```
annelo-msft, line 326: The semantics of DynamicJson.ToString() and JsonElement.GetString() are different, although the result is the same if the JSON property has a string value.  The latter will throw if the property doesn't have a string value, whereas DynamicJson.ToString() will always return the underlying JSON as a string.

For disambiguating the type to call Console.WriteLine(), the intended approach is either to use a cast to the expected type:
```csharp
Console.WriteLine((string)answer.answer);
``
Or an interpolated string, as illustrated in the generator samples: https://github.com/Azure/autorest.csharp/issues/3192#issue-1612375261

File: sdk/cognitivelanguage/Azure.AI.Language.Conversations/tests/Samples/Sample2_AnalyzeConversation_OrchestrationPrediction.cs [hunk 4]
```diff
@@ -50,29 +50,28 @@ public void AnalyzeConversationOrchestrationPredictionQuestionAnswering()
...
+                foreach (dynamic answer in questionAnsweringResponse.answers)
                 {
-                    Console.WriteLine(answer.GetProperty("answer").GetString());
+                    Console.WriteLine(answer.answer?.ToString());
                 }
             }
             #endregion
 
-            Assert.That(targetIntentResult.GetProperty("targetProjectKind").GetString(), Is.EqualTo("QuestionAnswering"));
+            Assert.That(targetIntentResult.targetProjectKind?.ToString(), Is.EqualTo("QuestionAnswering"));
```
annelo-msft, line 74:
```suggestion
            Assert.That((string)targetIntentResult.targetProjectKind, Is.EqualTo("QuestionAnswering"));
```

Curious: is there a reason not to use a cast here instead?

heaths: It can throw. See [comment above](https://github.com/Azure/azure-sdk-for-net/pull/35202#discussion_r1152471558).

sdk/cognitivelanguage/Azure.AI.Language.Conversations/README.md
heaths: Given input is camelCase, I still prefer output to also match the original (mostly) camelCase for consistency.

sdk/cognitivelanguage/Azure.AI.Language.Conversations/README.md
heaths: Thinking about this more, I'd rather leave `ToString`. All the interpolated strings above are calling `ToString` so this is consistent and, I bet, will be more common.

sdk/cognitivelanguage/Azure.AI.Language.Conversations/README.md
annelo-msft: I was thinking just `if (entity.resolutions is not null)` ... would that not work?

sdk/cognitivelanguage/Azure.AI.Language.Conversations/README.md
annelo-msft: Yes, happy to document that - good call out.

Sorry, no, I must have misspoken.  GetString() is not a valid call on DynamicJson - I only mentioned it because in the original code that you are changing, you had called it on the JsonElement.

Since these are samples that customers may copy, I think it would be better if we show the intended usage. I see your point that ToString() works here in a call to Console.WriteLine(), but I'd be concerned people might confuse it with the goal of GetString().  Do you think that is valid?  Maybe this isn't a concern.

heaths:
I'm going to break out the change to CLU tests into another PR that we can merge immediately. We may be shipping a new version of CLU and/or QA soon but not entirely sure, so we may not want to update our samples / docs just yet. I think it's great to update the tests, though, so we can keep some real-world use in rotation as changes are made to DynamicJson.

annelo-msft:
> I think it's great to update the tests, though, so we can keep some real-world use in rotation as changes are made to DynamicJson.

That's great.  Thanks @heaths!

heaths:
Converting to draft until we know if we are soon GA'ing a new CLU. /cc @kristapratico @mayurid

PR #35204, opened 2023-03-28
Fix escaping bug in MutableJsonDocument

annelo-msft:
### Problem

In [this AppConfig round-trip sample](https://github.com/Azure/azure-sdk-for-net/blob/dynamicjson_feedback/samples/AppConfigDemo/RoundTripSample.cs#L98), if the highlighted line is commented out, the AppConfig service will respond to the request using DynamicData with a 400, where the same request using JsonDocument or JsonNode will succeed with a 200.  This is because, prior to this PR, DynamicData was not handling escaping the same way JsonDocument does.

In the prior implementation, MutableJsonDocument.WriteTo() created a reader on the original JSON buffer, without doing any escaping, and passed reader.ValueSpan directly to writer.WriteStringValue().  Utf8JsonWriter will always escape string values (unless passed options that direct it otherwise), and JsonDocument deals with this by [unescaping a string](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Text.Json/src/System/Text/Json/Reader/JsonReaderHelper.Unescaping.cs#L393) prior to calling WriteString().

### Solution

Since all of the escaping logic in JsonDocument is held in internal types, the best we could do to duplicate the logic would be to copy those helper classes, or hand-write similar logic.  This is clearly error prone.  Instead, this PR moves the implementation of WriteTo() up a level of abstraction, so it can take advantage of the escaping logic in JsonDocument itself.

### Implementation

This PR:
- Removes the Utf8JsonReader/Utf8JsonWriter implementation of [MutableJsonDocument.WriteTo()](https://github.com/Azure/azure-sdk-for-net/pull/35204/files#diff-7cfa0a92a2ef507a2b25dd94a7d04806839fb54c7e66fa2bcc5e8665d9d06385)
- Adds an internal JsonDocument/Utf8JsonWriter implementation of [MutableJsonElement.WriteTo()](https://github.com/Azure/azure-sdk-for-net/pull/35204/files#diff-0b846adb4112f2d61c0ffb34cb025964feba66e862cd55e6e052e43f578dd256) and [calls through to it from MutableJsonDocument.WriteTo()](https://github.com/Azure/azure-sdk-for-net/pull/35204/files#diff-8989081de72d6619e2a4cca3196032f8bcd3ce142eb2a73be882e6cc06c3f346R78)
- Updates tests for WriteTo() to [compare the result of Mutable* WriteTo()](https://github.com/Azure/azure-sdk-for-net/pull/35204/files#diff-404b45ffe3711cbf1d5b4229cb6c300ffc0c1088e55ace215b8c2befd902e0f6R1113-R1126) methods to the result of JsonDocument.WriteTo() directly.

Contributes to https://github.com/Azure/azure-sdk-for-net/issues/33856


sdk/core/Azure.Core.Experimental/src/MutableJsonDocument.WriteTo.cs
```diff
@@ -186,7 +187,16 @@ private void WriteString(string path, int highWaterMark, ref Utf8JsonReader read
                 return;
             }
 
-            writer.WriteStringValue(reader.ValueSpan);
+#if NET6_0_OR_GREATER
+            Span<byte> span = stackalloc byte[reader.ValueSpan.Length + 2];
+            span[0] = (byte)'"';
+            reader.ValueSpan.CopyTo(span.Slice(1));
+            span[span.Length - 1] = (byte)'"';
+            writer.WriteRawValue(span);
```
annelo-msft, line 195: Utf8JsonReader has a [ValueIsEscaped property](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.utf8jsonreader.valueisescaped) we could check prior to these calls, but it's only available in .NET 7+, arg.

sdk/core/Azure.Core.Experimental/src/MutableJsonDocument.cs
```diff
@@ -56,15 +56,15 @@ public void WriteTo(Stream stream, StandardFormat format = default)
...
             if (!Changes.HasChanges)
             {
                 Write(stream, _original.Span);
                 stream.Flush();
                 return;
             }
 
-            WriteRootElementTo(writer);
+            Utf8JsonWriter writer = new(stream);
+            RootElement.WriteTo(writer);
```
jsquire, line 67: Does this automatically flush?   If not, should we consider doing so in order to keep behavior consistent with the `!Changes.HasChanges` path?

jsquire (approved):

sdk/core/Azure.Core.Experimental/src/MutableJsonElement.WriteTo.cs
```diff
@@ -0,0 +1,94 @@
...
+            }
+
+            originalElement.WriteTo(writer);
+        }
+
+        private void WriteObject(string path, int highWaterMark, JsonElement originalElement, Utf8JsonWriter writer)
+        {
+            writer.WriteStartObject();
+
+            foreach (JsonProperty property in originalElement.EnumerateObject())
```
KrzysztofCwalina, line 49: this enumerates original properties. What if a new property got added?

annelo-msft: Good catch!  Fixing...

annelo-msft: Oh wait, no, that was a copy and paste error in my test.  I think it does work, but let me come up with a clear explanation, and I can update the variable names to clarify as well.

annelo-msft: Ok, I've added tests for this [here](https://github.com/Azure/azure-sdk-for-net/pull/35204/files#diff-64c6d61f33cd5e11ec907e4f27f4b455c2434e70079d4f3f720d03c2e68dbe92R255) and [here](https://github.com/Azure/azure-sdk-for-net/pull/35204/files#diff-64c6d61f33cd5e11ec907e4f27f4b455c2434e70079d4f3f720d03c2e68dbe92R279).

The explanation for this is that when a property is added to an object, we [store a change to that object's path](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core.Experimental/src/MutableJsonElement.cs#L310-L317) with the new JsonElement holding the added property.  So, we enter [the first block in WriteElement](https://github.com/Azure/azure-sdk-for-net/pull/35204/files#diff-0b846adb4112f2d61c0ffb34cb025964feba66e862cd55e6e052e43f578dd256R19), and write out the entire property in the block.

I'll add some more tests with different sequencings of changes to make the test coverage a little more robust.

heaths (approved):
(Forgot to finish my review yesterday)

sdk/core/Azure.Core.Experimental/src/MutableJsonDocument.cs
```diff
@@ -56,15 +56,15 @@ public void WriteTo(Stream stream, StandardFormat format = default)
...
-            Utf8JsonWriter writer = new Utf8JsonWriter(stream);
             if (!Changes.HasChanges)
             {
                 Write(stream, _original.Span);
                 stream.Flush();
                 return;
             }
 
-            WriteRootElementTo(writer);
+            Utf8JsonWriter writer = new(stream);
```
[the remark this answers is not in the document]
annelo-msft: Oh awesome - thanks, Heath!

sdk/core/Azure.Core.Experimental/src/MutableJsonElement.WriteTo.cs
```diff
@@ -0,0 +1,94 @@
...
+{
+    public partial struct MutableJsonElement
+    {
+        internal void WriteTo(Utf8JsonWriter writer)
+        {
+            WriteElement(_path, _highWaterMark, _element, writer);
+            writer.Flush();
+        }
+
+        private void WriteElement(string path, int highWaterMark, JsonElement originalElement, Utf8JsonWriter writer)
```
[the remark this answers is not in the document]
annelo-msft: Ok, good.  You were right - this worked in the simplest case, but had a bug where we weren't processing descendant for a node that had direct changes.  I've updated this now, and simplified the implementation a bit.

sdk/core/Azure.Core.Experimental/src/MutableJsonDocument.cs
heaths: Now that `using` was added it will.

annelo-msft: It won't also flush the stream though, right?  I'll have a quick look at whether JsonElement.WriteTo() flushes to make sure we have parity in behavior here.

heaths: Not sure. The documentation doesn't say that, only that it flushes any remaining JSON scopes that may have been buffered.

heaths: Generally, though, you should always dispose a disposable. I'm surprise Roslyn Analyzers didn't flag that missing usage. I know I've seen those warnings before in our repo. Wonder if it was locally disabled e.g., for this project...and why.

annelo-msft: Yep, good call.  I know there are some differences in what's enforced in Experimental vs. Core, so I'm not too surprised.

annelo-msft: It looks like JsonDocument/JsonElement does not flush the writer themselves in WriteTo(), so I've removed places I'm doing that.  I think it makes sense that a perf-minded caller would want to own this themselves.

sdk/core/Azure.Core.Experimental/src/MutableJsonElement.WriteTo.cs
```diff
@@ -0,0 +1,73 @@
...
+        internal void WriteTo(Utf8JsonWriter writer)
+        {
+            WriteElement(_path, _highWaterMark, _element, writer);
+        }
+
+        private void WriteElement(string path, int highWaterMark, JsonElement element, Utf8JsonWriter writer)
+        {
+            if (Changes.TryGetChange(path, highWaterMark, out MutableJsonChange change))
+            {
+                element = change.AsJsonElement();
```
annelo-msft, line 20: We can potentially improve performance here by switching on the type of change.Value and writing the value directly to the writer if it is a primitive, e.g. doesn't have descendants.  This will save us a round-trip through JsonSerializer.Serialize() and JsonDocument.Parse().  Noting this optimization for a future PR.

PR #35210, opened 2023-03-29
Pass property name mapping to array, object enumerators

heaths:


sdk/core/Azure.Core.Experimental/src/DynamicJson.cs
```diff
@@ -100,7 +100,7 @@ private IEnumerable GetEnumerable()
         {
             return _element.ValueKind switch
             {
-                JsonValueKind.Array => new ArrayEnumerator(_element.EnumerateArray()),
+                JsonValueKind.Array => new ArrayEnumerator(_element.EnumerateArray(), _options),
```
annelo-msft, line 103: Would you be willing to add this to ObjectEnumerator as well?

sdk/core/Azure.Core.Experimental/tests/DynamicJsonTests.cs
```diff
@@ -225,6 +226,25 @@ public void CanSetArrayValueToNull()
             Assert.IsNull((int?)jsonData[0]);
         }
 
+        [Test]
+        public void CanEnumerateArrayWithProperCasing()
```
annelo-msft, line 230: Would you be willing to move this into the section of tests on casing?  (~line 478ish)

annelo-msft (commented):
Thanks for this, @heaths!

sdk/core/Azure.Core.Experimental/src/DynamicJson.cs
heaths: Possibly, but we should probably talk about the expected output. For example, should the `Key` property return the re-cased name or the underlying / original case? I was thinking about this last night and, upon reflection, am not entirely convinced what would be the right expectation. Initial thought was to transform, but if a caller is enumerating name/value pairs it's not necessarily in service of call site familiarity but likely enumerating the underlying data's name/value pairs.

annelo-msft: My instinct would be to make it a simple pass-through like the one you have on the `ArrayEnumerator`, e.g.:

public DynamicJsonProperty Current => new(_enumerator.Current.Name, new(_enumerator.Current.Value, **_cachedOptions**));

But I'm not visualizing a use case where this would be a problem - do you have one in mind, and if so, could you provide a sample that would behave unexpectedly so I can understand your thinking better?

heaths: Why pass it if we don't use it? With no use case, it seems like a waste of processing power and memory. If/when it's needed, it could be added then.

annelo-msft: I think we do use it, and not having it is probably a bug.

What I'm not understanding is, if we fix the bug by passing the options to the child value, what happens that you are concerned about?

heaths: `ObjectEnumerator` doesn't declare or use a casing scheme: https://github.com/Azure/azure-sdk-for-net/blob/a0a5af5b2679c13210fa428f7b7aca79a000753b/sdk/core/Azure.Core.Experimental/src/DynamicJson.ObjectEnumerator.cs#L42. In fact, it's a lightweight wrapper around a `DynamicJsonProperty` which also doesn't making use of any casing scheme: https://github.com/Azure/azure-sdk-for-net/blob/a0a5af5b2679c13210fa428f7b7aca79a000753b/sdk/core/Azure.Core.Experimental/src/DynamicJsonProperty.cs#L9, hence my confusion. Why would we pass something we don't? And even if we do end up using it, should we? Seems the `ObjectEnumerator` is for enumerating the property name/value pairs, and should those be transformed (re-cased) or left as-is? Aren't we just changing the case for properties to feel more C#-y? Enumerating raw property name/value pairs feels more like a "tell me what the data actually looks like" sort of operation, IMO.

annelo-msft: Passing in the options to the new DynamicJson() passed in as the value parameter to the DynamicJsonProperty constructor would fix the following bug (which you found through this work, and thank you for that!):

```csharp
            string json = """
                {
                    "first":
                    {
                        "one" : 1,
                        "two" : 2
                    },
                    "second": {
                        "one" : 1,
                        "two" : 2
                    }
                }
                """;

            dynamic value = new BinaryData(json).ToDynamic(DynamicJsonOptions.AzureDefault);

            foreach (dynamic property in value)
            {
                Assert.AreEqual(1, (int)property.Value.one);
                Assert.AreEqual(1, (int)property.Value.One);  // <-- This will throw will an NRE because the casing options are absent
                Assert.AreEqual(2, (int)property.Value.two);
                Assert.AreEqual(2, (int)property.Value.Two);  // <-- This will throw will an NRE because the casing options are absent
            }
```

heaths: I see - passing it all the way through to the `DynamicJson` created in the `Current` property. I missed that. One example where a simple `new(...)` leads to less readability. Wonder if we should avoid using that via .editorconfig 🤔

annelo-msft (approved):
Thank you so much for doing this, @heaths!

PR #35228, opened 2023-03-29
Use DynamicJson for CLU tests

heaths:


heaths:
This splits the CLU test updates out of #35202 so we can merge immediately. We may be shipping a new version of CLU and/or QA soon, so updating samples to use DynamicJson right now may not align with its own projected GA.

sdk/cognitivelanguage/Azure.AI.Language.Conversations/tests/ConversationAnalysisClientLiveTests.cs
```diff
@@ -20,152 +21,208 @@ public ConversationAnalysisClientLiveTests(bool isAsync, ConversationsClientOpti
...
 
             // deserialize
-            using JsonDocument json = await JsonDocument.ParseAsync(response.ContentStream);
-            ConversationalTaskResult conversationalTaskResult = ConversationalTaskResult.DeserializeConversationalTaskResult(json.RootElement);
+            dynamic conversationalTaskResult = response.Content.ToDynamic(DynamicJsonNameMapping.PascalCaseGetters);
             Assert.IsNotNull(conversationalTaskResult);
 
             // assert - prediction type
-            Assert.AreEqual(ProjectKind.Conversation, conversationalTaskResult.Result.Prediction.ProjectKind);
+            Assert.AreEqual("Conversation", conversationalTaskResult.Result.Prediction.ProjectKind?.ToString());
```
annelo-msft, line 53: Would you consider using a cast to string here?  Since you're comparing it to a string, it seems like the cast shouldn't fail.

sdk/cognitivelanguage/Azure.AI.Language.Conversations/tests/ConversationAnalysisClientLiveTests.cs
```diff
@@ -20,152 +21,208 @@ public ConversationAnalysisClientLiveTests(bool isAsync, ConversationsClientOpti
...
-            Response response = await Client.AnalyzeConversationAsync(conversationalTask.AsRequestContent());
+            Response response = await Client.AnalyzeConversationAsync(RequestContent.Create(data));
 
             // assert - main object
             Assert.IsNotNull(response);
 
             // deserialize
-            using JsonDocument json = await JsonDocument.ParseAsync(response.ContentStream);
-            ConversationalTaskResult conversationalTaskResult = ConversationalTaskResult.DeserializeConversationalTaskResult(json.RootElement);
+            dynamic conversationalTaskResult = response.Content.ToDynamic(DynamicJsonNameMapping.PascalCaseGetters);
```
annelo-msft, line 204:
```suggestion
            dynamic conversationalTaskResult = response.Content.ToDynamic(DynamicJsonOptions.AzureDefault);
```

annelo-msft (approved):

sdk/cognitivelanguage/Azure.AI.Language.Conversations/tests/ConversationAnalysisClientLiveTests.cs
heaths: I can do this, but I have a question/feedback: if the default - no parameter - is to use the original case, which I fully support because the re-casing is confusing, as we covered offline in the past, why do we have a value with "default" in the name that doesn't actually match the default behavior with no value specified? I get that it's "AzureDefault", but still seems potentially misleading.

heaths: Regardless of the name, I think the documentation for `AzureDefault` should reflect what it actually does. That's why I used `PascalCaseGetters`: it's explicit and obvoius what it does. I'm not personally a fan of using something that isn't explicit about its intent or is otherwise obvious.

heaths: I added this to the feedback thread. We can resume there.

PR #35236, opened 2023-03-30
Remove public-clients flag in synapse

### LINKED ISSUE #35109: Remove unsupported autogen flags from synapse SDKs and regenerate.
We are currently allowing an unsupported combination of flags in autorest when public-clients and generation1-convenience-client are both true. This combination is especially problematic when it comes to LRO since it exposes publicly the Start prefixed method names which we don't do any longer and doesn't use the WaitUntil parameter.

We need to make the following SDKs use only generation1-convenience-client and update LRO pattern for these SDKs.

 https://www.nuget.org/packages/Azure.Analytics.Synapse.Spark
 https://www.nuget.org/packages/Azure.Analytics.Synapse.Monitoring
 https://www.nuget.org/packages/Azure.Analytics.Synapse.ManagedPrivateEndpoints
 https://www.nuget.org/packages/Azure.Analytics.Synapse.Artifacts

pshao25:
Fix https://github.com/Azure/azure-sdk-for-net/issues/35109


sdk/synapse/Azure.Analytics.Synapse.Monitoring/src/autorest.md
```diff
@@ -11,9 +11,9 @@ require:
     - https://github.com/Azure/azure-rest-api-specs/blob/fc5e2fbcfc3f585d38bdb1c513ce1ad2c570cf3d/specification/synapse/data-plane/readme.md
 namespace: Azure.Analytics.Synapse.Monitoring
 generation1-convenience-client: true
-public-clients: true
 security: AADToken
 security-scopes: https://dev.azuresynapse.net/.default
+# use: D:/GIT/pshao25.csharp/artifacts/bin/AutoRest.CSharp/Debug/net6.0
```
wonner, line 16: Added by mistake?

pallavit:
Did you generate the SDK using the new autorest setting or did you do a manual rename of the files? I would have expected to see customization layer handholding as in this PR - https://github.com/Azure/azure-sdk-for-net/pull/35170

/cc: @m-nash

m-nash (commented):
Should move to `Custom` folder as a mirror to `Generated` folder.

PR #35246, opened 2023-03-30
Resolve PR feedback for #35228

heaths:
I had #35228 set to auto-merge and forgot to undo that as I was resolving PR feedback. This PR is to resolve that PR feedback from @annelo-msft


annelo-msft (approved):
Thanks!

PR #35278, opened 2023-03-31
[Core] Move DynamicJson into DynamicData

annelo-msft:
We have concerns that the explorations into extensibility (such as https://github.com/Azure/azure-sdk-for-net/pull/34810 and https://github.com/Azure/azure-sdk-for-net/pull/34774) may take some time to settle.  We believe we can postpone deciding on an approach by making the `DynamicData` type behave as only JSON for the time being, sealing it, and making constructors internal.

This PR implements this by removing the DynamicData base class,  and renaming DynamicJson and all related types to DynamicData.


sdk/core/Azure.Core.Experimental/api/Azure.Core.Experimental.net6.0.cs
```diff
@@ -111,85 +111,79 @@ namespace Azure.Core.Dynamic
...
         public static dynamic ToDynamic(this System.BinaryData data) { throw null; }
-        public static dynamic ToDynamic(this System.BinaryData data, Azure.Core.Dynamic.DynamicJsonNameMapping propertyNameCasing) { throw null; }
-        public static dynamic ToDynamic(this System.BinaryData data, Azure.Core.Dynamic.DynamicJsonOptions options) { throw null; }
-    }
-    public abstract partial class DynamicData
-    {
-        protected DynamicData() { }
-        internal abstract void WriteTo(System.IO.Stream stream);
-        public static void WriteTo(System.IO.Stream stream, Azure.Core.Dynamic.DynamicData data) { }
+        public static dynamic ToDynamic(this System.BinaryData data, Azure.Core.Dynamic.DynamicDataNameMapping propertyNameCasing) { throw null; }
```
KrzysztofCwalina, line 114: I wonder if these methods and the enum (parameter) are not JSON specific, i.e. maybe we want to call then ToDynamicFromJson and the parameter being as it was?

KrzysztofCwalina (approved):

sdk/core/Azure.Core.Experimental/api/Azure.Core.Experimental.net6.0.cs
christothes: In practice, they probably only apply to json but nothing about the API would restrict it to this. Is the thought that the more specific naming would help with discoverability?

annelo-msft: We can definitely be more explicit about this.

Supporting different options based on the data type (e.g. JSON, XML) makes sense - and I guess we don't need to substantially refactor today if the base class only supports JSON.  It will mean some options functionality exists only in subtypes and some in the base class - that makes sense.  The casing piece seems general across anything that has name/value properties - I guess we then copy those across options types and respect them where relevant, without making the assumption that all possible future formats will have the need for this option.

KrzysztofCwalina: If we think that the enum will scale to other formats, then we should keep the name generic. I was just observing that the rename to the main class does not necessitate renames to the extension methods and their parameters. These are independent.

sdk/core/Azure.Core.Experimental/src/DynamicDataOptions.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Azure.Core.Dynamic
+{
+    /// <summary>
+    /// Provides the ability for the user to define custom behavior when accessing data, such as JSON, through a dynamic layer.
+    /// </summary>
+    public struct DynamicDataOptions
+    {
+        /// <summary>
+        /// Gets the default <see cref="DynamicDataOptions"/> for Azure services.
```
heaths, line 12: Perhaps in a different PR, but could you clarify exactly what is considered "default" for Azure services? As a caller, I feel this is rather opaque and not likely to use it when I can be more specific with another overload.

sdk/core/Azure.Core.Experimental/api/Azure.Core.Experimental.net6.0.cs
KrzysztofCwalina: @annelo-msft, we would add a parameter to the method to select the format.

sdk/core/Azure.Core.Experimental/src/DynamicDataOptions.cs
annelo-msft: Sure! https://github.com/Azure/azure-sdk-for-net/issues/35281

sdk/core/Azure.Core.Experimental/api/Azure.Core.Experimental.net6.0.cs
```diff
@@ -110,86 +110,80 @@ namespace Azure.Core.Dynamic
...
+            public Azure.Core.Dynamic.DynamicData.ObjectEnumerator GetEnumerator() { throw null; }
             public bool MoveNext() { throw null; }
             public void Reset() { }
-            System.Collections.Generic.IEnumerator<Azure.Core.Dynamic.DynamicJsonProperty> System.Collections.Generic.IEnumerable<Azure.Core.Dynamic.DynamicJsonProperty>.GetEnumerator() { throw null; }
+            System.Collections.Generic.IEnumerator<Azure.Core.Dynamic.DynamicDataProperty> System.Collections.Generic.IEnumerable<Azure.Core.Dynamic.DynamicDataProperty>.GetEnumerator() { throw null; }
             System.Collections.IEnumerator System.Collections.IEnumerable.GetEnumerator() { throw null; }
         }
     }
-    public enum DynamicJsonNameMapping
+    public enum DynamicDataNameMapping
```
annelo-msft, line 166: I like keeping this one general because I think it would apply to any subtype/extension type.  Is there a format we might consider where values aren't keyed on property names?

sdk/core/Azure.Core.Experimental/src/DynamicData.ArrayEnumerator.cs
```diff
@@ -8,13 +8,13 @@
...
     {
         /// <summary>
         /// An enumerable and enumerator for the contents of a mutable JSON array.
         /// </summary>
         [DebuggerDisplay("{Current,nq}")]
-        public struct ArrayEnumerator : IEnumerable<DynamicJson>, IEnumerator<DynamicJson>
+        public struct ArrayEnumerator : IEnumerable<DynamicData>, IEnumerator<DynamicData>
         {
             private MutableJsonElement.ArrayEnumerator _enumerator;
             private readonly DynamicJsonOptions _options;
```
annelo-msft, line 20: It's a little awkward to hold a JSON-specific options here, but given this is all internal, I think it leaves space for extensibility.

sdk/core/Azure.Core.Experimental/src/DynamicData.cs
```diff
@@ -1,28 +1,202 @@
...
     /// <summary>
-    /// A dynamic abstraction over content data.  Deriving types are implemented
-    /// for a specific format, such as JSON or XML.
-    ///
-    /// This and related types are not intended to be mocked.
+    /// Dynamic layer over MutableJsonDocument.
     /// </summary>
-    public abstract class DynamicData
+    [DebuggerDisplay("{DebuggerDisplay,nq}")]
+    [JsonConverter(typeof(JsonConverter))]
```
annelo-msft, line 19: @KrzysztofCwalina, would we have any concerns with this type being JSON-serializable, in any of our extensibility scenarios?  If it were holding e.g. XML, I would expect JSON serialization to throw, but it doesn't feel like it precludes supporting serialization to other formats.

Thoughts?

sdk/core/Azure.Core.Experimental/src/DynamicData.ArrayEnumerator.cs
heaths: Though, it is `public`. Should probably be marked `private`. Enumerables in the .NET BCL almost always are. It only matters they implement `IEnumerable<T>` anyway.

annelo-msft: It is modeled after JsonElement's public [ArrayEnumerator](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsonelement.arrayenumerator?view=net-8.0), but you make a good point.  If everything else works, we can always make it public later if there is a reason.  Let me look at hiding it.  Thanks, @heaths!

sdk/core/Azure.Core.Experimental/api/Azure.Core.Experimental.net6.0.cs
annelo-msft: Updated.  Per this discussion, we have the option to keep a ToDynamic() method with a parameter that specifies the format, but this can be added at a later time.

sdk/core/Azure.Core.Experimental/src/DynamicData.ArrayEnumerator.cs
annelo-msft: Updated per your feedback, @heaths - thanks!

PR #35287, opened 2023-04-01
[AzureMonitorExporter] introduce platform abstraction

TimothyMothra:
## Background
Our exporter has a few instances where reading an Environment Variable or OS Name is necessary to make internal decisions.
We cannot reliably unit test these codeblocks as-is.

I want to introduce a platform abstraction to facilitate better unit testing. This is inspired by Application Insights' [IPlatform](https://github.com/microsoft/ApplicationInsights-dotnet/blob/main/BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/IPlatform.cs) and [PlatformImplementation](https://github.com/microsoft/ApplicationInsights-dotnet/blob/main/BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Platform/PlatformImplementation.cs)

## Changes

- New Interface `IPlatform`. 
  This abstracts EnvironmentVars and OS Name and facilitates more control over unit tests.
- New class `DefaultPlatform : IPlatform` which will be used at runtime. 
  This is initialized in the `TransmitterFactory`.
- New test class `MockPlatform : IPlatform` which will be used in unit tests. 
  Will add new tests in follow up PRs.

### DIFF
--- a/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TransmitterFactory.cs
+++ b/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TransmitterFactory.cs
@@ -2,7 +2,7 @@
 // Licensed under the MIT License.
 
 using System.Collections.Generic;
-using Azure.Core;
+using Azure.Monitor.OpenTelemetry.Exporter.Internals.Platform;
 
 namespace Azure.Monitor.OpenTelemetry.Exporter.Internals
 {
@@ -29,7 +29,7 @@ namespace Azure.Monitor.OpenTelemetry.Exporter.Internals
                 {
                     if (!_transmitters.TryGetValue(key, out transmitter))
                     {
-                        transmitter = new AzureMonitorTransmitter(azureMonitorExporterOptions);
+                        transmitter = new AzureMonitorTransmitter(azureMonitorExporterOptions, new DefaultPlatform());
 
                         _transmitters.Add(key, transmitter);
                     }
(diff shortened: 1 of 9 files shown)


sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TransmitterFactory.cs
```diff
@@ -29,7 +29,7 @@ public AzureMonitorTransmitter Get(AzureMonitorExporterOptions azureMonitorExpor
                 {
                     if (!_transmitters.TryGetValue(key, out transmitter))
                     {
-                        transmitter = new AzureMonitorTransmitter(azureMonitorExporterOptions);
+                        transmitter = new AzureMonitorTransmitter(azureMonitorExporterOptions, new DefaultPlatform());
```
vishweshbankwar, line 32: Any particular reason to have multiple instances of `DefaultPlatform` in case of multiple transmitters?

TimothyMothra: Can definitely change this to a static instance in the Factory. good call out!

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TransmitterFactory.cs
```diff
@@ -15,6 +15,7 @@ namespace Azure.Monitor.OpenTelemetry.Exporter.Internals
     internal class TransmitterFactory
     {
         public static TransmitterFactory Instance = new();
+        public static IPlatform platform = new DefaultPlatform();
```
rajkumar-rangaraj, line 18: Mark this as read-only.

rajkumar-rangaraj (approved):

vishweshbankwar (approved):

PR #35318, opened 2023-04-03
Update ACR TSG with info about retrying blob upload

### LINKED ISSUE #35253: [ACR] Update TSG with information about restarting upload when fails due to retry
See example in https://github.com/Azure/azure-sdk-for-java/pull/34277

annelo-msft:
Fixes https://github.com/Azure/azure-sdk-for-net/issues/35253

- Updates the Troubleshooting guide with a sample showing how to catch the specific `BLOB_UPLOAD_INVALID` exception
- Adds a custom error parser to the library to correctly write content to exception message, and set RequestFailedException.ErrorCode where applicable
- The parser buffers the content if the response content stream is unbuffered (e.g. on download), but will benefit from https://github.com/Azure/azure-sdk-for-net/issues/35355 when implemented.
- Per @pallavit suggestion, adds a link to the Troubleshooting guide to the exception text when the blob upload operation fails.


sdk/containerregistry/Azure.Containers.ContainerRegistry/TROUBLESHOOTING.md
```diff
@@ -160,4 +163,16 @@ The `CredentialUnavailableExcpetion` is a special exception type derived from `A
...
+
+When working with `ContainerRegistryContentClient`, you may get an `RequestFailedException` exception with
+message containing additional information and [Docker error code](https://docs.docker.com/registry/spec/api/#errors-2).
+
+### Getting BLOB_UPLOAD_INVALID
+
+In rare cases, a transient error (such as connection reset) can happen during blob upload which may lead to a `RequestFailedException` with a 404 status code being thrown with message similar to
+`{"errors":[{"code":"BLOB_UPLOAD_INVALID","message":"blob upload invalid"}]}`, and resulting in a failed upload.  In this case, upload should to be restarted from the beginning.
+
+To display detailed error messages, please see the code sample in [General troubleshooting](#general-troubleshooting) above.
```
lmolkova, line 178: nit: do we provide access to [`AcrErrors`](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/containerregistry/data-plane/Azure.ContainerRegistry/stable/2021-07-01/containerregistry.json#L2121) ? 

Assuming I'm a user and implementing restart logic in my custom code, I would not want to rely on exception message to retry call to service.
And also I don't want to retry all 404s from upload (e.g. if the repository does not exist), so is there a way to  get a code (`BLOB_UPLOAD_INVALID`) that users can retry upon?

here's what I did in JAva
```java
  if (ex.getCause() instanceof AcrErrorsException) {
      AcrErrorsException acrErrors = (AcrErrorsException) ex.getCause();
      for (AcrErrorInfo info : acrErrors.getValue().getErrors()) {
          System.out.printf("Uploaded blob failed: code '%s'\n", info.getCode());
      }
  }
```

lmolkova (approved):

sdk/containerregistry/Azure.Containers.ContainerRegistry/TROUBLESHOOTING.md
annelo-msft: .NET doesn't provide service-specific exceptions (that I know of), but you're correct that this is a rare instance where the caller would want to programmatically take action on the exception type.  I can update the sample to show how to retrieve the error code to check it in order to determine a retry case.

sdk/core/Azure.Core/src/Shared/ClientDiagnostics.cs
```diff
@@ -86,7 +86,7 @@ internal static HttpMessageSanitizer CreateMessageSanitizer(DiagnosticsOptions d
 
         public async ValueTask<RequestFailedException> CreateRequestFailedExceptionAsync(Response response, ResponseError? error = null, IDictionary<string, string>? additionalInfo = null, Exception? innerException = null)
         {
-            if (GetType() == typeof(ClientDiagnostics) && error is null && additionalInfo is null)
+            if (GetType() == typeof(ClientDiagnostics) && response.ContentStream is null && error is null && additionalInfo is null)
```
annelo-msft, line 89: @AlexanderSher, it looks like you had added this check in https://github.com/Azure/azure-sdk-for-net/pull/33001, so I want to make sure I'm not undoing what you intended here.

I'd like to add this check for response.ContentStream == null to address the case that the client is using an unbuffered network stream.  If it is, the generator isn't passing error or additionalInfo, so we will call the RFE constructor, but it will fail when it tries to read the content from the network stream.  This check ensures that if there is content to read, it is handled properly by the existing method in ClientDiagnostics.

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/ContainerRegistryRequestFailedDetailsParser.cs
```diff
@@ -0,0 +1,61 @@
...
+
+            BinaryData content = BinaryData.FromStream(response.ContentStream);
+            if (content.ToMemory().Length == 0)
+            {
+                return false;
+            }
+
+            // Buffer the stream to log the content in the message
+            // See: https://github.com/Azure/azure-sdk-for-net/issues/35355
+            MemoryStream stream = new(content.ToArray());
```
AlexanderSher, line 32: I think we can simply extract this part of the `ResponseBodyPolicy` into helper method: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/Pipeline/Internal/ResponseBodyPolicy.cs#L78-L110

Something like
```cs
internal static async ValueTask BufferResponse(Response response, bool async, CancellationTokenSource cts)
{
    var responseContentStream = message.Response.ContentStream;
    if (responseContentStream is null or MemoryStream || responseContentStream is MemoryStream)
    {
        return;
    }
    try
    {
        var bufferedStream = new MemoryStream();
        if (async)
        {
            await CopyToAsync(responseContentStream, bufferedStream, cts).ConfigureAwait(false);
        }
        else
        {
            CopyTo(responseContentStream, bufferedStream, cts);
        }

        responseContentStream.Dispose();
        bufferedStream.Position = 0;
        message.Response.ContentStream = bufferedStream;
    }
    // We dispose stream on timeout or user cancellation so catch and check if cancellation token was cancelled
    catch (Exception ex)
        when (ex is ObjectDisposedException
                  or IOException
                  or OperationCanceledException
                  or NotSupportedException)
    {
        ThrowIfCancellationRequestedOrTimeout(oldToken, cts.Token, ex, networkTimeout);
        throw;
    }
}
```

AlexanderSher: And then call it here with `BufferResponse(response, false, new CancellationTokenSource()).EnsureCompleted()`;

sdk/containerregistry/Azure.Containers.ContainerRegistry/src/ContainerRegistryRequestFailedDetailsParser.cs
```diff
@@ -0,0 +1,61 @@
...
+
+            try
+            {
+                // Optimistic check for JSON error object
+                if (content.ToMemory().Span[0] != (byte)'{')
+                {
+                    return false;
+                }
+
+                JsonDocument doc = JsonDocument.Parse(content);
```
AlexanderSher, line 43: `using JsonDocument doc = JsonDocument.Parse(content);`

sdk/containerregistry/Azure.Containers.ContainerRegistry/TROUBLESHOOTING.md
annelo-msft: Discussed offline.

lmolkova (approved):

Issue #22219, opened 2021-06-26
RecordMatcher fails with doubles on different runtime versions


heaths:
I have some doubles in both requests and reponses, and the `RecordMatcher` has a problem with doubles in requests on .NET 5.0 vs older versions. For example, here's what I get in a test that was recorded with .NET 5.0 but used on older runtimes:

```
  Message: 
    Azure.Core.TestFramework.TestRecordingMismatchException : Unable to find a record for the request POST https://wuppe.api.cognitive.microsoft.com/language/:query-knowledgebases?projectName=190a9e13-8ede-4e4b-a8fd-c4d7f2aeab6c&deploymentName=test&api-version=2021-05-01-preview
    Header differences:
        <Content-Length> values differ, request <360>, record <328>
    Body differences:
    Request and response bodies do not match at index 109:
         request: "shold":0.20000000000000001,"co"
         record:  "shold":0.2,"context":{"previou"
```

I can disable body comparisons, but I'm wondering if we could have the recording framework somehow inject a double converter to control precision.

For example, I believe Newtonsoft.Json uses `double.ToString("R")` so maybe we could as well, or even pass something like "f8" to always emit 8 decimal places (or whatever number) unconditionally.

heaths:
See https://github.com/dotnet/runtime/issues/435 for some discussion about this.

pakrym:
The recording framework is not the one that is serializing the request so it can't control the number formatting. Unfortunately this is not the first time we hit this issues so we do need a fix.

One possibility might be a semantic JSON comparer instead of bytes comparer.

heaths:
Another possibility might be something conceptually similar to what they did [here](https://github.com/dotnet/runtime/pull/35042/files#diff-a31e2a8eea664a0b94fff01d8bc216318a37636f42aca04c4dc43c23780654cfR7740-R7752): if we can safely (and we should probably make this behavior optional to mitigate risk) determine we are writing a double, we could precision to some degree as long as rounding works in our favor. While this change ignores ".0", we could use stateful comparisons to limit precision to 4 decimals or so. The only problem might be if on one machine we had a number like .3542 while on another .35419001. We should be able to mitigate that by handling rounding ourselves to whatever precision we define. Perhaps the degree of precision is an input to even enable this hypothetical encoder.

JoshLove-msft:
[...]

PR #35320, opened 2023-04-03
Fix some Cognitive Language test errors

heaths:
Fixes errors introduced in #35228

### DIFF
--- a/sdk/cognitivelanguage/Azure.AI.Language.Conversations/tests/ConversationAnalysisClientLiveTests.cs
+++ b/sdk/cognitivelanguage/Azure.AI.Language.Conversations/tests/ConversationAnalysisClientLiveTests.cs
@@ -34,7 +34,7 @@ namespace Azure.AI.Language.Conversations.Tests
                 },
                 parameters = new
                 {
-                    projectNam = TestEnvironment.ProjectName,
+                    projectName = TestEnvironment.ProjectName,
                     deploymentName = TestEnvironment.DeploymentName,
                 },
                 kind = "Conversation",
@@ -80,7 +80,7 @@ namespace Azure.AI.Language.Conversations.Tests
                 },
                 parameters = new
                 {
-                    projectNam = TestEnvironment.OrchestrationProjectName,
+                    projectName = TestEnvironment.OrchestrationProjectName,
                     deploymentName = TestEnvironment.OrchestrationDeploymentName,
                 },
                 kind = "Conversation",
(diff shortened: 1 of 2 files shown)


heaths:
@annelo-msft @KrzysztofCwalina this is a good example of what not having models might lead to. In part, though, it's because we don't validate inputs for our test framework and I had a typo I copied/pasted in several tests. The other issue was that an object was nested one level too deep - all things that models over anonymous types or dynamic types would've prevented.

annelo-msft (approved):
Thanks, @heaths, it's a good call-out, and I think a known risk of hand-authoring JSON (i.e. not using compiled model types) using any method.

I'm curious, did these tests pass before?  How did you discover the issue?

heaths:
> Thanks, @heaths, it's a good call-out, and I think a known risk of hand-authoring JSON (i.e. not using compiled model types) using any method.
> 
> I'm curious, did these tests pass before? How did you discover the issue?

The recorded tests pass because we don't validate requests - only responses. The live tests failed because I missed an "e" on the end of `projectName` and misplaced the `task` object one level too deep on another request.

pallavit (approved):

annelo-msft:
> The recorded tests pass because we don't validate requests - only responses. The live tests failed because I missed an "e" on the end of `projectName` and misplaced the `task` object one level too deep on another request.

That's good feedback.  We may want to recommend in our docs that end-to-end testing is a best practice to avoid this kind of error.  I'll make a note on the docs PR.

heaths:
> That's good feedback. We may want to recommend in our docs that end-to-end testing is a best practice to avoid this kind of error. I'll make a note on the docs PR.

We don't currently recommend they have end-to-end testing? Or do you mean that they run live tests prior to submitting initial tests? To record they would've had to. This is a case where we don't check the request like we do the responses. Perhaps we could by recording our own byte count for a quick test and drill into body matching for requests. /cc @JoshLove-msft @christothes

JoshLove-msft:
> Perhaps we could by recording our own byte count for a quick test and drill into body matching for requests. /cc @JoshLove-msft @christothes

Can you clarify why we are not matching requests here? Is it because it is disabled for this specific library? Request body matching is on by default.

JoshLove-msft:
> > Perhaps we could by recording our own byte count for a quick test and drill into body matching for requests. /cc @JoshLove-msft @christothes
> 
> Can you clarify why we are not matching requests here? Is it because it is disabled for this specific library? Request body matching is on by default.

Here is the [recommended workaround](https://github.com/Azure/azure-sdk-for-net/issues/22219#issuecomment-1267307531) - we should probably add this to the Test Framework docs.

PR #35324, opened 2023-04-04
[FormRecognizer] Disabling model factory generation

kinelski:
Since both factories in Form Recognizer are written manually we're disabling model factory generation for them.
- The methods in the old `FormRecognizerModelFactory` do not follow the same pattern as generated factory methods, so they need to be written manually (the factory generation feature was only added to CodeGen way after we had a factory in FR).
- Methods in the new `DocumentAnalysisModelFactory` do follow the same pattern as generated factory methods, but they're always generated in the `FormRecognizerModelFactory` class.


sdk/formrecognizer/Azure.AI.FormRecognizer/src/FormRecognizerClient/FormRecognizerModelFactory.cs
```diff
@@ -505,338 +503,5 @@ public static TrainingDocumentInfo TrainingDocumentInfo(string name, int pageCou
         {
             return new TrainingDocumentInfo(name, pageCount, errors?.ToList(), status, modelId);
         }
-
-        #region generated methods
```
kinelski, line 509: These methods we're removing existed only to hide autogenerated members that were placed in the `FormRecognizerModelFactory`. They correspond to models from the new API, but were being added to the old factory.

pallavit:
> Methods in the new DocumentAnalysisModelFactory do follow the same pattern as generated factory methods, but they're always generated in the FormRecognizerModelFactory class.

Is there a way to rename the model factory class for the new DocumentAnalysisModelFactory? I do not have any concerns with this change but wanted to see given the same method pattern if we could use generated code as much as possible.

kinelski:
> We can't remove the attribute from the old factory otherwise ApiCompat fails.

Thinking about it now, that seems a bit weird. Those CodeGen attributes are internal so it shouldn't interfere with the public API. I'll double check with Ted and, if it's not an issue, I'll try suppressing the error locally.

pallavit:
> > We can't remove the attribute from the old factory otherwise ApiCompat fails.
> 
> Thinking about it now, that seems a bit weird. Those CodeGen attributes are internal so they shouldn't interfere in the public API. I'll double check with Ted and, if it's not an issue, I'll try suppressing the error locally.

Sounds good. Do not spend too many cycles on it though. I do not have concerns with the change but as I said earlier if we can take benefit of auto generated code - the more the better :)

kinelski:
> Sounds good. Do not spend too many cycles on it though. I do not have concerns with the change but as I said earlier if we can take benefit of auto generated code - the more the better :)

I double checked with Ted and removing the attribute should not be considered a breaking change, so we may need to suppress the ApiCompat error. It doesn't seem very straightforward so I'll just open an issue for now to track this work. We can merge the current changes just to make sure the next couple of PRs I'll push won't make CodeGen generate a bunch of factory methods we'll need to manually hide.

pallavit:
>  It doesn't seem very straightforward so I'll just open an issue for now to track this work. 

Sounds like a plan :)

joseharriaga (approved):

PR #35341, opened 2023-04-04
[AzureMonitorExporter] fix net7 nullables

TimothyMothra:
Follow up to #34013.

Temporarily adding net7.0 to our TargetFrameworks identified more nullables to be addressed.
These come from newer framework libraries that have more nullables in their APIs and stricter enforcement of existing nullables.


sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Customizations/Models/MetricsData.cs
```diff
@@ -29,7 +29,7 @@ public MetricsData(int version, Metric metric, MetricPoint metricPoint) : base(v
                 {
                     // Note: if Key exceeds MaxLength or if Value is null, the entire KVP will be dropped.
 
-                    Properties.Add(new KeyValuePair<string, string>(tag.Key, tag.Value.ToString().Truncate(SchemaConstants.MetricsData_Properties_MaxValueLength)));
```
TimothyMothra, line 32: `ToString()` now returns `string?`

sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/HttpPipelineHelper.cs
```diff
@@ -23,15 +24,24 @@ internal static class HttpPipelineHelper
 
         internal static int GetItemsAccepted(HttpMessage message)
         {
-            return GetTrackResponse(message).ItemsAccepted.GetValueOrDefault();
+            return TryGetTrackResponse(message, out var trackResponse)
+                ? trackResponse.ItemsAccepted.GetValueOrDefault()
+                : default;
         }
 
-        internal static TrackResponse GetTrackResponse(HttpMessage message)
```
TimothyMothra, line 29: A few methods in this class needed to be change to a TryGet pattern because the dependent libraries now specify that they can return null.

rajkumar-rangaraj (approved):

PR #35364, opened 2023-04-05
[App Configuration] Update snapshot API

ShivangiReja:
Here's the following changes in the PR:

* Removed `SnapshotName` property from SettingSelector type
* Added GetConfigurationSettingsForSnapshot method and related tests
* Updated TSG

### DIFF
--- a/sdk/appconfiguration/Azure.Data.AppConfiguration/TROUBLESHOOTING.md
+++ b/sdk/appconfiguration/Azure.Data.AppConfiguration/TROUBLESHOOTING.md
@@ -6,6 +6,7 @@ This troubleshooting guide contains instructions to diagnose issues encountered 
 
 * [General troubleshooting](#general-troubleshooting)
   * [Enable client logging](#enable-client-logging)
+  * [Limit issues](#limit-issues)
 * [Troubleshooting authentication issues](#troubleshooting-authentication-issues)
   * [AuthenticationFailedException](#authenticationfailedexception)
   * [CredentialUnavailableException](#credentialunavailableexception)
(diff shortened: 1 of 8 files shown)


sdk/appconfiguration/Azure.Data.AppConfiguration/TROUBLESHOOTING.md
```diff
@@ -109,6 +110,19 @@ ConfigurationClientOptions options = new ConfigurationClientOptions()
 ```
 
 For more detail on capturing Azure SDK logs and additional examples, please see the [Azure SDK diagnostics sample](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/samples/Diagnostics.md).
+
+### Limit Issues
+
+The error thrown by the App Configuration client library includes a detailed response error object that provides specific useful insights into what went wrong and includes corrective actions to fix common issues. A common error encountered is HTTP status code 429 for exceeding the limit. Refer to the body of the 429 response for the specific reason why the request failed. The failures often happen under these circumstances:
```
jsquire, line 116: Do we have any docs that we could link to that describe the rate limit?

jsquire (approved):

sdk/appconfiguration/Azure.Data.AppConfiguration/TROUBLESHOOTING.md
ShivangiReja: I found these 2 docs:
* https://learn.microsoft.com/en-us/azure/azure-app-configuration/faq#are-there-any-limits-on-the-number-of-requests-made-to-app-configuration
* https://azure.microsoft.com/en-us/updates/general-availability-azure-app-configuration-now-offers-increased-hourly-request-limit/

Which one do you think we should link here? I think FAQ one is linked with AppConfig docs, so we can use it in our TSG.

jsquire: oohh... the first one!

ShivangiReja: Updated!

PR #35369, opened 2023-04-05
[Event Hubs Trigger] MinBatch WaitTime Adjustments

jsquire:
# Summary

The focus of these changes is to guard against the scenario where partition ownership changes while the trigger is waiting to build a batch of the requested size.  Because the new owner has no insight into how long the previous owner has already waited, the wait time will restart and obey the configured `MaxWaitTime`.  This could result in a batch for the partition not being dispatched in a timely manner.   In the worst case, where ownership of the partition changes multiple times, this could lead to a batch for the partition appearing to be severely delayed or to never be dispatched to the Function.

To guard against this, the trigger will now consider the time that the most recently read checkpoint was created and prefer dispatching a partial batch when the partition reads events for the first time and the checkpoint is older than the configured `MaxWaitTime` or when no checkpoint was found.

### DIFF
--- a/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubListener.PartitionProcessor.cs
+++ b/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubListener.PartitionProcessor.cs
@@ -15,6 +15,7 @@ using Microsoft.Azure.WebJobs.Host.Executors;
 using Microsoft.Extensions.Logging;
 using Newtonsoft.Json;
 using Azure.Core.Diagnostics;
+using System.Diagnostics;
 
 namespace Microsoft.Azure.WebJobs.EventHubs.Listeners
 {
@@ -27,18 +28,20 @@ namespace Microsoft.Azure.WebJobs.EventHubs.Listeners
         /// </summary>
         internal class PartitionProcessor : IEventProcessor, IDisposable
         {
+            private readonly CancellationTokenSource _cts = new();
+
             private readonly ITriggeredFunctionExecutor _executor;
             private readonly bool _singleDispatch;
             private readonly ILogger _logger;
-            private readonly CancellationTokenSource _cts = new();
             private readonly int _batchCheckpointFrequency;
             private int _batchCounter;
+            private bool _firstFunctionExecute;
             private bool _minimumBatchesEnabled;
             private bool _disposed;
-            private Task _cachedEventsBackgroundTask;
             private TimeSpan _maxWaitTime;
             private ValueStopwatch _currentCycle;
             private EventProcessorHostPartition _mostRecentPartitionContext;
+            private Task _cachedEventsBackgroundTask;
             private CancellationTokenSource _cachedEventsBackgroundTaskCts;
             private SemaphoreSlim _cachedEventsGuard;
 
@@ -53,17 +56,23 @@ namespace Microsoft.Azure.WebJobs.EventHubs.Listeners
                 _singleDispatch = singleDispatch;
                 _batchCheckpointFrequency = options.BatchCheckpointFrequency;
                 _logger = logger;
-                _minimumBatchesEnabled = options.MinEventBatchSize > 1; // 1 is the default
-                CachedEventsManager = new PartitionProcessorEventsManager(maxBatchSize: options.MaxEventBatchSize, minBatchSize: options.MinEventBatchSize);
+                _firstFunctionExecute = true;
                 _maxWaitTime = options.MaxWaitTime;
-                _cachedEventsGuard = new SemaphoreSlim(1, 1);
+                _minimumBatchesEnabled = options.MinEventBatchSize > 1; // 1 is the default
+
+                // Events are only cached when building a batch of minimum size.
+                if (_minimumBatchesEnabled)
+                {
+                    _cachedEventsGuard = new SemaphoreSlim(1, 1);
+                    CachedEventsManager = new PartitionProcessorEventsManager(maxBatchSize: options.MaxEventBatchSize, minBatchSize: options.MinEventBatchSize);
+                }
             }
 
             public Task CloseAsync(EventProcessorHostPartition context, ProcessingStoppedReason reason)
             {
                 // signal cancellation for any in progress executions and clear the cached events
                 _cts.Cancel();
-                CachedEventsManager.ClearEventCache();
+                CachedEventsManager?.ClearEventCache();
 
                 _logger.LogDebug(GetOperationDetails(context, $"CloseAsync, {reason}"));
                 return Task.CompletedTask;
@@ -127,6 +136,7 @@ namespace Microsoft.Azure.WebJobs.EventHubs.Listeners
                         };
 
                         await _executor.TryExecuteAsync(input, linkedCts.Token).ConfigureAwait(false);
+                        _firstFunctionExecute = false;
                         eventToCheckpoint = events[i];
                     }
                 }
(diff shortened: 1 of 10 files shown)


sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubListener.PartitionProcessor.cs
```diff
@@ -287,19 +304,52 @@ private async Task MonitorCachedEvents(CancellationTokenSource backgroundCancell
...
+                {
+                    return TimeSpan.Zero;
+                }
+
+                var timeSinceLastCheckpoint = DateTimeOffset.UtcNow.Subtract(lastCheckpointTime.Value);
+
+                // If the checkpoint was last recorded more than the 1.5 times configured wait time ago, ownership
+                // of the partition likely migrated and the function was not invoked on the expected schedule.
+                // Dispatch any available events immediately, as the wait time is presumed to have already elapsed.
+                if (timeSinceLastCheckpoint.TotalMilliseconds >= (configuredMaxWaitTime.TotalMilliseconds * 1.5))
```
m-redding, line 327: Just curious why 1.5* the wait time instead of the wait time?

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubListener.PartitionProcessor.cs
```diff
@@ -227,19 +239,24 @@ private async Task TriggerExecute(EventData[] events, EventProcessorHostPartitio
...
+            private async Task MonitorCachedEvents(DateTimeOffset? lastCheckpointTime, CancellationTokenSource backgroundCancellationTokenSource)
             {
+                Debug.Assert(_minimumBatchesEnabled, "Monitoring of cached events should only take place when minimum batches are enabled.");
+
                 var acquiredSemaphore = false;
+                var adjustedMaxWaitTime = GetAdjustedMaxWaitTime(_maxWaitTime, lastCheckpointTime, _firstFunctionExecute);
+
                 try
                 {
                     _currentCycle = ValueStopwatch.StartNew();
```
m-redding, line 254: Just a thought but could it be helpful to start the timer at the time since the last checkpoint if it's less than the max wait time? If we don't, we may still run into the case where we wait .9999\*maxwaittime and then migrate and since we're only checking time since checkpoint on the first invocation, we could still end up waiting 1.999*maxwaittime? Unless I'm thinking about that wrong

m-redding (approved):
I asked a couple of questions but otherwise it LGTM

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubListener.PartitionProcessor.cs
jsquire: This came out of our conversation.  The concern was that we may be too reactive when checkpointing is configured to be infrequent.   We had talked about doubling the max wait time, but that seemed like it may be too large - so I split the difference.  

Feeling different about padding the wait time now?  Should we consider snapping back to just the wait time?

m-redding: Ohh ok, I must have explained my thought wrong I was thinking
- if > max wait time but less than 2maxwait time then invoke immediately
- if > 2maxwaittime don't invoke immediately if less than minbatchsize, just start the timer and behave normally

I figured that a migration would take less than a full maxwaittime in most cases, so it would hopefully eliminate the case where a new app is starting up after awhile and the first batch is just small

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubListener.PartitionProcessor.cs
jsquire: Not sure that I'm following your scenario.   To make it concrete, let's say that we have a wait time of 60 seconds and we read a checkpoint from 30 seconds ago.  Since checkpointing takes place after a batch is processed, we can infer that a function was invoked 30 seconds ago.  If we wanted to adjust to honor the wait time, we'd adjust the wait time to be 30 seconds.   That's what you're describing, right?

I actually had logic in for that at one point, but I ended up removing it to reduce complexity.   I've got the wait time documented to indicate that you may see 2 * wait time as when ownership changes.  Think it's important and that we should add that back in?

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubListener.PartitionProcessor.cs
jsquire: The concern that I have with the second clause there is that we don't know how frequently an application is checkpointing.  The migration may be fresh but the old owner is checkpointing every 10 batches.  Generally, I think we're better dispatching a partial batch for the first execution, since we know that it's a one-time thing and we'll be able to make more accurate decisions starting with the second batch.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubListener.PartitionProcessor.cs
m-redding: Yeah that's what I was describing! And ok that makes sense. Having a maximum of 2maxwaittime for ownership changes sounds good to me since hopefully that perfectly timed migration won't be a very common issue.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubListener.PartitionProcessor.cs
m-redding: Yeah that makes sense, my thought was to try and avoid having the first small batch hit every application, but having the documentation right in the property that you're likely to get a small batch in the first invocation is a good compromise.

jsquire: ok.  I'm going to bump this back to the configured wait time and remove the multiplier while ignoring the second clause.

sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubListener.PartitionProcessor.cs
jsquire: I added this back in for discussion.   Think we should keep it or pull it back out?

m-redding: I think I like it more this way personally, since it seems like checkpointfrequency=1 is the most common case, so preventing the potential 2maxwaittime wait seems like a positive feature.

jsquire: Alrighty.  We'll keep it then.

PR #35371, opened 2023-04-05
Diagnostic Scope improvements

sourabh1007:
Making following improvements:

- Use of ConcurrentDict in DiagScope - Can be improved by passing the ActivitySource from DiagScopeFactory and giving ownership of ActivitySource to it.

- Change Scope state on the basis of activity state, to support parent-based sampling

### DIFF
--- a/sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
+++ b/sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
@@ -5,7 +5,6 @@
 
 using System;
 using System.Collections;
-using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
@@ -21,15 +20,9 @@ namespace Azure.Core.Pipeline
         internal const string OpenTelemetrySchemaAttribute = "az.schema_url";
         internal const string OpenTelemetrySchemaVersion = "https://opentelemetry.io/schemas/1.17.0";
         private static readonly object AzureSdkScopeValue = bool.TrueString;
-        private static readonly ConcurrentDictionary<string, object?> ActivitySources = new();
 
         private readonly ActivityAdapter? _activityAdapter;
         private readonly bool _suppressNestedClientActivities;
-
-        internal DiagnosticScope(string ns, string scopeName, DiagnosticListener source, ActivityKind kind, bool suppressNestedClientActivities) :
-            this(scopeName, source, null, GetActivitySource(ns, scopeName), kind, suppressNestedClientActivities)
-        {
-        }
 
         internal DiagnosticScope(string scopeName, DiagnosticListener source, object? diagnosticSourceArgs, object? activitySource, ActivityKind kind, bool suppressNestedClientActivities)
         {
@@ -44,34 +37,15 @@ namespace Azure.Core.Pipeline
                 IsEnabled &= !AzureSdkScopeValue.Equals(Activity.Current?.GetCustomProperty(AzureSdkScopeLabel));
             }
 
-            _activityAdapter = IsEnabled ? new ActivityAdapter(activitySource, source, scopeName, kind, diagnosticSourceArgs) : null;
+            _activityAdapter = IsEnabled ? new ActivityAdapter(
+                                                    activitySource: activitySource,
+                                                    diagnosticSource: source,
+                                                    activityName: scopeName,
+                                                    kind: kind,
+                                                    diagnosticSourceArgs: diagnosticSourceArgs) : null;
         }
 
         public bool IsEnabled { get; }
-
-        /// <summary>
-        /// This method combines client namespace and operation name into an ActivitySource name and creates the activity source.
-        /// For example:
-        ///     ns: Azure.Storage.Blobs
-        ///     name: BlobClient.DownloadTo
-        ///     result Azure.Storage.Blobs.BlobClient
-        /// </summary>
-        private static object? GetActivitySource(string ns, string name)
-        {
-            if (!ActivityExtensions.SupportsActivitySource())
-            {
-                return null;
-            }
-
-            string clientName = ns;
-            int indexOfDot = name.IndexOf(".", StringComparison.OrdinalIgnoreCase);
-            if (indexOfDot != -1)
-            {
-                clientName += "." + name.Substring(0, indexOfDot);
-            }
-
-            return ActivitySources.GetOrAdd(clientName, static n => ActivityExtensions.CreateActivitySource(n));
-        }
 
         public void AddAttribute(string name, string value)
         {
@@ -355,7 +329,7 @@ namespace Azure.Core.Pipeline
 
             private Activity? StartActivitySourceActivity()
             {
-                return ActivityExtensions.ActivitySourceStartActivity(
+                Activity? activity = ActivityExtensions.ActivitySourceStartActivity(
                     _activitySource,
                     _activityName,
                     (int)_kind,
@@ -364,6 +338,13 @@ namespace Azure.Core.Pipeline
                     links: GetActivitySourceLinkCollection(),
                     traceparent: _traceparent,
                     tracestate: _tracestate);
+
+                if (activity == null || ((bool?)ActivityExtensions.GetCustomProperty(activity, "IsAllDataRequested") ?? false))
+                {
+                    return null;
+                }
+
+                return activity;
             }
 
             public void SetStartTime(DateTime startTime)
--- a/sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
+++ b/sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
@@ -14,7 +14,7 @@ using System.Reflection;
 
 namespace Azure.Core.Pipeline
 {
-    internal readonly struct DiagnosticScope : IDisposable
+    internal struct DiagnosticScope : IDisposable
     {
         private const string AzureSdkScopeLabel = "az.sdk.scope";
         internal const string OpenTelemetrySchemaAttribute = "az.schema_url";
@@ -45,7 +45,7 @@ namespace Azure.Core.Pipeline
                                                     diagnosticSourceArgs: diagnosticSourceArgs) : null;
         }
 
-        public bool IsEnabled { get; }
+        public bool IsEnabled { get; private set; }
 
         public void AddAttribute(string name, string value)
         {
@@ -89,6 +89,11 @@ namespace Azure.Core.Pipeline
         public void Start()
         {
             Activity? started = _activityAdapter?.Start();
+
+            if (started != null)
+            {
+                this.IsEnabled = IsEnabled && started.GetIsAllDataRequestedFlag();
+            }
             started?.SetCustomProperty(AzureSdkScopeLabel, AzureSdkScopeValue);
         }
 
@@ -338,12 +343,6 @@ namespace Azure.Core.Pipeline
                     links: GetActivitySourceLinkCollection(),
                     traceparent: _traceparent,
                     tracestate: _tracestate);
-
-                if (activity == null || ((bool?)ActivityExtensions.GetCustomProperty(activity, "IsAllDataRequested") ?? false))
-                {
-                    return null;
-                }
-
                 return activity;
             }
 
@@ -421,6 +420,7 @@ namespace Azure.Core.Pipeline
         private static Action<Activity, string?>? SetTraceStateStringMethod;
         private static Action<Activity, int, string?>? SetErrorStatusMethod;
         private static Func<Activity, int>? GetIdFormatMethod;
+        private static Func<Activity, bool>? GetAllDataRequestedMethod;
         private static Action<Activity, string, object?>? ActivityAddTagMethod;
         private static Func<object, string, int, object?, ICollection<KeyValuePair<string, object>>?, IList?, DateTimeOffset, Activity?>? ActivitySourceStartActivityMethod;
         private static Func<object, bool>? ActivitySourceHasListenersMethod;
@@ -540,6 +540,25 @@ namespace Azure.Core.Pipeline
             }
 
             return GetTraceStateStringMethod(activity);
+        }
+
+        public static bool GetIsAllDataRequestedFlag(this Activity activity)
+        {
+            if (GetAllDataRequestedMethod == null)
+            {
+                var method = typeof(Activity).GetProperty("IsAllDataRequested")?.GetMethod;
+                if (method == null)
+                {
+                    GetAllDataRequestedMethod = _ => true;
+                }
+                else
+                {
+                    GetAllDataRequestedMethod = Expression.Lambda<Func<Activity, bool>>(
+                        Expression.Call(ActivityParameter, method),
+                        ActivityParameter).Compile();
+                }
+            }
+            return GetAllDataRequestedMethod(activity);
         }
 
         public static void SetTraceState(this Activity activity, string? tracestate)
--- a/sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
+++ b/sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
@@ -14,7 +14,7 @@ using System.Reflection;
 
 namespace Azure.Core.Pipeline
 {
-    internal struct DiagnosticScope : IDisposable
+    internal readonly struct DiagnosticScope : IDisposable
     {
         private const string AzureSdkScopeLabel = "az.sdk.scope";
         internal const string OpenTelemetrySchemaAttribute = "az.schema_url";
@@ -45,7 +45,7 @@ namespace Azure.Core.Pipeline
                                                     diagnosticSourceArgs: diagnosticSourceArgs) : null;
         }
 
-        public bool IsEnabled { get; private set; }
+        public bool IsEnabled { get; }
 
         public void AddAttribute(string name, string value)
         {
@@ -89,11 +89,6 @@ namespace Azure.Core.Pipeline
         public void Start()
         {
             Activity? started = _activityAdapter?.Start();
-
-            if (started != null)
-            {
-                this.IsEnabled = IsEnabled && started.GetIsAllDataRequestedFlag();
-            }
             started?.SetCustomProperty(AzureSdkScopeLabel, AzureSdkScopeValue);
         }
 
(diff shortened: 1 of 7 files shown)


sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
```diff
@@ -364,6 +338,13 @@ public void AddLink(string traceparent, string? tracestate, IDictionary<string,
                     links: GetActivitySourceLinkCollection(),
                     traceparent: _traceparent,
                     tracestate: _tracestate);
+
+                if (activity == null || ((bool?)ActivityExtensions.GetCustomProperty(activity, "IsAllDataRequested") ?? false))
```
lmolkova, line 342: `IsAllDataRequested` is a flag on the activity.

`GetCustomProperty` method calls into [Activity.GetCustomProperty](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Activity.cs#L1046), what you need is to call [Activity.IsAllDataRequested](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Activity.cs#L837).

You'd have to add a new method that gets `IsAllDataRequested` - you can use `ActivitySourceHasListeners` as an example on how to do it with reflection+caching for property

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
```diff
@@ -115,6 +89,11 @@ public void AddLink(string traceparent, string? tracestate, IDictionary<string,
         public void Start()
         {
             Activity? started = _activityAdapter?.Start();
+
+            if (started != null)
+            {
+                this.IsEnabled = IsEnabled && started.GetIsAllDataRequestedFlag();
```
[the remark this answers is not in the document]
JoshLove-msft: Should we just set _currentActivity to null if IsAllDataRequested is false? 
/cc @lmolkova

sdk/core/Azure.Core/tests/Azure.Core.Tests.csproj
```diff
@@ -14,6 +14,7 @@
     <PackageReference Include="NUnit" />
     <PackageReference Include="NUnit3TestAdapter" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" />
+    <PackageReference Include="OpenTelemetry" VersionOverride="1.4.0" />
```
JoshLove-msft, line 17: Instead of using VersionOverride, add this to [package.data.props](https://github.com/Azure/azure-sdk-for-net/blob/main/eng/Packages.Data.props#L185) instead in the tests section.

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
```diff
@@ -15,22 +14,16 @@
 
 namespace Azure.Core.Pipeline
 {
-    internal readonly struct DiagnosticScope : IDisposable
+    internal struct DiagnosticScope : IDisposable
```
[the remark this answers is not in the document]
sourabh1007: because I want to change the `isEnabled `flag on the basis of activity status. And since it was `readonly`, it was not allowing me to do so.

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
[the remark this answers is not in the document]
sourabh1007: We don't lose the activities due to sampling when we create it using activity source, we just get activities with `IsAllDataRequested` flag off. Hence, I did the same thing here. Otherwise, I am fine with setting activities as null (it would make this struct `readonly `again)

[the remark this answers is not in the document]
JoshLove-msft: But IsEnabled doesn't do anything - it isn't used anywhere.

sdk/core/Azure.Core/tests/ClientDiagnosticsTests.Net50.cs
```diff
@@ -468,6 +470,38 @@ public void FailedStopsActivityAndWritesExceptionEventActivitySource()
             CollectionAssert.Contains(activity.Tags, new KeyValuePair<string, string>("Attribute2", "2"));
             CollectionAssert.Contains(activity.Tags, new KeyValuePair<string, string>("az.namespace", "Microsoft.Azure.Core.Cool.Tests"));
         }
+
+        [Test]
+        [NonParallelizable]
+        public void OpenTelemetryCompatibilityWithParentBasedSampling()
```
lmolkova, line 476: please add a deterministic test instead.

I.e use 
- `AlwaysOff` sampler and check that no activities were created.

OR 

- use parent-based sampler and manually set `00-valid_trace_id-valid_span_id-00` as parentId and expect it to be sampled out. 
 
?

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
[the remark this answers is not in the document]
JoshLove-msft: Yes we do! https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/Shared/MessagingClientDiagnostics.cs#L173

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
```diff
@@ -288,7 +262,15 @@ public void AddLink(string traceparent, string? tracestate, IDictionary<string,
                 _currentActivity = StartActivitySourceActivity();
                 if (_currentActivity != null)
                 {
-                    _currentActivity.AddTag(OpenTelemetrySchemaAttribute, OpenTelemetrySchemaVersion);
+                    if (!(_currentActivity?.GetIsAllDataRequestedFlag() ?? true))
```
lmolkova, line 265: instead of stopping and setting current activity to null (which will break context propagation we care about when activities are sampled out), can we instead either:
- add a private `IsAllDataRequested` flag to DiagScope and set it once activity is started
- or use `_currentActivity?.GetIsAllDataRequestedFlag() == true` every time

to guard any expensive operation that DiagScope does with it?

I can see just a couple of places that's needed:
- [MarkFailed](https://github.com/Azure/azure-sdk-for-net/blob/2800d6ecab44c45489a7868bfb6536fef9dc3620/sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs#L367)
- [AddTag](https://github.com/Azure/azure-sdk-for-net/blob/2800d6ecab44c45489a7868bfb6536fef9dc3620/sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs#L203)

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
```diff
@@ -288,7 +267,7 @@ public void AddLink(string traceparent, string? tracestate, IDictionary<string,
                 _currentActivity = StartActivitySourceActivity();
                 if (_currentActivity != null)
                 {
-                    _currentActivity.AddTag(OpenTelemetrySchemaAttribute, OpenTelemetrySchemaVersion);
+                    _currentActivity?.AddTag(OpenTelemetrySchemaAttribute, OpenTelemetrySchemaVersion);
```
lmolkova, line 270:
```suggestion
                    _isAllDataRequested = _currentActivity.GetIsAllDataRequestedFlag();
                    _currentActivity.AddTag(OpenTelemetrySchemaAttribute, OpenTelemetrySchemaVersion);
```

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
```diff
@@ -206,6 +180,8 @@ private class ActivityAdapter : IDisposable
             private string? _traceparent;
             private string? _tracestate;
 
+            private bool _isAllDataRequested;
```
lmolkova, line 183:
```suggestion
            private bool _isAllDataRequested = true;
```

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
```diff
@@ -348,14 +327,16 @@ public void AddLink(string traceparent, string? tracestate, IDictionary<string,
                     _currentActivity.Start();
                 }
 
+                _isAllDataRequested = _currentActivity?.GetIsAllDataRequestedFlag() ?? true;
```
lmolkova, line 330:
```suggestion
```

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
```diff
@@ -374,14 +356,17 @@ public void SetStartTime(DateTime startTime)
...
+                        _diagnosticSource?.Write(_activityName + ".Exception", exception);
+                    }
 
-                if (ActivityExtensions.SupportsActivitySource())
-                {
-                    _currentActivity?.SetErrorStatus(exception?.ToString());
+                    if (ActivityExtensions.SupportsActivitySource())
+                    {
+                        _currentActivity?.SetErrorStatus(exception?.ToString());
+                    }
```
lmolkova, line 369:
```suggestion
                if (exception != null)
                {
                    _diagnosticSource?.Write(_activityName + ".Exception", exception);
                }

                if (_isAllDataRequested && ActivityExtensions.SupportsActivitySource())
                {
                    _currentActivity?.SetErrorStatus(exception?.ToString());
                }
```

sdk/core/Azure.Core/tests/ClientDiagnosticsTests.Net50.cs
```diff
@@ -468,6 +470,109 @@ public void FailedStopsActivityAndWritesExceptionEventActivitySource()
...
+        [Test]
+        [NonParallelizable]
+        public void OpenTelemetryCompatibilityWithTraceIdRatioBasedSampler()
+        {
+            using var _ = SetAppConfigSwitch();
+
+            // Open Telemetry Listener
+            using TracerProvider OTelTracerProvider = Sdk.CreateTracerProviderBuilder()
+                .AddSource($"Azure.*")
+                .SetSampler(new TraceIdRatioBasedSampler(0.1))
```
lmolkova, line 483: is this test validating`TraceIdRatioBasedSampler` behavior? If there is some Azure SDK behavior that's not covered by other tests, can we write deterministic tests for it?

sdk/core/Azure.Core/tests/ClientDiagnosticsTests.Net50.cs
```diff
@@ -468,6 +470,109 @@ public void FailedStopsActivityAndWritesExceptionEventActivitySource()
...
+                .SetSampler(new AlwaysOffSampler())
+                .Build();
+
+            DiagnosticScopeFactory clientDiagnostics = new DiagnosticScopeFactory("Azure.Clients", "Microsoft.Azure.Core.Cool.Tests", true, true);
+
+            int activeActivityCounts = 0;
+            for (int i = 0; i < 100; i++)
+            {
+                DiagnosticScope scope = clientDiagnostics.CreateScope("ClientName.ActivityName");
+                scope.Start();
```
lmolkova, line 521:
```suggestion
                Assert.IsNull(scope.Start());
```

sdk/core/Azure.Core/tests/ClientDiagnosticsTests.Net50.cs
```diff
@@ -468,6 +470,109 @@ public void FailedStopsActivityAndWritesExceptionEventActivitySource()
...
+                if (Activity.Current.IsAllDataRequested)
+                    activeActivityCounts++;
+                scope.Dispose();
```
lmolkova, line 524:
```suggestion
                scope.Dispose();
                Assert.IsNull(Activity.Current);
```

sdk/core/Azure.Core/tests/ClientDiagnosticsTests.Net50.cs
```diff
@@ -468,6 +470,109 @@ public void FailedStopsActivityAndWritesExceptionEventActivitySource()
...
+
+            DiagnosticScopeFactory clientDiagnostics = new DiagnosticScopeFactory("Azure.Clients", "Microsoft.Azure.Core.Cool.Tests", true, true);
+
+            int activeActivityCounts = 0;
+            for (int i = 0; i < 5; i++)
+            {
+                DiagnosticScope scope = clientDiagnostics.CreateScope("ClientName.ActivityName");
+                scope.AddLink($"00-6e76af18746bae4eadc3581338bbe8b{i}-2899ebfdbdce904b-00", "foo=bar");
+
+                scope.Start();
```
lmolkova, line 552:
```suggestion
                Assert.IsNull(scope.Start());
```

sdk/core/Azure.Core/tests/ClientDiagnosticsTests.Net50.cs
sourabh1007: Added 2 more tests as you suggested in prev comments.

sdk/core/Azure.Core/tests/ClientDiagnosticsTests.Net50.cs
sourabh1007: Start doesn't return anything...why are we asserting is to null? or I am missing something here?

sdk/core/Azure.Core/tests/ClientDiagnosticsTests.Net50.cs
sourabh1007: Start doesn't return anything...why are we asserting is to null? or I am missing something here?

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
JoshLove-msft: Just curious - is there a reason to prefer adding the IsAllDataRequested guards to each method vs setting to null? It seems like this will become easy to forget as we add new methods in the future.

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
lmolkova: in the line 270, right after starting activity.The it'd apply to Activity source path and would immediately prevent `_currentActivity.AddTag(OpenTelemetrySchemaAttribute, OpenTelemetrySchemaVersion);` from adding any overhead

sdk/core/Azure.Core/tests/ClientDiagnosticsTests.Net50.cs
lmolkova: is this test still needed? does it test anything related to Azure SDK behavior?

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
lmolkova: I don't have any preference, as long as we preserve Activity.Current and don't forget to close started one at the end. Agree that it might be easier to do with `_sampledOutActivity`

sourabh1007: assigning to new object and having an extra logic in dispose to handle it, sounds like an overhead. In case my test I can add more assertions to make sure we are not adding Tags in sampled out activities.

sdk/core/Azure.Core/tests/ClientDiagnosticsTests.Net50.cs
sourabh1007: I thought having one extra test to cover one more kind of sampler, should not be a problem as right now diagnostic scope has zero coverage for sampling. Anyway, I am removing it as you said it is not testing anything related to Azure SDK behaviour

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
JoshLove-msft: More tests are always good, but I was thinking more about new methods that we may add in the future to DiagnosticScope where we will need to remember to add this check.

sourabh1007: if we make sure that new methods are working with sampling then, we won't miss this check.

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
```diff
@@ -267,6 +266,17 @@ public void AddLink(string traceparent, string? tracestate, IDictionary<string,
                 _currentActivity = StartActivitySourceActivity();
                 if (_currentActivity != null)
                 {
+                    _isAllDataRequested = _currentActivity.GetIsAllDataRequestedFlag();
+
+                    if (!_isAllDataRequested)
+                    {
+                        _sampleOutActivity = _currentActivity;
+                        _currentActivity.Stop();
```
JoshLove-msft, line 274: We shouldn't stop the current activity.

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
```diff
@@ -396,6 +401,7 @@ public void Dispose()
                 {
                     _currentActivity.Stop();
                 }
+                _sampleOutActivity?.Stop();
```
JoshLove-msft, line 404: This whole method needs to be updated to work off of either activity, e.g.:
```c#
var activity = _currentActivity ?? _sampledOutActivity;
if (activity == null)
{
  return;
}
...
```

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
JoshLove-msft: Are tests passing currently? Ideally we would have a test that would fail with this change.

sourabh1007: yes..client diagnostics tests are passing with current change

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
```diff
@@ -288,6 +266,16 @@ public void AddLink(string traceparent, string? tracestate, IDictionary<string,
                 _currentActivity = StartActivitySourceActivity();
                 if (_currentActivity != null)
                 {
+                    _isAllDataRequested = _currentActivity.GetIsAllDataRequestedFlag();
```
JoshLove-msft, line 269: This class variable is not needed. We can just use the return of GetIsAllDataRequestedFlag directly.

sdk/core/Azure.Core/src/Shared/DiagnosticScope.cs
```diff
@@ -561,6 +552,25 @@ public static bool IsW3CFormat(this Activity activity)
             return GetTraceStateStringMethod(activity);
         }
 
+        public static bool GetIsAllDataRequestedFlag(this Activity activity)
```
JoshLove-msft, line 555:
```suggestion
        public static bool GetIsAllDataRequested(this Activity activity)
```

JoshLove-msft:
Looks great! Thanks @sourabh1007!

JoshLove-msft (approved):

lmolkova (approved):
LGTM, thank you!!

PR #35376, opened 2023-04-05
Update Microsoft.CodeAnalysis.NetAnalyzers to 7.0.1

christothes:


pallavit:
Curious what motivated this change?

christothes:
> Curious what motivated this change?

I only became aware of the newer released version because OmniSharp logs the following warning

```
[warn]: OmniSharp.MSBuild.ProjectLoader
        The .NET SDK has newer analyzers with version '7.0.1' than what version '7.0.0' of 'Microsoft.CodeAnalysis.NetAnalyzers' package provides. Update or remove this package reference.
```

heaths (approved):

heaths:
> @heaths had to revert an older version because of a regression. AFAIK it was fixed but just to double check :)

Thanks for looping me in. That issue to which you refer was only because we used a property that conflicted with one that newer package uses, but I fixed that. At the time, I believe 7.0.0 was the latest, so this change should be fine and seems to be passing so far.

@christothes just to make sure, though, if you load a project in VS, can you double check you don't see it constantly reloading packages again and again? My change didn't necessitate any changes to the package itself, but we might want to double check this one last time just to be sure.

christothes:
VS seems happy with the change 😄

heaths:
This should fix https://github.com/Azure/autorest.csharp/issues/3261 as well, since https://github.com/dotnet/roslyn-analyzers/pull/6114 fixed this in the newer NetAnalyzers. While my attempt to fix this in https://github.com/Azure/autorest.csharp/pull/3262 is still a valid solution, it has a non-obvious bug tail that is probably not worth the time.

However, we should consider attributing codegen attributes with a `ConditionalAttribute` so they don't end up in production and the (albeit it small) perf hit for anything enumerating attributes won't even hit these attributes. They aren't needed beyond code gen. I had talked with Pavel about this ages ago and the only issue was that we'd need to pass a symbol e.g., `CODEGEN`, in all areas when we generate code. @m-nash wouldn't the best place be in the generator itself such that anything that calls it by any means gets it for free?

heaths:
I opened https://github.com/Azure/autorest.csharp/issues/3321 to track.

PR #35382, opened 2023-04-05
Adding BlobContainerClient data movement extensions

schaabs:
- Adds directory upload and download extension methods to `BlobContainerClient`.
- Sample usage added to DatMovement.Blobs readme
- Still working on recorded tests, they will be added to the PR shortly.


sdk/storage/Azure.Storage.DataMovement.Blobs/src/BlobContainerClientExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+        /// <param name="client">The <see cref="BlobContainerClient"/> used for service operations.</param>
+        /// <param name="localDirectoryPath">The full path to the local directory where files will be dowloaded.</param>
+        /// <param name="blobDirectoryPrefix">Optionally restricts the downloaded content to blobs with the specified directory prefix.</param>
+        /// <param name="options">Options which control the container download.</param>
+        /// <returns></returns>
+        public static async Task<DataTransfer> StartDownloadDirectoryAsync(this BlobContainerClient client, string localDirectoryPath, string blobDirectoryPrefix = default, BlobContainerClientTransferOptions options = default)
+        {
+            StorageResourceContainer localDirectory = new LocalDirectoryStorageResourceContainer(localDirectoryPath);
+
+            StorageResourceContainer blobDirectory = new BlobDirectoryStorageResourceContainer(client, blobDirectoryPrefix, options?.BlobContainerOptions);
```
amnguye, line 47: if `blobDirectoryPrefix == default` should this be making a `BlobStorageResourceContainer` instead of a root directory storage resource container instead?

sdk/storage/Azure.Storage.DataMovement.Blobs/src/BlobContainerClientExtensions.cs
```diff
@@ -0,0 +1,52 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Threading.Tasks;
+using Azure.Storage.DataMovement;
+using Azure.Storage.DataMovement.Blobs;
+
+namespace Azure.Storage.Blobs
```
amnguye, line 9: Should this namespace be `Azure.Storage.Blobs.Specialized` or something?

sdk/storage/Azure.Storage.DataMovement.Blobs/src/BlobContainerClientExtensions.cs
jalauzon-msft: Maybe applies to the upload as well?

sdk/storage/Azure.Storage.DataMovement.Blobs/README.md
```diff
@@ -41,24 +41,7 @@ az storage account create --name MyStorageAccount --resource-group MyResourceGro
...
-// Create BlobTransferManager with event handler in Options bag
-TransferManagerOptions transferManagerOptions = new TransferManagerOptions();
-ContainerTransferOptions options = new ContainerTransferOptions()
-{
-    MaximumTransferChunkSize = 4 * Constants.MB,
-    CreateMode = StorageResourceCreateMode.Overwrite,
-};
-TransferManager transferManager = new TransferManager(transferManagerOptions);
-```
+Data movement authentication is controlled
```
jalauzon-msft, line 44: Was there meant to be more here?

sdk/storage/Azure.Storage.DataMovement.Blobs/src/BlobContainerClientTransferOptions.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Azure.Storage.DataMovement.Blobs;
+using Azure.Storage.DataMovement.Models;
+
+namespace Azure.Storage.Blobs
```
jaschrep-msft, line 7: This should probably be `Azure.Storage.Blobs.Models`. The only options classes we have not in a models namespace is the service-wide ClientOptions class as well as some stuff from Storage.Common.

sdk/storage/Azure.Storage.DataMovement.Blobs/src/BlobContainerClientTransferOptions.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        /// <summary>
+        /// Options pertaining to the blob storage container used in the data transfer.
+        /// </summary>
+        public BlobStorageResourceContainerOptions BlobContainerOptions { get; set; }
+
+        /// <summary>
+        /// Options pertaining to the data tranfer.
+        /// </summary>
+        public ContainerTransferOptions TransferOptions { get; set; }
```
jaschrep-msft, line 23: Are all of these sub-option-bags purely optional parameters?

sdk/storage/Azure.Storage.DataMovement.Blobs/src/BlobContainerClientExtensions.cs
jaschrep-msft: Do we have a reason to hide this in specialized even after someone has already included the datamovement package for their code that is dealing directly with blob clients?

sdk/storage/Azure.Storage.DataMovement.Blobs/README.md
```diff
@@ -81,17 +64,66 @@ We guarantee that all client instance methods are thread-safe and independent of
...
+Upload local directory to the root of the container
+```C# Snippet:ExtensionMethodSimpleUploadToRoot
+DataTransfer transfer = await container.StartUploadDirectoryAsync(localPath);
 
-### Start Upload from Local File to Block Blob
+await transfer.AwaitCompletion();
+```
+Upload local directory to a directory in the container by specifying a directory prefix
+```C# Snippet:ExtensionMethodSimpleUploadToDirectoryPrefix
+DataTransfer transfer2 = await container.StartUploadDirectoryAsync(localPath, blobDirectoryPrefix: Path.GetDirectoryName(localPath));
```
jaschrep-msft, line 83: can we get our sample files that are populating this reorganized a bit so we don't have some samples in the readme using this `transfer2` name? Kinda odd from the reader's perspective.

sdk/storage/Azure.Storage.DataMovement.Blobs/README.md
```diff
@@ -81,17 +64,66 @@ We guarantee that all client instance methods are thread-safe and independent of
...
+```
+Download the entire container to a local directory
+```C# Snippet:ExtensionMethodSimpleDownloadContainer
+DataTransfer transfer = await container.StartDownloadDirectoryAsync(localDirectoryPath);
+
+await transfer.AwaitCompletion();
+```
+Download a directory in the container by specifying a directory prefix
+```C# Snippet:ExtensionMethodSimpleDownloadContainerDirectory
+DataTransfer tranfer2 = await container.StartDownloadDirectoryAsync(localDirectoryPath2, blobDirectoryPrefix);
```
jaschrep-msft, line 95: Can we get our prefix samples to look the same? I like this one that takes some variable named `blobDirectoryPrefix`. The other one having us derive a value with some `Path` functionality seems to clutter the sample.

sdk/storage/Azure.Storage.DataMovement.Blobs/README.md
```diff
@@ -149,6 +181,8 @@ containerTransferOptions.TransferFailed += (TransferFailedEventArgs args) =>
 };
 ```
 
+
+
```
jaschrep-msft, line 185: probably want to remove these newlines?

sdk/storage/Azure.Storage.DataMovement.Blobs/src/BlobContainerClientExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        /// <summary>
+        /// Downloads the contents of a blob container.
+        /// </summary>
+        /// <param name="client">The <see cref="BlobContainerClient"/> used for service operations.</param>
+        /// <param name="localDirectoryPath">The full path to the local directory where files will be dowloaded.</param>
+        /// <param name="blobDirectoryPrefix">Optionally restricts the downloaded content to blobs with the specified directory prefix.</param>
+        /// <param name="options">Options which control the container download.</param>
+        /// <returns></returns>
+        public static async Task<DataTransfer> StartDownloadDirectoryAsync(this BlobContainerClient client, string localDirectoryPath, string blobDirectoryPrefix = default, BlobContainerClientTransferOptions options = default)
```
jaschrep-msft, line 43: Our existing clients in main Storage packages use the phrasing `DownloadTo` for downloads that are written to a destination for you. Perhaps `StartDownloadToDirectoryAsync` is the more consistent API name.

sdk/storage/Azure.Storage.DataMovement.Blobs/src/BlobContainerClientExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        /// <summary>
+        /// Uploads the entire contents of local directory to the blob container.
+        /// </summary>
+        /// <param name="client">The <see cref="BlobContainerClient"/> used for service operations.</param>
+        /// <param name="localDirectoryPath">The full path to the local directory to be uploaded.</param>
+        /// <param name="blobDirectoryPrefix">Optionally specifies the directory prefix to be prepended to all uploaded files.</param>
+        /// <param name="options">Options which control the directory upload.</param>
+        /// <returns>A <see cref="DataTransfer"/> instance which can be used track progress and wait for completion with <see cref="DataTransfer.AwaitCompletion"/>.</returns>
+        public static async Task<DataTransfer> StartUploadDirectoryAsync(this BlobContainerClient client, string localDirectoryPath, string blobDirectoryPrefix = default, BlobContainerClientTransferOptions options = default)
```
jaschrep-msft, line 26: Feels like `blobDirectoryPrefix` should be in the options bag. Storage aggressively tries to avoid optional parameters in the main signature.

sdk/storage/Azure.Storage.DataMovement.Blobs/src/BlobContainerClientExtensions.cs
amnguye: Oh you know what, I didn't realize this was in Azure.Storage.DataMovement.Blobs, my bad I only saw namespace Azure.Storage.Blobs

sdk/storage/Azure.Storage.DataMovement.Blobs/src/BlobContainerClientTransferOptions.cs
amnguye: Yes they should be since they're all optional.

sdk/storage/Azure.Storage.DataMovement.Blobs/src/BlobContainerClientExtensions.cs
amnguye: We wanted the `blobDirectoryPrefix` to not go in the options bag that way customers are well aware that they can upload to a specific directory instead of only at the root of the container.

I think the reasoning behind it is cause potentially customers could upload to the root of the container by accident(?) @schaabs

sdk/storage/Azure.Storage.DataMovement.Blobs/README.md
```diff
@@ -81,17 +64,66 @@ We guarantee that all client instance methods are thread-safe and independent of
...
+Create Instance of TransferManager
+```C# Snippet:CreateTransferManagerSimple
+TransferManager transferManager = new TransferManager(new TransferManagerOptions());
+```
+
+Create Instance of TransferManager with Options
+```C# Snippet:CreateTransferManagerWithOptions
+// Create BlobTransferManager with event handler in Options bag
+TransferManagerOptions transferManagerOptions = new TransferManagerOptions();
+ContainerTransferOptions options = new ContainerTransferOptions()
```
jalauzon-msft, line 110: Just a note that I just merged #35610 which will have some conflicts (be breaking) for this PR now (sorry!). We've combined `ContainerTransferOptions` and `SingleTransferOptions` into a new class `TransferOptions`.

sdk/storage/Azure.Storage.DataMovement.Blobs/src/BlobContainerClientExtensions.cs
schaabs: Fixed. However, it does beg the question do we need two different types for virtual directories and containers? Could the `BlobStorageResourceContainer` have an optional directory prefix, where `null` indicates the base of the container?

sdk/storage/Azure.Storage.DataMovement.Blobs/src/BlobContainerClientTransferOptions.cs
schaabs: Fixed

sdk/storage/Azure.Storage.DataMovement.Blobs/src/BlobContainerClientExtensions.cs
```diff
@@ -0,0 +1,52 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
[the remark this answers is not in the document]
schaabs: I'm not sure I understand where you want to add more new lines.

sdk/storage/Azure.Storage.DataMovement.Blobs/src/BlobContainerClientExtensions.cs
schaabs: Updated

sdk/storage/Azure.Storage.DataMovement.Blobs/src/BlobContainerClientExtensions.cs
schaabs: I've updated this to pull apart into two overloads a simplified one with no options parameter, where `blobDirectoryPrefix` is optional.

```c#
public static Task<DataTransfer> StartUploadDirectoryAsync(this BlobContainerClient client, string localDirectoryPath, string blobDirectoryPrefix = default)
```

Then I included `BlobDirectoryPrefix` on `BlobContainerClientTransferOptions`, and added the second overload which only takes `localDirectoryPath` and `options`, both required parameters.

```c#
public static async Task<DataTransfer> StartUploadDirectoryAsync(this BlobContainerClient client, string localDirectoryPath, BlobContainerClientTransferOptions options)
```

I think this simplifies the use of optional arguments substantially as it eliminates having two optional parameters in the same overload, while still having a convenience method where it's possible just to specify the source and, optionally, target directory. However, if we really wanted to avoid optional parameters all together we could split the first overload into to, resulting in the three overloads.

```c#
public static Task<DataTransfer> StartUploadDirectoryAsync(this BlobContainerClient client, string localDirectoryPath)
public static Task<DataTransfer> StartUploadDirectoryAsync(this BlobContainerClient client, string localDirectoryPath, string blobDirectoryPrefix)
public static async Task<DataTransfer> StartUploadDirectoryAsync(this BlobContainerClient client, string localDirectoryPath, BlobContainerClientTransferOptions options)
```

Thoughts?

amnguye (approved):

sdk/storage/Azure.Storage.DataMovement.Blobs/src/BlobContainerClientExtensions.cs
amnguye: Probably when we go to do `UploadContainer` and stuff we can add those. Since those might come with whatever attributes it could be made with. For now this looks good.

sdk/storage/Azure.Storage.DataMovement.Blobs/src/BlobContainerClientExtensions.cs
amnguye: Looking over the API changes, this looks good. This is currently what's updated in the PR right? I'm going to resolve this convo.

PR #35383, opened 2023-04-05
Add credscan suppression for blob recording

benbp:
The blob recording is getting marked as a credscan false positive for rule CSCAN-AZURE0041 (redis secret).

This may not be necessary, as a re-generation of the recording in https://github.com/Azure/azure-sdk-for-net/pull/35380 fixes the issue as well.


weshaggard (changes requested):
Lets not add this now that they regenerated the blob to not have the pattern

PR #35388, opened 2023-04-06
Drop the double quotes on the ETag serialization to support DNS service behavior

ArthurMa1978:
.Net SDK follow the [RFC](https://www.rfc-editor.org/rfc/rfc2616#section-14.19) that adds double quotes on an ETag string, but unfortunately DNS service doesn't follow the RFC which not accept an ETag with double quotes.

So this PR customize the ETag serialization logic which remove the double  qutoes.


ArthurMa1978:
Service team will fix this.

PR #35391, opened 2023-04-06
Add public Delay and RetryPolicy types

JoshLove-msft:
The Delay/RetryPolicy changes were pulled off https://github.com/Azure/azure-sdk-for-net/pull/33040 so that they can be shipped before implementing the LRO changes. This will allow us to avoid implementing the LRO changes using a project reference of Core in the mgmt libraries. Using a project reference would be tricky because the mgmt libraries have a different release lifecycle.


sdk/core/Azure.Core/src/Pipeline/RetryPolicy.cs
```diff
@@ -14,30 +14,24 @@ namespace Azure.Core.Pipeline
...
+        /// </summary>
+        protected Delay Delay { get; }
 
         /// <summary>
         /// Initializes a new instance of the <see cref="RetryPolicy"/> class.
         /// </summary>
-        /// <param name="options">The set of options to use for configuring the policy.</param>
-        protected RetryPolicy(RetryOptions? options = default)
+        /// <param name="maxRetries">The maximum number of retries to attempt.</param>
+        /// <param name="delay">The delay to use for computing the interval between attempts.</param>
```
christothes, line 30:
```suggestion
        /// <param name="delay">The delay to use for computing the interval between retry attempts.</param>
```

sdk/core/Azure.Core/src/Pipeline/RetryPolicy.cs
```diff
@@ -14,30 +14,24 @@ namespace Azure.Core.Pipeline
...
-        private readonly TimeSpan _maxDelay;
         private readonly int _maxRetries;
 
-        private readonly Random _random = new ThreadSafeRandom();
-
-        private const string RetryAfterHeaderName = "Retry-After";
-        private const string RetryAfterMsHeaderName = "retry-after-ms";
-        private const string XRetryAfterMsHeaderName = "x-ms-retry-after-ms";
+        /// <summary>
+        /// Gets the delay to use between retries.
```
christothes, line 22:
```suggestion
        /// Gets the delay to use for computing the interval between retry attempts.
```

christothes (commented):
Just some suggested doc tweaks

christothes (approved):

m-nash (approved):

PR #35410, opened 2023-04-06
Remove GetNextDelay from RetryPolicy

JoshLove-msft:


sdk/core/Azure.Core/src/Pipeline/RetryPolicy.cs
```diff
@@ -21,7 +21,7 @@ public class RetryPolicy : HttpPipelinePolicy
         /// <summary>
         /// Gets the delay to use for computing the interval between retry attempts.
         /// </summary>
-        protected Delay Delay { get; }
+        private readonly Delay _delay;
```
annelo-msft, line 24: Is the reason we make this no longer protected because the only place subclasses would have accessed it was in the GetNextDelay method?

JoshLove-msft: Yep, it's no longer useful for implementers as they cannot influence the delay that is returned besides the delay instance they pass in.

annelo-msft (approved):
I think it makes sense to remove these methods until we understand better how people would use the RetryPolicy.GetNextDelay differently  - esp. since there's some possible confusion about how it would interact with the Delay implementation passed in via the constructor.

sdk/core/Azure.Core/src/Pipeline/RetryPolicy.cs
annelo-msft: Cool - that's makes sense! + a nice additional simplification of the public API surface :)

PR #35433, opened 2023-04-07
[Messaging] README Updates

jsquire:
# Summary

The focus of these changes is to tweak the content and format of the Event Hubs README snippets to clarify lifetime guidance explicitly in each example and to add an illustration of registering with ASP.NET Core DI.  Tweaked the layout of the Service Bus examples, moving the ASP.NET Core section to an example to streamline the "Authenticating the client" section and focus on the core "Getting Started" scenario.

### DIFF
--- a/sdk/eventhub/Azure.Messaging.EventHubs/src/ApiCompatBaseline.txt
+++ b/sdk/eventhub/Azure.Messaging.EventHubs/src/ApiCompatBaseline.txt
@@ -1,2 +1,0 @@
-# Baselining these until the next GA. This is not a compatibility issue; the impacted member was inappropriately tagged as non-visible and provides value for customers being discoverable.
-CannotRemoveAttribute : Attribute 'System.ComponentModel.EditorBrowsableAttribute' exists on 'Azure.Messaging.EventHubs.Consumer.EventPosition.ToString()' in the contract but not the implementation.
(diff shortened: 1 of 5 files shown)


sdk/servicebus/Azure.Messaging.ServiceBus/README.md
```diff
@@ -56,41 +56,9 @@ await using var client = new ServiceBusClient(connectionString);
...
-To use the preceding code, add this to your configuration:
-
-```json
-{
-  "ConnectionStrings": {
-    "ServiceBus": "<connection_string>"
-  }
-}
-```
+For examples of how to authenticate for an ASP.NET Core application, view this [example](#registering-with-aspnet-core-dependency-injection).
```
jsquire, line 59: @JoshLove-msft :  The "how do I register for DI with a credential" came up twice this week in support issues.   When I went to add the snippet, it felt large and like it didn't quite fit into "authorize the client".   

I tried moving it down to an example scenario and leaving a note here.   If you don't like it, I can move things back.   Thoughts?

sdk/eventhub/Azure.Messaging.EventHubs/src/ApiCompatBaseline.txt
```diff
@@ -1,2 +0,0 @@
-# Baselining these until the next GA. This is not a compatibility issue; the impacted member was inappropriately tagged as non-visible and provides value for customers being discoverable.
```
JoshLove-msft, line 1: Interesting.. is this not needed anymore?

JoshLove-msft (approved):

sdk/eventhub/Azure.Messaging.EventHubs/src/ApiCompatBaseline.txt
jsquire: I'm hoping not.  It was supposed to just be needed for 1-2 versions and it's been hanging around for a year+.

m-redding (approved):

PR #35434, opened 2023-04-07
Rename to DelayStrategy

JoshLove-msft:


sdk/core/Azure.Core/api/Azure.Core.net461.cs
```diff
@@ -1035,7 +1035,7 @@ public sealed partial class RedirectPolicy : Azure.Core.Pipeline.HttpPipelinePol
     }
     public partial class RetryPolicy : Azure.Core.Pipeline.HttpPipelinePolicy
     {
-        public RetryPolicy(int maxRetries = 3, Azure.Core.Delay? delay = null) { }
+        public RetryPolicy(int maxRetries = 3, Azure.Core.DelayStrategy? delayStrategy = null) { }
```
annelo-msft, line 1038: nit: keep parameter name `delay`?  Totally your call.

sdk/core/Azure.Core/src/ExponentialDelayStrategyInternal.cs
```diff
@@ -5,11 +5,11 @@
 
 namespace Azure.Core
 {
-    internal class ExponentialDelay : Delay
+    internal class ExponentialDelayStrategyInternal : DelayStrategy
```
annelo-msft, line 8: Curious: why suffix the name with `Internal`?

JoshLove-msft: There is a whole mess with the internal delay strategy types used for LROs - this is a temporary rename until we can fix the LROs once the public DelayStrategy type is shipped.

sdk/core/Azure.Core/api/Azure.Core.net461.cs
JoshLove-msft: Oooh - good catch. No strong opinion here, but the other parameters that use `delay` are for TimeSPans.

sdk/core/Azure.Core/src/Shared/DelayStrategyInternal.cs
```diff
@@ -10,7 +10,7 @@ namespace Azure.Core
     /// <summary>
     /// Strategy to control delay behavior.
     /// </summary>
-    internal abstract class DelayStrategy
+    internal abstract class DelayStrategyInternal
```
annelo-msft, line 13: Curious: why do we have this class and the public class?

JoshLove-msft: These are internal types that will be switched over after the new type is shipped.

annelo-msft: I'm afraid I don't follow - can you help me understand why the public types can't be used now?

JoshLove-msft: Because this would require adding a project reference to Core in every mgmt library which we cannot do.

sdk/core/Azure.Core/tests/OperationInternalOfTTests.cs
```diff
@@ -511,7 +511,7 @@ public TestOperation(Func<bool, CancellationToken, ValueTask<OperationState<int>
             public ValueTask<OperationState<int>> UpdateStateAsync(bool async, CancellationToken cancellationToken) => _updateStateAsyncHandler(async, cancellationToken);
         }
 
-        private class CallCountStrategy : DelayStrategy
```
annelo-msft, line 514: Curious: would it make sense to validate that Operation will work with the new DelayStrategy implementation in tests?

JoshLove-msft: Yes! will do this once we switch over to the new implementation.

sdk/core/Azure.Core/src/Shared/DelayStrategyInternal.cs
JoshLove-msft: https://github.com/Azure/azure-sdk-for-net/issues/35437

sdk/core/Azure.Core/tests/OperationInternalOfTTests.cs
JoshLove-msft: https://github.com/Azure/azure-sdk-for-net/issues/35437

sdk/core/Azure.Core/src/ExponentialDelayStrategyInternal.cs
JoshLove-msft: https://github.com/Azure/azure-sdk-for-net/issues/35437

sdk/core/Azure.Core/tests/OperationInternalOfTTests.cs
JoshLove-msft: To be fair, this is already done in https://github.com/Azure/azure-sdk-for-net/pull/33040

sdk/core/Azure.Core/src/Shared/DelayStrategyInternal.cs
annelo-msft: Sorry, I'm still a little confused -if we change the name of this file, would we need to update the management plane libraries that depend on it, like is happening in the maps, etc, project files below?  

I'm also wondering, if we're planning to replace this internal type, would it make sense to have this type inherit from the final type to validate that functionality, rather than having them be independent?

JoshLove-msft: Well this file will be deleted, so we would need to remove the reference from any csprojs that reference it.

We can't have this type inherit from the final type because the signatures are different. But the functionality has already been validated in https://github.com/Azure/azure-sdk-for-net/pull/33040

KrzysztofCwalina (approved):

sdk/core/Azure.Core/src/ExponentialDelayStrategyInternal.cs
annelo-msft: Discussed offline.

sdk/core/Azure.Core/src/Shared/DelayStrategyInternal.cs
annelo-msft: Discussed offline.

sdk/core/Azure.Core/tests/OperationInternalOfTTests.cs
annelo-msft: Discussed offline.  Key point is that in https://github.com/Azure/azure-sdk-for-net/pull/33040, the call sites are updated to address the signature changes across types.

PR #35440, opened 2023-04-07
Place arguments from the token provider file first.

NCarlsonMSFT:
This allows for using a framework dependent dotnet app.


christothes (commented):
Could we add a test for this change?

christothes (approved):

PR #35445, opened 2023-04-10
Fix the ci definition for the Azure.ResourceManager.PaloAltoNetworks.Ngfw

ArthurMa1978:


ArcturusZhang (approved):
Could you please check the api file of this RP? Last time, this RP does not have an api file

PR #35451, opened 2023-04-10
Move event processor config from .github/workflows into .github

JimSuplizio:
There are two changes here:
1. Move the event-processor.config file from .github/workflows to .github. This is being done in preparation for managing the event yml files from a central location (azure-sdk-tools) without changing the repository specific config file.
2. Remove the now defunct fabricbot file.


jsquire (approved):

JimSuplizio:
I have to override check-enforcer, there's no pipeline or anything that will test these changes.

PR #35453, opened 2023-04-10
[Text Analytics] Implement `IsEnvironmentReadyAsync`

### LINKED ISSUE #35452: [Text Analytics] Fix tests failing with "401 (PermissionDenied)"
This exclusively affects tests running against the dynamic resource. The root cause is probably that tests start running before the resource has been fully created and stabilized.

```text
Azure.RequestFailedException : Access denied due to invalid subscription key or wrong API endpoint. Make sure to provide a valid key for an active subscription and use a correct regional API endpoint for your resource.
Status: 401 (PermissionDenied)
ErrorCode: 401
```

joseharriaga:
Resolves: https://github.com/Azure/azure-sdk-for-net/issues/35452.

Implemented `IsEnvironmentReadyAsync` with a simple request to the dynamic resource to confirm that permissions have propagated successfully, because otherwise we're at risk of having a bunch of test failures like this one:

```text
Azure.RequestFailedException : Access denied due to invalid subscription key or wrong API endpoint. Make sure to provide a valid key for an active subscription and use a correct regional API endpoint for your resource.
Status: 401 (PermissionDenied)
ErrorCode: 401
```

Additionally, while testing, I found another instance of the transient Internal Server Error. The difference is that the message is simply "Internal Server Error." while the previous known case said "Failed to process task after several retry":

```text
Azure.RequestFailedException : Internal Server Error.
Status: 200 (OK)
ErrorCode: InternalServerError

Additional Information:
Target: #/tasks/items/1
error-0: InternalServerError: Internal Server Error.
```

I added this one to the `ShouldRetry` method of `RetryOnInternalServerErrorAttribute`.

### DIFF
--- a/sdk/textanalytics/Azure.AI.TextAnalytics/tests/RetryOnInternalServerErrorAttributeTests.cs
+++ b/sdk/textanalytics/Azure.AI.TextAnalytics/tests/RetryOnInternalServerErrorAttributeTests.cs
@@ -14,18 +14,24 @@ namespace Azure.AI.TextAnalytics.Tests
     public class RetryOnInternalServerErrorAttributeTests
     {
         private const string FailedToProcessTaskExceptionMessage =
-            "Failed to process task after several retry,"
-            + " Status: 200 (OK),"
-            + " ErrorCode: InternalServerError";
+            "Failed to process task after several retry"
+            + ", Status: 200 (OK)"
+            + ", ErrorCode: InternalServerError";
 
         private const string InvalidTaskTypeExceptionMessage =
-            "Invalid Task Type,"
-            + " Status: 500 (Internal Server Error),"
-            + " ErrorCode: InternalServerError";
+            "Invalid Task Type"
+            + ", Status: 500 (Internal Server Error)"
+            + ", ErrorCode: InternalServerError";
+
+        private const string InternalServerErrorExceptionMessage =
+            "Internal Server Error."
+            + ", Status: 200 (OK)"
+            + ", ErrorCode: InternalServerError";
 
         [Test]
         [TestCase(FailedToProcessTaskExceptionMessage)]
         [TestCase(InvalidTaskTypeExceptionMessage)]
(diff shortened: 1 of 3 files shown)


sdk/textanalytics/Azure.AI.TextAnalytics/tests/RetryOnInternalServerErrorAttributeTests.cs
```diff
@@ -14,18 +14,24 @@ namespace Azure.AI.TextAnalytics.Tests
     public class RetryOnInternalServerErrorAttributeTests
     {
         private const string FailedToProcessTaskExceptionMessage =
-            "Failed to process task after several retry,"
-            + " Status: 200 (OK),"
-            + " ErrorCode: InternalServerError";
+            "Failed to process task after several retry"
```
heaths, line 17: "With great power comes great responsibility." :)

Have you been letting the service team know about these issues? When I initially did something like this with Key Vault, it was after effectively triaging the sporadic issues we were finding and investigations on their side.

heaths (approved):

deyaaeldeen (approved):
Looks good, however, the issue seems to be not TA specific. Should the pipeline add some latency after the resource creation step to mitigate it?

heaths:
> Looks good, however, the issue seems to be not TA specific. Should the pipeline add some latency after the resource creation step to mitigate it?

We don't want to add latency to all pipelines. Few, hopefully, need it. See my review for what I suggested to inject latency when needed.

mssfang (approved):

sdk/textanalytics/Azure.AI.TextAnalytics/tests/RetryOnInternalServerErrorAttributeTests.cs
joseharriaga: Yes, I filed bugs for each of these and provided all the information that I have.

PR #35460, opened 2023-04-10
TransferValidation Improvements: OpenWrite

jaschrep-msft:
`StorageWriteStream` now keeps a running calculation of the bytes in its buffer as they are copied into the buffer, ensuring earliest possible accounting of data.

Bug fix for keeping a running calculation of an MD5. Was not an issue previously as MD5 was only ever calculated all at once before now.

### DIFF
--- a/sdk/storage/Azure.Storage.Common/src/Shared/HashAlgorithmHasher.cs
+++ b/sdk/storage/Azure.Storage.Common/src/Shared/HashAlgorithmHasher.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.IO;
-using System.IO.Hashing;
 using System.Security.Cryptography;
 
 namespace Azure.Storage
@@ -15,7 +14,7 @@ namespace Azure.Storage
     {
         private readonly HashAlgorithm _hashAlgorithm;
 
-        public int HashSize => _hashAlgorithm.HashSize;
+        public int HashSizeInBytes => _hashAlgorithm.HashSize >> 3;
 
         public HashAlgorithmHasher(HashAlgorithm hashAlgorithm)
         {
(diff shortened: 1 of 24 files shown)


sdk/storage/Azure.Storage.Common/src/Shared/HashAlgorithmHasher.cs
```diff
@@ -15,7 +14,7 @@ internal class HashAlgorithmHasher : IHasher
     {
         private readonly HashAlgorithm _hashAlgorithm;
 
-        public int HashSize => _hashAlgorithm.HashSize;
+        public int HashSizeInBytes => _hashAlgorithm.HashSize >> 3;
```
amnguye, line 17: Curious why we shift by 3? Should this be in our constants?

sdk/storage/Azure.Storage.Common/src/Shared/StorageWriteStream.cs
```diff
@@ -35,15 +37,19 @@ internal abstract class StorageWriteStream : Stream
...
+            if (!transferValidation.PrecalculatedChecksum.IsEmpty)
             {
                 throw Errors.PrecalculatedHashNotSupportedOnSplit();
             }
 
             if (buffer != null)
             {
+                if (buffer.Position != 0)
+                {
+                    throw new ArgumentException("Buffer must be empty if provided.", nameof(buffer));
```
amnguye, line 51: Add this to `Errors`

amnguye (approved):

sdk/storage/Azure.Storage.Common/src/Shared/HashAlgorithmHasher.cs
jaschrep-msft: `x >> 3` == `x / 8`. We are just converting from bits to bytes. `HashAlgorithm.HashSize` is in bits for some reason. We convert this to bytes to be consistent with modern .NET tooling. This would probably be clearer if the .NET api used `HashSizeInBits` instead. I can leave a comment to clarify but asking for a constant is asking for `const int BitsInByte = 8` except now we have an even more unwieldy name like `BitToByteBitshiftValue` or something awful when if we just communicate we are converting units it will become very clear.

PR #35499, opened 2023-04-11
[AzureMonitorExporter] update dependent projects for Exporter

TimothyMothra:
Azure.Monitor.OpenTelemetry.Exporter.Perf has a dependency on Azure.Test.Perf.
I added that project to our solution.

image


rajkumar-rangaraj (approved):

reyang:
Curious, why do we want to do this?

TimothyMothra:
> Curious, why do we want to do this?

Our Perf project has a dependency on this project and will fail to build if the Azure.Test.Perf is not also built.

reyang:
> > Curious, why do we want to do this?
> 
> Our Perf project has a dependency on this project and will fail to build if the Azure.Test.Perf is not also built.

What if Azure.Test.Perf later added another dependency, are you going to keep maintaining the dependency chain?

reyang (changes requested):
Need to understand the direction https://github.com/Azure/azure-sdk-for-net/pull/35499#issuecomment-1503918461.

reyang:
> > What if Azure.Test.Perf later added another dependency, are you going to keep maintaining the dependency chain?
> 
> At the moment, yes. :(
> 
> This could be mitigated if Azure.Test.Perf was published and we could take a PackageDependency. But for as long as we have ProjectDependencies we can have issues like this.

Would you investigate what's the right approach? I suggest that we avoid any hacks that require ongoing obligations (e.g. "are you going to keep maintaining the dependency chain - yes" is something I'm very concerned about).

PR #35522, opened 2023-04-12
Abhishesingh msft/cont send dtmf updates

abhishesingh-msft:
Improvements in Continuous Dtmf & Send Dtmf
* remove Options pattern for sendDtmf request
* generate new model from swagger changes
* support ContinuousDtmfRecognitionStopped event
* adjust tests


bharat-kalyan-namburi (approved):

sdk/communication/Azure.Communication.CallAutomation/src/CallMedia.cs
```diff
@@ -463,18 +471,5 @@ public virtual Response StopContinuousDtmfRecognition(CommunicationIdentifier ta
                 throw;
             }
         }
-
-        private static SendDtmfRequestInternal CreateSendDtmfRequest(SendDtmfOptions sendDtmfOptions)
```
minwoolee-msft, line 467: Question: is there a reason why option class have been removed?

sdk/communication/Azure.Communication.CallAutomation/src/CallMedia.cs
```diff
@@ -409,17 +409,21 @@ public virtual Response StopContinuousDtmfRecognition(CommunicationIdentifier ta
...
         /// </summary>
-        /// <param name="sendDtmfOptions">Configuration attributes for SendDtmf.</param>
+        /// <param name="targetParticipant">Target participants for start continuous Dtmf Recognition.</param>
+        /// <param name="tones">List of Tones to be sent.</param>
+        /// <param name="operationContext"></param>
         /// <param name="cancellationToken"></param>
         /// <returns></returns>
-        public virtual async Task<Response<SendDtmfResult>> SendDtmfAsync(SendDtmfOptions sendDtmfOptions,
-            CancellationToken cancellationToken = default)
+        public virtual async Task<Response<SendDtmfResult>> SendDtmfAsync(CommunicationIdentifier targetParticipant, IReadOnlyList<DtmfTone> tones,
```
minwoolee-msft, line 417: for input, we want IEnumerable<T> for parameters.

See comments from Azure Review Board:
> We prefer IEnumerable<T> for parameters and IList<T> or IReadOnlyList<T> for properties

sdk/communication/Azure.Communication.CallAutomation/src/CallMedia.cs
abhishesingh-msft: As per [this](https://github.com/Azure/azure-sdk-for-net/pull/35242/files/192632cec82e3025d2f14dd0fb3fbac9446f7493#r1153796703) review comment from last PR, we removed options from startContinuousDtmfRecognition(), so for consistency it was thought to remove options from sendDtmf as well and this [workitem](https://skype.visualstudio.com/SPOOL/_workitems/edit/3224670) was created to remove options. Additionlly, we don't expect the arguments to grow hence we kept the method simple at this point without options.

If you suggest for options, we will add in next pr as this change will require a contract and a PMA change also.

sdk/communication/Azure.Communication.CallAutomation/src/CallMedia.cs
abhishesingh-msft: thank you, I have changed now.

sdk/communication/Azure.Communication.CallAutomation/src/CallMedia.cs
minwoolee-msft: Ah ok - just curiosity, is OperationContext required field here? for all other operations, this is optional value

abhishesingh-msft: It is not...instead of overloading another method, I have mentioned in method doc to pass null, if not applicable

minwoolee-msft: If it is optional value, we would have signature like this 
..., OperationContext = null, ...

This sets default value for OperationContext, and if the devs never pass this value in the first place, SDK will set it as default value. This has few benefits:

1. It clearly shows in signature that which values are required and which values are not required
2. Does not force the devs to pass null when devs do not care for this value
3. Without reading documentation, it gives devs idea that this value's default value is null
4. This aligns with all other methods in our SDK

By making OperationContext optional, would it break your change? if not, I suggest we make this value optional.

abhishesingh-msft: Sure, done...Thank you

minwoolee-msft (approved):

bharat-kalyan-namburi (approved):

sdk/communication/Azure.Communication.CallAutomation/src/CallAutomationClient.cs
```diff
@@ -83,10 +83,10 @@ private CallAutomationClient(Uri endpoint, HttpPipeline httpPipeline, CallAutoma
             _pipeline = httpPipeline;
             _resourceEndpoint = endpoint.AbsoluteUri;
             _clientDiagnostics = new ClientDiagnostics(options);
-            AzureCommunicationServicesRestClient = new AzureCommunicationServicesRestClient(_clientDiagnostics, httpPipeline, endpoint.AbsoluteUri, options.ApiVersion);
```
abhishesingh-msft, line 86: @minwoolee-msft this is the change which we discussed, resulted from "Endpoint"."format": "url" in swagger from [commit](https://skype.visualstudio.com/SCC/_git/sync_calling_concore-platform/commit/673e805f4502668a6baf4d765ae23b9dbdf70494?path=/Microsoft.Skype.Platform/Microsoft.Skype.Platform.ExecutionAgent/Azure.Communication.Service.ServerCalling/Swagger/ServerCallDocumentFilter.cs)

